Issue #531, opened 2014-08-11
clusterhq-release RPM doesn't set `gpgcheck=0`, and the resulting repository is therefore unuseable


itamarst:
Also I'm not sure how this was generated and therefore how to update it.

Issue #510, opened 2014-08-08
It's not possible to install `flocker-node` etc from our release repository in our Vagrant image


itamarst:
In our Vagrant image you can do:

```
$ yum localinstall -y http://archive.clusterhq.com/fedora/clusterhq-release$(rpm -E %dist).noarch.rpm
$ repoquery --repoid clusterhq '*'                                                                                                                                                                 
geard-0:0-0.12.1.1.git4f8177d.fc20.x86_64
python-characteristic-0:0.1.0-3.fc20.noarch
python-eliot-0:0.4.0-2.fc20.noarch
python-idna-0:0.8-1.fc20.noarch
python-netifaces-0:0.10.4-1.fc20.x86_64
python-service-identity-0:1.0.0-2.fc20.noarch
python-treq-0:0.2.1-1.fc20.noarch
python-twisted-0:12.2.0-3.fc20.noarch
python-twisted-0:14.0.0-2.fc20.x86_64
```

Notice how the flocker packages aren't listed. But on Ubuntu:

```
$ repoquery --repoid clusterhq --repofrompath clusterhq,http://archive.clusterhq.com/fedora/20/x86_64/ "*"
Not using downloaded repomd.xml because it is older than what we have:
  Current   : Fri Aug  8 13:05:00 2014
  Downloaded: Wed Jul 16 14:16:18 2014
flocker-cli-0:0.0.6-1.fc20.noarch
flocker-node-0:0.0.6-1.fc20.noarch
geard-0:0-0.12.1.1.git4f8177d.fc20.x86_64
python-characteristic-0:0.1.0-3.fc20.noarch
python-eliot-0:0.4.0-2.fc20.noarch
python-flocker-0:0.0.6-1.fc20.noarch
python-idna-0:0.8-1.fc20.noarch
python-netifaces-0:0.10.4-1.fc20.x86_64
python-service-identity-0:1.0.0-2.fc20.noarch
python-treq-0:0.2.1-1.fc20.noarch
python-twisted-0:12.2.0-3.fc20.noarch
python-twisted-0:14.0.0-2.fc20.x86_64
```

I really have no idea what's going on.

PR #532, opened 2014-08-11
Fedora server install instructions that might actuall work - fixes 521

### LINKED ISSUE #521: Instructions for installing flocker-node on Fedora do not work
Due to #510 I expect that our instructions for installing `flocker-node` on Fedora will fail. Unless we fix that for release we should replace current instructions with instructions that include the full path of the RPMs, as #511 did for tutorial Vagrant setup.

itamarst:
Fixes #521.

Motivated by #510 and #531.


exarkun:
Fedora support is long gone and these instructions are now significantly out of date anyway.  Deleting the corresponding branch.

PR #1016, opened 2014-12-11
[FLOC-660] Change Homebrew script to not conflict with Python

adamtheturtle:
Design for https://clusterhq.atlassian.net/browse/FLOC-660

I think that the Homebrew instructions for Python packages have been updated since I last looked.
See https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Formula-Cookbook.md#specifying-gems-python-modules-etc-as-dependencies.

Following their instructions solved this problem and is therefore the design I went for.

This design includes a `flocker.rb` file which would be in the `homebrew-tap` repository versioned as the next release version.
I started to make this file by copying the `jrnl` example linked from the Homebrew docs (https://github.com/Homebrew/homebrew/blob/master/Library/Formula/jrnl.rb).
I then used the `mkpydeps` script linked from the Homebrew docs to create the resource stanzas.
This does not look at specified versions and gets the resource URLs and sha1s of the latest versions of each dependency.
Therefore, I manually looked at each resource and compared it to its equivalent in Flocker's `setup.py`.
If there was an `==` equivalent in `setup.py` I replaced the URL and sha1.

This can be tested on OS X by running:
```
brew install --verbose --debug flocker.rb
brew test flocker.rb
```

You can do this with `brew install python` done before, after or not at all.
`flocker-deploy` should not already be on your PATH.

`brew uninstall flocker` afterwards to clean up.

I'd like a look-over / design review before I continue, in case people are not happy with the final result of my planned instructions (flocker.rb).

If this is suitable, I will write instructions to do what I explained above, and a follow up issue to improve the `mkpydeps` script to account for versions if possible.
This could potentially lead to the whole Homebrew script being generated automatically given a version number and/or a download link.

It may be useful to compare this to the latest script we ship https://raw.githubusercontent.com/ClusterHQ/homebrew-tap/master/flocker-0.3.2.rb.


adamtheturtle:
I think that the cost of automating the whole script creation process (that is https://clusterhq.atlassian.net/browse/FLOC-1150, perhaps without the CI) is so low compared to the cost of doing it manually that that should just be done now. I have emailed @tdsmith asking for a permissive licence to be added to `mkpydeps`. Else we can re-implement its functionality.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -324,33 +324,65 @@ Release
...
-          ed03a154c2fdcd19eca471c0e22925cf0d3925fb  dist/Flocker-0.1.2.tar.gz
+     Update the ``sha1`` checksum. Retrieve it with ``sha1sum``:
 
-     - Commit the changes and push:
+     .. code-block:: console
+
+         sha1sum "dist/Flocker-${VERSION}.tar.gz"
+         ed03a154c2fdcd19eca471c0e22925cf0d3925fb  dist/Flocker-0.1.2.tar.gz
+
+     Download and run the `mkpydeps`_ script.
```
adamtheturtle: Apart from the specific instruction `mkpydeps flocker`, everything is covered by the above step:

```
#. Create (if necessary) and activate the Flocker release virtual environment:

   .. note:: The following instructions use `virtualenvwrapper`_ but you can use `virtualenv`_ directly if you prefer.

   .. code-block:: console

      mkvirtualenv flocker-release-${VERSION}
      pip install --editable .[release]
```

Note that `setup.py` 's release section has added the new dependency.

adamtheturtle:
The latest changes cover some of the https://clusterhq.atlassian.net/browse/FLOC-1150 (originally created as feedback to review comments on this PR).

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -324,33 +324,65 @@ Release
...
-       .. code-block:: console
+     The version number is also included in the ``url`` part of the recipe and should be changed as appropriate.
 
-          sha1sum "dist/Flocker-${VERSION}.tar.gz"
-          ed03a154c2fdcd19eca471c0e22925cf0d3925fb  dist/Flocker-0.1.2.tar.gz
+     Update the ``sha1`` checksum. Retrieve it with ``sha1sum``:
 
-     - Commit the changes and push:
+     .. code-block:: console
+
```
adamtheturtle: I'm now confused about where this is now (confusing diff).

setup.py
```diff
@@ -126,7 +126,7 @@ def run(self):
 
         "psutil == 2.1.2",
         "netifaces >= 0.8",
-        "ipaddr == 2.1.10",
+        "ipaddr == 2.1.11",
```
adamtheturtle: https://pypi.python.org/pypi/ipaddr/2.1.10 doesn't have a pypi URL.

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python
+
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Create a Homebrew recipe for Flocker, using the VERSION environment variable.
+
+Invoke using e.g.:
+  $ export VERSION=0.1.2
+  $ ./make_homebrew_recipe.py > > flocker-${VERSION}.rb
```
adamtheturtle: Typo (double >).

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+    class_name = class_name.replace(disallowed_character, '')
+
+resources = []
+graph = Graph()
+graph.from_specifications(["flocker"])
+dependencies = []
+for name, node in sorted(graph.iteritems()):
+    if name == "flocker":
+        continue
+    else:
```
adamtheturtle: No need for this to be in `else`.

adamtheturtle:
@tdsmith kindly added the MIT licence to https://github.com/tdsmith/labmisc.

setup.py
adamtheturtle: A concern I have is about what happens if a package stops being served by PyPI while we depend on it. I suggested serving our own packages.

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+
+resources = []
+graph = Graph()
+graph.from_specifications(["flocker"])
+dependencies = []
+for name, node in sorted(graph.iteritems()):
+    if name == "flocker":
+        continue
+
+    requirement = node.dist.as_requirement()
```
adamtheturtle: This is how `tl.eggdeps` gets the versions for `print_list(graph, Options(version_numbers=True))` which has been mentioned in the comments.

setup.py
exarkun: > https://pypi.python.org/pypi/ipaddr/2.1.10 doesn't have a pypi URL.

This is confusing.  That **is** a PyPI URL?

adamtheturtle: The JSON release url metadata is not available for this release https://pypi.python.org/pypi/ipaddr/2.1.10/json - see "2.1.10" and "2.1.11" under "releases".

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+    operator, version = requirement.specs[0]
+    project_name = requirement.project_name
+    dependencies.append(project_name)
+    url = "http://pypi.python.org/pypi/{name}/{version}/json".format(
+            name=project_name,
+            version=version)
```
wallrj: pyflakes says `make_homebrew_recipe.py:48:13: E126 continuation line over-indented for hanging indent`

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+
+url = ("http://storage.googleapis.com/archive.clusterhq.com/"
+       "downloads/flocker/Flocker-{version}.tar.gz").format(version=version)
+tarball = urlopen(url)
+checksum = sha1(tarball.read()).hexdigest()
+tarball.close()
+
+class_name = 'Flocker' + version
+for disallowed_character in ['-', '.']:
+    class_name = class_name.replace(disallowed_character, '')
```
wallrj: I think you can use `translate` instead. 
```
In [14]: 'foo-bar.baz'.translate(None, '.-')
Out[14]: 'foobarbaz'
```

Although it's not as readable.

```
In [11]: x.translate?
Type:       builtin_function_or_method
String Form:<built-in method translate of str object at 0x22e3810>
Docstring:
S.translate(table [,deletechars]) -> string

Return a copy of the string S, where all characters occurring
in the optional argument deletechars are removed, and the
remaining characters have been mapped through the given
translation table, which must be a string of length 256 or None.
If the table argument is None, no translation is applied and
the operation simply removes the characters in deletechars.
```

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+
+resources = []
+graph = Graph()
+graph.from_specifications(["flocker"])
+dependencies = []
+for name, node in sorted(graph.iteritems()):
+    if name == "flocker":
```
wallrj: The thing I noticed when experimenting with this, is that `pip freeze` reports our package name as `Flocker`. So perhaps it'd be worth comparing against the lowercase name.

```
(660-test)(homebrew-failure-FLOC-660 ? =)[~/.../HybridLogic/flocker]$ pip freeze
Flocker==0.3.2-653-g23aa1fc
...
```

admin/make_homebrew_recipe.py
wallrj: Which comments? I can't see any in the code. Perhaps you should add this note as an explanatory comment.

setup.py
wallrj: I think the reason is that `ipaddr` used to link to their external google code download URL...which is frowned upon. 2.1.11 seems to be the first version where they've started properly hosting the package on PyPI and *properly* updating their metadata.

Here's a comment on the subject:

> There are some more notable names in the "unsafe" lists, but a few
spot checks on projects like PyGObject, PyGTK, biopython, dbus-python,
django-piston, ipaddr, matplotlib, and mayavi showed that a number of
them *have* switched to PyPI hosting for recent releases, but have
left older releases as externally hosted. (A few notable names, like
wxPython and Spyder, *did* show up as genuinely externally hosted.
Something that would be nice to be able to do, but isn't really
practical without a server side dependency graph, is to be able to
figure out how many packages have an externally hosted dependency
*somewhere in their dependency chain*, and *how many* other projects
are depending on particular externally hosted projects transitively).

https://mail.python.org/pipermail/distutils-sig/2014-May/024238.html

So I think you're doing the right thing by incrementing the version number here.

wallrj: Also https://gist.github.com/dstufft/5088915#file-missing-json-L6553

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -324,33 +323,19 @@ Release
...
-
-   - Update recipe file:
-
-     - Update the version number:
-
-       The version number is included in the class name with all dots and dashes removed.
-       e.g. ``class Flocker012 < Formula`` for Flocker-0.1.2
-
-     - Update the URL:
+   - Create a recipe file using a Flocker admin script:
```
wallrj: Add a note explaining how to install the flocker admin tools `pip install Flocker[release]`

admin/make_homebrew_recipe.py
```diff
@@ -0,0 +1,97 @@
...
+from urllib2 import urlopen
+from hashlib import sha1
+from tl.eggdeps.graph import Graph
+
+version = environ.get("VERSION")
+if version is None:
+    raise Exception("Set the VERSION environment variable.")
+
+url = ("http://storage.googleapis.com/archive.clusterhq.com/"
+       "downloads/flocker/Flocker-{version}.tar.gz").format(version=version)
```
wallrj: This should be an `https` link.

wallrj: And ideally (I think) we'd just supply the Flocker sdist package as an argument to this script. 
The locally built package is the "known good" package. It just happens to be uploaded to google storage in our current process. 
`make_homebrew_recipe.py Flocker-0.3.3.tgz`

wallrj: There should at least be some validation and error reporting when the supplied sdist URL / file is invalid. I got 
```
$ VERSION=0.3.3 ./admin/make_homebrew_recipe.py 
  File "./admin/make_homebrew_recipe.py", line 27, in <module>
    tarball = urlopen(url)
  File "/usr/lib64/python2.7/urllib2.py", line 127, in urlopen
    return _opener.open(url, data, timeout)
  File "/usr/lib64/python2.7/urllib2.py", line 410, in open
    response = meth(req, response)
  File "/usr/lib64/python2.7/urllib2.py", line 523, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib64/python2.7/urllib2.py", line 448, in error
    return self._call_chain(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 382, in _call_chain
    result = func(*args)
  File "/usr/lib64/python2.7/urllib2.py", line 531, in http_error_default
    raise HTTPError(req.get_full_url(), code, msg, hdrs, fp)
urllib2.HTTPError: HTTP Error 404: Not Found

```

Which didn't help much.

wallrj:
Thanks @adamtheturtle 

I think this looks good:

I suggest:
 * some command line help (`twisted.python.usage`)
 * script should be moved to a module leaving `./admin/bin/make_homebrew_script` as a stub script (in line with the other admin tools)
 * it should take a positional argument containing the path to a local Flocker sdist package (I think that's what @tomprince would prefer if we get round to running this automatically from Buildbot build artifacts.)
 * Split the script into some functions 
 * Add a little more validation and error reporting

You choose which of those things to do now and which can be punted to followup issues.

Then resubmit for a final review.

admin/make_homebrew_recipe.py
adamtheturtle: > And ideally (I think) we'd just supply the Flocker sdist package as an argument to this script. 
The locally built package is the "known good" package. It just happens to be uploaded to google storage in our current process. 

I see the advantage, but this way tests that the package is actually available at the download URL, so I'd rather keep it with better error reporting.

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: ` pip install --editable .[release]` is above this line already so I won't.

admin/make_homebrew_recipe.py
adamtheturtle: I meant comments on this PR .`print_list` was mentioned at https://github.com/ClusterHQ/flocker/pull/1016#issuecomment-66764446 and is in the link posted on https://github.com/ClusterHQ/flocker/pull/1016#discussion_r21965191.

adamtheturtle:
> it should take a positional argument containing the path to a local Flocker sdist package (I think that's what @tomprince would prefer if we get round to running this automatically from Buildbot build artifacts.)

I suspect that a few changes could be made to better support the CI use case but I'll modify https://clusterhq.atlassian.net/browse/FLOC-1150 and refer to this comment.

exarkun:
> You choose which of those things to do now and which can be punted to followup issues.

Keep in mind this is an infrequently used internal tool.  We don't need to polish it all the way right now.

admin/make_homebrew_recipe.py
adamtheturtle: This can be changed if there is ever an issue.

adamtheturtle:
I have addressed the inline comments (either by replying or changing the code), and the following final comments have been done:

 * script should be moved to a module leaving `./admin/bin/make_homebrew_script` as a stub script (in line with the other admin tools)
* Split the script into some functions 
* Add a little more validation and error reporting

The following have not:

 * some command line help (`twisted.python.usage`)
 * it should take a positional argument containing the path to a local Flocker sdist package (I think that's what @tomprince would prefer if we get round to running this automatically from Buildbot build artifacts.)
 
If changes need to be made to help this run on Buildbot, they can be made in https://clusterhq.atlassian.net/browse/FLOC-1150.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
+#!/usr/bin/env python
```
wallrj: No need for shebang line now that this is a module.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
...
+
+    :param unicode application: The name of an application to get the
+        dependencies of.
+
+    :return tl.eggdeps.graph.Graph: Graph of Python dependencies, not
+        including ``application``.
+    """
+    dependency_graph = Graph()
+    dependency_graph.from_specifications([application])
+    dependency_graph.pop(application, None)
```
wallrj: This looks a bit odd. Maybe add a comment explaining what this does, why we're discarding the result and what the `None` argument is for...maybe supply keyword arguments to make it clear.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
...
+    if version is None:
+        raise Exception("Set the VERSION environment variable.")
+    return version
+
+
+def get_checksum(url):
+    """
+    Given the URL of a file, download that file and return its sha1 hash.
+
+    :param unicode url: A URL of a file.
```
wallrj: URLs must always be bytes...I think.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
...
+
+def get_formatted_dependency_list(dependency_graph):
+    """
+    :param tl.eggdeps.graph.Graph dependency_graph: Graph of Python
+        dependencies.
+
+    :return unicode: Space separated list of dependency names.
+    """
+    dependencies = []
+    for name, node in sorted(dependency_graph.iteritems()):
```
wallrj: I think you're intending to sort the returned  names alphabetically, so I'd say iterate of the `keys` alphabetically or better still do the sorting at the end when you join the list into  a string.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
...
+  resource "{project_name}" do
+    url "{url}"
+    sha1 "{checksum}"
+  end
+"""
+    for name, node in sorted(dependency_graph.iteritems()):
+        requirement = node.dist.as_requirement()
+        operator, version = requirement.specs[0]
+        project_name = requirement.project_name
+        url = u"http://pypi.python.org/pypi/{name}/{version}/json".format(
```
wallrj: Url should be bytes. So encode this to ascii.

admin/homebrew.py
```diff
@@ -0,0 +1,174 @@
...
+
+def get_resource_stanzas(dependency_graph):
+    """
+    :param tl.eggdeps.graph.Graph dependency_graph: Graph of Python
+        dependencies.
+
+    :return unicode: The part of the Homebrew recipe which defines the Python
+        packages to install.
+    """
+    resources = u""
```
wallrj: Might have been nice to make this a list and then join it at the end.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -307,50 +311,19 @@ Release
...
-   - Checkout the `homebrew-tap`_ repository:
+#. Create a version specific ``Homebrew`` recipe for this release:
 
-     .. code-block:: console
-
-        git clone git@github.com:ClusterHQ/homebrew-tap.git
+   XXX This should be automated https://clusterhq.atlassian.net/browse/FLOC-1150
 
-   - Create a release branch:
+   - Create a recipe file and push it tos the `homebrew-tap`_ repository:
```
wallrj: Typo "tos"

wallrj:
Thanks @adamtheturtle 

Sorry this took so long....too many interruptions today.

I've left one or two inline comments.

I don't know how you tested this without uploading an unreleased Flocker package to Google Cloud Storage.

I followed the new instructions using version 0.3.2 and it works up to the point of actually running flocker-deploy.

That doesn't work, because it was built against a different version of ipaddr.

```
ClusterHQs-Mac:~ ClusterHQVM$ flocker-deploy 
  File "/usr/local/Cellar/flocker-0.3.2/0.3.2/libexec/bin/flocker-deploy", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/usr/local/Cellar/flocker-0.3.2/0.3.2/libexec/vendor/lib/python2.7/site-packages/pkg_resources.py", line 2829, in <module>
    working_set = WorkingSet._build_master()
  File "/usr/local/Cellar/flocker-0.3.2/0.3.2/libexec/vendor/lib/python2.7/site-packages/pkg_resources.py", line 451, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/local/Cellar/flocker-0.3.2/0.3.2/libexec/vendor/lib/python2.7/site-packages/pkg_resources.py", line 464, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/local/Cellar/flocker-0.3.2/0.3.2/libexec/vendor/lib/python2.7/site-packages/pkg_resources.py", line 639, in resolve
    raise DistributionNotFound(req)
pkg_resources.DistributionNotFound: ipaddr==2.1.10
```

This is another reason (I think) for supplying a Flocker sdist *file* rather than assuming it can be downloaded from our repo....but in the interests of getting this finished, perhaps you can just create a followup issue for that.

I'm also not sure about the handling of "dependencies of dependencies". Presumably this script installs the lowest allowed version of those. I guess that's fine, but it may result in Mac users getting a different (even more different) set of dependencies than people installing on Linux.
I think ultimately, when we use peep, we may end up adding the list of all dependencies and specific versions to our requirements.txt file, at which point it'll be more stable between operating systems...I think.
I'd say create an issue about that, but I'm not really sure how to describe the problem.

Anyway, I'm fairly convinced that this will work when we use it during next weeks dev release, but if not we can always create some bug fix issues then.

So please answer or address the points above and then merge.

wallrj:
Oh, and I also noticed that the script adds some trailing whitespace to the ruby file. See screenshot:
screenshot from 2015-01-08 15 34 21

adamtheturtle:
> This is another reason (I think) for supplying a Flocker sdist file rather than assuming it can be downloaded from our repo....but in the interests of getting this finished, perhaps you can just create a followup issue for that.

The existing followup issue (for getting this running with CI) links to a comment which mentions that as an option.

adamtheturtle:
I've tested this by creating a new .tar.gz

```
  url "http://cl.ly/1Y1V3C1b2i10/Flocker-0.3.2-668-gd7a00e2.tar.gz"
  sha1 "452f6e321a20e68259f81ce6753fe8be3a4fbbea"
```

and this seems to work.

I have also addressed the minor comments (but left the resource creation the same).

Thanks for the many reviews everyone.

PR #1024, opened 2014-12-16
[FLOC-1161] Disable functional packaging tests.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1153


wallrj:
It think the issue link above is wrong.
The correct issue is https://clusterhq.atlassian.net/browse/FLOC-1161 which is now closed "won't do".
So I think we should close this PR.

PR #1025, opened 2014-12-18
Clean nodes faster in acceptance tests

dwgebler:
This simple change has cut the average time it takes to run the acceptance test suite for me by over 4 minutes.

Fixes FLOC-1168


flocker/acceptance/testtools.py
```diff
@@ -157,6 +151,10 @@ def _clean_node(test_case, node):
     # http://doc-dev.clusterhq.com/advanced/cleanup.html#removing-zfs-volumes
```
adamtheturtle: This is no longer a valid anchor link, the heading has been changed to bold text.

flocker/acceptance/testtools.py
```diff
@@ -157,6 +151,10 @@ def _clean_node(test_case, node):
...
     # A tool or flocker-deploy option to purge the state of a node does
     # not yet exist. See https://clusterhq.atlassian.net/browse/FLOC-682
+    containers = _run_SSH(22, 'root', node, ['docker', 'ps', '-aq'], None)
```
adamtheturtle: http://doc-dev.clusterhq.com/advanced/cleanup.html says that to achieve this, three steps must be performed (in order):
* Stop Containers
* Remove Containers
* Remove ZFS Volumes

Please either change this to exactly match those docs' commands, or explain why not.

Also, I get that this messy option is faster and we do it a lot in these scripts, but if running `flocker_deploy` with "clean" configurations does the same job but slower, can we not replace those docs with an instruction to run `flocker_deploy` with "clean" configurations.

In summary, I think that the code should match and link to the docs, or have comments explaining why it doesn't.

flocker/acceptance/testtools.py
```diff
@@ -157,6 +151,10 @@ def _clean_node(test_case, node):
...
+    containers = containers.split(b"\n")[:-1]
+    for c in containers:
```
adamtheturtle: One letter variable names are bad for readability and maintainability (in my opinion), please change this to `container`.

adamtheturtle:
This is worse for readability and maintainability but measurably faster to run. Please provide metrics to support this being merged (i.e. the downsides would outweigh the upsides for a 1% speed improvement, but perhaps not a 90% speed improvement, 4 minutes out of how much). As this was done a while ago, I ask for the metrics to be determined again now as at least one thing has changed in master since meant to speed up flocker-deploy: `b"-o", b"IgnoreUnknown=GSSAPIAuthentication"` has been added to SSH commands.

adamtheturtle:
Thanks, please address the above comments and resubmit.

dwgebler:
In respect of your question about metrics on how much faster this actually is, a variety of factors will impact total execution time for the acceptance tests, so simply running the tests both with and without this change and noting the difference isn't sufficient to be reliably sure the entire time difference was due only to how we're cleaning the nodes. On average, I was getting times 25-30% quicker this way (in comparison to master at the time of writing), but either way, what we do know for sure is that the work done in these commands is considerably less computational work than `flocker-deploy` performs, because that's the equivalent of these commands plus a bunch of other stuff.

flocker/acceptance/testtools.py
dwgebler: Stopping the containers, removing them, then deleting ZFS volumes is what happens here, in that order. It's done with slightly different Docker commands because we cannot use pipe over these SSH calls - the underlying mechanism uses `Popen` and it's a potential security risk to injection to enable shell access that way.
There might be a good case for simplifying the docs user process for cleaning a node to deploying empty Flocker configs, I don't think that's in the scope of this issue, which is just to make the automated tests a bit faster, but I'd certainly suggest raising that as an issue. The less people following our docs have to do, the better.

flocker/acceptance/testtools.py
dwgebler: This line isn't a change in this branch/story. If a link is wrong, please create a separate issue for it.

flocker/acceptance/testtools.py
adamtheturtle: I have raised this issue at https://clusterhq.atlassian.net/browse/FLOC-1262.

itamarst:
No longer relevant to current design, unfortunately, so closing.

PR #1026, opened 2015-01-02
[FLOC-697] Dataset model - design review

itamarst:
Haven't updated `flocker.volume` at this point, that will end up in follow up issue.

https://clusterhq.atlassian.net/browse/FLOC-697

### DIFF
--- a/flocker/node/_config.py
+++ b/flocker/node/_config.py
@@ -62,7 +62,7 @@ class IApplicationConfiguration(Interface):
         Returns the ``Application`` instances parsed from the supplied
         configuration.
 
-        This method should only be called after valdiating the format
+        This method should only be called after validating the format
         with a call to ``is_valid_format``.
 
         This method should only be called once, in that calling it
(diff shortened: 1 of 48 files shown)


itamarst:
The changes to `_config.py` are kind of a mess. But basic idea is that there's optional top-level key for datasets.

flocker/node/_config.py
```diff
@@ -1311,4 +1346,5 @@ def marshal_configuration(state):
         "version": 1,
         "applications": result,
         "used_ports": sorted(state.used_ports),
+        # XXX "manifestations": ...
```
itamarst: And also need to output additional `datasets` key in global config combined by `flocker-deploy`.

wallrj:
Thanks @itamarst

I had some trouble understanding the motivation for ``Manifestation`` but after our meeting this afternoon I think I get it.

It would have been nice to have seen some documentation explaining this motivation...perhaps a link to the design document.

Here's my understanding.
A ``Dataset`` is a cluster-wide representation of a bag of data which may be attached to an ``Application``.
A ``Dataset`` may not be attached to any ``Application``s but it will still be managed and reported by Flocker.
A ``Dataset``s must always have one ``primary`` ``Manifestation`` and *may* also have  non-primary (replica) ``Manifestation``s; if the storage backend supports replicas.

A ``Volume`` then becomes less overloaded and represents anything that can be supplied as the host path in a ``docker run --volume`` value.

A user can configure ``Dataset`` s and their ``Manifestation`` s implicitly by configuring application volumes.

They might also configure an unattached ``Dataset`` by explicitly configuring ``manifestations`` in their ``applications.yml`` file. But how will they choose which node the manifestation belongs on? Will there be a ``node`` subkey for each manifestation? Or is this something that ought to be in ``deployment.yml``.

Then again what about ``flocker-reportstate``? which is node specific. Presumably it will output  ``applications.yml`` content containing only the  ``manifestations`` on that node. In which case, there'll need to be code in ``flocker-deploy`` for merging the reported manifestations and there locations.

I suppose it'd be nice to start the implementation by writing some documentation explaining how all this is going to fit together with links to followup issues...but that's easier said than done.

That's all I've got. It'll be interesting to see how this turns out. Please proceed.

PR #1028, opened 2015-01-05
[FLOC-1170] Pre-pull tutorial images

wallrj:
All the images used in the tutorial examples are now included in the tutorial vagrant image.

This means that the separate pull steps can be removed from the documentation.

The acceptance tests don't seem to simulate the separate pull steps so no changes are needed there.

Fixes https://clusterhq.atlassian.net/browse/FLOC-1170

### DIFF
--- a/docs/gettingstarted/examples/environment.rst
+++ b/docs/gettingstarted/examples/environment.rst
@@ -25,29 +25,15 @@ If you have since shutdown or destroyed those VMs, boot them up again:
    Bringing machine 'node1' up with 'virtualbox' provider...
    ==> node1: Importing base box 'clusterhq/flocker-dev'...
 
-Download the Docker Image
-=========================
+Launch MySQL
+============
 
-The Docker image used by this example is quite large, so you should pre-fetch it to your nodes.
-
-.. code-block:: console
-
-   alice@mercury:~/flocker-tutorial$ ssh -t root@172.16.255.250 docker pull mysql:5.6.17
-   ...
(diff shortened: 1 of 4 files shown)


itamarst:
I think this change might actually have a negative impact on tutorial completion.

The goal here is to improve completion rates of the tutorial. Analytics suggest people don't get past installation phase of tutorial. For some reason @lukemarsden decided that speeding up downloads of other examples is relevant to the tutorial, but I'm not really sure why - no relevant analytics evidence was presented. Insofar as increasing the size of the vagrant image makes it take longer to download I can't see how that would help tutorial completion rates.

1. How much bigger is the vagrant image?
2. @lukemarsden why did you propose doing this?

lukemarsden:
@itamarst I think something got lost here. For starters, @wallrj wasn't suppose to pick this up because it was marked as assigned to me.

I am proposing changing the tutorial Vagrant image so that it *only* includes the Docker images for the new getting started guide: https://clusterhq.atlassian.net/browse/FLOC-1167

This means (a) the tutorial Vagrant image will be as small as it can be for people following the new getting started guide given (b) all the waiting will happen at the start, when downloading the Vagrant image, rather than later, during flocker-deploy runs.

itamarst:
Thanks for the clarification. FYI at the moment our vagrant image does in fact only include the tutorial images, so not sure if change is necessary on that front.

lukemarsden:
@itamarst To be clear, the new getting started guide proposes using different Docker images. I will cross that bridge (getting them into the tutorial Vagrant image) when I come to it.

PR #1031, opened 2015-01-05
[FLOC-921] Sphinx apidocs generation

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-921

1. I'm not sure this is the best way to change the directive to support non-hardcoded schema store and examples, but lacking an actual usage of the code it's not clear what a better way is.
2. Directive is untested, as it was in original code this is copied from...

I suggest limiting this branch to what it is (minimal port of existing code) and when we have issues actually using it (as is likely), fix those in a follow-up issue.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -34,6 +34,7 @@ commands =
     sphinx-build -a -b spelling docs/ docs/_build/spelling
     sphinx-build -a -b html docs/ docs/_build/html
     sphinx-build -a -b linkcheck docs/ docs/_build/linkcheck
+    trial --rterror flocker.restapi.docs
 
 [testenv:admin]
 basepython = python2.7
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,7 @@ basepython = python2.7
 basepython = python2.7
 deps = flake8
 changedir = {toxinidir}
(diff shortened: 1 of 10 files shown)


tox.ini
```diff
@@ -34,6 +34,7 @@ commands =
     sphinx-build -a -b spelling docs/ docs/_build/spelling
     sphinx-build -a -b html docs/ docs/_build/html
     sphinx-build -a -b linkcheck docs/ docs/_build/linkcheck
+    trial --rterror flocker.restapi.docs
```
wallrj: Does this really belong here? This build environment is for generating the documentation. This new step is testing some of our documentation  infrastructure as far as I can tell. Maybe move it to `[testenv]`.

wallrj:
Thanks @itamarst 

I'm not really sure how to review this. I've run the `flocker.restapi.docs`  tests and they pass.
Builds are mostly green, apart from the usual suspects.

Only thing that struck me is that the new modules should have the HybridLogic copyright header added...assuming it's our code.

> I suggest limiting this branch to what it is (minimal port of existing code) and when we have issues actually using it (as is likely), fix those in a follow-up issue.

That seems fine to me. So please merge after answering or addressing the points above.

PR #1033, opened 2015-01-06
State change acceptance tests floc 267

dwgebler:
Design sketch (with one test implemented as example) for FLOC-267
Also implements adding environment info to flocker-reportstate, please see my comment for more detail on https://clusterhq.atlassian.net/browse/FLOC-1174

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -69,7 +69,7 @@ def create_application(name, image, ports=frozenset(), volume=None,
     """
     Instantiate an ``Application`` with the supplied parameters and return it.
     """
-    if not ':' in image:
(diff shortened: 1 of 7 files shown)


flocker/acceptance/testtools.py
```diff
@@ -69,8 +69,10 @@ def create_application(name, image, ports=frozenset(), volume=None,
     """
     Instantiate an ``Application`` with the supplied parameters and return it.
     """
+    if ':' not in image:
```
wallrj: `DockerImage.from_string` already does this parsing so this should never have been necessary. Double check and if not create an issue for fixing that.

itamarst:
Unit tests are probably sufficient.

PR #1037, opened 2015-01-06
Misc release doc fixes

exarkun:
Fixes FLOC-1177

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -172,12 +172,18 @@ This review step is to ensure that all acceptance tests pass on the release bran
 .. warning:: This process requires ``Vagrant`` and should be performed on your own workstation;
             **not** on a :doc:`Flocker development machine <vagrant>`.
 
+#. Export the version number of the release being reviewed as an environment variable for later use:
+
+   .. code-block:: console
+
+      export VERSION=0.1.2
+
 #. Do the acceptance tests:
 
    - Add the tutorial vagrant box that BuildBot has created from the release branch.
 
      The URL can be found by examining the "upload-base-box" step of the ``flocker-vagrant-tutorial-box`` builder.
-     The URL will look like ``http://build.clusterhq.com/results/vagrant/<RELEASE_BRANCH>/flocker-tutorial.json``.
+     The URL will look like ``http://build.clusterhq.com/results/vagrant/release/flocker-${VERSION}/flocker-tutorial.json``.
 
      .. code-block:: console
 
@@ -201,8 +207,8 @@ This review step is to ensure that all acceptance tests pass on the release bran
 
         git clone git@github.com:ClusterHQ/flocker.git
         cd flocker
-        git checkout -b *release branch*
-        mkvirtualenv flocker-release-${VERSION}
+        git checkout "release/flocker-${VERSION}"
+        mkvirtualenv "flocker-release-${VERSION}"
         pip install --editable .[dev]
 
    - Install `PhantomJS`_:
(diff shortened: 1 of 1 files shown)


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -172,12 +172,18 @@ This review step is to ensure that all acceptance tests pass on the release bran
...
+
+   .. code-block:: console
+
+      export VERSION=0.1.2
+
 #. Do the acceptance tests:
 
    - Add the tutorial vagrant box that BuildBot has created from the release branch.
 
      The URL can be found by examining the "upload-base-box" step of the ``flocker-vagrant-tutorial-box`` builder.
```
adamtheturtle: The URL cannot be found by examining the "upload-base-box" step of the ``flocker-vagrant-tutorial-box`` builder, so this should either be removed, or changed to reference the "tutorial box" link under ``write-base-box-metadata``.

Or to clarify what I was suggesting on the release PR, this line and the line below ("The URL will look like...") can be removed, and the console block below can be changed from:
```
.. code-block:: console

        vagrant box add <URL>
```

to

```
.. code-block:: console

        vagrant box add http://build.clusterhq.com/results/vagrant/release/flocker-${VERSION}/flocker-tutorial.json
```

I think that this is the best solution.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -201,8 +207,8 @@ This review step is to ensure that all acceptance tests pass on the release bran
 
         git clone git@github.com:ClusterHQ/flocker.git
         cd flocker
-        git checkout -b *release branch*
-        mkvirtualenv flocker-release-${VERSION}
+        git checkout "release/flocker-${VERSION}"
```
adamtheturtle: :question: The engineer doing the release does

```
$ git clone git@github.com:ClusterHQ/flocker.git "flocker-${VERSION}"
$ cd flocker-${VERSION}
```

Perhaps it would be better to have these consistent - I guess the reason to clone into a directory called "flocker-${VERSION}" and not just "flocker" is to avoid conflict with the "flocker" directory that the release engineer will likely have.

adamtheturtle:
Thanks Jean-Paul, please address the comments as you see fit and merge when Buildbot passes (enough).

I still think that "Pre-tag Review Process" would be better off with a "Check 'Files changed' on the release PR against the expected changes from 'Preparing For a Release'" step. Else it is saying that the release can be continued even without checking for "What's New" etc. One reason that I think it is important to be so explicit with this is that as we automate each step we remove the documentation to do it manually - everything currently in "Pre-tag Review Process" can and will be automated, but checking "NEWS" cannot.

adamtheturtle:
`FLOC-965` is mentioned in an XXX in the review step as it automates part of the review process. This issue has been done and so the review process should be updated to reflect that.

wallrj:
I notice there's a *genuine* broken link in the release docs: 

```
gettinginvolved/infrastructure/release-process.rst:40: [broken] https://vagrantcloud.com/organization/clusterhq/teams: HTTP Error 404: Not Found
```

http://build.clusterhq.com/builders/flocker-docs/builds/3363/steps/build-linkcheck/logs/errors

exarkun:
> FLOC-965 is mentioned in an XXX in the review step as it automates part of the review process. This issue has been done and so the review process should be updated to reflect that.

The resolution of FLOC-965 was automating the tutorial acceptance tests on Rackspace cloud.  The acceptance tests have already been run by this point in the release process though.  Is the intent that they be run again?

If that's the case, I think I'll leave this as-is for now and add a comment to FLOC-957 and FLOC-958 that they should include an update to this document to fix the instructions in this section.

PR #1038, opened 2015-01-07
[FLOC-909] Clarify that `zfs destroy` should be run on the remote node

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-909

It might also be worth making it clear that the cleanup steps need to be performed in sequence ie all the containers must  be stopped before removing their zfs volumes. Otherwise you get this:

```
(flocker-tutorial)[~/flocker-tutorial]$ ssh root@172.16.255.250 'zfs destroy -r flocker'
cannot destroy 'flocker/7b766437-2c5a-46aa-afb1-abedf99a04d0.default.mongodb-volume-example': dataset is busy
```

The steps could be converted from headings to a numbered list.

### DIFF
--- a/docs/advanced/cleanup.rst
+++ b/docs/advanced/cleanup.rst
@@ -34,7 +34,7 @@ Removing Containers
 .. code-block:: console
 
    alice@mercury:~/flocker-mysql$ ssh root@172.16.255.250 'docker ps -aq | xargs --no-run-if-empty docker rm'
-   
+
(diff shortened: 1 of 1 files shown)


adamtheturtle:
This looks good and I've tested that it works. Either merge when Buildbot passes to your satisfaction, or make your proposed changes and then merge when Buildbot passes.

wallrj:
Thanks, I decided to convert the steps to a numbered list as discussed. The doc builder passes so I'll merge.

PR #1040, opened 2015-01-07
[FLOC-1182] New terminology

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1182

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -23,18 +23,18 @@ from ..volume.service import VolumeName
 from ..common import gather_deferreds
 
 
-def _to_volume_name(name):
+def _to_volume_name(dataset_id):
     """
     Convert unicode name to ``VolumeName`` with ``u"default"`` namespace.
 
     To be replaced in https://github.com/ClusterHQ/flocker/issues/737 with
     real namespace support.
 
-    :param unicode name: Volume name.
+    :param unicode dataset_id: Dataset ID.
 
     :return: ``VolumeName`` with default namespace.
     """
-    return VolumeName(namespace=u"default", id=name)
+    return VolumeName(namespace=u"default", dataset_id=dataset_id)
 
 
 class IStateChange(Interface):
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -333,8 +333,9 @@ class Deployer(object):
         def map_volumes_to_size(volumes):
             managed_volumes = dict()
             for volume in volumes:
-                if volume.uuid == self.volume_service.uuid:
-                    managed_volumes[volume.name.id] = volume.size.maximum_size
+                if volume.node_id == self.volume_service.node_id:
+                    managed_volumes[volume.name.dataset_id] = (
+                        volume.size.maximum_size)
             return managed_volumes
         volumes.addCallback(map_volumes_to_size)
         d = gatherResults([self.docker_client.list(), volumes])
--- a/flocker/node/test/test_deploy.py
+++ b/flocker/node/test/test_deploy.py
@@ -2534,7 +2534,8 @@ class WaitForVolumeTests(SynchronousTestCase):
                                   mountpoint=FilePath(u"/var")))
         wait.run(deployer)
         self.assertEqual(result,
-                         [VolumeName(namespace=u"default", id=u"myvol")])
+                         [VolumeName(namespace=u"default",
+                                     dataset_id=u"myvol")])
 
     def test_return(self):
         """
--- a/flocker/node/test/test_deploy.py
+++ b/flocker/node/test/test_deploy.py
@@ -961,7 +961,7 @@ class DeployerDiscoverNodeConfigurationTests(SynchronousTestCase):
                     activation_state=u'active')
         units = {unit.name: unit}
 
-        volume = Volume(uuid=unicode(uuid4()),
+        volume = Volume(node_id=unicode(uuid4()),
                         name=_to_volume_name(u"site-example.com"),
                         service=self.volume_service)
         self.successResultOf(volume.service.pool.create(volume))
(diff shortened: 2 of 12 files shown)


flocker/node/_deploy.py
```diff
@@ -23,18 +23,18 @@
 from ..common import gather_deferreds
 
 
-def _to_volume_name(name):
+def _to_volume_name(dataset_id):
```
exarkun: A number of callers of `_to_volume_name` remain in this file which pass `self.volume.name` where `self.volume` is an `AttachedVolume` instance.  `AttachedVolume.name` is still documented as a short, human-readable identifier.

flocker/node/test/test_deploy.py
```diff
@@ -961,7 +961,7 @@ def test_discover_remotely_owned_volumes_ignored(self):
                     activation_state=u'active')
         units = {unit.name: unit}
 
-        volume = Volume(uuid=unicode(uuid4()),
+        volume = Volume(node_id=unicode(uuid4()),
                         name=_to_volume_name(u"site-example.com"),
```
exarkun: And some callers of `_to_volume_name` like this remain.

flocker/volume/_ipc.py
```diff
@@ -79,7 +79,7 @@ def acquire(volume):
         :param Volume volume: The volume which will be acquired by the
             remote volume manager.
 
-        :return: The UUID of the remote volume manager (as ``unicode``).
+        :return: The node ID of the remote volume manager (as ``unicode``).
```
exarkun: I wonder if (not now) it would make sense to represent UUIDs as UUID instances instead of unicode.

flocker/volume/_ipc.py
```diff
@@ -132,7 +132,7 @@ def receive(self, volume):
         return self._destination.run([b"flocker-volume",
                                       b"--config", self._config_path.path,
                                       b"receive",
-                                      volume.uuid.encode(b"ascii"),
+                                      volume.node_id.encode(b"ascii"),
```
exarkun: just a note(this is pre-existing, clearly) - `...encode(b"ascii")` is a hypercorrection (should just be `"ascii"`).

flocker/volume/service.py
```diff
@@ -48,8 +48,8 @@ class VolumeName(object):
     :ivar unicode namespace: The namespace of the volume,
         e.g. ``u"default"``. Must not include periods.
 
-    :ivar unicode id: The id of the volume,
-        e.g. ``u"mypostgresdata"``. Since volume ids must match Docker
+    :ivar unicode dataset_id: The unique id of the
+        dataset. ``u"mypostgresdata"``. Since volume ids must match Docker
```
exarkun: `u"mypostgresdata"` isn't a good example anymore

exarkun: The need to document character constraints here is probably a good point in favor of (eventually) representing these as UUID instances instead of unicode.

flocker/volume/service.py
```diff
@@ -255,9 +255,9 @@ def enumerated(filesystems):
                 #    https://github.com/ClusterHQ/flocker/issues/78
                 basename = filesystem.get_path().basename()
                 try:
-                    uuid, name = basename.split(b".", 1)
+                    node_id, name = basename.split(b".", 1)
                     name = VolumeName.from_bytes(name)
-                    uuid = UUID(uuid)
+                    UUID(node_id)
```
exarkun: Maybe now worth a comment that this is here for validation purposes.

flocker/volume/test/test_ipc.py
```diff
@@ -208,9 +209,9 @@ def pushed(ignored):
             created.addCallback(got_volume)
             return created
 
-        def test_acquire_returns_uuid(self):
+        def test_acquire_returns_node_id(self):
             """
-            ``acquire()`` returns the UUID of the remote volume manager.
+            ``acquire()`` returns the NODE_ID of the remote volume manager.
```
exarkun: casing/punctuation weird

flocker/volume/test/test_service.py
```diff
@@ -492,22 +493,22 @@ def test_receive_local_uuid(self):
         service.startService()
 
         self.assertRaises(ValueError, service.receive,
-                          service.uuid.encode("ascii"), b"lalala", None)
+                          service.node_id.encode("ascii"), b"lalala", None)
 
     def test_receive_creates_volume(self):
-        """Receiving creates a volume with the given uuid and name."""
+        """Receiving creates a volume with the given node_id and name."""
```
exarkun: Could also fix the formatting of this docstring

flocker/volume/test/test_service.py
```diff
@@ -613,12 +616,13 @@ def test_enumerate_skips_other_filesystems(self):
         service = VolumeService(FilePath(self.mktemp()), pool, reactor=Clock())
         service.startService()
 
-        name = VolumeName(namespace=u"mynspaces", id=u"good volume name")
+        name = VolumeName(namespace=u"mynspaces",
+                          dataset_id=u"good volume name")
```
exarkun: This value doesn't conform to the documented constraints (spaces are supposedly not allowed).

exarkun:
Thanks.  I think these points are all straightforward except perhaps for the first one about `AttachedVolume.name`.

I think I'd be okay with a doc-only solution to that here (just document name as being an opaque unique identifier thing instead of what it says now).  If that makes sense to you, go ahead with that change and addressing the other review comments and then and merge.

If that doesn't make sense then maybe we should talk about alternate solutions.  Thanks again.

flocker/node/_deploy.py
itamarst: That's all going to be fixed in FLOC-697, including the documentation.

flocker/node/test/test_deploy.py
itamarst: Again hopefully fixed in FLOC-697,or maybe followup task depending how big it is.

PR #1041, opened 2015-01-08
[FLOC-1189] Sketch of new API that doesn't make my brain ache

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1189

I needed to change some of this code, and found it almost impossible to reason about, so here's a sketch of what a nicer API would look like.

There's a minor amount of extra effort (fig YAML will be parsed twice) but whatever, not on critical path.

### DIFF
--- a/flocker/cli/script.py
+++ b/flocker/cli/script.py
@@ -20,9 +20,9 @@ from characteristic import attributes
 
 from ..common.script import (flocker_standard_options, ICommandLineScript,
                              FlockerScriptRunner)
-from ..node import (FlockerConfiguration, ConfigurationError,
+from ..node import (ConfigurationError,
                     FigConfiguration, applications_to_flocker_yaml,
-                    model_from_configuration)
+                    parse_desired_configuration)
 
 from ..common import ProcessNode, gather_deferreds
 from ._sshconfig import DEFAULT_SSH_DIRECTORY, OpenSSHConfiguration
@@ -96,16 +96,8 @@ class DeployOptions(Options):
                 self['application_config'] = (
                     applications_to_flocker_yaml(applications)
                 )
-            else:
-                configuration = FlockerConfiguration(app_config_obj)
-                if configuration.is_valid_format():
-                    applications = configuration.applications()
-                else:
-                    raise ConfigurationError(
-                        "Configuration is not a valid Fig or Flocker format."
-                    )
-            self['deployment'] = model_from_configuration(
-                applications=applications,
+            self['deployment'] = parse_desired_configuration(
+                application_config=app_config_obj,
                 deployment_configuration=deploy_config_obj)
         except ConfigurationError as e:
             raise UsageError(str(e))
--- a/flocker/node/_config.py
+++ b/flocker/node/_config.py
@@ -1190,9 +1190,12 @@ class FlockerConfiguration(object):
             self._applications[application_name] = Application(**attributes)
 
 
-def deployment_from_configuration(deployment_configuration, all_applications):
+def _apply_deployment_to_applications(
+        deployment_configuration, all_applications):
     """
-    Validate and parse a given deployment configuration.
+    Validate and parse a given deployment configuration, and use that and
+    applications to create corresponding ``Node`` objects describing which
+    application runs where.
 
     :param dict deployment_configuration: The intermediate configuration
         representation to load into ``Node`` instances.  See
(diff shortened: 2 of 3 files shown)


flocker/cli/script.py
```diff
@@ -96,16 +96,8 @@ def parseArgs(self, deployment_config, application_config):
...
-                if configuration.is_valid_format():
-                    applications = configuration.applications()
-                else:
-                    raise ConfigurationError(
-                        "Configuration is not a valid Fig or Flocker format."
-                    )
-            self['deployment'] = model_from_configuration(
-                applications=applications,
+            self['deployment'] = parse_desired_configuration(
+                application_config=app_config_obj,
```
exarkun: This seems like a worthwhile refactoring.

flocker/node/_config.py
```diff
@@ -1190,9 +1190,12 @@ def _parse(self):
             self._applications[application_name] = Application(**attributes)
 
 
-def deployment_from_configuration(deployment_configuration, all_applications):
+def _apply_deployment_to_applications(
```
exarkun: The conceptual model that suggests "apply" in this name isn't apparent to me.

exarkun: I suspect something like `Cluster.from_deployment_and_application_config` might make sense as the API for this functionality?  (No, `Cluster` does not exist yet.)

flocker/node/_config.py
```diff
@@ -1245,35 +1248,50 @@ def deployment_from_configuration(deployment_configuration, all_applications):
     return set(nodes)
 
 
-def model_from_configuration(applications, deployment_configuration):
+def parse_desired_configuration(application_configuration,
```
exarkun: `Deployment.from_deployment_and_application_config`?  Now I'm trying to figure out how a `Deployment` differs from a hypothetical `Cluster`.  Probably it doesn't, I guess.  So maybe this is `Deployment.from_deployment_and_application_config` (or just `from_config`?) and the above API is some other private helper.

flocker/node/_config.py
```diff
@@ -1245,35 +1248,50 @@ def deployment_from_configuration(deployment_configuration, all_applications):
...
+            raise ConfigurationError(
+                "Configuration is not a valid Fig or Flocker format.")
+
+    nodes = _apply_deployment_to_applications(
         deployment_configuration, applications)
     return Deployment(nodes=frozenset(nodes))
 
 
-def current_from_configuration(current_configuration):
+def parse_current_cluster_state(current_cluster_state):
```
exarkun: As long as we stick with `Deployment`, mixing the word "cluster" in here is probably confusing.  So `parse_current_deployment_state` at a minimum - but even better, I think, `Deployment.from_state`?  Or maybe `Deployment.from_current_state` (but while "current" seems to clarify things, I think it actually obscures the fact that there's really no way to know when this state was present on the cluster and we just **hope** it's still the same as the current state).

flocker/node/_config.py
```diff
@@ -1289,14 +1307,14 @@ def current_from_configuration(current_configuration):
...
-def marshal_configuration(state):
+def marshal_node_state(state):
     """
-    Generate representation of a node's applications using only simple Python
-    types.
+    Generate representation of a node's current node state using
+    only simple Python types.
 
-    :param NodeState state: The configuration state to marshal.
+    :param NodeState state: The deployment state to marshal.
```
exarkun: Considering the explicit and sole purpose of a "model" type is to represent state, I find the name `NodeState` somewhat redundant.  Guess that point isn't really in scope here though.

itamarst:
I guess I should clarify some of my motivation a bit:

We have two pieces of high level information we represent - desired configuration and actual cluster state. We also have a class called `Deployment` can store **either** of these, and a "deployment YAML" config file. The problem is that the word "deployment" can if we're not careful imply "actual cluster state", even though instances of `Deployment` can also be used to represent desired configuration.

So I tried to choose naming that reduced the confusion above. Perhaps unsuccesfully, and perhaps we should rename `Deployment`?

flocker/node/_config.py
```diff
@@ -1289,14 +1307,14 @@ def current_from_configuration(current_configuration):
     return Deployment(nodes=frozenset(nodes))
 
 
-def marshal_configuration(state):
+def marshal_node_state(state):
```
exarkun: If you're convinced by any of the `from_config`-style suggestions above then a `to_config`-style name might make more sense here.  `node_state_to_config`?  I'm not 100% behind this idea though, since we seem to be trying to draw a distinction between "desired config" and "deployment state" even though in practice the representation of these two things is pretty similar.  Since this is deployment state, not desired config, perhaps putting "config" in the name is confusing (and the parallel idea, "to_deployment_state" doesn't really make sense).

exarkun:
Out of time for this for now unfortunately.  Given your comment above, maybe the suggestions I've made don't entirely make sense.  Even considering that information, though, I'm not sure I think the new names do a much better job of conveying the intent than the old names did.

One other random thought I had is that having two types to represent exactly what `Deployment` represents might help.  `DesiredConfiguration` and `DeployedState` (or something, not sure `DeployedState` is the best name for that).  These could be no-op `Deployment` subclasses (or something with `proxyForInterface` maybe - woo composition) and would be interchangeable in practice - but the fact that they have different names would help make it clear exactly what is going on anywhere they're used.  Like I said, just a random thought, I haven't looked carefully at the relevant code to see how this might look in specific instances.

I guess I'll let you think this over some more and re-submit for design review.  Thanks.

itamarst:
My feeling is that the proposal is a quick change that improves things. Maybe it's not the final API we want, but mostly I just want to *vaguely* understand what I'm modifying as part of 697. So I'd like to submit that since this is better than what we have now it's worth doing as is, even if not perfect and temporary.

itamarst:
After further thought: I probably don't need this to finish 697. So maybe it's worth taking the time to do this the right way.

PR #1044, opened 2015-01-09
Sketch: Create new flocker.control with corresponding flocker-control script, in charge of managing the cluster.

exarkun:


flocker/control/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Control service for managing a Flocker cluster.
```
exarkun: This probably merits a more expansive description, or at least a link to some other architecture/design documentation (but I don't think we have any linkable right now).

flocker/control/script.py
```diff
@@ -0,0 +1,40 @@
...
+    flocker_standard_options, FlockerScriptRunner, main_for_service)
+
+
+@flocker_standard_options
+class ControlOptions(Options):
+    """
+    Command line options for ``flocker-control`` cluster management process.
+    """
+    optParameters = [
+        ["port", "p", 4523, "The port to listen on.", int],
```
exarkun: Probably should be a string endpoint description I guess (this might even be of practical value, since it should make introducing TLS easier).

flocker/control/script.py
```diff
@@ -0,0 +1,40 @@
...
+        ]
+
+
+class ControlScript(object):
```
exarkun: I think product has difficulty with calling this the "control service".  They want to call it the "API server".  Nothing necessarily has to change here, just a note that we might need to keep working to convince them to change their minds.

flocker/node/script.py
```diff
@@ -301,13 +223,10 @@ class ServeScript(object):
...
     def main(self, reactor, options, volume_service):
-        api_service = create_api_service(
-            TCP4ServerEndpoint(reactor, options["port"]))
-        parent_service = _ServeService(volume_service, api_service)
-        return _main_for_service(reactor, parent_service)
+        return main_for_service(reactor, volume_service)
 
 
-def flocker_serve_main():
+def flocker_volume_main():
```
exarkun: We've been talking about "agents".  Perhaps this should be something like `flocker_volume_agent_main`?  Or maybe `flocker_volume_convergence_main`.  Although since it doesn't do any converging yet I'm not quite sure if that's the right level of abstraction.

setup.py
```diff
@@ -103,7 +103,8 @@ def run(self):
             'flocker-deploy = flocker.cli.script:flocker_deploy_main',
             'flocker-changestate = flocker.node.script:flocker_changestate_main',
             'flocker-reportstate = flocker.node.script:flocker_reportstate_main',
-            'flocker-serve = flocker.node.script:flocker_serve_main',
+            'flocker-volume = flocker.node.script:flocker_volume_main',
```
exarkun: If the main function has its name adjusted, possibly the command line script name should be adjusted correspondingly.

exarkun:
Thanks.

To summarize my understanding, this will:

  * introduce a new `flocker.control` package where the cluster-wide control service will be implemented and which will eventually expose the public API
  * move the existing API server (such as it is) into that package and creates a script for running it
  * take the existing flocker-serve script and main function and re-purposes them to eventually be entry points for a per-node volume convergence agent process
  * factor some some shared service-y code out to a common module for re-use between the control service and the volume convergence agent.

If that's correct, this looks good.  Please proceed with the implementation.

flocker/control/script.py
itamarst: Well... except we're going to have a very specific security policy we want, which is *not* expressable in the string format at the moment. I suggest considering that when we've actually implemented the TLS policy and see if that level of configuration makes sense.

flocker/node/script.py
itamarst: Yeah... just leave it as is, nothing's use it yet anyway, and we can rename it later.

PR #1045, opened 2015-01-09
[FLOC-1178] Control service refactoring

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1178


flocker/control/__init__.py
```diff
@@ -0,0 +1,12 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Control service for managing a Flocker cluster.
+
+A Flocker cluster is composed of a number of nodes. The control service is
+in charge of managing the desired configuration, and exposes a public API
```
exarkun: Extraneous comma

flocker/control/test/test_script.py
```diff
@@ -0,0 +1,54 @@
...
+        options = ControlOptions()
+        options.parseOptions([])
+        self.assertEqual(options["port"], 4523)
+
+    def test_custom_port(self):
+        """
+        The ``--port`` command-line option allows configuring the port.
+        """
+        options = ControlOptions()
+        options.parseOptions(["--port", 1234])
```
exarkun: `b"1234"`

exarkun: Guess that was a pre-existing defect since this code is just being moved.  Still probably sensible to fix it now though.

flocker/node/script.py
```diff
@@ -211,104 +208,26 @@ def flocker_reportstate_main():
...
-        "before", "shutdown", _chain_stop_result, service, stop)
-    return stop
-
-
 @flocker_standard_options
 @flocker_volume_options
-class ServeOptions(Options):
+class VolumeServeOptions(Options):
     """
     Command line options for ``flocker-serve`` cluster management process.
```
exarkun: This has a different name now I think.

setup.py
```diff
@@ -103,7 +103,8 @@ def run(self):
             'flocker-deploy = flocker.cli.script:flocker_deploy_main',
             'flocker-changestate = flocker.node.script:flocker_changestate_main',
             'flocker-reportstate = flocker.node.script:flocker_reportstate_main',
-            'flocker-serve = flocker.node.script:flocker_serve_main',
```
exarkun: There's still a functional test for `flocker-serve` existing which is now failing.  It should be replaced with tests for the two differently named scripts being introduced here, I guess.

exarkun:
Thanks.  Just those few very minor points.  Go ahead and merge once they're addressed.

PR #1047, opened 2015-01-12
[FLOC-697] Dataset model

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-697

This diverged significantly from original design.

Some unfiled followup issues that should probably exist:

1. Fix usage of `_to_volume_name` wherever it doesn't do a UUID. (Not relevant, turns out).
2. In retrospect maximum_size should also have been metadata. Fix that. https://clusterhq.atlassian.net/browse/FLOC-1215
3. There's no way to express datasets that have no manifestations. This only applies to OpenStack backend though.
4. Get rid of deepcopying, replace with persistent datastructures.


flocker/node/_model.py
```diff
@@ -189,6 +164,39 @@ class Application(object):
...
 
 
+@attributes(["dataset", "primary"])
+class Manifestation(object):
+    """
+    A dataset that is mounted on a node.
+
+    :ivar Dataset: The dataset being mounted.
+
+    :ivar bool primary: If true, this is a primary, otherwise it is a replica.
```
exarkun: This appears to be unused.  Lacking any users, it's not clear if this is the right structure for this piece of information.  For example, another approach that might make sense here is to have `Master` and `Replica` types - perhaps with alternate implementations of some `IManifestation` interface.  Or maybe a bool is really the right approach - but it's hard to tell right now.

flocker/node/_model.py
```diff
@@ -189,6 +164,39 @@ class Application(object):
...
+
+    :ivar dataset_id: A unique identifier, as ``unicode``. May also be ``None``
+        if this is coming out of human-supplied configuration, in which
+        case it will need to be looked up from actual state for existing
+        datasets, or a new one generated if a new dataset will need tbe
+        created.
+
+    :ivar PMap metadata: Mapping between ``unicode`` keys and
+        corresponding values. Typically there will be a ``"name"`` key whose
+        value is a a human-readable name, e.g. ``"main-postgres"``.
```
exarkun: If/when we open this up to arbitrary key/value pairs, we should be sure to reserve some of the key-space for ourselves.  For example, all upper-case identifiers or all identifiers beginning with `_` or whatever.  We both want to be able to expand the metadata Flocker itself provides without stomping on user-data and we don't want Flocker to unintentionally consume some metadata a user provides that just happens to use a key we picked too.

Probably nothing to do here now ... except make a note of this point somehow/somewhere?

flocker/node/_model.py
```diff
@@ -189,6 +164,39 @@ class Application(object):
...
+
+    :ivar Dataset: The dataset being mounted.
+
+    :ivar bool primary: If true, this is a primary, otherwise it is a replica.
+    """
+
+
+@attributes(["dataset_id",
+             Attribute("maximum_size", default_value=None),
+             Attribute("metadata", default_factory=pmap)])
```
exarkun: minor point... a `default_value=pmap()` probably works just as well since all empty pmaps are identical anyway.

flocker/node/_model.py
```diff
@@ -189,6 +164,39 @@ class Application(object):
...
+        case it will need to be looked up from actual state for existing
+        datasets, or a new one generated if a new dataset will need tbe
+        created.
+
+    :ivar PMap metadata: Mapping between ``unicode`` keys and
+        corresponding values. Typically there will be a ``"name"`` key whose
+        value is a a human-readable name, e.g. ``"main-postgres"``.
+
+    :ivar int maximum_size: The maximum size in bytes of this dataset, or
+        ``None`` if there is no specified limit.
```
exarkun: Add an issue and link to it in a comment around here about how this should be metadata instead.

flocker/node/_deploy.py
```diff
@@ -424,6 +453,34 @@ def calculate_necessary_state_changes(self, desired_state,
...
+                    dataset = application.volume.dataset
+                    name = dataset.metadata[u"name"]
+                    current_datasets_by_name[name] = dataset
+        for dataset in datasets_with_no_id:
+            matching = current_datasets_by_name.get(dataset.metadata[u"name"])
+            if matching:
+                dataset.dataset_id = matching.dataset_id
+            else:
+                dataset.dataset_id = unicode(uuid4())
+
```
exarkun: This could probably be refactored.  A method to compute `datasets_with_no_id`.  A method to compute `current_datasets_by_name`.  A method to copy `desired_state`, mutate the copy given the previous two pieces of information, and return the copy.  This at least keeps the growth of this already too-long method to a minimum.

flocker/node/test/test_config.py
```diff
@@ -1866,7 +1890,28 @@ def test_volume_max_size_fractional(self):
...
+        If a volume has a ``dataset_id`` attribute then it is set on the
+        created ``Dataset`` object.
+        """
+        dataset_id = unicode(uuid4())
+        config = dict(
+            version=1,
+            applications={
+                'mysql-hybridcluster': {
+                    'image': 'clusterhq/mysql:v1.0.0',
+                    'volume': {'mountpoint': b'/var/lib/mysql',
```
exarkun: The mountpoint should probably be unicode?

flocker/node/test/test_config.py
```diff
@@ -2923,6 +2976,44 @@ def test_application_with_volume_includes_max_size(self):
...
+                ),
+            )
+        ]
+        result = marshal_configuration(
+            NodeState(running=applications, not_running=[]))
+        expected = {
+            'used_ports': [],
+            'applications': {
+                'mysql-hybridcluster': {
+                    'volume': {'mountpoint': b'/var/mysql/data',
```
exarkun: And again here?

flocker/node/test/test_config.py
```diff
@@ -2923,6 +2976,44 @@ def test_application_with_volume_includes_max_size(self):
...
+                               'dataset_id': dataset_id},
+                    'image': u'flocker/mysql:v1.0.0',
```
exarkun: `__future__.unicode_literals` is on.  I'm growing to dislike that future feature, I guess.  It's annoying to always have to scroll up to the top of the file any time I want to know what type a string is (and seeing a '' next to a u'' always makes me want to check).

flocker/node/_deploy.py
```diff
@@ -424,6 +453,34 @@ def calculate_necessary_state_changes(self, desired_state,
...
+                        datasets_with_no_id.append(dataset)
+        current_datasets_by_name = {}
+        for node in current_cluster_state.nodes:
+            for application in node.applications:
+                if application.volume:
+                    dataset = application.volume.dataset
+                    name = dataset.metadata[u"name"]
+                    current_datasets_by_name[name] = dataset
+        for dataset in datasets_with_no_id:
+            matching = current_datasets_by_name.get(dataset.metadata[u"name"])
```
exarkun: I guess we should at least make a note here that this heuristic leaves something to be desired in the case where multiple volumes have been given the same name.  Long term this doesn't concern me much, since long term I think we're going to throw this heuristic away.  But I'm not quite sure how long it will be until we can do that.

exarkun:
Okay, thanks!  I left some comments inline.  They're all pretty minor, I think, except maybe the one about the `primary` flag - but we sort of discussed that issue elsewhere and I think probably it doesn't make sense to make any changes wrt that here.  Feel free to do or not do the suggested refactoring (I don't think that code has long to live either way).  So merge when you and buildbot are happy.

flocker/node/test/test_config.py
itamarst: Maybe it should *always* be on?

PR #1048, opened 2015-01-12
[FLOC-1159] Design sketch - config persistence

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1159

Depends on #1045 being merged first.


flocker/control/_config.py
```diff
@@ -0,0 +1,25 @@
+# XXX move flocker.node._config here, and also add:
+
+
+def marshal_to_application_config_format(deployment):
+    """
+    Convert ``Deployment`` into application configuration format.
+
+    :param Deployment deployment: The current desired configuration.
+
+    :return: Simple Python types suitable for serialization to YAML, in
```
exarkun: I forget if we talked about this already - but maybe JSON instead of YAML?  Perhaps I shouldn't care about this ... but I do :japanese_goblin: :mouse: .

flocker/control/_persistence.py
```diff
@@ -0,0 +1,40 @@
...
+    def startService(self):
+        # Load application and deployment YAML files from disk, set to
+        # self._deployment.
+        pass
+
+    def save(self, deployment):
+        """
+        Save and flush new deployment to disk.
+
+        :return Deferred: Fires when write is finished.
```
exarkun: Hee hee!  Asynchronous filesystem I/O for the win!  :fireworks:

flocker/control/script.py
```diff
@@ -0,0 +1,42 @@
...
+        ]
+
+
+class ControlScript(object):
+    """
+    A command to start a long-running process to control a Flocker
+    cluster.
+    """
+    def main(self, reactor, options):
+        # XXX also start persistence service
```
exarkun: And I guess the API service should get a reference to the persistence service?  Though it won't do anything with it yet... Perhaps it should?  Should this branch come with an implementation of the `get the desired configuration` API endpoint, perhaps?  It seems like that should be a pretty trivial addition to the other stuff that's going to happen and it might be a good way to demonstrate persistence (not the only way though - having the config survive process restarts would demonstrate it too, except I don't think there's any way to observe that it has survived right now ..)

flocker/control/script.py
```diff
@@ -0,0 +1,42 @@
...
+    flocker_standard_options, FlockerScriptRunner, main_for_service)
+
+
+@flocker_standard_options
+class ControlOptions(Options):
+    """
+    Command line options for ``flocker-control`` cluster management process.
+    """
+    optParameters = [
+        # XXX also choose where configuration will be stored
```
exarkun: This will be a config file like the current `/etc/flocker/config.yml` right?  And we're going to gradually expand that to cover the rest of the service (not deployment or application) configuration data we need to track?

Eventually perhaps the config file should be the only thing accepted on the command line and everything else, like port numbers and certificate/key paths, should be in the config file.  But that's definitely eventually, not right away.  Which raises the question - does this branch need the config file at all?  If the control service accepts a port number to listen on as a command line argument, what other configuration does it need now?  I can't think of anything.  (So maybe punt config file handling down the line a bit.)

exarkun:
Thanks.  This basically looks good - but please clarify in the two places where I've left questions.  (To be clear, "rejecting" this design but basically only because I'd like to make sure we're on the same page in those two places before accepting it.  I'll probably accept it once you answer those questions.)

flocker/control/script.py
itamarst: No, this isn't a configuration file. This is data *written* by the server. So it'd go in `/var/lib/flocker` or something. The word "configuration" is a bit misleading, it's configuration of the cluster, not of the control service process as stuff in `/etc/` would be.

flocker/control/_config.py
itamarst: Yeah, JSON is fine.

flocker/control/script.py
exarkun: Okay ... maybe all this new stuff (ie, the port and the path to the stored cluster configuration) **should** be in the control service process configuration file?  It's not a big deal - but I don't see a good reason to have /etc/flocker/config.yml for some configuration and command line options for other configuration.

I guess one difference is that the volume manager uuid is generated and written whereas everything else is going to be a static default value or user-supplied values... But the control service doesn't have a volume manager uuid.

flocker/control/script.py
itamarst: Sure, I can add the API endpoint to scope.

flocker/control/script.py
itamarst: Maybe, but... seems like "config file vs. command line options" is a different issue, and different discussion?

exarkun: Wee, hidden comment.

> Maybe, but... seems like "config file vs. command line options" is a different issue, and different discussion?

Yea, that's fine.

PR #1049, opened 2015-01-13
[FLOC-1202] Clean up, if the acceptance runner receives a signal.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1202

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -30,7 +30,7 @@ def safe_call(command, **kwargs):
     try:
         return process.wait()
     except:
-        process.kill()
+        process.terminate()
 
 
 def check_safe_call(command, **kwargs):
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -7,6 +7,7 @@ from subprocess import call, check_call
 import sys
 import os
 import yaml
+import signal
 
 from zope.interface import Interface, implementer
 from characteristic import attributes
@@ -289,6 +290,16 @@ class RunOptions(Options):
         )
 
 
+def signal_handler(signal, frame):
+    """
+    Exit gracefully when receiving a singal.
+
+    :param int signal: The signal that was received.
+    :param frame: The running frame.
+    """
+    raise SystemExit(1)
+
+
 def main(args, base_path, top_level):
     """
     :param list args: The arguments passed to the script.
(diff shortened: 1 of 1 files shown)


admin/acceptance.py
```diff
@@ -289,6 +317,16 @@ def postOptions(self):
         )
 
 
+def signal_handler(signal, frame):
+    """
+    Exit gracefully when receiving a singal.
```
adamtheturtle: Typo singal > signal.

admin/acceptance.py
```diff
@@ -19,6 +20,33 @@
...
+def safe_call(command, **kwargs):
+    """
+    Run a process and kill it if the process is interrupted.
+
+    Takes the same arguments as ``subprocess.Popen``.
+    """
+    process = Popen(command, **kwargs)
+    try:
+        return process.wait()
+    except:
```
exarkun: What's expected here?  `SystemExit`?  Something wrapping `EINTR`?  A random oops-this-code-isn't-signal-safe exception from the `Popen` implementation?  And why doesn't this exception get re-raised after `process.terminate()`?

admin/acceptance.py
```diff
@@ -305,6 +343,8 @@ def main(args, base_path, top_level):
 
     runner = options.runner
 
+    signal.signal(signal.SIGTERM, signal_handler)
```
exarkun: It might be nice to walk a future maintainer through the logic here a little bit.  Signals can be subtle (and I'll admit that I'm not 100% sure I understand what's going to happen when a `SIGTERM` is delivered to this process).

exarkun:
Thanks.  Please add some comments for future maintainers in the places where I indicated uncertainty and then re-submit and I'll see if I can convince myself I understand what the consequences of this change are going to be.

admin/acceptance.py
tomprince: I added some. I could add some more, referring to the C implementation, but I think what I added is probably sufficient.

exarkun:
Thanks.  I think that I'd personally have an easier time reasoning about behavior in this case if this used Twisted's process management APIs - so I'm just going to apologize in advance for the obscure edge case I missed while reviewing this code. :wink:   The comments seem like they adequately convey the intent of the code now so I'm happy to see this merged (as long as buildbot is happy, which I think it is despite the red `native-rpm-fedora-20` because we've since eliminated that).

PR #1050, opened 2015-01-13
[FLOC-1064] Fail skips

adamtheturtle:
This is a design for part of FLOC-1064. I'd like it to be looked at before I go too deep into overriding all the ways to skip a test.

There are multiple ways to skip a test:
* `@skipIf` / `@skipUnless` decorators
* `raise SkipTest(reason)`
* `test_case.skipTest(reason)`
* `test_case.skip = reason`

and maybe others?

However, not all of these are used in our codebase. Is it worthwhile writing code to cover those cases which are not used. It seems entirely possible to avoid some of these options (e.g. avoid `raise SkipTest(reason)` and only use `test_case.skipTest(reason)`.

Of course implementing this would involve changing all the current skips in the codebase, as well as changing the buildbot and tox configurations.

These are test tools, and so I didn't add tests for them, but in doing an initial implementation sketch it became quite tedious testing manually, so I think that this should have tests. Thoughts?

If the work remaining is as suggested above, I believe that this will take more than the 2 hour estimate to complete, which is one reason why I am requesting a look over before designing most of the issue.

This is the work of me and @wallrj .

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -76,8 +76,6 @@ Running Tests
 
 You can run all unit tests by doing:
 
-.. Document the FLOCKER_TEST_NO_SKIPS environment variable here
-
 .. code-block:: console
 
    $ tox
@@ -91,6 +89,8 @@ In addition, ``tox`` needs to be run as root:
 
 Since these tests involve global state on your machine (filesystems, ``iptables``, Docker containers, etc.) we recommend running them in the development Vagrant image.
 
+The ``FLOCKER_TEST_NO_SKIPS`` environment variable can be used to specify tests with which should be counted as failing if they are skipped.
+For example, ``FLOCKER_TEST_NO_SKIPS=flocker.acceptance tox`` will cause test failures if the acceptance tests are skipped.
 
 Documentation
 =============
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -89,8 +89,8 @@ In addition, ``tox`` needs to be run as root:
 
 Since these tests involve global state on your machine (filesystems, ``iptables``, Docker containers, etc.) we recommend running them in the development Vagrant image.
 
-The ``FLOCKER_TEST_NO_SKIPS`` environment variable can be used to specify tests with which should be counted as failing if they are skipped.
-For example, ``FLOCKER_TEST_NO_SKIPS=flocker.acceptance tox`` will cause test failures if the acceptance tests are skipped.
+The ``FLOCKER_TEST_FAIL_IF_SKIPS`` environment variable can be used to specify tests with which should be counted as failing if they are skipped.
+For example, ``FLOCKER_TEST_FAIL_IF_SKIPS=flocker.acceptance tox`` will cause test failures if the acceptance tests are skipped.
 
 Documentation
 =============
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -76,6 +76,8 @@ Running Tests
 
 You can run all unit tests by doing:
 
+.. Document the FLOCKER_TEST_NO_SKIPS environment variable here
+
 .. code-block:: console
 
    $ tox
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -91,6 +91,7 @@ Since these tests involve global state on your machine (filesystems, ``iptables`
 
 The ``FLOCKER_TEST_FAIL_IF_SKIPS`` environment variable can be used to specify tests with which should be counted as failing if they are skipped.
 For example, ``FLOCKER_TEST_FAIL_IF_SKIPS=flocker.acceptance tox`` will cause test failures if the acceptance tests are skipped.
+To support this environment variable, all tests which need skip logic must use either ``flocker.testtools.skips.skipIf``, ``flocker.testtools.skips.skipUnless`` or ``flocker.testtools.skips.SkipTest``.
 
 Documentation
 =============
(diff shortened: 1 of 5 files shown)


flocker/testtools/skips.py
```diff
@@ -0,0 +1,45 @@
...
+
+    :param test_object: test method or class with a skip decorator.
+
+    :return: True if ``test_object`` is covered by environment variable
+        FLOCKER_TEST_NO_SKIPS, else False.
+    """
+    return False
+
+
+def skipUnless(condition, reason):
```
adamtheturtle: These functions need suitable names. Ideally those names would read well, like how `skipUnless(SELENIUM_INSTALLED)` reads like "skip unless selenium is installed". One option is to have them called `skipIf` and `skipUnless`, like the unittest methods, and then only import lines must be changed. Another option is `skipOrFailUnless` and `skipOrFailIf`.

itamarst:
I would suggest scoping this to only cover *one* way of skipping, and then telling people This Is The Only Way You May Skip in Flocker.

CONTRIBUTING.rst
```diff
@@ -89,6 +89,8 @@ In addition, ``tox`` needs to be run as root:
 
 Since these tests involve global state on your machine (filesystems, ``iptables``, Docker containers, etc.) we recommend running them in the development Vagrant image.
 
+The ``FLOCKER_TEST_NO_SKIPS`` environment variable can be used to specify tests with which should be counted as failing if they are skipped.
```
itamarst: Maybe... `FLOCKER_TEST_FAIL_IF_SKIPS`.

adamtheturtle:
@itamar on the JIRA issue you asked for example usage. This branch now has a set of sample tests. When running `trial flocker.sample` all of these tests fail if `fail_if_skipped` returns `True`, but skip if it returns `False`.

> I would suggest scoping this to only cover one way of skipping, and then telling people This Is The Only Way You May Skip in Flocker.

Because the decorators are very similar to one another, they are both included in this design. As for the other three options (`raise SkipTest(reason)`, `test_case.skipTest(reason)` and  `test_case.skip = reason`), I see no problem with telling people to only use `raise SkipTest`, however this design includes passing `SkipTest` a test_item which it can then fail or skip. Perhaps it is possible to work this out without the parameter.

Note that my comment on an "outdated diff" is still a relevant question about naming these things https://github.com/ClusterHQ/flocker/pull/1050#discussion_r22856191.

flocker/testtools/skips.py
```diff
@@ -0,0 +1,75 @@
...
+    """
+    Check whether to fail the given test object if it will be skipped.
+
+    :param test_object: test method or class with a skip decorator.
+
+    :return: True if ``test_object`` is covered by environment variable
+        FLOCKER_TEST_FAIL_IF_SKIPS, else False.
+    """
+    test_fail_if_skips = environ.get("FLOCKER_TEST_FAIL_IF_SKIPS")
+    return True
```
adamtheturtle: This can be changed to see the effect, when running `trial flocker.sample` of changing the `FLOCKER_TEST_FAIL_IF_SKIPS` environment variable once this is implemented.

The design is that running:
`FLOCKER_TEST_FAIL_IF_SKIPS=flocker.sample trial flocker.sample` would cause all the skipping tests in that module to fail, while running `FLOCKER_TEST_FAIL_IF_SKIPS=flocker.another_module trial flocker.sample` would cause no tests to fail.

flocker/sample/test_sample.py
```diff
@@ -0,0 +1,33 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Sample tests for skipping.
+"""
+from flocker.testtools.skips import skipUnless, skipIf, SkipTest
+from twisted.trial.unittest import TestCase
+
+skip_this_test_with_skipUnless = skipUnless(
```
itamarst: How would I specify that I want to not skip this?

itamarst: I.e. what would the env variable look like.

flocker/sample/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
exarkun: Please move this package somewhere else.

flocker/sample/test_sample.py
adamtheturtle: https://github.com/ClusterHQ/flocker/pull/1050/files#r23014698 gave an example of the environment variable being used.

It is not that the skip is ignored, however. I thought that debugging would be much easier if the test failed when it was otherwise meant to skip.

flocker/sample/__init__.py
adamtheturtle: This won't be in the final code for sure, and it probably won't be this but moved. If this design is deemed appropriate then there can be tests for the test tools which do something similar to eyeballing the results of these tests with different variables supplied.

itamarst:
The internet here, I *think*, is that you skip by module or TestCase class? It's not really clear.

tomprince:
I think that the when picking what skips to allow, it should be by abstract identifier, rather than by test-case name. I think the motivating example for this originally was `flocker.testtools.skip_on_broken_permissions` and `if_root`, both of which occur sporadically throughout the codebase. So rather than giving a package name, the environment variable should specify something like `broken_permissions` or `need_root` or something.

Also, while the default should be for skips to not cause failures, if this feature is being used ignoring a skip should opt-in, rather than opt-out. The reasoning is, if a new skip is added, it should cause a failure by default on buildbot. So, looking at the current buildbot results, the only thing that should be skipped is the acceptance tests, so that should be the only thing that needs to be listed, as opposed to all the various other reasons for skips.

tomprince:
We will eventually need some prose documentation for this. I think writing it now, as part of the design would make it clear what is intended. Please write some, and resubmit for review.

itamarst:
Re-open if/when we return to this.

PR #1046, opened 2015-01-09
[FLOC-1162] Volume API version endpoint

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1162

PR #1052, opened 2015-01-13
[FLOC-1162] Volume API version endpoint

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1162

This is a copy of the branch from #1046, which was accidentally merged and then reverted.


flocker/volume/schema/endpoints.yml
```diff
@@ -0,0 +1,11 @@
+$schema: http://json-schema.org/draft-04/schema#
+id: http://api.clusterhq.com/v1/endpoints.json
+definitions:
+  versions:
```
itamarst: This isn't really tested, i.e. we should have tests validating the schema does what we think it does. We had some in HybridCluster you can use as basis.

setup.py
```diff
@@ -94,6 +94,9 @@ def run(self):
             'env-docker/*',
             'retry-docker/*'
         ],
+        # These data files are used by the volumes API to define input and
```
itamarst: Is this necessary if it's already in MANIFEST.in?

flocker/volume/schema/endpoints.yml
wallrj: Done. And this allowed us to simplify and improve the versions schema.

setup.py
wallrj: Yes, it seemed to be. 
Adding the files to MANIFEST ensures that they are included in an sdist package.
Adding to `package_data` ensures that those files are actually installed.

flocker/volume/httpapi.py
```diff
@@ -10,27 +13,41 @@
...
+class DatasetAPIUserV1(object):
     """
     A user accessing the API.
     """
     app = Klein()
 
-    @app.route("/noop")
-    @structured({}, {})
-    def noop(self):
+    @app.route("/version")
```
adamtheturtle: Use `@userDocumentation` as per http://doc-dev.hybridcluster.net/public-api.html

wallrj: I'll leave that for the followup api documentation ticket https://clusterhq.atlassian.net/browse/FLOC-101

flocker/volume/test/test_httpapi.py
```diff
@@ -15,26 +15,27 @@
...
+from ..httpapi import DatasetAPIUserV1, create_api_service
+from ... import __version__
 
 
 class APITestsMixin(object):
     """
     Integration tests for the Dataset Manager API.
     """
-    def test_noop(self):
+    def test_version(self):
```
adamtheturtle: If following the HybridCluster guide, this should perhaps be in `class VersionTestsMixin` or similar.

> Tests should be grouped together by endpoint and purpose. For example, all of the tests for the website creation endpoint should be defined in one class.

flocker/volume/httpapi.py
wallrj: Actually, that was the wrong link. 
The REST API documentation infrastructure has been ported in https://clusterhq.atlassian.net/browse/FLOC-921
And there's a followup issue for figuring out how to use it https://clusterhq.atlassian.net/browse/FLOC-1171
So I'll add a note there and if that's not the right place, we can create a separate issue.

flocker/volume/test/test_httpapi.py
wallrj: It can be refactored later if necessary.

setup.py
wallrj: And it was the same in HybridCluster too.

PR #1054, opened 2015-01-15
[FLOC-1219] Include API docs - design sketch

itamarst:
Basically works; for final branch:

1. Output still looks a little iffy.
2. Had to add another monkey-patch, alas. Not sure if there's other solution though.
3. Support for non-referenced (inline) properties isn't tested.


exarkun:
Thanks.  Seems fine.  Go ahead.

PR #1056, opened 2015-01-15
yumdownloader and createrepo and gsutil are all Software requirements.

exarkun:
Fixes FLOC-1190

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -32,8 +32,6 @@ Software
 - An up-to-date clone of the `Flocker repository <https://github.com/ClusterHQ/flocker.git>`_.
 - An up-to-date clone of the `homebrew-tap repository <https://github.com/ClusterHQ/homebrew-tap.git>`_.
 - `gsutil Python package <https://pypi.python.org/pypi/gsutil>`_.
-- ``yumdownloader`` (from the Fedora or Ubuntu ``yum-utils`` package)
-- ``createrepo`` (from the Fedora or Ubuntu ``createrepo`` package)
 
 Access
 ~~~~~~
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -38,7 +38,12 @@ Access
 
 - A Read the Docs account (`registration <https://readthedocs.org/accounts/signup/>`_),
   with `maintainer access <https://readthedocs.org/dashboard/flocker/users/>`_ to the Flocker project.
-- Access to `Google Cloud Storage`_ using `gsutil`_ on your workstation.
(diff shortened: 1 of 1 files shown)


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -31,6 +31,9 @@ Software
 - A web browser.
 - An up-to-date clone of the `Flocker repository <https://github.com/ClusterHQ/flocker.git>`_.
 - An up-to-date clone of the `homebrew-tap repository <https://github.com/ClusterHQ/homebrew-tap.git>`_.
+- `gsutil <https://pypi.python.org/pypi/gsutil>`_
```
tomprince: This should be installed via the `release` extra.

I guess perhaps this is for outside of a VM. In which case https://clusterhq.atlassian.net/browse/FLOC-1036 would be the correct solution.

tomprince:
There appears to be some confusion about where the requirements need to be installed. Please clear that up, and resubmit for review.

adamtheturtle:
I have clarified that `gsutil` must be installed on your workstation, and asked on the JIRA issue whether the other tools are needed only for http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/release-process.html?highlight=release#appendix-pre-populating-rpm-repository, in which case they should be preinstalled on the Flocker development machine. If not, and they are needed on the local workstation, then using Linux to do the release is a requirement.

adamtheturtle:
As per the latest comment on the JIRA issue, I think that only a part of this spec is necessary, and have committed what I think is necessary.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -31,13 +31,14 @@ Software
...
 - An up-to-date clone of the `homebrew-tap repository <https://github.com/ClusterHQ/homebrew-tap.git>`_.
+- `gsutil Python package <https://pypi.python.org/pypi/gsutil>`_ on your workstation.
 
 Access
 ~~~~~~
 
 - A Read the Docs account (`registration <https://readthedocs.org/accounts/signup/>`_),
   with `maintainer access <https://readthedocs.org/dashboard/flocker/users/>`_ to the Flocker project.
-- Access to `Google Cloud Storage`_ using `gsutil`_.
+- Access to `Google Cloud Storage`_ using `gsutil`_ on your workstation.
```
tomprince: I think it needs to be configured both on the workstation, and in the VM.

tomprince: Specifically the upload-rpms script uses gsutil, which is run in the VM.

tomprince:
There is a comment part way through the process saying that gsutil needs to be installed on the workstation. That comment can perhaps be removed, now that it is a duplicate of the one at the beginning.

tomprince:
Address the two minor points, and merge.

PR #1057, opened 2015-01-15
Remove python-flocker.spec support code

exarkun:
Fixes FLOC-1221


setup.py
```diff
@@ -112,9 +85,7 @@ def run(self):
     },
 
     install_requires=[
-        # Any changes here must be reflected in ``python-flocker.spec.in`` so
-        # that RPM dependencies match.
-        # They must also be reflected in the yumdownloader lines in
+        # Any changes here must be reflected in the yumdownloader lines in
```
tomprince: We shouldn't need any ``yumdownloader`` lines for python dependencies, for omnibus builds. Only non-python dependencies.

setup.py
```diff
@@ -16,35 +16,8 @@
...
-    def initialize_options(self):
-        pass
-    def finalize_options(self):
-        pass
-    def run(self):
-        with open('python-flocker.spec.in', 'r') as source:
-            spec = source.read()
-
-        flocker_version = versioneer.get_version()
-        version, release = make_rpm_version(flocker_version)
```
tomprince: We should probably file a ticket about making this use ``characteristic`` instead of ``namedtuple``, now that it isn't being used in `setup.py`.

tomprince: The mention that this is used in `setup.py` should be removed now, though.

tomprince: https://clusterhq.atlassian.net/browse/FLOC-1223

setup.py
tomprince: I guess that means we should adjust the lines there to reflect that fact too.

tomprince:
Please make the above changes and re-submit.

setup.py
exarkun: All of the packages used with the `yumdownloader` commands in the `Pre-populating RPM Repository` currently appear to be Python packages.  Does that mean we don't currently need any of the packages in that RPM repository and until we introduce a non-Python dependency we won't need to do any pre-populating?

exarkun: Okay I don't really know what to do here.  It seems like we could drop the whole `Pre-populating RPM Repository` section right now and not be any worse off.  Perhaps at the cost of having to re-introduce these steps later if we gain a non-standard dependency.  We've managed to avoid that so far, though, it seems...?

tomprince: We do have some zfs packages in there unfortunately Which I guess should probably be listed there, but aren't currently. :frowning: That is my fault, I guess. Although, those aren't the steps that were used to upload those zfs packages. I guess that means that section can probably just go.

tomprince:
Merge once the build is green.

wallrj:
We forgot to remove the documentation for distribution maintainers:
 * http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/packaging.html#instructions-for-distribution-maintainers

ryao:
@wallrj We still haven't done that.

ryao:
For those from the future reading this, the command you want is likely `python setup.py bdist_rpm`.

exarkun:
Nope.  `admin/build-package`

PR #1059, opened 2015-01-16
[FLOC-1125] Disable linkcheck.

tomprince:
Note that the dev-box build should be passing after this lands.

https://clusterhq.atlassian.net/browse/FLOC-1225

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -148,6 +148,8 @@ Preparing For a Release
 
    Go to the `BuildBot web status`_ and force a build on the just-created branch.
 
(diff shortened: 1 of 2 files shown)


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -148,6 +148,8 @@ Preparing For a Release
 
    Go to the `BuildBot web status`_ and force a build on the just-created branch.
 
+   In addition, review the link-check step of the documentation builder to ensure that all the errors are expected.
```
tomprince: This makes me sad, but a red builder makes me sadder.

I think the trade-off of doing this during the release is worth it for the moment. Long-term, we should fix the link-check. (I guess a ticket should be filed).

PR #1060, opened 2015-01-16
[FLOC-1195] Fix test_moving_data.

tomprince:


admin/acceptance.py
```diff
@@ -106,6 +106,13 @@ def start_nodes(self):
             cwd=self.vagrant_path.path,
             env=extend_environ(FLOCKER_BOX_VERSION=box_version))
 
+        from flocker.provision._install import run, task_pull_docker_images
```
adamtheturtle: Why is this import here and not at the top of the file?

admin/acceptance.py
```diff
@@ -106,6 +106,13 @@ def start_nodes(self):
...
+        for node in self.NODE_ADDRESSES:
+            run(
+                username='root',
+                address=node,
+                commands=task_pull_docker_images()
```
adamtheturtle: This, and I guess other uses of ``task_pull_docker_images``, differ from the docs. In the documented example of how to move data, at http://doc-dev.clusterhq.com/gettingstarted/tutorial/volumes.html, the images are not pulled.

If we are only guaranteeing (via tests) that these commands work if `docker pull` is used first, then the docs should have those pull commands. If the docs don't say to pull the images first then another option is to increase the test timeout.

adamtheturtle:
Ah I saw this was marked "Code review ready" before noticing that it is not in Sprint 4.

I was going to say:
> Please address the comments to make the tests match the docs, and then resubmit.

But I guess that the first step is to write in FLOC-1195 why it should be in the current sprint, if you think it should.

admin/acceptance.py
tomprince: But each of the other examples do including pulling the image before hand.

http://doc-dev.clusterhq.com/gettingstarted/examples/environment.html#download-the-docker-image

I think (given an infinite, or longer timeout), all the tests would pass without this (but I haven't tested that). I think the only reason only one test fails due to this, is that only one test happens to trigger twisted timeout code. (I think it only happens when you return a deferred, and only for the time that deferred takes; the tests are mostly synchronous).

Perhaps it makes sense to move this logic elsewhere in the acceptance tests, but I'm not entirely sure where the best place is for it. (And if that is the case, all the other providers should get the same change). I can file an issue for doing this, but I think this fix is sufficient for the moment.

itamarst: I'm fine with this differeing marginally from docs on presumption that a green buildbot is more important than testing purity, and that Docker doing implicit pulls isn't going to stop happening anytime soon. Increasing timeouts is pretty terrible, we want tests to fail quickly.

admin/acceptance.py
```diff
@@ -106,6 +106,14 @@ def start_nodes(self):
             cwd=self.vagrant_path.path,
             env=extend_environ(FLOCKER_BOX_VERSION=box_version))
 
+        # Import this here, since `run` isn't a good generic name.
+        from flocker.provision._install import run, task_pull_docker_images
```
itamarst: How about putting imports at top and doing `import run as run_on_provisioned_node`?

admin/acceptance.py
itamarst: `ACCEPTANCE_IMAGES` ought to be in `flocker.acceptance` somewhere, and therefore presumably not implicit in the provisioning code.

itamarst:
I'd accept this, except the acceptance tests never even ran... Address above, re-run, and if they pass you can hit Accept.

admin/acceptance.py
tomprince: https://clusterhq.atlassian.net/browse/FLOC-1219

PR #1061, opened 2015-01-16
Add a test for ResizeVolume.run.

exarkun:
Fixes FLOC-1231.

### DIFF
--- a/flocker/node/test/test_deploy.py
+++ b/flocker/node/test/test_deploy.py
@@ -2512,6 +2512,48 @@ class CreateVolumeTests(SynchronousTestCase):
             _to_volume_name(u"myvol")))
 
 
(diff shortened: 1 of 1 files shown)


flocker/node/test/test_deploy.py
```diff
@@ -2512,6 +2512,48 @@ def test_return(self):
             _to_volume_name(u"myvol")))
 
 
+class ResizeVolumeTests(SynchronousTestCase):
```
itamarst: Oh, uh, this shouldn't be SynchronousTestCase if you're reutrning a Deferred...

exarkun: Argh.  Thanks.  TDD fail. :cry:

PR #1063, opened 2015-01-16
[FLOC-1219] Include REST API docs

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1219

### DIFF
--- a/docs/_themes/bootstrap/static/sphinx-bootstrap.css
+++ b/docs/_themes/bootstrap/static/sphinx-bootstrap.css
@@ -327,3 +327,9 @@ a.download:before {
 .sidebar-heading {
     font-size: 24px;
 }
+
+/* Necessary for "Request JSON Schema" etc. to look right in generated REST API
+   documentation. */
+th.field-name {
+    text-align: left;
+}
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -1,0 +1,39 @@
+# The whole file is a list of examples.  Unless otherwise indicated, examples
+# are independent of each other.  For complicated examples that involve state
+# that can't be established using a single HTTP request, requirements (in the
+# form of other examples) can be declared.
+#
+# Some variable substitution is done to some of the content in this file.  The
+# details of what and how are explained below but here is the list of variables
+# that are globally defined and always available to be substituted:
+#
+# DOMAIN - The domain name of the cluster API server.
+
+-
+  # Every element must have a unique identifier.  Documentation can use these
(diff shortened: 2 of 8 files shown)


flocker/restapi/docs/publicapi.py
```diff
@@ -420,7 +407,13 @@ def run(self):
 
         appContainer = namedAny(self.arguments[0])
 
-        examples_path = FilePath(self.options["examples_path"])
+        # For now we ignore this and just hardcode, due to
+        # https://github.com/sphinx-doc/sphinx/issues/1685
+        # examples_path = FilePath(self.src).preauthChild(
+        #     options["examples_path"]))
```
tomprince: I haven't tested it, but I think ``self.state_machine.get_source(self.lineno)`` will get the right file.

docs/advanced/api_examples.yml
```diff
@@ -0,0 +1,39 @@
...
+  # validation.  A User-Agent should automatically be inserted.  %(VARIABLES)s
+  # in the request body get values substituted in for them.
+  request: |
+    GET /v1/version HTTP/1.1
+
+  # The other half of the example is the HTTP response you'll get if you do
+  # this.  Ditto about the harness translating pseudo-HTTP into real HTTP.
+  # %(VARIABLES)s are substituted in the response body as well.
+  response: |
+    HTTP/1.1 200 Created
```
itamarst: Should be `200 OK`

docs/_themes/bootstrap/static/sphinx-bootstrap.css
```diff
@@ -327,3 +327,9 @@ a.download:before {
 .sidebar-heading {
     font-size: 24px;
 }
+
+/* Necessary for "Request JSON Schema" etc. to look right in generated REST API
+   documentation. */
```
exarkun: TYVM for this comment.

flocker/restapi/docs/publicapi.py
exarkun: Looks like it does the right thing.  To make it easier to use this, I've also changed the value expected by the `examples_path` option.  It's relative to the source file it's in, now, instead of relative to the root of the repository.

exarkun:
Another issue is that the API framework applies its generic success/failure wrapping around results.  That makes the docs generated here (or the source material) not quite correct.

I think we should ignore that until we're automatically verifying the docs, though.  Then we'll have something showing us what examples are wrong and we can decide how we want to fix them (perhaps by removing that wrapping layer and using HTTP response codes, for example).

PR #1066, opened 2015-01-20
[FLOC-1159] Configuration persistence

itamarst:
Originally this involved the current configuration format, and there was an endpoint that returned current configuration in that format, so a bunch of changes relevant to that were made. I ripped most of that out since we decided to switch to (it's so terrible but it's very temporary) pickle, but left some of it in since it's generally a good idea. E.g. config and model really do belong in flocker.control.


flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Persistence of cluster configuration.
+"""
+
+from cPickle import dumps, loads
```
exarkun: Probably no benefit to using cPickle over pickle.  Probably only obscure downsides to using cPickle (bad profiler interaction, bad exception reporting, slightly greater chance of memory corruption, etc).  So not a big deal, but it's easy to switch. :smile:

flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
...
+
+from twisted.application.service import Service
+
+from ._model import Deployment
+
+
+# These should not use Pickle!@!
+# https://clusterhq.atlassian.net/browse/FLOC-1241x
+def serialize_deployment(deployment):
```
exarkun: Perhaps these are both private implementation details of `ConfigurationPersistenceService` (iow - underscore prefix?)

flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
...
+
+class ConfigurationPersistenceService(Service):
+    """
+    Persist configuration to disk, and load it back.
+
+    :ivar Deployment _deployment: The current desired deployment configuration.
+    """
+    def __init__(self, reactor, path):
+        """
+        :param reactor: Reactor to use for thread pool.
```
exarkun: There's no thread pool here yet though.

flocker/control/_persistence.py
itamarst: They're public on purpose, so https://clusterhq.atlassian.net/browse/FLOC-1214 can use an API that isn't pickle.

flocker/control/_persistence.py
itamarst: Oh yeah, I was going to file another issue, right. Unless you want this done in this branch.

flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def save(self, deployment):
+        """
+        Save and flush new deployment to disk.
+
+        :return Deferred: Fires when write is finished.
+        """
+        self._sync_save(deployment)
+        self._deployment = deployment
+        # Switch to thread at some point in future:
```
exarkun: Maybe?  We might not use local file storage long enough to get to the point where we care about blocking for a few milliseconds on the local disk.  Maybe instead the informative comment here would be one about how this might involve synchronizing state with remote systems participating in a distributed configuration tracking system.

flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
...
+        from twisted.internet.defer import succeed
```
exarkun: This can move to the top.

flocker/control/_persistence.py
```diff
@@ -0,0 +1,86 @@
...
+        return succeed(None)
+
+    def get(self):
+        """
+        :return Deployment: The current desired configuration.
+        """
+        return self._deployment
```
exarkun: Hm.  This makes me wish we used `apply_immutable` more consistently throughout the model.  Maybe it doesn't matter.  We're not going to let anyone mutate `Deployment` instances right?

Alternatively, `get` could re-load the deployment on every call.

Up to you.

flocker/control/test/test_httpapi.py
```diff
@@ -23,19 +26,44 @@ class APITestsMixin(object):
     """
     Integration tests for the Dataset Manager API.
     """
-    def test_version(self):
+    def initialize(self):
```
exarkun: I guess this isn't `setUp` because `buildIntegrationTests` doesn't have a cooperative `setUp`?  Sad.  Wish we had proper fixture support.  Oh well.

flocker/control/test/test_persistence.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_empty_on_start(self):
+        """
+        If no configuration was previously saved, starting a service results
+        in an empty ``Deployment``.
+        """
+        service = self.service(FilePath(self.mktemp()))
+        self.assertEqual(service.get(), Deployment(nodes=frozenset()))
+
+    def test_directory_is_created(self):
```
exarkun: What if the directory does exist?

flocker/control/test/test_persistence.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        If a directory does not exist in given path, it is created.
+        """
+        path = FilePath(self.mktemp())
+        self.service(path)
+        self.assertTrue(path.isdir())
+
+    def test_file_is_created(self):
```
exarkun: What if the file does exist?

flocker/control/test/test_persistence.py
```diff
@@ -0,0 +1,89 @@
...
+        """
+        service = self.service(FilePath(self.mktemp()))
+        d = service.save(TEST_DEPLOYMENT)
+        d.addCallback(lambda _: service.get())
+        d.addCallback(self.assertEqual, TEST_DEPLOYMENT)
+        return d
+
+    def test_persist_across_restarts(self):
+        """
+        A configuration that was saved can be loaded from a new service.
```
exarkun: Hm.  This creates a second service instance with the same path so I guess it exercises the case where things exist already.

exarkun:
Thanks.  Looks pretty good as far as it goes.  Just minor comments inline.  Apart from that, acceptance tests are failing.  Please fix and re-submit.

flocker/control/test/test_httpapi.py
itamarst: Yeah, it happens in wrong order.

flocker/control/_persistence.py
itamarst: I'll add comment about not mutating, but yeah, we want to switch to persistent data structures.

flocker/control/_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+from pickle import dumps, loads
+
+from twisted.application.service import Service
+from twisted.internet.defer import succeed
+
+from ._model import Deployment
+
+
+# These should not use Pickle!@!
+# https://clusterhq.atlassian.net/browse/FLOC-1241x
```
tomprince: trailing `x`

flocker/control/_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+    def startService(self):
+        if not self._path.exists():
+            self._path.makedirs()
+        self._config_path = self._path.child(b"current_configuration.json")
+        if self._config_path.exists():
+            self._deployment = deserialize_deployment(
+                self._config_path.getContent())
+        else:
+            self._deployment = Deployment(nodes=frozenset())
+            self._sync_save(self._deployment)
```
tomprince: I'm curious how this will be implemented once saving isn't synchronous, but involved network communication.

flocker/control/_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+        :param reactor: Reactor to use for thread pool.
+        :param FilePath path: Directory where desired deployment will be
+            persisted.
+        """
+        self._path = path
+
+    def startService(self):
+        if not self._path.exists():
+            self._path.makedirs()
+        self._config_path = self._path.child(b"current_configuration.json")
```
tomprince: json?

flocker/control/_persistence.py
```diff
@@ -0,0 +1,92 @@
...
+    Convert a ``Deployment`` object to ``bytes``.
+
+    :param Deployment deployment: Object to serialize.
+
+    :return bytes: Serialized object.
+    """
+    return dumps(deployment)
+
+
+def deserialize_deployment(data):
```
tomprince: Do we want to do any version checking here? I know we strongly suggest we don't provide compatability, but I'm certain that people will try anyway. It would be nice to error instead of carrying on with possibly invalid data.

itamarst: Hm. I feel like we should be doing FLOC-1214 ASAP (we'll be using pickle over the network otherwise :fearful:), which will solve that. Only reason I didn't do it first is that it's easily parallelizable with the rest of the work, so didn't seem worth making it a bottleneck.

flocker/control/_persistence.py
itamarst: Worry about it when we get there. Maybe I'll finally have a reason to finish that dependency injection framework.

flocker/control/_persistence.py
tomprince: A simple expedient solution is just to store json with a version and a pickle field, and check that on import.

itamarst: We can do that if it becomes an issue, but this is not code anyone can run yet so it's not worth worrying about.

tomprince:
Fix the link and filename. Consider adding some simple version checking.
If you do, please submit for a quick review (of just that code). Otherwise, merge after buildbot is green.

flocker/control/_persistence.py
exarkun: The other loading and saving code can always get smarter about handling "no configuration exists at all" cases.  It might not even need to be that special since we may end up using an API that just gives back an empty result when there is nothing saved (instead of, say, exploding - like attempting to read a non-existent file would).

PR #1067, opened 2015-01-20
[FLOC-1181] Unattached manifestations design sketch

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1181
Wooo :ghost:

Depends on FLOC-1159... or at least, I thought it did? Anyway it's a diff against that, fine for reviewing.


flocker/control/_model.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from characteristic import attributes, Attribute
-from pyrsistent import pmap
+from pyrsistent import pmap, pset
```
exarkun: is pset better than frozenset?

itamarst: It has an `add()` method? But who knows maybe not.

flocker/node/_deploy.py
```diff
@@ -420,8 +420,8 @@ def applications_from_units(result):
                 running=running,
                 not_running=not_running,
                 used_ports=self.network.enumerate_used_ports(),
-                # FLOC-1181 will want to add here the manifestations that
-                # are not attached to any application.
+                # FLOC-1181 will want to add here the remaining
+                # manifestations that are not attached to any application.
```
exarkun: That is to say, all the manifestations that the above loop didn't consume from `available_manifestations`?

flocker/control/_model.py
exarkun: That's probably useful eventually.  I'm happy to see pset used but I guess I'll be sad if we randomly use frozenset in some places and pset in others.  Maybe just stick with frozenset until pset is actually a win for something?

exarkun:
Thanks, looks good, please proceed.

PR #1068, opened 2015-01-20
[FLOC-1186] New docker-py

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1186

Skipped design review phase since there's nothing else left to do.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -13,6 +13,7 @@ from zope.interface import Interface, implementer
 
 from docker import Client
 from docker.errors import APIError
+from docker.utils import create_host_config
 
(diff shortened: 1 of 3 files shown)


wallrj:
Thanks @itamarst 

Some of the tests are failing which use the docker-py `Client` directly. `Client` seems to be using the 1.16 API version by default now so we need to force the older API version in those tests. I might quickly fix that and push so that we can see if there are any other build failures.

Not sure if we're still maintaining the `python-flocker.spec` file? If yes, update that too and ask @tomprince to package the new docker-py.

Otherwise this looks fine. Fix those tests and then merge.

PR #1070, opened 2015-01-21
Sketch of protocol design.

exarkun:


flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+        pass
+
+    def toString(self, node_state):
+        #return json.dumps(marshal_configuration(node_state))
+        pass
+
+
+# XXX This should suffice for both both communicating global current state
+# and the deployment configuration from control service to convergence
+# agent. We don't need to transfer used ports because the control service
```
exarkun: We might need to transfer used ports because ports might be used for some reason other than that they've been assigned to a Flocker-managed Docker container.  For example, port 22 is **probably** in use on any node you're going to deploy to so the control service should reject configuration changes that try to publish that port from a container.  I don't know of a good alternative to having the nodes report this information in order to handle this case.

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+class DeploymentArgument(Argument):
+    """
+    AMP argument that takes a ``Deployment`` object.
+    """
+    def fromString(self, in_bytes):
+        #values = json.loads(in_bytes))
+        #deployment_config = values["deployment"]
+        #application_config = values["applications"]
+        # XXX this is wrong function cause _config is terrible, but basic idea:
+        #return deployment_from_configuration(deployment_config, application_config)
```
exarkun: This all gets replaced by the equivalent FLOC-1159-introduced deserialization API I expect?

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+class VersionCommand(Command):
+    """
+    Return configuration protocol version of the control service.
+
+    Semantic versioning: Major version changes implies incompatibility,
+    minor version implies compatible extension.
+    """
+    arguments = []
+    response = [('major', Integer()),
+                ('minor', Integer())]
```
exarkun: Hm.  When will be bump minor version and why?  Does this really matter?  A decision might be made based on the value of the major version: your major version does not match mine, I will not try to talk to you.  What decision might be made on the minor version?

I suspect we should skip minor version for the time being.

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+                 ('state', DeploymentArgument())]
+    response = []
+
+
+class NodeStateCommand(Command):
+    """
+    Used by a convergence agent to update the control service about the
+    status of a particular node.
+    """
+    arguments = [('hostname', String()),
```
exarkun: At some point we should definitely replace hostname with a better node identifier.  But one step at a time... yea.

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+    def cluster_updated(self, configuration, state):
+        self.agent.cluster_updated(configuration, state)
+
+
+# In addition to the above I would need to implement IConvergenceAgent, a
+# testing-oriented IConvergenceAgent implementation, and minimal
+# ClusterState (the latter would be just enough to test, would be fleshed
+# out in later issues).
+
+# Might end up porting tiny bit of AMP testing infrastructure from HybridCluster too.
```
exarkun: There might be some minor useful bit of testing infrastructure in lp:game... Not much, it's sort of a primitive (though thorough) approach to AMP testing.  `CommandTestMixin` might be worth something, though.

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    def __init__(self, agent):
+        """
+        :param IConvergenceAgent agent: Convergence agent to notify of changes.
+        """
+        self.agent = agent
+
+    @ClusterStatusCommand.responder
+    def cluster_updated(self, configuration, state):
+        self.agent.cluster_updated(configuration, state)
```
exarkun: I guess two blockers for doing much more in this direction are having an agent type with a `cluster_updated` method and having a cluster type with a `node_changed` method?

flocker/node/_protocol.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    Used by a convergence agent to update the control service about the
+    status of a particular node.
+    """
+    arguments = [('hostname', String()),
+                 ('node_state', NodeStateArgument())]
+    response = []
+
+
+class ControlServiceLocator(CommandLocator):
```
exarkun: The AMP protocol definitions will be shared between control service and convergence agents.  So the code will run in two different places.  Which maybe suggests it belongs in a package we're explicitly sharing between nodes and control service.  Could just be `flocker.common` I guess, though I wouldn't mind an explicit protocol package either.

Probably suggests there will be some explicit interfaces too, to avoid making `flocker.common` depend on `flocker.node` or `flocker.control` for the ... business logic ... types.

exarkun:
Thanks.  Looks like a good start.  I think it should be updated to reflect what's happened elsewhere meanwhile and some code should shuffle around a little bit.  Please submit again once that's happened.

exarkun:
Oh.  Also, I think it would be nice to (eventually) include a high-level view of the intended interactions along with this.  Can just be module docstring or something, not user-facing docs (since this protocol is internal, after all).  That might make a good part of the design since it will probably do a better job of conveying the overall idea here than is done by the individual little bits and pieces.

flocker/node/_protocol.py
itamarst: The agents can report this information to the control service, that's what the NodeState argument and NodeStateCommand are for. My point was the that the control service doesn't need to tell agents about used ports.

flocker/node/_protocol.py
itamarst: Yeah.

flocker/node/_protocol.py
itamarst: OK.

flocker/node/_protocol.py
exarkun: Ah yes, I see.

flocker/node/_protocol.py
itamarst: Well... I was thinking I'd do them after this one? And just have test fakes in this branch. Because e.g. the thing you want the control service `node_changed` to do is tell the protocol to send new information to al the other nodes...

flocker/node/_protocol.py
itamarst: How about we just move this module to `flocker.control`? It seems reasonable to me to have convergence agents borrow some code from the control service package.

itamarst:
OK, addressed or commented as relevant.

flocker/node/_protocol.py
exarkun: My concern here is that without implementing `cluster_updated` it's hard to tell if this is the right interface or not.  I guess if the interface isn't quite right we can fix it later...

This might be the kind of thing where we'd benefit from a quick spike or just a more fleshed-out design (ie, what is cluster_updated actually going to do?).  Then we'd be in a better position to look at a subset of things and say they make sense (or not).

Up to you - if you have a good idea of what's going to go into these methods, go ahead with this.  If you're not quite sure, maybe do a spike (maybe not even a working one) to find out.

flocker/node/_protocol.py
exarkun: That's probably okay for now.  In general the tricky thing there might be dependency management - `flocker.control` might end up with some dependencies that are both irrelevant to the convergence agents and worth actually not installing (I'm thinking of zookeeper).  But we haven't gotten there yet, so not a big deal.

exarkun:
Thanks.  Basically good.  I'm going to re-assign the JIRA issue to you to click "accept" or "reject" on, based on what you think about that comment about `cluster_updated` et al.

itamarst:
FLOC-1255 (PR #1082) suggests that the notifications for the agent should also include connected and disconnected events. So might want to do the protocol after 1255 is done.

PR #1071, opened 2015-01-21
Use options supported by OS X.

exarkun:
Fixes FLOC-1184

### DIFF
--- a/flocker/common/_ipc.py
+++ b/flocker/common/_ipc.py
@@ -112,12 +112,16 @@ class ProcessNode(object):
             b"-o", b"StrictHostKeyChecking=no",
             # The tests hang if ControlMaster is set, since OpenSSH won't
             # ever close the connection to the test server.
-            b"-oControlMaster=no",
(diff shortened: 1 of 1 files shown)


itamarst:
My Ubuntu 14.10 machine no longer appears to have the slow SSH issue present in older versions, so not sure what to say.

itamarst:
I am unable to review this in any particularly useful way.

exarkun:
Okay.  I'm going to say this is sufficiently self-contained (and therefore ... I hope ... simple) yet important and time-sensitive (we can't do a release until this is fixed) that the testing I've done so far is sufficient (checking that the different SSH options causes `flocker-deploy` to not fail with the "Bad configuration option: IgnoreUnknown" SSH error).  If it somehow ends up introducing a performance regression on Ubuntu (which I don't **think** it does but I also can't reliably reproduce the environment that necessitated this fix) then we'll deal with that when it shows up (hopefully with a better understanding of and more documentation explaining the cause).  Meanwhile, at least it will be possible to use `flocker-deploy` on OS X again.

PR #1064, opened 2015-01-19
[FLOC-1166] New documentation theme

msh100:
https://clusterhq.atlassian.net/browse/FLOC-1166

PR #1072, opened 2015-01-22
[FLOC-1232] [WIP] Sphinx extension for bootstrap components.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1232


flocker/docs/bootstrap/_simple.py
```diff
@@ -0,0 +1,116 @@
...
+        <div class="container"><div class="row"><div class="col-md-12">
+        """),
+        post=dedent("""\
+        </div></div></div>
+        """),
+        match_titles=True,
+    ))
+
+
+mobile_label, MobileLabelDirective, mobile_label_setup = (
```
tomprince: This probably wants to change to use an argument, rather than content, but this is good enough for a first pass.

sample-quickstart.rst
```diff
@@ -0,0 +1,217 @@
...
+
+         Install the flocker-cli client on your Linux machine:
+
+         .. code-block:: console
+
+            $ virtualenv flocker-tutorial && \
+              flocker-tutorial/bin/pip install --upgrade pip && \
+              flocker-tutorial/bin/pip install --quiet flocker-cli
+
+      .. container:: arrow-down center-block invisible
```
tomprince: `container` needs content, so this should probably be replaced by some sort of dedicated directive.

tomprince:
The generated page doesn't look at all right, but that is because the styles and stuff aren't there. Hopefully this branch demonstrates how to have different layout for different pages. Although the layout should be based on #1064 rather than what is here.

sample-quickstart.rst
dwgebler: Yes, another problem with `container` is that it will output HTML with a `container` class on the `div`, e.g.

```
<div class="arrow-down arrow-offset center-block container"></div>
```

`container` is a reserved class name in the Bootstrap framework so this will cause display problems.

dwgebler:
Apologies @tomprince there was a mistake in the RST to HTML conversion guide I sent you for the new directives - `intro-text` was given as wrapping the content in `<header><h1>` tags, whereas it should be:

```
intro-text::
content
```

maps to 

```
    <div class="jumbotron jumbo-flocker">
      <div class="container">
    	<div class="row">
		<div class="col-md-9">
	       	<p>content</p>
		</div>
    	</div>
      </div>
    </div>

```

It is in fact the top

```
=====================
Getting Started with Flocker
=====================
```

that should map to

```
<header>
        <h1 class="text-center">Getting started with Flocker</h1>    
 </header>
```

if we can't make a regular RST header enclosed by two lines of `=` come out with that, I guess we need a custom header directive to.

Another considerable problem is that Sphinx insists on creating permalink anchors next to any heading of any level. I guess worst-case-scenario I could style these to not be rendered on the page at all, but it'd also be easier to not have them in the HTML for this layout, if that's possible.

Tables in the RST need to come out with `table` and `table-striped` classes on the HTML, is there anything built in to docutils to handle that, or would we need another custom `table` directive?
I know you can add `:class:` attributes to any standard directives, but tables are rendered just from the text formatting without any preceding directive so perhaps there's not a built-in mechanism to add classes to those?

The directive extensions for tabs are slightly off in the HTML they produce, currently producing two divs with `tab-content` classes. What we want is one `tab-content` classed element containing two `tab-pane` elements:

```
<div id="install-cli" class="tab-content">
                <div class="tab-pane active" id="cli-osx">
                    <p>Install the flocker-cli client on your Mac (requires homebrew):</p>
                    <samp>
$ brew update &amp;&amp; \
    brew tap clusterhq/flocker &amp;&amp; \
    brew install flocker-0.3.2
                    </samp>
                </div>

                <div class="tab-pane" id="cli-linux">
                    <p>Install the flocker-cli client on your Linux machine:</p>
                    <samp>
$ virtualenv flocker-tutorial &amp;&amp; \
    flocker-tutorial/bin/pip install --upgrade pip &amp;&amp; \
    flocker-tutorial/bin/pip install --quiet flocker-cli
                    </samp>
                </div>
</div>
```

Tricky thing is that the first occurrence of a `tab-pane` should also have an `active` class, but if we can't address that at the Sphinx level, I should be able to do some JavaScript magic there to correct it at the time of page load.

I'll continue to test and review this, but there's some initial issues that can be addressed or thought about. Thanks.

tomprince:
> Another considerable problem is that Sphinx insists on creating permalink anchors next to any heading of any level. I guess worst-case-scenario I could style these to not be rendered on the page at all, but it'd also be easier to not have them in the HTML for this layout, if that's possible.

I think they should be left. With the default sphinx theme, they only show up when you hover over the title, and are in any case useful.

tomprince:
I beleive I've addressed the issue with tabs.

Regarding the tables, alternating rows already have 'row-odd' and 'row-even' classes. Is that sufficient to get the appropriate styling?

The other changes are mechanical changes that should be straightforward to implement.

dwgebler:
>I beleive I've addressed the issue with tabs.

Thanks Tom
 
> Regarding the tables, alternating rows already have 'row-odd' and 'row-even' classes. Is that sufficient to get the appropriate styling?

I can work with that, yeah, it'll take a little adjustment but probably a lot easier and quicker than the work you'd have to put in to generate the tables differently.

dwgebler:
@tomprince I've added a branch off this branch at https://github.com/ClusterHQ/flocker/compare/docs-getting-started-diverged-FLOC-1232?expand=1

In that, I've added the required static assets for the getting started page, filled in the base "home page" layout with the required header/nav/footers that surround the generated doc content, added a "hero-header" directive and altered the "intro-text" directive. This has got us a bit closer to what we want the final Sphinx generated files to come out like. We can merge that branch in to this one if you like, I just didn't want to commit straight to your WIP pull request.

dwgebler:
@tomprince if you want to be freed up for other stuff now, I'm pretty confident I can take this over now and continue working on my branch of this branch. I'll keep committing to that, and then perhaps if you want to get involved in reviewing the final state of things on Monday afternoon (so your Monday morning I guess), I think I'll have a reasonable working solution in place by then.

tomprince:
> "hero-header" directive

Is there any chance this could be done just with CSS? It is already the only `h1` in the document.

dwgebler:
> Is there any chance this could be done just with CSS? It is already the only h1 in the document.

Unfortunately the RST file needs a top h1 to have a Sphinx-recognised title such that it can be included in a table of contents.

```
============================
Getting started with Flocker
============================

.. header-hero::

   Getting started with Flocker
```

I wanted to be able to do it with pure CSS but unfortunately was problematic while preserving correct Sphinx behaviour.

tomprince:
> Unfortunately the RST file needs a top h1 to have a Sphinx-recognised title such that it can be included in a table of contents.

My point is, that you are already styling the sphinx-provided h1 (to hide it). Couldn't you just apply the style you need to that h1, instead of having a duplicate one created by a directive?

dwgebler:
> Couldn't you just apply the style you need to that h1, instead of having a duplicate one created by a directive?

Oh right, unfortunately not; the `header-hero` is to allow the `h1` to be enclosed in a `header` tag.

tomprince:
Is that strictly necessary?

dwgebler:
> Is that strictly necessary?

I suppose it's to some extent relative how "strictly necessary" anything is in software technique, the best I can promise is that time permitting, I will endeavour to streamline the RST and remove any custom directives we can get away with before this whole "new docs, new website, new tutorial" project has to be completed and delivered.

tomprince:
> Is that strictly necessary?

I think I manged to remove the need for it, just by applying the styles that were applied to `header` and `h1` to the sphinx-generated `h1 instead.

tomprince:
Closing in favour of #1083.

PR #1073, opened 2015-01-22
[FLOC-1181] Unattached manifestations

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1181


flocker/node/_deploy.py
```diff
@@ -196,90 +196,89 @@ def run(self, deployer):
...
+            size=VolumeSize(maximum_size=self.dataset.maximum_size)
         )
         return deployer.volume_service.set_maximum_size(volume)
 
 
 @implementer(IStateChange)
-@attributes(["volume"])
-class WaitForVolume(object):
+@attributes(["dataset"])
+class WaitForDataset(object):
```
exarkun: This seems like the right thing for now.  I wonder if we'll want to split this into `WaitForPrimaryManifestation` and `WaitForReplica` later, though.

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        """
+        ``Deployer.calculate_necessary_state_changes`` specifies that a new
+        dataset must be created if the desired configuration specifies
+        that a dataset that previously existed nowhere is going to be
+        on this node.
+        """
+        hostname = u"node1.example.com"
+
+        current = Deployment(nodes=frozenset({
+            Node(hostname=hostname, applications=frozenset()),
```
exarkun: empty frozenset is the default for `applications` fwiw

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+                dataset=MANIFESTATION.dataset)])])
+        self.assertEqual(expected, changes)
+
+    def test_volume_wait(self):
+        """
+        ``Deployer.calculate_necessary_state_changes`` specifies that the
+        dataset previously stored on another node must be waited for, in
+        anticipation of that node handing it off to us.
+        """
+        # The application is not running here - therefore there is no container
```
exarkun: If it's in `other_manifestations` on `node2` then there's no application involved at all, right?  Otherwise it would be attached to the application instead.

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+            hostname=u"node1.example.com",
+            applications=frozenset(),
+        )
+        another_node = Node(
+            hostname=u"node2.example.com",
+            other_manifestations=frozenset({MANIFESTATION}),
+        )
+
+        # The discovered current configuration of the cluster reveals the
+        # application is running somewhere else.
```
exarkun: same comment as above

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        api = Deployer(
+            create_volume_service(self), docker_client=docker,
+            network=make_memory_network()
+        )
+
+        desired = Deployment(nodes=frozenset({
+            Node(hostname=node.hostname,
+                 other_manifestations=frozenset({MANIFESTATION})),
+            Node(hostname=another_node.hostname,
+                 applications=frozenset()),
```
exarkun: the default value again

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+            hostname=u"node1.example.com",
+            other_manifestations=frozenset({MANIFESTATION}),
+        )
+        another_node = Node(
+            hostname=u"node2.example.com",
+            applications=frozenset(),
+        )
+
+        # The discovered current configuration of the cluster reveals the
+        # application is running here.
```
exarkun: same comment as above

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        """
+        ``Deployer.calculate_necessary_state_changes`` specifies no work for
+        the dataset if it was and continues to be on the node.
+        """
+        volume_service = create_volume_service(self)
+        self.successResultOf(volume_service.create(
+            volume_service.get(
+                _to_volume_name(DATASET.dataset_id))
+            )
+        )
```
exarkun: Why does the volume need to be created?

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        will be resized if a dataset which was previously hosted on this
+        node continues to be on this node but specifies a dataset maximum_size
+        that differs to the existing dataset size.
+        """
+        volume_service = create_volume_service(self)
+        self.successResultOf(volume_service.create(
+            volume_service.get(
+                _to_volume_name(DATASET.dataset_id))
+            )
+        )
```
exarkun: Same question

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        dataset will be resized if it is to be relocated to a different
+        node but specifies a maximum_size that differs to the existing
+        size. The dataset will be resized before moving.
+        """
+        volume_service = create_volume_service(self)
+        self.successResultOf(volume_service.create(
+            volume_service.get(
+                _to_volume_name(DATASET.dataset_id))
+            )
+        )
```
exarkun: Same question

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        node but specifies a maximum_size that differs to the existing
+        size. The dataset on the new target node will be resized after it
+        has been received.
+        """
+        volume_service = create_volume_service(self)
+        self.successResultOf(volume_service.create(
+            volume_service.get(
+                _to_volume_name(DATASET.dataset_id))
+            )
+        )
```
exarkun: Same question

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+
+        api = Deployer(
+            volume_service, docker_client=docker,
+            network=make_memory_network()
+        )
+
+        calculating = api.calculate_necessary_state_changes(
+            desired_state=desired,
+            current_cluster_state=current,
+            hostname="node2.example.com",
```
exarkun: unicode

flocker/node/test/test_deploy.py
```diff
@@ -2291,13 +2336,371 @@ def test_dataset_id_generated(self):
...
+        changes = self.successResultOf(calculating)
+        dataset = MANIFESTATION_WITH_SIZE.dataset
+
+        expected = Sequentially(changes=[
+            InParallel(changes=[WaitForDataset(dataset=dataset)]),
+            InParallel(changes=[ResizeDataset(dataset=dataset)]),
+        ])
+        self.assertEqual(expected, changes)
+
+
```
exarkun: Not exercised by this `TestCase` is the case where `primary` changes value.  Though `primary` can't ever be anything but `True` now so that's probably okay.  And we don't have an `IStateChange` to handle that either.  And maybe the ZFS backend doesn't do anything for that change anyway (or maybe it switches between being mounted read-write and read-only?).

So maybe there's nothing to talk about here.  I guess I'm still not entirely clear how `Manifestation.primary` is going to interact with the rest of the system.  I guess it's derived from some configuration we store somewhere... but we don't have that yet, nor specs for any APIs that might manipulate it.  It's probably related to two-phase push, though, somehow - which we do have ...  Hm.

Can state discovery produce `primary=False` `Manifestations`?  (It can't but maybe it should be able to?  The ZFS dataset left after a volume "moves" to another node is probably not a `primary=True` `Manifestation` ...)

flocker/node/test/test_deploy.py
```diff
@@ -2626,15 +3029,7 @@ def created(ignored):
...
-            change = ResizeVolume(
-                volume=AttachedVolume(
-                    manifestation=Manifestation(
-                        dataset=dataset,
-                        primary=True,
-                    ),
-                    mountpoint=FilePath(b"/opt"),
-                )
-            )
+            change = ResizeDataset(dataset=dataset)
```
exarkun: yay

exarkun:
Thanks.  Some minor points noted inline, mostly about stuff that can be deleted I think.  One kind of big question but it doesn't seem like a blocker for this branch - maybe something to turn into an issue or a comment on some existing issue?

Address those to  your satisfaction and merge when buildbot is green.

flocker/node/test/test_deploy.py
itamarst: Replicas (non-primaries) were descoped and moved to https://clusterhq.atlassian.net/browse/FLOC-1240 - there's no need to handle them at this point (and it's also not clear *how* to handle them).

This is the way things were before, and the way we dealt with it then (and now) was making push automatically incremental if there was an existing replica. So existence of replicas is just optimization at the moment.

flocker/node/test/test_deploy.py
itamarst: These tests were semi-copy/pasted, so there's occasional junk left over.

flocker/node/test/test_deploy.py
itamarst: To be more explicit: they are not discoverable at the moment.

PR #1074, opened 2015-01-22
A design sketch for the dataset creation endpoint

exarkun:
Provides design for FLOC-1152.


flocker/control/httpapi.py
```diff
@@ -54,6 +54,44 @@ def version(self):
...
+    )
+    def create_dataset(self):
+        """
+        Create a new dataset on the cluster.
+        """
+        # Use persistence_service to get a Deployment for the cluster
+        # configuration.
+        #
+        # Create a new Dataset with given parameters (generate dataset_id if
+        # necessary).
```
itamarst: We need to error out on duplicate dataset_ids, too. This could be in separate issue if it makes this one too big.

flocker/control/httpapi.py
```diff
@@ -54,6 +54,44 @@ def version(self):
...
+        #
+        # Add the new Manifestation to a Node in the Deployment
+        # (other_manifestations) (need to merge FLOC-1214 for this).
+        #
+        # Persist the new Deployment using persistence_service.
+        #
+        # Return information about the new Manifestation???  But this is
+        # create_dataset.  I guess we'll squish information from the
+        # manifestation (eg the address of the primary) into the representation
+        # of the dataset.
```
itamarst: I wonder if having different API and internal representation is going to be a problem. I guess we'll see.

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,46 @@ definitions:
...
+
+      dataset_id:
+        title: "Unique identifier"
+        description: |
+          An opaque identifier, unique across the cluster, identifying a
+          particular dataset.  If not given, a new identifier will be generated
+          and returned.
+        type: string
+        # The length of a stringified uuid
+        maxLength: 36
```
itamarst: minLength too?

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,46 @@ definitions:
...
+
+      metadata:
+        title: "Data about a dataset"
+        description: |
+          Additional key/value data describing the dataset.  These items are
+          not interpreted by Flocker.  If not given, no metadata will be
+          associated with the new dataset.
+        type: object
+        additionalProperties:
+          type: string
```
itamarst: `object`, not `string`.

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,46 @@ definitions:
...
+        title: "Primary manifestation"
+        description: |
+          The address of the node which will be given the primary manifestation
+          of the newly created dataset.  This must be the address of a node
+          that has introduced itself to the cluster.
+        type: string
+        oneOf:
+          - format: ipv4
+
+      required:
```
itamarst: My guess is that the default schema will end up being pretty lenient, and specific storage backends will add additional constraints somehow.

flocker/control/test/test_httpapi.py
```diff
@@ -62,10 +64,40 @@ def test_version(self):
...
+    def test_wrong_schema(self):
+        # If the input data doesn't match the schema for a dataset, fail with
+        # an error.
+        pass
+
+    def test_dataset_id_collision(self):
+        # Given dataset_id is already in use, fail with an error.
+        pass
+
+    def test_unknown_primary_node(self):
```
itamarst: We don't have a way to validate this currently...

flocker/control/httpapi.py
```diff
@@ -54,6 +54,44 @@ def version(self):
         return {u"flocker":  __version__}
 
 
+    @app.route("/datasets", methods=['POST'])
+    @user_documentation("""
+        Create a new dataset.
```
itamarst: Going to need a lot of expansion.

itamarst:
Looks fine, thanks.

flocker/control/httpapi.py
```diff
@@ -54,6 +69,106 @@ def version(self):
...
+    )
+    def create_dataset(self, primary, dataset_id=None, maximum_size=None, metadata=None):
+        """
+        Create a new dataset in the cluster configuration.
+
+        :param unicode primary: The address of the node on which the primary
+            manifestation of the dataset will be created.
+
+        :param unicode dataset_id: A unique identifier to assign to the
+            dataset.  This is typically a UUID.  If no value is given, one will
```
tomprince: "typically"? When might it not be, and why? Does anything care?

flocker/control/httpapi.py
```diff
@@ -54,6 +69,106 @@ def version(self):
...
+
+        # Use persistence_service to get a Deployment for the cluster
+        # configuration.
+        deployment = self.persistence_service.get()
+        for node in deployment.nodes:
+            for manifestation in node.manifestations():
+                if manifestation.dataset.dataset_id == dataset_id:
+                    raise DATASET_ID_COLLISION
+
+        # XXX Check cluster state to determine if the given primary node
```
tomprince: :exclamation:  X marks the spot

flocker/control/httpapi.py
```diff
@@ -54,6 +69,106 @@ def version(self):
...
+            maximum_size=maximum_size,
+            metadata=pmap(metadata)
+        )
+        manifestation = Manifestation(dataset=dataset, primary=True)
+
+        primary_nodes = list(
+            node for node in deployment.nodes if primary == node.hostname
+        )
+        if len(primary_nodes) == 0:
+            primary_node = Node(hostname=primary)
```
tomprince: :exclamation: It seem like the comment above suggests this should raise `PRIMARY_NODE_NOT_FOUND`.

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,61 @@ definitions:
...
+
+      metadata:
+        title: "Data about a dataset"
+        description: |
+          Additional key/value data describing the dataset.  These items are
+          not interpreted by Flocker.  If not given, no metadata will be
+          associated with the new dataset.
+        type: object
+        patternProperties:
+          ".{1,256}":
```
tomprince: The regex isn't bounded, so if you are trying to limit the length of the key, it should be wrapped in `^$`, I think. See
https://spacetelescope.github.io/understanding-json-schema/reference/object.html#pattern-properties

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,61 @@ definitions:
...
+            type: string
+            # Arbitrary
```
tomprince: :question: I think this and the following comment should be combined and expanded. Something like "we limit the length/number of properties for this and this reason. The specific limits are arbitrary."

flocker/control/test/test_httpapi.py
```diff
@@ -45,25 +71,263 @@ def assertGoodResult(self, method, path, expected_good_result):
...
+
+RealTestsVersion, MemoryTestsVersion = buildIntegrationTests(
+    VersionTestsMixin, "Version", _build_app)
+
+
+class CreateDatasetTestsMixin(APITestsMixin):
+    """
+    Tests for the dataset creation endpoint at ``/datasets``.
+    """
+    NODE_A = b"10.0.0.1"
```
tomprince: :question: We should use one of the TEST-NET blocks defined in [RFC 5737](https://tools.ietf.org/html/rfc5737). One of
- 192.0.2.0/24 (TEST-NET-1)
- 198.51.100.0/24 (TEST-NET-2),
- 203.0.113.0/24 (TEST-NET-3)

flocker/control/test/test_httpapi.py
```diff
@@ -45,25 +71,263 @@ def assertGoodResult(self, method, path, expected_good_result):
...
+        """
+        return self.assertBadResult(
+            b"POST", b"/datasets", {u"primary": self.NODE_A},
+            BAD_REQUEST, {
+                u"description":
+                    u"The provided primary node is not part of the cluster."
+            }
+        )
+    test_unknown_primary_node.todo = (
+        "XXX File a ticket.  Make this pass by inspecting cluster state "
```
tomprince: :exclamation: X marks the spot

flocker/control/schema/endpoints.yml
tomprince: There does appear to be a test for this, so I'm puzzled why it isn't failing.

Also, it appears that if you want to *disallow* properties that don't match, you also need `additionalProperties: false`. Otherwise, this only says that properties that do match this must match a schema, and says nothing about properties that don't.

flocker/restapi/docs/publicapi.py
```diff
@@ -296,6 +296,7 @@ def _formatRouteBody(data, schema_store):
     """
     baseSubstitutions = {
         u"DOMAIN": u"example.com",
+        u"NODE_0": u"node000.example.com",
```
tomprince: :exclamation: This should be an IP address, otherwise, the examples should fail validation.

flocker/control/test/test_httpapi.py
```diff
@@ -45,25 +71,263 @@ def assertGoodResult(self, method, path, expected_good_result):
...
+        """
+        return self.assertBadResult(
+            b"POST", b"/datasets", {u"primary": self.NODE_A, u"junk": u"garbage"},
+            BAD_REQUEST,
+            {u'description': u"The provided JSON doesn't match the required schema.",
+             u'errors': [
+                 u"Additional properties are not allowed (u'junk' was unexpected)"]}
+        )
+
+    def test_dataset_id_collision(self):
```
tomprince: :exclamation: There should perhaps be a test where the collision occurs with an unrelated node (i.e. not the one that is going to be hosting the new volume.

flocker/control/httpapi.py
```diff
@@ -54,6 +69,106 @@ def version(self):
...
+        else:
+            (primary_node,) = primary_nodes
+
+        new_node_config = Node(
+            hostname=primary_node.hostname,
+            applications=primary_node.applications,
+            other_manifestations=
+                primary_node.other_manifestations | frozenset({manifestation})
+        )
```
tomprince: :exclamation: There aren't any tests that this doesn't disturb other nodes, or other configuration on the node being modified.

flocker/control/schema/endpoints.yml
exarkun: Oops, the test was buggy. :cry:  Since I wrote the schema as part of the design, it didn't get developed TDD.  I'll try to be more careful about that in the future.

docs/advanced/api_examples.yml
```diff
@@ -37,3 +43,103 @@
...
+
+  request: |
+    POST /v1/datasets HTTP/1.1
+
+    {"primary": "%(NODE_0)s"}
+
+  response: |
+    HTTP/1.1 201 Created
+
+    {"dataset_id": "a5f75af7-3fb9-4c1a-81ce-efeeb9f2c788", "primary": "%(NODE_0)s", "metadata": {}}
```
itamarst: As noted for version endpoint example, this isn't actually accurate representation of output until https://clusterhq.atlassian.net/browse/FLOC-1254 is done.

flocker/control/httpapi.py
```diff
@@ -53,6 +68,116 @@ def version(self):
...
+        ]
+    )
+    @structured(
+        inputSchema={'$ref': '/v1/endpoints.json#/definitions/datasets'},
+        outputSchema={'$ref': '/v1/endpoints.json#/definitions/datasets'},
+        schema_store=SCHEMAS
+    )
+    def create_dataset(self, primary, dataset_id=None, maximum_size=None,
+                       metadata=None):
+        """
```
itamarst: I would suggest that the bulk of the documentation be not here, but rather in the `user_documentation` decorator where it's more useful.

flocker/control/httpapi.py
```diff
@@ -53,6 +68,116 @@ def version(self):
...
+        Create a new dataset in the cluster configuration.
+
+        :param unicode primary: The address of the node on which the primary
+            manifestation of the dataset will be created.
+
+        :param unicode dataset_id: A unique identifier to assign to the
+            dataset.  This is expected to be a UUID.  If no value is given, one
```
itamarst: "expected to be a UUID" is a little vague, there's a few alternate formats I believe.

flocker/control/schema/endpoints.yml
```diff
@@ -12,3 +12,65 @@ definitions:
...
+        title: "Maximum size"
+        description: |
+          The upper limit on how much data the dataset will be allowed to
+          store, as an integer number of bytes.
+        type: number
+        # 64 MiB.  Sort of an arbitrary limit but inspired by the lower bound
+        # on what this can be set to for the ZFS backend.  Could probably be
+        # considered more carefully.
+        minimum: 67108864
+        # This is how you require integers, of course.
```
itamarst: Of course.

flocker/control/httpapi.py
```diff
@@ -53,6 +68,116 @@ def version(self):
...
+        def saved(ignored):
+            result = {
+                u"dataset_id": dataset_id,
+                u"primary": primary,
+                u"metadata": metadata,
+            }
+            if maximum_size is not None:
+                result[u"maximum_size"] = maximum_size
+            return result
+        saving.addCallback(saved)
```
itamarst: Examples suggest response code is 201, actual code (and tests) verify 200.

flocker/control/test/test_httpapi.py
```diff
@@ -45,25 +72,344 @@ def assertGoodResult(self, method, path, expected_good_result):
...
+            return self.assertBadResult(
+                b"POST", b"/datasets",
+                {u"primary": primary, u"dataset_id": dataset_id},
+                CONFLICT,
+                {u"description": u"The provided dataset_id is already in use."}
+            )
+        posting = saving.addCallback(saved)
+
+        def failed(reason):
+            deployment = self.persistence_service.get()
```
itamarst: Couldn't you just stash the original `Deployment` you create above, and then `self.assertEqual(current_deployment, original_deployment)`?

flocker/control/test/test_httpapi.py
```diff
@@ -35,7 +44,25 @@ def initialize(self):
         self.persistence_service.startService()
         self.addCleanup(self.persistence_service.stopService)
 
-    def assertGoodResult(self, method, path, expected_good_result):
+    def assertResponseCode(self, method, path, request_body, expected_code):
```
itamarst: You rely on the result a lot, and it's supposed to be something specific.

1. Rename method to reflect fact that it's doing an action?
2. Docstring.

flocker/control/test/test_httpapi.py
```diff
@@ -45,25 +72,344 @@ def assertGoodResult(self, method, path, expected_good_result):
...
+
+        def saved(ignored):
+            return self.assertResponseCode(
+                b"POST", b"/datasets", {u"primary": self.NODE_B}, OK
+            )
+        saving.addCallback(saved)
+
+        def created(ignored):
+            deployment = self.persistence_service.get()
+            (node_a,) = (
```
itamarst: Perhaps `Deployment` should grow a function to look up a node by hostname?

flocker/restapi/docs/publicapi.py
```diff
@@ -296,6 +296,7 @@ def _formatRouteBody(data, schema_store):
     """
     baseSubstitutions = {
         u"DOMAIN": u"example.com",
+        u"NODE_0": u"192.0.2.1",
```
itamarst: Examples comment claimed NODE_1, NODE_2 etc. also exist

flocker/control/httpapi.py
itamarst: On second thought, after looking at API docs maybe this is OK - we extract property types and descriptions from the JSON Schema so it's probably OK as is.

itamarst:
Please address above as necessary, and then merge (assuming Buildbot is happy, anyway).

flocker/control/test/test_httpapi.py
exarkun: I wrote it this way because I was afraid of one or the other object being mutated by another part of the code.  This doesn't happen now, as far as I know, but who knows what the future might bring.  When we get to a place where `Deployment` is wholly immutable I think I'd be happy to see the test written the way you suggest.

flocker/control/test/test_httpapi.py
exarkun: Or we should fix https://clusterhq.atlassian.net/browse/FLOC-266 somehow and start using it

PR #1075, opened 2015-01-22
[FLOC-1249] Cluster state service - design sketch

itamarst:


flocker/control/service.py
```diff
@@ -0,0 +1,7 @@
+"""
+Main entry point for control service.
+"""
+
+# - Move service creation code out of script.py and into here.
+# - Add in cluster state service.
+# - Maybe make services look-up-able by class? I.e. use class as service name.
```
exarkun: Maybe fqpn of the class - just to keep the service name a string?

exarkun: Or ... maybe not.  See above.

flocker/control/httpapi.py
```diff
@@ -31,6 +31,9 @@ class DatasetAPIUserV1(object):
     """
     app = Klein()
 
+    # Take the main service instead of a persistence_service, loopkup
+    #individual services by service class, e.g.
+    # self.persistence_service = main_service.getServiceByName(ConfigurationPersistenceService)
```
exarkun: `IServiceCollection` is maybe a bad interface to require of an argument unless you're writing some service-management code, I think.  It's kind of like accepting a single dict as an argument and reaching inside it to get the arguments you're actually interested in... It works but sometimes it means you've gone wrong somewhere.

How about just taking the `ClusterStateService` as another argument?  If we end up with more than a few of these then maybe we have another problem and we should consider doing something different - but I don't see the problem with an `__init__` that takes two arguments for two different objects it needs to operate on.

exarkun:
Thanks.  Looks good.  Inline comments for your consideration.  Please proceed.

flocker/control/httpapi.py
itamarst: OK, I can do that.

PR #1082, opened 2015-01-23
[FLOC-1255] Convergence loop design sketch

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1255

As a first pass, just explain the state machine. I think this covers the necessary edge cases; if it's hard to understand I can turn it into more concrete sketched code... and regardless I may end up with second design review pass, we'll see.

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -5,13 +5,9 @@ The convergence agent runs a loop that attempts to converge the local
 state with the desired configuration as transmitted by the control
 service. This involves two state machines: ClusterStatus and AgentOperation.
 The ClusterStatus state machine receives inputs from the connection to the
-control service, and sends outputs to the AgentOperation state machine.
+control service, and sends inputs to the AgentOperation state machine.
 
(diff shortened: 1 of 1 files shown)


flocker/node/_loop.py
```diff
@@ -0,0 +1,60 @@
...
+If connection failed try to connect again.
+
+IGNORANT:
+
+The status of the cluster is unknown.
+If desired configuration and cluster state are received send a GO input
+symbol to the AgentOperation state machine. switch to KNOWN.
+If disconnected then switch to DISCONNECTED.
+
+KNOWLEDGEABLE:
```
exarkun: KNOWN / KNOWLEDGEABLE are the same I guess

exarkun:
What functionality will be provided when this is complete?

itamarst:
Given an AMP protocl implementation, a `flocker-p2p-agent` script that runs as a convergence agent, I guess.

exarkun:
That can actually converge a node using our current convergence loop implementation?  Okay cool.

PR #1083, opened 2015-01-24
[FLOC-1260] Update getting started guide.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1260

This builds on #1080.


dwgebler:
In your last commit, you changed the getting started RST to use `empty-div` but accidentally left the directive registered in the extension as `general-division`

Anyway, I'll review this whole branch properly on Monday morning UK time.

exarkun:
Consider the errors reported by an html5 validator (eg <https://html5.validator.nu/?doc=http%3A%2F%2Fbuild.clusterhq.com%2Fresults%2Frelease-maintenance%2Fflocker-0.3.2%2Fdocs-getting-started-FLOC-1260%2Fflocker-3792%2Fdocs%2Fgettingstarted%2Fquickstart.html&showimagereport=yes&showsource=yes>) and fix those that seems important (eg "Duplicate ID `tabs`" seems like more than bad style).

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -0,0 +1,197 @@
...
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <title>{{ title|striptags|e }}{{ titlesuffix }}</title>
+
+    <link href="{{ pathto('_static/css/bootstrap.min.css', 1) }}" rel="stylesheet">
+    <link href="{{ pathto('_static/css/' + style, 1) }}" rel="stylesheet">
+    <link href="{{ pathto('_static/css/pygments.css', 1) }}" rel="stylesheet">
+    
+    <script type="text/javascript" src="//fast.fonts.net/jsapi/b20e6cd5-c1a4-4878-9d69-da3806cb86b8.js"></script>
```
exarkun: Scheme-independent URIs seem like a good thing to me.  Some URIs lower down on this page hard-code `https` - perhaps they should be adjusted to be schema-independent as well?

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -0,0 +1,197 @@
...
+                        {%- if next %}
+                            <a href="{{ next.link|e }}" title="{{ next.title|striptags|e }}" class="button rel pull-right">Next</a>
+                        {%- endif %}
+                    </div>
+                </div>
+                
+                <!-- Disqus -->
+                {% if theme_disqus_shortname %}
+                    <div id="disqus_thread"></div>
+                    <noscript>Please enable JavaScript to view the <a href="http://disqus.com/?ref_noscript">comments powered by Disqus.</a></noscript>
```
exarkun: There's an awful lot of JavaScript on this page.  Why single out disqus for this support and ignore the rest?  Is the page usable at all without JavaScript?  For example, you can't switch between the platform tabs without JavaScript (clicking silently does nothing).

docs/_themes/clusterhq/layout_docs.html
exarkun: The goal should be to avoid mixed-content warnings from browsers when the page is loaded over HTTPS.

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -0,0 +1,197 @@
...
+    </div>
+    
+    <!-- Footer area -->
+    <div class="footer">
+        <div class="container">
+            <div class="col-xs-3">
+                <img src="{{ pathto('_static/images/small-logo@2x.png', 1) }}" width="47" height="52">
+            </div>
+            <div class="col-xs-9 text-right">
+                &copy; Copyright 2014, ClusterHQ. Created using <a href="http://sphinx.pocoo.org/">Sphinx</a> {{ sphinx_version }}.
```
exarkun: 2015 fwiw

docs/_themes/clusterhq/layout_homepage.html
```diff
@@ -0,0 +1,169 @@
...
+        {%- block htmltitle %}
+        <title>{{ title|striptags|e }}{{ titlesuffix }}</title>
+        {%- endblock %}
+        <meta name="viewport" content="width=device-width, initial-scale=1">
+
+        <link rel="stylesheet" href="{{ pathto('_static/css/bootstrap.min.css', 1) }}">
+        <link type="text/css" rel="stylesheet" href="http://fast.fonts.net/cssapi/b20e6cd5-c1a4-4878-9d69-da3806cb86b8.css"/>
+        <link rel="stylesheet" href="{{ pathto('_static/css/homepage.css', 1) }}">
+
+        <script src="{{ pathto('_static/js/vendor/modernizr-2.6.2-respond-1.1.0.min.js"', 1) }}></script>
```
exarkun: Quotes on this line are screwed up...

exarkun: Skipping review of the rest of this file, too hard with broken syntax coloring.

docs/_themes/clusterhq/static/css/bootstrap-theme.css
```diff
@@ -0,0 +1,470 @@
+/*!
```
exarkun: Not reviewing any of the imported bootstrap.

docs/_themes/clusterhq/layout_docs.html
dwgebler: Hmm, scheme-relative URLs are useful when you don't know what protocol someone will be accessing a page with, but the best technique is simply if an asset is available over HTTPS, you include it via HTTPS. Mixed content warnings should occur only when the browser requested a URL via HTTPS and the document served in the response subsequently requests assets over unencrypted HTTP. You shouldn't get a browser warning when accessing the page over HTTP but internally that document is including some HTTPS assets.

So the short version is: it's preferable to hard-code all the assets in the document as HTTPS than to leave them scheme-relative.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,221 @@
...
+         Linux
+         ^^^^^
+
+         Install the flocker-cli client on your Linux machine:
+
+         .. code-block:: console
+
+            $ virtualenv flocker-tutorial && \
+              flocker-tutorial/bin/pip install --upgrade pip && \
+              flocker-tutorial/bin/pip install --quiet flocker-cli
```
exarkun: This step fails because there is no `flocker-cli` package on PyPI (nor have we any plans for one, as far as I know :/).

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,221 @@
...
+
+            $ git clone \
+              https://github.com/clusterhq/vagrant-flocker && \
+              cd vagrant-flocker && \
+              vagrant up
+
+         AWS
+         ^^^
+
+         Please see our separate `AWS install instructions <http://docs.clusterhq.com/en/latest/gettingstarted/installation.html#using-amazon-web-services>`_ to get started.
```
exarkun: Uh.  This ... seems like it sort of defeats the purpose of the complex, specific layout we've gone to very much trouble to create.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,221 @@
...
+         $ git clone https://github.com/clusterhq/flocker-tutorial
+         $ cd flocker-tutorial
+
+      fig.yml
+      -------
+
+      .. code-block:: yaml
+
+         web:
+           image: lmarsden/flask:v0.16
```
exarkun: This should be replaced by a `clusterhq`-managed image.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,221 @@
...
+
+      .. code-block:: yaml
+
+         "version": 1
+         "nodes":
+           "172.16.255.250": ["web", "redis"]
+           "172.16.255.251": []
+
+
+      The fig.yml describes your distributed application. The deployment.yml describes which containers to deploy where.
```
exarkun: Some newlines missing here and below.

flocker/docs/bootstrap/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
exarkun: Some docstrings missing from various classes and functions in modules in this package.

docs/gettingstarted/quickstart.rst
dwgebler: I agree, I was just following Michael's comment https://clusterhq.atlassian.net/browse/FLOC-1167?focusedCommentId=15905&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-15905 in respect of the content

docs/gettingstarted/quickstart.rst
exarkun: Make this `pip install --quiet https://archive.clusterhq.com/downloads/flocker/Flocker-0.3.2-py2-none-any.whl` I guess.  Or a shorter version, if you can manage, so it better fits in the box (perhaps a bit.ly URL?  dunno)

exarkun:
Thanks.  Please address inline comments and re-submit.

docs/gettingstarted/quickstart.rst
tomprince: If we want a shorter URL, it should be somewhere we manage, probably?

docs/_themes/clusterhq/layout_docs.html
dwgebler: The primary docs (those using this layout file, i.e. not the getting started page) are usable with JS disabled, with the exceptions of the search facility and Disqus comments, both of which will display a warning that they require JS enabled.

The getting started page's tabs and "Get involved" signup form will not work without JS. I can include non-JS alternatives for the tabs, but there is no static alternative for the signup form (Intercom, as far as I can tell from their docs, relies on JS exclusively for the integration to work). So the best I can suggest with that is to hide the signup form and its enclosing banner by default and use JS to unhide it immediately when the page loads.

I will do this in a revision commit.

flocker/docs/bootstrap/_logo.py
```diff
@@ -0,0 +1,51 @@
+
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Logo directive
```
tomprince: This directive should probably just use the helper in simple (it can go there). It was originally written before the helper.

flocker/docs/bootstrap/_extension.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    for module in [_simple, _tabs]:
+        module.setup(app)
+
+
+class HTMLWriter(HTMLTranslator):
+    """
+    Overrides part of the default HTMLTranslator to provide specific
+    class names on some generated HTML elements.
```
tomprince: It looks like these methods are copied from sphinx (or maybe docutils?, with some changes. We should give attribution, and include where they came from, and what changes where made.

flocker/docs/bootstrap/_extension.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+
+    def visit_table(self, node):
+        self._table_row_index = 0
+        self.context.append(self.compact_p)
+        self.compact_p = True
+        classes = ' '.join(['docutils', 'table', 'table-striped',
+                            self.settings.table_style]).strip()
```
tomprince: It seems like we could just override `table_style` instead of this function.

Alternatively, defining a node visitor (See `_tabs`, that adds the class to the node; `docutils.writers.html4css1.Writer.starttag` is where this is used).

dwgebler: Probably, I did this as a very quick hack to make tables come out right. The changes required are that tables come out in the HTML with classes of `table table-striped` on the `table` tag and cell content wrapped in `<p>` tags. If you know how to do this better, please feel free to commit an improvement to this branch :)

flocker/docs/bootstrap/_extension.py
dwgebler: I think one method was copied from Sphinx's `HTMLTranslator` and the other one from the docutils base class that Sphinx's class inherits. They were then modified to add a couple of class names to the table tag and to wrap cell content in paragraph tags. As per my other comment below, if you want to change/improve this, please do. Otherwise we can add a comment attributing the original source.

docs/_themes/clusterhq/layout_docs.html
exarkun: Hm.  I'm not sure what the point of loading things over HTTPS when the page itself (or any of the content, really) is loaded over HTTP.  Little or no extra protection is granted but caching is worsened and page load times are slightly slowed.  However, I don't think those things are worth making changes for.  So if this isn't going to give anyone mixed content warnings now, great, scratch this comment.

docs/gettingstarted/quickstart.rst
exarkun: Please file a follow-up issue for fixing this box.

docs/gettingstarted/quickstart.rst
exarkun: Or better yet, as Itamar suggested, this should all be changed back to one of our existing examples.

flocker/docs/bootstrap/_extension.py
exarkun: Please be very careful copying code from other projects.  It **must** always be properly attributed.

dwgebler:
> Or better yet, as Itamar suggested, this should all be changed back to one of our existing examples.

I've addressed all comments except this one; Luke's migration diagram in the getting started guide is modeled on the flask/redis example so I can't change the sample application config used without a new diagram being produced to go with it. I can also understand the rationale of using an example that can be quickly checked in a web browser, so perhaps the better solution is for `lmarsden/flask:v0.16` to be moved in to the clusterhq namespace. Marketing may argue that `dockerfile/redis` also shows that Flocker works with standard Docker images and not just those prepared by ClusterHQ.

docs/_themes/clusterhq/layout_docs.html
tomprince: scheme-relative URLs seem break viewing ``file://`` URLs (which you get when viewing the documentation locally).

dwgebler:
> I've addressed all comments except this one

My understanding is now that we are going with the Flask example for the time being, with a view to changing it later. For now the tutorial vagrant box will pre-include the required Flask and Redis images.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,268 @@
+:layout: homepage
```
tomprince: Any reason this isn't just in `gettingstarted/index.rst`?

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,268 @@
...
+
+         Vagrant
+         ^^^^^^^
+
+         Simulate a Flocker cluster with virtual machines on your laptop (requires Vagrant, VirtualBox):
+
+         .. code-block:: console
+
+            $ git clone \
+              https://github.com/clusterhq/vagrant-flocker && \
```
tomprince: If we are going to have another repo (:sad:) for this, then the release-process instructions need to be adjusted to update the version number in that repository. Having a separate repository for this also means it isn't possible to have multiple versions available.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,268 @@
...
+   ============================
+   
+   .. tutorial-step-condensed::
+
+      You should have flocker-cli installed on your laptop and flocker-node installed on some servers: either VMs on your laptop, or real instances on cloud infrastructure.
+      Now you can try our simple tutorial: a Python web application and a Redis server.
+
+      .. code-block:: console
+
+         $ git clone https://github.com/clusterhq/flocker-quickstart
```
tomprince: As above, having a separate repository for this also means it isn't possible to have multiple versions available. In particular, if we want to change this, we can't update the main branch of the repository until we do another release.

docs/gettingstarted/quickstart.rst
dwgebler: I had Sphinx complaining at one point (as a warning, but still complaining) about something not being included in any toctree. Possibly if docs/index has gettingstarted/index in its toctree, quickstart.rst can be renamed gettingstarted/index.rst and that problem won't happen.

docs/gettingstarted/quickstart.rst
dwgebler: Multiple versions of the Vagrantfile? Yes the release process doc will need updating, but other than that I don't really know how to answer this (or the comment below about the `flocker-quickstart` repo, this is how @lukemarsden wanted it and how he wrote the content for the guide, for the purpose of the instructions being simple and concise copy/paste commands.

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -0,0 +1,197 @@
...
+            <!-- Main contents area -->
+            <div class="col-sm-8 col-sm-offset-1 page white-bg">
+                {%- if pagename != "search" %}
+                <form class="search" action="{{ pathto('search') }}" method="get" _lpchecked="1">
+                    <input type="text" name="q" placeholder="Type here to search docs..." id="search">
+                    <input type="hidden" name="check_keywords" value="yes">
+                    <input type="hidden" name="area" value="default">
+                </form>
+                {%- endif %}
+                <!-- DOCS PAGE DATA HERE -->
```
exarkun: please standardize on a single start/end comment style and make all of these comments follow it

adamtheturtle:
A few issues I noticed (not a proper review):

* Our other headings are title case.
* Hovering over this title shows a pilcrow unlike other titles in our site.
* At a small-medium widths "Try Flocker" in the header loses its border and then just looks like another link in the header - but it is vertically misaligned.
* You can change the width of the window so that only the nose of the bird is shown in the green section.
* The bird should either move/shrink so you can see it all, or go away entirely.
* Syntax highlight the yml files and bash commands
* At some widths the linux install instructions flow out of their container.
* When I make the window width small enough that only one column can fit, there are white breaks in the right hand grey border, and there is no border on the left which is odd.
* Some images are blurry on a retina display.
* "Homebrew" should link to http://brew.sh
* When I make the window width small enough that only one column can fit, there is an orange block saying "Local" at the top of the Flocker Node Installation section which I don't understand, which overlaps a grey downward arrow.

Please fix these or make follow up issues as you see appropriate.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,313 @@
...
+      .. noscript-content::
+
+         Ubuntu / Debian
+         ^^^^^^^^^^^^^^^
+
+         Install the dependencies:
+         
+         .. code-block:: console
+         
+            $ sudo apt-get install gcc python2.7 python-virtualenv python2.7-dev
```
exarkun: Problems I encountered:

  * my user is not a sudoer so this step failed (fixed by getting root some other way)
  * my package list is out of date so apt-get couldn't find the packages (fixed by running `apt-get update`)
  * `apt-get install` prompted me to continue (fixed by running `apt-get install -y ...`)

I think the first is "too bad" there's nothing we can reasonably do about that.
I think the second two points can reasonably be addressed by changing the command we suggest.

Also I think the intent of these panels is to provide a single command the user can copy/paste to get Flocker so please combine the fixed version of this command with the rest of the install steps with `&&` as is used below.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,313 @@
...
+
+         Next, install the flocker-cli client on your Linux machine:
+
+         .. code-block:: console
+
+            $ virtualenv flocker-tutorial && \
```
exarkun: I copy/pasted this command and the `$` was copied.  I think there's an easy fix for this, something to do with `.. prompt:: bash $`.  If that fixes it and still looks okay, please make that change.  If it's any harder than that, file a follow-up issue instead and leave this as-is for now.

docs/gettingstarted/quickstart.rst
exarkun: Also `git` is another dependency.  When I ran this step, `-bash: git: command not found`.  I'm inclined to believe relevant folks in our audience will know how to deal with this, though.  So I don't think we should bother to do anything about it (though we could add `git` to the list of `apt-get` packages at the top).

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,313 @@
...
+           "172.16.255.251": []
+
+
+      The fig.yml describes your distributed application.
+      The deployment.yml describes which containers to deploy where.
+      If you are using real servers on AWS, you'll need to change the IP addresses in the deployment file.
+
+      .. code-block:: console
+
+         $ flocker-deploy deployment-node1.yml fig.yml
```
exarkun: `flocker-deploy` is not `$PATH` at this point.  This suggests we really actually do just want to activate the virtualenv.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,313 @@
...
+      .. mobile-label::
+
+         Live
+
+      .. noscript-content::
+
+         Vagrant
+         ^^^^^^^
+
+         Simulate a Flocker cluster with virtual machines on your laptop (requires Vagrant, VirtualBox):
```
exarkun: ... and how do I get those :/

docs/gettingstarted/quickstart.rst
exarkun: Also this step proceeded to fail anyway:

```
/home/exarkun/vagrant-flocker/Vagrantfile:6:in `<top (required)>': undefined method `require_version' for Vagrant:Module (NoMethodError)
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:115:in `load'
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:115:in `block in procs_for_source'
        from /usr/lib/ruby/vendor_ruby/vagrant/config.rb:41:in `block in capture_configures'
        from <internal:prelude>:10:in `synchronize'
        from /usr/lib/ruby/vendor_ruby/vagrant/config.rb:36:in `capture_configures'
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:114:in `procs_for_source'
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:51:in `block in set'
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:45:in `each'
        from /usr/lib/ruby/vendor_ruby/vagrant/config/loader.rb:45:in `set'
        from /usr/lib/ruby/vendor_ruby/vagrant/environment.rb:377:in `block in load_config!'
        from /usr/lib/ruby/vendor_ruby/vagrant/environment.rb:392:in `call'
        from /usr/lib/ruby/vendor_ruby/vagrant/environment.rb:392:in `load_config!'
        from /usr/lib/ruby/vendor_ruby/vagrant/environment.rb:327:in `load!'
        from /usr/bin/vagrant:40:in `<main>'
```

exarkun: Since there were no installation instructions for Vagrant and virtualbox, I ran `apt-get install vagrant virtualbox`.  Ubuntu 14.04 packages vagrant 1.0.3 which is far too old to work with this tutorial.

exarkun: So instead of just saying we require vagrant, we should probably tell people to install a particular version.  The `Vagrantfile` in the git repo suggests >= 1.6.2.  I'm pretty sure there are one or two versions newer than that that are also broken - but the newest version is not, afaik.

We should probably refer people to https://www.vagrantup.com/downloads.html which has installers for various platforms.  We'll have to keep on top of new Vagrant releases to make sure they still work (but really we should be doing that anyway so nothing new there).  Please do file an issue for that task, though.

docs/gettingstarted/quickstart.rst
exarkun: I guess we're going to go with this for now.  Please file an issue for integrating the flocker-quickstart work into the release process.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+              flocker-tutorial/bin/pip install --quiet https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker/Flocker-0.3.2-py2-none-any.whl
+
+         Fedora 20
+         ^^^^^^^^^
+
+         Install the flocker-cli client on your Linux machine:
+         
+         .. code-block:: console
+         
+            $ sudo yum install @buildsys-build python python-devel python-virtualenv && \
```
exarkun: Likewise, this prompts for confirmation.  `yum install -y ...` addresses that.

docs/gettingstarted/quickstart.rst
exarkun: Ugh.  Neither virtualbox nor vagrant appears to be in fedora 21's package repository.  So a link to https://www.virtualbox.org/wiki/Linux_Downloads here is probably also a good idea.

exarkun: (which doesn't have packages for Fedora 20 or Fedora 21)

exarkun: installing VirtualBox on Fedora is supposedly like:

```
wget -q https://www.virtualbox.org/download/oracle_vbox.asc
rpm --import oracle_vbox.asc
curl http://download.virtualbox.org/virtualbox/rpm/fedora/virtualbox.repo > /etc/yum.repos.d/virtualbox.repo
yum install VirtualBox-4.3
```

However, while this gave me a working `VBoxManage` command it didn't give me a working virtualbox backend for vagrant...

exarkun: I fixed this by running this blob from our other Fedora install docs:

```
UNAME_R=$(uname -r)
PV=${UNAME_R%.*}
KV=${PV%%-*}
SV=${PV##*-}
ARCH=$(uname -m)
yum install -y https://kojipkgs.fedoraproject.org/packages/kernel/${KV}/${SV}/${ARCH}/kernel-devel-${UNAME_R}.rpm
```

With the right kernel headers package installed the VirtualBox install succeeded for real instead of only pretending to succeed.

I don't know if all of these instructions fit (or if we even want to recommend people install kernel headers from koji).  Maybe instead we should just make AWS the default tab and hope that only people who already like Vagrant and have it working click on the Vagrant tab.

exarkun: Oops!  Forgot there's nothing useful on the AWS.  Hmmm.

dwgebler:
In reference to some earlier comments:

> At a small-medium widths "Try Flocker" in the header loses its border and then just looks like another link in the header - but it is vertically misaligned.

That should be fixed now.

> You can change the width of the window so that only the nose of the bird is shown in the green section. The bird should either move/shrink so you can see it all, or go away entirely.

This is at least partially fixed, might need some minor further adjustment.

> Syntax highlight the yml files and bash commands

That should be fixed now.

> At some widths the linux install instructions flow out of their container.

That should be fixed now.

> When I make the window width small enough that only one column can fit, there are white breaks in the right hand grey border, and there is no border on the left which is odd.

Looking in to this,display seems at least passable on small devices for now so possibly a follow-up issue for fine tuning.

> Some images are blurry on a retina display.

The docs build via Sphinx slightly complicates the issue of automatically adjusting images for retina displays. I'm working on it, but this will probably go in a follow-up issue.

> When I make the window width small enough that only one column can fit, there is an orange block saying "Local" at the top of the Flocker Node Installation section which I don't understand, which overlaps a grey downward arrow.

That's fixed, the label is for mobile displays only, indicating which installation instructions are for the local CLI and which are for the live node tools.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+
+         Vagrant
+         ^^^^^^^
+
+         Simulate a Flocker cluster with virtual machines on your laptop (requires `Vagrant <http://www.vagrantup.com/downloads>`_, `VirtualBox <https://www.virtualbox.org/wiki/Downloads>`_):
+
+         .. code-block:: console
+
+            you@laptop:~$ git clone \
+              https://github.com/clusterhq/vagrant-flocker && \
```
exarkun: Looks like no one ever wrote the README for this repository.  Please file a follow-up issue for this work.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
```
exarkun: Likewise for the README in this repository.

docs/_themes/clusterhq/layout_docs.html
exarkun: Please file a follow-up issue for addressing this.

docs/gettingstarted/quickstart.rst
dwgebler: https://clusterhq.atlassian.net/browse/FLOC-1305

docs/gettingstarted/quickstart.rst
dwgebler: https://clusterhq.atlassian.net/browse/FLOC-1305

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+
+         Ubuntu / Debian
+         ^^^^^^^^^^^^^^^
+
+         Install the flocker-cli client on your Linux machine:
+         
+         .. code-block:: console
+         
+            you@laptop:~$ sudo apt-get update && apt-get install -y gcc python2.7 python-virtualenv python2.7-dev && \         
+              virtualenv flocker-tutorial && \
```
exarkun: Please file a follow-up issue for moving the `activate ...` to immediately after this line and changing the `pip` invocations on the subsequent lines to rely on `${PATH}` instead of specifying a path.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+              flocker-tutorial/bin/pip install --upgrade pip && \
+              flocker-tutorial/bin/pip install --quiet https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker/Flocker-0.3.2-py2-none-any.whl && source flocker-tutorial/bin/activate
```
exarkun: Please file a follow-up issue for providing a shorter URL for this command.

docs/_themes/clusterhq/layout_docs.html
dwgebler: https://clusterhq.atlassian.net/browse/FLOC-1306

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+         ^^^^^^^^^
+
+         Install the flocker-cli client on your Linux machine:
+         
+         .. code-block:: console
+         
+            you@laptop:~$ sudo yum install -y @buildsys-build python python-devel python-virtualenv && \
+              virtualenv flocker-tutorial && \
+              flocker-tutorial/bin/pip install --upgrade pip && \
+              flocker-tutorial/bin/pip install --quiet https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker/Flocker-0.3.2-py2-none-any.whl && source flocker-tutorial/bin/activate
```
exarkun: Above mentioned tickets should cover similar fixes to this section as well.

exarkun: Please file a ticket for adding automated test coverage for these two command blocks.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+         OS X
+         ^^^^
+
+         Install the flocker-cli client on your Mac (requires Homebrew):
+
+         .. code-block:: console
+
+            you@laptop:~$ brew update && \
+              brew tap clusterhq/flocker && \
+              brew install flocker-0.3.2
```
exarkun: Oops.  This is another install command.  Amend my above comment about two command blocks to cover this third block as well.

exarkun: Please file a ticket for automatically substituting the latest released version of flocker into this documentation at build time.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+         ^^^^^^^^^
+
+         Install the flocker-cli client on your Linux machine:
+         
+         .. code-block:: console
+         
+            you@laptop:~$ sudo yum install -y @buildsys-build python python-devel python-virtualenv && \
+              virtualenv flocker-tutorial && \
+              flocker-tutorial/bin/pip install --upgrade pip && \
+              flocker-tutorial/bin/pip install --quiet https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker/Flocker-0.3.2-py2-none-any.whl && source flocker-tutorial/bin/activate
```
exarkun: Please file a ticket for de-duplicating the install content that appears both here and above under the `noscript-content` block.

docs/gettingstarted/quickstart.rst
exarkun: Please file a ticket for integrating `clusterhq/vagrant-flocker`  updates into the release process document.

docs/gettingstarted/quickstart.rst
exarkun: Please file a ticket for de-duplicating the content that appears both here and under the `noscript-content` block.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+   ============================
+   
+   .. tutorial-step-condensed::
+
+      You should have flocker-cli installed on your laptop and flocker-node installed on some servers: either VMs on your laptop, or real instances on cloud infrastructure.
+      Now you can try our simple tutorial: a Python web application and a Redis server.
+
+      .. code-block:: console
+
+         you@laptop:~$ git clone https://github.com/clusterhq/flocker-quickstart
```
exarkun: Please file a ticket for integrating `clusterhq/flocker-quickstart` updates into the release process.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+           image: clusterhq/flask
+           links:
+            - "redis:redis"
+           ports:
+            - "80:80"
+         redis:
+           image: dockerfile/redis
+           ports:
+            - "6379:6379"
+           volumes: ["/data"]
```
exarkun: Please file a ticket for keeping this code block up-to-date with the content in `flocker-quickstart`.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+
+      deployment-node1.yml
+      --------------------
+
+      .. code-block:: yaml
+
+         "version": 1
+         "nodes":
+           "172.16.255.250": ["web", "redis"]
+           "172.16.255.251": []
```
exarkun: That ticket should cover this block as well.

docs/gettingstarted/quickstart.rst
```diff
@@ -0,0 +1,293 @@
...
+
+      deployment-node2.yml
+      --------------------
+
+      .. code-block:: yaml
+
+         "version": 1
+         "nodes":
+           "172.16.255.250": ["web"]
+           "172.16.255.251": ["redis"]
```
exarkun: And this one.

flocker/docs/bootstrap/_extension.py
exarkun: I think you probably need to add docutils and sphinx copyright lines to the top of this file.

exarkun:
Thanks.  Please file the requested tickets.  Mark them as blockers of <https://clusterhq.atlassian.net/browse/FLOC-1208>.  Make the one suggested change (copyright statements) and then go ahead and merge.

PR #1091, opened 2015-01-27
[FLOC-1193] Changed push commands

adamtheturtle:
Fixes FLOC-1193 and changes another occurrence of the same push command which caused the issue.


exarkun:
I failed to reproduce the problem with a brand new repository. :cry: 

I think this change makes sense anyway because it makes the command follow the grammar documented in the `git push` man page - even if the other argument order works.  Maybe there's something else to fix as well - like we need to specify a particular git configuration for our repositories or there's some interaction with a different local workflow that can result in this behavior in certain cases that I don't know about. :/

exarkun:
Please merge this as-is, anyway.  If I learn anything more about the problem I'll raise another issue.

PR #1093, opened 2015-01-27
[FLOC-1255] Convergence loop

itamarst:
Building on top of the AMP protocol, a more detailed sketch of convergence loop.


flocker/node/_loop.py
```diff
@@ -0,0 +1,100 @@
...
+If changes finish switch to STOPPED.
+If GO is received switch to CHANGING.
+"""
+
+
+from ._protocol import AgentClient
+
+
+class AgentLoopService(Service):
+    def __init__(self, host, port):
```
exarkun: endpoints please

flocker/node/_loop.py
```diff
@@ -0,0 +1,100 @@
...
+        self.agent_operation = AgentOperation()
+        self.cluster_status = ClusterStatus(self.agent_operation)
+
+    def startService(self):
+        self.factory = ReconnectingClientFactory()
```
exarkun: sigh.

Should I go implement a reconnecting tool for endpoints?

flocker/node/_loop.py
```diff
@@ -0,0 +1,100 @@
...
+        self.factory.protocol = lambda: AgentClient(self)
+        reactor.connectTCP(self.host, self.port)
+
+    def stopService(self):
+        # stop factory
```
exarkun: And close (abort?) the current connection, I think?  And wait for it to receive disconnect notification.

flocker/node/_loop.py
```diff
@@ -0,0 +1,100 @@
+"""
+Convergence loop for the agent.
```
exarkun: Probably worth saying "dataset" at least once somewhere here since this is only the (zfs?) dataset convergence loop.  Although most of this code looks like it is generic and applicable to any of our convergence agents.  I imagine we'll refactor as soon as we want a second kind of agent.

flocker/node/_loop.py
```diff
@@ -0,0 +1,100 @@
...
+        pass
+
+
+# Rich inputs for CONNECTED (AgetnClient) and UPDATE (two Deployments, desired and actual)
+# Rich input for GO (AgentClient)
+
+
+# AgentOperation as in state machine description above; uses AgentClient to send discovered NodeState
+
+class AgentOperation(object):
```
exarkun: Using machinist?  Does this build the FSM, proxy to its `receive` method, and have the `IOutputExecutor` as a private implementation detail?

If it uses `AgentClient` I guess it should take it as an argument - so it can't be instantiated until a connection is made, and it has to be re-constructed (or at least mutated) whenever a connection is lost/re-established.  This probably merits a bit more fleshing-out.

exarkun:
Thanks.  Please clarify the state machine and its interaction with the client and re-submit.

flocker/node/_loop.py
itamarst: This is why I didn't use endpoints. There's a branch up for review, which has some wierdness (takes a ClientFactory, which is silly) but otherwise probably sufficient, so you could review it. But even with that done it's easier to not use endpoints in this pass.

itamarst: Twisted branch, that is.

flocker/node/_loop.py
itamarst: I guess I should actually relearn how machinist works, eh.

itamarst: The `AgentClient` gets passed in using the `GO` rich input, so you get new one every time reconnection happens. I'll try to sketch this a bit more though.

PR #1096, opened 2015-01-28
[FLOC-1254] Don't wrap REST API results in additional dictionary - design sketch

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1254

This may actually be all the work that is necessary for this issue, but I was fairly superficial doing the work and may've missed something.

### DIFF
--- a/flocker/restapi/docs/test/test_publicapi.py
+++ b/flocker/restapi/docs/test/test_publicapi.py
@@ -123,7 +123,7 @@ class MakeRstTests(SynchronousTestCase):
                 u"response":
                     u"HTTP/1.1 200 OK\n"
                     u"\n"
-                    u'{"error": false, "result": "%(DOMAIN)s"}\n',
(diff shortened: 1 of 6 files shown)


flocker/restapi/docs/test/test_publicapi.py
```diff
@@ -123,7 +123,7 @@ def hasExamples():
                 u"response":
                     u"HTTP/1.1 200 OK\n"
                     u"\n"
-                    u'{"error": false, "result": "%(DOMAIN)s"}\n',
+                    u'"%(DOMAIN)s"\n',
```
exarkun: Maybe stick some other arbitrary structure around this value now.  A bare string isn't legal JSON, I think.

exarkun:
Thanks.  Basically looks good.  Unfortunately master has diverged meanwhile (in a somewhat non-trivial way).  Please resolve that conflict and submit for code review (accepting this as a sufficient design).

PR #1098, opened 2015-01-29
Make more CLI tools available from omnibus package

exarkun:
Fixes FLOC-1291

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,9 @@ setup(
     },
 
     entry_points = {
-        # Command-line programs we want setuptools to install:
+        # These are the command-line programs we want setuptools to install.
+        # Don't forget to modify the omnibus packaging tool
+        # (admin/packaging.py) if you make changes here.
         'console_scripts': [
             'flocker-volume = flocker.volume.script:flocker_volume_main',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -73,7 +73,9 @@
     },
 
     entry_points = {
-        # Command-line programs we want setuptools to install:
+        # These are the command-line programs we want setuptools to install.
+        # Don't forget to modify the omnibus packaging tool
+        # (admin/packaging.py) if you make changes here.
```
wallrj: I guess we could have the list of entrypoints centrally, maybe in `admin/somewhere` which can be imported in `setup.py` and in the `packaging` module and its tests....as a followup issue though.

wallrj:
Looks good to me.
 * All the package builds pass
 * There's a failed httpapi test when run against the fedora-20 omnibus package installation, but it must be unrelated.
 * I test installed the centos7 packages from this branch and found that `flocker-control` was available and runnable.
 * It seems confusing that `flocker-zfs-agent` uses a main function called `flocker_volume_main`, but that's not related to this branch.

Please merge.

PR #1099, opened 2015-01-29
[FLOC-1038] Version in docs

adamtheturtle:
Fixes FLOC-1038


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -32,6 +32,16 @@ Software
...
+- ``envsubst`` available on your PATH.
+
+  .. note:: ``envsubst`` requires ``gettext`` to be installed on OS X.
+
+             Install `Homebrew`_.
+
+             .. code-block:: console
+
+                $ brew install gettext
+                $ brew link --force gettext
```
adamtheturtle: Having to ``--force`` isn't great:

```
Warning: gettext is keg-only and must be linked with --force
Note that doing so can interfere with building software.
```

It is also possible to instead:

```
echo export PATH=\${PATH}:/usr/local/opt/gettext/bin >> ~/.bash_profile
source ~/.bash_profile
```

flocker/docs/version_code_block.py
```diff
@@ -0,0 +1,43 @@
...
+class VersionCodeBlock(CodeBlock):
+    """
+    Similar to CodeBlock but replaces |RELEASE| with the latest release
+    version.
+    """
+    def run(self):
+        parsed_version = parse_version(version)
+        latest = parsed_version.release
+
+        if parsed_version.weekly_release is not None:
```
adamtheturtle: I guess that this logic could be in a property of flocker.docs._version.FlockerVersion but it is kind of weird because it is trying to get the last version which e.g. Hombrew will have a script for. We don't make install scripts for documentation releases.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -93,23 +103,13 @@ Preparing For a Release
...
-   Commit the changes:
+#. Update the version numbers in the install and Vagrant scripts:
 
    .. code-block:: console
 
-      $ git commit -am "Bumped version numbers"
-
-   .. This should be automated. See https://clusterhq.atlassian.net/browse/FLOC-1038
+      envsubst < docs/gettingstarted/linux-install.sh.template > docs/gettingstarted/linux-install.sh
+      envsubst < docs/gettingstarted/tutorial/Vagrantfile.template > docs/gettingstarted/tutorial/Vagrantfile
```
tomprince: Rather than adding another dependency to this process, how about a python solution to this? The script could even automate the commit.

adamtheturtle: Perhaps, although is Python better than a shell script? Note that Linux (or at least Richard's Fedora machine and the Flocker dev environment) have `envsubst` already. We have been discussing recently not supporting OS X as an environment for doing this process and as I'm currently the only one on OS X I'm not sure that the dependency is a problem.

flocker/docs/version_code_block.py
tomprince: It would probably make sense to get the version that sphinx thinks this is, which already has that processing done.

Alternatively, `get_doc_release` will strip the `+docX` from a documentation release. (I guess this may be a slightly confusing name. The idea is get the version to be displayed in the docs).

flocker/docs/version_code_block.py
```diff
@@ -0,0 +1,43 @@
...
+            latest = latest + 'dev' + parsed_version.weekly_release
+        elif parsed_version.pre_release is not None:
+            latest = latest + 'pre' + parsed_version.pre_release
+
+        self.content = [item.replace(u'|RELEASE|', latest) for item in
+                        self.content]
+        block = CodeBlock(self.name, self.arguments, self.options,
+                          self.content, self.lineno, self.content_offset,
+                          self.block_text, self.state, self.state_machine)
+        return block.run()
```
tomprince: If this is subclassing `CodeBlock`, can't you just `return CodeBlock.run(self)` here?

tomprince:
I think maybe we should be using `|version|` instead of `|release|` (and then we could make `|release|` include the `+docX`. (Not that either of those are currently displayed anywhere in the docs).

tomprince:
I bet sphinx could be made to do the substitutions we want for the downloads. But I doubt it is worth the effort here.  (https://github.com/sphinx-doc/sphinx/issues/1698)

docs/gettinginvolved/infrastructure/release-process.rst
tomprince: Python is better that shell, yes. It is non-trivial to get quoting correct in shell, for example.

tomprince:
Please
- [ ] use python rather than envsubst
- [ ] switch to `|version|`
- [ ] simplify the the code in `run`

adamtheturtle:
Would you prefer that the python does the commit, as suggested in your comment?

Is it not odd to have |version| in the docs be different from |version| in a code block?
Version in the docs should have e.g. `-dirty` but these code blocks should not (because e.g. there is no -dirty Homebrew script).

tomprince:
> Would you prefer that the python does the commit, as suggested in your comment?

I think it might as well, we will want it to do it eventually, as we automate more of the process.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -93,23 +103,13 @@ Preparing For a Release
    The release may require certain changes to be back ported from the master branch.
    See :ref:`back-porting-changes`\ .
 
-#. Update the version numbers in:
-
-   .. note:: Skip this step for a documentation release.
```
tomprince: This line shouldn't be lost.

tomprince:
In sphinx convention, `|release|` is more specific that `|version|`. 

I guess it is reasonable to have the docs point at installing that last release. In which case, we should probably substitute something other than `|version|` *or* `|release|`.

adamtheturtle:
How about `|latest-packaged-version|`?

tomprince:
Sure.

wallrj:
> use python rather than envsubst

It seems like ultimately, these OSX installation steps could be implemented as `flocker.provision.tasks` and run automatically. The only thing missing, when I discussed it with @adamtheturtle, was that you can't supply sample output to the tasks. But that should be simple enough to implement.
So I wonder, why go to the trouble of creating a python tool at this stage?

tomprince:
Part of the point of using python rather than `envsubst`, it can be reused as part of a more fully automated release tool. That it happens to remove a need for an extra dependency for doing the release process on OSX is a happy side-effect.

flocker/docs/version_code_block.py
adamtheturtle: > It would probably make sense to get the version that sphinx thinks this is

Is that not what gets rendered when `|version|` is with prose, which is not what is wanted here?

tomprince: When I wrote that comment, I was assuming that it was what was wanted. My point was, don't try and generate the same thing as sphinx does, but defer to what sphinx is configured to use (so that there is only one place that needs to be changed, to change what gets displayed).

In any case, the version should be based on the version that sphinx has. And if there has to be logic to format the version, it should live in `flocker.docs._version`.

docs/gettingstarted/installation.rst
```diff
@@ -100,10 +100,10 @@ You can see the Homebrew recipe in the `homebrew-tap`_ repository.
 
 The ``flocker-deploy`` command line program will now be available:
 
-.. code-block:: console
+.. version-code-block:: console
 
    alice@mercury:~$ flocker-deploy --version
-   0.3.2
+   |latest-packaged-version|
```
adamtheturtle: Perhaps "client-version" or similar which references flocker-cli

tomprince: I'm not sure if it occurs in the PR, but this also refers to the version of the node software (they should always be the same, for the moment).

docs/gettingstarted/tutorial/Vagrantfile.template
```diff
@@ -12,7 +12,7 @@ ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtualbox'
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = "clusterhq/flocker-tutorial"
-  config.vm.box_version = "= 0.3.2"
+  config.vm.box_version = "= |cli-release|"
```
tomprince: This isn't the `cli-release` version.

adamtheturtle:
> Please
> - [ ] use python rather than envsubst
> - [ ] switch to `|version|`
> - [ ] simplify the the code in `run`

1. `envsubst` is gone but instead of a Python replacement for it Sphinx now generates the files / include text. This allows there to be one less step in the release process.

2. We discussed that Sphinx's `|version|` is not exactly what is wanted. Therefore it is now `|latest-installable|` as this represents the latest installable CLI and node packages. The logic for working this out has also been moved (and is now tested).

3. Done

dwgebler:
Thanks @adamtheturtle this mostly seems OK to me. In reference to your comment about generated files in the `docs/` directory that have been version-substituted, because of the way these extensions work, I believe the generated files such as `linux-install.sh` have to be written to the path they're currently going in, because when you hand over processing back to the Sphinx handlers you inherit from (e.g. `LiteralInclude`) it expects to find those files to copy them to the downloads directory of the HTML output. So writing them to something like `tmp/linux-install.sh` probably won't work.

One option I can think of is to add an event handler to clean up those generated files after the rest of the build process is completed, so something like:

```
versioned_files = list()

def make_changed_file(rel_filename):
    with ....
        ....
    versioned_files.append(new_rel_filename)
```

and then 

```
def PurgeVersionedFiles(app, *args):
    for filename in set(versioned_files):
        os.remove(filename)
```

and

```
def setup(app):
    ....
    app.connect('build-finished', PurgeVersionedFiles)
```

Also, one other thing I found is that using `sphinx-build` on the command line, with the code in this branch, I am unable to build from the root `flocker` directory with `sphinx-build -bhtml docs/ docshtml/` because it will complain about being unable to find files such as `indepth/linux-install.sh.template` when it searches for them on a relative path (so I'd have to go in to `flocker/dpcs/` and run `sphinx-build -bhtml . ../docshtml/` for example)

adamtheturtle:
That second point is a big deal because it means that `tox -e sphinx`, our documented way of building the docs, is broken.

adamtheturtle:
I have addressed the above comments.

dwgebler:
This works correctly for me now. Note @tomprince 's comment at https://github.com/ClusterHQ/flocker/pull/1099/files#r23859249 it looks like the note in the docs he refers to there still needs to go back in, please address that or leave a comment to say why not if it doesn't need to change.

Other than that, I'm happy if Buildbot is happy, please merge.

adamtheturtle:
That comment was made by Tom when there was still a section in the docs for this (albeit a reduced section). No step needs to be skipped for documentation releases now.

PR #1089, opened 2015-01-27
[FLOC-1214] AMP protocol for communication between control service and convergence agents

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1214

PR #1085, opened 2015-01-26
[FLOC-1249] Cluster state service

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1249

PR #1105, opened 2015-01-29
[FLOC-1281] Control service AMP server - design sketch

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1281

Depends on #1089 and #1085.


exarkun:
I think all the proposed changes were in https://github.com/ClusterHQ/flocker/commit/1708346062ffe1611d09dc8bbb0626bc75238c45 so that's the only revision I reviewed.  Very minor comments inline, otherwise looks good.  Please proceed.  Thanks.

PR #1111, opened 2015-01-30
[FLOC-1255] Convergence loop design sketch

itamarst:


exarkun:
Thanks.  Glad to see someone else is going to try using rich inputs. :wink:  Looks good to proceed, please do.

flocker/node/_loop.py
```diff
@@ -0,0 +1,282 @@
...
+            lambda node_state: self.input(_DiscoveredStatus(node_state)))
+        # XXX error case
+
+    def output_REPORT_NODE_STATE(self, symbol, context):
+        self.client.callRemote(NodeStateCommand, node_state=context.node_state)
+
+    def output_CHANGE(self, symbol, context):
+        # XXX need to refactor Deployer slightly so you can pass in NodeState...
+        d = self.deployer.change_node_state(self.configuration, self.cluster_state,
+                                            HOSTNAME)
```
exarkun: While thinking about how the OpenStack agent is going to work, it occurred to me that if we fix the signature of `change_node_state` to omit hostname then this FSM might be usable to drive either the P2P or OpenStack dataset agents.  This would require the P2P deployer changing to be told what host it is operating on via `__init__` instead of `change_node_state`.  That seems fairly easy though.

PR #1119, opened 2015-02-03
[FLOC-1321] New interface for Deployers

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1321

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -66,10 +66,45 @@ class IStateChange(Interface):
         """
 
 
+# XXX will have very minimal set of interface verification tests
+class IDeployer(Interface):
+    """
+    An object that can discover local state and calculate necessary
+    changes to bring local state and desired cluster configuration into
+    alignment.
+    """
+    def discover_local_state(self):
+        """
+        Discover the local state, i.e. the state which is exclusively under
+        the purview of the convergence agent running this instance.
+
+        :return: A ``Deferred`` which fires with an object describing
+             local state.
+        """
+
+    def calculate_necessary_state_changes(local_state,
+                                          desired_configuration,
+                                          current_cluster_state):
+        """
+        Calculate the state changes necessary to make the local state match
+        the desired cluster configuration.
+
+        :param local_state: The recent output of ``discover_local_state``.
+        :param Deployment desired_configuration: The intended
+            configuration of all nodes.
+        :param Deployment current_cluster_state: The current state of all
+            nodes. While technically this may also includes the local
+            state, that information is likely out of date so should be
+            overriden by ``local_state``.
+
+        :return: A ``IStateChange`` provider.
+        """
+
+
 @implementer(IStateChange)
 @attributes(["changes"])
 class Sequentially(object):
-    """
+        """
     Run a series of changes in sequence, one after the other.
 
     Failures in earlier changes stop later changes.
@@ -307,7 +342,8 @@ class SetProxies(object):
         return gather_deferreds(results)
 
 
-class Deployer(object):
+@implementer(IDeployer)
+class P2PNodeDeployer(object):
     """
     Start and stop applications.
 
@@ -317,6 +353,7 @@ class Deployer(object):
     :ivar INetwork network: The network routing API to use in
         deployment operations. Default is iptables-based implementation.
     """
+    # XXX take hostname as argument
     def __init__(self, volume_service, docker_client=None, network=None):
         if docker_client is None:
             docker_client = DockerClient()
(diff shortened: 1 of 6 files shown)


exarkun:
Thanks, looks good.

flocker/node/_deploy.py
```diff
@@ -44,13 +44,13 @@ def _to_volume_name(dataset_id):
...
     """
-    An operation that changes the state of the local node.
+    An operation that changes local state.
     """
     def run(deployer):
         """
         Run the change.
 
-        :param Deployer deployer: The ``Deployer`` to use.
+        :param IDeployer deployer: The ``IDeployer`` to use.
```
exarkun: Not a good docstring before.  Not a good docstring now. :cry

exarkun: Not a good docstring before.  Not a good docstring now. :cry

flocker/node/_deploy.py
```diff
@@ -66,6 +66,40 @@ def __ne__(other):
...
+    changes to bring local state and desired cluster configuration into
+    alignment.
+    """
+    def discover_local_state():
+        """
+        Discover the local state, i.e. the state which is exclusively under
+        the purview of the convergence agent running this instance.
+
+        :return: A ``Deferred`` which fires with an object describing
+             local state.
```
exarkun: Of unspecified type?  Are there any constraints at all?

flocker/node/_deploy.py
```diff
@@ -307,17 +341,22 @@ def run(self, deployer):
         return gather_deferreds(results)
 
 
-class Deployer(object):
+@implementer(IDeployer)
+class P2PNodeDeployer(object):
```
exarkun: :question: characteristic

flocker/node/_deploy.py
```diff
@@ -530,137 +570,132 @@ def calculate_necessary_state_changes(self, desired_state,
...
+                ResizeDataset(dataset=dataset)
+                for dataset in dataset_changes.coming]))
+        if dataset_changes.creating:
+            phases.append(InParallel(changes=[
+                CreateDataset(dataset=dataset)
+                for dataset in dataset_changes.creating]))
+        start_restart = start_containers + restart_containers
+        if start_restart:
+            phases.append(InParallel(changes=start_restart))
+        return Sequentially(changes=phases)
```
exarkun: I hope this was merely de-dented to reflect it no longer needs to run in a callback.

flocker/node/functional/test_deploy.py
```diff
@@ -35,8 +35,10 @@ def test_restart(self):
         """
         name = random_name()
         docker_client = DockerClient()
-        deployer = Deployer(create_volume_service(self), docker_client,
-                            make_memory_network())
+        deployer = P2PNodeDeployer(
+            u"localhost",
```
exarkun: :question: Our docs say "hostname" all over the place but in practice the value is always going to be an IP address.

flocker/node/script.py
```diff
@@ -198,8 +196,13 @@ def __init__(self, docker_client=None, network=None):
         self._network = network
 
     def main(self, reactor, options, volume_service):
-        deployer = Deployer(volume_service, self._docker_client, self._network)
-        d = deployer.discover_node_configuration()
+        # Discovery doesn't actually care about hostname, so don't bother
+        # figuring out correct one. Especially since this code is going
+        # away someday soon.
```
exarkun: An issue link might be nice here.

exarkun:
Thanks.  Looks pretty good.  Only thing I notice is there aren't any simple, direct unit tests for the new `change_node_state` function.  It is exercised a lot by complicated functional tests but a simple unit test for it (independent of any complicated `IDeployed` behavior) would probably be good.

Please add that and then merge.

flocker/node/_deploy.py
itamarst: For the initialization at least that means I have to change all constructors to be keyword-based, don't have the energy to do that at the moment.

flocker/node/_deploy.py
itamarst: Pretty much. Other change was referring to `self.hostname` instead of passed in one, and using passed in local state. But no significant changes (which is why all the unit tests changes were trivial).

flocker/node/functional/test_deploy.py
itamarst: Is it? We've seen issues with external IP detection, for example, and I would argue that hostnames are more stable than IPs. Which could be completely wrong. Anyway seems like something to discuss separately.

itamarst:
Pretty sure `ChangeNodeStateTests` are isolated unit tests that cover `change_node_state` in the way you want. Not the best tests ever, but it's coverage.

PR #1123, opened 2015-02-04
[FLOC-1250] Custom 404 generated by Sphinx

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1250

 * Thought it might be neat to generate the error page using sphinx to avoid duplicating the content in the template.
 * Only trouble is that the 404 may be served from anywhere in the URL hierarchy, so relative links in the template will not always work.
 * So I conditionally generate a `<base href` tag to set the actual URL of the 404 page.
 * Only trouble is that you have to specify an absolute base url, which means that the links on the 404 page will point to the current (old) docs.

Alternative is to create a completely separate template where all the links are absolute or  "root-relative".

Or update  the main template to generate absolute URLs.

To properly test this, I (or reviewer) need to upload the generated 404.html page to https://s3.amazonaws.com/clusterhq-staging-docs/404.html using gsutil.

Also note that the error page will be shown for all 40x errors so it may not make sense to call it `404.html`. Then again, it's possible that we could configure the custom error page using cloudfront, which does allow separate error pages for each error status code.


docs/error_pages/404.rst
```diff
@@ -0,0 +1,8 @@
+ Error 404
+===========
+
+The page you requested does not exist.
+
+The page may have moved.
+
+Try searching or click the navigation links on the left.
```
adamtheturtle: Navigation links aren't always on the left (on small window sizes they are above, in a menu).

exarkun:
This should probably be against `release/flocker-0.3.2` since we'll have to do a doc release to deploy it.

docs/error_pages/404.rst
```diff
@@ -0,0 +1,8 @@
+ Error 404
+===========
```
exarkun: Are the extra spaces around this heading important?  I'm not familiar with that rst construct.

exarkun:
Sphinx is going to complain about an unlinked page, I expect.

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -2,6 +2,9 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
+    {% if pagename == 'error_pages/404' %}
+    <base href="http://docs.clusterhq.com/en/latest/error_pages/404.html">
+    {% endif %}
```
exarkun: Probably worth leaving a comment here explaining what's going on.

docs/error_pages/404.rst
tomprince: It looks like they shouldn't be there, as this is rendering as a header (and there is no title in the generate document).

tomprince:
For compairson, (this)[https://docs.clusterhq.com/404.html] is what it currently looks like (obtained from grabbing a random non-existent URL from the main site).

wallrj:
Ready for another review.

exarkun:
Thanks.  This looks pretty good.  However,

> This should probably be against release/flocker-0.3.2 since we'll have to do a doc release to deploy it.

tomprince:
> This should probably be against release/flocker-0.3.2 since we'll have to do a doc release to deploy it.

The 404 page isn't actually part of the documentation. It lives at /404.html in s3 bucket, not under `/en/<version>/`, so we don't need to base it on 0.3.2 to deploy it.

wallrj:
I modified the `publish-docs` command to also modify the error_key in the target bucket. 

To test, I:
 * Built docs locally and uploaded to the dev bucket using `$ s3cmd sync --no-mime-magic --guess-mime-type --delete-removed --no-preserve html/ s3://clusterhq-dev-docs/0.0.0dev4/`
 * Disabled the cloudfront related steps and changed the staging bucket in the script to a new `test` bucket.

You can see the results here:
 * http://clusterhq-testing-docs.s3-website-us-east-1.amazonaws.com/en/0.0.0dev4/index.html
 * http://clusterhq-testing-docs.s3-website-us-east-1.amazonaws.com/en/0.0.0dev4/page/not/found

admin/release.py
```diff
@@ -202,6 +205,12 @@ def publish_docs(flocker_version, doc_version, environment):
                      prefix=version_prefix,
                      keys=existing_version_keys - new_version_keys))
 
+
+    # Update the key used for error pages
+    yield Effect(
```
tomprince: I wonder if we want to do this for dev releases?

wallrj:
In https://github.com/ClusterHQ/flocker/pull/1123#discussion_r25279900 @tomprince wrote:
> I wonder if we want to do this [update 404 error page] for dev releases?

Hmm. I see what you mean.
Perhaps I can update the 404 page on docs.staging.clusterhq.com, so that we can observe how template changes affect the error page.
I'll make some changes.

admin/test/test_release.py
```diff
@@ -661,3 +661,56 @@ def test_publish_non_release_fails(self):
...
+            s3_buckets={
+                'clusterhq-staging-docs': {},
+                'clusterhq-dev-docs': {},
+            }
+        )
+        expected_version = '0.3.1dev5'
+        expected_error_key = 'en/{}/error_pages/404.html'.format(
+            expected_version)
+        self.publish_docs(aws, '0.3.0-444-gf01215b', expected_version,
+                          environment=Environments.STAGING)
```
exarkun: Does this unit test actually change some live S3 configuration?

exarkun: Oh, looks like not.  Hence `FakeAWS`.

admin/release.py
exarkun: Seems like most likely not.  As I understand things, this `Effect` will cause the error page (404 or otherwise?) for the entire `docs.clusterhq.com` site to be changed to the 404 error page in the package being released by this code.

If this code were only updating the error page for the version of the documentation corresponding to the release being made, it would make sense to do the update (I don't know off hand whether that's actually possible - I could imagine being allowed to specify an error page to handle errors that occur below some prefix, but I haven't looked at CloudFront to learn if that's actually supported or not).  I don't think it makes sense to do the update for all error pages on the entire site - since the point of dev releases isn't that we have something polished that we want users to see, just that we want to exercise the release process.

Of course, if we don't do this update then we're not exercising this part of the release process...  So that's a bit sad.  We do having the staging bucket, though.  So probably that's how we'll exercise this code?

Long story short, I agree that it makes sense to skip this update if `is_dev` is true.

exarkun:
Thanks.  Sorry about the long turn-around.  I think this is great apart from the question of whether to update the error page for development releases (and as I mentioned inline, I think we probably shouldn't).

Please address that point and then merge.  Thanks.

adamtheturtle:
There is a TODO for this issue in docs/gettinginvolved/infrastructure/documentation.rst, so this PR should remove or update that.

wallrj:
@tomprince Would you mind taking one more look at this. 

I've made some changes to the tests to verify the behaviour that we decided upon.

I tested the script as follows:

```
tox -e sphinx
s3cmd sync --no-mime-magic --guess-mime-type --delete-removed --no-preserve docs/_build/html/ s3://clusterhq-dev-docs/0.999.0/
./admin/publish-docs --doc-version=0.999.0 --flocker-version=0.999.0
./admin/publish-docs --doc-version=0.999.1dev1 --flocker-version=0.999.0
```

You should then see the new error page when you visit:
 * https://docs.staging.clusterhq.com/en/latest/foo/bar/baz/qux
 * https://docs.staging.clusterhq.com/en/devel/foo/bar/baz/qux

While experimenting, I published the docs to 0.0.0 but there was already a redirect from 0.0.0 > latest. 
And I'm getting some lingering infinite redirects because of that. (see attached screenshot)

Is this due to a bug in our cloudfront cache invalidation Effect? Perhaps those files aren't being included in the invalidation request. I haven't investigated that yet.

screenshot from 2015-04-15 12 44 19

docs/gettinginvolved/infrastructure/documentation.rst
```diff
@@ -151,15 +151,6 @@ There are empty files at ``/index.html`` and ``/en/index.html`` that redirect to
    gsutil -h x-amz-website-redirect-location:/en/latest/ cp - s3://clusterhq-docs/index.html </dev/null
    gsutil -h x-amz-website-redirect-location:/en/latest/ cp - s3://clusterhq-docs/en/index.html </dev/null
 
-.. TODO - Specify where this is versioned. https://clusterhq.atlassian.net/browse/FLOC-1250
-
-There is an ``error.html`` uploaded to the root of the bucket. It is uploaded with:
```
tomprince: We should document where the error page comes from now, and when it is updated.

admin/test/test_release.py
```diff
@@ -688,6 +688,138 @@ def test_publish_non_release_fails(self):
...
+        empty_routes = {
+            'en/devel/': 'en/0.0.0/',
+            'en/latest/': 'en/0.0.0/',
+        }
+        # In all the S3 buckets.
+        empty_routing_rules = {bucket_name: empty_routes.copy() for bucket_name in bucket_names}
+        # And that all the buckets themselves are empty.
+        empty_buckets = {bucket_name: {} for bucket_name in bucket_names}
+        # And that all the buckets have an empty error_key
+        empty_error_keys = {bucket_name: {} for bucket_name in bucket_names}
```
tomprince: An empty `dict` isn't the right type for a default value. In any case, the dev bucket probably shouldn't have an error key set.

admin/test/test_release.py
```diff
@@ -688,6 +688,138 @@ def test_publish_non_release_fails(self):
...
+                DOCUMENTATION_CONFIGURATIONS[e].dev_bucket
+            )
+        # Pretend that both devel and latest aliases are currently pointing to
+        # an older version.
+        empty_routes = {
+            'en/devel/': 'en/0.0.0/',
+            'en/latest/': 'en/0.0.0/',
+        }
+        # In all the S3 buckets.
+        empty_routing_rules = {bucket_name: empty_routes.copy() for bucket_name in bucket_names}
```
tomprince: The dev bucket doesn't have any routes.

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -2,6 +2,10 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
+    {% if pagename == 'error_pages/404' %}
+    <!-- The 404 error page may be served at any depth in the URL hierarchy. -->
+    <base href="/en/latest/error_pages/404.html">
```
tomprince: I think it doesn't matter, but it may be worth a comment: When publishing a dev release to the staging bucket, this base-ref will still refer to the latest marketing release.

tomprince:
Couple of minor comments, then merge.

PR #1125, opened 2015-02-04
Increased ipaddr required version

adamtheturtle:
Partial fix for FLOC-1337

There is a current discussion about whether the problem that this solves can be solved without changing anything in the Python code, because that would mean having to create a new wheel.


adamtheturtle:
No longer necessary.

PR #1117, opened 2015-02-03
[FLOC-1300] Release Flocker 0.3.2+doc1

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1300

Pre-tag Review Checklist
- [ ] Do the acceptance tests. (although perhaps this can be skipped, as no changes were made to the running code)
- [ ] Check documentation.
  - [ ] The documentation should be available at https://docs.staging.clusterhq.com/en/0.3.2/
  - [ ] The following URLs should redirect to the above URL.
     - https://docs.staging.clusterhq.com/
     - https://docs.staging.clusterhq.com/en/
     - https://docs.staging.clusterhq.com/en/latest/
 
    In addition, check that deep-links to /en/latest/ work. https://docs.staging.clusterhq.com/en/latest/authors.html should redirect to https://docs.staging.clusterhq.com/en/0.3.2/authors.html
- [ ] Additional checks, since this is the first release on s3:
   - Check that, for all versions on [readthedocs](https://readthedocs.org/projects/flocker/), we redirect to the latest version, both for the root, and for deep links.

I'm not sure the above list is exhaustive.

The branch should not be merged until the release process is complete.

PR #1126, opened 2015-02-04
Last minute bug fixes in master.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1328
https://clusterhq.atlassian.net/browse/FLOC-1330
https://clusterhq.atlassian.net/browse/FLOC-1329


docs/gettingstarted/index.rst
```diff
@@ -173,7 +173,9 @@ Getting started with Flocker
             you@laptop:~$ git clone \
               https://github.com/clusterhq/vagrant-flocker && \
               cd vagrant-flocker && \
-              vagrant up
+              vagrant up && \
+              [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent) && \
+              ssh-add ~/.vagrant.d/insecure_private_key
```
wallrj: What we found yesterday is that if an agent isn't running, it will be started in the second console that was used for starting the vagrant nodes, but the agent socket path and the loaded key will not then be available in the first console where the user has installed `flocker-deploy`.

Those of us testing on Ubuntu / Fedora usually don't see this because the desktop an alternative key agent eg gnome-keyring-daemon, which exports the SSH_AUTH_SOCK environment variable, which is  inherited by all child shells.

```
[~/flocker-tutorial-ubuntu]$ sudo lsof -p $(pidof gnome-keyring-daemon) | grep "$SSH_AUTH_SOCK" 
gnome-key 1696 richard   14u     unix 0xffff8800c5ffcd00      0t0   30296 /run/user/1000/keyring-nkE1ut/ssh
```

docs/gettingstarted/index.rst
```diff
@@ -95,7 +95,7 @@ Getting started with Flocker
 
          .. code-block:: console
 
-            you@laptop:~$ sudo apt-get update && apt-get install -y gcc python2.7 python-virtualenv python2.7-dev && \
+            you@laptop:~$ sudo apt-get update && sudo apt-get install -y gcc python2.7 python-virtualenv python2.7-dev && \
```
wallrj: Why do we need to include the `apt-update` step? I imagine that might have been included because the author was testing in a docker container, which starts up without apt-update ever having been run.

wallrj: I notice we don't do `yum update` before running the Fedora installation commands.

docs/gettingstarted/index.rst
```diff
@@ -270,6 +274,7 @@ Getting started with Flocker
    .. tutorial-step-condensed::
 
       Now we are going to use a different deployment configuration to show moving the Redis container with its data volume.
+      The web server will remain deployed on the first host and remain accessible via either host's address.
```
wallrj: Personally, I think the tutorial should end with another clickable link and a screen shot showing that the page has been "visited 2 times"...ie the redis data for the first visit has been moved to the new node.
 * https://github.com/ClusterHQ/flocker/pull/1117#issuecomment-72848241

wallrj:
Thanks @tomprince 

I think:
 * the additional of `sudo` in the ubuntu installation instructions is fine
 * the ssh-agent changes need more thought (see above)
 * the instructions to revisit the web app need further explanation (see comment above)

docs/gettingstarted/index.rst
exarkun: The docs don't tell you to launch multiple shells.  It's certainly true that if you do, this isn't going to help you.  We could explicitly tell people to run all the commands in one shell.  I wonder what prompted the idea that you should use multiple shells for the guide, though?

Since this is what's actually deployed now, I don't think this should block the merge (but that doesn't mean we shouldn't do some follow-up work to make this better).

docs/gettingstarted/index.rst
exarkun: `yum` updates repository data automatically.  `apt-get` doesn't.  If any of the package versions have changed since you ran `apt-get update` last then you might get an error from the `apt-get install` command.

wallrj:
Ignore previous comment. 
I hadn't realised that this was a proposed forward merge of already reviewed changes in the 0.3.2doc5 release branch.
I'll merge it.

PR #1128, opened 2015-02-05
readme: fix tutorial link

TomasTomecek:
https://docs.clusterhq.com/en/latest/gettingstarted/tutorial/ is 404


exarkun:
Thanks for finding this.  While looking into this, @ferrantim noticed that another link wasn't ideal and created #1129 which also fixes this problem so I'm going to close this PR in favor of that one.

PR #1118, opened 2015-02-03
[FLOC-1316] Release/flocker 0.3.3dev5

jongiddy:
This branch should not be merged until the release process is complete.

https://clusterhq.atlassian.net/browse/FLOC-1316

PR #1131, opened 2015-02-05
[FLOC-1356] Quickstart version numbers

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1356

This could / should also be merged into #1118


adamtheturtle:
Note that this fixes part of https://clusterhq.atlassian.net/browse/FLOC-1307?jql=text%20~%20%22release%20process%22%20and%20status%20!%3D%20done (for which you have another PR in design review).

adamtheturtle:
This seems to work for all instances of the version number both with and without JavaScript, thanks.

PR #1133, opened 2015-02-06
[FLOC-1307] Integrate quickstart configuration files 

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1307

* An (unfinished) attempt to import Vagrantfile and flocker configuration files which are currently held in separate repositories.
* If they stay in separate repo's then we'll need to tag those repos with each released version and add a `checkout <version>` step to the quickstart guide....which is a pain.
* Instead, this is an attempt to use the `|latest-version|` sphinx helper extensions, but
* I haven't quite figured out how to render the Vagrantfile.template file, without including it's content in the quickstart guide...
* I just want the template rendered and then to include eg `wget https://docs.clusterhq.com/en/|latest-installable|/gettingstarted/nodes/Vagrantfile` in the instructions.
* Also not sure how to make that wget command easily testable at the "pre-tag" review stage of each release.

May need help from @adamtheturtle .


dwgebler:
Sphinx doesn't like having two downloadable files with the same name, because it copies everything in to one `downloads` directory in its output, so if we have two `Vagrantfile` files, one of them will get renamed to something like `Vagrantfile.1` (I forget off hand exactly what the convention is).

dwgebler:
Yep, http://build.clusterhq.com/results/downloadable-vagrantfiles-FLOC-1307/flocker-4084/docs/_downloads/  -see `Vagrantfile1`

adamtheturtle:
> Also not sure how to make that wget command easily testable at the "pre-tag" review stage of each release.

What in particular do you want to test?

adamtheturtle:
These aren't wonderful ideas (just before the weekend) but:
* Include the template in an orphan unseen file
* Make a new directive which doesn't display the content and inline the files above when you want to use them `.. new-directive:: Vagrantfile.template`
* Make a function to create the files you want on http://sphinx-doc.org/extdev/appapi.html#event-builder-inited.

adamtheturtle:
> If they stay in separate repo's then we'll need to tag those repos with each released version and add a checkout <version> step to the quickstart guide....which is a pain.

You can download a file from a repository without cloning a repository (e.g. `$ wget https://raw.githubusercontent.com/ClusterHQ/homebrew-tap/master/flocker-0.3.2.rb`).

docs/gettingstarted/index.rst
```diff
@@ -168,14 +168,16 @@ Getting started with Flocker
...
 
-            you@laptop:~$ git clone \
-              https://github.com/clusterhq/vagrant-flocker && \
-              cd vagrant-flocker && \
-              vagrant up && \
-              [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent) && \
-              ssh-add ~/.vagrant.d/insecure_private_key
+            mkdir vagrant-flocker
+            cd vagrant-flocker
+            wget https://docs.clusterhq.com/en/|latest-installable|/gettingstarted/nodes/Vagrantfile
```
adamtheturtle: OS X at least does not have `wget` without e.g. `brew install wget`.

docs/gettingstarted/index.rst
```diff
@@ -168,14 +168,16 @@ Getting started with Flocker
...
+            vagrant up
+            [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent)
```
adamtheturtle: :question: Optional because this is an unrelated change but a comment explaining what this does would make it seem less risky to paste into my shell.

docs/gettingstarted/index.rst
```diff
@@ -189,14 +191,16 @@ Getting started with Flocker
 
          Simulate a Flocker cluster with virtual machines on your laptop (requires `Vagrant <http://www.vagrantup.com/downloads>`_, `VirtualBox <https://www.virtualbox.org/wiki/Downloads>`_):
 
-         .. code-block:: console
+         :version-download:`nodes/Vagrantfile.template`
```
adamtheturtle: To confirm, what you want is to have this generate the Vagrantfile but not have a download link, right?

docs/gettingstarted/index.rst
```diff
@@ -189,14 +191,16 @@ Getting started with Flocker
...
-              https://github.com/clusterhq/vagrant-flocker && \
-              cd vagrant-flocker && \
-              vagrant up && \
-              [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent) && \
-              ssh-add ~/.vagrant.d/insecure_private_key
+         .. version-code-block:: console
+
+            mkdir vagrant-flocker
+            cd vagrant-flocker
+            wget https://docs.clusterhq.com/en/|latest-installable|/gettingstarted/nodes/Vagrantfile
```
adamtheturtle: Instead of generating the file, perhaps we can use the release branch (fragile?), e.g. `$ wget https://raw.githubusercontent.com/ClusterHQ/flocker/release/flocker-|latest-installable|/README.rst`, but with a path to the Vagrantfile instead of the README.

adamtheturtle:
I've put a few ideas down. If any of them appeal, go with it and re-submit. Else ask me and I'll pair with you on finding a better solution.

docs/gettingstarted/index.rst
exarkun: It does ship with curl though (which is probably better in general).

docs/gettingstarted/index.rst
exarkun: The previous scripts use `&&` to join the commands intentionally.  That pattern stops the command sequence on the first failure.

adamtheturtle:
The Vagrantfile issue can be worked around with `VAGRANT_VAGRANTFILE` http://docs.vagrantup.com/v2/other/environmental-variables.html.

adamtheturtle:
This was used as a starting point for follow-up issues, so I'm closing.

PR #1136, opened 2015-02-06
[FLOC-1255] Convergence loop

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1255

This does not include any end-to-end testing, since the branch is too big as it is. That will be done in follow up issue, and will no doubt find some more problems.


flocker/node/_loop.py
```diff
@@ -0,0 +1,344 @@
...
+    # The system is shutting down:
+    SHUTDOWN = NamedConstant()
+
+
+@attributes(["client"])
+class _ClientConnected(trivialInput(ClusterStatusInputs.CLIENT_CONNECTED)):
+    """
+    A rich input indicating the client has connected.
+
+    :ivar AMP client: An AMP client connected to the control service.
```
tomprince: `client` seems a little confusing, since it is a connection to the controller.

flocker/node/_loop.py
```diff
@@ -0,0 +1,344 @@
...
+        self.convergence_loop_fsm.receive(
+            _ClientStatusUpdate(client=self.client,
+                                configuration=context.configuration,
+                                state=context.state))
+
+    def output_STOP(self, context):
+        self.convergence_loop_fsm.receive(ConvergenceLoopInputs.STOP)
+
+    def output_DISCONNECT(self, context):
+        self.client.transport.loseConnection()
```
tomprince: Do we want to set `client` to `None` here?

flocker/node/_loop.py
```diff
@@ -0,0 +1,344 @@
...
+    :ivar port: Port to connect to.
+    :ivar cluster_status: A cluster status FSM.
+    :ivar factory: The factory used to connect to the control service.
+    """
+
+    def __init__(self):
+        MultiService.__init__(self)
+        convergence_loop = build_convergence_loop_fsm(self.deployer)
+        self.cluster_status = build_cluster_status_fsm(convergence_loop)
+        self.factory = ReconnectingClientFactory()
```
tomprince: :exclamation: Use `Factory.forProtocol`?

flocker/node/_loop.py
```diff
@@ -0,0 +1,344 @@
...
+    O = ClusterStatusOutputs
+    table = TransitionTable()
+    # We may be shut down in any state, in which case we disconnect if
+    # necessary.
+    table = table.addTransitions(
+        S.DISCONNECTED, {
+            # Store the client, then wait for cluster status to be sent
+            # over AMP:
+            I.CLIENT_CONNECTED: ([O.STORE_CLIENT], S.IGNORANT),
+            I.SHUTDOWN: ([], S.SHUTDOWN),
```
tomprince: :question: While first looking at this, I wondered why this doesn't stop trying to connect the client. It looks like that happens outside of the state machine. Perhaps a comment *somewhere* (I don't know where) would be appropriate.

tomprince: Particularly since the state machine *does* handle disconnecting.

flocker/node/_loop.py
```diff
@@ -0,0 +1,344 @@
...
+    # Store AMP client, desired configuration and cluster state for later
+    # use:
+    STORE_INFO = NamedConstant()
+    # Start an iteration of the covergence loop:
+    CONVERGE = NamedConstant()
+
+
+class ConvergenceLoop(object):
+    """
+    World object the convergence loop state machine, executing the actions
```
tomprince: :exclamation: Missing some words here.

tomprince:
:exclamation: It would be *really* nice if there was some prose documentation (in `gettinginvolved` maybe) explaining the structure. I'm finding that I have reconstruct the architecture, to understand the code enough to review it. Perhaps this can be done in a later issue though (although not having it probably makes doing follow up issues more complicated).

setup.py
```diff
@@ -81,7 +81,7 @@
             'flocker-deploy = flocker.cli.script:flocker_deploy_main',
             'flocker-changestate = flocker.node.script:flocker_changestate_main',
             'flocker-reportstate = flocker.node.script:flocker_reportstate_main',
-            'flocker-zfs-agent = flocker.node.script:flocker_volume_main',
+            'flocker-zfs-agent = flocker.node.script:flocker_zfs_agent_main',
```
tomprince: :exclamation:  This probably requires an adjustment to the packaging.

flocker/control/test/test_protocol.py
```diff
@@ -317,12 +318,8 @@ def test_configuration_change(self):
 class FakeAgent(object):
     """
     Fake agent for testing.
-
-    Not a full verified fake since
-    https://clusterhq.atlassian.net/browse/FLOC-1255 may change this a
-    little.
     """
-    def connected(self):
+    def connected(self, client):
```
tomprince: :exclamation: This should record the client and something should test that it gets passed the correct object.

flocker/node/test/test_loop.py
```diff
@@ -0,0 +1,583 @@
...
+    def assertConvergenceLoopInputted(self, expected):
+        """
+        Assert that that given set of symbols were input to the agent
+        operation FSM.
+        """
+        self.assertEqual(self.agent_operation.inputted, expected)
+
+    def test_creation_no_side_effects(self):
+        """
+        Creating the FSM has no side effects.
```
tomprince: :exclamation: The test is only testing for one specific kind of side effect.

tomprince:
:question: The issue suggests that the local status should be sent on connect. The implementation only sends it once it has received the cluster status.

flocker/node/test/test_loop.py
```diff
@@ -0,0 +1,583 @@
...
+        notification is needed for agent operation FSM.
+        """
+        self.fsm.receive(_ClientConnected(client=build_protocol()))
+        self.fsm.receive(ClusterStatusInputs.CLIENT_DISCONNECTED)
+        self.assertConvergenceLoopInputted([])
+
+    def test_disconnect_after_status_update(self):
+        """
+        If the client disconnects after a status update is received then the
+        agent operation is FSM is notified.
```
tomprince: :exclamation: `is` appears to be a typo.

flocker/node/test/test_loop.py
itamarst: Since it has no references to any other objects, talking to convergence loop is the only side-effect we care about. I mean yes it could do `os.remove` or whatever, but you can't test *everything*.

flocker/node/test/test_loop.py
```diff
@@ -0,0 +1,583 @@
...
+        for local_state in local_states:
+            client.register_response(
+                NodeStateCommand, dict(node_state=local_state),
+                {"result": None})
+        return client
+
+    def test_convergence_done_notify(self):
+        """
+        A FSM doing convergence that gets a discovery result notifies the last
+        received client.
```
tomprince: :exclamation: of what?

flocker/node/test/test_loop.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+        A FSM doing a convergence iteration does another iteration when
+        applying changes is done.
+        """
+        local_state = object()
+        local_state2 = object()
+        configuration = object()
+        state = object()
+        action = ControllableAction(succeed(None))
+        action2 = ControllableAction(Deferred())
```
tomprince: :exclamation:  It isn't at all obvious that argument here controls whether this action ends up running or not. It took a while of staring at the code to figure out why this tests runs two loops and the above test only runs one.

tomprince: Particularly, I was puzzled why if it ran twice, it didn't run an infinite number of times.

flocker/node/test/test_loop.py
```diff
@@ -0,0 +1,583 @@
...
+        # Action finally finishes, and we can move on to next iteration,
+        # which happens with second set of client, desired configuration
+        # and cluster state:
+        action.result.callback(None)
+        self.assertEqual(
+            (deployer.calculate_inputs, client.calls, client2.calls),
+            ([(local_state, configuration, state),
+              (local_state2, configuration2, state2)],
+             [(NodeStateCommand, dict(node_state=local_state))],
+             [(NodeStateCommand, dict(node_state=local_state2))]))
```
tomprince: :question: Something about testing that
- if multiple status updates are received before an iteration is done, that last result is used
- the second iteration doesn't start until the first is done.

tomprince:
There are a number of minor changes, and things that could be clarified, but it mostly looks good. Please address the above points to your satisfaction and merge.

setup.py
itamarst: Looks fine, AFAICT.

itamarst:
Sending local status only after global cluster state is received is fine, since there's logic on control service side to send that automatically to whoever connects (`flocker.control._protocol.ControlAMPService.connected`).

setup.py
tomprince: My eyes must have glazed over; I thought the binary had changed.

PR #1138, opened 2015-02-08
Use extra to install flake8 in tox.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1372

Otherwise, we automatically get newer versions, which cause spurious failures.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,9 +22,10 @@ basepython = python2.7
 
 [testenv:lint]
 basepython = python2.7
-deps = flake8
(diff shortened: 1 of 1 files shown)


jongiddy:
Patch looks fine.  Only problem is failing build of tutorial box, but is this due to the version problem from upgrades to `pip` in Buildbot?  If so, we have a chicken'n'egg problem, where the fix for FLOC-1366 fails due to lint being broken, and this patch fails due to FLOC-1366.

PR #1140, opened 2015-02-09
[FLOC-1370] Dataset creation acceptance test

itamarst:
Depends on #1141 being merged.


wallrj:
Thanks @itamarst 

Looks like this is built on top of the Convergance agent branch, so I've only reviewed `flocker/acceptance/test_api.py`.

I agree that it looks a little ugly as designed, but I'm sure the end result will be more elegant.

Consider:
 * Documenting the node requirements for the test to run (number of nodes and how to configure the IP address of the node under test)
 * Perhaps test the situation when a volume with the requested name already exists on the node. (it looked like that was why this is blocked by #1141...but I may be wrong)
 * Are there followup tickets for packaging `flocker-node` so that flocker-control and flocker-zfs-agent are started automatically by the distro init system? Rather than having to start them as background SSH processes.

Anway, it's exciting that we're almost ready to demo the API. Please proceed.

itamarst:
I eventually got a passing test run: http://build.clusterhq.com/builders/flocker%2Facceptance%2Fvagrant%2Ffedora-20/builds/505

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+def wait_for_api(hostname):
+    """
+    Wait until REST API is available.
+
+    :return Deferred: Fires when REST API is available.
+    """
+    def api_available():
+        try:
+            s = socket.socket()
+            s.connect((hostname, 4523))
```
exarkun: Please factor away the duplication of this port number.

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+
+from twisted.trial.unittest import TestCase
+
+from treq import get, post, content
+
+from .testtools import get_nodes, _run_SSH
+from ..testtools import loop_until
+
+
+def wait_for_api(hostname):
```
exarkun: missing docs for hostname

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+
+class DatasetAPITests(TestCase):
+    """
+    Tests for the dataset API.
+    """
+    def test_dataset_creation(self):
+        """
+        A dataset can be created on a specific node.
+        """
+        # This is blocking for now, may as well due this the succinct way:
```
exarkun: Seems like the wrong shortcut to take.  If it's blocking, maybe it shouldn't return a Deferred.  If it returns a Deferred, you should add a callback.  `successResultOf` is not for this.

flocker/node/script.py
```diff
@@ -227,14 +227,20 @@ class ZFSAgentOptions(Options):
...
+        "Usage: flocker-zfs-agent [OPTIONS] <local-hostname> "
+        "<control-service-hostname>")
 
     optParameters = [
         ["destination-port", "p", 4524,
          "The port on the control service to connect to.", int],
     ]
 
-    def parseArgs(self, host):
+    def parseArgs(self, hostname, host):
```
exarkun: Oh hey that's useful.  Now a two node cluster is possible with `socat`. :wink:

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+                      b"", None, True)
+        # https://clusterhq.atlassian.net/browse/FLOC-1382
+        p2 = _run_SSH(22, 'root', node_1,
+                      [b"flocker-zfs-agent", node_1, b"localhost"],
+                      b"", None, True)
+
+        def close(process):
+            process.stdin.close()
+            kill(process.pid, SIGINT)
+        self.addCleanup(close, p1)
```
exarkun: Probably should do this cleanup before trying `p2` in case that `_run_SSH` fails.

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+        self.addCleanup(close, p2)
+
+        d = wait_for_api(node_1)
+
+        uuid = unicode(uuid4())
+        dataset = {u"primary": node_1,
+                   u"dataset_id": uuid,
+                   u"metadata": {u"name": u"my_volume"}}
+        d.addCallback(
+            lambda _: post("http://{}:4523/v1/datasets".format(node_1),
```
exarkun: another duplication of that port number

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+                           persistent=False))
+        d.addCallback(content)
+
+        def got_result(result):
+            result = loads(result)
+            self.assertEqual(dataset, result)
+        d.addCallback(got_result)
+
+        def created():
+            result = get("http://{}:4523/v1/state/datasets".format(node_1),
```
exarkun: port again

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Tests for the control service REST API.
+"""
+
```
exarkun: `from __future__ import unicode_literal`, maybe?  The code does some text formatting on bytes as written now.

exarkun: Or just check every string and make sure it's either b"" or u"" and not just "".

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+        d.addCallback(got_result)
+
+        def created():
+            result = get("http://{}:4523/v1/state/datasets".format(node_1),
+                         persistent=False)
+            result.addCallback(content)
+
+            def got_body(body):
+                body = loads(body)
+                # Current state listing doesn't include metadata.
```
exarkun: Or rather, it includes empty metadata all the time.  A link to https://clusterhq.atlassian.net/browse/FLOC-1386 might make sense here.

flocker/acceptance/testtools.py
```diff
@@ -117,7 +117,8 @@ def get_node_state(node):
     return FlockerConfiguration(state).applications()
 
 
-def _run_SSH(port, user, node, command, input, key=None):
+def _run_SSH(port, user, node, command, input, key=None,
+             background=False):
```
exarkun: `background` is undocumented and definitely not obvious what it does from the name :smile: Good thing we're going to get rid of all this SSH stuff soon.

flocker/node/test/test_script.py
```diff
@@ -416,7 +416,7 @@ def test_main_starts_service(self):
         """
         service = Service()
         options = ZFSAgentOptions()
-        options.parseOptions([b"example.com"])
+        options.parseOptions([b"1.2.3.4", b"example.com"])
```
exarkun: :question: Positional arguments are so lame, though.  `--local-name ... --control-service ...` is so much more readable.

flocker/acceptance/test_api.py
exarkun: At your request, this is now `/v1/configuration/datasets` :wink:

exarkun:
Thanks.  Looks great.  Please address those minor inline points and then merge.

flocker/acceptance/test_api.py
```diff
@@ -0,0 +1,96 @@
...
+from signal import SIGINT
+from os import kill
+from uuid import uuid4
+from json import dumps, loads
+
+from twisted.trial.unittest import TestCase
+
+from treq import get, post, content
+
+from .testtools import get_nodes, _run_SSH
```
adamtheturtle: Importing _run_SSH means that it shouldn't be private.

flocker/node/test/test_script.py
itamarst: OTOH it's less clear that they're required? We can change that in some other issue if you want. In practice I doubt humans will run these much.

PR #1101, opened 2015-01-29
[FLOC-1271] API endpoint to get current dataset state

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1271

PR #1141, opened 2015-02-09
[FLOC-389] Make convergence loop take into account local state

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-389

### DIFF
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -253,7 +253,9 @@ class Deployment(object):
 
     def update_node(self, node):
         """
-        Replace existing ``Node`` with updated version.
+        Create new ``Deployment`` based on this one which replaces existing
+        ``Node`` with updated version, or just adds given ``Node`` if no
+        existing ones have matching hostname.
 
         :param Node node: An update for ``Node`` with same hostname in
              this ``Deployment``.
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -260,7 +260,9 @@ class Deployment(object):
 
         :return Deployment: Updated with new ``Node``.
         """
-        # XXX implement, test
+        return Deployment(nodes=frozenset(
+            list(n for n in self.nodes if n.hostname != node.hostname) +
+            [node]))
 
 
 @attributes(['internal_port', 'external_port'])
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -251,10 +251,21 @@ class Deployment(object):
             for application in node.applications:
                 yield application
 
+    def update_node(self, node):
+        """
+        Replace existing ``Node`` with updated version.
+
+        :param Node node: An update for ``Node`` with same hostname in
+             this ``Deployment``.
+
+        :return Deployment: Updated with new ``Node``.
+        """
+        # XXX implement, test
+
 
 @attributes(['internal_port', 'external_port'])
 class Port(object):
-    """
+        """
     A record representing the mapping between a port exposed internally by an
     application and the corresponding port exposed to the outside world.
 
@@ -337,3 +348,13 @@ class NodeState(object):
         are present on the node but are not attached as volumes to any
         applications.
     """
+    def to_node(self):
+        """
+        Convert into a ``Node`` instance.
+
+        :return Node: Equivalent ``Node`` object.
+        """
+        # XXX will need tests
+        return Node(hostname=self.hostname,
+                    other_manifestations=self.other_manifestations,
+                    application=frozenset(self.running + self.not_running))
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -269,7 +269,7 @@ class Deployment(object):
 
 @attributes(['internal_port', 'external_port'])
 class Port(object):
-        """
+    """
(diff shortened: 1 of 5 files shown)


wallrj:
Thanks @itamarst 

I had a notion that we already double checked the local enumerated volumes rather than rely on the cluster state provided by `flocker-deploy` but I couldn't find it.

I can see that this will address the issue described, but I'm not clear about how this blocks the Volumes API acceptance test? Can you elaborate?

And if the local state is found to be different than the supplied cluster state, what should happen? The API currently only supports volume creation and enumeration, so are you planning a test that triggers an error when a volume of the requested name already exists.

Anyway, please proceed.

itamarst:
This blocks acceptance test because it causes it to fail: without this there is a race condition where the convergence agent on its *second* loop tries to create the dataset again, because it already exists but the agent doesn't notice since it's relying on still stale cluster state.

If local state is different than cluster state that's fine and expected; there's lag in control service being notified of convergence agent's local state. The discovered local state therefore unconditionally overrides cluster state.

flocker/control/_model.py
```diff
@@ -251,10 +251,23 @@ def applications(self):
             for application in node.applications:
                 yield application
 
+    def update_node(self, node):
+        """
+        Replace existing ``Node`` with updated version.
```
wallrj: Perhaps the docstring could be a little clearer that this returns an updated copy of this Deployment rather than updating in-place.

flocker/control/_model.py
```diff
@@ -251,10 +251,23 @@ def applications(self):
...
+        """
+        return Deployment(nodes=frozenset(
+            list(n for n in self.nodes if n.hostname != node.hostname) +
+            [node]))
+
 
 @attributes(['internal_port', 'external_port'])
 class Port(object):
-    """
+        """
```
wallrj: Looks like a typo.

flocker/control/test/test_model.py
```diff
@@ -201,6 +243,53 @@ def test_applications(self):
                          sorted(list(node.applications) +
                                 list(another_node.applications)))
 
+    def test_update_node_new(self):
+        """
+        When donig ``update_node()``, if the given ``Node`` has hostname not
```
wallrj: Typo "donig"

flocker/control/test/test_model.py
```diff
@@ -201,6 +243,53 @@ def test_applications(self):
...
+        )
+        original = Deployment(nodes=frozenset([node]))
+        updated = original.update_node(another_node)
+        self.assertEqual((original, updated),
+                         (Deployment(nodes=frozenset([node])),
+                          Deployment(nodes=frozenset([node, another_node]))))
+
+    def test_update_node_replace(self):
+        """
+        When donig ``update_node()``, if the given ``Node`` has hostname in
```
wallrj: Typo "donig"

flocker/control/test/test_model.py
```diff
@@ -201,6 +243,53 @@ def test_applications(self):
...
 
+    def test_update_node_new(self):
+        """
+        When donig ``update_node()``, if the given ``Node`` has hostname not
+        in existing ``Deployment`` then just add new ``Node`` to new
+        ``Deployment``.
+        """
+        node = Node(
+            hostname=u"node1.example.com",
+            applications=frozenset({Application(name=u'mysql-clusterhq',
```
wallrj: @exarkun is keen to remove mysql examples from the test suite. Maybe replace this one:
 * https://github.com/ClusterHQ/flocker/pull/1101/files#r23867389

...although I note that we forgot to update the mysql reference in that PR before merging.

flocker/node/test/test_deploy.py
```diff
@@ -1573,9 +1573,13 @@ def test_volume_handoff(self):
         # application is running here.
         current = Deployment(nodes=frozenset([node, another_node]))
 
+        volume_service = create_volume_service(self)
+        self.successResultOf(volume_service.create(
+            volume_service.get(_to_volume_name(DATASET_ID))))
+
```
wallrj: I'm not suggesting changing it here, but I found this, and the other updated tests, difficult to read. It'd be nice to use variable names like `node1`, `node2`, `node1_volume_service`, `node1_deployer` to make it clear what situation we're trying to simulate.

flocker/node/test/test_deploy.py
```diff
@@ -2324,27 +2349,18 @@ def test_dataset_id_populated(self):
...
-                              docker_client=FakeDockerClient(),
+
+        api = P2PNodeDeployer(u"node", volume_service,
+                              docker_client=docker,
                               network=make_memory_network())
         result = api.calculate_necessary_state_changes(
             self.successResultOf(api.discover_local_state()), desired, actual)
 
         # Desired configuration was changed to set the correct dataset ID,
         # which is why a resize is happening:
```
wallrj: I don't understand this. I know it's not related to this branch, but could you add some further information about why updating dataset_id causes a ResizeDataset step?

wallrj:
Thanks @itamarst 

This all looks good to me.

I left a few inline comments above. Please answer or address those and then merge.

PR #1143, opened 2015-02-09
Use prompt:bash consistently in release process documentation.

jongiddy:


adamtheturtle:
Thanks Jon. For the `vagrant box list` block, add a comment why this was left unchanged. Also, make any follow up issues you think would be useful in terms of applying similar changes elsewhere in the documentation.

PR #1146, opened 2015-02-10
Get dataset configuration

exarkun:
https://clusterhq.atlassian.net/browse/FLOC-1387

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -74,6 +74,31 @@ class DatasetAPIUserV1(object):
         """
         return {u"flocker":  __version__}
 
+
+    @app.route("/datasets", methods=['GET'])
+    @user_documentation(
+        """
+        Get the cluster's dataset configuration.
+        """,
+        examples=[u"get configured datasets"],
+    )
+    @structured(
+        inputSchema={},
+        outputSchema={
+            '$ref': '/v1/endpoints.json#/definitions/datasets_array',
+        },
+        schema_store=SCHEMAS,
+    )
+    def get_dataset_configuration(self):
+        """
+        Get the configured datasets.
+
+        :return: A ``list`` of ``dict`` representing each of dataset
+            that is configured to exist anywhere on the cluster.
+        """
+        return list(datasets_from_deployment(self.persistence_service.get()))
+
+
     @app.route("/datasets", methods=['POST'])
     @user_documentation(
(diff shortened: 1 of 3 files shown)


flocker/control/httpapi.py
```diff
@@ -74,6 +74,29 @@ def version(self):
         """
         return {u"flocker":  __version__}
 
+    @app.route("/datasets", methods=['GET'])
```
itamarst: I would suggest this and the `POST` get changed to `/configuration/datasets` as you suggested at some point, so they are more symmetric with `/state/datasets` and it's immediately clear one is referring to *configuration*, not state.

(Might want to change this again when we have API that can actually tell you when a change has been applied successfully, but that may be a while...)

itamarst:
Additional things to fix:
1. Both this and the current state endpoint don't include the JSON schema in the generated documentation. I believe this is because our documentation generator doesn't handle arrays, and assumes output is always a JSON object/struct. File an issue separately for this, I guess.
2. Missing an acceptance test. This should be easy to write once FLOC-1397 is done. Up to you whether to do it in this branch or another, but we should require acceptance tests for each new API endpoint from here on out.

flocker/control/test/test_httpapi.py
```diff
@@ -451,10 +451,146 @@ def get_dataset_ids(deployment):
...
+        expected = [
+            api_dataset_from_dataset_and_node(
+                manifestation_a.dataset, self.NODE_A
+            ),
+            api_dataset_from_dataset_and_node(
+                manifestation_b.dataset, self.NODE_A
+            ),
+        ]
+        return self._dataset_test(deployment, expected)
+
```
itamarst: Maybe add unit test for dataset that has metadata, so we can see that the schema verification doesn't reject it.

itamarst:
Please accept if you're doing acceptance test as separate issue, or reject if you're doing acceptance test in this branch.

exarkun:
> 2. Missing an acceptance test. This should be easy to write once FLOC-1397 is done. Up to you whether to do it in this branch or another, but we should require acceptance tests for each new API endpoint from here on out.

I think this will probably make sense once we have a more sophisticated configuration storage system (ZooKeeper or what have you).  For now I think the acceptance test would basically exercise exactly the same code path as the integration test so I'm going to skip it.

PR #1147, opened 2015-02-10
[FLOC-936] Sketch of implementation of Eliot logging

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-936

### DIFF
--- a/flocker/common/script.py
+++ b/flocker/common/script.py
@@ -5,12 +5,13 @@
 import sys
 import os
 
+from eliot import MessageType, fields, Logger
+
 from twisted.internet import task
 from twisted.internet.defer import Deferred, maybeDeferred
 from twisted.python import usage
 from twisted.python.filepath import FilePath
-from twisted.python.log import (
-    addObserver, removeObserver, FileLogObserver, msg)
+from twisted.python.log import addObserver, removeObserver, textFromEventDict
 
 from zope.interface import Interface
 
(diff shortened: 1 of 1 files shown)


flocker/common/script.py
```diff
@@ -139,9 +158,10 @@ def main(self):
         self._react(self.script.main, (options,), _reactor=self._reactor)
 
         # Not strictly necessary, but nice cleanup for tests:
-        if observer is not None:
-            removeObserver(observer)
-            log_file.close()
+        # XXX FLOC-936
+        # 0. If self.logging is False, don't do anything. Otherwise:
+        # 1. Remove EliotObserver
+        # 2. Stop ThreadedLogWriter
```
exarkun: :question: If you make some context managers I guess you could use a `with` block here and have slightly neater cleanup.

flocker/common/script.py
```diff
@@ -80,16 +106,18 @@ class FlockerScriptRunner(object):
     """
     _react = staticmethod(task.react)
 
-    # Location where logs will be written, overrideable by tests:
-    log_directory = FilePath(b"/var/log/flocker/")
-
-    def __init__(self, script, options, reactor=None, sys_module=None):
+    def __init__(self, script, options, logging=True,
+                 reactor=None, sys_module=None):
```
exarkun: We should talk about API compatibility (for our own benefit) at some point.

flocker/common/script.py
```diff
@@ -118,19 +146,10 @@ def _parse_options(self, arguments):
...
-                                os.getpid()))
-            log_file = log_path.open("a")
-            observer = FileLogObserver(log_file).emit
-            addObserver(observer)
-            msg("Arguments: %s" % (self.sys_module.argv,))
-        except (OSError, IOError):
-            pass
+        # XXX FLOC-936
+        # 0. If self.logging is False, don't do anything. Otherwise:
+        # 1. Create ThreadedLogWriter writing to stdout
```
exarkun: I wonder if there'd be any value (in general, most likely not for this issue) in having a non-threaded stdout logger using `twisted.internet.stdio`.

exarkun:
Thanks.  Seems reasonable.  Please proceed.

PR #1150, opened 2015-02-11
[FLOC-1275] Transfer Eliot context with AMP commands

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1275


wallrj:
Thanks @itamarst 

This looks good to me.

The only thing I wondered is whether it might be possible to define a new `EliotAction` AMP argument type which took care of calling `serialize_task_id()`.

Please proceed.

wallrj:
@itamarst I finished testing the ``with_eliot_context`` decorator. Now need your guidance on how to update the existing tests and how / where to test the messages logged by remote AMP commands. 

See `trial flocker.control.test.test_protocol.AgentClientTests.test_cluster_updated` in this branch for an unfinished attempt to update one of the existing tests.

wallrj:
Ok. I think this is ready for a review.

itamarst:
Buildbot was red last I checked.

flocker/control/_protocol.py
```diff
@@ -90,10 +98,30 @@ class NodeStateCommand(Command):
...
-    arguments = [('node_state', NodeStateArgument())]
+    arguments = [('node_state', NodeStateArgument()),
+                 ('eliot_context', Unicode())]
     response = []
 
 
+def with_eliot_context(function):
+    """
+    Decorator for responders that accept an ``eliot_context`` argument
+    that deserializes the given Eliot context and runs the wrapped
```
itamarst: first `that` in this line is ambiguous - does it refer to the decorator or the argument?

flocker/control/_protocol.py
```diff
@@ -253,7 +313,19 @@ def __init__(self, agent):
         CommandLocator.__init__(self)
         self.agent = agent
 
+    @property
```
itamarst: Can't you just override two different loggers in the test?

flocker/control/test/test_protocol.py
```diff
@@ -163,9 +175,12 @@ def test_connection_made_send_cluster_status(self):
         When a connection is made the cluster status is sent to the new client.
         """
         sent = []
-        self.patch(self.protocol, "callRemote",
-                   lambda *args, **kwargs: sent.append((args, kwargs))
-                   or succeed(None))
+        self.patch(
```
itamarst: This and latter variants are all the same, just with different lists. Maybe the utility method should be something that does patching? `self.patch_callremote(sent)`.

flocker/control/test/test_protocol.py
```diff
@@ -290,6 +314,27 @@ def test_stop_service_connections(self):
              [c.transport.disconnecting for c in connections]),
             ([False] * 3, [True] * 3))
 
+    def assertDictEqual(self, expected, actual):
```
itamarst: You can just use built-in version of this as provided by stdlib `TestCase`, no?

flocker/control/test/test_protocol.py
```diff
@@ -290,6 +314,27 @@ def test_stop_service_connections(self):
...
+            actual_value = actual.pop(expected_key)
+            self.assertEqual(
+                expected_value, actual_value,
+                'Non-equal dictionary value. '
+                'Key: {!r} Expected: {!r} Actual: {!r}'.format(
+                    expected_key, expected_value, actual_value)
+            )
+        self.assertEqual({}, actual)
+
+    def assertArgsEqual(self, expected, actual):
```
itamarst: Docstring.

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+            return server.handle_request(
+                eliot_context=action.serialize_task_id(),
+                **kwargs
+            )
+
+
+class ServerProcess(object):
+    logger = None
+
+    @with_eliot_context
```
itamarst: My guess is if you remove this line and add a eliot_context keyword argument that is ignored, these tests will still pass, because HANDLE_REQUEST is being run in the context of SEND_REQUEST irrespective of the decorator. If I'm correct that means these tests aren't really enough to prove the code works.

itamarst: Specifically the `assert_child_action` assertion doesn't prove anything.

itamarst: The solution is to move `server.handle_request` out of the `with SEND_REQUEST` block:

```
with SEND_REQUEST(self.logger) as action:
    task_id = action.serialize_task_id()
return server.handle_request(eliot_context=task_id, **kwargs)
```

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+    def assert_child_action(self, logger):
+        """
+        The Client sets the logging context which means that Server Actions
+        appear as children of the Client Action.
+        """
+        # There should only be one...
+        (client_action,) = LoggedAction.of_type(logger.messages, SEND_REQUEST)
+        (server_action,) = LoggedAction.of_type(
+            logger.messages, HANDLE_REQUEST)
+        for child_action in client_action.descendants():
```
itamarst: `self.assertIn(child_action, client_action.descendants())`?

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+        """
+        ``cluster_updated`` is decorated using ``with_eliot_context`` and
+        therefore requires an eliot_context argument. The supplied
+        eliot_context is used as the context for messages logged in that
+        method.
+        """
+        fake_agent = FakeAgent()
+        self.patch(fake_agent, 'logger', logger)
+        locator = _AgentLocator(agent=fake_agent)
+        with SEND_REQUEST(logger) as action:
```
itamarst: Same issue as above: you need to move `cluster_updated` call out of the action context.

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+        fake_control_amp_service = build_control_amp_service(self)
+        self.patch(fake_control_amp_service, 'logger', logger)
+        self.patch(
+            fake_control_amp_service, 'node_changed', lambda node_state: None)
+
+        locator = ControlServiceLocator(
+            control_amp_service=fake_control_amp_service
+        )
+
+        with SEND_REQUEST(logger) as action:
```
itamarst: Same problem again.

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+        action and a LOG_SEND_TO_AGENT action for the remote calls to each of
+        its connections.
+        """
+        control_amp_service = build_control_amp_service(self)
+        self.patch(control_amp_service, 'logger', logger)
+
+        connection_protocol = ControlAMP(control_amp_service)
+        connection_protocol.makeConnection(StringTransport())
+
+        # XXX Maybe supply multiple connections and check that each is logged?
```
itamarst: Sure?

flocker/control/test/test_protocol.py
```diff
@@ -451,3 +518,295 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+            LOG_SEND_CLUSTER_STATE,
+            succeeded=True,
+            startFields={
+                "configuration": (
+                    control_amp_service.configuration_service.get()
+                ),
+                "state": control_amp_service.cluster_state.as_deployment()
+            }
+        )
+        # XXX Check that there's an action for each connection.
```
itamarst: Looks already done.

flocker/node/_loop.py
```diff
@@ -261,7 +274,10 @@ def output_CONVERGE(self, context):
         d = self.deployer.discover_local_state()
 
         def got_local_state(local_state):
-            self.client.callRemote(NodeStateCommand, node_state=local_state)
+            with LOG_SEND_TO_CONTROL_SERVICE(
```
itamarst: 1. LOG_SEND_TO_CONTROL_SERVICE is a MessageType, should be ActionType.
2. This passes in field that is not defined there.

itamarst:
Thanks! Please address the above and resubmit.

dwgebler:
Thanks @itamarst I'm putting this up for review again; this is not suitable for merging yet, actually, but I could use a couple more pointers on some of the Eliot aspects. It seems the `FakeAMPClient` in `flocker.testtools` is going to need to a method (possibly just a `__repr__` that returns a fake, knowable IP address of the control service for the convergence loop to use as its fields in the logging (see the XXX comment in `flocker.node._loop.LOG_SEND_TO_CONTROL_SERVICE`) and I have some broken tests in `test_loop` that I'm still working on, but if you could review the changes I've made today and highlight if anything's not on the right track, that'd be much appreciated. Going to go back to pairing with @wallrj on this later.

itamarst:
Having `repr` as a serializer for connections isn't great. What we really want is some sort of identity ("zfs agent for node 123-545-679"), but we don't have infrastructure for that yet. So maybe use host/port (`address = client.transport.getPeer()`) and you can add a fake transport to the fake AMP client using a `twisted.test.proto_helpers.StringTransport`.

flocker/control/test/test_protocol.py
```diff
@@ -137,7 +141,41 @@ def build_control_amp_service(test):
...
+    def patch_call_remote(self, capture_list, protocol=None):
+        """
+        Patch the callRemote method for this test case's protocol.
+
+        :param capture_list: A `list` to which results will be added.
+
+        :param protocol: Either `None` to default to self.protocol, or
+            a ``ControlAMP`` instance.
+        """
+        if protocol is None:
```
itamarst: :question: The tests are a lot more readable as a result of this, but I would suggest omitting this default and explicitly passing in `self.protocol`; too much succinctness makes things less understandable.

flocker/node/_loop.py
```diff
@@ -231,6 +233,21 @@ class ConvergenceLoopOutputs(Names):
...
+    lambda client: repr(client),
+    "The AMP connection to control service")
+
+LOG_SEND_TO_CONTROL_SERVICE = ActionType(
+    u"flocker:agent:send_to_control_service",
+    # XXX fields would just be the address of the control service, we can
+    # rely on the higher-level logged action of the convergence loop to
+    # log the local state (in FLOC-1379).
+    # XXX what's the success field here?
+    [_FIELD_CONNECTION], [],
```
itamarst: Could be address of control service? Or not if that's easier, for now, since in short term there's only one control service address anyway. Oh, another useful thing would be the unique identifier for the convergence loop FSM (`fsm.identifier()`). But you can feel free to omit this in this branch and just have comment referring to followup.

So, some stuff in there would be useful, but not necessary to complete this branch. So just omit it.

itamarst:
After further thought, I suggest not bothering adding any fields to that message at all. So forget about connection address, we'll add that someday in the future.

In general looking better.

dwgebler:
I *think*, barring the things @itamarst said to ignore, this is now ready for review again.

flocker/control/_protocol.py
```diff
@@ -253,7 +315,19 @@ def __init__(self, agent):
...
+    def logger(self):
+        """
+        ``with_eliot_context`` assumes that the decorated method's bound
+        instance has a logger, which it could, but for testing, it needs to use
+        the same logger as the supplied agent.
+
+        XXX This seems ugly.
+        """
+        return self.agent.logger
```
exarkun: `IAgent` doesn't have a `logger` attribute.  I see how this can work in the test suite, since `FakeAgent` has been given this attribute - but I don't see where it will come from in production when using an actual agent from Twisted.

flocker/control/_protocol.py
```diff
@@ -81,7 +88,8 @@ class ClusterStatusCommand(Command):
     in the convergence agent during startup.
     """
     arguments = [('configuration', DeploymentArgument()),
-                 ('state', DeploymentArgument())]
+                 ('state', DeploymentArgument()),
+                 ('eliot_context', Unicode())]
```
exarkun: A custom `IArgument` implementation for eliot actions probably makes sense.  This would let application code pass an action as an argument to the `callRemote` call and let the `IArgument` figure out how to serialize and unserialize the object.  This will be increasingly convenient as there are more AMP calls (with just two it's not so bad but we're going to have more, probably for result reporting and definitely for node-to-node communication).

exarkun: This should be a simple refactoring.  A `Unicode` subclass with `toString` and `fromString` overridden to call the relevant eliot APIs - `serialize_task_id` and `continue_task`.  The eliot `Logger` object is maybe a little tricky...   Possibly this will need `toStringProto` and `fromStringProto` instead, in order to have access to the `ControlAMP` instance (which should be sufficient to reach the right `Logger` instance).  Relying on the protocol instance to provide this information is not really ideal - but that's what this version of the code does too, so at least it's not really worse.

Another option would be to have a module-scope logger object but maybe that's worse (since it will be shared between more things and since it's global state that the tests will have to replace and then restore).

flocker/control/_protocol.py
```diff
@@ -90,10 +98,30 @@ class NodeStateCommand(Command):
     Used by a convergence agent to update the control service about the
     status of a particular node.
     """
-    arguments = [('node_state', NodeStateArgument())]
+    arguments = [('node_state', NodeStateArgument()),
+                 ('eliot_context', Unicode())]
```
exarkun: Same here, of course.

flocker/control/_protocol.py
```diff
@@ -90,10 +98,30 @@ class NodeStateCommand(Command):
...
     response = []
 
 
+def with_eliot_context(function):
+    """
+    Decorator for responders who accept an ``eliot_context`` argument
+    that deserializes the given Eliot context and runs the wrapped
+    function using that as the Eliot context.
+
+    No support for returned ``Deferreds`` at this point.
```
exarkun: I suggest making this a noisy failed.  If a `Deferred` is returned, raise an exception.

flocker/control/_protocol.py
exarkun: Sorry.  That's all wrong, of course.  Disregard it.  I mistakenly thought this was a `twisted.web.client.Agent` but it's an `IConvergenceAgent`.

exarkun: The correct comment to make here, then, is that the `logger` attribute isn't part of `IConvergenceAgent`.  And `FakeAgent` is an unverified fake so nothing demonstrates that it's compatible with the useful implementation of `IConvergenceAgent` (`ControlAMPService`).

That's too bad.  And more generally, `FakeAgent` is an example of not-very-good testing methodology that we should avoid.  It's basically a minimally functional, completely custom mock.  So it has the correctness drawbacks of mocks but without any of the convenience of mocks.  Sad.

The interface should be fixed and a test added that can be shared between `IConvergenceAgent` implementations demonstrating that (I suppose) they have a `logger` attribute that is an instance of `Logger`.

itamarst: Or just have a single module-level logger? I'm still not convinced (for Eliot in general, and here specifically) that one ever needs more than one.

flocker/control/_protocol.py
exarkun: It seems like a bit of a headache to have a million different `Logger` instances and have to keep track of which log messages go to which ones.  Particularly in this case, where `_AgentLocator` is private.  Forcing the tests to know that this private helper's logger attribute is the one where some log events will go seems like an unfortunately leaky abstraction.

It seems like it kinda makes sense to say "Log events related to this one `IConvergenceAgent` provider you've got will all go to its `logger` attribute" and then make all the things hanging off of that object, including this AMP code, use the object specified by that attribute.  In which case this isn't really ugly, it's just how you have to do things with eliot.

Maybe we should think about whether we like how eliot encourages you to have tons of `Logger` instances all of the place, though.

flocker/control/test/test_protocol.py
```diff
@@ -137,7 +141,39 @@ def build_control_amp_service(test):
...
+            kwargs.pop('eliot_context')
+            capture.append((args, kwargs))
+            return succeed(None)
+
+        self.patch(
+            protocol,
+            "callRemote",
+            lambda *args, **kwargs: capture_call_remote(
+                capture_list, *args, **kwargs)
+        )
```
exarkun: This ... is just refactoring of existing code, I guess.  This (the pre-existing code or the refactored code) is really not how we want to be writing tests though.  This is a tedious, error prone way to test.  It's another semi-re-implementation of the mock testing pattern.  The exact arguments passed to `callRemote` aren't the important thing.  The behavior achieved is.  This level of testing only exercises the former, though.

This is a pre-existing issue so it doesn't block this branch, though.

exarkun:
There's a couple comments, at least.  Unfortunately not able to finish this right now.  I'll pick it up again ASAP but sending it back in case you get some time to address the comments I did make.

flocker/control/_protocol.py
itamarst: I kinda feel the whole "let's patch a `Logger`" idiom was a mistake and that really the default in tests should just be "capture all log messages generated by the test".

itamarst:
I ripped out a lot of code based on @exarkun suggestions, and then ripped out a lot of fragile fiddly unit tests that sadly didn't prove much evidence of actual correctness.

1. Acceptance tests should validate that this change didn't break anything.
2. https://clusterhq.atlassian.net/browse/FLOC-1507 should eventually test cross-process logging properties the correct way.
3. Until that is done: it probably works, and if not we're no worse off than before. And we're wasting too much time on this.

flocker/control/test/test_protocol.py
```diff
@@ -452,3 +517,122 @@ class FakeAgentInterfaceTests(iconvergence_agent_tests_factory(
...
+    """
+    Tests for ``NodeStateCommand``.
+    """
+    def test_command_arguments(self):
+        """
+        ``NodeStateCommand`` requires the following arguments.
+        """
+        self.assertEqual(
+            sorted(['node_state', 'eliot_context']),
+            sorted(v[0] for v in NodeStateCommand.arguments))
```
exarkun: This is what `TestCase.assertItemsEqual` is for I think

flocker/control/_protocol.py
```diff
@@ -137,12 +163,32 @@ def connectionLost(self, reason):
         self.control_amp_service.disconnected(self)
 
 
+DEPLOYMENT_CONFIG = Field(u"configuration", repr, u"The cluster configuration")
+CLUSTER_STATE = Field(u"state", repr, u"The cluster state.")
+
+LOG_SEND_CLUSTER_STATE = ActionType(
+    "flocker:controlservice:send_cluster_state",
```
exarkun: Again I find myself wondering how these strings are constructed and how anyone will make sense of them.

:question:

flocker/control/_protocol.py
```diff
@@ -137,12 +163,32 @@ def connectionLost(self, reason):
         self.control_amp_service.disconnected(self)
 
 
+DEPLOYMENT_CONFIG = Field(u"configuration", repr, u"The cluster configuration")
+CLUSTER_STATE = Field(u"state", repr, u"The cluster state.")
```
exarkun: One description with a full stop, the other without.  These **aren't** sentences, so...

flocker/control/_protocol.py
```diff
@@ -137,12 +163,32 @@ def connectionLost(self, reason):
...
+    [],
+    "Sending the configuration and state of the cluster.")
+
+AGENT = Field(u"agent", repr, u"The agent we're sending to")
+
+LOG_SEND_TO_AGENT = ActionType(
+    "flocker:controlservice:send_state_to_agent",
+    [AGENT],
+    [],
+    "Send the configuration and state of the cluster to a specific agent.")
```
exarkun: The description for this action is in a different ... mood?  Or something.  As compared to the description for `LOG_SEND_CLUSTER_STATE` above.

exarkun:
Thanks.  Please address remaining minor inline comments and then, if buildbot is green, merge.

PR #1151, opened 2015-02-11
[FLOC-1371] Document the architecture

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1371

Note this talks about datasets, which aren't yet documented anywhere - this is covered by separate issue, https://clusterhq.atlassian.net/browse/FLOC-1165

### DIFF
--- a/docs/advanced/architecture.rst
+++ b/docs/advanced/architecture.rst
@@ -54,6 +54,7 @@ For example, imagine the control service notifies the agent on node A that node 
 
 #. The agent discovers there are no datasets on the node.
 #. The agent tells the control service that there exist no datasets on the node.
+   The agent always reports its latest local state to the control service to ensure it is up-to-date, even if it may change in near future.
 #. The agent decides it needs to create dataset D, and does so.
 #. The loop begins again - the agent discovers that dataset D exists on the node.
 #. The agent tells the control service that dataset D exists on the node.
--- a/docs/advanced/architecture.rst
+++ b/docs/advanced/architecture.rst
@@ -1,0 +1,56 @@
+============================
+Flocker Cluster Architecture
+============================
+
+This document describes the Flocker cluster's architecture.
+More accurately, it describes the architecture Flocker is moving towards, a transition that is still in progress.
+
+The Flocker cluster is composed of two sets of services:
+
+1. **The control service** that you can interact with using a :ref:`HTTP API<api>` to modify the desired configuration of the cluster.
+2. **Convergence agents** in charge of modifying the cluster state to match the desired configuration.
+   For example, if you're using Flocker's ZFS storage backend you will have ZFS-specific agents running on each node in the cluster.
+
+
+Control service
+===============
+
+The control service is the integration point between the desires of whoever is configuring the cluster, e.g. a human administrator or an orchestration framework, and the convergence agents that attempt to execute these desires.
+In particular it controls the configuration of the cluster.
+
+The service consists of three parts:
+
+* A data storage system stores the configuration of the system.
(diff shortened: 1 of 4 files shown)


docs/advanced/architecture.rst
```diff
@@ -0,0 +1,56 @@
...
+
+1. **The control service** that you can interact with using a :ref:`HTTP API<api>` to modify the desired configuration of the cluster.
+2. **Convergence agents** in charge of modifying the cluster state to match the desired configuration.
+   For example, if you're using Flocker's ZFS storage backend you will have ZFS-specific agents running on each node in the cluster.
+
+
+Control service
+===============
+
+The control service is the integration point between the desires of whoever is configuring the cluster, e.g. a human administrator or an orchestration framework, and the convergence agents that attempt to execute these desires.
```
wallrj: Not sure about "desires". Maybe instead something like:

> The control service is the integration point: 
> * at which the configuration changes, requested by e.g. a human administrator or an orchestration framework, are merged and stored, and 
> * from which convergence agents receive configuration change notifications before enacting those changes on each node.

docs/advanced/architecture.rst
```diff
@@ -0,0 +1,56 @@
...
+All three of the above are encapsulated in a single server, for the moment limited to running on a single machine.
+
+
+Convergence agents
+==================
+
+Convergence agents ensure that the state of the cluster eventually converges with the configuration.
+They control the actual system state but are not be able to modify the configuration.
+
+Convergence agents may be in charge of state on a specific node, e.g. a ZFS agent may be in charge of ZFS datasets.
```
wallrj: I hadn't understood, until @exarkun explained it to me, that there may be multiple agents running on a node, each responsible for managing a particular aspect of the overall state. Eg `flocker-docker-agent`, if we continue to develop our own orchestration framework.

docs/advanced/architecture.rst
```diff
@@ -0,0 +1,56 @@
...
+Or they may be in charge of some cluster-wide state.
+Each convergence agent runs a loop to converge the state is in charge of with the desired configuration.
```
wallrj: Typo "state *that it* is in charge of"

docs/advanced/architecture.rst
```diff
@@ -0,0 +1,56 @@
...
+#. Checks the local state it is in charge of (e.g. by listing local ZFS filesystems).
+#. Notifies the control service of the local state.
+#. Calculates the actions necessary to make local state match desired configuration.
+#. Executes these actions.
+#. Starts the loop again.
+
+For example, imagine the control service notifies the agent on node A that node A should have dataset D, and that as far as it knows no dataset D exists in the cluster.
+
+#. The agent discovers there are no datasets on the node.
+#. The agent tells the control service that there exist no datasets on the node.
```
wallrj: I don't understand why agent needs to report to the control service at this point.
I imagine that the loop would already be paused here, awaiting configuration changes.

wallrj:
Looks good:
 * I spotted a few typos, fix those.
 * I left a few comments, but feel free to ignore them.
 * The `docs` builder reports a spelling error, fix that.

Then merge. (This is up for design review, but I guess we can skip that step here)

itamarst:
`effect` is not being installed. When I do a manual install (`pip install effect=0.1a13`) it does install however. I suspect a pip bug/feature involving non-release packages being skipped? In which case it'll go away at some point when `effect` has a less alpha release. Not a blocker, anyway, since we're not using `effect` in product itself.

itamarst:
The release instructions don't mention need for `pip install --edit .[release]`. This is not a blocker, and should become unnecessary if buildbot can run acceptance tests.

PR #1152, opened 2015-02-12
Add whitespace into "Managing Volumes" section of docs/introduction.rst

dsalisbury:
Sphinx currently interprets it as continuing from the previous line ("Data Model") without the line break, and hence looks a tad ugly.

Link to current published docs for reference: https://docs.clusterhq.com/en/0.3.2/introduction.html#managing-volumes

### DIFF
--- a/docs/introduction.rst
+++ b/docs/introduction.rst
@@ -116,6 +116,7 @@ Managing Volumes
   * Association is done based on container names.
 
 * Data model
+
   * Volumes are owned by a specific node.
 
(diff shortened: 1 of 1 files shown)


exarkun:
Thanks!  Great catch.

itamarst:
@dsalisbury If you send an email to thankyou@clusterhq.com with your t-shirt size, mailing address and a phone number (to be used only for filling out the shipping form) we'll be happy to send you a t-shirt as a thank you.

PR #1154, opened 2015-02-13
This won't be necessary when buildbot runs all the acceptance tests auto...

adamtheturtle:
...matically.

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -230,55 +230,6 @@ This review step is to ensure that all acceptance tests pass on the release bran
 
       export VERSION=0.1.2
 
-#. Do the acceptance tests:
-
-   - Add the tutorial vagrant box that BuildBot has created from the release branch.
-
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Remove `.. _PhantomJS: http://phantomjs.org/download.html` below these changes. Resolve merge conflicts and merge when Buildbot is green and the blocking issue is done.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -230,55 +230,6 @@ This review step is to ensure that all acceptance tests pass on the release bran
 
       export VERSION=0.1.2
```
adamtheturtle: You don't really need this exported in this section, but then again it is useful(ish) for the link checking to know what $VERSION is meant to mean.

adamtheturtle:
The "Post-Release Review Process" has a step "Remove the Vagrant box which was added as part of Pre-tag Review Process:". This should be removed.

adamtheturtle:
Merging as the docs builder passes and that is the only change.

PR #1158, opened 2015-02-16
[FLOC-1405] Modify documented command to work with old and new setuptools.

jongiddy:
Newer versions of setuptools PEP440 normalise package version numbers.

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -343,10 +343,13 @@ Release
 
    .. note:: The following instructions use `virtualenvwrapper`_ but you can use `virtualenv`_ directly if you prefer.
 
+   .. note:: The final command ensures that setuptools is a version that does not normalize version numbers according to PEP440.
+
    .. prompt:: bash $
 
       mkvirtualenv flocker-release-${VERSION}
       pip install --editable .[release]
+      pip install setuptools==3.6
 
 #. Tag the version being released:
 
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -373,7 +373,7 @@ Release
       python setup.py sdist bdist_wheel
       gsutil cp -a public-read \
           "dist/Flocker-${VERSION}.tar.gz" \
-          "dist/Flocker-$(python setup.py --version)-py2-none-any.whl" \
+          "dist/Flocker-${VERSION}-py2-none-any.whl" \
           gs://archive.clusterhq.com/downloads/flocker/
 
 #. Build RPM packages and upload them to ``archive.clusterhq.com``
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -373,7 +373,7 @@ Release
       python setup.py sdist bdist_wheel
       gsutil cp -a public-read \
           "dist/Flocker-${VERSION}.tar.gz" \
-          "dist/Flocker-${VERSION}-py2-none-any.whl" \
+          "dist/Flocker-$(python setup.py --version)-py2-none-any.whl" \
           gs://archive.clusterhq.com/downloads/flocker/
 
 #. Build RPM packages and upload them to ``archive.clusterhq.com``


adamtheturtle:
This works to output a `.whl` file with an extra `.` in it. However that `whl` file is installed by anyone following either of our getting started guides. See the `flocker-tutorial/bin/pip install` command in http://doc-dev.clusterhq.com/indepth/installation.html#linux for example. This uses a template with `|latest-installable|` on it. I guess you could also make `|latest-whl|` but that seems worse than somehow keeping the `whl` like it used to be, and waiting until we update our versioning system.

adamtheturtle:
Possible fixes include:
* `pip install setuptools=={some_working_version}`
* Rename the file if it has the extra `.`.

tomprince:
The other option is to specify the version and use `--find-links Flocker==x.y.z`.

wallrj:
@tomprince I tried but it fails because there's no directory listing:

```
(test3)[vagrant@localhost vagrant]$ pip install --find-links=https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker Flocker==0.3.2
Downloading/unpacking Flocker==0.3.2
  Real name of requirement Flocker is flocker
  Could not find a version that satisfies the requirement Flocker==0.3.2 (from versions: 0.0.0-696-ged513f7)
Cleaning up...
No distributions matching the version for Flocker==0.3.2
Storing debug log for failure in /home/vagrant/.pip/pip.log
```


```
(test3)[vagrant@localhost vagrant]$ head -n20 /home/vagrant/.pip/pip.log 
------------------------------------------------------------
/home/vagrant/.virtualenvs/test3/bin/pip run on Thu Feb 19 02:51:04 2015
Downloading/unpacking Flocker==0.3.2
  Getting page https://pypi.python.org/simple/Flocker/
  Could not fetch URL https://pypi.python.org/simple/Flocker/: 404 Client Error: Not Found
  Will skip URL https://pypi.python.org/simple/Flocker/ when looking for download links for Flocker==0.3.2
  Getting page https://pypi.python.org/simple/
  Real name of requirement Flocker is flocker
  URLs to search for versions for Flocker==0.3.2:
  * https://pypi.python.org/simple/Flocker/0.3.2
  * https://pypi.python.org/simple/flocker/
  * https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker
  Skipping link file:///var/cache/wheelhouse (from -f); not a file
  Skipping link https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker (from -f); not a file
  Getting page https://pypi.python.org/simple/Flocker/0.3.2
  Could not fetch URL https://pypi.python.org/simple/Flocker/0.3.2: 404 Client Error: Not Found
  Will skip URL https://pypi.python.org/simple/Flocker/0.3.2 when looking for download links for Flocker==0.3.2
  Getting page https://pypi.python.org/simple/flocker/
  Could not fetch URL https://pypi.python.org/simple/flocker/: 404 Client Error: Not Found
  Will skip URL https://pypi.python.org/simple/flocker/ when looking for download links for Flocker==0.3.2

```

 * https://pip.pypa.io/en/latest/reference/pip_wheel.html#cmdoption-f

Perhaps it would be nice to fix that though. And maintain our own wheel (house) index from which our omnibus packages would pull their requirements.

I guess that's related to FLOC-1046 "Omnibus package builder should only install verified dependencies"
 * https://clusterhq.atlassian.net/browse/FLOC-1046

wallrj:
* Created new vagrant-dev machine
* Checked installed version of setuptools

```
(test3)[vagrant@localhost vagrant]$ pip show setuptools
---
Name: setuptools
Version: 3.6

```

* Installed `python-virtualenvwrapper`
* Verified that it uses that version of setuptools
* Verified that the `python setup.py --version` reports non-normalised version number
* Upgraded setuptools inside the virtualenv, using pip
* verified that `python setup.py --version` now reports the normalised version
* verified that `pip install setuptools==3.6` was able to downgrade setuptools

So as far as I can tell, these instructions will work.

Is there a followup issue for switching to a pep440 style tagging scheme and making the necessary changes to our docs and release tools?

I remember @tomprince saying that buildbot will still be using the latest (version normalising) setuptools. And I guess that means that we need to change some of our packaging scripts...or has that already been done? I vaguely recall reviewing a change where we get the version by importing `flocker.__version__` instead of relying on `python setup.py --version` so perhaps it is.

Anyway, please merge unless any of my comments ring any alarm bells.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -343,10 +343,13 @@ Release
 
    .. note:: The following instructions use `virtualenvwrapper`_ but you can use `virtualenv`_ directly if you prefer.
 
+   .. note:: The final command ensures that setuptools is a version that does not normalize version numbers according to PEP440.
```
adamtheturtle: :question: link to https://clusterhq.atlassian.net/browse/FLOC-1331

adamtheturtle:
> Is there a followup issue for switching to a pep440 style tagging scheme and making the necessary changes to our docs and release tools?

Yes https://clusterhq.atlassian.net/browse/FLOC-1331.

docs/gettinginvolved/infrastructure/release-process.rst
jongiddy: I think I'll leave that link out. Knowing that we might one day move to PEP440 versioning is not necessary to understand the note, which is itself merely a distraction to getting on with releasing.

PR #1161, opened 2015-02-17
Reject pre tag floc 1377

adamtheturtle:
Fixes FLOC-1377


jongiddy:
Fine, but the line: "Else, add any problems ..." sounds a bit awkward.  I would rephrase as:

Otherwise, add comments to the Pull Request for any problems, and add that they must be resolved before repeating the pre-tag review process.

PR #1167, opened 2015-02-19
[FLOC-1357] Design for convergence loop error logging

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1357

Stacked on #1150

Use `git diff amp-logging-FLOC-1275` to see only design changes.


itamarst:
One thing that's missing right now is error handling (i.e. log and continue) for not just discovery but e.g. applying the calculated changes.

Basic idea is how it would work, though, yes.

PR #1169, opened 2015-02-20
OS X test fixes

adamtheturtle:
These are changes to Flocker to get the test suite to pass on OS X. We only care about the CLI working so any skips which are not in the CLI are fine.

See tests passing / skipping on OS X at http://build.staging.clusterhq.com/builders/flocker-osx-10.10/builds/3.


flocker/node/testtools.py
```diff
@@ -4,14 +4,26 @@
...
 
-# This is terible (https://clusterhq.atlassian.net/browse/FLOC-85):
-if_docker_configured = skipIf(Popen([b"docker", b"version"]).wait(),
-                              "Docker must be installed and running.")
+
+def docker_running():
+    """
+    :return: ``True`` if Docker is installed and running, else ``False``.
+
+    XXX This is terrible (https://clusterhq.atlassian.net/browse/FLOC-85)
```
tomprince: It would probably be better to use something like https://github.com/ClusterHQ/flocker/commit/6c0dcb8c980c35f7cfb4989b1dbd82fd14a6c778 since we probably *don't* want to use boot2docker in the tests. (Maybe along with https://github.com/ClusterHQ/flocker/blob/master/flocker/node/_docker.py#L271).

setup.py
```diff
@@ -35,6 +36,50 @@
...
+    "selenium==2.44.0",
+
+    # The cloud acceptance test runner needs these
+    "fabric==1.10.0",
+    "apache-libcloud==0.16.0",
+]
+
+# The test suite uses network namespaces
+# nomenclature cannot be installed on OS X
+if platform.system() != 'Darwin':
```
tomprince: Better to check for `Linux`, since nomenclature can only be installed there.

flocker/cli/functional/test_sshconfig.py
```diff
@@ -64,7 +64,8 @@ def test_connection_failed(self):
 
         exc = self.assertRaises(CalledProcessError,
                                 self.configure_ssh, b"127.0.0.1", port)
-        self.assertIn(b"refused", exc.output)
+        self.assertTrue(b"refused" in exc.output or "Operation timed out" in
+                        exc.output)
```
tomprince: I guess a bound socket that isn't listening behaves differently on OSX. I guess there should be a comment explaining the possible behaviors.

tomprince:
Please address the above comments. I'm rejecting it to review the docker skip changes.

adamtheturtle:
I have addressed the review comments, making minor changes to the linked previously-existing code. If this is suitable, check if https://clusterhq.atlassian.net/browse/FLOC-85 should be marked fixed too.

fabfile.py
```diff
@@ -0,0 +1,8 @@
+from fabric.api import task, run, local
+
+@task
+def brew():
```
tomprince: This file doesn't belong in the root like this.

flocker/cli/functional/test_sshconfig.py
```diff
@@ -64,7 +64,13 @@ def test_connection_failed(self):
 
         exc = self.assertRaises(CalledProcessError,
                                 self.configure_ssh, b"127.0.0.1", port)
-        self.assertIn(b"refused", exc.output)
+        # There are different error messages on different platforms.
```
tomprince: It'd be nice if there was some comment on *why* there are different error messages.

flocker/node/testtools.py
```diff
@@ -4,14 +4,43 @@
...
+import socket
+from unittest import skipUnless
 
+from ._docker import BASE_DOCKER_API_URL
 from ..testtools import loop_until
 
-# This is terible (https://clusterhq.atlassian.net/browse/FLOC-85):
-if_docker_configured = skipIf(Popen([b"docker", b"version"]).wait(),
-                              "Docker must be installed and running.")
+SOCKET_PATH = BASE_DOCKER_API_URL.split(':/')[-1]
```
tomprince: This should probably have `DOCKER` in the name.

start_homebrew_machine
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
tomprince: Same with these scripts.

flocker/node/testtools.py
```diff
@@ -4,14 +4,43 @@
...
+        if e.errno == errno.EACCES:
+            return False
+        if e.errno == errno.ENOENT:
+            # Docker is not installed
+            return False
+        raise
+    else:
+        return True
+
+if_docker_configured = skipUnless(
```
tomprince: This now duplicates https://github.com/ClusterHQ/flocker/commit/6c0dcb8c980c35f7cfb4989b1dbd82fd14a6c778, so we should get rid of the other one (and just use this).

adamtheturtle:
Apologies @tomprince , I had WIP that was meant to be in another branch and pushed it here, which is why the fabric stuff is here. I will clean this up.

flocker/cli/functional/test_sshconfig.py
adamtheturtle: A quick search shows nothing and I don't know - I'm not going to dig into this although I agree that it would be nice.

flocker/node/testtools.py
adamtheturtle: That code seems to be gone from `master`.

adamtheturtle:
Passing build.

adamtheturtle:
http://build.clusterhq.com/builders/flocker-osx-10.10/builds/5

PR #1171, opened 2015-02-20
[FLOC-1442] Update copyright and add test for current copyright

jongiddy:
Rather than attempt to update copyright automatically, this code causes the doc build to fail if the current year does not exisdt in the copyright. So, after New Year, doc tests will fail until someone changes the copyright (or removes the assert).

Design question - should this fail as above, or emit a warning (practically invisible to all but most observant watchers), or should we handle it a differewnt way?

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1171)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,9 +13,9 @@
 
 from twisted.python.filepath import FilePath
 
+import datetime
 import sys
 import os
-import re
 
 sys.path.insert(0, FilePath(__file__).parent().parent().path)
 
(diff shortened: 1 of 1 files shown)


tomprince:
Given that this should only trigger once a year, the exact details probably don't matter. This probably is as good as any other option.

*But* this doesn't actually change the generated docs, as our current theme hard-codes the copyright. It should be updated to include this copyright (https://github.com/sphinx-doc/sphinx/blob/master/sphinx/themes/basic/layout.html#L191 is the upstream code that does this).

PR #1172, opened 2015-02-20
Design proposal for FLOC-1206

adamtheturtle:


homebrew-proposal.rst
```diff
@@ -0,0 +1,28 @@
...
+
+It would not provide the other benefits of Continuous Integration.
+That is, CI should be able to tell you before you merge a commit into master whether that commit will break something.
+This is not possible with the current Homebrew set up even if Homebrew builds were run from Flocker pushes, because the Homebrew recipe downloads the latest released .tar.gz from GCS and installs it.
+Unless there is a release which uploads the .tar.gz file to GCS, the Homebrew recipe will not change.
+
+Testing often with CI would test that the links to the dependencies have not gone down, and that a change to Homebrew itself does not interfere with our recipe.
+
+This led to the conclusion that this issue should be covered by an admin script (perhaps put in the current ``admin/make-homebrew-recipe``).
+This script would be run during the release process, in the place of the current Homebrew steps (including the review step).
```
itamarst: What you are suggesting is having two completely different ways of testing things on a Mac, with corresponding duplication of code, configuration, etc.:
1. Buildbot (up for review another branch)
2. Manual script

Looking at the bigger picture for the release process, right now we have two completely different ways of testing:
1. Buildbot
2. Manual testing for the release

Both in the specific case and in the general case this seems non-ideal. Better if everything ran a single way. If it's Buildbot we're using for automated testing then it's Buildbot we should be using for release testing. "The tarball isn't usually available" is true, but Buildbot can do conditional builds, and in fact already does some conditional building depending on whether you're in a release branch or not (documentation, IIRC). Presumably we can do conditional builds like "oh, this has progressed far enough in the release process I can expect X to be the case". You could also modify homebrew recipe to accept alternate URL (but not alternate checksum!) for interim testing, etc..

@tomprince any thoughts?

homebrew-proposal.rst
```diff
@@ -0,0 +1,28 @@
+The initial design by Jon and me is the following prose:
+
+Linking Travis-CI or Buildbot to the homebrew-tap repository would save:
+   #. The release engineer from logging into the Mac Mini and testing the recipe.
+   #. A reviewer of the PR to homebrew-tap’s master logging into the Mac mini and testing the recipe.
+
+It would not provide the other benefits of Continuous Integration.
+That is, CI should be able to tell you before you merge a commit into master whether that commit will break something.
```
tomprince: That depends on exactly what is being tested. It should be possible to generate a homebrew script targeting any particular commit, install that, and run tests against it.

homebrew-proposal.rst
```diff
@@ -0,0 +1,28 @@
...
+This is not possible with the current Homebrew set up even if Homebrew builds were run from Flocker pushes, because the Homebrew recipe downloads the latest released .tar.gz from GCS and installs it.
+Unless there is a release which uploads the .tar.gz file to GCS, the Homebrew recipe will not change.
+
+Testing often with CI would test that the links to the dependencies have not gone down, and that a change to Homebrew itself does not interfere with our recipe.
+
+This led to the conclusion that this issue should be covered by an admin script (perhaps put in the current ``admin/make-homebrew-recipe``).
+This script would be run during the release process, in the place of the current Homebrew steps (including the review step).
+It would:
+   * Log into the Mac mini.
```
tomprince: Given that we have a buildslave running on the mini, this step isn't necessary.

homebrew-proposal.rst
```diff
@@ -0,0 +1,28 @@
...
+   * Start an OS X VM with a script very similar to https://github.com/ClusterHQ/internal-tools/blob/master/bin/start_homebrew_machine.
+   * Instead of just SSHing into the VM and leaving the user to test the recipe, the script would use fabric (or similar) to run the brew test commands on the VM.
```
tomprince: The above two steps do seem like something it would be worthwhile to make into a script. (But that script would likely be run by buildbot on the mac-mini.

homebrew-proposal.rst
```diff
@@ -0,0 +1,28 @@
...
+   * It would merge the recipe into the homebrew-tap master if the test passes.
```
tomprince: This seems like it would be something worth automating, but probably in a separate ticket.

admin/homebrew.py
```diff
@@ -174,5 +183,57 @@ def install
...
+    # This can likely be found another way, but I got it from within the VM:
+    # System Preferences > Network
+    VM_ADDRESS = "172.18.140.54"
+
+    # XXX Requires https://github.com/msteinhoff/vmfusion-python/pull/4 to be
+    # merged
+    vmrun.revertToSnapshot(YOSEMITE_VMX_PATH, 'homebrew-clean')
+    vmrun.start(YOSEMITE_VMX_PATH, gui=False)
+
+    recipe_path = "flocker-{version}.rb".format(version=version)
```
adamtheturtle: @jongiddy and @dwgebler If you want to use this, it might be useful to know that you could replace this with a URL to the recipe, if you get rid of the `Put` below.

PR #1173, opened 2015-02-20
[FLOC-1453] First class datasets

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1453

As a nice side-effect the invariant on `Node` also ensures you can't e.g. move a dataset from one node to another if you're not correspondingly modifying the application that uses it.

I'm really liking pyrsistent.

### DIFF
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -8,7 +8,7 @@ Record types for representing deployment models.
 from characteristic import attributes, Attribute
 
 from pyrsistent import (
-    pmap, PRecord, field, PMap, PSet, pset, PVector, pvector, v,
+    pmap, PRecord, field, PMap, PSet, pset,
     )
 
 from zope.interface import Interface, implementer
@@ -239,8 +239,8 @@ class Node(PRecord):
             if not isinstance(app, Application):
                 return (False, '%r must be Appplication' % (app,))
             if app.volume is not None:
-                 if app.volume.manifestation not in manifestations:
-                     return (False, '%r manifestation is not on node' % (app,))
+                if app.volume.manifestation not in manifestations:
+                    return (False, '%r manifestation is not on node' % (app,))
         for key, value in self.manifestations.items():
             if key != value.dataset_id:
                 return (False, '%r is not correct key for %r' % (key, value))
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -361,9 +361,9 @@ class NodeState(PRecord):
     The current state of a node.
 
     :ivar unicode hostname: The hostname of the node.
-    :ivar running: A ``PVector`` of ``Application`` instances on this node
+    :ivar running: A ``PSet`` of ``Application`` instances on this node
         that are currently running or starting up.
-    :ivar not_running: A ``PVector`` of ``Application`` instances on this
+    :ivar not_running: A ``PSet`` of ``Application`` instances on this
         node that are currently shutting down or stopped.
     :ivar used_ports: A ``PSet`` of ``int``\ s giving the TCP port numbers
         in use (by anything) on this node.
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -13,6 +13,9 @@ from pyrsistent import (
 
 from zope.interface import Interface, implementer
 
+# Not necessary once pyrsistent 0.9 is out:
+PSet.__reduce__ = lambda self: (pset, (list(self),))
+
 
 @attributes(["repository", "tag"], defaults=dict(tag=u'latest'))
 class DockerImage(object):
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -7,7 +7,9 @@ Record types for representing deployment models.
 
 from characteristic import attributes, Attribute
 
-from pyrsistent import pmap, PRecord, field, PMap
+from pyrsistent import (
+    pmap, PRecord, field, PMap, PSet, pset, PVector, pvector, v,
+    )
 
 from zope.interface import Interface, implementer
 
@@ -173,6 +175,12 @@ class Manifestation(object):
 
     :ivar bool primary: If true, this is a primary, otherwise it is a replica.
     """
+    @property
+    def dataset_id(self):
+        """
+        :return unicode: The dataset ID of the dataset.
+        """
+        return self.dataset.dataset_id
 
 
 @attributes(["dataset_id",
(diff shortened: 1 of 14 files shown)


itamarst:
If this is accepted I would recommend a follow up issue to convert the rest of the model classes to use `pyrsistent`.

flocker/control/_model.py
```diff
@@ -215,20 +217,27 @@ class Node(object):
...
+    """
+    def __invariant__(self):
+        manifestations = self.manifestations.values()
+        for app in self.applications:
+            if app.volume is not None:
+                if app.volume.manifestation not in manifestations:
+                    return (False, '%r manifestation is not on node' % (app,))
+        return (True, "")
+
+    hostname = field(type=unicode)
```
tomprince: Probably most of these want to be specified as mandatory.

flocker/control/_model.py
```diff
@@ -215,20 +217,27 @@ class Node(object):
...
+    # XXX with invariant ensuring correct type
+    applications = field(type=frozenset)
```
tomprince: Any reason this isn't a `PSet` now?

flocker/control/httpapi.py
```diff
@@ -288,20 +284,14 @@ def update_dataset(self, dataset_id, primary=None):
...
-                applications=target_node.applications,
-                other_manifestations=(
-                    target_node.other_manifestations
-                    | frozenset({primary_manifestation})
-                )
-            )
+            new_target_node = target_node.transform(
+                ("manifestations", dataset_id), primary_manifestation)
 
         deployment = deployment.update_node(new_target_node)
```
tomprince: I guess this will change once a deployment is using `pyrsistent`.

tomprince:
This change seems reasonable. I haven't checked that the code changes are correct.

flocker/control/_model.py
```diff
@@ -200,14 +202,14 @@ class Dataset(object):
     """
 
 
-@attributes(["hostname",
-             Attribute("applications", default_value=frozenset()),
-             Attribute("other_manifestations", default_value=frozenset())])
-class Node(object):
+class Node(PRecord):
```
exarkun: Neat.  Wonder why we ever even used characteristic.

itamarst: `PRecord` got added in January, I think.

itamarst:
I have no idea why those tests are failing, they don't fail locally for me :cry:

wallrj:
I see the test failures with Python 2.7.5 on Fedora20

There are only three failures, so changing the expected value to pset seems like a good approach.

```
(1453)(first-class-datasets-FLOC-1453 ✕? =)[~/.../HybridLogic/flocker]$ git diff
diff --git a/flocker/node/test/test_deploy.py b/flocker/node/test/test_deploy.py
index c030fdf..e2fda19 100644
--- a/flocker/node/test/test_deploy.py
+++ b/flocker/node/test/test_deploy.py
@@ -9,7 +9,7 @@ from uuid import uuid4
 from zope.interface.verify import verifyObject
 
-from pyrsistent import pmap
+from pyrsistent import pmap, pset
 
 from twisted.internet.defer import fail, FirstError, succeed, Deferred
 from twisted.trial.unittest import SynchronousTestCase, TestCase
@@ -1192,13 +1192,13 @@ class DeployerDiscoverNodeConfigurationTests(SynchronousTestCase):
         d = api.discover_local_state()
 
         self.assertEqual(
-            {Manifestation(
+            pset({Manifestation(
                 dataset=Dataset(
                     dataset_id=DATASET_ID,
                     metadata=pmap({u"name": u"site-example.com"})),
                 primary=True),
              Manifestation(dataset=Dataset(dataset_id=DATASET_ID2),
-                           primary=True)},
+                           primary=True)}),
             self.successResultOf(d).manifestations)
 
 
(1453)(first-class-datasets-FLOC-1453 ✕? =)[~/.../HybridLogic/flocker]$ python --version
Python 2.7.5
(1453)(first-class-datasets-FLOC-1453 ✕? =)[~/.../HybridLogic/flocker]$ trial flocker.node.test.test_deploy.DeployerDiscoverNodeConfigurationTests.test_discover_datasets
flocker.node.test.test_deploy
  DeployerDiscoverNodeConfigurationTests
    test_discover_datasets ...                                             [OK]

Ran 1 tests in 0.005s

PASSED (successes=1)

```

wallrj:
@itamarst discovered that the reason tests fail on buildbot and pass on local workstation is down to a difference in python version. See https://github.com/tobgu/pyrsistent/issues/27#issuecomment-76281428 for discussion

flocker/control/test/test_config.py
```diff
@@ -2666,17 +2666,24 @@ def test_error_on_unrecognized_application_name(self):
...
         """
-        ``_deployment_from_config`` returns a set of ``Node`` objects. One for
-        each key in the supplied nodes dictionary.
+        ``deployment_from_configuration`` returns a set of ``Node``
+        objects. One for each key in the supplied nodes dictionary.
         """
+        manifestation = Manifestation(
+            dataset=Dataset(
+                dataset_id=dataset_id_from_name(
+                    "mysq-hybridcluster"),
```
wallrj: Typo "mysq"

flocker/control/test/test_config.py
```diff
@@ -2666,17 +2666,24 @@ def test_error_on_unrecognized_application_name(self):
...
+            primary=True)
+
         applications = {
             'mysql-hybridcluster': Application(
                 name='mysql-hybridcluster',
-                image=Application(
-                    name='mysql-hybridcluster',
-                    image=DockerImage(repository='flocker/mysql',
-                                      tag='v1.0.0'))
-            )
```
wallrj: So was this test broken? The diff makes it look like we used to supply an `Application` as the `image` attribute of an `Application`.

wallrj:
Thanks @itamarst 

The switch to Pyrsistent  looks exciting...although yet another library for us all to learn.

Can you briefly explain the main motivation for introducing it here? Perhaps at today's standup.

I made a few inline comments.

Looks like you figured out the cause of broken tests on buildbot. Seems to me that would be ok update those broken assertions with pset as the expected type.

Do that, if you agree, and then feel free to merge.

Maybe file followup issues for tracking the eventual fix for that comparison problem
And an issue for switching model classes  from `Characteristic` to `PRecord` (as you mentioned above).

flocker/control/test/test_config.py
itamarst: Yeah, the test was broken previously. `pyrsistent` will help us with that since it allows for type enforcement.

itamarst:
I was mostly afraid of fixing broken tests when I didn't understand the reason. Since I understand the issue now then switching to `pset` should fix it, and is a reasonable thing to do.

PR #1174, opened 2015-02-24
[FLOC-1466] Configration for opening firewall for flocker-control.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1466


admin/package-files/flocker-control-agent.firewalld.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<service>
+  <short>Flocker Control Service Agent Port</short>
+  <description>The Flocker control service manages the state of a flocker cluster.</description>
+  <port protocol="tcp" port="4524"/>
```
tomprince: These are split so that they can be put in different zones.

admin/package-files/flocker-control-agent.ufw
```diff
@@ -0,0 +1,4 @@
+[Samba]
```
tomprince: I haven't tested the ubuntu configuration.

tomprince:
There is an issue with file permissions for packaged files. I'm not yet sure what the best way to fix that is. Probably copying the files somewhere temporary, and setting the appropriate permissions there.

admin/package-files/after-install.sh
```diff
@@ -0,0 +1,2 @@
+# https://git.fedorahosted.org/cgit/firewalld.git/tree/config/macros.firewalld
```
itamarst: That doesn't quite match what that URL suggests?

admin/package-files/flocker-control-agent.firewalld.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
itamarst: Perhaps we should have separate folders for different OSes, instead of mixing CentOS and Ubuntu in the same folder?

admin/packaging.py
```diff
@@ -864,7 +880,23 @@ def omnibus_package_builder(
             BuildPackage(
                 package_type=distribution.package_type(),
                 destination_path=destination_path,
-                source_paths={flocker_node_path: FilePath("/usr/sbin")},
+                source_paths={
+                    flocker_node_path: FilePath("/usr/sbin"),
+                    # Fedora/CentOS firewall configuration
+                    package_files.child('flocker-control-api.firewalld.xml'):
```
itamarst: Wouldn't we only want applicable files for each OS to be installed, rather than installing everything? Not nice to dump random crap in `/etc` if it's for different OS.

admin/packaging.py
```diff
@@ -878,6 +910,7 @@ def omnibus_package_builder(
                 category=category,
                 dependencies=make_dependencies(
                     'node', rpm_version, distribution),
+                after_install=package_files.child('after-install.sh'),
```
itamarst: Likewise, this script is Fedora/CentOS specific?

flocker/provision/_install.py
```diff
@@ -141,17 +141,38 @@ def task_enable_docker():
     ]
 
 
-def task_disable_firewall():
+def configure_firewalld(rule):
```
itamarst: If we're relying on firewall rules in the package, why do we need to do it in the provisioning system?

itamarst:
May've misunderstood some stuff, so could you answer questions above? Thanks.

admin/package-files/flocker-control-agent.firewalld.xml
tomprince: I would be more inclined to split things up by tool (so `firewalld/flocker-control-agent.xml`), if we were to split things up.

itamarst: Makes sense. But I do think they should be split.

admin/packaging.py
tomprince: At least firewalld is available on [ubuntu](http://packages.ubuntu.com/search?keywords=firewalld). So, it seems to make sense to install the configuration files for it, even if it isn't used by default.

itamarst: OK, I'll buy that.

itamarst:
One more question: how will this be tested? Is the presumption that it'll make e.g. rackspace acceptance tests start passing?

flocker/provision/_install.py
tomprince: The firewalls rules aren't active by default, so we need to turn them on. The configuration that is relevant to this PR is https://github.com/ClusterHQ/flocker/pull/1174/files#diff-bd81b442286984f45ec7fe47f8aa0584R172 which enables the defined service by name, rather than hard-coding the name.

This also allows (although I don't think we should bother documenting) using slightly more complicated commands to only allow access to these services from some addresses.

admin/package-files/after-install.sh
tomprince: That file defines a macro, which expands to the line that I have. In a fedora/centos package, you would just put `%firewalld_reload`, but
1. I wanted this to be cross-platform.
2. It requires the firewalld-filesystem to be installed.
I think more sense to just install the snippet.

tomprince:
> One more question: how will this be tested? Is the presumption that it'll make e.g. rackspace acceptance tests start passing?

The API tests, yes.

itamarst:
Thanks, please proceed.

admin/package-files/after-install.sh
tomprince: Also, it appears that CentOS doesn't have that macro.

tomprince:
> There is an issue with file permissions for packaged files. I'm not yet sure what the best way to fix that is. Probably copying the files somewhere temporary, and setting the appropriate permissions there.

Turns out, we were setting a umask on buildbot, so that mock could write to an appropriate directory. Since we no longer use mock, we can remove that. That means that the build will depend on the umask that was used to checkout the repository. This isn't ideal but `lintian` will catch if this breaks (which is how I noticed it).

flocker/provision/_rackspace.py
```diff
@@ -64,7 +68,7 @@ def rackspace_provisioner(username, key, region, keyname):
             "ex_config_drive": "true",
         },
         provision=provision_rackspace,
-        default_size="performance1-2",
+        default_size="performance1-8",
```
itamarst: What's the reason for this change?

docs/advanced/api.rst
```diff
@@ -7,6 +7,19 @@ Flocker REST API
 We are currently in the process of developing an external HTTP-based REST API for Flocker.
 While this API is not yet directly accessible in a standard Flocker setup, the documentation below will give a sense of what will eventually be available.
 
+
+Installation
+============
+
+To allow remote access to the control agent REST API, and for agent connections,
```
itamarst: Maybe make more explicit that this refers to platforms using `firewalld` like CentOS or Fedora.

itamarst:
I'll assume Rackspace testing will catch any problems with these actually working.

itamarst:
OK, address minor comment and merge - thanks.

flocker/provision/_rackspace.py
tomprince: We suggest 8G machines, but the one we were using was only 2G.

tomprince:
> I'll assume Rackspace testing will catch any problems with these actually working.

Yes, You can see in the rackspace test run that the API tests are passing.

docs/advanced/api.rst
tomprince: I made it explicit, by making a Fedora/CentOS section.

PR #1177, opened 2015-02-24
[FLOC-1470] Sketch of design for busting firewalls.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1470

### DIFF
--- a/docs/indepth/installation.rst
+++ b/docs/indepth/installation.rst
@@ -356,14 +356,6 @@ To enable and start Docker, run the following commands in a root console:
 .. task:: enable_docker
    :prompt: [root@node]#
 
-To enable Flocker to forward ports between nodes, the firewall needs to be configured to allow forwarding.
-On a typical fedora installation, the firewall is configured by `firewalld <https://fedoraproject.org/wiki/FirewallD>`_.
-(Note: The Fedora AWS images don't have firewalld installed, as there is an external firewall configuration.)
-The following commands will configure firewalld to enable forwarding:
-
-.. task:: disable_firewall
-   :prompt: [root@node]#
-
 Flocker requires a ZFS pool named ``flocker``.
 The following commands will create a 10 gigabyte ZFS pool backed by a file.
 Paste them into a root console:
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -347,6 +347,31 @@ class SetProxies(object):
         return gather_deferreds(results)
 
 
+@implementer(IStateChange)
+@attributes(["ports"])
+class OpenPorts(object):
+    """
+    Set the ports which will have the firewall opened.
+
+    :ivar ports: A collection of ``Port`` objects.
+    """
+    def run(self, deployer):
+        results = []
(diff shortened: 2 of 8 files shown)


docs/indepth/installation.rst
```diff
@@ -356,14 +356,6 @@ To enable and start Docker, run the following commands in a root console:
...
    :prompt: [root@node]#
 
-To enable Flocker to forward ports between nodes, the firewall needs to be configured to allow forwarding.
-On a typical fedora installation, the firewall is configured by `firewalld <https://fedoraproject.org/wiki/FirewallD>`_.
-(Note: The Fedora AWS images don't have firewalld installed, as there is an external firewall configuration.)
-The following commands will configure firewalld to enable forwarding:
-
-.. task:: disable_firewall
-   :prompt: [root@node]#
-
```
exarkun: Hooray for removing a manual setup step.

flocker/node/_deploy.py
```diff
@@ -347,6 +347,31 @@ def run(self, deployer):
         return gather_deferreds(results)
 
 
+@implementer(IStateChange)
+@attributes(["ports"])
```
exarkun: I guess we decided to use pyrsistent by default.

vagrant/tutorial/cache.sh
```diff
@@ -17,3 +17,16 @@ for image in busybox clusterhq/mongodb dockerfile/redis clusterhq/flask; do
...
+systemctl enable firewalld
+systemctl start firewalld
+for PORT in 3306 9200 5000 80 5432 27018; do # List of ports used in acceptance tests
+    for DEST in 172.16.255.240 172.16.255.241; do # List of vagrant nodes
+        # Extra rules for proxies
+        firewall-cmd --permanent --add-rule ipv4 filter FORWARD 0 --destination $DEST --protocol tcp --desitnation-port $PORT -j ACCEPT
+    done
+    # New rule for open ports
+    firewall-cmd --permanent --add-rule ipv4 filter INPUT 0 --protocol tcp --desitnation-port $PORT -j ACCEPT
+done
```
exarkun: I don't quite understand why these rules are being added.  I suppose they are necessary because we no longer simply disable the system firewall **and** the new iptables rules added by this design only accept traffic between the docker container and its host, not traffic between the two vagrant instances?

If that's so, then don't people installing Flocker on their own systems (eg CentOS 7) with a firewall still need to be told they have to allow certain traffic through the firewall (thus, there still needs to be something in the install documentation)?

exarkun:
Thanks.  The basic idea here, do some more iptables configuration as part of the deployment logic, certainly makes sense.  I wouldn't mind hearing a bit more in regards to the last comment I left inline but I don't think the answer is going to change much about this design - so please proceed.

vagrant/tutorial/cache.sh
tomprince: Looks like it didn't end up building, but these rules are/were supposed to mimic what would be implemented by this branch, to demonstrate that the proposed rules work. This section will be deleted in the final branch.

PR #1178, opened 2015-02-24
[FLOC-1199] Generate uuids in configuration parser

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1199


exarkun:
Yep, makes sense.  New unit tests will be straightforward, I expect (much like the old ones but a lot simpler).  Thanks.  Please proceed.

flocker/control/test/test_config.py
```diff
@@ -3339,3 +3346,29 @@ def test_unicode(self):
...
         )
+
+
+class DatasetIdFromNameTests(SynchronousTestCase):
+    """
+    Tests for ``dataset_id_from_name``.
+    """
+    def test_uuid(self):
+        """
+        ``dataset_id_from_name`` returns a UUID.
```
exarkun: And by UUID you mean the unicode representation of a UUID? :smile:

flocker/control/_config.py
```diff
@@ -295,7 +297,8 @@ def convert_volume(self):
                 u'mountpoint': self._application.volume.mountpoint.path
             }
             dataset = self._application.volume.dataset
-            if dataset.dataset_id is not None:
+            if dataset.dataset_id != dataset_id_from_name(
+                    self._application.name):
```
exarkun: This doesn't seem strictly necessary?  Except it makes some overly specific tests continue to pass instead of breaking them with a change they shouldn't actually be concerned with, which maybe is a sufficient reason to do it...

It probably merits a comment, at least - either saying it's only done to avoid having to do a bunch of boring test maintenance, or else giving another good reason if there is one.

exarkun: The reason it seems unnecessary to me is that if the dataset_id is included in the output then the next thing that looks at this data structure will just go ahead and use that id.  And if it isn't included then it will just re-generate the exact same value.  Either way, the result is the same.

But if we decide to change how the dataset_id is generated - for example, to support some kind of weak namespacing-like feature for `flocker up` (to let different developers use it simultaneously against one cluster) then it might get hard for the server to re-generate the same uuid.  Not that that means we should change this now but maybe it does mean we will eventually have to change it and fix all those tests.

exarkun:
Some tests are failing - they create `Dataset` instances with `dataset_id` set to `None`.  The deployer used to catch and fix this, now it doesn't.  Probably the tests just need the minor obvious tweak.

The rest looks good.  Please merge after addressing in the inline comments, fixing those test failures, and seeing a green buildbot.  Thanks.

PR #1181, opened 2015-02-25
[FLOC-1290] Add sketch of variant support for provisioning.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1290

This is a sketch of adding support for the acceptance runner to run tests with newer versions docker.

Running it will be done in https://clusterhq.atlassian.net/browse/FLOC-1304


itamarst:
If we don't already have one, might also want a separate issue for running our non-acceptance tests (specifically functional tests, but easiest to just do everything) against latest Docker.

tomprince:
:frowning: For some reason, I thought this had been accepted.

PR #1184, opened 2015-02-25
[FLOC-1465] Dataset deletion endpoint

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1465


wallrj:
Thanks @itamarst 

This all looks good, please proceed.

A couple of questions: 
1. When does a deleted dataset get removed from the configuration?
2. What will be the API response if I attempt to delete a deleted dataset? I imagine that it should be the same as if I attempt to delete an unknown dataset. 

Consider adding an extra example to demonstrate that behaviour. (idempotent or 404 behaviour)

itamarst:
1. Not clear how deleted dataset gets removed from configuration. It'd be nice to have, but not essential to the feature, so for now we're not prioritizing AFAIK.
2. Idempotent, yeah. An example... might be a little weird given our current example setup which is a single request/response per example? I will definitely make sure to document it in the prose.

flocker/control/httpapi.py
```diff
@@ -200,17 +200,97 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
-                result[u"maximum_size"] = maximum_size
+            result = api_dataset_from_dataset_and_node(dataset, primary)
             return EndpointResponse(CREATED, result)
         saving.addCallback(saved)
         return saving
 
+    def _find_manifestation_and_node(self, dataset_id):
+        """
+        Given the ID of a dataset, find its primary manifestation and the node
+        it's one.
```
wallrj: typo "one"

flocker/control/httpapi.py
```diff
@@ -200,17 +200,97 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
+
+        return primary_manifestation, origin_node
+
+    @app.route("/configuration/datasets/<dataset_id>", methods=['DELETE'])
+    @user_documentation(
+        """
+        Delete an existing dataset.
+
+        Deletion is idempotent: deleting a dataset multiple times will
+        result in the same the response.
```
wallrj: Typo "the same the response"

docs/advanced/api_examples.yml
```diff
@@ -208,6 +208,43 @@
...
+  doc: |
+    Delete a dataset.
+
+  requires:
+    - "create dataset with dataset_id"
+
+  request: |
+    DELETE /v1/configuration/datasets/886ed03a-5606-453a-94a9-a1cbaf35164c HTTP/1.1
+
+    {"primary": "%(NODE_1)s"}
```
wallrj: I don't think client supplies a request body. Perhaps this is a copy / paste error.

docs/advanced/api_examples.yml
```diff
@@ -208,6 +208,43 @@
...
+  id:
+    "delete dataset with unknown dataset id"
+
+  doc: |
+    An attempt to delete an unknown dataset results in a 404 Not Found response.
+
+  request: |
+    DELETE /v1/configuration/datasets/<unknown_dataset_id> HTTP/1.1
+
+    {"primary": "%(NODE_1)s"}
```
wallrj: Same again.

flocker/control/httpapi.py
```diff
@@ -200,17 +200,97 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
+        schema_store=SCHEMAS
+    )
+    def delete_dataset(self, dataset_id):
+        """
+        Delete an existing dataset in the cluster configuration.
+
+       :param unicode dataset_id: The unique identifier of the dataset.  This
+            is a string giving a UUID (per RFC 4122).
+
+        :return: A ``dict`` describing the dataset which has been added to the
```
wallrj: "added to" > "deleted from"

flocker/control/httpapi.py
```diff
@@ -200,17 +200,97 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
+        :param unicode dataset_id: The unique identifier of the dataset.  This
+            is a string giving a UUID (per RFC 4122).
+
+        :return: Tuple containing the primary ``Manifestation`` and the
+            ``Node`` it is on.
+        """
+        # Get the current configuration.
+        deployment = self.persistence_service.get()
+
+        # Lookup the node that has a primary Manifestation (if any)
```
wallrj: This comment (I think I wrote it originally) seems misleading or at least in the wrong place. Should be two lines below, or probably deleted.

flocker/control/httpapi.py
```diff
@@ -200,17 +200,97 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
+        manifestations_and_nodes = manifestations_from_deployment(
+            deployment, dataset_id)
```
wallrj: I mentioned it in the design review, but here's another case where it seems like deleted datasets should be removed from the deployment.

An attempt to delete a deleted dataset, should result in DATASET_NOT_FOUND.

And an attempt to update a deleted dataset should result in DATASET_NOT_FOUND.

flocker/control/test/test_httpapi.py
```diff
@@ -751,6 +756,131 @@ def saved(ignored):
...
+    Tests for the dataset deletion endpoint at
+    ``/configuration/datasets/<dataset_id>``.
+    """
+    def test_unknown_dataset(self):
+        """
+        NOT_FOUND is returned if the requested dataset_id doesn't exist.
+        The error includes the requested dataset_id.
+        """
+        unknown_dataset_id = unicode(uuid4())
+        deleting = self.assertResponseCode(
```
wallrj: I think `assertResult` could be used here.

flocker/control/test/test_httpapi.py
```diff
@@ -751,6 +756,131 @@ def saved(ignored):
...
+        return deleting
+
+    def _test_delete(self, dataset, deployment):
+        """
+        Helper method which makes an API call to delete the supplied
+        ``dataset`` from ``origin`` and finally asserts that the API call
+        returned the expected result and that the persistence_service has
+        been updated.
+
+        :param Dataset dataset: The dataset which will be moved.
```
wallrj: Typo "moved"

flocker/control/httpapi.py
itamarst: Deleting a deleted dataset should give same response, if we want this API to be idempotent (and we do, I think). So I think that's OK.

The second case is definitely a bug. I'll fix it when I get in.

wallrj:
Thanks @itamarst 

This makes sense.

I'm still not sure about the idempotent DELETE behaviour. I left some comments about that above, in particular about what should happen if I attempt to update a deleted dataset.

There were a few typos.

The PR diff seems to contain changes from various other branches, probably just because it needs merging forward.

DELETE API examples contain some request bodies, which seems wrong.

There's missing coverage of the IndexError when there are no primary manifestations, but I think we deliberately removed the test for that behaviour in the UPDATE datasets API branch.
 * http://build.clusterhq.com/results/complete-coverage/delete-dataset-API-FLOC-1465/flocker_control_httpapi.html#n237

...so I guess it can wait until we do https://clusterhq.atlassian.net/browse/FLOC-1403

Answer or address those points, file followup issues then merge, or put it back up for review if you decide to change the DELETE deleted datasets behaviour.

flocker/cli/functional/test_sshconfig.py
```diff
@@ -64,7 +64,13 @@ def test_connection_failed(self):
...
         exc = self.assertRaises(CalledProcessError,
                                 self.configure_ssh, b"127.0.0.1", port)
-        self.assertIn(b"refused", exc.output)
+        # There are different error messages on different platforms.
+        # On Linux the error may be:
+        # 'ssh: connect to host 127.0.0.1 port 34716: Connection refused\r\n'
+        # On OS X the error may be:
+        # 'ssh: connect to host 127.0.0.1 port 56711: Operation timed out\r\n'
+        self.assertTrue(b"refused" in exc.output or "Operation timed out" in
+                        exc.output)
```
exarkun: One of these is `b""` and the other is `""`

itamarst: I didn't touch this code in this branch...

itamarst:
Thanks! Definitely something weird about the diff...

PR #1185, opened 2015-02-25
[FLOC-1476] ZFS "volume" deletion.

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1476

Just enough to get it working.

### DIFF
--- a/flocker/volume/filesystems/interfaces.py
+++ b/flocker/volume/filesystems/interfaces.py
@@ -135,6 +135,17 @@ class IStoragePool(Interface):
             problems.
         """
 
+    def destroy(volume):
+        """
+        Destroy the filesystem for the given volume.
+
+        :param volume: The volume whose filesystem should be deleted.
+        :type volume: :class:`flocker.volume.service.Volume`
+
+        :return: Deferred that fires on filesystem deletion, or errbacks
+            if deletion failed.
+        """
+
     def set_maximum_size(volume):
         """
         Set the maximum size of a filesystem for the given volume.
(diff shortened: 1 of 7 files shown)


wallrj:
Thanks @itamarst 

This looks good.

1. I liked it when you included proposed tests in your previous designs, but I guess you're intending to do the implementation here, so not necessary.
2. Will there need to be handling of the case where we attempt to delete a mounted / in-use dataset?
3. Or are we going to assume that the convergence agent ensures that applications are stopped and datasets unmounted before making any attempt to delete?

Anyway, please proceed or add some further test ideas if you'd like someone else to write the code.

itamarst:
This isn't even on the level of convergence, this is just a low-level operation. The convergence agent (next issue queued up after this one) will ensure stuff is detached, yeah.

wallrj:
Thanks @itamarst 

This looks great.
* Tests pass on buildbot.
* Coverage is complete

Please merge.

PS I was a bit puzzled by the coverage change report:
 * http://build.clusterhq.com/results/change-coverage/zfs-deletion-FLOC-1476/flocker/volume/service.py%2Ccover

...which seems to have got mixed up with the changes to flocker.volume.service from another branch.

```
      # XXX or maybe caller should just do that directly?
      def destroy(self, volume):
          """
          Destroy an existing volume.
  
          :param Volume volume: The ``Volume`` instance to delete from the
              storage pool.
  
          :return: A ``Deferred`` that fires when deletion is done.
          """
          #return self.pool.destroy(volume)
!         pass
```

 @tomprince Any ideas?

PR #1186, opened 2015-02-26
Intermittent failure floc 1317

jongiddy:
This PR ensures the start of a Docker container is not indicated as done until the container is in state Running.  This may fix the problem with intermittent failures due to inability to remove the Docker containers.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -7,7 +7,7 @@ Docker API client.
 
 from __future__ import absolute_import
 
-from time import sleep
+from time import time, sleep
 
 from zope.interface import Interface, implementer
 
@@ -442,7 +442,7 @@ class DockerClient(object):
             # stop on this container Docker might well complain it knows
             # not the container of which we speak. To prevent this we poll
             # until it does exist.
-            self._blocking_wait(self._blocking_exists, container_name)
+            self._blocking_wait(5, self._blocking_exists, container_name)
 
             self._client.start(container_name)
             # Tests fail intermittently because containers cannot be
@@ -453,7 +453,7 @@ class DockerClient(object):
             # returns before the container is fully created.
             # This is currently only a theory, as the error only occurs
             # on Buildbot, and cannot be reliably re-created.
-            self._blocking_wait(self._blocking_started, container_name)
+            self._blocking_wait(10, self._blocking_started, container_name)
         d = deferToThread(_add)
 
         def _extract_error(failure):
@@ -465,7 +465,7 @@ class DockerClient(object):
         d.addErrback(_extract_error)
         return d
 
-    def _blocking_wait(self, is_ready, arg):
+    def _blocking_wait(self, timeout, is_ready, arg):
         """
         Wait until a condition is satisfied.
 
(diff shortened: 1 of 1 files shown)


flocker/node/_docker.py
```diff
@@ -457,6 +465,20 @@ def _extract_error(failure):
...
+        Wait until a condition is satisfied.
+
+        :param function is_ready: Function testing whether condition is
+            satisfied.
+        :param unicode arg: An argument to pass to the is_ready function.
+        """
+        time_limit = time() + timeout
+        while not is_ready(arg):
+            if time() > time_limit:
+                raise RuntimeError('Timeout > %s seconds' % timeout)
```
exarkun: minor point, `RuntimeError` is for errors *in* the runtime (ie, CPython) rather than errors *at* runtime.

flocker/node/_docker.py
```diff
@@ -442,10 +442,18 @@ def _add():
...
+
             self._client.start(container_name)
+            # Tests fail intermittently because containers cannot be
+            # removed after being stopped.  It may be that the stop
+            # command sees the container before it is fully started,
+            # and responds with OK, since the container is not running.
+            # This would match with the experience above, where create
+            # returns before the container is fully created.
+            # This is currently only a theory, as the error only occurs
+            # on Buildbot, and cannot be reliably re-created.
```
exarkun: Can we instrument the code/tests so that a failing run on Buildbot reveals the exact request/response information exchanged with Docker?

flocker/node/_docker.py
```diff
@@ -442,10 +442,18 @@ def _add():
             # stop on this container Docker might well complain it knows
             # not the container of which we speak. To prevent this we poll
             # until it does exist.
-            while not self._blocking_exists(container_name):
-                sleep(0.001)
-                continue
+            self._blocking_wait(5, self._blocking_exists, container_name)
```
exarkun: This and the other new timeout value make me a little uneasy.  Of course, the code being replaced here is "block forever" which could certainly have its own problems.

The downfall of timeouts is that sometimes you hit them.  And the really nasty thing is that you're pretty much guaranteed not to hit them during development (because of course you select a timeout that's long enough).  Then when the code runs in some heavily loaded production environment and things slow down enough... you've got some obscure exception you've never seen before.

I'd really like to see what state the container is in, as reported by docker, when this problem is triggered.  If it reveals that it just takes a little while for docker to get the container into the state we asked for, great - I think we don't need the timeout or we can have a really, really long one.

*Alternatively*, this may be the wrong layer at which to apply a fix.  Perhaps Docker calls simply fail sometimes and there's nothing to be done to make them reliable.  Fortunately this code already executes in the context of a loop.  It's possible the best solution is to let these calls fail sometimes and make sure the way the convergence agent handles those failures is by performing another convergence iteration, in which hopefully the calls will fail.

The main thing to figure out is whether these are really transient failures that do go away after docker catches up or if we're doing something basically wrong that's causing persistent error conditions.

exarkun:
Thanks.  I can see how this change might address the problem but I think I'd like the problem to be a little more completely understood before deciding on a strategy for addressing it.

What do you think about introducing some extra logging?  I haven't looked at the Python Docker library we're using to find out what logging capabilities it actually has - but hopefully there's at least some?

jongiddy:
Closed in favour of #1315

PR #1187, opened 2015-02-26
[FLOC-1460] Include dataset path in cluster state endpoint

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1460

Depends on #1173 being merged first.


wallrj:
I can see how this would work.

* The failed builds on the sketch branch suggest that there might be a problem serialising the `_PathMap` `FilePath` values for transmission by AMP from agent to control service.

* I'm not sure whether to ask this question, for fear of sounding dumb, but if mount path derived from dataset ID,  and if flocker nodes always use the same mount root, then why can't this all be calculated by the control service?

* I'm also a bit confused about about whether the new definition of `Manifestation` is a "A dataset that is mounted on a node" (as in the docstring) ...in which case....
 
* I'd expect the mountpoint to be an attribute of the Manifestation rather than AttachedVolume....or 

* Perhaps that's planned for a future model

* I wonder how this fits in with the IaaS BlockDevice agent that @exarkun  and I are working on. If a dataset exists on an *unattached* blockdevice, does that count as a Manifestation?

* And where in the Deployment model will such unattached manifestations be listed?

Submit for a further design review if any of these questions ring any bells or proceed if you're confident about how it's all going to fit together.

wallrj:
Oh, and I merged forward to get a more succinct diff.

itamarst:
* "I'm not sure whether to ask this question, for fear of sounding dumb, but if mount path derived from dataset ID, and if flocker nodes always use the same mount root, then why can't this all be calculated by the control service?"
  It's a reasonable point. My thought that on OpenStack the volume manager UUID and whole concept of "ownership" doesn't really make any sense. I can also imagine constraints on 
* "I'd expect the mountpoint to be an attribute of the Manifestation rather than AttachedVolume....".
  No, remember that there's *two* paths: the one on the host filesystem, and the one inside the container. `AttachedVolume` is the latter.
* "I wonder how this fits in with the IaaS BlockDevice agent that @exarkun and I are working on. If a dataset exists on an unattached blockdevice, does that count as a Manifestation?"
  No, that's not a manifestation. That's just a dataset. With block devices you can have datasets with no manifestations at all.
* "And where in the Deployment model will such unattached manifestations be listed?"
  I can't find an issue, but there's going to have be one. My guess is all datasets will be stored on the `Deployment` object.

exarkun:
> >
> > And where in the Deployment model will such unattached manifestations be listed?
> >
>
> I can't find an issue, but there's going to have be one. My guess is all datasets will be stored on the Deployment object.
>

Yes.  This will definitely make sense when get to the point where we want to limit IaaS API calls to be made only from a single process.  It can discover these datasets without manifestations (by listing the volumes that exist) and send them up to the control service where they can be included in the state `Deployment` and sent out to any other interested parties (the other half of the IaaS dataset convergence implementation which runs on all storage cluster nodes, presumably, as well as API consumers).

However, perhaps that's still a ways off.  The decision for now is to make IaaS API calls from all of the dataset convergence agents.  This is a lot of extra API calls and it will be good to switch to the scheme outlined above - but it will work for now and it's both simpler and less work.  In this scheme, each agent can just query the IaaS for extant volumes (inferring extant datasets from this).

That still leaves the requirement for the control service to publish this information to API consumers, though.  So at least one dataset convergence agent needs to push this information up to the control service and the control service needs to persist it.

docs/advanced/api_examples.yml
```diff
@@ -253,4 +254,9 @@
   response: |
     HTTP/1.1 200 OK
 
-    [{"dataset_id": "47440eff-e933-4de0-b56c-d3469b61421f", "primary": "%(NODE_0)s", "maximum_size": 1073741824, "metadata": {}, "deleted": false}]
+    [{"dataset_id": "47440eff-e933-4de0-b56c-d3469b61421f",
+      "primary": "%(NODE_0)s",
+      "maximum_size": 1073741824,
+      "metadata": {},
+      "deleted": false,
+      "path": "/flocker/47440eff"}]
```
exarkun: intentionally or accidentally truncated uuid?  if intentionally, that seems like an interesting decision - care to explain?

flocker/control/schema/endpoints.yml
```diff
@@ -34,6 +34,13 @@ definitions:
           guaranteed to exist.
         type: boolean
 
+      path:
+        title: "Mount path"
+        description: |
+          The filesystem path where the dataset is mounted on a node.
+          Can only be present in cluster state, not in configuration.
```
exarkun: Sounds like it might be time to split up the schema for these two endpoints?

exarkun: Probably worth saying this will be an absolute path.

docs/advanced/api_examples.yml
```diff
@@ -245,7 +245,8 @@
     "get state datasets"
 
   doc: |
-    Get a list of all datasets in a deployment.
+    Get a list of all datasets in a deployment. The filesystem path where
+    the dataset is mounted will be included.
```
exarkun: Lots of other stuff is included too.  What's the motivation behind singling this property out?  Is it to do with the fact that the schema definition is currently unclear about where the path property is actually valid/possible?

docs/advanced/api_examples.yml
exarkun: regarding the reformatting of this line, thank you.  in general, it would be nice if the doc renderer pretty-printed our json for us, I bet.  This is an easier fix though.

flocker/control/_model.py
```diff
@@ -356,6 +357,17 @@ class DatasetChanges(object):
...
 
+class _PathMap(CheckedPMap):
+    """
+    A mapping between dataset IDs and the paths where they are mounted.
+
+    See https://github.com/tobgu/pyrsistent/issues/26 for more succinct
+    idiom combining this with ``field()``.
+    """
+    __key_type__ = unicode
+    __value_type__ = FilePath
```
exarkun: Kind of unrelated, but I wonder if there's a way to make these interface-based instead of type-based.

docs/advanced/api_examples.yml
itamarst: Intentionally, because there's no actual guarantee what the path will be so I don't want to give impression you can just assume a particular value based on the UUID.

flocker/control/_model.py
itamarst: Yeah, there's custom invariants that are just a function doing whatever, so we could write our own factory that creates classes that check by interface.

flocker/node/_deploy.py
```diff
@@ -397,6 +397,8 @@ def applications_from_units(result):
             running = []
             not_running = []
             manifestations = []  # Manifestation objects we're constructing
+            manifestation_paths = {dataset_id: path for (path, (dataset_id, _))
+                                   in available_manifestations.items()}
```
exarkun: maybe decoding into unicode here, rather than in the API, would be better.

flocker/node/_deploy.py
```diff
@@ -473,6 +475,7 @@ def applications_from_units(result):
                 not_running=not_running,
                 used_ports=self.network.enumerate_used_ports(),
                 manifestations=manifestations,
+                paths=manifestation_paths,
```
exarkun: I forget why this isn't represented using a new `path` attribute on `Manifestation`.  Can you write that down somewhere?

itamarst: Because this isn't something you can configure, and we want `Manifestation` etc. to be only stuff shared by both configuration and state. I'll right it down.

flocker/control/schema/endpoints.yml
itamarst: I'll split the schemas up.

docs/advanced/api_examples.yml
itamarst: It's the one thing that is fundamentally different between this endpoint and all the others that related to datasets. This is true regardless of whether we have two schemas or one.

flocker/node/_deploy.py
exarkun: Oh right but it's represented as a `FilePath` throughout and `FilePath` restricts us to bytes-only still.  Nevermind.

docs/advanced/api_examples.yml
exarkun: Maybe using a completely unrelated random byte sequence would be good, then, to avoid the appearance that what's going on is truncation.

This will make automatic example verification a little difficult later but I guess we'll figure out what to do about that when we actually get around to implementing that.

docs/advanced/api_examples.yml
exarkun: There's metadata, too. :smile:

exarkun:
Thanks.  Looks good.  Buildbot has some trouble with the acceptance tests but they **look** spurious...  Address inline comments, make sure tests really are passing on buildbot, then merge.

docs/advanced/api_examples.yml
itamarst: Metadata has been excised from the state schema.

PR #1188, opened 2015-02-26
[FLOC-1477] Sketch of deployer deletion action and calculation

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1477

This presumes #1185 and #1184 for low-level ZFS destroy operation and deleted attribute on `Dataset` class respectively.

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -320,6 +320,35 @@ class PushDataset(object):
 
 
 @implementer(IStateChange)
+@attributes(["dataset"])
+class DeleteDataset(object):
+    """
+    Delete all local copies of the dataset.
+
+    A better action would be one that deletes a specific manifestation
+    ("volume" in flocker.volume legacy terminology). Unfortunately
+    currently "remotely owned volumes" (legacy terminology), aka
+    non-primary manifestations or replicas, are not exposed to the
+    deployer, so we have to enumerate them here.
+
+    :ivar Dataset: The dataset to delete.
+    """
+    def run(self, deployer):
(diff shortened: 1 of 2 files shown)


flocker/node/_deploy.py
```diff
@@ -320,6 +320,35 @@ def run(self, deployer):
...
+    def run(self, deployer):
+        # service = deployer.volume_service
+        # d = service.enumerate()
+
+        # def got_volumes(volumes):
+        #     deletions = []
+        #     for volume in volumes:
+        #         if volume.name.dataset_id == self.dataset.dataset_id:
+        #             deletions.append(service.pool.destroy(volume))
+        #     return gatherResults(deletions)
```
exarkun: Some error handling/logging here would probably be nice (given the error-handling behavior of `gatherResults`).

flocker/node/_deploy.py
```diff
@@ -735,5 +768,8 @@ def find_dataset_changes(hostname, current_state, desired_state):
         local_current_dataset_ids | remote_current_dataset_ids)
     creating = set(dataset for dataset in local_desired_datasets
                    if dataset.dataset_id in creating_dataset_ids)
+
+    # XXX all datasets with deleted=True attribute should be added to deleted
+    # set on the DatasetChanges
```
exarkun: `find_dataset_changes` is getting somewhat long.  Any way we can factor any of those code into separate helpers?

exarkun:
Thanks.  This looks great.  Just those two minor inline comments, integrate as you think makes sense.  Please proceed.

PR #1190, opened 2015-03-01
[FLOC-1066] CentOS Package Repo on S3

adamtheturtle:
This is a design to resolve https://clusterhq.atlassian.net/browse/FLOC-1066 and incidentally covers https://clusterhq.atlassian.net/browse/FLOC-506.

In summary it:
* Moves Fedora RPM hosting to S3,
* Adds CentOS hosting (also on S3),
* Separates development and marketing RPMs (and changes to the docs to show the relevant URLs),
* Replaces ``subprocess.check_call``s with Python libraries,
* Changes the ``upload-rpms`` script to be a stub like other `admin` scripts.

This goes with a branch in another repository, https://github.com/ClusterHQ/fedora-packages/compare/centos-repo-FLOC-1066.
However I would like feedback on a few things before continuing.
- In GCS there is a SRPMS/ directory. What has to be done to replicate this? Does this have to be done?
- Instead of hardcoding fedora in `ClusterHQ/fedora-packages/clusterhq.repo` it is possible to use a custom variable e.g. `$osname` and have the user run, for example ``$ echo "fedora" > /etc/yum/vars/osname`` in the install instructions for Fedora, and similar in the instructions for fedora.
However, would it be better to have two separate rpms and `clusterhq.repo` entries. Perhaps created from two separate Git repositories. If not, the fedora-packages repository has an outdated name.

What remains:
- There is no testing on this branch. I would like feedback on a suitable scope for testing. The code that this replaces has no testing, but there is now a framework for testing part of this (`FakeAWS`).
- I have done manual testing but would appreciate a reviewer doing the same to check that the code works as expected.


admin/release.py
```diff
@@ -303,3 +319,178 @@ def publish_docs_main(args, base_path, top_level):
...
+    :param bytes target_key: Path within S3 bucket to upload repository to.
+    :param bytes source_repo: Repository on the build server to get packages
+        from.
+    :param list packages: List of bytes, each specifying the name of a package
+        to upload to the repository.
+    """
+    # Import these here so that this file can be imported when yum is not
+    # available.
+    from yum import YumBase
+    from pakrat import repo
```
tomprince: I think it makes more sense to use the external programs. Theses packages aren't on PyPI, and it seems that using them is somewhat more complicated that using external programs.

tomprince:
> In GCS there is a SRPMS/ directory. What has to be done to replicate this? Does this have to be done?

We probably don't need this, since fpm-built packages don't have corresponding source packages.

tomprince:
> Instead of hardcoding fedora in ClusterHQ/fedora-packages/clusterhq.repo it is possible to use a custom variable e.g. $osname and have the user run, for example $ echo "fedora" > /etc/yum/vars/osname in the install instructions for Fedora, and similar in the instructions for fedora. However, would it be better to have two separate rpms and clusterhq.repo entries. Perhaps created from two separate Git repositories. If not, the fedora-packages repository has an outdated name.

We can't have the users run a command like that. Which, I guess, means that we need separate packages. The `fedora-packages name is out-dated, although I'm not sure if it is worth changing. On the other hand, perhaps the still relevant files should just be moved here.

admin/release.py
```diff
@@ -303,3 +319,178 @@ def publish_docs_main(args, base_path, top_level):
...
+                                               patterns=packages)
+    flocker_repo.pkgdir = os.path.join(rpm_directory.path, target_key)
+    base.downloadPkgs(yum_packages)
+
+    # Update repository metadata
+    flocker_repo.pkgdir = os.path.join(flocker_repo.pkgdir, 'repodata')
+    repo.create_metadata(repo=flocker_repo)
+
+    # Upload updated repository
+    for root, dirs, files in os.walk(rpm_directory.path):
```
tomprince: `FilePath.walk`

tomprince: Although, we only want to upload
1. New RPMs.
2. New yum metadata: `repodata/repomod.xml` and the files referenced there. (It probably doesn't make sense to parse `repomod.xml`, just grab the new files in `repodata`.

admin/release.py
```diff
@@ -303,3 +319,178 @@ def publish_docs_main(args, base_path, top_level):
...
+    rpm_directory.createDirectory()
+    s3 = boto.connect_s3()
+    try:
+        # Does not work if there is a '.' in the name
+        bucket = s3.get_bucket(bucket_name=target_bucket)
+    except boto.exception.S3ResponseError:
+        bucket = s3.create_bucket(bucket_name=target_bucket)
+
+    # Download existing repository
+    for item in bucket.list(prefix=target_key):
```
tomprince: Given the comments below, we perhaps only want to download RPMs here. If we do download metadata, we should keep track of it, so we don't need to upload it again later.

tomprince:
Since I propose there being some logic in what to upload and download, I think there should be some testing of this (probably using effect).

admin/release.py
```diff
@@ -303,3 +319,178 @@ def publish_docs_main(args, base_path, top_level):
...
                          % (base_path.basename(),))
         raise SystemExit(1)
+
+
+class UploadOptions(Options):
+    """
+    Options for uploading packages.
+    """
+    optParameters = [
+        ["target", None, b'clusterhq-yum-repository',
```
tomprince: I think this should be something more generic. I think we will be uploading DEBs here as well, and perhaps also python sdists and/or wheels.

admin/release.py
adamtheturtle: If this is just a matter of efficiency, I think that it is best left as a follow-up in https://clusterhq.atlassian.net/browse/FLOC-1506. This code does not yet need to be optimised for performance. Or is there something else I'm missing?

adamtheturtle:
> The fedora-packages name is out-dated, although I'm not sure if it is worth changing. On the other hand, perhaps the still relevant files should just be moved here.

How can one tell what is still a relevant file? I can't think of any downsides to moving the files here, can you?

> Since I propose there being some logic in what to upload and download, I think there should be some testing of this (probably using effect).

As per my comment at https://github.com/ClusterHQ/flocker/pull/1190#discussion_r25684526 this seems like it would be a set of tests for optimisations, which can be, in my opinion, left for https://clusterhq.atlassian.net/browse/FLOC-1506. Do you agree? If so, I can address the remaining comments (awaiting feedback about "fedora-packages") and resubmit.

admin/release.py
adamtheturtle: I've gone with `clusterhq-archive`.

tomprince:
> How can one tell what is still a relevant file?

Just the ones needed for this branch.

admin/release.py
tomprince: I don't think it is significantly more complex or time consuming to do the more efficient thing, so I don't think it is worth not doing now. That is, even if it takes a bit of time, we should do it now.

admin/release.py
```diff
@@ -303,3 +322,180 @@ def publish_docs_main(args, base_path, top_level):
...
+    :param list packages: List of bytes, each specifying the name of a package
+        to upload to the repository.
+    """
+    rpm_directory.createDirectory()
+    s3 = boto.connect_s3()
+    try:
+        # Does not work if there is a '.' in the name
+        bucket = s3.get_bucket(bucket_name=target_bucket)
+    except boto.exception.S3ResponseError:
+        bucket = s3.create_bucket(bucket_name=target_bucket)
```
tomprince: I think assuming the bucket exists is a reasonable choice.

adamtheturtle:
> Although, we only want to upload
> 1. New RPMs.
> 2. New yum metadata: repodata/repomod.xml and the files referenced there. (It probably doesn't make sense to parse repomod.xml, just grab the new files in repodata.

@tomprince Are we concerned about RPMs which have been changed but have the same filename? Or is looking for new keys enough? Same question for the yum metadata (apart from the `xml` file which we know is updated).

If we are concerned, any advice on going about this? Hashing files etc.?

adamtheturtle:
This has got big enough that I'd like it to be looked at before I continue.

As mentioned above I'm not quite sure about the desired logic of what to upload when updating the repository.
However, the code is in such a state that it should be trivial to e.g. only upload packages if a package with that name is already in S3, which I think is desired but won't do until confirmed.

Something I don't quite understand is `archive.clusterhq.com`, how it works, why it works and how the changes which are made here (to the `.spec` and `.repo` files).

Other nice-to-haves I'd like feedback on (whether they should be done here or in a follow-up or neither):
* Error handling. Particularly for the case where the build server does not have all of the desired packages.
* Logging. This code logs nothing.
* `admin.test.test_release.UploadRPMsTests.test_real_yum_utils` is skipped on buildbot and needs `createrepo` tools to run.
* Deletion of metadata files which are no longer necessary; that is, metadata files which do not exist locally after `createrepo` is called but exist on S3.
* Drop characteristic for pyrsistent

If this is accepted what needs to happen:
* Create a branch for https://github.com/ClusterHQ/fedora-packages - to change the README to say that it is a legacy repository and point to the Flocker docs for packaging.
* See the tutorial box test work - the 0.3.2 tag should be built before running `export VERSION=0.3.2; admin/upload-rpms "${VERSION}"`. Unfortunately buildbot is failing to build the packages for this version because of a missing file. http://build.clusterhq.com/builders/flocker-omnibus-fedora-20/builds/1764/steps/build-package/logs/stdio.

admin/release-packaging/README.rst
```diff
@@ -0,0 +1,3 @@
+Packages for Flocker that aren't available in each supported Operating System.
```
tomprince: This should probably be something about packages that point at our repositories, instead.

admin/release-packaging/centos/development/clusterhq-release.spec
```diff
@@ -0,0 +1,44 @@
+# See README for instructions on how to build.
```
tomprince: :question: Maybe don't have two layers of indirection.

admin/release-packaging/centos/development/clusterhq.repo
```diff
@@ -0,0 +1,5 @@
+[clusterhq]
```
tomprince: We shouldn't have both a development and a marketing `clusterhq-release` package. Instead, we should just have a single package that has two repositories defined, one enabled and the other disabled by default.

admin/release.py
```diff
@@ -101,7 +120,14 @@ class NotTagged(Exception):
 
 class NotARelease(Exception):
     """
-    Raised if trying to publish to a version that isn't a release.
+    Raised if trying to publish to a version that isn't a release or upload
```
tomprince: :question: Both docs and RPMs are being published.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+        return 'file://' + source_repo.path
+
+    def setUp(self):
+        self.scratch_directory = FilePath(tempfile.mkdtemp())
+        self.addCleanup(self.scratch_directory.remove)
+        self.rpm_directory = self.scratch_directory.child(
+            b'distro-version-arch')
+        self.target_key = 'test/target/key'
+        self.target_bucket = 'test-target-bucket'
+        self.build_server = 'http://test-build-server.com'
```
tomprince: `test-build-server.example`: We should use a domain name that is reserved for examples/tests.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+        self.packages = ['clusterhq-flocker-cli', 'clusterhq-flocker-node']
```
tomprince: `clusterhq-python-flocker`?

admin/release.py
```diff
@@ -303,3 +329,176 @@ def publish_docs_main(args, base_path, top_level):
...
+    optParameters = [
+        ["target", None, ARCHIVE_BUCKET,
+         "The bucket to upload packages to."],
+        ["build-server", None,
+         b'http://build.clusterhq.com',
+         "The URL of the build-server."],
+    ]
+
+    def parseArgs(self, version):
+        self['version'] = version
```
tomprince: This throws me every time. Is there any reason this isn't consistent with the `publish-docs` options?

admin/release.py
```diff
@@ -303,3 +329,176 @@ def publish_docs_main(args, base_path, top_level):
...
+        target_path=rpm_directory,
+        filter_extensions=('.rpm',)))
+
+    yield Effect(DownloadPackagesFromRepository(
+        source_repo=source_repo,
+        target_path=rpm_directory,
+        packages=packages,
+        ))
+
+    downloaded_packages = yield Effect(ListPackages(
```
tomprince: This will get all the packages, both the ones that already exists, as well as the new ones we need to add.

Since we do know the version of the packages we are downloading,
- we could just download them directly, rather than have to use yum
- we can easily generate the list of download packages.

That would also alleviate the issue the packages being downloaded will be the exact version we are expecting (although, it shouldn't be possible to build from a tag and get a different version).

admin/release.py
```diff
@@ -303,3 +329,176 @@ def publish_docs_main(args, base_path, top_level):
...
+
+    yield Effect(CreateRepo(
+        repository_path=rpm_directory,
+        ))
+
+    new_metadata = yield Effect(ListMetadata(
+        repository_path=rpm_directory,
+    ))
+
+    changed_metadata = new_metadata - old_metadata
```
tomprince: This dance seems like it would be reasonable to put in `CreateRepo`? It could just walk `repodata` before and after updating, and return the list of new files along with `repomod.xml`.

vagrant/tutorial/bootstrap.py
```diff
@@ -24,11 +24,12 @@
     'fedora/zfs-release%s.noarch.rpm') % (rpm_dist,)
 check_call(['yum', 'install', '-y',  zfs_repo_url])
 
-clusterhq_repo_url = (
-    'https://storage.googleapis.com/'
-    'archive.clusterhq.com/'
+clusterhq_marketing_repo_url = (
```
tomprince: Given the above comments about the `clusterhq-release` package, we don't need to specify that this is the /marketing/ repo url.

flocker/provision/_install.py
```diff
@@ -11,12 +11,27 @@
...
 
 ZFS_REPO = ("https://s3.amazonaws.com/archive.zfsonlinux.org/"
             "fedora/zfs-release$(rpm -E %dist).noarch.rpm")
-CLUSTERHQ_REPO = ("https://storage.googleapis.com/archive.clusterhq.com/"
-                  "fedora/clusterhq-release$(rpm -E %dist).noarch.rpm")
+
+if is_release(get_doc_version(flocker_version)):
+    key = 'marketing'
+else:
+    key = 'development'
```
tomprince: We don't need this either (but if we did, we should use the version being installed, instead of the version that is running things.

docs/gettinginvolved/infrastructure/packaging.rst
```diff
@@ -60,3 +60,44 @@ The commands above require the ``rpmdevtools`` and ``yum-utils`` packages instal
...
+To build and upload these packages, on a machine with the operating system which the package is for
+(an easy way to do this is to use the :doc:`Flocker development machine <vagrant>` and an old Fedora 20 version of it),
+set up `gsutil` with S3 credentials,
+go to the relevant directory in `admin/release-packaging` and run:
+
+.. code-block:: sh
+
+   # Modify the below to reference the S3 key which should hold the package
+   # using the release type and distro name.
+   export S3KEY=marketing/centos
```
tomprince: Release type isn't needed.

docs/gettinginvolved/infrastructure/packaging.rst
```diff
@@ -60,3 +60,44 @@ The commands above require the ``rpmdevtools`` and ``yum-utils`` packages instal
 Flocker depends on a number of packages which aren't available in fedora.
 These packages are available from `our Copr repository <https://copr.fedoraproject.org/coprs/tomprince/hybridlogic/>`_.
 To enable yum to find them, put the `repo file <https://copr.fedoraproject.org/coprs/tomprince/hybridlogic/repo/fedora-20-x86_64/tomprince-hybridlogic-fedora-20-x86_64.repo>`_ in :file:`/etc/yum.repos.d/`.
+
+
+Package Hosting
+===============
+
+Some packages are hosted on Google Cloud Storage and others are hosted on Amazon S3.
```
tomprince: Maybe just say, hosted on S3, with a note about legacy versions.

admin/yum.py
```diff
@@ -0,0 +1,192 @@
...
+        b'clean',
+        b'metadata'])
+
+    check_call([
+        b'yumdownloader',
+        b'--config', yum_repo_config.path,
+        b'--disablerepo=*',
+        b'--enablerepo=flocker',
+        b'--quiet',
+        b'--destdir', intent.target_path.path] + intent.packages)
```
tomprince: As mentioned elsewhere, we should probably just explicitly download the files we care about.

admin/upload-rpms
```diff
@@ -1,125 +1,13 @@
 #!/usr/bin/env python
```
tomprince: :question: I'm tempted to suggest this should actually be called `publish-rpms`.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            target_bucket=self.target_bucket,
+            target_key=self.target_key,
+            source_repo=repo_uri,
+            packages=self.packages,
+        )
+
+        files = [
+            'clusterhq-flocker-cli-0.3.3-0.dev.7.noarch.rpm',
+            'clusterhq-flocker-node-0.3.3-0.dev.7.noarch.rpm',
+            'repodata/1988e623dfb7204ec981d2a2ab1a38da6c0f742717d184088dd0a0f344f6e89c-filelists.sqlite.bz2',  # noqa
```
tomprince: I guess the fact that these files match those in the test-repo is just because the repos are identical, rather than that they get copied from one to another. Perhaps a comment would be useful here.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            target_bucket=self.alternative_bucket,
+            target_key=self.target_key,
+            source_repo=self.create_fake_repository(files=repo_contents),
+            packages=self.packages,
+        )
+
+        index_path = os.path.join(self.target_key, 'repodata', 'repomd.xml')
+        self.assertNotEqual(
+            aws.s3_buckets[self.target_bucket][index_path],
+            aws.s3_buckets[self.alternative_bucket][index_path])
```
tomprince: I'm not sure I understand what this test is asserting or trying to test.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            target_bucket=self.target_bucket,
+            target_key=self.target_key,
+            source_repo=self.create_fake_repository(files=repo_contents),
+            packages=self.packages,
+        )
+
+        expected_keys = existing_s3_keys.copy()
+        expected_keys.update({
+            os.path.join(self.target_key, package): repo_contents[package]
+            for package in repo_contents})
```
tomprince: This seems superfluous; that is, the new packages being there are already covered by other tests.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            rpm_directory=self.rpm_directory,
+            target_bucket=self.target_bucket,
+            target_key=self.target_key,
+            source_repo=self.create_fake_repository(files={}),
+            packages=self.packages,
+        )
+
+        self.assertEqual(
+            aws.s3_buckets[self.target_bucket][existing_metadata_file],
+            existing_s3_keys[existing_metadata_file])
```
tomprince: This assertion doesn't make sense to me. Maybe it would be better if the fake `CreateRepository` actually included (fake) hashes in the names of the files, and this checked that the old ones are still there with the old data.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            source_repo=self.create_fake_repository(files={}),
+            packages=self.packages,
+        )
+
+        repodata_files = [
+            key for key in aws.s3_buckets[self.target_bucket] if
+            key.startswith(os.path.join(self.target_key, 'repodata'))]
+
+        # What matters is that there is more than just the index.
+        self.assertGreater(len(repodata_files), 1)
```
tomprince: We should check at least one of the metadata files (although, it would probably be enough to have one fake one (but with a hash in the filename).

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+            target_bucket=self.alternative_bucket,
+            target_key=self.target_key,
+            source_repo=self.create_fake_repository(files={}),
+            packages=self.packages,
+        )
+
+        index_path = os.path.join(self.target_key, 'repodata', 'repomd.xml')
+        self.assertNotEqual(
+            aws.s3_buckets[self.target_bucket][index_path],
+            aws.s3_buckets[self.alternative_bucket][index_path])
```
tomprince: I'm not sure I understand what this is asserting.

admin/release.py
adamtheturtle: I think I might be under the wrong impression. Did we not need all of the packages which exist in order to create a valid `repodata.xml`?

admin/test/test_release.py
adamtheturtle: I'll put a note here. I didn't want to include that in the repository because it is significantly bigger than the other binaries. It doesn't seem necessary to have all the right package names here. If you think that it is, I will change this and then use the subset `['clusterhq-flocker-cli', 'clusterhq-flocker-node']` for the test which uses the real yum tools.

docs/gettinginvolved/infrastructure/packaging.rst
adamtheturtle: Below this there is a mention that OS X uses GCS. Later there is a "Legacy" section.

flocker/provision/_install.py
adamtheturtle: To clarify: Do you mean that we should not separate marketing and testing RPMs in S3? If so, would you prefer that I remove this feature?

adamtheturtle:
The CentOS packages don't work :disappointed: 

```
$ sudo yum install -v -y https://s3.amazonaws.com/clusterhq-archive/development/centos/clusterhq-release.el7.centos.noarch.rpm
Loading "fastestmirror" plugin
Config time: 0.006
Yum version: 3.4.3
rpmdb time: 0.000

Cannot open: https://s3.amazonaws.com/clusterhq-archive/development/centos/clusterhq-release.el7.centos.noarch.rpm. Skipping.
```

admin/release-packaging/centos/development/clusterhq-release.spec
adamtheturtle: I chose to do this because I thought that this will continue to have a README but the packaging file might get moved (so there is only one reference to change).

admin/release-packaging/README.rst
adamtheturtle: Please could you suggest an ideal comment - I'm not 100% clear on what this does.

I think "Packages which point to repositories where Flocker packages are stored".

admin/release.py
tomprince: We need to download them, but we don't need to upload them again (which is what this variable gets used for).

admin/test/test_release.py
tomprince: That seems reasonable.

flocker/provision/_install.py
tomprince: We should have separate repositories. But we should only have a single `clusterhq-release` package (that has configuration for both of them). We probably won't have any extra packages that matter for acceptance testing in the development repo (if we do, the work in #1181 can perhaps be used to support it).

That is, we will upload the RPMs to separate repositories, but since we aren't installing those RPMs anyway, we don't need to care about the distinction in `flocker.provision`.

tomprince:
> Ah that is an S3 permissions denied problem

I guess we should probably have a bucket permission thing (like for the doc buckets; see the documentation for details).

admin/upload-rpms
adamtheturtle: Agreed.

tomprince: Or even, `publish-packages`.

admin/yum.py
adamtheturtle: Thinking about this there should probably be an Exception raised if ≥1 of those files are not available. Do you agree?

admin/release.py
adamtheturtle: No, I will change it.

admin/yum.py
tomprince: Yes.

admin/yum.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+@sync_performer
+def perform_download_packages_from_repository(dispatcher, intent):
+    """
+    See :class:`DownloadPackagesFromRepository`.
+    """
+    # TODO don't use yum tools. Just download files.
+    import urllib2
+    import urllib
```
tomprince: I'm a little sad to suggest another dependency, but I think I should say, just use http://docs.python-requests.org/en/latest/

admin/yum.py
```diff
@@ -0,0 +1,176 @@
...
+    packages = ['clusterhq-flocker-cli-0.3.3-0.dev.7.noarch.rpm',
+                'clusterhq-flocker-node-0.3.3-0.dev.7.noarch.rpm']
+    # TODO use intent.packages, but get versioned files
```
tomprince: There is a function in `admin.packaging` that gets the right filename.

admin/release.py
adamtheturtle: I think that it can be simpler than that. Because the existing metadata is not downloaded, just listed, walking `repodata` beforehand will always be empty.

admin/test/test_release.py
```diff
@@ -661,3 +667,637 @@ def test_publish_non_release_fails(self):
...
+                    'development',
+                    operating_system['distro'],
+                    operating_system['version'],
+                    operating_system['arch'],
+                    file,
+                )
+                expected_files.add(path)
+
+        files_on_s3 = aws.s3_buckets[self.target_bucket].keys()
+        self.assertTrue(expected_files.issubset(set(files_on_s3)))
```
adamtheturtle: Tests with "issubset" fail just telling you that False does not equal True. Mitigation options for this problem include:

* Put a descriptive message in
* Compare sorted lists instead of sets

Any preference / suggestion @tomprince ?

adamtheturtle:
I think that I have addressed all the review comments (some by adding comments to existing tests) and so I am submitting for another review. Sorry if I've missed something already mentioned.

There doesn't seem (to me) to be a need to put hashes in fake metadata filenames. The metadata files are distinct in that they change when differently named RPMs are in the repository.

@tomprince Thank you for the review and also the work in simplifying the repositories. Is this work done (as in the repo and spec files are all that is needed and nothing needs to be disabled / to document how to enable any repository)?

There are outstanding questions at https://github.com/ClusterHQ/flocker/pull/1190#issuecomment-77866449 about what follow-up work to do now and later.

I'd like help with choosing the right logging and permissions settings for the S3 bucket before setting them up and documenting them.

admin/aws.py
```diff
@@ -187,18 +189,158 @@ def perform_list_s3_keys(dispatcher, intent):
...
+
+
+@sync_performer
+@do
+def perform_upload_s3_key_recursively(dispatcher, intent):
+    """
+    See :class:`UploadToS3Recursively`.
+    """
+    for path in intent.source_path.walk():
+        if path.basename() in intent.files:
```
tomprince: If you have the list of files, you don't actually need to walk the tree. In any case, `files` should be a list of paths relative to `source_path`, so this won't properly grab files in subdirectories.

admin/packaging.py
```diff
@@ -53,6 +53,11 @@ class PackageTypes(Values):
     },
 }
 
+ARCH_REQUIREMENT = {
```
tomprince: Maybe `PACKAGE_ARCHITECTURE`?

admin/yum.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+    List the filenames of new and changed repository metadata files.
+
+    :param FilePath repository_path: Location of repository to list repository
+        metadata from.
+    :param set existing_metadata: Filenames of existing metadata files.
+    """
+    all_metadata = set([os.path.basename(path.path) for path in
+                        repository_path.child('repodata').walk()])
+    new_metadata = all_metadata - existing_metadata
```
tomprince: Since
- we don't download old metadata
- metadata files are small
- we are unlikely to generate exactly the same metadata twice

we probably don't need to bother excluding existing metadata here.
If we were going to do anything like this, it would probably make sense to list pre-existing *local* metadata, and exclude that (this could be done as part of this effect).

admin/yum.py
```diff
@@ -0,0 +1,175 @@
...
+def _list_new_metadata(repository_path, existing_metadata):
+    """
+    List the filenames of new and changed repository metadata files.
+
+    :param FilePath repository_path: Location of repository to list repository
+        metadata from.
+    :param set existing_metadata: Filenames of existing metadata files.
+    """
+    all_metadata = set([os.path.basename(path.path) for path in
+                        repository_path.child('repodata').walk()])
```
tomprince: As mentioned elsewhere, this should probably return a list of relative paths; `segementsFrom` is probably useful in that regard.

Incidentally, `os.path.basename(path.path)` can be spelled `path.basename()`.

admin/release.py
```diff
@@ -303,3 +327,179 @@ def publish_docs_main(args, base_path, top_level):
...
+
+    :param FilePath scratch_directory: Temporary directory to download
+        repository to.
+    :param bytes target_bucket: S3 bucket to upload repository to.
+    :param bytes version: Version to download RPMs for.
+    :param bytes build_server: Server to download new RPMs from.
+    """
+    if not (is_release(version)
+            or is_weekly_release(version)):
+        raise NotARelease
```
tomprince: Raise an object, not a class (i.e. `NotARelease()`).

admin/release.py
```diff
@@ -303,3 +327,179 @@ def publish_docs_main(args, base_path, top_level):
...
+        source_repo=os.path.join(build_server, b'results/omnibus', version,
+                                 b'fedora-20'),
+        packages=FLOCKER_PACKAGES,
+        flocker_version=version,
+        distro_name='fedora',
+        distro_version='20',
+    )
+
+    yield update_repo(
+        rpm_directory=scratch_directory.child(b'centos-7-x86_64'),
```
tomprince: This seems like we could loop over the list `[('fedora', '20', 'x86_64'), ('centos', '7', 'x86_64')]`.

admin/yum.py
```diff
@@ -0,0 +1,175 @@
...
+    """
+
+
+@sync_performer
+def perform_download_packages_from_repository(dispatcher, intent):
+    """
+    See :class:`DownloadPackagesFromRepository`.
+    """
+    # TODO avoid circular imports when importing these somewhere else
+    # Reviewer - before I do, this might be easier if the logic to get
```
tomprince: https://clusterhq.atlassian.net/browse/FLOC-1223 suggests moving into `flocker`, so probably just link to that here.

setup.py
```diff
@@ -71,6 +71,10 @@
     # The cloud acceptance test runner needs these
     "fabric==1.10.0",
     "apache-libcloud==0.16.0",
+
+    # Packages are downloaded from Buildbot
+    "requests==2.5.3",
```
tomprince: We also need this as part of the release process.

tomprince:
> There doesn't seem (to me) to be a need to put hashes in fake metadata filenames. The metadata files are distinct in that they change when differently named RPMs are in the repository.

All the metadata files (except `repodata.xml`) are named with hashes. The reason for this is so that if somebody is downloading data while new metadata is being uploaded, the old files are still around and the downloader can complete the update. We should test that this actually happens. The fake `CreateRepository` currently has fixed names, so it can't be used to test this.

admin/test/test_release.py
```diff
@@ -661,3 +669,704 @@ def test_publish_non_release_fails(self):
...
+            packages=self.packages,
+            flocker_version=self.dev_version,
+            distro_name=self.operating_systems[0]['distro'],
+            distro_version=self.operating_systems[0]['version'],
+        )
+
+        # This tests that filelists.xml.gz, a metadata file which is not the
+        # index, still has the content 'old_metadata_content', even though
+        # CreateRepo changes this content. This shows that this file which
+        # already existed on S3 is not re-uploaded.
```
tomprince: I think this isn't something we care about (uploading a metadata file twice isn't a big issue), but we *don't* want to overwrite existing differently named (due to hashses) metadata.

admin/test/test_release.py
```diff
@@ -661,3 +669,704 @@ def test_publish_non_release_fails(self):
...
+        )
+
+        repodata_files = [
+            key for key in aws.s3_buckets[self.target_bucket] if
+            key.startswith(os.path.join(self.target_key, 'repodata'))]
+
+        # What matters is that there is more than just the index.
+        # Other tests cover the creation and modification of particular
+        # metadata filenames, but these change because of `createrepo` options
+        # such as whether to use hashes in filenames or not.
```
tomprince: We are using a fake CreateRepository, so we know the names of all the metadata files that are created.

admin/test/test_release.py
```diff
@@ -661,3 +669,704 @@ def test_publish_non_release_fails(self):
...
+        cli_package = 'clusterhq-flocker-cli-0.3.3-0.dev.7.noarch.rpm'
+        existing_s3_keys = {
+            os.path.join(self.target_key, cli_package): 'old-cli-package',
+        }
+
+        aws = FakeAWS(
+            routing_rules={},
+            s3_buckets={
+                self.target_bucket: {},
+                self.alternative_bucket: existing_s3_keys.copy(),
```
tomprince: I'm not sure what this extra bucket is for.

admin/test/test_release.py
tomprince: Or why we care that the index of `target_bucket` is different that `alternative_bucket`.

admin/yum.py
```diff
@@ -0,0 +1,175 @@
...
+    @sync_performer
+    def _perform_create_repository(self, dispatcher, intent):
+        """
+        See :class:`CreateRepo`.
+        """
+        metadata_directory = intent.repository_path.child('repodata')
+        metadata_directory.createDirectory()
+        packages = set([
+            os.path.basename(path.path) for path in
+            intent.repository_path.walk() if path.isfile()])
```
tomprince: `repository_path.listdir() if path.endswith('rpm')`?

admin/test/test_release.py
```diff
@@ -661,3 +669,704 @@ def test_publish_non_release_fails(self):
...
+                source_repo=self.create_fake_repository(
+                    files=self.repo_contents),
+                packages=self.packages,
+                flocker_version=self.dev_version,
+                distro_name=self.operating_systems[0]['distro'],
+                distro_version=self.operating_systems[0]['version'],
+            )
+
+        self.assertEqual(
+            u'404 Client Error: None',
```
tomprince: `requests` probably provides the code as an attribute. We should check that.

admin/release.py
```diff
@@ -303,3 +327,179 @@ def publish_docs_main(args, base_path, top_level):
...
+        target_path=rpm_directory,
+        packages=packages,
+        flocker_version=flocker_version,
+        distro_name=distro_name,
+        distro_version=distro_version,
+        ))
+
+    existing_metadata = yield Effect(
+        ListS3Keys(bucket=target_bucket,
+                   prefix=os.path.join(target_key, 'repodata/')))
```
tomprince: As mentioned elsewhere, this shouldn't be necessary.

admin/aws.py
tomprince: Which perhaps suggests that the intent should be `UploadKeysToS3` and the following one `UploadKeyToS3`.

tomprince:
Other than the above comments, this looks good. After addressing them, if buildbot is green, and you've tested that the tool works, feel free to merge. (Although a second pair of eyes wouldn't hurt).

admin/yum.py
adamtheturtle: Unless I'm misunderstanding I think that it is not necessary to list pre-existing local metadata because we have a fresh scratch directory with downloaded packages and nothing else there.

adamtheturtle:
http://build.clusterhq.com/builders/flocker-omnibus-centos-7/builds/1752/steps/build-repo-metadata/logs/stdio shows ``Critical: Temporary repodata directory: ./.repodata/ already exists! (Another createrepo process is running?)``. I guess that this might be because `admin.test.test_release.UploadRPMsTests.test_real_yum_utils` uses `createrepo`. I'll look into this because I think that it is a useful test.

adamtheturtle:
The above did not happen in the latest builds. I'm not sure if this is a related failure which would cause intermittent failures on `master` if this is merged, or unrelated.

adamtheturtle:
I *think* that all of the review comments have been addressed. This leaves:
* Maybe making some tests pass, we'll see if they don't
* Manual testing
* Finding out and documenting what has to be done differently, and when, to enable the -testing repository when one wants to install a non-marketing release.

adamtheturtle:
For manual testing I've done the following:

```
export VERSION=0.3.3dev7
admin/publish-packages --flocker-version "${VERSION}"
sudo yum install -y https://s3.amazonaws.com/clusterhq-archive/centos/clusterhq-release.el7.centos.noarch.rpm
sudo yum --disablerepo=clusterhq --enablerepo=clusterhq-testing clusterhq-flocker-cli
flocker-deploy --version
# outputs: 0.3.2+doc1-2381-gdb5fed8
```

adamtheturtle:
I created and merged https://github.com/ClusterHQ/fedora-packages/pull/3 - changing the fedora-packages README to say that it is deprecated.

PR #1191, opened 2015-03-02
[FLOC-1463] Initial stub script for Flocker CLI

dwgebler:
Fixes FLOC-1463. See https://clusterhq.atlassian.net/browse/FLOC-1463?focusedCommentId=16380&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16380

### DIFF
--- a/flocker/cli/script.py
+++ b/flocker/cli/script.py
@@ -6,7 +6,7 @@ The command-line ``flocker-deploy`` tool.
 
 from subprocess import CalledProcessError
 
-from twisted.internet.defer import DeferredList
+from twisted.internet.defer import DeferredList, succeed
 from twisted.internet.threads import deferToThread
 from twisted.python.filepath import FilePath
 from twisted.python.usage import Options, UsageError
@@ -250,9 +250,48 @@ class DeployScript(object):
         return DeferredList(results)
 
 
+@flocker_standard_options
+class CLIOptions(Options):
+    """
+    Command line options for ``flocker`` CLI.
+
+    """
+    longdesc = """flocker does nothing (yet).
+
+    """
+
+    synopsis = ("Usage: flocker [OPTIONS] "
+                "\n"
+                "If you have any issues or feedback, you can talk to us: "
+                "http://docs.clusterhq.com/en/latest/gettinginvolved/"
+                "contributing.html#talk-to-us")
+
+
+@implementer(ICommandLineScript)
+class CLIScript(object):
+    """
+    A command-line script to interact with a cluster via the API.
+    """
+    def main(self, reactor, options):
+        """
+        See :py:meth:`ICommandLineScript.main` for parameter documentation.
+
+        :return: A ``Deferred`` which fires when the deployment is complete or
+                 has encountered an error.
+        """
+        return succeed(None)
+
+
 def flocker_deploy_main():
     return FlockerScriptRunner(
         script=DeployScript(),
         options=DeployOptions(),
+        logging=False,
+    ).main()
+
+def flocker_cli_main():
+    return FlockerScriptRunner(
+        script=CLIScript(),
+        options=CLIOptions(),
         logging=False,
     ).main()
--- a/flocker/cli/script.py
+++ b/flocker/cli/script.py
@@ -28,6 +28,13 @@ from ..common import ProcessNode, gather_deferreds
 from ._sshconfig import DEFAULT_SSH_DIRECTORY, OpenSSHConfiguration
 
 
+FEEDBACK_CLI_TEXT = (
+    "\n"
+    "If you have any issues or feedback, you can talk to us: "
+    "http://docs.clusterhq.com/en/latest/gettinginvolved/"
+    "contributing.html#talk-to-us")
+
+
 @attributes(['node', 'hostname'])
 class NodeTarget(object):
     """
@@ -48,10 +55,7 @@ class DeployOptions(Options):
 
     synopsis = ("Usage: flocker-deploy [OPTIONS] "
                 "DEPLOYMENT_CONFIGURATION_PATH APPLICATION_CONFIGURATION_PATH"
-                "\n"
-                "If you have any issues or feedback, you can talk to us: "
-                "http://docs.clusterhq.com/en/latest/gettinginvolved/"
-                "contributing.html#talk-to-us")
+                "{feedback}").format(feedback=FEEDBACK_CLI_TEXT)
 
     def parseArgs(self, deployment_config, application_config):
         deployment_config = FilePath(deployment_config)
@@ -254,17 +258,11 @@ class DeployScript(object):
 class CLIOptions(Options):
     """
     Command line options for ``flocker`` CLI.
-
     """
-    longdesc = """flocker does nothing (yet).
+    longdesc = """flocker does nothing (yet)."""
 
-    """
-
-    synopsis = ("Usage: flocker [OPTIONS] "
-                "\n"
-                "If you have any issues or feedback, you can talk to us: "
-                "http://docs.clusterhq.com/en/latest/gettinginvolved/"
-                "contributing.html#talk-to-us")
+    synopsis = "Usage: flocker [OPTIONS] {feedback}".format(
+        feedback=FEEDBACK_CLI_TEXT)
 
 
 @implementer(ICommandLineScript)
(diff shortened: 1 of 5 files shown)


adamtheturtle:
This design does not address the part of the issue description "we need to be able to create packages for this script that do not drag in all of the server-side dependencies so that the script is easily installable on systems that cannot support the server-side software (eg OS X).". If that is a result of discussion and not oversight, please note here the summary/conclusions of discussions to change the spec of this. Else, one suggestion I have is to separate the `dev` dependencies into `dev` and `dev-linux-only` `extras_require` sections and install those extras only on the appropriate builders.

adamtheturtle:
Sorry if this is me being slow today, but I'm failing to work out what here is "design" and what is being submitted as finished code.

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
...
+    Command line options for ``flocker`` CLI.
+
+    """
+    longdesc = """flocker does nothing (yet).
+
+    """
+
+    synopsis = ("Usage: flocker [OPTIONS] "
+                "\n"
+                "If you have any issues or feedback, you can talk to us: "
```
adamtheturtle: :question: This feedback clause is repeated elsewhere and it is best that those remain consistent. Therefore I recommend using a variable.

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
...
+@implementer(ICommandLineScript)
+class CLIScript(object):
+    """
+    A command-line script to interact with a cluster via the API.
+    """
+    def main(self, reactor, options):
+        """
+        See :py:meth:`ICommandLineScript.main` for parameter documentation.
+
+        :return: A ``Deferred`` which fires when the deployment is complete or
```
adamtheturtle: This comment is the same as in `DeployScript.main`. Is this `CLIScript` going to be only for deployments? If so, perhaps it should take the place (and name?) of `flocker-deploy`. Else this comment should be changed. It is also at odds with the `longdesc` "flocker does nothing (yet).". I'm not quite sure what the best-practice is for a stub: should the description be honest and say that it does nothing, or say what the plan is? Either way the comments and docstrings should be consistent in this.

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
...
         options=DeployOptions(),
         logging=False,
     ).main()
+
+
+def flocker_cli_main():
+    return FlockerScriptRunner(
+        script=CLIScript(),
+        options=CLIOptions(),
+        logging=False,
```
adamtheturtle: My understanding is that we are taking the approach now of the more logging the better. Please either add basic logging or link to a follow-up issue to do so.

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
         return DeferredList(results)
 
 
+@flocker_standard_options
+class CLIOptions(Options):
+    """
+    Command line options for ``flocker`` CLI.
+
```
adamtheturtle: Extra blank line

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
...
+    """
+    longdesc = """flocker does nothing (yet).
+
```
adamtheturtle: Extra blank line not there in the other `longdesc`s I've just looked at.

flocker/cli/script.py
```diff
@@ -250,9 +250,49 @@ def _changestate_on_nodes(self, deployment, deployment_config,
...
+                "\n"
+                "If you have any issues or feedback, you can talk to us: "
+                "http://docs.clusterhq.com/en/latest/gettinginvolved/"
+                "contributing.html#talk-to-us")
+
+
+@implementer(ICommandLineScript)
+class CLIScript(object):
+    """
+    A command-line script to interact with a cluster via the API.
```
adamtheturtle: :question: It seems to be that the key differentiator of this script to the others (apart from `flocker-deploy` which this will replace) is that it is user facing, while the others are run either automatically or by the developers of `Flocker`. This might be a useful thing to document.

dwgebler:
> Sorry if this is me being slow today, but I'm failing to work out what here is "design" and what is being submitted as finished code.

What's here is "finished code" to the end that it's a working, installable, do-nothing, stub script, but it's a design in that are a few more things to be added or polished before we'd want to merge this, a few of which you've highlighted in other comments (things like the script's long description, logging, etc.) - for some of those though, the ticket isn't clear enough on what's intended to be included at this phase and what will be done in related follow-up issues, which I'm intending to clear up in the engineering meeting today.

Current indications from the last engineering meeting are that for at least the scope of this issue, you don't need to be able to develop on OS X.

flocker/cli/script.py
dwgebler: If we're going to keep as-is for the new CLI, absolutely, but I don't know if we're keeping this text the same or not yet, another thing I need to ask about in an upcoming engineering meeting.

setup.py
```diff
@@ -126,6 +126,7 @@
             'flocker-reportstate = flocker.node.script:flocker_reportstate_main',
             'flocker-zfs-agent = flocker.node.script:flocker_zfs_agent_main',
             'flocker-control = flocker.control.script:flocker_control_main',
+            'flocker = flocker.cli.script:flocker_cli_main',
```
adamtheturtle: This section says:

>         # Don't forget to modify the omnibus packaging tool
        # (admin/packaging.py) if you make changes here.

adamtheturtle: I'm guessing that will be a follow up issue, please put an "XXX" and a link to that here if so.

flocker/cli/script.py
dwgebler: This is something that's eventually replacing `flocker-deploy` (as far as I understand the plan), but will also be developed while we're still keeping the old deploy script. I think this comment string remains accurate so far as the intention of what the new CLI will be doing, just how it's doing deployments will be different (interacting with an API rather than YAML over SSH). So the comment string may need to change, but again I also don't know yet. For the moment, as a stub script, a "successful deployment" is actually just returning a succeeded deferred.

adamtheturtle:
As the comment replies seem to indicate that the scope / intention of this will be determined in a later meeting I will mark this as rejected. However, some of the comments I have made should be applicable no matter what so those can be done before that meeting if you like.

setup.py
dwgebler: This is included in this branch now.

dwgebler:
Please note: This is now in JIRA in a design review ready state, however I don't think there's much else (if anything) that can be done to fulfil the implementation. The original reason for the design workflow rather than going straight to coding is that when the ticket was first created, there was an intention that this script would not be merely another entrypoint in our existing package. That decision was reversed in an engineering meeting.

flocker/cli/script.py
```diff
@@ -28,6 +28,13 @@
 from ._sshconfig import DEFAULT_SSH_DIRECTORY, OpenSSHConfiguration
 
 
+FEEDBACK_CLI_TEXT = (
+    "\n"
+    "If you have any issues or feedback, you can talk to us: "
+    "http://docs.clusterhq.com/en/latest/gettinginvolved/"
```
itamarst: Should be `https`.

flocker/cli/script.py
```diff
@@ -250,9 +254,44 @@ def _changestate_on_nodes(self, deployment, deployment_config,
         return DeferredList(results)
 
 
+@flocker_standard_options
+class CLIOptions(Options):
+    """
+    Command line options for ``flocker`` CLI.
+    """
+    longdesc = """flocker does nothing (yet)."""
```
itamarst: Maybe add reference to `flocker-deploy` as alternative until this actually does stuff.

flocker/cli/script.py
itamarst: I also suggest adding an extra line break at the end to separate synopsis from description.

itamarst:
Thanks!

1. Some minor comments above.
2. Buildbot is red.

Fix those, then merge (no need for separate code review).

PR #1194, opened 2015-03-03
[FLOC-1497] Update vagrant dev-box to use CentOS.

adamtheturtle:
https://clusterhq.atlassian.net/browse/FLOC-1497

This:
* Changes the dev box to use CentOS
* Removes the need to have an equivalent to the `fedora20-updated` box by using a Vagrant plugin installed which allows reloading of the box during provisioning
* Removes legacy installs no longer necessary with the omnibus packaging

If this is accepted then:
* At least one Vagrant plugin must be installed on Buildbot (see https://github.com/ClusterHQ/build.clusterhq.com/compare/centos-dev-FLOC-1497?expand=1 for a start)
* The box that Buildbot builds can be uploaded to Atlas


vagrant/dev/bootstrap.sh
```diff
@@ -5,17 +5,17 @@
 set -e
 
 # Install useful yum repos
-yum install -y https://s3.amazonaws.com/archive.zfsonlinux.org/fedora/zfs-release$(rpm -E %dist).noarch.rpm
-curl https://copr.fedoraproject.org/coprs/tomprince/hybridlogic/repo/fedora-20-x86_64/tomprince-hybridlogic-fedora-20-x86_64.repo >/etc/yum.repos.d/hybridlogic.repo
+yum install -y http://archive.zfsonlinux.org/epel/zfs-release$(rpm -E %dist).noarch.rpm
```
tomprince: We should still use `https` here.

vagrant/dev/Vagrantfile
```diff
@@ -19,4 +19,33 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+
+  # This allows the VM to be restarted during provisioning, necessary after
+  # updating the kernel
+  unless Vagrant.has_plugin?("vagrant-reload")
+    raise "vagrant-reload plugin is not installed"
+  end
+
+=begin
+  With some versions of VirtualBox, the following error may occur when running
+  vagrant up.
```
tomprince: This perhaps has nothing to do with versions of VirtualBox, but rather the fact that we are updating the kernel.

vagrant/dev/bootstrap.sh
```diff
@@ -5,17 +5,17 @@
...
-	kernel-headers kernel-devel rpmdevtools \
+	dkms kernel-headers kernel-devel rpmdevtools \
 	zlib-devel libuuid-devel libselinux-devel \
 	automake autoconf libtool \
 	rpm-devel rpmlint mock createrepo \
-	docker-io \
+	docker \
 	python-devel python-tox \
 	python-virtualenv python-virtualenvwrapper python-pip \
 	python-cffi libffi-devel \
```
tomprince: `python-cffi` is probably unneeded.

vagrant/dev/bootstrap.sh
```diff
@@ -5,17 +5,17 @@
...
-yum install -y https://s3.amazonaws.com/archive.zfsonlinux.org/fedora/zfs-release$(rpm -E %dist).noarch.rpm
-curl https://copr.fedoraproject.org/coprs/tomprince/hybridlogic/repo/fedora-20-x86_64/tomprince-hybridlogic-fedora-20-x86_64.repo >/etc/yum.repos.d/hybridlogic.repo
+yum install -y http://archive.zfsonlinux.org/epel/zfs-release$(rpm -E %dist).noarch.rpm
+yum install -y https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
 
 # Install packages
 yum install -y \
 	@buildsys-build git \
-	kernel-headers kernel-devel rpmdevtools \
+	dkms kernel-headers kernel-devel rpmdevtools \
```
tomprince: I would be inclined to move the `dkms` install to before the reboot, so the guest additions get installed with `dkms`.

vagrant/dev/Vagrantfile
tomprince: (maybe)

vagrant/dev/Vagrantfile
```diff
@@ -19,4 +19,33 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   if Vagrant.has_plugin?("vagrant-cachier")
     config.cache.scope = :box
   end
+
+  # This allows the VM to be restarted during provisioning, necessary after
+  # updating the kernel
+  unless Vagrant.has_plugin?("vagrant-reload")
```
tomprince: If you don't have the plugin installed, will the above provisioning that calls it fail? If so, this should be moved before that. Otherwise, here is fine.

vagrant/dev/bootstrap.sh
tomprince: Perhaps everything but installing `zfs` can be moved before rebooting?

vagrant/dev/Vagrantfile
adamtheturtle: I'll make it clear in the comment that there is uncertainty, it doesn't seem worth investigating why this happens when there is such a simple workaround.

adamtheturtle:
I have addressed the above comments.

vagrant/dev/bootstrap.sh
```diff
@@ -1,51 +1,25 @@
...
+# box must be rebooted.
 
 set -e
 
+yum update -y
 # Install useful yum repos
-yum install -y https://s3.amazonaws.com/archive.zfsonlinux.org/fedora/zfs-release$(rpm -E %dist).noarch.rpm
-curl https://copr.fedoraproject.org/coprs/tomprince/hybridlogic/repo/fedora-20-x86_64/tomprince-hybridlogic-fedora-20-x86_64.repo >/etc/yum.repos.d/hybridlogic.repo
+yum install -y https://s3.amazonaws.com/archive.zfsonlinux.org/epel/zfs-release$(rpm -E %dist).noarch.rpm
+yum install -y https://dl.fedoraproject.org/pub/epel/7/x86_64/e/epel-release-7-5.noarch.rpm
```
tomprince: `epel-release` should be avaiable by default on a fresh CentOS system, so no need for a URL.

tomprince:
The changes look good, but I haven't tested building the box yet.

tomprince:
tox fails on the new box due to a spelling error.

adamtheturtle:
I have addressed the spelling error. This does remind me that dictionaries differ across platforms (https://clusterhq.atlassian.net/browse/FLOC-717).

This is still in the "Code review" state in JIRA.

PR #1195, opened 2015-03-03
[FLOC-1477] [FLOC-1478] Deployer deletion

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1477
https://clusterhq.atlassian.net/browse/FLOC-1478

This should finish off https://clusterhq.atlassian.net/browse/FLOC-1467

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -404,13 +404,13 @@ class DatasetAPITests(TestCase):
 
         def delete_dataset(result):
             cluster, dataset = result
-            deleted = cluster.delete_dataset(dataset.dataset_id)
+            deleted = cluster.delete_dataset(dataset["dataset_id"])
 
             def not_exists():
                 request = cluster.datasets_state()
                 request.addCallback(
-                    lambda actual_datasets: dataset.dataset_id not in
-                    (d.dataset_id for d in actual_datasets))
+                    lambda actual_datasets: dataset["dataset_id"] not in
+                    (d["dataset_id"] for d in actual_datasets))
                 return request
             deleted.addCallback(lambda _: loop_until(not_exists))
             return deleted
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -12,7 +12,7 @@ from uuid import uuid4
 from json import dumps, loads
 
 from twisted.trial.unittest import TestCase
-from treq import get, post, content, delete
+from treq import get, post, content, delete, json_content
 from characteristic import attributes
 
 from .testtools import get_nodes, run_SSH
@@ -128,8 +128,7 @@ class Cluster(object):
             the state of the cluster.
         """
         request = get(self.base_url + b"/state/datasets", persistent=False)
-        request.addCallback(content)
-        request.addCallback(loads)
+        request.addCallback(json_content)
         return request
 
     def wait_for_dataset(self, dataset_properties):
@@ -213,6 +212,8 @@ class Cluster(object):
         Delete a dataset.
 
         :param unicode dataset_id: The uuid of the dataset to be modified.
+
+        :returns: A 2-tuple of (cluster, api_response)
         """
         request = delete(
             self.base_url + b"/configuration/datasets/%s" % (
@@ -222,8 +223,7 @@ class Cluster(object):
             persistent=False
         )
 
-        request.addCallback(content)
-        request.addCallback(loads)
+        request.addCallback(json_content)
         # Return cluster and API response
         request.addCallback(lambda response: (self, response))
         return request
@@ -313,7 +313,7 @@ class DatasetAPITests(TestCase):
     """
     Tests for the dataset API.
     """
-    def _test_create(self):
+    def _create_test(self):
         """
         Create a dataset on a single-node cluster.
 
@@ -351,7 +351,7 @@ class DatasetAPITests(TestCase):
         """
         A dataset can be created on a specific node.
         """
-        return self._test_create()
+        return self._create_test()
 
     def test_dataset_move(self):
         """
@@ -400,7 +400,7 @@ class DatasetAPITests(TestCase):
         """
         A dataset can be deleted, resulting in its removal from the node.
         """
-        created = self._test_create()
+        created = self._create_test()
 
         def delete_dataset(result):
             cluster, dataset = result
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -11,6 +11,8 @@ from characteristic import attributes
 
 from pyrsistent import pmap, PRecord, field
 
+from eliot import write_failure, Logger
+
 from twisted.internet.defer import gatherResults, fail, succeed
 
 from ._docker import DockerClient, PortMap, Environment, Volume as DockerVolume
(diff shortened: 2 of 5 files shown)


flocker/acceptance/test_api.py
```diff
@@ -120,6 +120,18 @@ def base_url(self):
...
+    def datasets_state(self):
+        """
+        Return the actual dataset state of the cluster.
+
+        :return: ``Deferred`` firing with a list of dataset dictionaries,
+            the state of the cluster.
+        """
+        request = get(self.base_url + b"/state/datasets", persistent=False)
+        request.addCallback(content)
+        request.addCallback(loads)
```
exarkun: fwiw there's also `treq.json_content`  :question:

itamarst:
I will have fix for failing acceptance test shortly.

flocker/acceptance/test_api.py
```diff
@@ -198,6 +208,26 @@ def update_dataset(self, dataset_id, dataset_properties):
         request.addCallback(lambda response: (self, response))
         return request
 
+    def delete_dataset(self, dataset_id):
+        """
+        Delete a dataset.
+
+        :param unicode dataset_id: The uuid of the dataset to be modified.
```
exarkun: return docs

flocker/acceptance/test_api.py
```diff
@@ -283,9 +313,13 @@ class DatasetAPITests(TestCase):
     """
     Tests for the dataset API.
     """
-    def test_dataset_creation(self):
+    def _test_create(self):
```
exarkun: :question: I like the `_foo_test` naming convention for these so that the name is more than a single `_` away from being a valid test name the runner will discover.

flocker/node/_deploy.py
```diff
@@ -320,6 +327,36 @@ def run(self, deployer):
...
+    """
+    Delete all local copies of the dataset.
+
+    A better action would be one that deletes a specific manifestation
+    ("volume" in flocker.volume legacy terminology). Unfortunately
+    currently "remotely owned volumes" (legacy terminology), aka
+    non-primary manifestations or replicas, are not exposed to the
+    deployer, so we have to enumerate them here.
+
+    :ivar Dataset: The dataset to delete.
```
exarkun: `ivar Dataset dataset: The dataset to delete.`

flocker/node/_deploy.py
```diff
@@ -320,6 +327,36 @@ def run(self, deployer):
...
+    def run(self, deployer):
+        service = deployer.volume_service
+        d = service.enumerate()
+
+        def got_volumes(volumes):
+            deletions = []
+            for volume in volumes:
+                if volume.name.dataset_id == self.dataset.dataset_id:
+                    deletions.append(service.pool.destroy(volume).addErrback(
+                        write_failure, _logger, u"flocker:p2pdeployer:delete"))
```
exarkun: I think we really need to document what we log.

Can we start a reference document that gives an exhaustive list of "systems" and what they indicate?  Or should people just refer to the implementation when they want to know what a particular log event represents?

flocker/node/test/test_deploy.py
```diff
@@ -1497,6 +1506,48 @@ def test_volume_created(self):
...
+        dataset must be deleted if the desired configuration specifies
+        that the dataset has the ``deleted`` attribute set to True.
+
+        Note that for now this happens regardless of whether the node
+        actually has the dataset, since the deployer doesn't know about
+        replicas... see FLOC-1240.
+        """
+        docker = FakeDockerClient(units={})
+        node = Node(
+            hostname=u"node1.example.com",
```
exarkun: Should be an IPv4 literal.  The API won't let you use a hostname.

flocker/node/_deploy.py
itamarst: Probably the `system` thing is a bogus feature that should be removed.
1. You should be able to deduce what caused the problem from the logging action context.
2. It's a traceback! It tells you exactly where it came from!

I'll blame sleep deprivation.

flocker/node/test/test_deploy.py
```diff
@@ -1497,6 +1506,48 @@ def test_volume_created(self):
...
+            self.successResultOf(api.discover_local_state()),
+            desired_configuration=desired,
+            current_cluster_state=current,
+        )
+
+        expected = Sequentially(changes=[
+            InParallel(changes=[DeleteDataset(dataset=DATASET.set(
+                "deleted", True))])
+            ])
+        self.assertEqual(expected, changes)
```
exarkun: :question: Should there also be a test that verifies deletions happen after other things?  eg, if a node only sees a dataset for the first time in the same pass as it sees it is deleted, I think it will create and then delete the dataset.  That's fine - as long as it actually does it in that order (instead of trying and failing to delete it, then creating it).  Although...  I suppose it will just delete it on the next pass anyway so maybe it doesn't really matter.

flocker/node/test/test_deploy.py
```diff
@@ -3044,6 +3095,52 @@ def test_return(self):
...
+            network=make_memory_network())
+
+        id1 = unicode(uuid4())
+        self.volume1 = self.volume_service.get(_to_volume_name(id1))
+        id2 = unicode(uuid4())
+        self.volume2 = self.volume_service.get(_to_volume_name(id2))
+        self.successResultOf(self.volume_service.create(self.volume1))
+        self.successResultOf(self.volume_service.create(self.volume2))
+
+    def test_creates(self):
```
exarkun: misnamed I think

flocker/node/test/test_deploy.py
```diff
@@ -3044,6 +3095,52 @@ def test_return(self):
...
+        """
+        ``CreateDataset.run()`` deletes volumes whose ``dataset_id`` matches
```
exarkun: wrong class name

flocker/node/test/test_deploy.py
```diff
@@ -3044,6 +3095,52 @@ def test_return(self):
...
+            [self.volume1])
+
+    @validate_logging(
+        lambda test, logger: logger.flush_tracebacks(RuntimeError))
+    def test_failed_create(self, logger):
+        """
+        Failed deletions of volumes are swallowed.
+        """
+        self.patch(self.volume_service.pool, "destroy",
+                   lambda fs: fail(RuntimeError()))
```
exarkun: nit, `RuntimeError` is for errors in the runtime (ie, CPython), not errors that happen at run time.

exarkun: not **just** swallowed though, in fact not really swallowed at all: actually logged.  This test probably passes whether they're logged or silently discarded?

flocker/node/_deploy.py
exarkun: > You should be able to deduce what caused the problem from the logging action context.

That's a good point.  I think you specifically mean the `system` argument accepted by `write_failure` here?  Maybe documentation for systems in the other Eliot APIs still makes sense.

itamarst: Yeah, the `write_failure`/`write_traceback` API. In other Eliot logging we'll be including documentation of messages and fields, so in theory we can have some code that extracts that and generates some logging reference documentation.

exarkun:
Thanks.  Looks good overall.  Please address minor inline comments to your satisfaction, verify buildbot is green, and then merge.

flocker/node/_deploy.py
itamarst: https://github.com/ClusterHQ/eliot/issues/147

flocker/node/test/test_deploy.py
itamarst: Yeah. On the other hand deletion has to happen after application stop, otherwise it will fail and then prevent application from stopping. So I'll test that.

PR #1196, opened 2015-03-04
[FLOC-1521] Test homebrew script

dwgebler:
This is part of FLOC-1464, subtask 1521. This is the script that will be triggered by Buildbot inside a virtual environment containing a flocker installation on a OS X host machine. The script expects the host (the office Mac mini) to contain an OS X virtual machine image in a particular location and for that VM to have a snapshot called homebrew-clean.

The script reverts the VM to this snapshot, starts it up, attempts to run a sequence of brew commands via SSH (terminating with non-zero exit code should any of those commands exit with an error), then stops the VM and terminates with a success code when this entire process is completed.

The script expects to receive a fully qualified URL to a homebrew recipe file as its only argument - this should be supplied by Buildbot when triggering the script's execution.


flocker/common/_ssh.py
```diff
@@ -0,0 +1,82 @@
...
+    host_string = "%s@%s" % (username, address)
+    with settings(
+            connection_attempts=24,
+            timeout=5,
+            pty=False,
+            host_string=host_string):
+
+        for command in commands:
+            handlers[type(command)](command)
+    disconnect_all()
```
wallrj: Let's not move these functions.

You can make them available in the new script by adding @tomprince 's preamble as the first import, which will make the flocker packages available on the path.

See other admin scripts for example.

wallrj: And those provisioning functions are actually a link between documented installation commands and commands that get run in our acceptance tests.

You should be able to modify the command examples in:
 * http://doc-dev.clusterhq.com/indepth/installation.html#os-x

...with snippets that you define in `_install.py` eg https://github.com/ClusterHQ/flocker/blob/master/flocker/provision/_install.py#L99

wallrj:
Various flake8 warnings.

```
testbrew.py:51:1: F841 local variable 'recipe_resource' is assigned to but never used
testbrew.py:52:1: F841 local variable 'revert_result' is assigned to but never used
testbrew.py:55:1: F841 local variable 'start_result' is assigned to but never used
testbrew.py:63:1: F841 local variable 'stop_result' is assigned to but never used
testbrew.py:80:1: W391 blank line at end of file
```

admin/testbrew.py
```diff
@@ -0,0 +1,80 @@
...
+
+VM_USERNAME = "ClusterHQVM"
+
+VM_HOST = "10.0.126.88"
+
+
+def main():
+    try:
+        if len(sys.argv) < 2:
+            raise Exception("URL of homebrew recipe not specified")
```
wallrj: Use `twisted.options` here for consistency.

See https://github.com/ClusterHQ/flocker/blob/master/admin/acceptance.py#L344 for example.

And make `VM_HOST` , `VM_USERNAME` `YOSEMITE_SNAPSHOT` `YOSEMITE_VMX_PATH` etc options defaults

admin/testbrew.py
```diff
@@ -0,0 +1,80 @@
...
+        # We do not need to read its content.
+        recipe_resource = urllib2.urlopen(recipe_url)
+        revert_result = check_output([
+            "vmrun", "revertToSnapshot", YOSEMITE_VMX_PATH, YOSEMITE_SNAPSHOT,
+        ])
+        start_result = check_output([
+            "vmrun", "start", YOSEMITE_VMX_PATH, "nogui",
+        ])
+        run_with_fabric(VM_USERNAME, VM_HOST, commands=[
+            Run(command="brew update"),
```
wallrj: These commands can be defined as a single `task_` in https://github.com/ClusterHQ/flocker/blob/master/flocker/provision/_install.py and then included in the documentation in place of the the commands that follow "Make sure Homebrew has no issues:" in the documentation.

admin/testbrew.py
```diff
@@ -0,0 +1,80 @@
...
+            cmd=" ".join(e.cmd), code=e.returncode
+        )
+        sys.exit(1)
+    except Exception as e:
+        print "Error: {error}.".format(error=str(e))
+        sys.exit(1)
+
+if __name__ == "__main__":
+    main()
+
```
wallrj: Separate the script stub from the homebrew module containing a main function. As in:
 * https://github.com/ClusterHQ/flocker/blob/master/admin/build-package

wallrj:
Thanks @dwgebler 

I left some inline comments and suggestions. I think most of them should be simple to address.

Then submit for another review.

admin/testbrew.py
```diff
@@ -0,0 +1,80 @@
...
+    try:
+        BASEPATH, TOPLEVEL
+        if len(sys.argv) < 2:
+            raise Exception("URL of homebrew recipe not specified")
+        recipe_url = sys.argv[1]
+        # Open the recipe URL just to validate and verify that it exists.
+        # We do not need to read its content.
+        urllib2.urlopen(recipe_url)
+        check_output([
+            "vmrun", "revertToSnapshot", YOSEMITE_VMX_PATH, YOSEMITE_SNAPSHOT,
```
adamtheturtle: Why not use `vmfusion` library as per https://github.com/ClusterHQ/flocker/pull/1172/files? In the meeting IIRC it was because of a desire to have no dependencies, but that is not an issue now.

dwgebler: I don't know, what benefit does using `vmfusion` and having the extra package dependency confer over just executing the CLI `vmrun` tool?

itamarst:
If you're doing PRs that are only part of an issue, could you please do it via sub-issues? That way it's much clearer what the other tasks in the issue are, and what the scope of the review should be.

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+            raise UsageError((
+                "URL of homebrew recipe not specified. "
+                "Run with --help for usage"
+            ))
+        else:
+            self['recipe_url'] = args[0]
+
+
+def main():
+    try:
```
adamtheturtle: :question: Personal style preference: In (some?) other admin scripts where a whole `main` function wants to be in a `try...catch` clause there is a new function made with a more descriptive name than `main` and `main` calls that in a `try...catch`.

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+YOSEMITE_SNAPSHOT = "homebrew-clean"
+
+VM_USERNAME = "ClusterHQVM"
+
+VM_HOST = "10.0.126.88"
+
+
+class TestBrewOptions(Options):
+    description = ("Start a OS X Virtual Machine and test a brew recipe "
+                   "installatiion.")
```
adamtheturtle: Typo "installatiion" > "installation".

:question: brew recipe > Homebrew recipe

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+
+YOSEMITE_SNAPSHOT = "homebrew-clean"
+
+VM_USERNAME = "ClusterHQVM"
+
+VM_HOST = "10.0.126.88"
+
+
+class TestBrewOptions(Options):
+    description = ("Start a OS X Virtual Machine and test a brew recipe "
```
adamtheturtle: Probably best here or somewhere else to specify that it must be a VMWare VM

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+    "{home}/Documents/Virtual Machines.localized/"
+    "OS X 10.10.vmwarevm/OS X 10.10.vmx"
+    '"'
+).format(home=os.getenv("HOME"))
+
+YOSEMITE_SNAPSHOT = "homebrew-clean"
+
+VM_USERNAME = "ClusterHQVM"
+
+VM_HOST = "10.0.126.88"
```
adamtheturtle: :question: Because people might not be very familiar with OS X it might be useful to specify how to find this.

adamtheturtle:
Follow up to a now-buried comment thread at https://github.com/ClusterHQ/flocker/pull/1196#discussion_r25776399

> I don't know, what benefit does using vmfusion and having the extra package dependency confer over just executing the CLI vmrun tool?

Python is more readable (to me) and more maintainable (in my experience) and it means that the work of making a wrapper for the shell commands is deferred to someone else. Also, in the particular case of VMWare Fusion, some of the options are particularly undescriptive. The `vmfusion` wrapper does a good job of helping the understandability.

I admit that these concerns aren't major - this is test infrastructure code, but that the code to use this tool is already written and was replaced by shell to avoid having to install dependencies - something which is done now anyways.

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+         'IP address or hostname of the Virtual Machine'],
+        ['vmuser', 'u', VM_USERNAME,
+         'Username for the Virtual Machine'],
+        ['vmpath', 'p', YOSEMITE_VMX_PATH,
+         'Full path to the Virtual Machine image'],
+        ['vmsnapshot', 's', YOSEMITE_SNAPSHOT,
+         'Snapshot identifier for the Virtual Machine'],
+    ]
+
+    synopsis = ('Usage: testbrew.py <recipe URL> [--vmhost <host>] '
```
adamtheturtle: Our other `admin` scripts follow a pattern which in this case would be something like:

test-homebrew:

```
#!/usr/bin/env python
"""
Test a Homebrew script.
"""

from _preamble import TOPLEVEL, BASEPATH

import sys

if __name__ == '__main__':
    from admin.test_homebrew import main
    main(sys.argv[1:], top_level=TOPLEVEL, base_path=BASEPATH)
```

test_homebrew.py: What is here now.

It is probably worth glancing over one or two to keep consistency. Note that at least `upload-rpms` is an early script which is different but there is WIP to change that.

admin/testbrew.py
```diff
@@ -0,0 +1,116 @@
...
+        options = TestBrewOptions()
+        try:
+            options.parseOptions()
+        except UsageError as e:
+            sys.stderr.write("Error: {error}.\n".format(error=str(e)))
+            sys.exit(1)
+        recipe_url = options['recipe_url']
+        # Open the recipe URL just to validate and verify that it exists.
+        # We do not need to read its content.
+        urllib2.urlopen(recipe_url)
```
adamtheturtle: Is this better in some way than having Homebrew fail because the recipe does not exist? If so please add that as a comment.

dwgebler:
> something which is done now anyways

@adamtheturtle can you elaborate on this? My understanding is that Buildbot will trigger this script from inside a virtual environment on the host OS that's had Flocker's dependencies installed, so even if `vmfusion` is installed as a Python package on the host, the virtualenv might not have access to it, so it would have to be a Flocker dependency. Is that what's already been done?

admin/testbrew.py
dwgebler: There's no point going to the overhead of launching a VM to run brew commands if you can already be 100% certain they're going to fail because the recipe location doesn't exist or isn't accessible. I don't imagine that needs a separate comment other than the one already in the code.

adamtheturtle:
> @adamtheturtle can you elaborate on this?

There is now known to be a virtualenv with Flocker's dependencies on the slave so you can add just one more dependency.

No need for the PR to be merged. You can use composition, inheritance or a pip feature https://pip.pypa.io/en/latest/reference/pip_install.html#git.

Yes, `vmfusion`'s underlying code is very similar to the code in this branch but it does the work (and more, as mentioned e.g. checking that the VM exists) for us.

dwgebler:
> There is now known to be a virtualenv with Flocker's dependencies on the slave so you can add just one more dependency.

So it would need to be added as another dependency of Flocker for the purposes of this one script; personally I don't see sufficient benefit at this stage.

> No need for the PR to be merged. You can use composition, inheritance or a pip feature https://pip.pypa.io/en/latest/reference/pip_install.html#git.

Hmm, not something I've ever done and it sounds to me like the kind of thing we wouldn't do; our dependencies should be standard, released packages.

> Yes, vmfusion's underlying code is very similar to the code in this branch but it does the work

`vmfusion-python` does it in a non-blocking manner though, which is not what we want here.

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env python
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Script executed by Buildbot on the Mac mini build slave, this instantiates the
```
exarkun: This is an operational concern.  It's interesting that this is for OS X and it's interesting that it depends on, say, VMware.  It's not so interesting that Buildbot runs it (maybe we'll switch to Jenkins) or that it runs on a Mac mini (maybe we'll move it to run on a VM or an OS X cloud service).

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
...
+        ['vmpath', 'p', YOSEMITE_VMX_PATH,
+         'Full path to the Virtual Machine image'],
+        ['vmsnapshot', 's', YOSEMITE_SNAPSHOT,
+         'Snapshot identifier for the Virtual Machine'],
+    ]
+
+    synopsis = ('Usage: test-brew-recipe <recipe URL> [--vmhost <host>] '
+                '[--vmuser <username>] '
+                '[--vmpath <path>] '
+                '[--vmsnapshot <name>]')
```
exarkun: The typical idiom is `[options]`.  `Options` will automatically generate help text for the various options that actually exist.  This way there's only one version of truth to maintain.

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
...
+
+from flocker.provision._install import run_with_fabric, task_test_homebrew
+
+
+YOSEMITE_VMX_PATH = (
+    '"'
+    "{home}/Documents/Virtual Machines.localized/"
+    "OS X 10.10.vmwarevm/OS X 10.10.vmx"
+    '"'
+).format(home=os.getenv("HOME"))
```
exarkun: `os.path.expanduser("~/Documents...")` also works here and is maybe more idiomatic.  :question:

exarkun: The extra quotes around this value are a bit confusing.  They make this string not actually be the path it seems to be.  If they're necessary later, for example to be able to pass this to a shell command, they should be added later, not here.  I don't think they should be necessary anywhere, though (since you're using the non-shell form of `check_output`).  Does this default value actually work?  I would guess not... Perhaps it's not worth supplying a default value for this parameter.

exarkun: Paths should be represented as `twisted.python.filepath.FilePath` instances.

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
...
+                '[--vmuser <username>] '
+                '[--vmpath <path>] '
+                '[--vmsnapshot <name>]')
+
+    def parseArgs(self, *args):
+        if len(args) < 1:
+            raise UsageError((
+                "URL of homebrew recipe not specified. "
+                "Run with --help for usage"
+            ))
```
exarkun: Interesting approach.  It strikes me that `Options` could support writing this code in the normal usual way (ie, `def parseArgs(self, recipe_url): ...` and automatically generate a message like this.  Perhaps worth filing a ticket against Twisted for.

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
...
+    except CalledProcessError as e:
+        sys.stderr.write(
+            (
+                "Error: Command {cmd} terminated with exit status {code}.\n"
+            ).format(cmd=" ".join(e.cmd), code=e.returncode)
+        )
+        sys.exit(1)
+    except Exception as e:
+        sys.stderr.write("Error: {error}.\n".format(error=str(e)))
+        sys.exit(1)
```
exarkun: is there a particular exception you wanted this block to handle or is it really the catch-all fallback case it seems to be?  Since this is a developer-facing tool I think it's probably better to let the unexpected exception reach the top-level and get a traceback reported.  It's going to be hard to debug if this handler is left in because only the exception message will be visible - no context.  Throwing tracebacks at end-users isn't nice because they'll get confused but throwing them at developers is nice because then they have more information about how to fix the bug.

admin/testbrew.py
```diff
@@ -0,0 +1,112 @@
...
+        * Homebrew installed and available
+"""
+
+import os
+import sys
+import urllib2
+
+from twisted.python.usage import Options, UsageError
+
+from subprocess import check_output, CalledProcessError
```
exarkun: nit pick - it's nice to have all stdlib imports grouped together, followed by all third-party imports grouped together, followed by all internal imports grouped together.  :question:

admin/testbrew.py
exarkun: If this weren't just test code I'd probably suggest that it makes sense to save and **use** the recipe content that gets downloaded by this step.  That way the failure happens before the VM boots, if it is going to happen, but you also avoid (1) double downloading the recipe and (2) possibly failing inside the VM anyway if the remote resource changes.

However, that's probably not worth doing in this context and this check is just fine as-is.

exarkun:
Thanks.  This looks pretty good.  I haven't actually tried it myself, I'll assume you've done the necessary manual testing to demonstrate it actually accomplishes what we want.  Please address the inline comments and then merge.

adamtheturtle:
I think that there is one still-relevant comment buried by GitHub from @wallrj - https://github.com/ClusterHQ/flocker/pull/1196#discussion_r25769514 - use the newly defined `task_` in the docs.

dwgebler:
@adamtheturtle I'm going to file a separate issue if necessary for task directive docs integration, as there is a discrepancy at the moment between what this script runs and what the doc instructions are. I looked in to it but our current implementation of the task directive doesn't support everything we'd need to resolve this.

flocker/provision/_install.py
```diff
@@ -96,6 +96,14 @@ def run_with_fabric(username, address, commands):
 run = run_with_fabric
 
 
+def task_test_homebrew(recipe_url='flocker'):
```
adamtheturtle: :question: Is this default necessary? Without `brew tap clusterhq/flocker` it won't work, and even then it is just happenstance that it would work (the `flocker` recipe should be deleted).

jongiddy:
Tested on Mac Mini using:
```
./pip-install.sh
source pytest/bin/activate
git clone https://github.com/ClusterHQ/flocker.git
cd flocker/
git checkout test-homebrew-script-FLOC-1464
pip install -e .[dev]
admin/test-brew-recipe <url>
deactivate
cd
./pip-uninstall.sh
rm -fr flocker
```

PR #1197, opened 2015-03-04
Copy key in a way which preserves metadata

adamtheturtle:
See FLOC-1471 comments for information and linked Buildbot PR.

### DIFF
--- a/admin/aws.py
+++ b/admin/aws.py
@@ -119,30 +119,46 @@ class CopyS3Keys(object):
     """
 
 
+EXTENSION_MIME_TYPES = {
+    '.eot': 'application/vnd.ms-fontobject',
+    '.gif': 'image/gif',
+    '.html': 'text/html',
+    '.jpg': 'image/jpeg',
+    '.js': 'application/javascript',
+    '.css': 'text/css',
+    '.png': 'image/png',
+    '.sh': 'text/plain',
+    '.svg': 'image/svg+xml',
+    '.ttf': 'application/x-font-ttf',
+    '.txt': 'text/plain',
+    '.woff': 'application/font-woff',
+    '.yml': 'text/plain',
+}
+
+
 @sync_performer
 def perform_copy_s3_keys(dispatcher, intent):
     """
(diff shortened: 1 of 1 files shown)


admin/aws.py
```diff
@@ -126,11 +126,23 @@ def perform_copy_s3_keys(dispatcher, intent):
     """
     s3 = boto.connect_s3()
     destination_bucket = s3.get_bucket(intent.destination_bucket)
+    source_bucket = s3.get_bucket(intent.source_bucket)
     for key in intent.keys:
-        destination_bucket.copy_key(
-            new_key_name=intent.destination_prefix + key,
-            src_bucket_name=intent.source_bucket,
-            src_key_name=intent.source_prefix + key)
+        if '/_images/' in key and key.endswith('.svg'):
```
itamarst: Check for `/_images/` seems fragile? And I can't imagine we'd ever upload something to docs that ends with `.svg` that isn't a SVG.

adamtheturtle: I didn't want to touch fonts as I'd heard no complaints about those, which is why I had this here.

admin/aws.py
```diff
@@ -126,11 +126,23 @@ def perform_copy_s3_keys(dispatcher, intent):
...
+            # s3cmd isn't great at choosing Content-Type. This fixes the one
+            # symptom of that we know of.
```
tomprince: I've done some poking around, and it appears that with the latest version of `s3cmd`, if we *don't* pass `--no-mime-magic`, it does the right thing for svg.

tomprince:
I think, given that `s3cmd` can be made to do the right thing, this branch is unneeded.

tomprince:
I wonder if, rather than checking just some files, and fixing their mime time, if it would make sense to just override the mime-type and not try and depend on something external to get the rest right.

The following is a list of the extant extensions:
- eot
- gif
- html
- jpg
- js
- css
- png
- sh
- svg
- ttf
- txt
- woff
- yml

itamarst:
Sounds good.

PR #1200, opened 2015-03-06
insert a one second sleep as a temporary solution to this issue

exarkun:
Fixes FLOC-1378 (at least, well enough for now)

### DIFF
--- a/flocker/node/test/test_loop.py
+++ b/flocker/node/test/test_loop.py
@@ -460,18 +460,28 @@ class ConvergenceLoopFSMTests(SynchronousTestCase):
         # Calculating actions happened, action is run, but waits for
         # Deferred to be fired... Meanwhile a stop input is received!
         loop.receive(ConvergenceLoopInputs.STOP)
-        # Action finally finishes, and nothing more happens; if another
-        # iteration did happen the loop would attempt discovery, which
-        # would make deployer.discover_local_state() to fail since it was
-        # only configured with one result.
+        # Action finally finishes:
         action.result.callback(None)
         reactor.advance(1.0)
 
-        self.assertEqual(
-            (deployer.calculate_inputs, client.calls, loop.state),
-            ([(local_state, configuration, state)],
-             [(NodeStateCommand, dict(node_state=local_state))],
-             ConvergenceLoopStates.STOPPED))
+        # work is scheduled:
+        expected = (
+            # The actions are calculated
+            [(local_state, configuration, state)],
+            # And the result is run
+            [(NodeStateCommand, dict(node_state=local_state))],
+            # The state machine gets to the desired state.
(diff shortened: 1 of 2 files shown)


flocker/node/test/test_loop.py
```diff
@@ -338,10 +341,33 @@ def test_convergence_done_changes(self):
...
+        state = object()
+        action = ControllableAction(succeed(None))
+        deployer = ControllableDeployer([succeed(local_state)], [action])
+        client = self.successful_amp_client([local_state])
+        reactor = Clock()
+        loop = build_convergence_loop_fsm(reactor, deployer)
+        loop.receive(_ClientStatusUpdate(
+            client=client, configuration=configuration, state=state))
+        # Calculating actions happened, result was run... and then we did
+        # whole thing again:
```
tomprince: We don't do the whole thing again.

flocker/node/test/test_loop.py
```diff
@@ -434,6 +466,8 @@ def test_convergence_stop(self):
         # would make deployer.discover_local_state() to fail since it was
         # only configured with one result.
         action.result.callback(None)
+        reactor.advance(1.0)
```
tomprince: Would it make sense to assert that there are no pending calls on the clock (i.e. nothing will ever happen), instead of advancing and asking if anything has happened yet?

tomprince:
Too minor comments, then merge.

flocker/node/test/test_loop.py
exarkun: Sorry.  Copy/paste is awful. :disappointed:

flocker/node/test/test_loop.py
exarkun: I think sometimes that's a reasonable strategy.  The troublesome part comes when it's legitimate for **some** stuff to be scheduled to happen, just not the one thing you don't want to happen.  It's hard/gross/fragile to sort out and ignore other unrelated delayed calls so you can assert just that the one you don't want to exist doesn't exist.

In this case it may be okay because it's really a unit test (so only a narrow bit of code gets run) and because there's no other use of `IReactorTime` anywhere near this code.  So, for right now, there's no sorting to do: the correct assertion to make is just that there are no delayed calls at all.

And actually it turns out that I think we want to do what I already made this test do **and** make that additional assertion.  Reading the test again now, I see that I asserted that we get to stopped but I didn't assert that nothing additional happens afterwards.  FWIW, the `advance(1.0)` is necessary even to get to the stopped state because I put the delay into a funny place in the chain of events.  Unfortunately it's much less convenient to put it somewhere where it doesn't get in the way of transitioning to the `STOPPED` state.  I think this is okay because transitioning to `STOPPED` is infrequent, you're not in a rush to do much of anything once you get there, and because we'll get rid of this hack at some point.

PR #1202, opened 2015-03-06
[FLOC-1241] Pyrsistent-based network serialization

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1241

This will require converting remaining model classes to use pyrsistent, i.e. `Application` and contained objects.

Note that the NodeState serialization/deserialization test passes with this sketch.


wallrj:
Thanks @itamarst 

All looks good. Here are a few notes that @exarkun and I made while thinking about the design:

* How to serialise / de-serialise FilePath
  * produce bytes from path attribute
  * takes bytes filepath or FilePath
  * A pathmap_field function?

* How to serialise / de-serialise stateless records such as IRestartPolicy?
   * maybe see `ApplicationMarshaller.convert_restart_policy`
   * _parse_restart_policy
   * Refactor these from _config

* Maybe create an `NodeState.__invariant__` for checking applications are only in one of running and not_running (or as followup issue)

* Do you intend to only convert AMP protocol related types in this branch and convert other Characteristic types later, as and when needed?
 * Application, Port, Link etc needs to be converted
   * Looks trivial

Please proceed.

itamarst:
The plan is to convert everything to `pyrsistent` since it's pretty easy, especially if you leave out type checking.

flocker/control/_model.py
```diff
@@ -34,6 +47,8 @@ class DockerImage(object):
     :ivar unicode full_name: A readonly property which combines the repository
         and tag in a format that can be passed to `docker run`.
     """
+    repository = field(mandatory=True)
+    tag = field(mandatory=True, initial=u"latest")
```
exarkun: type constraints for these two fields?

flocker/control/_model.py
```diff
@@ -65,22 +80,35 @@ def from_string(cls, input):
...
 
-    :ivar Manifestation manifestation: The ``Manifestation`` that is being
-        attached as a volume. For now this is always from a ``Dataset``
-        with the same as the name of the application it is attached to
-        https://clusterhq.atlassian.net/browse/FLOC-49).
+    :ivar int internal_port: The port number exposed by the application.
+    :ivar int external_port: The port number exposed to the outside world.
+    """
+    internal_port = field(mandatory=True, type=int)
+    external_port = field(mandatory=True, type=int)
```
exarkun: I wonder if `type=int` will cause trouble when things are accidentally `long`.

flocker/control/_model.py
```diff
@@ -65,22 +80,35 @@ def from_string(cls, input):
...
+        This is used to determine the environment variables to populate in the
+        container.
+    :ivar int remote_port: The port exposed externally by the remote
+        application.
+    :ivar unicode alias: Environment variable prefix to use for exposing
+        connection information.
+    """
+    local_port = field(mandatory=True, type=int)
+    remote_port = field(mandatory=True, type=int)
+    alias = field(mandatory=True)
```
exarkun: `type=unicode`?

flocker/control/_model.py
```diff
@@ -90,37 +118,31 @@ class IRestartPolicy(Interface):
...
     """
     A restart policy that restarts an application when it fails.
 
-    :ivar int maximum_retry_count: The number of times the application is
-        allowed to fail, before the giving up.
+    :ivar maximum_retry_count: The number of times the application is
+        allowed to fail before giving up, or ``None`` if there is no
+        maximum.
     """
+    maximum_retry_count = field(mandatory=True, initial=None)
```
exarkun: `type`?

flocker/control/_model.py
```diff
@@ -128,24 +150,17 @@ def __init__(self):
         """
         if self.maximum_retry_count is not None:
             if not isinstance(self.maximum_retry_count, int):
-                raise TypeError(
-                    "maximum_retry_count must be an integer or None, "
-                    "got %r" % (self.maximum_retry_count,))
+                return (False,
+                        "maximum_retry_count must be an integer or None, "
+                        "got %r" % (self.maximum_retry_count,))
```
exarkun: Hm I guess this does type enforcement.  Any reason not to switch this to built-in pyrsistent type enforcement?

flocker/control/_model.py
```diff
@@ -174,6 +189,16 @@ class Application(object):
...
+    name = field(mandatory=True)
+    image = field(mandatory=True, type=DockerImage)
+    ports = pset_field(Port)
+    volume = field(mandatory=True, initial=None)
+    links = pset_field(Link)
+    memory_limit = field(mandatory=True, initial=None)
+    cpu_shares = field(mandatory=True, initial=None)
+    restart_policy = field(mandatory=True, initial=RestartNever())
+    environment = field(mandatory=True, initial=pmap(), factory=pmap,
+                        type=PMap)
```
exarkun: various missing `type`s

flocker/control/_model.py
```diff
@@ -400,14 +416,10 @@ class NodeState(PRecord):
         node. Maps ``dataset_id`` to a ``FilePath``.
     """
     hostname = field(type=unicode, factory=unicode, mandatory=True)
```
exarkun: `factory=unicode` seems dangerous?  But not related to this branch I guess...

flocker/control/_persistence.py
```diff
@@ -4,37 +4,65 @@
...
     """
-    Convert a ``Deployment`` object to ``bytes``.
-
-    :param Deployment deployment: Object to serialize.
+    JSON encoder that can encode the configuration model.
+    """
+    def default(self, obj):
+        if isinstance(obj, PRecord):
+            result = dict(obj)
+            result[u"$__class__$"] = obj.__class__.__name__
```
exarkun: pretty.  maybe make this a module constant.

flocker/control/_persistence.py
```diff
@@ -4,37 +4,65 @@
...
+    def decode_object(dictionary):
+        class_name = dictionary.get("$__class__$", None)
+        if class_name == u"FilePath":
+            return FilePath(dictionary.get(u"path").encode("utf-8"))
+        elif class_name in classes:
+            dictionary = dictionary.copy()
+            dictionary.pop("$__class__$")
+            return classes[class_name].create(dictionary)
+        else:
+            return dictionary
```
exarkun: Is there a test for disallowed classes not being deserialized?  I didn't see one.  There probably ought to be.

flocker/control/_protocol.py
```diff
@@ -56,10 +54,10 @@ class DeploymentArgument(Argument):
     AMP argument that takes a ``Deployment`` object.
     """
     def fromString(self, in_bytes):
-        return deserialize_deployment(in_bytes)
+        return wire_decode(in_bytes)
 
     def toString(self, deployment):
-        return serialize_deployment(deployment)
+        return wire_encode(deployment)
```
exarkun: These two AMP types are pretty similar now.  Maybe even exactly the same.  Perhaps we only need one of them?  Or perhaps the way they should differ now is to disallow objects of the wrong type.

flocker/cli/functional/test_deploy_script.py
```diff
@@ -69,7 +69,7 @@ def test_installs_public_sshkeys(self):
             nodes=frozenset([
```
exarkun: These could also all turn into regular lists now, right?  :question:

flocker/cli/functional/test_deploy_script.py
```diff
@@ -69,7 +69,7 @@ def test_installs_public_sshkeys(self):
...
                 Node(
                     hostname=str(self.server.ip),
```
exarkun: And these should really all be `unicode` instead of `str` (or, like, neither... structured objects for the win).  :question:

flocker/control/_persistence.py
```diff
@@ -57,7 +85,7 @@ def startService(self):
             self._path.makedirs()
         self._config_path = self._path.child(b"current_configuration.pickle")
```
exarkun: The extension probably shouldn't be `.pickle` anymore.

flocker/control/test/test_httpapi.py
```diff
@@ -1535,11 +1535,13 @@ def test_application_volumes(self):
 
         node = Node(
             hostname=expected_hostname,
-            applications=frozenset({Application(name=u'mysql-clusterhq',
-                                                image=object()),
-                                    Application(name=u'site-clusterhq.com',
-                                                image=object(),
-                                                volume=volume)}),
+            applications=frozenset({
```
exarkun: Doesn't need to be a `frozenset` anymore, right?

flocker/control/test/test_model.py
```diff
@@ -64,12 +64,13 @@ def test_repr(self):
...
+        image = repr(DockerImage(repository=u'clusterhq/flocker',
+                                 tag=u'release-14.0'))
         self.assertEqual(
-            "<DockerImage(repository=u'clusterhq/flocker', "
-            "tag=u'release-14.0')>",
-            repr(image)
+            [image.startswith("DockerImage"),
+             "clusterhq/flocker" in image,
+             "release-14.0" in image],
+            [True, True, True],
```
exarkun: I wonder why we ever cared about this.

exarkun:
Thanks.

It looks like this is also meant to resolve FLOC-1528?  Or perhaps only partially, since it doesn't introduce a versioning scheme.  But if you rename `current_configuration.pickle` to `current_configuration.v1.json` then that's a versioning scheme...

Please address inline comments and then merge.

flocker/control/_model.py
itamarst: I've tried to minimize adding constraints so that I don't have to update a bazillion tests that pass in bytes or whatever.

flocker/control/_model.py
itamarst: Yep. Could add a `factory=int`, or figure out how to do multi-type constraints.

itamarst: But doesn't seem urgent.

flocker/control/_model.py
itamarst: See previous comment about not adding too many constraints in this branch.

flocker/control/_model.py
itamarst: Need to support both `None` and `int`. Built-in type validation does not seem to support that.

itamarst: Filed https://github.com/tobgu/pyrsistent/issues/28

flocker/control/_model.py
itamarst: Filed https://clusterhq.atlassian.net/browse/FLOC-1552

flocker/control/_model.py
exarkun: I think you just pass a tuple to `type`, eg `type=(int, NoneType)`

itamarst:
I don't think it resolves FLOC-1528 because there's still no upgrade mechanism, nor detection of incompatible changes etc..

PR #1203, opened 2015-03-09
[FLOC-1533] Modify Homebrew recipe creation to accept explicit path to distribution.

jongiddy:
Modify the existing script to accept a command line argument specifying the URL from which to download the sdist code.  Using the explicit argument enables CI by giving more flexibility to where the sdist file is located.  For backwards compatibility, providing no argument continues to use the release location on Google Storage.

Note, this change is equivalent to the change proposed in PR #1192, but separated from the additional design work to actually enable CI.

### DIFF
--- a/admin/homebrew.py
+++ b/admin/homebrew.py
@@ -134,6 +134,15 @@ def get_resource_stanzas(dependency_graph):
 def main():
     """
     Print a Homebrew recipe for the Flocker distribution.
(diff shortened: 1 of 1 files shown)


itamarst:
I suggest documenting the new command line argument (just a sentence in the docstring, say). Otherwise looks good, thanks!

PR #1205, opened 2015-03-10
[FLOC-1534] Flags to set version and output file for Homebrew recipe creation

jongiddy:
BuildBot prefers command arguments to environment variables and file re-direction. Add some flags to the Homebrew creation script to rely on command options, while keeping backwards compatibility to currently documented release process.

Also, add some logging for progress.

### DIFF
--- a/admin/homebrew.py
+++ b/admin/homebrew.py
@@ -6,10 +6,12 @@ Create a Homebrew recipe for Flocker, using the VERSION environment variable.
 Inspired by https://github.com/tdsmith/labmisc/blob/master/mkpydeps.
 """
 
+import argparse
+import logging
 import sys
 from os import environ
 from json import load
-from urllib2 import HTTPError, urlopen
+from urllib2 import urlopen
 from hashlib import sha1
 from tl.eggdeps.graph import Graph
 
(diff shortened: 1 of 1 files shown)


admin/homebrew.py
```diff
@@ -144,17 +131,39 @@ def main():
     If no command line argument is provided, use the standard release
     location for the indicated version.
     """
-    version = get_version()
-    if len(sys.argv) < 2:
+    logging.basicConfig(level=logging.INFO)
+
+    parser = argparse.ArgumentParser(description='Process some integers.')
```
dwgebler: This could be a bit more descriptive, if I run `--help` at the moment I get
```
usage: make-homebrew-recipe [-h] [--flocker-version FLOCKER_VERSION]
                            [--sdist SDIST] [--output-file OUTPUT_FILE]

Process some integers.
```

admin/homebrew.py
```diff
@@ -144,17 +131,39 @@ def main():
...
+    parser.add_argument(
+        '--flocker-version', help='a version number for the Homebrew recipe')
+    parser.add_argument(
+        '--sdist', help='source distribution')
```
dwgebler: Likewise, probably worth giving the help description as something like "URL of source distribution"

dwgebler:
Not resulting from this branch, but a flake8 error present in the file:
`admin/homebrew.py:175:80: E501 line too long (91 > 79 characters)`

Other than that, I've tested this the old way (with VERSION environment variable) and the new way with CLI options, both with and without custom `sdist` locations and it all works (although I've not verified the recipe files generated, which seems unnecessary for the changes in this branch).

There were two minor inline comments, please address and merge.

admin/homebrew.py
jongiddy: Ooops. Copy-paste error.

PR #1206, opened 2015-03-10
[FLOC-1535] Update cluster local state before supplying it to calculate_necessary_state_changes

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1535

We'd also planned to remove the `local_state` argument of `calculate_necessary_state_changes`, thinking that the `cluster_state` would have all the required information. But actually, the `NodeState` attributes `running` and `not_running` and `used_ports` are not available from `cluster_state`. It is a `Deployment` and  only has `Node` records. 

There should probably be a followup issue to create a `DeploymentState` record which has `NodeStates` rather than `Nodes` and a corresponding update to the control service and AMP protocol to broadcast `DeploymentState` records.

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -266,7 +266,9 @@ class ConvergenceLoop(object):
         def got_local_state(local_state):
             # Current cluster state is likely out of date as regards the
             # local state, so update it accordingly:
-            self.cluster_state = self.cluster_state.update_node(local_state.to_node())
+            # self.cluster_state = self.cluster_state.update_node(
+            #     local_state.to_node()
+            # )
             self.client.callRemote(NodeStateCommand, node_state=local_state)
             action = self.deployer.calculate_necessary_state_changes(
                 local_state, self.configuration, self.cluster_state
@@ -291,7 +293,7 @@ class ConvergenceLoop(object):
         # https://clusterhq.atlassian.net/browse/FLOC-1357
 
 
-def build_convergence_loop_fsm(reactor, deployer, loop_factory=ConvergenceLoop):
+def build_convergence_loop_fsm(reactor, deployer):
     """
     Create a convergence loop FSM.
 
@@ -319,10 +321,10 @@ def build_convergence_loop_fsm(reactor, deployer, loop_factory=ConvergenceLoop):
             I.ITERATION_DONE: ([], S.STOPPED),
         })
 
-    loop = loop_factory(reactor, deployer)
+    loop = ConvergenceLoop(reactor, deployer)
     fsm = constructFiniteStateMachine(
         inputs=I, outputs=O, states=S, initial=S.STOPPED, table=table,
-        richInputs=[_ClientStatusUpdate], inputContext={},
+        richInputs=[_ClientStatusUpdate], inputContext ={},
         world=MethodSuffixOutputer(loop))
     loop.fsm = fsm
     return fsm
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -264,6 +264,10 @@ class ConvergenceLoop(object):
         d = self.deployer.discover_local_state()
 
         def got_local_state(local_state):
+            # Update cluster_state here and then no need to pass local state to calculate_necessary_state_changes.
+            # Avoids having to modify all IDeployers and needs only one test.
+            # Add unit test for ConvergenceLoop
+            self.cluster_state = self.cluster_state.update_node(local_state.to_node())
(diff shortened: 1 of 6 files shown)


wallrj:
Ah, there are some tests to fix. :frowning:

wallrj:
Ok. So some points for review / discussion:

 * @itamarst and @exarkun discussed the case of a convergence agents that is not node specific. In this case, what is the `local_state`?

Perhaps an example of such an agent might be an `flocker-openstack-agent` which carries out only the attach and detach of openstack block devices leaving the mounting (and if required formatting) of the volume (inside the node VM) to flocker-dataset-agent. 

But it seems to me that it'd still `discover_local_state` and that would still need merging with the cluster state received from the control service....so perhaps that's not a good example. 

 * It looks like this is going to require a ton of changes to the existing tests, which supply dummy local and cluster state `objects` instead of real `Deployment` and `NodeState` objects. Which prevents the ConvergenceLoop doing anything with those arguments.

I need some ideas about how to refactor those tests, if we choose to go ahead with this.

flocker/node/_loop.py
```diff
@@ -264,9 +264,14 @@ def output_CONVERGE(self, context):
         d = self.deployer.discover_local_state()
 
         def got_local_state(local_state):
+            # Update cluster_state here and then no need to pass local state to calculate_necessary_state_changes.
```
itamarst: Remove these comments, add back comment that used to be in P2PNodeDeployer about why we're doing this.

flocker/node/test/test_loop.py
```diff
@@ -319,15 +321,57 @@ def test_convergence_done_notify(self):
...
+        local_node_state = NodeState(
+            hostname=local_node_hostname,
+            manifestations=[discovered_manifestation]
+        )
+        client = self.successful_amp_client([local_node_state])
+        action = ControllableAction(Deferred())
+        deployer = ControllableDeployer([succeed(local_node_state)], [action])
+
+        loops = []
+        def spy_loop_factory(reactor, deployer):
```
itamarst: Instead of testing this way, I suggest passing in a `ControllableDeployer` and making sure its calculate function is called with updated state... both in this loop *and* in a future loop. The principle here is that as far as possible one should test public interfaces, and the interface in this case is that the IDeployer calculate function is called.

Additionally, `ControllableDeployer` should have its `IDeployer`-ness verified using `ideployer_tests_factory` from `test_deploy.py`.

flocker/node/_loop.py
```diff
@@ -287,7 +292,7 @@ def got_local_state(local_state):
         # https://clusterhq.atlassian.net/browse/FLOC-1357
 
 
-def build_convergence_loop_fsm(reactor, deployer):
+def build_convergence_loop_fsm(reactor, deployer, loop_factory=ConvergenceLoop):
```
itamarst: Comment elsewhere suggests this is unnecessary.

itamarst:
Fix failing tests and address comments, resubmit.

wallrj:
Made suggested changes and also moved ControllableAction to `flocker.node.testtools`.

flocker/node/test/test_loop.py
```diff
@@ -304,62 +268,118 @@ def successful_amp_client(self, local_states):
...
+    def test_convergence_done_update_local_state(self):
+        """
+        An FSM doing convergence that gets a discovery result supplies an
+        updated ``cluster_state`` to ``calculate_necessary_state_changes``.
+        """
+        local_node_hostname = u'192.0.2.123'
+        # Control service reports that this node has no manifestations.
+        received_node = Node(hostname=local_node_hostname)
+        received_cluster_state = Deployment(
+            nodes=frozenset([received_node])
```
itamarst: `frozenset` isn't necessary any more.

flocker/node/test/test_loop.py
```diff
@@ -304,62 +268,118 @@ def successful_amp_client(self, local_states):
...
+            hostname=local_node_hostname,
+            manifestations=[discovered_manifestation]
+        )
+        client = self.successful_amp_client([local_node_state])
+        action = ControllableAction(result=Deferred())
+        deployer = ControllableDeployer([succeed(local_node_state)], [action])
+
+        fsm = build_convergence_loop_fsm(Clock(), deployer)
+        fsm.receive(_ClientStatusUpdate(client=client,
+                                        configuration=object(),
```
itamarst: We decided to stop using wrong object types in tests, this should be a `Deployment` instance.

itamarst:
Looks good, address one or two minor comments then merge.

PR #1207, opened 2015-03-11
[FLOC-999] Installation instructions for CentOS

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-999


docs/indepth/installation.rst
```diff
@@ -347,10 +347,37 @@ You must also install the ZFS package repository.
...
 
-.. task:: install_flocker
+.. task:: install_flocker fedora-20
    :prompt: [root@node]#
 
-Installing ``clusterhq-flocker-node`` will automatically install Docker, but the ``docker`` service may not have been enabled or started.
+Installing on CentOS 7
+----------------------
+
+Flocker requires the latest available kernel.
```
exarkun: I'm guessing that since CentOS is a fairly stable (ie "slow moving" I suppose) distro, the chances of it introducing a new kernel that's grossly incompatible with ZFS are pretty low and we're counting on that?

What about testing things before our users get them?  I suppose that's what the centos distro-testing build configuration is about - we'll get proposed kernel updates before users of the stable distro do, and if they're broken we have some window to ship a fix...?

flocker/provision/_install.py
```diff
@@ -13,10 +13,17 @@
...
 from ._common import PackageSource
 
-ZFS_REPO = ("https://s3.amazonaws.com/archive.zfsonlinux.org/"
-            "fedora/zfs-release$(rpm -E %dist).noarch.rpm")
-CLUSTERHQ_REPO = ("https://storage.googleapis.com/archive.clusterhq.com/"
-                  "fedora/clusterhq-release$(rpm -E %dist).noarch.rpm")
+ZFS_REPO = {
+    'fedora-20': "https://s3.amazonaws.com/archive.zfsonlinux.org/"
+                 "fedora/zfs-release$(rpm -E %dist).noarch.rpm",
+    'centos-7': "http://archive.zfsonlinux.org/epel/zfs-release.el7.noarch.rpm"
```
exarkun: Do they have https?  Do they sign the package?

flocker/provision/_install.py
```diff
@@ -122,6 +129,17 @@ def task_upgrade_kernel():
...
 
 
+def task_upgrade_kernel_centos():
+    return [
+        Run.from_args([
+            "yum", "install", "-y", "kernel-devel", "kernel"]),
+        # For dkms and ... ?
+        Run.from_args([
+            "yum", "install", "-y", "epel-release"]),
+        Run.from_args(['sync'])
```
exarkun: Aren't you supposed to run it three times? :grin:

docs/indepth/installation.rst
exarkun: Oh.  But there is no distro-testing for CentOS 7.  Did we look around for an alternative to that and not find anything?  That would be sad but maybe not much we can do about it...

docs/indepth/installation.rst
```diff
@@ -347,10 +347,37 @@ You must also install the ZFS package repository.
...
+
+.. task:: upgrade_kernel_centos
+
+Flocker requires ZFS, and installing ZFS requires that the running kernel be the one that will eventually be used.
+Thus we need to reboot into the new kernel.
+
+.. prompt:: bash [root@node]#
+
+   shutdown -r now
```
exarkun: Should this also be a task?  If not then I guess these docs are diverging slightly from the things the provisioning code actually sets up cloud resources to run the acceptance tests (or, if not diverging, duplicating - albeit minorly).

flocker/provision/_rackspace.py
```diff
@@ -15,6 +16,14 @@ def provision_rackspace(node, package_source, distribution):
     """
     Provision flocker on this node.
     """
+    if distribution in ('centos-7',):
+        run(
+            username='root',
+            address=node.address,
+            commands=task_upgrade_kernel_centos(),
```
exarkun: This seems to be the Rackspace-specific provisioner.  Does the CentOS 7 kernel on the other platforms not need to be upgraded?  The docs seem to suggest the upgrade is required regardless of cloud provider.

flocker/provision/_install.py
```diff
@@ -122,6 +129,17 @@ def task_upgrade_kernel():
     ]
 
 
+def task_upgrade_kernel_centos():
+    return [
+        Run.from_args([
+            "yum", "install", "-y", "kernel-devel", "kernel"]),
+        # For dkms and ... ?
+        Run.from_args([
+            "yum", "install", "-y", "epel-release"]),
```
exarkun: I wonder how people will feel about installing this repository and the software it provides.  Presumably they're running CentOS 7 because they like it more than Fedora.  I could imagine there being some push back here, maybe?

I just talked to Luke about this and he seems to think installing from EPEL is cool though.  So nothing to worry about here.

docs/indepth/installation.rst
```diff
@@ -347,10 +347,37 @@ You must also install the ZFS package repository.
 The following commands will install the two repositories and the ``clusterhq-flocker-node`` package.
 Paste them into a root console on the target node:
 
-.. task:: install_flocker
+.. task:: install_flocker fedora-20
    :prompt: [root@node]#
 
-Installing ``clusterhq-flocker-node`` will automatically install Docker, but the ``docker`` service may not have been enabled or started.
+Installing on CentOS 7
```
exarkun: The various cloud provider-specific installation instructions all end with a suggestion to jump to the generic Fedora 20 installation instructions.  Maybe they can be changed to give a link to a generic installation instruction section that starts with a mini-index with links to sections for each OS (Fedora 20, CentOS 7 for now, Ubuntu eventually, etc).  Then folks who want non-Fedora 20 don't get confused when they reach the end of the AWS/rackspace/etc sections.

docs/indepth/installation.rst
exarkun: @wallrj showed me that in the automation we use libcloud reboot API - because, we think, fabric gets upset if you use it to reboot the machine.  That seems like a reasonable technical justification for doing things this way for now, so if that's the reason here then nothing needs to change.

flocker/provision/_rackspace.py
exarkun: Now that I've digested more of this, I realize the docs that say to upgrade the kernel are meant to support any random CentOS 7 machine.  Since we don't know what kernel such a machine will have, a blanket statement to upgrade seems alright.  And then this code is specifically to support Rackspace, and maybe we know that that system in particular does need a kernel upgrade, and we also know that the AWS and DigitalOcean CentOS 7 images don't need the upgrade so we can skip this for them.

Is that right?

exarkun:
Thanks.  This seems basically sound.  Took me a little while to get my head entirely around the context but I guess I finally did.

Some additional comments beyond what I mentioned inline:

  * I think eventually we want automated test coverage for the documentation itself.  Would it make sense to, instead of hard-coding a number of calls to various `task_` functions in the acceptance test provisioning code, write a tool to collect tasks from the Sphinx document and then run them to do the provisioning?  (more or less, anyway)  Then if they don't actually work to provision stuff the acceptance tests will fail...  modulo minor details like the `node.reboot()` vs `shutdown -r now` stuff.
  * I expect we need to do something about the quick start guide.  I suspect someone is expecting that this work would include those improvements but looking at these changes it seems like it's actually a pretty separate chunk of work (in particular because we don't actually have **any** coherent "quick start" style docs for the server-side install component).

Anyway, overall, looks fine.  So please proceed.

docs/indepth/installation.rst
tomprince: Yes.

exarkun:
> I expect we need to do something about the quick start guide. 

I talked to @robhaswell about this and he said it's fine to disregard the quick start guide with respect to this work for now.  @alicesowerby is working to figure out how we want to improve that guide and when she has worked that out we'll figure out the next steps we need to take.

flocker/provision/_install.py
tomprince: The fact that CentOS includes the `epel-release` package in its repositories suggests that many people on CentOS do install EPEL as well. Since EPEL specifically doesn't replace upstream packages, installing this only allows you to install packages you wouldn't otherwise be able to install.

flocker/provision/_rackspace.py
tomprince: This was just an oversight for AWS, and digitialocean support hasn't landed yet.

tomprince:
> I think eventually we want automated test coverage for the documentation itself. [...]

This is probably true, with a bit of emphasis on eventually.

docs/indepth/installation.rst
tomprince: The instruction don't work exactly as written for CentOS, currently (and the story didn't include cloud cases).

tomprince: I did add CentOS 7 to the list of installation methods at the beginning of the larger section.

exarkun:
Thanks.  This looks good to me.  I didn't follow the steps myself.  If you think it'd be a good idea for someone to do that, maybe @adamtheturtle will get a chance to do so in a timely manner (I'll be on planes/buses for a while).  But it seems to me it's probably fine to merge this now so go ahead with that if you're comfortable it's correct.

PR #1208, opened 2015-03-11
[FLOC-1540] Remove container API endpoint

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1540

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -504,8 +504,7 @@ class DatasetAPIUserV1(object):
         ]
     )
     @structured(
-        inputSchema={
-            '$ref': '/v1/endpoints.json#/definitions/configuration_container'},
+        inputSchema={},
         outputSchema={},
         schema_store=SCHEMAS
     )
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -500,6 +500,7 @@ class DatasetAPIUserV1(object):
         """,
         examples=[
             u"remove a container",
+            u"remove a container with unknown name",
         ]
     )
     @structured(
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -44,6 +44,8 @@ SCHEMAS = {
 CONTAINER_NAME_COLLISION = make_bad_request(
     code=CONFLICT, description=u"The container name already exists."
 )
+CONTAINER_NOT_FOUND = make_bad_request(
+    code=NOT_FOUND, description=u"Container not found.")
 DATASET_ID_COLLISION = make_bad_request(
     code=CONFLICT, description=u"The provided dataset_id is already in use.")
 PRIMARY_NODE_NOT_FOUND = make_bad_request(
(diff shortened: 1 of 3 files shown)


dwgebler:
The previous branch, create container endpoint, that this branched off has been merged now so I've merged `master` in to this branch to make review a bit easier.

flocker/control/httpapi.py
```diff
@@ -485,6 +490,49 @@ def saved(_):
...
+        restarted again.
+        """,
+        examples=[
+            u"remove a container",
+            u"remove a container with unknown name",
+        ]
+    )
+    @structured(
+        inputSchema={
+            '$ref': '/v1/endpoints.json#/definitions/configuration_container'},
```
dwgebler: :question: Is this actually required for a `DELETE` request? The tests will pass with an empty `inputSchema`. Also this is a question more broadly relating to how our JSON schemas are generated from the YAML files; specifying the input schema here would in one sense suggest that there ought to be a request body with the endpoint's defined required fields such as the container's host and image. Obviously we don't supply those and the API calls succeed on a `DELETE` without a request body and the `name` specified in the request path (which is entirely the correct behaviour I'd expect), but this is the case even with this input schema present, in which case I'm left wondering what purpose it serves here. So this is more a question of curiosity as to how this works.

dwgebler:
Thanks @itamarst the OS X build failure is expected, I've also left an inline question that would be nice to hear your thoughts on, but otherwise this looks good to merge.

flocker/control/test/test_httpapi.py
```diff
@@ -345,6 +345,90 @@ def test_response(self):
...
+        d = self.assertResponseCode(
+            b"POST", b"/configuration/containers",
+            {
+                u"host": self.NODE_A, u"name": name,
+                u"image": u"postgres"
+            }, CREATED
+        )
+        d.addCallback(lambda _: self.assertResult(
+            b"DELETE",
+            b"/configuration/containers/{}".format(name.encode("ascii")), None,
```
exarkun: FWIW:

```
Python 3.4.0 (default, Apr 11 2014, 13:05:11) 
[GCC 4.8.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> b"{}".format(b"hello")
AttributeError: 'bytes' object has no attribute 'format'
>>> u"{}".format(b"hello")
"b'hello'"
>>> u"{}".format(u"hello").encode("ascii")
b'hello'
>>> 
```

flocker/control/test/test_httpapi.py
```diff
@@ -345,6 +345,90 @@ def test_response(self):
...
+        return self.assertResult(
+            b"DELETE",
+            b"/configuration/containers/%s" % (
+                unknown_name.encode('ascii'),),
+            None, NOT_FOUND,
+            {u"description": u'Container not found.'})
+
+    def _delete_test(self, name):
+        """
+        Create and then delete a container, ensuring the expected response
```
exarkun: What response is expected?

flocker/control/test/test_httpapi.py
```diff
@@ -345,6 +345,90 @@ def test_response(self):
...
+        d.addCallback(lambda _: self.assertResult(
+            b"DELETE",
+            b"/configuration/containers/{}".format(name.encode("ascii")), None,
+            OK, None,
+        ))
+        return d
+
+    def test_delete(self):
+        """
+        The ``DELETE`` action removes the given container from the
```
exarkun: minor point - the HTTP spec calls them methods, not actions.

PR #1212, opened 2015-03-12
[FLOC-1495] IBlockDeviceAPI and a Loopback implementation

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1495

Originally developed in https://github.com/ClusterHQ/flocker/compare/loopback-create-FLOC-1502 but broken out for ease of review.

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -261,7 +261,9 @@ class LoopbackBlockDeviceAPI(object):
                 host_directory.makedirs()
             except OSError:
                 pass
-            old_path.moveTo(host_directory.child(blockdevice_id))
+            new_path = host_directory.child(blockdevice_id)
+            old_path.moveTo(new_path)
+            check_output(["losetup", "--find", new_path.path])
             attached_volume = volume.set(host=host)
             return attached_volume
 
@@ -304,7 +306,5 @@ class LoopbackBlockDeviceAPI(object):
         volume_path = self._attached_directory.descendant(
             [volume.host, volume.blockdevice_id]
         )
-        device_path = device_for_path(volume_path)
-
-        if device_path is not None:
-            return FilePath(device_path)
+        # May be None if the file hasn't been used for a loop device.
+        return device_for_path(volume_path)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -240,12 +240,12 @@ class LoopbackBlockDeviceAPI(object):
         if volume.host is None:
             old_path = self._unattached_directory.child(blockdevice_id)
             host_directory = self._attached_directory.child(host)
-            attached_volume = volume.set(host=host)
             try:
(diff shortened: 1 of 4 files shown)


wallrj:
Ok. I think this is ready for a review, but there are a couple of questions:
 
1. On reflection, `IBlockDeviceAPI.get_device_path` seems wrong. The openstack API requires you to supply a device path when you attach a block device. So, assuming that BlockDeviceAPI is going to be run on the node, we should just roll the selection of the device path into the `attach_volume` method and add the selected device_path to the returned `BlockDeviceVolume` instance.

2. losetup on Ubuntu1404 doesn't support the `--columns` option, so if we're going to support that platform, we'll have to add a little more parsing.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+        List all the block devices available via the back end API.
+
+        :returns: A ``list`` of ``BlockDeviceVolume``s.
+        """
+
+    def get_device_path(blockdevice_id):
+        """
+        Calculate the path at which ``blockdevice_id`` will be exposed on the
+        host when attached.
```
wallrj: I think this part of the interface is wrong. Device path should be either supplied to ``attach_volume`` or it should be chosen automatically by ``attach_volume``, but that'll only work if the provider of this interface is run on host to which the block device is about to be attached.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,270 @@
...
+    :ivar unicode blockdevice_id: The unique identifier of the block device.
+    :ivar int size: The size of the block device.
+    :ivar bytes host: The IP address of the host to which the block device is
+        attached or ``None`` if it is currently unattached.
+    """
+    blockdevice_id = field(type=unicode, mandatory=True)
+    size = field(type=int, mandatory=True)
+    # XXX: Should be hostname, for consistency of is host a better name since
+    # we currently only expect IP addresses?
+    host = field(type=(bytes, type(None)), initial=None)
```
wallrj: ...and we probably need to add an optional ``device_path`` attribute here.

flocker/node/agents/blockdevice.py
wallrj: I think I've changed my mind again. What confused me was that we were creating the loopback device here in our `LoopbackBlockDeviceAPI`, which seems unrealistic.

With libcloud interface to OpenStack blocks, you have to supply the device path when you attach the device.
But that can be chosen automatically by the OpenStack implementation of BlockDeviceAPI.

Then `OpenStackBlockDeviceAPI.get_device_path` will be able to get that path from the libcloud API.

flocker/node/agents/blockdevice.py
wallrj: I tried this, but it makes the tests difficult, because the device_path can't be predicted...and in trying it this way, I've convinced myself that the original get_device_path API is correct.

PR #1214, opened 2015-03-12
[FLOC-1544] Create container with environment

dwgebler:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1544


docs/advanced/api_examples.yml
```diff
@@ -281,6 +281,29 @@
...
+  doc: |
+    Create a new container.
+
+  request: |
+    POST /v1/configuration/containers HTTP/1.1
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "environment": {"CONFIG_FILE": "/etc/nginx/nginx.conf", "USER": "webserver"}}
```
wallrj: I was curious if these environment variables were actually supported by the official Nginx docker image. It looks like they're not.
 * https://github.com/nginxinc/docker-nginx/blob/master/Dockerfile

The `CONFIG_FILE` variable seems realistic. 
But the `USER` seems a little contrived. Why would an administrator care which user the webserver runs as inside the container. 
Maybe remove that one.

docs/advanced/api_examples.yml
```diff
@@ -281,6 +281,29 @@
 
 -
   id:
+    "create container with environment"
+
+  doc: |
+    Create a new container.
```
wallrj: This isn't unique. And I think this prevents it being shown in the docs:
 * http://build.clusterhq.com/results/docs/create-container-with-environment-FLOC-1544/advanced/api.html

wallrj: That should be flagged by our documentation builder. Create an issue for that.

docs/advanced/api_examples.yml
```diff
@@ -281,6 +281,29 @@
...
+     "image": "nginx:latest",
+     "environment": {"CONFIG_FILE": "/etc/nginx/nginx.conf", "USER": "webserver"}}
+
+  response: |
+    HTTP/1.1 201 Created
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "environment": {"CONFIG_FILE": "/etc/nginx/nginx.conf", "USER": "webserver"}}
```
wallrj: This pretty printed response isn't strictly what the server will return. 
I see that the other container API examples are the same.
But in the future, when the we test these examples automaticallyu, will they pass or fail.
It also makes me think that the example was hand crafted rather than actually tested.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,57 @@ def can_connect():
...
+        waiting_for_cluster = wait_for_cluster(test_case=self, node_count=1)
+
+        def create_container(cluster, data):
+            data[u"host"] = cluster.nodes[0].address
+            return cluster.create_container(data)
+
+        d = waiting_for_cluster.addCallback(create_container, data)
+
+        def check_result(result):
+            response = result[1]
```
wallrj: I think it would be slightly clearer to unpack the cluster, response in the callback signature eg

```
def check_result((cluster, response)):
    ...
```

And that ensures that you'd get an error if the result ever contains an unexpected number of parts.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,57 @@ def can_connect():
...
+                try:
+                    remote_data = run_SSH(
+                        22, 'root', data[u"host"],
+                        [
+                            b"docker",
+                            b"inspect",
+                            b"flocker--{0}".format(data[u"name"])
+                        ], None
+                    )
+                except Exception:
```
wallrj: Can this be a specific exception? Maybe `CalledProcessError` ?

docs/advanced/api_examples.yml
dwgebler: :question: The actual response body from the server doesn't have the line breaks and may have the keys in a slightly different order, but I think that's just a documentation convenience.

Actual response as bytes:

```
{"environment": {"CONFIG_FILE": "/etc/nginx/nginx.conf", "USER": "webserver"}, "host": "192.0.2.2", "name": "nginx", "image": "nginx:latest"}
```

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,57 @@ def can_connect():
...
+                    return False
+                container_data = loads(remote_data)
+                env_data = container_data[0]['Config']['Env']
+                for var in env_data:
+                    try:
+                        label, value = var.split("=")
+                        if [(label, value)] == data[u"environment"].items():
+                            return True
+                    except ValueError:
```
wallrj: I guess the `ValueError` occurs if the split yields too few or too many parts, so move the except clause directly beneath the `label`, `value` unpacking.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,57 @@ def can_connect():
...
+                        continue
+                return False
+            dl = loop_until(can_connect)
+            self.assertEqual(response, data)
```
wallrj: Is this assertion in the right place? Surely it need to be performed only after the `loop_until` result fires.

flocker/control/httpapi.py
```diff
@@ -457,6 +459,10 @@ def create_container_configuration(self, host, name, image, ports=()):
         :param list ports: A ``list`` of ``dict`` objects, mapping internal
             to external ports for the container.
 
+        :param dict environment: A ``dict`` of key/value pairs to be supplied
+            to the container as environment variables. Values must be
+            ``unicode``.
```
wallrj: Keys have to be unicode too, right?

flocker/control/httpapi.py
```diff
@@ -492,11 +498,15 @@ def create_container_configuration(self, host, name, image, ports=()):
                 external_port=port['external']
             ))
 
+        if environment:
```
wallrj: Better to specifically check `if environment is not None` ie not the placeholder value in the function signature. Other Falsy values should be  used and cause a failure if they happen to be supplied.

dwgebler: :question: Surely if `environment` is not `None`, it's guaranteed to be a dictionary because the JSON schema type checking has already been enforced before we get to the endpoint?

flocker/control/schema/types.yml
```diff
@@ -59,7 +59,17 @@ definitions:
...
           type: integer
     uniqueItems: true
-    
+
+  environment:
+    title: "Container environment variables"
+    description: "Mapping of label and value pairs for environment variables"
+    type: object
+    patternProperties:
+      "^.{1,256}$":
```
wallrj: Is this for the environment variable key or for the value or both?

It seems a little too permissive. 

I checked HybridCluster schema for similar types, but couldn't find anything. But double check this.

flocker/acceptance/test_api.py
dwgebler: Not without modifying `run_ssh` in `testtools`, which I think is outside the scope of this issue:

```
raise Exception('Command Failed', command, process.returncode)
```

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,79 @@ def test_container_name_collision_different_node(self):
...
+        """
+        saving = self.persistence_service.save(Deployment(
+            nodes={
+                Node(
+                    hostname=self.NODE_A,
+                    applications=[]
+                ),
+                Node(hostname=self.NODE_B),
+            }
+        ))
```
wallrj: What's this bit for? To wipe out the configuration before performing the test? 
Add an explanatory comment.
Move this to `setUp` ?

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,79 @@ def test_container_name_collision_different_node(self):
...
+
+        environment = {
+            u'USER': u'webserver',
+            u'CONFIG_FILE': u'/etc/nginx/nginx.conf',
+        }
+
+        saving.addCallback(lambda _: self.assertResponseCode(
+            b"POST", b"/configuration/containers",
+            {
+                u"host": self.NODE_A, u"name": u"nginx",
```
wallrj: You used the name "webserver" elsewhere. That might be clearer here too.

flocker/control/test/test_schemas.py
```diff
@@ -98,6 +98,23 @@
...
+        # Environment given but at least one entry is not a string
+        {
+            'host': '192.168.0.3',
+            'image': 'postgres',
+            'name': 'postgres',
+            'environment': {
+                'POSTGRES_USER': 'admin',
+                'POSTGRES_VERSION': 9.4
+            }
+        },
```
wallrj: I think the schema limits the key / value to 256 characters. Add a test for that.

flocker/acceptance/test_api.py
dwgebler: No, this assertion doesn't rely on anything that happens in the `loop_until` call; this is just checking the API response matches up to the supplied and expected JSON correctly. The separate looping call ensures that the test will either succeed when it can connect via SSH and inspect the container and finds a match to the injected environment variable keys and values, or the whole test will just continuously run until it times out through trial.

wallrj: Ok. But I think `loop_until` returns the result of the predicate when it's not `False`, so if `can_connect` returned the result of the successful `docker inspect` then you'd be able to make an assertion once the container is found.

Otherwise, in the case where the environment variables returned by docker inspect do not match, the test just loops for ever and eventually times out without any meaningful error.

wallrj:
Thanks @dwgebler 

 * I ran the acceptance tests and saw them pass.

I made some comments above. Please answer or address those and resubmit for final review.

flocker/control/test/test_httpapi.py
dwgebler: Yes, but some of the tests in this class use a different initial cluster configuration, so I won't move it to `setUp`, however the `applications=[]` is unnecessary here and will be removed.

flocker/control/schema/types.yml
dwgebler: It's for both and I now have JSON schema tests to show that. I'm not aware of what characters might not be allowed in an environment variable so I don't know what other constraints I could put on this; I used the same pattern for generic properties (JSON schema equivalent of an arbitrary dictionary) as we use in our schema for dataset metadata.

flocker/control/test/test_schemas.py
dwgebler: Tests added, they're just below these ones in the diff

flocker/control/schema/types.yml
wallrj: Yeah, ok. There doesn't seem to be any standard so I suppose it's fine to let the user supply that they like.

http://stackoverflow.com/questions/2821043/allowed-characters-in-linux-environment-variable-names

But I suppose it's also limited to characters that are allowed in JSON keys...I'm not sure what that is.

The 256 character limit may be a little low though, especially for environment variable values which may contain things like public ssh keys.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+                    22, 'root', data[u"host"],
+                    [
+                        b"docker",
+                        b"inspect",
+                        b"flocker--{0}".format(data[u"name"])
+                    ], None
+                )
+            except Exception:
+                return False
+            container_data = loads(remote_data)
```
wallrj: Personally, I'd `return container_data` here and perform all the environment variable parsing in the `verify_environment` function below.

As it stands, `loop_until` will treat an empty dictionary as a signal to loop again, so if the container has somehow been started without any environment or if inspect incorrectly returns an empty environment, the test will time out without a meaningful error.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+            for var in env_data:
+                try:
+                    label, value = var.split("=")
+                except ValueError:
+                    continue
+                result[label] = value
+            return result
+
+        def verify_environment(actual, expected):
+            for key, value in expected.iteritems():
```
wallrj: Why not just compare the dictionaries here? That way trial can present a nicely formatted, single failure result in case of item mismatches.

As I said above, I'd move the environment parsing here and then finish with a single `assertEqual` of the expected and actual environment dicts.

flocker/control/httpapi.py
wallrj: I agree that json schema validation should prevent anything other than a dictionary here, but I still think it's better to be explicit about what's expected.

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,76 @@ def test_container_name_collision_different_node(self):
         """
         return self._container_name_collision_test(self.NODE_A, self.NODE_B)
 
+    def test_create_container_with_environment(self):
+        """
+        A valid API request to create a container including environment
+        variables results in an updated configuration.
+        """
+        saving = self.persistence_service.save(Deployment(
```
wallrj: I think I asked this before, but what's the point of this initial step? Seems to me that the configuration should be clean /empty  at the beginning of each test.
And remember, that users are allowed to request a new container on a node that has not (as yet) been seen.

I think this can be removed, but if not, then add an explanatory comment.

wallrj: Oh, perhaps I understand. We're testing that *existing* configuration is updated rather than replaced, right?

wallrj: I'd have thought that behaviour would be tested, once, separately, not for every variation of the create container parameters.

But if you want to keep it this way, perhaps it could be made clearer by assigning to a variable called `original_configuration` and then using the `Deployment.update_node` method to derive the expected `updated_configuration`. Up to you. 

But at least make the test docstring a little clearer.

dwgebler: Right :)

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,76 @@ def test_container_name_collision_different_node(self):
...
+            nodes={
+                Node(hostname=self.NODE_A),
+                Node(hostname=self.NODE_B),
+            }
+        ))
+
+        environment = {
+            u'USER': u'webserver',
```
wallrj: Maybe remove or change the `USER` env variable which seems a little contrived (as mentioned in previous review)

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,76 @@ def test_container_name_collision_different_node(self):
...
+                }
+            )
+            self.assertEqual(deployment, expected)
+
+        saving.addCallback(created)
+        return saving
+
+    def test_create_container_with_environment_response(self):
+        """
+        A valid API request to create a container including environment
```
wallrj: Maybe remove the word "valid" here and above. It seems unnecessary.

flocker/acceptance/test_api.py
dwgebler: The actual environment dict will contain a bunch of keys (or some arbitrary number of keys) that are not part of expected; expected only contains the key/value pairs we've injected in to the container, whereas actual will have potentially other stuff too (usually at least PATH), e.g. they might look like:

```
expected: {"CONFIG_PATH": "/etc/nginx/nginx.conf"}

actual: {"CONFIG_PATH": "/etc/nginx/nginx.conf", "PATH": "/usr/local/bin;/etc..."}

File: flocker/acceptance/test_api.py [hunk 19]
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+                    ], None
+                )
+            except Exception:
+                return False
+            container_data = loads(remote_data)
+            env_data = container_data[0]['Config']['Env']
+            result = dict()
+            for var in env_data:
+                try:
+                    label, value = var.split("=")
```
wallrj: Since we allow `=` in the value, I think this should be `var.split("=", 1)` and add an `=` to the environment variable that you use in the test, to check that it's allowed through.

flocker/acceptance/test_api.py
wallrj: Ah, ok. Well add a comment to that effect, or rustle up an `assertDictContians` method.

wallrj:
Thanks @dwgebler 

I left some further comments above.

We discussed in the standup last week, that acceptance tests will be be simplified when you have a container list API. Isn't it worth doing that next?

Anyway, please answer or address the comments as you see fit and then merge.

flocker/control/httpapi.py
exarkun: Why have this check at all then?  It looks like it can only introduce complexity - and an extra bug.  If `environment` is `{}` then it won't get converted to a frozenset, so it's passed to `Application` as a dict.  Maybe that doesn't break anything, but why go out of our way to do it?

exarkun: I suspect it actually can be `None` though since it's not defined as required by the schema.  If no environment is supplied at all then it should come through as `None`.

flocker/control/httpapi.py
```diff
@@ -576,6 +587,8 @@ def container_configuration_response(application, node):
             result['ports'].append(dict(
                 internal=port.internal_port, external=port.external_port
             ))
+    if application.environment:
```
exarkun: Similar comments about truth testing vs identity check against `None` apply here as well.

flocker/control/schema/types.yml
```diff
@@ -59,7 +59,17 @@ definitions:
...
+
+  environment:
+    title: "Container environment variables"
+    description: "Mapping of label and value pairs for environment variables"
+    type: object
+    patternProperties:
+      "^.{1,4096}$":
+        type: string
+        maxLength: 4096
+    additionalProperties: false
```
exarkun: This schema doesn't limit the number of properties on the object - so the total size of the environment is not limited.  Limiting the total size of the environment seems like the primary motivation to have any limits at all in the schema here.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+                remote_data = run_SSH(
+                    22, 'root', data[u"host"],
+                    [
+                        b"docker",
+                        b"inspect",
+                        b"flocker--{0}".format(data[u"name"])
+                    ], None
+                )
+            except Exception:
+                return False
```
exarkun: This is a really bad pattern to use.  Don't ever handle `Exception` and then squash it down to a simple value like `False`.  Always handle the most specific exception type you can.  `Exception` is almost as far on the opposite end of this suggestion as you can get.  The exception that actually gets handled here could be anything from what you expected (what did you expect, by the way?) to a `NameError` because the function name you expected to call was misspelled to a `RuntimeError` because hardware failed and a random bit of memory got flipped.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+
+        def inspect_container():
+            try:
+                remote_data = run_SSH(
+                    22, 'root', data[u"host"],
+                    [
+                        b"docker",
+                        b"inspect",
+                        b"flocker--{0}".format(data[u"name"])
+                    ], None
```
exarkun: This test side-steps the public interface of the thing being tested.  We discussed some of the downsides of this kind of testing last week and tentatively agreed to stop doing things this way.

So... what is the public interface of the thing being tested?  It's ... containers.  Which is a pretty wide interface.  You could run a container that runs a web server that spits out its environment in response to a request and then you could issue a request to it.

The advantage of writing the test that way is that when we change everything around - for example, we get rid of automatic SSH access or we change the naming convention for containers - the test will still work.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+            except Exception:
+                return False
+            container_data = loads(remote_data)
+            env_data = container_data[0]['Config']['Env']
+            result = dict()
+            for var in env_data:
+                try:
+                    label, value = var.split("=")
+                except ValueError:
+                    continue
```
exarkun: This drops part of the environment for no particular reason.  Just `split("=", 1)` to avoid the `ValueError` when the environment variable contains extra `=`.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,61 @@ def can_connect():
...
+                result[label] = value
+            return result
+
+        def verify_environment(actual, expected):
+            for key, value in expected.iteritems():
+                self.assertEqual(actual[key], value)
```
exarkun: Assertions in a loop like this make for very difficult to debug test failures.  By doing an assertion in a loop, if one assertion fails you don't get to see the results of any later iterations (the normal problem with multiple assertions in one test).  And on top of that, the exception that gets raised doesn't tell you which key the problem was with - so you don't really know how the test failed, only that it did.

Also, in this case, it seems like there's only actually one item in `expected` anyway.  So it would probably be better to just make an assertion about that one specific item.  Alternatively, use one of the useful assertion helpers like `assertDictContainsSubset`.

flocker/control/httpapi.py
```diff
@@ -431,7 +431,8 @@ def state_datasets(self):
         examples=[
             u"create container",
             u"create container with duplicate name",
-            u"create container with ports"
+            u"create container with ports",
+            u"create container with environment"
```
exarkun: Put a trailing comma on this line **now** and the line won't show up as modified when you add an example for the next branch. :wink: :question:

flocker/control/test/test_schemas.py
```diff
@@ -98,6 +98,41 @@
...
+        # Environment given but a key is too long (>4096 characters)
+        {
+            'host': '192.168.0.3',
+            'image': 'postgres',
+            'name': 'postgres',
+            'environment': {
+                'MY_'+('X'*4094): 'somevalue',
```
exarkun: whitespace around operators.  Also, this construct seems needlessly complicated.  Why not just `'X' * 4097`?

flocker/acceptance/test_api.py
dwgebler: It's just for `loop_until` - the `run_ssh` method throws `Exception` explicitly, so I'm looping until the `docker inspect` command succeeds.

flocker/acceptance/test_api.py
dwgebler: @exarkun this acceptance test is temporary until we've implemented the endpoint to retrieve information about a container. Until we have that endpoint, the test does technically do its job of verifying the presence of injected environment variables. Is it really worth creating a custom Docker image to run a webserver and rely on the output of that server being a correct representation of the environment for a test we're going to throw away and replace in the very near future anyway?

flocker/control/schema/types.yml
dwgebler: @exarkun save me some Googling :smile:  how does one limit the number of arbitrary properties in a JSON schema object? And what would you regard as a sensible limit?

flocker/acceptance/test_api.py
exarkun: The problem is that just about anything else that raises an exception is *also* going to raise `Exception`.  `run_SSH` is a terrible API.  It is doing a terrible thing raising `Exception`.  But this is layering on more terrible.

flocker/acceptance/test_api.py
exarkun: I'm not sure that's what we'll actually do.  The "list containers" endpoint isn't sufficient to demonstrate containers are actually being run.  It just tells you what the cluster thinks is running.  I suspect the test I described is the one we actually want to have in the long run - in addition to the internal-consistency test you're thinking of that only uses the "create container" and "list containers" endpoints.

flocker/control/schema/types.yml
exarkun: `maxProperties` - `metadata` in the types schema uses this if you want an example.

There's some maximum environment size supported by `exec(3)`.  Picking a limit at or below that is probably sensible.  I don't think the limit is actually defined by any standard, though (so it's a case of "whatever random value some random version of Linux picked", unfortunately).  On 3.13, seems to be in the ballpark of 2 MiB.  If we stick to 4096 for key and value limit, that's 256 items.  So let's go with that.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,54 @@ def can_connect():
...
+        def create_container(cluster, data):
+            data[u"host"] = cluster.nodes[0].address
+            return cluster.create_container(data)
+
+        d = waiting_for_cluster.addCallback(create_container, data)
+
+        def check_result((cluster, response)):
+            self.assertEqual(response, data)
+
+        def inspect_container(host, port):
```
itamarst: Function name is no longer accurate.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,54 @@ def can_connect():
...
+            def can_connect():
+                s = socket.socket()
+                conn = s.connect_ex((host, port))
+                return False if conn else True
+
+            dl = loop_until(can_connect)
+            return dl
+
+        def query_environment(host, port):
```
itamarst: Function name is no longer accurate.

itamarst: Or at least you should have comment explaning why this request gives you back environment.

itamarst: Or at least you should have comment explaning why this request gives you back environment.

flocker/acceptance/test_api.py
```diff
@@ -369,6 +369,54 @@ def can_connect():
...
+            req = get(
+                "http://{host}:{port}".format(host=host, port=port),
+                persistent=False
+            ).addCallback(json_content)
+            return req
+
+        d.addCallback(check_result)
+        d.addCallback(lambda _: inspect_container(data[u"host"], 8080))
+        req = d.addCallback(lambda _: query_environment(data[u"host"], 8080))
```
itamarst: Don't really need the `req =` bit, can just stick to `d`.

itamarst:
A few really minor comments - address and merge, since Buildbot seems happy.

itamarst:
Oh, actually: I just merged branch that turned `Application.environment` into a dictionary-like object. So you should also merge forward and make appropriate tweaks.

PR #1215, opened 2015-03-13
[FLOC-1546] Clean up test-brew scripts

jongiddy:


adamtheturtle:
Fixes FLOC-1546

admin/test-brew-recipe
```diff
@@ -1,11 +1,9 @@
 #!/usr/bin/env python
-# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# Copyright 2015 ClusterHQ.  See LICENSE file for details.
```
adamtheturtle: :question: Probably don't have the year. Is it elsewhere / standard / necessary? Is there a method for updating it annually? Or an issue to create one?

adamtheturtle:
Optionally address the comment and merge.

admin/test-brew-recipe
```diff
@@ -1,11 +1,9 @@
...
 """
 Test a brew recipe generated by Buildbot.
 """
 
-from _preamble import TOPLEVEL, BASEPATH
```
adamtheturtle: These are used in other scripts to provide context for error messages. Perhaps they should be kept and that context should be added to the relevant scripts./

PR #1216, opened 2015-03-16
[FLOC-1547] Sort the expected logged errors before comparing

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1547

I couldn't reproduce the thread ordering problem locally, so introduced some time delays to force the issue.

I should probably remove those delays before merging.

Or alternatively, I could somehow chain the firing of failures so that the errors are logged in predictable order.

But probably not worth spending time on.

### DIFF
--- a/flocker/cli/functional/test_deploy_script.py
+++ b/flocker/cli/functional/test_deploy_script.py
@@ -4,6 +4,7 @@
 Functional tests for the ``flocker-deploy`` command line tool.
 """
 from subprocess import check_output, CalledProcessError
+import time
 from unittest import skipUnless
 
 from twisted.python.procutils import which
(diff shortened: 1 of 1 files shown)


adamtheturtle:
> Or alternatively, I could somehow chain the firing of failures so that the errors are logged in predictable order.
> But probably not worth spending time on.

Is there any advantage at all in practice to having a predictable order?

adamtheturtle:
> I should probably remove those delays before merging.

Yes.

Apart from this I can't think of any advantages to chaining the firing of failures. Unless you can think of some important ones address your own comment and merge. Perhaps add a comment above the assertion about why the lists are sorted. If you can think of an advantage, either chain the failures and resubmit or create a follow-up issue.

PR #1217, opened 2015-03-16
[FLOC-1490, FLOC-1494] Initial IBlockDeviceAPI and an initial Loopback implementation for creating, attaching and listing block devices

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1490
Fixes https://clusterhq.atlassian.net/browse/FLOC-1494

This introduces an (incomplete) interface for block-device backends.  It also implements that interface using a local-only loopback block device provider.  This was split out of a larger branch which also includes new `IDeployer` implementation that knows how to use this new interface to provide block-device-based storage for a datasets convergence agent (that code to be up for review shortly, you can see it in https://github.com/ClusterHQ/flocker/tree/iblockdevice-deployer-FLOC-1502 (currently contains all of this code and some more code besides)).

The new interface and implementation also covers the "list existing block devices" functionality that's required but there's no separate JIRA issue for that work.

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -244,7 +244,8 @@ class LoopbackBlockDeviceAPI(object):
 
     def create_volume(self, size):
         """
-        Create a file of some size and put it in the ``unattached`` directory.
+        Create a "sparse" file of some size and put it in the ``unattached``
+        directory.
 
         See ``IBlockDeviceAPI.create_volume`` for parameter and return type
         documentation.
@@ -253,9 +254,10 @@ class LoopbackBlockDeviceAPI(object):
             blockdevice_id=unicode(uuid4()),
             size=size,
         )
-        self._unattached_directory.child(
+        with self._unattached_directory.child(
             volume.blockdevice_id.encode('ascii')
-        ).setContent(b'\0' * volume.size)
+        ).open('wb') as f:
+            f.truncate(size)
         return volume
 
     def _get(self, blockdevice_id):
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -261,7 +261,9 @@ class LoopbackBlockDeviceAPI(object):
                 host_directory.makedirs()
             except OSError:
                 pass
-            old_path.moveTo(host_directory.child(blockdevice_id))
+            new_path = host_directory.child(blockdevice_id)
+            old_path.moveTo(new_path)
+            check_output(["losetup", "--find", new_path.path])
             attached_volume = volume.set(host=host)
             return attached_volume
 
@@ -304,7 +306,5 @@ class LoopbackBlockDeviceAPI(object):
         volume_path = self._attached_directory.descendant(
             [volume.host, volume.blockdevice_id]
         )
-        device_path = device_for_path(volume_path)
-
-        if device_path is not None:
-            return FilePath(device_path)
+        # May be None if the file hasn't been used for a loop device.
+        return device_for_path(volume_path)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -284,7 +284,9 @@ class LoopbackBlockDeviceAPI(object):
         volume = self._get(blockdevice_id)
         if volume.host is None:
             old_path = self._unattached_directory.child(blockdevice_id)
-            host_directory = self._attached_directory.child(host.encode("utf-8"))
+            host_directory = self._attached_directory.child(
+                host.encode("utf-8")
+            )
             try:
                 host_directory.makedirs()
             except OSError:
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -240,12 +240,12 @@ class LoopbackBlockDeviceAPI(object):
         if volume.host is None:
             old_path = self._unattached_directory.child(blockdevice_id)
             host_directory = self._attached_directory.child(host)
-            attached_volume = volume.set(host=host)
             try:
                 host_directory.makedirs()
             except OSError:
                 pass
             old_path.moveTo(host_directory.child(blockdevice_id))
+            attached_volume = volume.set(host=host)
             return attached_volume
 
         raise AlreadyAttachedVolume(blockdevice_id)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -47,6 +47,13 @@ class AlreadyAttachedVolume(VolumeException):
     """
 
 
+class UnattachedVolume(VolumeException):
+    """
+    An attempt was made to operate on an unattached volume but the operation
+    requires the volume to be attached.
+    """
+
+
 class IBlockDeviceAPI(Interface):
     """
     Common operations provided by all block device backends.
@@ -78,6 +85,16 @@ class IBlockDeviceAPI(Interface):
         List all the block devices available via the back end API.
 
         :returns: A ``list`` of ``BlockDeviceVolume``s.
+        """
+
+    def get_device_path(blockdevice_id):
+        """
+        Calculate the path at which ``blockdevice_id`` will be exposed on the
+        host when attached.
+
+        :param unicode blockdevice_id: The unique identifier for the block
+            device.
+        :returns: A ``FilePath`` for the device.
         """
 
 
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -81,7 +81,7 @@ class IBlockDeviceAPI(Interface):
 
         :param unicode blockdevice_id: The unique identifier for the block
             device being attached.
-        :param bytes host: The IP address of a host to attach the volume to.
+        :param unicode host: The IP address of a host to attach the volume to.
         :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
             exist.
         :raises AlreadyAttachedVolume: If the supplied ``blockdevice_id`` is
@@ -118,12 +118,12 @@ class BlockDeviceVolume(PRecord):
 
     :ivar unicode blockdevice_id: The unique identifier of the block device.
     :ivar int size: The size, in bytes, of the block device.
-    :ivar bytes host: The IP address of the host to which the block device is
+    :ivar unicode host: The IP address of the host to which the block device is
         attached or ``None`` if it is currently unattached.
     """
     blockdevice_id = field(type=unicode, mandatory=True)
     size = field(type=int, mandatory=True)
-    host = field(type=(bytes, type(None)), initial=None)
+    host = field(type=(unicode, type(None)), initial=None)
 
 
 def _losetup_list_parse(output):
@@ -284,7 +284,7 @@ class LoopbackBlockDeviceAPI(object):
         volume = self._get(blockdevice_id)
         if volume.host is None:
             old_path = self._unattached_directory.child(blockdevice_id)
-            host_directory = self._attached_directory.child(host)
+            host_directory = self._attached_directory.child(host.encode("utf-8"))
             try:
                 host_directory.makedirs()
             except OSError:
@@ -316,7 +316,7 @@ class LoopbackBlockDeviceAPI(object):
             volumes.append(volume)
 
         for host_directory in self._root_path.child('attached').children():
-            host_name = host_directory.basename().encode('ascii')
+            host_name = host_directory.basename().decode('ascii')
             for child in host_directory.children():
 
                 volume = BlockDeviceVolume(
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -82,6 +82,7 @@ class IBlockDeviceAPI(Interface):
 
         :param unicode blockdevice_id: The unique identifier for the block
             device being attached.
+        :param bytes host: The IP address of a host to attach the volume to.
         :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
             exist.
         :raises AlreadyAttachedVolume: If the supplied ``blockdevice_id`` is
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -268,7 +268,15 @@ class LoopbackBlockDeviceAPI(object):
 
     def attach_volume(self, blockdevice_id, host):
         """
-        Move an existing ``unattached`` file into a per-host directory.
+        Move an existing ``unattached`` file into a per-host directory and
+        create a loopback device backed by that file.
+
+        Note: Although `mkfs` can format files directly and `mount` can mount
+        files directly (with the `-o loop` option), we want to simulate a real
+        block device which will be allocated a real block device file on the
+        host to which it is attached. This allows the consumer of this API to
+        perform formatting and mount operations exactly the same as for a real
+        block device.
 
         See ``IBlockDeviceAPI.attach_volume`` for parameter and return type
         documentation.
@@ -283,6 +291,8 @@ class LoopbackBlockDeviceAPI(object):
                 pass
             new_path = host_directory.child(blockdevice_id)
             old_path.moveTo(new_path)
+            # The --find option allocates the next available /dev/loopX device
+            # name to the device.
             check_output(["losetup", "--find", new_path.path])
             attached_volume = volume.set(host=host)
             return attached_volume
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -114,23 +114,70 @@ class BlockDeviceVolume(PRecord):
     host = field(type=(bytes, type(None)), initial=None)
 
 
-def device_for_path(file_path):
+def losetup_list_parse(output):
     """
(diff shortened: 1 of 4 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+
+
+class UnattachedVolume(VolumeException):
+    """
+    An attempt was made to operate on an unattached volume but the operation
+    requires the volume to be attached.
+    """
+
+
+# TODO: Introduce a non-blocking version of this interface and an automatic
```
itamarst: Link to followup issue.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+# thread-based wrapper for adapting this to the other.  Use that interface
+# anywhere being non-blocking is important (which is probably lots of places).
+class IBlockDeviceAPI(Interface):
+    """
+    Common operations provided by all block device backends.
+    """
+    def create_volume(size):
```
itamarst: How many block device APIs let you specify 500 byte sizes? Just saying :grin:

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+
+        :param int size: The size of the new block device in bytes.
+        :returns: A ``BlockDeviceVolume``.
+        """
+
+    def attach_volume(blockdevice_id, host):
+        """
+        Attach ``blockdevice_id`` to ``host``.
+
+        :param unicode blockdevice_id: The unique identifier for the block
```
itamarst: Document `host` parameter.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+    def list_volumes():
+        """
+        List all the block devices available via the back end API.
+
+        :returns: A ``list`` of ``BlockDeviceVolume``s.
+        """
+
+    def get_device_path(blockdevice_id):
+        """
+        Calculate the path at which ``blockdevice_id`` will be exposed on the
```
itamarst: I am skeptical of this method's specification. I suspect in some cases the device path will vary depending on what devices are already attached to a specific host.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+        :returns: A ``FilePath`` for the device.
+        """
+
+
+class BlockDeviceVolume(PRecord):
+    """
+    A block device that may be attached to a host.
+
+    :ivar unicode blockdevice_id: The unique identifier of the block device.
+    :ivar int size: The size of the block device.
```
itamarst: Presumably "in bytes".

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+    :ivar bytes host: The IP address of the host to which the block device is
+        attached or ``None`` if it is currently unattached.
+    """
+    blockdevice_id = field(type=unicode, mandatory=True)
+    size = field(type=int, mandatory=True)
+    # XXX: Should be hostname, for consistency of is host a better name since
```
itamarst: Neither, specifying by IP is probably the wrong thing to do :grin: There's no reason to expect them to be stable, so e.g. AWS nodes have an identity different than IP.

We can fix this later though.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+    # we currently only expect IP addresses?
+    host = field(type=(bytes, type(None)), initial=None)
+
+
```
itamarst: Loopback is presumably a *specific* kind of block device. This module's docstring says it's for code that is shared by all block device backends, so presumably loopback code belongs elsewhere.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+def losetup_list_parse(output):
```
itamarst: This should be private function.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+       supplied ``IBlockDeviceAPI`` provider.
+    """
+    class Tests(IBlockDeviceAPITestsMixin, SynchronousTestCase):
+        def setUp(self):
+            self.api = blockdevice_api_factory(test_case=self)
+
+    return Tests
+
+
+def losetup_detach(device_file):
```
itamarst: This should be private function.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+
+        # Remove the space that may have been between the path and the deleted
+        # flag.
+        backing_file = backing_file.rstrip()
+        backing_file = FilePath(backing_file)
+        devices.append((device_file, backing_file))
+    return devices
+
+
+def losetup_list():
```
itamarst: And this should be private.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+    :returns: A ``list`` of
+        2-tuple(FilePath(device_file), FilePath(backing_file))
+    """
+    output = check_output(
+        ["losetup", "--all"]
+    ).decode('utf8')
+    return losetup_list_parse(output)
+
+
+def device_for_path(expected_backing_file):
```
itamarst: And this.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+        ``None`` if no device exists.
+    """
+    for device_file, backing_file in losetup_list():
+        if expected_backing_file == backing_file:
+            return device_file
+    return None
+
+
+@implementer(IBlockDeviceAPI)
+@attributes(['root_path'])
```
itamarst: 1. Use `pyrsistent` for new code.
2. This attribute should be private.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+        try:
+            self._attached_directory.makedirs()
+        except OSError:
+            pass
+
+    def create_volume(self, size):
+        """
+        Create a file of some size and put it in the ``unattached`` directory.
+
+        See ``IBlockDeviceAPI.create_volume`` for parameter and return type
```
itamarst: :question: You don't need to say this bit, really.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+        documentation.
+        """
+        volume = BlockDeviceVolume(
+            blockdevice_id=unicode(uuid4()),
+            size=size,
+        )
+        self._unattached_directory.child(
```
itamarst: Eeeeeeeeee. Means we can't test large files without using massive amounts of RAM. I suggest doing this the slightly longer way by opening the file and using... `seek()` and then `write()`? However it is you create giant files.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,313 @@
...
+        if volume.host is None:
+            old_path = self._unattached_directory.child(blockdevice_id)
+            host_directory = self._attached_directory.child(host)
+            try:
+                host_directory.makedirs()
+            except OSError:
+                pass
+            new_path = host_directory.child(blockdevice_id)
+            old_path.moveTo(new_path)
+            check_output(["losetup", "--find", new_path.path])
```
itamarst: Might be worth explaining why you're doing this.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+        been created.
+        """
+        self.assertEqual([], self.api.list_volumes())
+
+    def test_created_is_listed(self):
+        """
+        ``create_volume`` returns a ``BlockVolume`` that is returned by
+        ``list_volumes``.
+        """
+        new_volume = self.api.create_volume(size=1024)
```
itamarst: This fails on real block storage, due to being too small. E.g. for EBS: "Constraints: If the volume type is io1, the minimum size of the volume is 4 GiB; otherwise, the minimum size is 1 GiB. The maximum volume size is 1024 GiB. If you specify a snapshot, the volume size must be equal to or larger than the snapshot size."

itamarst: (This could be fixed in follow up issue).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+    def test_attach_unknown_volume(self):
+        """
+        An attempt to attach an unknown ``BlockDeviceVolume`` raises
+        ``UnknownVolume``.
+        """
+        self.assertRaises(
+            UnknownVolume,
+            self.api.attach_volume,
+            blockdevice_id=unicode(uuid4()),
+            host=b'192.0.2.123'
```
itamarst: Fix here, or more plausibly in relevant followup issue: the host or whatever should not be hardcoded, you can't presume specific values in cloud.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+        volume1 = self.api.create_volume(size=1024)
+        volume2 = self.api.create_volume(size=1024)
+        attached_volume1 = self.api.attach_volume(
+            volume1.blockdevice_id, host=expected_host
+        )
+        attached_volume2 = self.api.attach_volume(
+            volume2.blockdevice_id, host=expected_host
+        )
+
+        self.assertEqual(
```
itamarst: I think `assertItemsEqual` might suffice and you don't need custom sort.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+        the attached volume.
+        """
+        new_volume = self.api.create_volume(size=1024 * 50)
+        attached_volume = self.api.attach_volume(
+            new_volume.blockdevice_id,
+            b'192.0.2.123'
+        )
+        device_path = self.api.get_device_path(attached_volume.blockdevice_id)
+        self.assertTrue(
+            device_path.isBlockDevice(),
```
itamarst: The idea is you're attaching block devices to some other machine. The path is therefore the path on that other machine, not locally... so `isBlockDevice()` isn't useful for real block devices.

Maybe... create a filesystem on the device? Of course if you have bugs that may destroy something :worried:

itamarst: In any case the validation has to happen on the node itself.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+            u"Not a block device. Path: {!r}".format(device_path)
+        )
+
+    def test_get_device_path_device_exists(self):
```
itamarst: Test name isn't quite right.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+    :returns: A ``TestCase`` with tests that will be performed on the
+       supplied ``IBlockDeviceAPI`` provider.
+    """
+    class Tests(IBlockDeviceAPITestsMixin, SynchronousTestCase):
+        def setUp(self):
+            self.api = blockdevice_api_factory(test_case=self)
+
+    return Tests
+
+
```
itamarst: The loopback tests probably belong elsewhere.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -0,0 +1,459 @@
...
+        )
+):
+    """
+    Interface adherence Tests for ``LoopbackBlockDeviceAPI``.
+    """
+
+
+class LoopbackBlockDeviceAPIImplementationTests(SynchronousTestCase):
+    """
+    Implementation specific tests for ``LoopbackBlockDeviceAPI``.
```
itamarst: Do we actually care, so long as interface tests pass and given it's only used for testing?

itamarst:
Looking forward to seeing the whole thing! Please address and resubmit.

flocker/node/agents/blockdevice.py
wallrj: I added a note to the Interface docstring.

flocker/node/agents/blockdevice.py
wallrj: The docstring was misleading.
It doesn't "calculate the path that will be used" it "Returns the device path that has been allocated to the attached block device on the host"

flocker/node/agents/blockdevice.py
wallrj: True, but I think we'll refactor this later.

flocker/node/agents/blockdevice.py
wallrj: It's not really a record class. I only used characteristic to avoid having to create an initialiser method.
But that's not a good reason, so I just added an __init__ method.

flocker/node/agents/blockdevice.py
wallrj: I thought it was useful here since the docstring contains some implementation details.

flocker/node/agents/blockdevice.py
wallrj: Python `file` has a `truncate` method. And I added a test too.

flocker/node/agents/test/test_blockdevice.py
wallrj: I switched to sparse files and made all the tests use realistic sizes

```
(1490)[vagrant@localhost vagrant]$ du -hs _trial_temp/
460K	_trial_temp/
(1490)[vagrant@localhost vagrant]$ du -hs --apparent-size _trial_temp/
61G	_trial_temp/

```

flocker/node/agents/test/test_blockdevice.py
wallrj: I'll leave it for a followup issue, but won't create a separate issue. We'll fix this when we implement the first real block device backend. I added a FIXME comment though.

flocker/node/agents/test/test_blockdevice.py
wallrj: Perhaps I should move this to the implementation specific test case for `LoopbackBlockDeviceAPI`

Not sure how we *can* perform this test generally. 

But, as you say, it can be tested generally in the tests for deployers that use the IBlockDeviceAPI providers, when attempts are made to format and mount the attached device.

So lets say we'll cross that bridge when we get to it. But it is a good example of why we should develop the real block APIs in lock step with the loop API, as you mentioned elsewhere.

flocker/node/agents/test/test_blockdevice.py
wallrj: We'll refactor later.

flocker/node/agents/test/test_blockdevice.py
wallrj: The tests were useful during TDD of the implementation, so they may as well stay, right?

wallrj:
Thanks @itamarst 

I addressed or answered all your comments, I think.

Ready for another review.

flocker/node/agents/blockdevice.py
wallrj: Done.

flocker/node/agents/blockdevice.py
wallrj: Done.

flocker/node/agents/test/test_blockdevice.py
itamarst: Yep.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,340 @@
...
+        :returns: A ``BlockDeviceVolume``.
+        """
+
+    def attach_volume(blockdevice_id, host):
+        """
+        Attach ``blockdevice_id`` to ``host``.
+
+        :param unicode blockdevice_id: The unique identifier for the block
+            device being attached.
+        :param bytes host: The IP address of a host to attach the volume to.
```
itamarst: Elsewhere this is unicode, I think.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,340 @@
...
+def _losetup_list_parse(output):
+    """
+    Parse the output of ``losetup --all`` which varies depending on the
+    privileges of the user.
+
+    :param unicode output: The output of ``losetup --all``.
+    :returns: A ``list`` of
+        2-tuple(FilePath(device_file), FilePath(backing_file))
+    """
+    devices = []
```
itamarst: The strings in this function should be explicitly unicode.

flocker/node/agents/blockdevice.py
```diff
@@ -0,0 +1,340 @@
...
+        block device which will be allocated a real block device file on the
+        host to which it is attached. This allows the consumer of this API to
+        perform formatting and mount operations exactly the same as for a real
+        block device.
+
+        See ``IBlockDeviceAPI.attach_volume`` for parameter and return type
+        documentation.
+        """
+        volume = self._get(blockdevice_id)
+        if volume.host is None:
```
itamarst: Next time you do this - might be easier to read if logic was "if volume.host is not None: raise" and then rest of logic at function level. Not necessary to change in this case, though.

PR #1218, opened 2015-03-17
Create container with policy floc 1545

dwgebler:
Fixes FLOC-1545

Please wait for #1214 to be merged before reviewing this, it will make the diff a lot smaller.


docs/advanced/api_examples.yml
```diff
@@ -281,6 +281,56 @@
...
+    This is a positive integer specifying the maximum number of attempts to restart a failed container.
+    If a restart policy is not specified, the container will be created with a restart policy of "never".
+
+  request: |
+    POST /v1/configuration/containers HTTP/1.1
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "restart_policy": {"name": "on-failure","maximum_retry_count": 5}}
```
itamarst: Space after comma.

docs/advanced/api_examples.yml
```diff
@@ -281,6 +281,56 @@
...
+
+  response: |
+    HTTP/1.1 201 Created
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "restart_policy": {"name": "on-failure","maximum_retry_count": 5}}
```
itamarst: Space after comma.

flocker/control/_model.py
```diff
@@ -90,9 +90,7 @@ class IRestartPolicy(Interface):
 
 
 @implementer(IRestartPolicy)
-@attributes([], apply_immutable=True,
-            # https://github.com/hynek/characteristic/pull/22
-            apply_with_init=False)
+@attributes([Attribute("name", default_value=u"never")], apply_immutable=True)
```
itamarst: I don't think this should be done this way. We want to separate model details from external API.

For the configuration (see `_config.py`) we had a mapping; probably want to something similar in `httpapi.py` (or share code with `_config.py`).

itamarst: If you ever *do* need to do something like this, class attribute would be better rather than instance attribute:

```
class MyClass:
    name = "blah"
```

flocker/control/httpapi.py
```diff
@@ -568,14 +606,24 @@ def container_configuration_response(application, node):
     """
     result = {
         "host": node, "name": application.name,
-        "image": application.image.full_name
+        "image": application.image.full_name,
+        "restart_policy": {
+            "name": application.restart_policy.name
```
itamarst: See comment above; I think we should dictionary for this, not rely on attribute of `IRestartPolicy`.

flocker/control/_model.py
itamarst: And if you ever do something like this where everything implementing an interface has a required attribute, you can and should document it:

```
from zope.interface import Interface, Attribute

class MyInterface(Interface):
    name = Attribute()
```

flocker/control/httpapi.py
```diff
@@ -492,11 +510,31 @@ def create_container_configuration(self, host, name, image, ports=()):
...
+            environment = frozenset(environment.items())
+
+        if restart_policy is None:
+            restart_policy = dict(name=u"never")
+
+        maximum_retry_count = None
+        if "maximum_retry_count" in restart_policy:
+            maximum_retry_count = restart_policy["maximum_retry_count"]
+
+        application_restart_policies = {
```
itamarst: Could be shared with similar code in `_config.py`.

flocker/control/httpapi.py
```diff
@@ -568,14 +606,24 @@ def container_configuration_response(application, node):
     """
     result = {
         "host": node, "name": application.name,
-        "image": application.image.full_name
+        "image": application.image.full_name,
+        "restart_policy": {
```
itamarst: If restart policy is *always* included you should update examples appropriately. Alternatively, you could omit restart policy if it's the default of never.

flocker/control/test/test_httpapi.py
```diff
@@ -242,6 +242,155 @@ def test_container_name_collision_different_node(self):
...
+        container_json_result = {
+            u"host": self.NODE_B, u"name": u"webserver",
+            u"image": u"nginx:latest", u"environment": environment,
+            u"restart_policy": {u"name": u"never"}
+        }
+        return self.assertResult(
+            b"POST", b"/configuration/containers",
+            container_json, CREATED, container_json_result
+        )
+
```
itamarst: The test coverage here isn't complete - you don't check policy of "always".

itamarst: Or "on failure" with no max.

itamarst:
Looks pretty good, but probably enough changes that it merits another quick review when done.

flocker/control/httpapi.py
dwgebler: I mulled over what to do in respect of responses for a while and eventually decided the best thing is to always return the full information about a container in a response, so I'll update the other response examples in the docs. In a response, I think it's ok for example to leave out a `ports` object if there are no exposed ports, but there is always a restart policy whether you explicitly gave it in the request or not, so yeah I think restart policy should always be included in the JSON response.

flocker/control/schema/types.yml
```diff
@@ -45,6 +45,29 @@ definitions:
...
+  restart_policy:
+    title: "Container restart policy"
+    type: object
+    description: "Policy for whether or not a container should be restarted after exit."
+    oneOf:
+    - additionalProperties: false
+      properties:
+        maximum_retry_count: {minimum: 1, type: integer}
+        name:
+          enum: [on-failure]
```
exarkun: please stick to the `-` list syntax and quote all strings

flocker/control/test/test_httpapi.py
```diff
@@ -305,13 +306,124 @@ def test_create_container_with_environment_response(self):
...
+        saving.addCallback(lambda _: self.assertResponseCode(
+            b"POST", b"/configuration/containers",
+            {
+                u"host": self.NODE_A, u"name": u"webserver3",
+                u"image": u"nginx", u"restart_policy": {
+                    u"name": u"on-failure"
+                }
+            }, CREATED
+        ))
+
```
itamarst: This doesn't cover `RestartNever` being the default or a possible input.

itamarst: And it's getting rather long for a single test.

flocker/control/schema/types.yml
```diff
@@ -45,6 +45,32 @@ definitions:
...
     additionalProperties: false
 
+  restart_policy:
+    title: "Container restart policy"
+    type: object
+    description: "Policy for whether or not a container should be restarted after exit."
+    oneOf:
+    - additionalProperties: false
+      properties:
+        maximum_retry_count: {minimum: 1, type: integer}
```
itamarst: Sounds like Jean-Paul wants us to use consistent style, so switch this to indentation based.

itamarst:
Two minor comments: address, merge when buildbot is happy. Thanks!

PR #1219, opened 2015-03-17
[FLOC-1542] Separate container discovery into its own IDeployer

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1542


flocker/control/_model.py
```diff
@@ -410,34 +420,41 @@ class NodeState(PRecord):
...
-        :return Node: Equivalent ``Node`` object.
-        """
-        return Node(hostname=self.hostname,
-                    manifestations={m.dataset_id: m
-                                    for m in self.manifestations},
-                    applications=self.running | self.not_running)
+    def update_deployment(self, deployment):
+        # Find the matching Node, replace its manifestations with
+        # self.manifestations
+        pass
```
exarkun: This doesn't seem to agree with how `NodeManifestations` is being handled in `update_cluster_state`.  Maybe you meant to get rid of the `isinstance` check in that method (that seems like it would be kind of nice)?

flocker/control/_clusterstate.py
```diff
@@ -9,6 +9,7 @@
 from ._model import Deployment
 
 
+# XXX This still needs to be updated for new model.
```
exarkun: What updating is that?

flocker/control/_model.py
itamarst: Notice that `ClusterStateService` wants to do *two* things:

1. convert state updates into a `Deployment`
2. get access to information that is *not* expressible via `Deployment` and `Node`. For example, paths to datasets. And this additional info differs depending on the type of update.

itamarst: Whereas the convergence loop only cares about the first of these.

flocker/control/_clusterstate.py
```diff
@@ -19,18 +20,21 @@ class ClusterStateService(Service):
...
         XXX: Multiple nodes may report being primary for a dataset. Enforce
         consistency here. See https://clusterhq.atlassian.net/browse/FLOC-1303
 
-        :param NodeState node_state: The state of the node.
+        :param IClusterStateUpdate cluster_state_update: Update to cluster state.
         """
-        self._nodes[node_state.hostname] = node_state
+        if isinstance(cluster_state_update, NodeManifestations):
+            self._node_paths[cluster_state_update.hostname] = cluster_state_update.paths
+        self._state = cluster_state_update.update_cluster_state(self._state)
```
exarkun: Looks like this method got called `update_deployment` on the interface.

flocker/node/_deploy.py
```diff
@@ -405,13 +427,22 @@ def __init__(self, hostname, volume_service, docker_client=None,
...
+            cluster.
+
+        :returns: A ``Deferred`` which fires with a ``NodeApplications``
             instance.
         """
+        # XXX delete manifestations code that is unrelated to
+        # applications, return NodeApplications instance.  Application's
+        # attached volumes/manifestations can be calculated based on
+        # passed-in cluster_state. This has a potential race condition
+        # unfortunately, but probably not an issue in practice.
```
exarkun: This comment is kind of unclear but I'm guessing that you plan to delete the code that calls `volume_service.list()` and replace it with code that looks up manifestations from the `cluster_state` object?  If so, sounds great.

Concerning the race condition, it seems like it's okay for this iteration of the implementation to have that problem but I think we should have a plan for fixing it later (it strikes me that this is the kind of misbehavior that we're often complaining about from Docker :smile:).

One simple (to think about but maybe hard to implement) solution seems like it would be to make sure this deployer (ie, the one for the container convergence agent) never gets asked to do anything until the dataset deployer has had a chance to discover and report (such that its state is part of the `cluster_state` passed here).  This means the this deployer (container convergence agent) will always be making decisions based on a real (if possibly slightly stale - but that doesn't matter) state instead of a likely erroneous state (ie, "there are no manifestations") which is not just state but simply wrong.  This, I think, eliminates the race condition.

The complications I can think of relating to this are:

  * Currently, the race condition goes both ways.  The dataset convergence agent depends on application state to make decisions correctly ("Is there an application running that is using this manifestation?") and vice versa ("Does a manifestation exist that matches this volume attached to this container?").  So you'd actually want both agents to get a chance to report their state to the other before either one started making decisions - but state discovery for applications depends on state discovery for datasets so bleugh it's all a big tangled mess.  The saving grace here is that eventually we'll use leases instead of having the datasets convergence agent peek into the application state which probably untangles the circularity.  So possibly this complication can be ignored in the long run.
  * Having the agents run separately (which we certainly want) makes it harder to actually implement this ordering.  I can imagine ways that the protocol might change to provide enough information to let the control service do some ordering stuff but that only helps once the agents are event driven (since right now they'll do discovery whenever they want, not just when the control service tells them to).  Also, doing this without making the agents heterogeneous and putting knowledge of them into the control service means extending the protocol even further to let agents themselves somehow declare their ordering requirements and there's more bleugh big tangled mess.

I think I had an idea for a supposedly less simple solution when I started typing this comment but now I seem to have lost it.  Maybe we should discuss this on a higher bandwidth channel and you can either explain some other solution to me or we can work one out together.

itamarst: The failure mode in this race condition is the container agent says "hm, these apps should be using manifestations as their volumes but they aren't... and there's no manifestations!" So what does it do? Presumably it stops them. If we're in situation where cluster state is missing info about the local node that means the agent was disconnected from the control service, perhaps for a while... a situation in which stopping the applications  (which will then soon be restarted) doesn't seem like it's end of the world.

Currently cluster state service never clears its memory of node state; in future it might expire after a while. Maybe convergence loop startup logic should be updated with "do a couple of iterations of discovery only until cluster state you get from control service includes local state". This'll make the race condition still there, but extremely unlikely... and combined with fact that it just means quick restart of containers we're probably fine?

exarkun:
Something that seems maybe worth thinking about eventually is that a storage system (ie, Cinder or VMware - not just Flocker's adapter for the system) has a problem and a manifestation is forcibly removed from a node (I don't know whether or under what real world conditions this actually happens).  If the dataset deployer discovers this and tries to report it to the control service while the container using the corresponding volume is still running, the invariant on `Node` might be invalidated and the control service will probably explode horribly.  Dealing with this right now might not be much of a priority (I don't really know how we should be prioritizing failure recovery work - I feel like we should actually be paying attention to it ... but ...).

flocker/node/_deploy.py
exarkun: > The failure mode in this race condition is the container agent says "hm, these apps should be using manifestations as their volumes but they aren't... and there's no manifestations!" So what does it do? Presumably it stops them.

Yep, got that.  There's a second consequence of this failure mode, though: the containers are reported back to the control service as the node's container state as running and as having no attached volume.  This is then possibly exposed to end-users who query the API for information about containers.  This will be quickly corrected (as soon as the agent stops the containers and re-reports its state) but it leaves a window open where the API is basically giving back incorrect results.

> If we're in situation where cluster state is missing info about the local node that means the agent was disconnected from the control service, perhaps for a while... a situation in which stopping the applications (which will then soon be restarted) doesn't seem like it's end of the world.

The other circumstance where it occurs is the one where the agent has just started up.  Considering the extra user-facing consequence of the race condition that I described above, this means that maybe users get randomly corrupt container information any time a container agent gets restarted.

itamarst: Restarts are fine as long as the control service is still running, since it will cache the info in memory (and someday expire it I guess).

flocker/control/_model.py
```diff
@@ -410,34 +420,41 @@ class NodeState(PRecord):
...
+@implementer(IClusterStateUpdate)
+class NodeManifestations(PRecord):
+    """
+    The current state of a node's manifestations.
+
+    :ivar PSet manifestations: All ``Manifestation`` instances that
+        are present on the node.
+    :ivar PMap paths: The filesystem paths of the manifestations on this
+        node. Maps ``dataset_id`` to a ``FilePath``.
+    """
```
exarkun: This probably needs a `hostname` field too.

flocker/control/_protocol.py
```diff
@@ -201,6 +201,8 @@ def disconnected(self, connection):
         """
         self.connections.remove(connection)
 
+    # XXX this isn't node_changed anymore really, it's cluster state
+    # updates; other than naming it's pretty much the same logic though
```
exarkun: This probably does require splitting `NodeStateCommand` into two different AMP commands, though (one that accepts a `NodeManifestations` and one that accepts a `NodeApplications`)?  Although that implies the two different kinds of agents are speaking different protocols... Which maybe isn't what we want?  On the other hand, they're conveying different state (with different structure) so it's sort of difficult to make them speak the same protocol.  We could paper over it in various ways - eg make `SerializableArgument` accepts multiple types to allow instead of just one (actually it does already because the way `isinstance` works, wee).  But that's not really helping the underlying issues (of heterogeneous network comms).

Maybe papering over the difference is fine for now though.  We don't really anticipate introducing any more kinds of convergence agents any time soon and this isn't a public interface of the cluster (so we can change it however we want later).

I can almost imagine a nice fix where we make the different agents represent their different kinds of state using a unified structure (they almost do already ... json is json right??  he he he).  But, yea, probably doesn't make a practical difference at this point.

exarkun:
Thanks.  I think some of the points raised here are deep enough that this should go through another round of design/review.  Reading this made me realize that there are some correctness issues, mostly with respect to the newly introduced concurrency.  It seems like it's worth making sure the design does address those issues before proceeding further.

itamarst:
I'm going to start from scratch on another branch.

PR #1220, opened 2015-03-17
[FLOC-1537] Release 0.3.3dev8

adamtheturtle:
This should not be merged until the release process is complete.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,5 @@
-Flocker Next Release (2015-02-11)
-=================================
+Flocker 0.3.3dev7 (2015-03-17)
+==============================
 
 - New getting started guide. (FLOC-1261)
 - Updated to Twisted 15.0. (FLOC-1302)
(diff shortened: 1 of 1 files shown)


NEWS
```diff
@@ -1,5 +1,5 @@
-Flocker Next Release (2015-02-11)
-=================================
+Flocker 0.3.3dev7 (2015-03-17)
+==============================
```
dwgebler: dev8

dwgebler:
The vagrant-dev-box failure `docker.errors.APIError: 500 Server Error: Internal Server Error ("You cannot remove a running container. Stop the container before attempting removal or use -f")` is the result of an ongoing known intermittent failure on the `IDockerClient` test suite, so if that specific failure happens again on any of the builders, it's still ok to proceed with this release process. Otherwise just update the number s/7/8 in the news file. Thanks.

exarkun:
Documentation looks good.  News looks good.  Pre-tag review :+1:

exarkun:
* Good:
    * CLI installs and works on Fedora 20
    * CLI installs and works on Ubuntu 14.04
    * CLI installs and works on OS X
    * Flocker itself works on the tutorial box (that I built myself from the release branch)
    * ELK tutorial works
  * Maybe not good
    * I don't see a 0.3.3dev8 tutorial Vagrant box on Atlas.  I do see 0.3.3dev6 and 0.3.3dev7.  And the release process seems to suggest the intent is that dev release boxes are uploaded.  The review process doesn't explicitly call for me to check this or use the result but I thought using the tutorial box would be an easy way to get some properly configured nodes to verify that `flocker-deploy` in this release actually functions.
    * BuildBot failed the dev box build, it seems.  Twice.  In different ways. :/

If the Vagrant box missing from Atlas is correct and just my misunderstanding, go ahead and merge.  Maybe also explain (in a new release-process issue?) why it's correct for the box to not be uploaded.  If it should have been uploaded by now, maybe explain why it wasn't and file an issue for addressing that if you think it's appropriate.  Then take a look at the failed dev box builds.  The fact that the failures were different suggests to me they're intermittent, not new problems with this release, so they shouldn't block the release.  If you can confirm that, then don't let them block this being merged.

Summary:

  * Fix Atlas thing
  * Maybe file issue for process docs regarding Atlas thing
  * Maybe file issue for dev box build failures
  * Merge

Thanks.

adamtheturtle:
Good spot, I hadn't pressed the final "Release version" button on Atlas.

adamtheturtle:
I think that this release has not introduced the dev box issue - it is occurring on `master`. I think that the issue may be covered elsewhere but I will look into it and make any necessary issues. Thanks for the thorough review.

PR #1221, opened 2015-03-17
[FLOC-1555] Container with dataset

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1555

Depends on #1218 being merged for nice diff.

There's no acceptance test because it's a pain to do lacking e.g. ability to poll for currently running containers. So will want to return to it later when it's easier. Or, ideas welcome.

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -59,6 +59,8 @@ DATASET_NOT_FOUND = make_bad_request(
     code=NOT_FOUND, description=u"Dataset not found.")
 DATASET_DELETED = make_bad_request(
     code=METHOD_NOT_ALLOWED, description=u"The dataset has been deleted.")
+DATASET_ON_DIFFERENT_NODE = make_bad_request(
+    code=CONFLICT, description=u"The dataset is on another node.")
 
 
 class ConfigurationAPIUserV1(object):
@@ -427,11 +429,12 @@ class ConfigurationAPIUserV1(object):
             del dataset[u"deleted"]
         return datasets
 
-    def _get_attached_volume(self, volume):
+    def _get_attached_volume(self, host, volume):
         """
         Create an ``AttachedVolume`` given a volume dictionary.
 
-        :param dict: Parameters for specific volume passed to creation
+        :param unicode host: The host where the volume should be.
+        :param dict volume: Parameters for specific volume passed to creation
             endpoint.
 
         :return AttachedVolume: Corresponding instance.
@@ -440,8 +443,11 @@ class ConfigurationAPIUserV1(object):
 
         instances = list(manifestations_from_deployment(
             deployment, volume[u"dataset_id"]))
+
         if not list(m for (m, _) in instances if not m.dataset.deleted):
             raise DATASET_NOT_FOUND
+        if not list(n for (_, n) in instances if n.hostname == host):
+            raise DATASET_ON_DIFFERENT_NODE
 
     @app.route("/configuration/containers", methods=['POST'])
     @user_documentation(
@@ -516,7 +522,7 @@ class ConfigurationAPIUserV1(object):
         # Find the volume, if any:
         attached_volume = None
         if volumes:
-            attached_volume = self._get_attached_volume(volumes[0])
+            attached_volume = self._get_attached_volume(host, volumes[0])
 
         # Find the node.
         node = self._find_node_by_host(host, deployment)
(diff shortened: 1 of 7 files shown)


dwgebler:
Merged master in to this branch and resolved conflicts to make the diff easier for review.

flocker/control/httpapi.py
```diff
@@ -510,19 +549,22 @@ def create_container_configuration(
...
-
         for node in deployment.nodes:
             for application in node.applications:
                 if application.name == name:
                     raise CONTAINER_NAME_COLLISION
 
+        # Find the volume, if any:
+        attached_volume = None
+        if volumes:
+            attached_volume = self._get_attached_volume(host, volumes[0])
```
dwgebler: Might be worth adding a comment just to underline again that we only support one attached volume right now, hence why we only care about `volumes[0]`

flocker/control/schema/endpoints.yml
```diff
@@ -28,6 +28,16 @@ definitions:
...
+      volumes:
+        title: Volumes to attach to the container
+        description: |
+          Volumes to attach to the container. Required keys are "dataset_id" and
+          "mountpoint". For now only one volume per container is supported.
+        type: array
+        items:
+          '$ref': 'types.json#/definitions/volume'
+        minItems: 0
+        maxItems: 1
```
dwgebler: :question: possibly easier to locate and maintain the detailed definitions if they're all in `types.yml` and `volumes` sticks to the same convention of just using a `$ref` here.

flocker/control/schema/types.yml
```diff
@@ -140,12 +140,33 @@ definitions:
...
+      - 'dataset_id'
+      - 'mountpoint'
+    additionalProperties: false
+
+  volume_path:
+    title: "Volume mount path"
+    description: |
+      The filesystem path where a volume is mounted inside a container.
+    type: string
+
```
dwgebler: Is there a pattern this could be constrained to for better validation? I'm guessing it has to be an absolute path.
`^((?:\/[a-zA-Z0-9]+(?:_[a-zA-Z0-9]+)*(?:\-[a-zA-Z0-9]+)*)+)$` :smile:

dwgebler:
Thanks. There are a few minor optional comments to address if you want to, otherwise merge.

exarkun:
> There's no acceptance test because it's a pain to do lacking e.g. ability to poll for currently running containers. So will want to return to it later when it's easier. Or, ideas welcome.

Can you file an issue for adding this and mark it as blocking FLOC-1139 so we don't lose track of the remaining work?

flocker/control/httpapi.py
```diff
@@ -435,6 +441,35 @@ def state_datasets(self):
...
+        :return AttachedVolume: Corresponding instance.
+        """
+        deployment = self.persistence_service.get()
+
+        instances = list(manifestations_from_deployment(
+            deployment, volume[u"dataset_id"]))
+
+        if not list(m for (m, _) in instances if not m.dataset.deleted):
+            raise DATASET_NOT_FOUND
+        if not list(n for (_, n) in instances if n.hostname == host):
```
exarkun: Perhaps `not any(n.hostname == host ...)` is a little more readable?

flocker/control/httpapi.py
exarkun: Particularly with a reference to the issue for lifting this limitation (do we have one  yet?)

flocker/control/test/test_schemas.py
```diff
@@ -289,6 +289,55 @@
...
+            'name': 'postgres',
+            'volumes': [{'dataset_id': 123,
+                         'mountpoint': '/var/db'}],
+        },
+        # Volume with mountpoint of wrong type
+        {
+            'host': '192.168.0.3',
+            'image': 'postgres',
+            'name': 'postgres',
+            'volumes': [{'dataset_id': "x" * 36,
```
exarkun: These aren't quite valid uuids.  It might be nicer if they were.

exarkun: Although I guess we decided not to require actual uuids... so it would be reason to say this should be fixed elsewhere first.

itamarst:
There's a blocking acceptance issue I filed yesterday... FLOC-1567.

flocker/control/schema/types.yml
itamarst: Hm... I'm going to say that starting with a `/` is a must, and everything else is not.

PR #1222, opened 2015-03-17
[FLOC-1229] Allow passing arguments to trial run by tox

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1229

Now you can do `tox -e py27 flocker.control.test.test_httpapi` or whatever. Lack of this was constant annoyance for me.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -80,6 +80,12 @@ You can run all unit tests by doing:
 
    $ tox
 
+You can also run specific tests in a specific environment:
+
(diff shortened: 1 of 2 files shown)


adamtheturtle:
Thanks @itamarst . Your example works and Buildbot is as happy as can be expected, so this can be merged. However, I think that there should be documentation for this in http://doc-dev.clusterhq.com/gettinginvolved/contributing.html?highlight=tox#running-tests. Please add that and merge if you're happy, or resubmit if you prefer.

PR #1223, opened 2015-03-18
[FLOC-1556] Create container with CPU shares

dwgebler:
Fixes FLOC-1556

Please wait for #1218 to be merged before reviewing, it will make the diff a lot smaller.

### DIFF
--- a/flocker/control/_config.py
+++ b/flocker/control/_config.py
@@ -210,7 +210,17 @@ class ApplicationMarshaller(object):
         if volume:
             config['volume'] = volume
         config['restart_policy'] = self.convert_restart_policy()
+        cpu_shares = self.convert_cpu_shares()
+        if cpu_shares:
+            config['cpu_shares'] = cpu_shares
         return config
+
+    def convert_cpu_shares(self):
+        """
+        :returns: ``None`` if there is CPU shares weighting on this
+        application, otherwise an ``int`` representing the CPU shares.
+        """
+        return self._application.cpu_shares
 
     def convert_restart_policy(self):
         """
--- a/flocker/control/_config.py
+++ b/flocker/control/_config.py
@@ -210,17 +210,7 @@ class ApplicationMarshaller(object):
         if volume:
             config['volume'] = volume
         config['restart_policy'] = self.convert_restart_policy()
-        cpu_shares = self.convert_cpu_shares()
-        if cpu_shares:
-            config['cpu_shares'] = cpu_shares
         return config
-
-    def convert_cpu_shares(self):
-        """
-        :returns: ``None`` if there is CPU shares weighting on this
-        application, otherwise an ``int`` representing the CPU shares.
-        """
-        return self._application.cpu_shares
 
     def convert_restart_policy(self):
         """
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -45,6 +45,13 @@ definitions:
     maxProperties: 16
     additionalProperties: false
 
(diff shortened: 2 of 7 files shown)


flocker/control/_config.py
```diff
@@ -210,8 +210,18 @@ def convert(self):
         if volume:
             config['volume'] = volume
         config['restart_policy'] = self.convert_restart_policy()
+        cpu_shares = self.convert_cpu_shares()
```
itamarst: Doing this change in `_config.py` means it impacts our configuration language... but there's no test coverage from that direction. Given we're likely going to kill our configuration language I am of two minds about whether we should be updating it or making this change in `httpapi.py`. Most of the code is going to be thrown out (other than Fig) and so anything used by HTTP API will end up being moved to `httpapi.py` eventually anyway.

I already have to munge result for volumes (where JSON schema doesn't quite match config format) so I'd say do this in `httpapi.py`. And eventually we'll merge into one nice function.

flocker/control/schema/types.yml
```diff
@@ -45,6 +45,39 @@ definitions:
     maxProperties: 16
     additionalProperties: false
 
+  cpu_shares:
+    title: "Container CPU shares"
+    description: "A number specifying the relative weighting of CPU cycles given to a container."
```
itamarst: Add link to relevant Docker documentation.

flocker/control/test/test_httpapi.py
```diff
@@ -305,13 +306,186 @@ def test_create_container_with_environment_response(self):
...
+        """
+        A valid API request to create a container including CPU shares
+        returns the CPU shares supplied in the request in the response
+        JSON.
+        """
+        container_json = {
+            u"host": self.NODE_B, u"name": u"webserver",
+            u"image": u"nginx:latest", u"cpu_shares": 512
+        }
+        container_json_response = {
```
itamarst: You could do:

```
container_json_response = container_json.copy()
container_json_response[u'restart_policy'] = {...}
```

Or:
```
container_json = pmap({
    u"host": self.NODE_B, u"name": u"webserver",
    u"image": u"nginx:latest", u"cpu_shares": 512
})
container_json_response = container_json.set(u'restart_policy': {...})

# Later on do dict(container_json) to convert it to real dictionary.
```

```

File: flocker/control/test/test_schemas.py [hunk 4]
```diff
@@ -156,6 +220,38 @@
...
+            'name': 'postgres',
+            'restart_policy': {
+                'name': 'on-failure', 'maximum_retry_count': 5
+            }
+        },
+        {
+            'host': '192.168.0.3',
+            'image': 'docker/postgres:latest',
+            'name': 'postgres',
+            'cpu_shares': 512
```
itamarst: Maybe we should one success case that's an ombinus with all the things.

itamarst:
Looks good, address and merge when buildbot is happy.

flocker/control/_config.py
dwgebler: Well, it impacts `ApplicationMarshaller`, which doesn't directly alter the behaviour of our configuration language, because this class deals only with converting `Application` instances in to some `dict` representation. Should we maybe just move the marshalling class in to `httpapi`?

itamarst: I assume `ApplicationMarshaller` is used *somewhere* in the flocker-deploy execution path (`flocker-reportstate`, at a gues). So I'd rather not mess with it without adding tests for that code path... but that's a code path we want to kill.

PR #1225, opened 2015-03-18
[FLOC-1311] Control service logging

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1311

Adds logging of:
* Saving configuration and loading it on startup.
* Incoming JSON requests and outgoing JSON response, as well as more details on HTTP request (method, code).
* Error logging/handling in a couple of places.


flocker/control/test/test_persistence.py
```diff
@@ -123,6 +134,24 @@ def saved_again(_):
...
 
+    @validate_logging(
+        lambda test, logger:
+        test.assertEqual(len(logger.flush_tracebacks(ZeroDivisionError)), 1))
+    def test_register_for_callback_failure(self, logger):
+        """
+        Failed callbacks don't prevent later callbacks from being called.
+        """
+        service = self.service(FilePath(self.mktemp()), logger)
+        l = []
```
exarkun: anything but `l` as a variable name

flocker/control/test/test_persistence.py
```diff
@@ -39,15 +43,18 @@ class ConfigurationPersistenceServiceTests(TestCase):
...
         """
         Start a service, schedule its stop.
 
         :param FilePath path: Where to store data.
+        :param logger: Optional eliot ``Logger`` to set before startup.
 
         :return: Started ``ConfigurationPersistenceService``.
         """
         service = ConfigurationPersistenceService(reactor, path)
+        if logger:
```
exarkun: `if logger is not None:`

flocker/control/test/test_persistence.py
```diff
@@ -123,6 +134,24 @@ def saved_again(_):
         d.addCallback(saved_again)
         return d
 
+    @validate_logging(
+        lambda test, logger:
+        test.assertEqual(len(logger.flush_tracebacks(ZeroDivisionError)), 1))
```
exarkun: if this fails because, say, two tracebacks are logged ... is the test failure meaningful (I'm guessing not... because `2 != 1`?)

flocker/control/test/test_protocol.py
```diff
@@ -627,7 +627,7 @@ def test_logging(self, logger):
         self.patch(control_amp_service, 'logger', logger)
 
         connection_protocol = ControlAMP(control_amp_service)
-        connection_protocol.makeConnection(StringTransport())
+        connection_protocol.callRemote = lambda *args, **kwargs: succeed({})
```
exarkun: Bleugh bleugh bleugh.

flocker/control/test/test_protocol.py
```diff
@@ -654,3 +654,29 @@ def test_logging(self, logger):
...
+    def test_error_sending(self, logger):
+        """
+        An error sending to one agent does not prevent others from being
+        notified.
+        """
+        control_amp_service = build_control_amp_service(self)
+        self.patch(control_amp_service, 'logger', logger)
+
+        connection_protocol = ControlAMP(control_amp_service)
+        connection_protocol.callRemote = lambda *args, **kwargs: succeed({})
```
exarkun: Bleugh

flocker/control/test/test_protocol.py
```diff
@@ -654,3 +654,29 @@ def test_logging(self, logger):
...
+
+        error = ConnectionLost()
+        disconnected_protocol = ControlAMP(control_amp_service)
+        disconnected_protocol.callRemote = lambda *args, **kwargs: fail(
```
exarkun: Bleugh

flocker/control/_protocol.py
```diff
@@ -229,16 +230,16 @@ def _send_state_to_connections(self, connections):
                                     configuration=configuration,
                                     state=state):
             for connection in connections:
-                with LOG_SEND_TO_AGENT(
-                        self.logger, agent=connection) as action:
-                    connection.callRemote(
+                action = LOG_SEND_TO_AGENT(self.logger, agent=connection)
+                with action.context():
+                    d = DeferredContext(connection.callRemote(
```
exarkun: Sort of unrelated but does `repr(connection)` actually tell anyone anything useful?  I'm guessing this ends up looking something like `<ControlAMP at 0x7fe094b57650>`... Can you file an issue for turning this field into something useful?  Probably by logging the node identifier this connection is associated with, I'd guess.

flocker/restapi/_logging.py
```diff
@@ -13,12 +13,26 @@
 
 LOG_SYSTEM = u"api"
 
+METHOD = Field.forTypes(
+    u"method", [unicode, bytes], u"The HTTP method of the request.")
```
exarkun: Can we just pick one?

flocker/restapi/_logging.py
```diff
@@ -13,12 +13,26 @@
...
-    [REQUEST_PATH],
+    LOG_SYSTEM + u":request",
+    [REQUEST_PATH, METHOD],
     [],
     u"A request was received on the public HTTP interface.")
+JSON_REQUEST = ActionType(
+    LOG_SYSTEM + u":json_request",
+    [JSON],
+    [RESPONSE_CODE, JSON],
+    u"A request containing JSON request and response.")
```
exarkun: something about "bodies" in here somewhere I think

exarkun: `JSON_REQUEST` doesn't have `REQUEST_PATH` or `METHOD` because it's expected to be a child of `REQUEST` which already has that information, right?  Perhaps `RESPONSE_CODE` should also be part of the `REQUEST` action instead of being part of `JSON_REQUEST`?

itamarst: Yeah, except implementation-wise that would unfortunately require more significant refactoring so this was expedient. In practice we only really care about response code for actual API methods.

Follow-up issue?

flocker/restapi/test/test_infrastructure.py
```diff
@@ -40,6 +42,123 @@ class ArbitraryException(Exception):
     """
 
 
+class ResultHandlingApplication(object):
```
itamarst: This was just moved from further down.

flocker/restapi/_logging.py
exarkun: > Follow-up issue?

Okay

flocker/control/test/test_protocol.py
```diff
@@ -654,3 +654,29 @@ def test_logging(self, logger):
...
+            error)
+
+        control_amp_service._send_state_to_connections(
```
exarkun: `control_amp_service.save(...)`?

exarkun: eh.  I guess I mean `control_amp_service.configuration_service.save(...)` which probably isn't actually any better.

exarkun: "Public APIs" is the point here though.

flocker/control/test/test_protocol.py
exarkun: `connected_protocol` is probably the name that parallels `disconnected_protocol` best.

flocker/restapi/test/test_infrastructure.py
```diff
@@ -90,6 +211,16 @@ def test_illegalResult(self, logger):
...
         """
+        # Workaround the message validation code in @validateLogging,
+        # otherwise the invalid message being logged throws an exception!
+        # See https://github.com/ClusterHQ/eliot/issues/150
+        def cleanup():
+            for message in logger.messages:
+                if message.get("action_type") == "api:json_request":
+                    if message["json"].get("foo"):
+                        message["json"]["foo"] = None
+        self.addCleanup(cleanup)
```
exarkun: I don't quite get it.  This test produces an illegal JSON response object (it has the un-encodeable `object` instance, specified a couple lines below).  So the server generates an internal error code and logs the encoding error.

The `@validateLogging` decorator verifies the encoding error got logged.

But ... there's still a problem?  Because now the response object also gets logged in the `REQUEST_JSON` success fields.  Or rather, it would - but it can't be encoded there, either.  So... there's a second traceback logged for the problem?  Do we want two tracebacks for the same problem logged every time?

exarkun: Or maybe the second traceback only happens **because** `validateLogging` is being used?

exarkun: (But then what happens when you log a bogus value when running this code for real?)

flocker/restapi/test/test_infrastructure.py
```diff
@@ -354,7 +392,10 @@ def test_name(self):
         application = self.Application(None, None)
         self.assertEqual("foo", application.foo.__name__)
 
-    @validateLogging(_assertRequestLogged(b"/foo/bar"))
+    @validateLogging(
+        assertJSONLogged, b"PUT", b"/foo/bar",
+        {"foo": "bar", "baz": ["quux"]}, None, 200
```
exarkun: `OK` instead of `200`

flocker/restapi/test/utils.py
```diff
@@ -39,12 +39,14 @@ def __eq__(self, other):
     u"The provided JSON doesn't match the required schema.")
 
 
-def _assertRequestLogged(path):
+def _assertRequestLogged(path, method="GET"):
```
exarkun: `b"GET"`

exarkun:
Thanks.  I kinda ran out of steam partway through (and some of the inline comments are clearly not very constructive).  Sorry.  Please address the other inline comments and re-submit (mostly for my benefit, so I can try reading this again with a clearer head).

flocker/control/test/test_persistence.py
itamarst: Well, you'll get the line number of the test module, so doesn't seem like a problem to me.

flocker/control/_protocol.py
itamarst: Done, FLOC-1585.

flocker/restapi/test/test_infrastructure.py
itamarst: Second traceback only happens because of `validateLogging`, yeah - it does extra validation. In real code the log message will be ... dropped I guess. Until https://github.com/ClusterHQ/eliot/issues/45 is fixed at which point it will actually log some useful error log message. That's on next eliot milestone.

flocker/control/test/test_persistence.py
exarkun: I guess it's not quite as bad as I first though since you're only flushing `ZeroDivisionError`.  Any other unexpected traceback written by this test won't get flushed, will fail the test, and will get reported in the test summary.  So yea okay.

flocker/control/test/test_persistence.py
```diff
@@ -76,17 +83,21 @@ def test_file_is_created(self):
         self.service(path)
         self.assertTrue(path.child(b"current_configuration.v1.json").exists())
 
-    def test_save_then_get(self):
+    @validate_logging(assertHasAction, _LOG_SAVE, True,
```
exarkun: keyword argument for that `True` value makes this easier to understand

flocker/restapi/_infrastructure.py
```diff
@@ -58,16 +74,10 @@ def _logging(original):
...
-            logger = _logger
-        else:
-            if logger is None:
-                logger = _logger
+        logger = _get_logger(self)
 
-        path = repr(request.path).decode("ascii")
-        action = REQUEST(logger, request_path=path)
+        action = REQUEST(logger, request_path=request.path,
+                         method=request.method)
```
exarkun: Did Eliot ever gain the ability to serialize a single field into multiple outputs?  `request=request` would be more convenient here.

I guess even if Eliot serializers aren't allowed to do that, there could still be an `IRequest` field serializer that spits out something like `{"path": ..., "method", ...}`.  That makes it a lot easier to improve the logging without fixing all of the (oh so very, very many...) call sites of `REQUEST`.

flocker/restapi/_infrastructure.py
```diff
@@ -58,16 +74,10 @@ def _logging(original):
...
-        try:
-            logger = self.logger
-        except AttributeError:
-            logger = _logger
-        else:
-            if logger is None:
-                logger = _logger
+        logger = _get_logger(self)
 
-        path = repr(request.path).decode("ascii")
```
exarkun: This line looked stupid, I guess.  The point was to handle non-ascii request paths though (I think).  I don't think this change is particularly for the better.  At the very least it seems worth a separate issue and some new tests.

flocker/control/test/test_protocol.py
```diff
@@ -655,3 +655,32 @@ def test_logging(self, logger):
...
+        self.patch(control_amp_service, 'logger', logger)
+
+        connected_protocol = ControlAMP(control_amp_service)
+        connected_protocol.callRemote = lambda *args, **kwargs: succeed({})
+
+        error = ConnectionLost()
+        disconnected_protocol = ControlAMP(control_amp_service)
+        results = [succeed({}), fail(error)]
+        disconnected_protocol.callRemote = (
+            lambda *args, **kwargs: results.pop(0))
```
exarkun: There's `flocker.testtools.FakeAMPClient` which is probably marginally better than this.  It would also seem reasonable to have an `AlwaysSucceedAMPClient` and an `AlwaysFailAMPClient` that implement `callRemote` in the ways implied by their names.  (As far as I can tell nothing about `control_amp_service.connected` cares what kind of `AMP` subclass is passed to it.)  Can be a follow-up ticket if you want. If so link to it from here.

exarkun:
Thanks.  One additional comment.  I noticed `_protocol.py` still has the pickle warning.  I think this is entirely obsolete at this point.  It's not really related but you're changing `_protocol.py` anyway so...

Otherwise, looks good.  Please address the inline comments and then merge.

flocker/restapi/_infrastructure.py
itamarst: No multi-field serializers so far, unfortunately. And given only one call site...

PR #1227, opened 2015-03-19
Copy git config to VM

adamtheturtle:
The commit for this PR was with a new dev box

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,8 +17,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box_version = "> 0.3.2.1714"
 
   config.vm.provision :shell, :inline => $bootstrap, :privileged => true
-  # Use the git configuration from the host in the VM
-  config.vm.provision "file", source: "~/.gitconfig", destination: ".gitconfig"
+  # Use the git configuration from the host in the VM, if it is in the expected
+  # location
+  if File.exists?(File.join(Dir.home, ".gitconfig"))
+    config.vm.provision "file", source: File.join(Dir.home, ".gitconfig"), destination: ".gitconfig"
+  end
 
   if Vagrant.has_plugin?("vagrant-cachier")
     config.cache.scope = :box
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -17,6 +17,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box_version = "> 0.3.2.1714"
 
   config.vm.provision :shell, :inline => $bootstrap, :privileged => true
+  # Use the git configuration from the host in the VM
+  config.vm.provision "file", source: "~/.gitconfig", destination: ".gitconfig"
```
itamarst: Won't this break for developers who don't have that file?

adamtheturtle: This gives:

```
There are errors in the configuration of this machine. Please fix
the following errors and try again:

File provisioner:
* File upload source file /Users/Adam/.gitconfig must exist
```

What is ideal? Do you think that it is best to:
* Have the error as it stands
* If the config doesn't exist there, ask the user where it does exist
* Ignore the error and continue without copying any config file if one does not exist at `~/.gitconfig`
?

itamarst: We should take a step back and look at the bigger issue: we're using this image for two things, developer-facing test box and as part of release process.

Insofar as this particular piece of functionality is irrelevant or detrimental to first use case, perhaps it should be done in way that is specific to second use case only. Which is to say, not in the Vagrantfile. Or perhaps we want separate Vagrantfiles for the different use case.

adamtheturtle: I'm not sure why this is irrelevant or detrimental to using this as a development box. In fact it would have saved me time in the past.

itamarst: Maybe you don't have a `.gitconfig` (3rd party just playing with Flocker), and now you're getting a confusing error.

adamtheturtle:
I have made some changes since this was submitted:

* This now uses what is apparently a more standard convention for getting the home directory.
* There is no change if ~/.gitconfig is not available (nothing is copied, no error occurs)
* That this is copied is documented in the release process

Hopefully that addresses the concerns.

itamarst:
Thanks, merge when green.

PR #1228, opened 2015-03-19
Add enchant and openssl-dev dependencies

adamtheturtle:
I'm hoping that this is done but Buildbot will pass or fail hopefully within a few hours.


exarkun:
Thanks.  In the future, it would probably be nice to split up fixes for separate issues into separate PRs.  Even though these changes are all necessary to resolve the issue, "Failing tox on dev box", that's mostly because the issue is overly broad.

BuildBot is green and the changes seem reasonable (insofar as "reasonable" is a thing that happens in software).  Merging.  Thanks again.

PR #1229, opened 2015-03-19
[FLOC-1558] Create container with links

dwgebler:
Fixes FLOC-1558

### DIFF
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -282,6 +282,30 @@
 
 -
   id:
+    "create container with links"
+
+  doc: |
+    Create a new container with links to one or more other containers.
+
+  request: |
+    POST /v1/configuration/containers HTTP/1.1
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "links": [{"alias":"postgres", "local_port": 5432, "remote_port": 54320}]}
+
+  response: |
+    HTTP/1.1 201 Created
+
+    {"host": "%(NODE_0)s",
+     "name": "webserver",
+     "image": "nginx:latest",
+     "restart_policy": {"name": "never"},
+     "links": [{"alias":"postgres", "local_port": 5432, "remote_port": 54320}]}     
+
+-
+  id:
     "create container with memory limit"
 
   doc: |
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -286,6 +286,7 @@
 
   doc: |
     Create a new container with links to one or more other containers.
+    See https://docs.clusterhq.com/en/latest/advanced/configuration.html for more information on Flocker's implementation of container links.
 
   request: |
     POST /v1/configuration/containers HTTP/1.1
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -286,7 +286,9 @@
 
   doc: |
     Create a new container with links to one or more other containers.
-    See https://docs.clusterhq.com/en/latest/advanced/configuration.html for more information on Flocker's implementation of container links.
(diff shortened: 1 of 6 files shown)


docs/advanced/api_examples.yml
```diff
@@ -282,6 +282,31 @@
 
 -
   id:
+    "create container with links"
+
+  doc: |
+    Create a new container with links to one or more other containers.
+    See https://docs.clusterhq.com/en/latest/advanced/configuration.html for more information on Flocker's implementation of container links.
```
itamarst: That's a bit of a vague link (there's lot of stuff in configuration) and we're going to be de-emphasizing the configuration language going forward. I recommend just writing some stuff here that explains what's going on in the context of this particular example. "Presuming we have another container exposed on port 54320, this will setup some env varaibles (link to docker docs) like so" and so on.

flocker/control/httpapi.py
```diff
@@ -52,6 +52,13 @@
 CONTAINER_PORT_COLLISION = make_bad_request(
     code=CONFLICT, description=u"A specified external port is already in use."
 )
+LINK_PORT_COLLISION = make_bad_request(
+    code=CONFLICT,
+    description=u"Local port links in a container must be unique."
```
itamarst: That's a little confusing... maybe "The local ports in a container's links must be unique."

flocker/control/httpapi.py
```diff
@@ -455,7 +463,8 @@ def state_datasets(self):
     )
     def create_container_configuration(
         self, host, name, image, ports=(), environment=None,
-        restart_policy=None, cpu_shares=None, memory_limit=None
+        restart_policy=None, cpu_shares=None, memory_limit=None,
+        links=None
```
itamarst: If you default to empty tuple you can get rid of "if links is not None" check.

flocker/control/httpapi.py
```diff
@@ -544,7 +576,8 @@ def create_container_configuration(
             environment=environment,
             restart_policy=policy,
             cpu_shares=cpu_shares,
-            memory_limit=memory_limit
+            memory_limit=memory_limit,
+            links=frozenset(application_links)
```
itamarst: Don't need the `frozenset`; pyrsistent is just going to convert whatever you pass in to a `PSet`.

flocker/control/schema/types.yml
```diff
@@ -60,6 +60,32 @@ definitions:
     minimum: 1
     maximum: 1024
 
+  container_links:
+    title: "Container links"
+    description: |
+        A mapping of links between containers, specifying the local and remote ports.
+        See https://docs.clusterhq.com/en/latest/advanced/configuration.html for more information.
```
itamarst: As mentioned above, I think we should include actual docs here. Insofar as there is a link it should be to Docker link environment variable docs.

flocker/control/schema/types.yml
```diff
@@ -60,6 +60,32 @@ definitions:
...
+      type: object
+      required:
+        - alias
+        - local_port
+        - remote_port
+      additionalProperties: false
+      properties:
+        local_port:
+          type: integer
+          minimum: 1
```
itamarst: max 65535

flocker/control/test/test_httpapi.py
```diff
@@ -530,6 +530,124 @@ def test_create_container_with_cpu_shares_response(self):
...
+    def test_create_container_with_links_response(self):
+        """
+        An API request to create a container including links to be injected in
+        to the container returns the link information in the response JSON.
+        """
+        container_json = pmap({
+            u"host": self.NODE_B, u"name": u"webserver",
+            u"image": u"nginx:latest", u"links": [
+                {
+                    'alias': 'postgres',
```
itamarst: Unicode strings here and in next test.

itamarst:
Looks good, address and merge.

PR #1231, opened 2015-03-19
Ignore lintian error about package revision.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1566


adamtheturtle:
Thanks @tomprince this seems sensible and also like something we will find out soon enough that we need to change if we do - if we submit the packages to some repository for example.

PR #1232, opened 2015-03-19
[FLOC-1569] Application can express running state

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1569

### DIFF
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -188,6 +188,8 @@ class Application(PRecord):
 
     :ivar IRestartPolicy restart_policy: The restart policy for this
         application.
+
+    :ivar bool running: Whether or not the application is running.
     """
     name = field(mandatory=True)
     image = field(mandatory=True, type=DockerImage)
@@ -199,6 +201,7 @@ class Application(PRecord):
     restart_policy = field(mandatory=True, initial=RestartNever())
     environment = field(mandatory=True, initial=pmap(), factory=pmap,
                         type=PMap)
+    running = field(mandatory=True, initial=True, type=bool)
 
 
 class Dataset(PRecord):
@@ -404,8 +407,7 @@ class NodeState(PRecord):
     configuration models.
 
     :ivar unicode hostname: The hostname of the node.
-    :ivar running: A ``PSet`` of ``Application`` instances on this node
-        that are currently running or starting up.
+    :ivar applications: A ``PSet`` of ``Application`` instances on this node.
     :ivar not_running: A ``PSet`` of ``Application`` instances on this
         node that are currently shutting down or stopped.
     :ivar used_ports: A ``PSet`` of ``int``\ s giving the TCP port numbers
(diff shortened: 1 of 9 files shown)


flocker/control/_model.py
```diff
@@ -199,6 +201,7 @@ class Application(PRecord):
     restart_policy = field(mandatory=True, initial=RestartNever())
     environment = field(mandatory=True, initial=pmap(), factory=pmap,
                         type=PMap)
+    running = field(mandatory=True, initial=True, type=bool)
```
dwgebler: :question: Deployer discovery will determine the actual running state of an application, but should we maybe assume that an `Application` is not running until a host confirms otherwise?

exarkun: Docker containers have a  bunch of states they can be in apart from just "running" and "not running".  Maybe addressing that is out of scope for this issue - it's not like we were representing any of those other states prior to this change, after all.

If I were building an application on top of our container API I'd probably want to see lots more information about what `docker inspect` calls `State`.  For example, the way restart policies are reflected is running=false, restarting=true.  I'd probably want to do something differently in that case and just having running=false doesn't tell me enough.

Yea, the more I write here the more this seems out of scope.

What is the use-case for having a default here?  I'm guessing "convenience in maintaining the test suite"?

exarkun:
Wee... so many test changes.  Thanks.  This looks pretty good.  I'm curious about the answer to the question about the use-case for the `running` default.  If I guessed the right answer then I guess it's too bad but it's probably the expedient choice...

flocker/control/_model.py
itamarst: Our docker client currently exposes those two states, so we'd have to fix that first.

Default is for convenience in maintaining test suite, yes... but also a bunch of code elsewhere e.g. the create a container endpoint.

itamarst: Also note that in discovery the `running` value is not true by default, but rather is set based on what Docker says state of container is. The default is purely an internal Python API thing, not a semantic default exposed to the HTTP API.

exarkun: > but also a bunch of code elsewhere e.g. the create a container endpoint.

Yea... that's the kind of thing that worries me.  Whether you get a running container or a not running container by using the API doesn't seem like it should depend on the default value we put into a model type.

flocker/control/_model.py
```diff
@@ -404,8 +407,7 @@ class NodeState(PRecord):
     configuration models.
 
     :ivar unicode hostname: The hostname of the node.
-    :ivar running: A ``PSet`` of ``Application`` instances on this node
-        that are currently running or starting up.
+    :ivar applications: A ``PSet`` of ``Application`` instances on this node.
     :ivar not_running: A ``PSet`` of ``Application`` instances on this
         node that are currently shutting down or stopped.
```
wallrj: I guess this docstring should have been removed too.

PR #1233, opened 2015-03-19
[FLOC-1570] Allow expressing ignorance about manifestations and applications on a Node

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1570


flocker/control/_model.py
```diff
@@ -301,9 +347,8 @@ def __invariant__(self):
         return (True, "")
 
     hostname = field(type=unicode, factory=unicode, mandatory=True)
-    applications = pset_field(Application)
-    manifestations = field(type=PMap, initial=pmap(), factory=pmap,
-                           mandatory=True)
+    applications = pset_field(Application, optional=True)
+    manifestations = pmap_field(unicode, Manifestation, optional=True)
```
exarkun: I don't see how the `__invariant__` can succeed if someone actually tries to instantiate a `Node` without one of these fields.

flocker/control/_model.py
```diff
@@ -17,27 +17,71 @@
...
+    :param value: The required type for the values of the map.
+    :param bool optional: If true, ``None`` can be used as a value for
+        this field.
+
+    :return: A ``field`` containing a ``CheckedPMap``.
+    """
+    class TheMap(CheckedPMap):
+        __key_type__ = key_type
+        __value_type__ = value_type
+    TheMap.__name__ = key_type.__name__ + value_type.__name__ + "PMap"
```
exarkun: I'm tempted to say something about underscores.  What's worse, `unicodeFilePathPMap` or `unicode_FilePath_PMap`?  Hard to say.  :question:

flocker/control/_model.py
exarkun: Probably means a few more unit tests would be appropriate to cover what we want to happen here, and then corresponding `__invariant__` fixes if they prove necessary.

exarkun: Do we actually want these changes to be applied to `Node` or do they actually belong on `NodeState`?  The goal is to let the dataset convergence agent send back just manifestations and the container convergence agent send back just applications, right?  `P2PNodeDeployer.discover_local_state` returns a `NodeState`, not a `Node`.

exarkun:
Thanks.  One of us is confused about `Node` vs `NodeState` here, I think.  Please explain or fix and re-submit.

flocker/control/_model.py
itamarst: We will definitely want them on `NodeState`, but I was going to do that in followup branch.

It needs to be in `Node` because that's how current cluster state is communicated to agents, and the goal is to allow the container agent to say "oh, manifestation state is missing, I should wait until that's available".

exarkun: Ah, indeed.  Thanks, that makes sense.

itamarst: Oh, and `optional=True` means "None is allowed". It doesn't mean you can omit the attribute (that's pyrsistent terminology, I reused it for consistency). The attribute still needs to be set.

itamarst: So there needs to be test and invariant needs to be fixed, but it needs to handle `None`, not absence of attribute.

exarkun: > We will definitely want them on NodeState, but I was going to do that in followup branch.

Which issue will that be?  FLOC-1570 just says "model objects" which seems to suggest this branch should complete the work.  I don't mind if there's another one to do the rest but maybe update the jira issue to describe the actual intended scope of this work.

itamarst: As part of reworked proposal for FLOC-1542.

exarkun:
Okay.  The `__invariant__` makes sense if we apply it to configuration only.

Do you want to change ... `ClusterStateService.as_deployment` to not use `to_node` as part of this branch?  Is that all that's necessary to switch to using `NodeState` instead of `Node` for representing cluster state in the control -> agent communication?  (I'm sure it couldn't be that easy...)

itamarst:
The invariant applies to state as well... but I still think there's no way state could or should express that an application has a manifestation if it doesn't have knowledge of manifestations on the node.

`as_node()` is insufficient insofar as `NodeState.manifestations` is not a mapping. But that's probably easy to fix so I'll try to do that.

exarkun:
> The invariant applies to state as well... but I still think there's no way state could or should express that an application has a manifestation if it doesn't have knowledge of manifestations on the node.

Ideally, yes.  And I agree this probably isn't something we **should** express (or want to).  I don't think I agree it's something we **couldn't** express - in particular, unintentionally.  We're trying to change `NodeState` so that it is constructed by two separate pieces of a distributed system.  I think if we want the invariant that `Node.__invariant__` implements to apply to state as well then we probably have more work to do.  Otherwise when some part of the system violates the invariant, the result is that pyrsistent refuses to let us represent the information and the system goes down some untested, unconsidered code path.

itamarst:
If we express cluster state using a different class (`NodeState`) then it needn't have an invariant. So that solves that problem too?

exarkun:
> If we express cluster state using a different class (NodeState) then it needn't have an invariant. So that solves that problem too?

Yes, I think that makes sense (maybe I misinterpreted your previous comment, I thought "The invariant applies to state as well" meant you wanted a similar `NodeState.__invariant__`).

itamarst:
OK, will do.

flocker/control/_model.py
```diff
@@ -284,13 +329,14 @@ class Node(PRecord):
...
-    :ivar frozenset applications: A ``frozenset`` of ``Application`` instances
-        describing the applications which are to run on this ``Node``.
+    :ivar applications: A ``PSet`` of ``Application`` instances describing
+        the applications which are to run on this ``Node``.
 
     :ivar PMap manifestations: Mapping between dataset IDs and
         corresponding ``Manifestation`` instances that are present on the
         node. Includes both those attached as volumes to any applications,
-        and those that are unattached.
+        and those that are unattached.w ``None`` if this information is
```
wallrj: Typo "unattached.w"

flocker/control/_model.py
```diff
@@ -17,27 +17,72 @@
...
 from pyrsistent import (
     pmap, PRecord, field, PMap, CheckedPSet, CheckedPMap,
+    optional as optional_type
     )
 
 from zope.interface import Interface, implementer
 
 
-def pset_field(klass):
+def pset_field(item_type, optional=False):
```
exarkun: `optional` is too close to "the opposite of `mandatory`" for this API to be very clear.  How about `allow_none` or something similar?

flocker/control/_model.py
```diff
@@ -17,27 +17,72 @@
...
+        def factory(argument):
+            if argument is None:
+                return None
+            else:
+                return TheSet(argument)
+    else:
+        factory = TheSet
+    return field(type=optional_type(TheSet) if optional else TheSet,
+                 factory=factory, mandatory=True,
                  initial=TheSet())
```
exarkun: Do `mandatory=True` and `initial=...` make sense when combined?

flocker/node/test/test_loop.py
```diff
@@ -37,6 +39,18 @@ def build_protocol():
...
+def to_node(node_state):
+    """
+    Convert a ``NodeState`` to a corresponding ``Node``.
+
+    :param NodeState node_state: Object to convert.
+    :return Node: Equivalent node.
+    """
+    return Node(hostname=node_state.hostname,
+                applications=node_state.applications,
+                manifestations=node_state.manifestations)
```
wallrj: What if NodeState.applications is None? Isn't Node going to blow up?

wallrj: Same goes for manifestations.

wallrj:
Thanks @itamarst 

Please:
 * merge forward
 * resolve trivial test failures in test_blockdevice
 * answer or address comments above.

Then merge when green.

flocker/control/_model.py
itamarst: `optional` is what pyrsistent calls it. The pyrsistent author has already stated he's willing to accept this upstream, so I decided to stick to what would make sense there.

flocker/control/_model.py
itamarst: Yeah, the combination means you can't *remove* the attribute.

flocker/node/test/test_loop.py
itamarst: This is not a public API, this is a utility to make testing easier. Or rather, make tests that used to rely on `NodeState.to_node` continue to work. "Don't do that" basically :grinning:

exarkun:
Test failures are spurious / unrelated.  Merging (so I can review #1247).

PR #1237, opened 2015-03-20
[FLOC-1502] Implement an IDeployer that works with IBlockDeviceAPI

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1502

The existing dataset creation acceptance test should exercise this functionality.  It's possible to do this now if you set up some acceptance testing nodes manually.  I filed a follow-up issue for making `admin/run-acceptance-tests` help out with this.
 * https://clusterhq.atlassian.net/browse/FLOC-1574


flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
 
+DATASET = Field(
+    u"dataset",
+    lambda dataset: dataset.dataset_id,
+    u"The unique identifier of a dataset."
+)
+
+DATASET_ID = Field(
+    u"dataset_id",
+    lambda dataset_id: unicode(dataset_id),
```
itamarst: It's already unicode... `Field.forTypes()` maybe?

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+
+BLOCK_DEVICE_ID = Field(
+    u"block_device_id",
+    lambda id: unicode(id),
+    u"The unique identifier if the underlying block device."
+)
+
+BLOCK_DEVICE_SIZE = Field(
+    u"block_device_size",
+    lambda size: size,
```
itamarst: :question: there's a utility function in eliot for this I think. doesn't necessarily save much typing though :grin:

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+class CreateBlockDeviceDataset(PRecord):
+    """
+    An operation to create a new dataset on a newly created volume with a newly
+    initialized filesystem.
+
+    :ivar Dataset dataset: The dataset for which to create a block device.
+    :ivar FilePath mountpoint: The path at which to mount the created device.
+    :ivar Logger logger: An Eliot ``Logger``.
+    """
+    dataset = field()
```
itamarst: Mandatory, type.

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+    mountpoint = field(mandatory=True)
```
itamarst: Type.

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+
+    logger = Logger()
+
+    def run(self, deployer):
+        """
+        Create a block device, attach it to the host of the supplied
```
itamarst: This is a little misleading, since the code below only works if you're running this on the host where it will be attached. So maybe say "local host".

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+            )
+
+            # This will be factored into a separate IStateChange to support the
+            # case where the volume exists but is not attached.  That object
+            # will be used by this one to perform this work.  FLOC-1575
+            volume = api.attach_volume(
+                volume.blockdevice_id, deployer.hostname
+            )
+            device = api.get_device_path(volume.blockdevice_id)
+            self.mountpoint.makedirs()
```
itamarst: If it already exists this will throw exception. Follow up issue, or add requirement to existing one.

flocker/node/agents/blockdevice.py
```diff
@@ -120,10 +247,48 @@ class BlockDeviceVolume(PRecord):
     :ivar int size: The size, in bytes, of the block device.
     :ivar unicode host: The IP address of the host to which the block device is
         attached or ``None`` if it is currently unattached.
+    :ivar UUID dataset_id: The Flocker dataset ID associated with this volume.
     """
     blockdevice_id = field(type=unicode, mandatory=True)
     size = field(type=int, mandatory=True)
     host = field(type=(unicode, type(None)), initial=None)
+    dataset_id = field(type=UUID, mandatory=True)
+
```
itamarst: If these are specific to loopback volume then make them utility functions, not class methods for a class that will be used by all backends.

itamarst: Or methods on the loopback backend class.

flocker/node/agents/blockdevice.py
```diff
@@ -340,3 +504,113 @@ def get_device_path(self, blockdevice_id):
         )
         # May be None if the file hasn't been used for a loop device.
         return _device_for_path(volume_path)
+
+
+def _manifestation_from_volume(volume):
```
itamarst: This on the other hand seems like it might be useful on the `BlockDeviceVolume` class?

flocker/node/agents/blockdevice.py
```diff
@@ -340,3 +504,113 @@ def get_device_path(self, blockdevice_id):
...
+
+    :param unicode hostname: The IP address of the node that has this deployer.
+    :param IBlockDeviceAPI block_device_api: The block device API that will be
+        called upon to perform block device operations.
+    :ivar FilePath mountroot: The directory where block devices will be
+        mounted.
+    """
+    hostname = field(type=unicode, mandatory=True)
+    block_device_api = field(mandatory=True)
+    mountroot = field(type=FilePath, initial=FilePath(b"/flocker"))
```
itamarst: Mandatory.

flocker/node/agents/blockdevice.py
```diff
@@ -340,3 +504,113 @@ def get_device_path(self, blockdevice_id):
...
+
+        paths = {}
+        for manifestation in manifestations:
+            dataset_id = manifestation.dataset.dataset_id
+            mountpath = self._mountpath_for_manifestation(manifestation)
+            paths[dataset_id] = mountpath
+
+        state = NodeState(
+            hostname=self.hostname,
+            running=(),
```
itamarst: Woo, merge issues waiting to happen. I guess whoever merges second gets to fix this.

flocker/node/agents/blockdevice.py
exarkun: We made it a `UUID` instance on the new backend type (and then maybe re-used this field for both kinds of things ...)

flocker/node/agents/blockdevice.py
```diff
@@ -340,3 +504,113 @@ def get_device_path(self, blockdevice_id):
...
+                                          current_cluster_state):
+        potential_configs = list(
+            node for node in desired_configuration.nodes
+            if node.hostname == self.hostname
+        )
+        if len(potential_configs) == 0:
+            this_node_config = Node(hostname=None)
+        else:
+            [this_node_config] = potential_configs
+
```
itamarst: Maybe:
```
If len(potential_configs) == 0:
    manifestations = {}
else:
    manifestations = potential_configs[0].manifestations
```

Seems nicer than creating empty `Node` object.

flocker/node/agents/blockdevice.py
```diff
@@ -340,3 +504,113 @@ def get_device_path(self, blockdevice_id):
...
+            manifestation.dataset.dataset_id for manifestation in
+            local_state.manifestations
+        )
+
+        manifestations_to_create = set(
+            this_node_config.manifestations[dataset_id] for dataset_id in
+            configured_dataset_ids.difference(local_dataset_ids)
+        )
+
+        # TODO check for non-None size on dataset; cannot create block devices
```
itamarst: I think this should be enforced on configuration side (what are you going to do here?), see FLOC-1579.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -10,18 +10,401 @@
...
 
+if not platform.isLinux():
+    # The majority of Flocker isn't supported except on Linux - this test
+    # module just happens to run some code that obviously breaks on some other
+    # platforms.  Rather than skipping each test module individually it would
+    # be nice to have some single global solution.  FLOC-1560, FLOC-1205
+    skip = "flocker.node.agents.blockdevice is only supported on Linux"
+
+
+class BlockDeviceDeployerTests(SynchronousTestCase):
```
itamarst: There's a minimal IDeployer testcase factory that does this check already, you should use it.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -10,18 +10,401 @@
...
+
+        :raise: A test failure exception if the manifestations are not what is
+            expected.
+        """
+        discovering = deployer.discover_local_state()
+        state = self.successResultOf(discovering)
+        expected_paths = {}
+        for manifestation in expected_manifestations:
+            dataset_id = manifestation.dataset.dataset_id
+            mountpath = deployer._mountpath_for_manifestation(manifestation)
```
itamarst: This class claims to be for `BlockDeviceDeployer` generically, but here it's using a private method that is loopback specific (`_mountpath_for_manifestation`).

I don't think asserting *specific* values for paths is important (it's not part of public API), you should merely assert that the paths are present.

itamarst: Actually... never mind.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -10,18 +10,401 @@
...
+        the deployer's node, no state changes are calculated.
+        """
+        dataset_id = unicode(uuid4())
+        manifestation = Manifestation(
+            dataset=Dataset(dataset_id=dataset_id), primary=True
+        )
+        node = u"192.0.2.1"
+        other_node = u"192.0.2.2"
+        configuration = Deployment(
+            nodes=frozenset({
```
itamarst: `frozenset` unnecessary.

flocker/node/agents/blockdevice.py
```diff
@@ -53,6 +59,125 @@ class UnattachedVolume(VolumeException):
...
+    [DATASET, MOUNTPOINT],
+    [DEVICE_PATH, BLOCK_DEVICE_ID, DATASET_ID, BLOCK_DEVICE_SIZE,
+     BLOCK_DEVICE_HOST],
+    u"A block-device-backed dataset is being created.",
+)
+
+
+@implementer(IStateChange)
+class CreateBlockDeviceDataset(PRecord):
+    """
```
itamarst: When you break out the components into sub-actions, I recommend doing the thing we discussed where `CreateBlockDeviceDataset` continues to exist... but is a compound `IStateChange` that runs sub-`IStateChange` instances.

That means high-level tests can still just assert "if dataset isn't created then do this high-level action", and you can have separate test that is "this high-level action is composed of these sub-actions".

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -10,18 +10,401 @@
...
+        ``BlockDeviceDeployer.calculate_necessary_state_changes`` bases its
+        decision about whether it is necessary to create a dataset by
+        inspecting the ``local_state`` argument.  It does this, instead of
+        inspecting the ``current_cluster_state`` argument, because the
+        ``local_state`` is presumed to be the most up-to-date information.
+        ``current_cluster_state`` has traveled a round-trip between the node
+        and the control service and had that period of time to fall behind the
+        actual node state.
+        """
+        local_hostname = u"192.0.2.1"
```
itamarst: Having refactored the code in convergence loop to apply local state to cluster state, this test seems unnecessary?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -10,18 +10,401 @@
...
+
+        # If Deployer is buggy and not overriding cluster state with local
+        # state this would result in a dataset creation action:
+        expected_changes = InParallel(changes=[])
+
+        self.assertEqual(expected_changes, actual_changes)
+
+    def test_match_configuration_to_state_of_datasets(self):
+        """
+        ``BlockDeviceDeployer.calculate_necessary_state_changes`` determines
```
itamarst: I don't think this is a good generic test; the description should probably be changed to specifically talk about metadata. A different size, in particular, would result in an action, so this is purely about metadata.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -476,3 +925,111 @@ def test_remove_inode(self):
...
+            )
+        )
+
+    def test_run(self):
+        """
+        ``CreateBlockDeviceDataset.run`` uses the ``IDeployer``\ 's API object
+        to create a new volume, attach it to the deployer's node, initialize
+        the resulting block device with a filesystem, and mount the filesystem.
+        """
+        host = u"192.0.2.1"
```
itamarst: 127.0.0.1 seems more accurate.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -476,3 +925,111 @@ def test_remove_inode(self):
...
+        mountpoint = mountroot.child(unicode(dataset_id).encode("ascii"))
+        dataset = Dataset(
+            dataset_id=unicode(dataset_id),
+            maximum_size=REALISTIC_BLOCKDEVICE_SIZE
+        )
+        change = CreateBlockDeviceDataset(
+            dataset=dataset, mountpoint=mountpoint
+        )
+        change.run(deployer)
+
```
itamarst: Seems like these should be two separate tests?

flocker/node/script.py
```diff
@@ -1,21 +1,22 @@
 # Copyright Hybrid Logic Ltd.  See LICENSE file for details.
-# -*- test-case-name: flocker.node.test.test_script -*-
+# -*- test-case-name: flocker.node.test.test_script,flocker.node.functional.test_script -*- # noqa
 
 """
-The command-line ``flocker-changestate`` and ``flocker-reportstate``
-tools.
+The command-line ``flocker-changestate`` and ``flocker-reportstate`` tools.
```
itamarst: Maybe add agents here too.

flocker/node/test/test_script.py
```diff
@@ -462,44 +467,213 @@ def test_starts_convergence_loop(self):
...
-    def test_default_port(self):
+    def test_get_service(self):
         """
-        The default AMP destination port configured by ``ZFSAgentOptions`` is
-        4524.
+        ``DatasetAgentServiceFactory.get_service`` creates ``AgentLoopService``
+        configured with the destination given by the options.
         """
-        options = ZFSAgentOptions()
-        options.parseOptions([b"1.2.3.4", b"example.com"])
```
itamarst: I half-feel most of these should be functional tests, not unit tests. But they're already written, so that's fine.

run-iblockdevice-acceptance-tests
```diff
@@ -0,0 +1,90 @@
+#!/bin/sh
```
itamarst: Not going to spend much time reviewing this since it's going to be thrown out soon, I hope.

run-iblockdevice-acceptance-tests
```diff
@@ -0,0 +1,90 @@
...
+set -e
+
+FLOCKER_ACCEPTANCE_NODE="172.16.255.250"
+FLOCKER_ACCEPTANCE_CONTROL_NODE="${FLOCKER_ACCEPTANCE_NODE}"
+FLOCKER_ACCEPTANCE_AGENT_NODE="${FLOCKER_ACCEPTANCE_NODE}"
+REPO="${RANDOM}"
+DISTRIBUTION="fedora-20"
+BRANCH="iblockdevice-deployer-redux-FLOC-1502"
```
itamarst: Except that hard coding branch seems like a problem?

itamarst:
Looks good, address and merge.

flocker/node/agents/blockdevice.py
wallrj: I linked to this comment from https://clusterhq.atlassian.net/browse/FLOC-1575 and https://clusterhq.atlassian.net/browse/FLOC-1576

flocker/node/agents/blockdevice.py
wallrj: Added a further comment and a note to https://clusterhq.atlassian.net/browse/FLOC-1498

flocker/node/agents/blockdevice.py
wallrj: I'll leave it for now, since it only creates primary manifestations.

flocker/node/agents/blockdevice.py
wallrj: Not sure about this. We want the default mountroot to be /flocker but to be overridable for tests.

flocker/node/agents/test/test_blockdevice.py
wallrj: I think @exarkun and I discussed this and thought it was useful to have the implementation specific test...but I can't remember our thinking.

flocker/node/agents/test/test_blockdevice.py
wallrj: I don't agree. It's never going to be 127.0.0.1 in real life.

flocker/node/agents/blockdevice.py
wallrj: Arghh! You got there first I guess :frowning:

wallrj:
The dataset create API acceptance tests pass:

```
(1502)(iblockdevice-deployer-redux-FLOC-1502 ? =)[~/.../HybridLogic/flocker]$ ./run-iblockdevice-acceptance-tests 
run-api-tests.sh                                                                                                    100% 1841     1.8KB/s   00:00    
rm '/etc/systemd/system/multi-user.target.wants/flocker-control.service'
rm '/etc/systemd/system/sockets.target.wants/flocker-control-agent.socket'
rm '/etc/systemd/system/sockets.target.wants/flocker-control-api.socket'
ln -s '/usr/lib/systemd/system/flocker-control.service' '/etc/systemd/system/multi-user.target.wants/flocker-control.service'
ln -s '/usr/lib/systemd/system/flocker-control-agent.socket' '/etc/systemd/system/sockets.target.wants/flocker-control-agent.socket'
ln -s '/usr/lib/systemd/system/flocker-control-api.socket' '/etc/systemd/system/sockets.target.wants/flocker-control-api.socket'
flocker.acceptance.test_api
  DatasetAPITests
    test_dataset_creation ... mke2fs 1.42.8 (20-Jun-2013)
                                             [OK]

Ran 1 tests in 3.251s

PASSED (successes=1)
REMOTE_AGENT_PID: 24949
Killed by signal 15.

```

wallrj:
So I think I'll remove that script and merge once the non-acceptance test builders have passed.

PR #1238, opened 2015-03-23
[FLOC-1174] Docker discover environment

dwgebler:
Fixes FLOC-1174

This branch adds environment information to `DockerClient.list`, disregarding the `HOME` and `PATH` variables that are added to all containers, thereby returning only custom environment variables that have been injected as part of an image, links or when launching the container (i.e. an `Application` environment).

Also adds functionality to discover custom environment variables in `P2PNodeDeployer`, distinguishing between variables injected to represent links and custom environment.


flocker/acceptance/test_postgres.py
```diff
@@ -40,6 +40,12 @@
         Port(internal_port=POSTGRES_INTERNAL_PORT,
              external_port=POSTGRES_EXTERNAL_PORT),
         ]),
+    environment=frozenset({
+        'LANG': 'en_US.utf8',
+        'PGDATA': '/var/lib/postgresql/data',
+        'PG_VERSION': '9.4.1-1.pgdg70+1',
```
itamarst: 1. You can and should just pass a dictionary these days.
2. Why was this change made? If the expectation s that `PG_VERSION` will be set and therefore this is required for comparisons to work, that seems super-fragile. This may also indicate a weakness in our comparison logic that needs to be fixed in this branch - if we're saying "oh this container is no good because it has extra environment variables that weren't configured" then that's a problem.

flocker/node/_docker.py
```diff
@@ -554,6 +554,20 @@ def _list():
...
                 else:
                     continue
+                # Retrieve environment variables for this container,
+                # disregarding the HOME and PATH environment variables that are
+                # included in every container and do not translate back to the
+                # deployed Flocker applications.
+                unit_environment = []
+                environment_data = data[u"Config"][u"Env"]
+                for environment in environment_data:
+                    if not environment.startswith(("HOME", "PATH")):
```
itamarst: `.startswith()` seems like a bad idea. Why not equality?

itamarst: Or actually , you probably want `("HOME=", "PATH=")` so you don't skip things like `"HOME_BLAH"` by mistake.

flocker/node/test/test_deploy.py
```diff
@@ -787,6 +787,42 @@ def test_discover_multiple(self):
...
+                     environment=environment,
+                     activation_state=u'active')
+        units = {unit1.name: unit1}
+
+        fake_docker = FakeDockerClient(units=units)
+        applications = [
+            Application(
+                name=unit1.name,
+                image=DockerImage.from_string(unit1.container_image),
+                environment=frozenset(environment_variables)
```
itamarst: Environment is a dictionary now (or rather, PMap).

flocker/acceptance/test_postgres.py
itamarst: Yeah, I guess that's the issue. So we need to change calculation logic so you don't need to hardcode env variables from the docker image in your config.

I guess... if there's *extra* env variables beyond ones manually configured, that should cause no side effects. And then you can just include e.g. `PATH` and `HOME`. Alternatively, if you can omit *all* env variables that were auto-added outside the config (is that visible in docker inspect at all?) that would work too.

itamarst:
I don't think we can merge this as is due to the calculation issue. We can discuss offline if you want.

flocker/node/functional/test_docker.py
```diff
@@ -385,6 +385,28 @@ def got_list(units):
...
+
+            def check_units(units):
+                unit = [unit for unit in units if unit.name == name][0]
+                self.assertIsNone(unit.environment)
+
+            deferred_units.addCallback(check_units)
+        d.addCallback(started)
+        return d
+
     def test_add_with_volumes(self):
```
itamarst: 1. Need test for inclusion of custom environment variables.
2. Need test for exclusion of environment variables present in the image.

itamarst: I guess first one is covered by the generic tests below, but second one still needs a test.

flocker/node/test/test_deploy.py
```diff
@@ -788,6 +788,42 @@ def test_discover_multiple(self):
...
+            u'example.com',
+            self.volume_service,
+            docker_client=fake_docker,
+            network=self.network
+        )
+        d = api.discover_local_state()
+
+        self.assertItemsEqual(pset(applications),
+                              sorted(self.successResultOf(d).applications))
+
```
itamarst: Might be worth having a test that combines both links and environment variables, since that code has to handle both.

itamarst:
Looks good, except for a bunch of missing tests. Once those are in and buildbot is green merge (don't want to merge this until we're sure acceptance tests pass!).

PR #1239, opened 2015-03-23
[FLOC-1578] Move container API

dwgebler:
Fixes FLOC-1578


docs/advanced/api_examples.yml
```diff
@@ -292,6 +292,26 @@
...
   id:
+    "move container"
+
+  doc: |
+    Move an existing named container to a new host.
+
+  request: |
+    POST /v1/configuration/containers/webserver HTTP/1.1
+
+    {"host": "%(NODE_0)s"}
```
itamarst: Maybe NODE_1?

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,143 @@ def test_container_with_volume(self):
...
+    Tests for the container configuration update endpoint at
+    ``/containers/<containername>``.
+    """
+    def _create_container(self):
+        """
+        Utility function to create a container configuration via the API.
+        """
+        saving = self.persistence_service.save(Deployment(
+            nodes={
+                Node(hostname=self.NODE_A),
```
itamarst: Maybe these should each have unrelated application on them, so that you can see that it's actually moving and not wiping out existing unrelated applications.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,143 @@ def test_container_with_volume(self):
...
+        """
+        An API request to update a named container's host to a different host
+        results in an error if the named container does not exist.
+        """
+        return self.assertResult(
+            b"POST", b"/configuration/containers/somecontainer",
+            {u"host": self.NODE_A}, NOT_FOUND,
+            {u"description": u"Container not found."},
+        )
+
```
itamarst: Test for non-existent host.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,143 @@ def test_container_with_volume(self):
...
+                u"restart_policy": {u"name": u"never"}
+            }
+            return self.assertResult(
+                b"POST", b"/configuration/containers/mycontainer",
+                request, OK, result
+            )
+
+        d.addCallback(update_container)
+        return d
+
```
itamarst: You didn't implement or test the bit where datasets are moved along with the application.

itamarst:
Looks good, other than minor comments... but is missing a chunk of functionality. Address and resubmit.

flocker/control/test/test_httpapi.py
dwgebler: Test and implementation added.

flocker/control/test/test_httpapi.py
dwgebler: New test added.

flocker/control/httpapi.py
```diff
@@ -664,6 +664,89 @@ def saved(_):
...
+            been updated.
+        """
+        deployment = self.persistence_service.get()
+
+        for node in deployment.nodes:
+            for application in node.applications:
+                if application.name == name:
+                    target_node = self._find_node_by_host(host, deployment)
+                    # We only need to perform a move if the node currently
+                    # hosting the container is not the node it's already on.
```
dwgebler: Oops, this comment doesn't make any sense. Obvious what I meant to write but I'll quickly do a commit to correct it.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+                            ),
+                            Application(
+                                name='mycontainer',
+                                image=DockerImage.from_string('busybox')
+                            ),
+                        ]
+                    ),
+                    Node(hostname=self.NODE_B),
+                }
+            )
```
exarkun: You could probably replace this `Deployment` instantiation with a `self.persistence_service.get()` call **before** issuing the API call.  Then the assertion is more faithfully "the configuration is unchanged" because you're comparing the configuration before to the configuration afterwards.  The result is the same for now but it's less code and it's more robust to future incidental `Deployment` / configuration changes that would invalidate the assumption made here that this **particular** `Deployment` is exactly what the configuration looks like before the no-op move is made.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+                        hostname=self.NODE_B,
+                        applications=[
+                            Application(
+                                name='mycontainer',
+                                image=DockerImage.from_string('busybox')
+                            ),
+                        ]
+                    ),
+                }
+            )
```
exarkun: Similarly, you might be able to grab the config before making the POST and then use a much shorter `transform` call to construct the expected value.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+    def test_update_moves_dataset(self):
+        """
+        An API request to update a named container's host to a different host
+        where the container has an attached dataset results in an updated
+        configuration, with the dataset's primary host also moved to the new
+        container host.
+        """
+        manifestation = _manifestation()
+        application = Application(
+            name='postgres',
```
exarkun: `u"postgres"`

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+            image=DockerImage.from_string('postgres'),
```
exarkun: `u"postgres"`

flocker/control/test/test_schemas.py
```diff
@@ -28,6 +28,30 @@
...
+        # Host wrong type
+        {'host': 1},
+        # Host not a host
+        {'host': 'idonotexist'},
+        # Extra properties
+        {'host': '192.168.0.3', 'image': 'nginx:latest'},
+    ],
+    passing_instances=[
+        {'host': '192.168.0.3'},
+    ],
```
exarkun: All of these strings should be unicode, I think.

docs/advanced/api_examples.yml
```diff
@@ -292,6 +292,26 @@
 
 -
   id:
+    "move container"
+
+  doc: |
+    Move an existing named container to a new host.
```
exarkun: Maybe worth mentioning that datasets are moved with it.

flocker/control/httpapi.py
```diff
@@ -664,6 +664,89 @@ def saved(_):
...
+                        # Finally we can now remove the manifestation from the
+                        # current host too.
+                        if application.volume is not None:
+                            dataset_id = application.volume.dataset.dataset_id
+                            node = node.transform(
+                                ("manifestations", dataset_id), discard
+                            )
+                        # Before updating the deployment and saving.
+                        deployment = deployment.update_node(node)
+                        deployment = deployment.update_node(target_node)
```
exarkun: I think that from `if node.hostname != host` to here would make a good method of `Deployment`.  `move_container` or something.  Do all this logic and return the new `Deployment` instance.  This method gets nicely shorter and the configuration mangling logic gets directly unit testable and re-usable (which is potentially beneficial to the API test suite).

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+                    Node(
+                        hostname=self.NODE_A,
+                        applications=[
+                            Application(
+                                name='leavemealone',
+                                image=DockerImage.from_string('busybox'),
+                            ),
+                            Application(
+                                name='mycontainer',
+                                image=DockerImage.from_string('busybox')
```
exarkun: These four strings should be unicode.

flocker/control/test/test_httpapi.py
```diff
@@ -1309,6 +1309,263 @@ def test_container_with_volume(self):
...
+                                image=DockerImage.from_string('busybox'),
+                            ),
+                        ]
+                    ),
+                    Node(
+                        hostname=self.NODE_B,
+                        applications=[
+                            Application(
+                                name='mycontainer',
+                                image=DockerImage.from_string('busybox')
```
exarkun: These four strings should be unicode.

exarkun:
Thanks.  This looks great.  Please address the remaining inline comments and then merge when buildbot is green (or as green as it can be at the moment :cry:).

PR #1240, opened 2015-03-23
[FLOC-1581] List containers state

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1581

### DIFF
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -268,7 +268,7 @@
 
   requires:
     - "create container"
-    - "create container with restart policy"
+    - "create container with attached volume"
 
   request: |
     GET /v1/configuration/containers HTTP/1.1
@@ -285,10 +285,12 @@
       },
       {
         "host": "%(NODE_0)s",
-        "name": "database",
-        "image": "postgres:latest",
-        "restart_policy": {"name": "on-failure", "maximum_retry_count": 5}
-      },
+        "name": "webserver",
+        "image": "nginx:latest",
+        "volumes": [{"dataset_id": "886ed03a-5606-453a-94a9-a1cbaf35164c",
+                     "mountpoint": "/usr/share/nginx/html"}],
+        "restart_policy": {"name": "never"}
+      }
     ]
 
 -
--- a/docs/advanced/api_examples.yml
+++ b/docs/advanced/api_examples.yml
@@ -284,8 +284,8 @@
         "restart_policy": {"name": "never"}
       },
       {
-        "host": "%(NODE_0)s",
-        "name": "webserver",
+        "host": "%(NODE_1)s",
+        "name": "webserver2",
         "image": "nginx:latest",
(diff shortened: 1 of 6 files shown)


docs/advanced/api_examples.yml
```diff
@@ -292,6 +292,39 @@
 
 -
   id:
+    "get actual containers"
+
+  doc: |
+    Get a list of all containers that are present in the cluster.
+
+  requires:
+    - "create container"
```
exarkun: Also `create container with restart policy` it looks like?  Since there are two containers in the response.

flocker/control/test/test_httpapi.py
```diff
@@ -2666,3 +2666,118 @@ def test_deleted(self):
...
+                applications={expected_application},
+            )
+        )
+        expected_dict = dict(
+            name=u"myapp",
+            host=expected_hostname,
+            image=u"busybox:latest",
+            running=True,
+            restart_policy={u"name": u"never"},
+        )
```
exarkun: There's more fields that could possibly be returned in the response.  Pretty sure that will work since you re-used `container_configuration_response` but it might not be bad to have a test exercise them.

docs/advanced/api_examples.yml
```diff
@@ -292,6 +292,39 @@
...
+        "host": "%(NODE_0)s",
+        "name": "database",
+        "image": "postgres:latest",
+        "restart_policy": {"name": "on-failure", "maximum_retry_count": 5},
+        "running": true
+      },
+    ]
+
+-
+  id:
```
exarkun: These examples don't show off all the fields that are available.  We should think about how to share the most information we can here.  Complete examples are a really easy way for people to understand how the API works.

flocker/control/httpapi.py
```diff
@@ -467,6 +467,38 @@ def get_containers_configuration(self):
...
+    def get_containers_state(self):
+        """
+        Get the containers present in the cluster.
+
+        :return: A ``list`` of ``dict`` representing each of the containers
+            that are configured to exist anywhere on the cluster.
+        """
+        result = []
+        deployment = self.cluster_state_service.as_deployment()
+        for node in deployment.nodes:
```
exarkun: `deployment.applications()` will do one of these for loops for you.

exarkun: Ah but it destroys host information.  Oh well nevermind.

exarkun:
Thanks.  Looks good apart from those minor points.  Apart from that ... end-to-end tests?  Are we doing those for the container API?  Later?

Please address inline comments, maybe file a follow-up issue for end-to-end-testing, and merge.

itamarst:
First acceptance we have is FLOC-1567, and that will need to use this code path too so should do.

docs/advanced/api_examples.yml
itamarst: There is field and schema documentation... and too many fields and the details get overwhelmning. But maybe it's worth having something with completely different fields than first item, so it's clear there's a variety of options.

PR #1242, opened 2015-03-24
Remove known hosts when running vagrant tests

adamtheturtle:
Fixes FLOC-1584

Jean-Paul pointed out that this is related to https://clusterhq.atlassian.net/browse/FLOC-1385, and it seems that this has been fixed too. This branch modifies what I think is the fix for that, so I think that should be closed with FLOC-1584.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -69,6 +69,16 @@ def extend_environ(**kwargs):
     return env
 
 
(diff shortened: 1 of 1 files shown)


adamtheturtle:
While two of the acceptance tests fail on this branch, this seems unrelated.

itamarst:
Looks good, please merge.

PR #1243, opened 2015-03-24
[FLOC-1567] Acceptance test: container with dataset

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1567

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -16,11 +16,31 @@ from unittest import SkipTest
 from treq import get, post, content, delete, json_content
 from pyrsistent import PRecord, field, CheckedPVector
 
-from ..testtools import (
-    loop_until, MONGO_IMAGE, require_mongo, verify_socket, get_mongo_client,
+from ..testtools import loop_until
+from .testtools import (
+    MONGO_IMAGE, require_mongo, get_mongo_client,
 )
 from ..node.agents.test.test_blockdevice import REALISTIC_BLOCKDEVICE_SIZE
 from ..control.httpapi import REST_API_PORT
+
+
+def verify_socket(host, port):
+    """
+    Wait until the destionation can be connected to.
+
+    :param bytes host: Host to connect to.
+    :param int port: Port to connect to.
+
+    :return Deferred: Firing when connection is possible.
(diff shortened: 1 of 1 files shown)


flocker/acceptance/test_api.py
```diff
@@ -17,10 +17,32 @@
...
+    """
+    def can_connect():
+        s = socket.socket()
+        conn = s.connect_ex((host, port))
+        print "Connection status", conn
+        return False if conn else True
+
+    dl = loop_until(can_connect)
+    return dl
+
```
dwgebler: Moving `verify_socket` is not strictly related to the issue or used in the "create container with dataset" test. Still, seems a small enough inline improvement that it's worth keeping in this branch.

dwgebler:
The only thing I wonder is if we should maybe have some cleanup code added to these tests to leave the VMs / acceptance nodes in the same state they were before the test suite was run. It's fine if you run these on newly spun-up VMs each time but you can't run the `test_api` suite twice on the same node instances without failures resulting from the lack of cleanup.

Otherwise, tests pass and this looks good to merge.

flocker/acceptance/test_api.py
```diff
@@ -17,10 +17,32 @@
...
+    :param bytes host: Host to connect to.
+    :param int port: Port to connect to.
+
+    :return Deferred: Firing when connection is possible.
+    """
+    def can_connect():
+        s = socket.socket()
+        conn = s.connect_ex((host, port))
+        print "Connection status", conn
+        return False if conn else True
```
exarkun: Such a terrible way to say `return conn != 0`

exarkun:
> It's fine if you run these on newly spun-up VMs each time but you can't run the test_api suite twice on the same node instances without failures resulting from the lack of cleanup.

This seems pretty significant.  It's already amazingly hard to run the acceptance tests.  Let's not make it even harder.

flocker/acceptance/test_api.py
itamarst: Just moved that code, it was already there!

itamarst:
I added some relevant cleanup code in #1246; I'm going to open a follow up issue to do it right.

PR #1245, opened 2015-03-25
Fix cloud-known hosts too.

tomprince:


adamtheturtle:
Good spot, merge when Buildbot is happy enough.

adamtheturtle:
Actually this is "obviously right" :laughing: so I'll merge.

PR #1247, opened 2015-03-25
[FLOC-1542] Split container discovery

itamarst:
Will be more readable once #1233 is merged (which can be done as soon as buildbot is green).

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -425,8 +425,9 @@ class P2PManifestationDeployer(object):
     :ivar VolumeService volume_service: The volume manager for this node.
     """
     def discover_local_state(self, local_state):
-        # XXX Move manifestation discovery code out of ApplicationNodeDeployer,
-        # Return a NodeState instance with applications=None.
+        # XXX Move manifestation and path discovery code out of
+        # ApplicationNodeDeployer, Return a NodeState instance with
+        # applications=None and ports=None.
         pass
 
     def calculate_necessary_state_changes(self, *args, **kwargs):
@@ -471,8 +472,8 @@ class ApplicationNodeDeployer(object):
             ``Application``. ``NodeState.manifestations`` and
             ``NodeState.paths`` will not be filled in.
         """
-        # XXX manifestations lookup will be moved away, will return
-        # NodeState with manifestations=None.
+        # XXX manifestations and path lookup will be moved away, will return
+        # NodeState with manifestations=None and paths=None.
 
         # Add real namespace support in
         # https://clusterhq.atlassian.net/browse/FLOC-737; for now we just
@@ -865,7 +866,7 @@ class P2PNodeDeployer(object):
         self.manifestations_deployer = ...
         self.applications_deployer = ...
 
-    def discover_local_state(self, cluster_state):
+    def discover_local_state(self, local_state):
         # 1. Lookup local state via self.manifestations_deployer
         # 2. Update cluster_state
         # 3. pass that to applications_deployer.discover_local_state
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -519,8 +519,8 @@ class ApplicationNodeDeployer(object):
                     # we assume all volumes are datasets
                     docker_volume = list(unit.volumes)[0]
                     try:
-                        manifestation = path_to_manifestations.get(
-                            docker_volume.node_path)
+                        manifestation = path_to_manifestations[
+                            docker_volume.node_path]
                     except KeyError:
                         # Apparently not a dataset we're managing, give up.
                         volume = None
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -7,7 +7,7 @@ Deploy applications on nodes.
 
 from itertools import chain
 
-from zope.interface import Interface, implementer
+from zope.interface import Interface, implementer, Attribute
 
 from characteristic import attributes
 
@@ -80,6 +80,9 @@ class IDeployer(Interface):
     changes to bring local state and desired cluster configuration into
     alignment.
     """
+    hostname = Attribute(
+        "The hostname of the node this deployer is managing.")
+
     def discover_local_state(local_state):
         """
         Discover the local state, i.e. the state which is exclusively under
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -892,16 +892,28 @@ class P2PNodeDeployer(object):
     """
     def __init__(self, hostname, volume_service, docker_client=None,
                  network=None):
-        self.manifestations_deployer = None
-        self.applications_deployer = None
+        self.manifestations_deployer = P2PManifestationDeployer(
+            hostname, volume_service)
+        self.applications_deployer = ApplicationNodeDeployer(
+            hostname, docker_client, network)
+        self.hostname = hostname
+        self.volume_service = self.manifestations_deployer.volume_service
+        self.docker_client = self.applications_deployer.docker_client
+        self.network = self.applications_deployer.network
 
     def discover_local_state(self, local_state):
-        # 1. Lookup local state via self.manifestations_deployer
-        # 2. Update cluster_state
-        # 3. pass that to applications_deployer.discover_local_state
-        # 3. Combine the two results
-        # Result is NodeState that looks like existing code.
-        pass
+        d = self.manifestations_deployer.discover_local_state(local_state)
+
+        def got_manifestations_state(manifestations_state):
+            app_discovery = self.applications_deployer.discover_local_state(
+                manifestations_state)
+            app_discovery.addCallback(
+                lambda app_state: app_state.set(
+                    "manifestations", manifestations_state.manifestations).set(
+                    "paths", manifestations_state.paths))
+            return app_discovery
+        d.addCallback(got_manifestations_state)
+        return d
 
     def calculate_necessary_state_changes(self, *args, **kwargs):
         # Calculation will be split up in FLOC-1553.
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -425,56 +425,6 @@ class P2PManifestationDeployer(object):
     :ivar VolumeService volume_service: The volume manager for this node.
     """
     def discover_local_state(self, local_state):
-        # XXX Move manifestation and path discovery code out of
-        # ApplicationNodeDeployer, Return a NodeState instance with
-        # applications=None and ports=None.
-        pass
-
-    def calculate_necessary_state_changes(self, *args, **kwargs):
-        # Does nothing in this branch. Follow up will move
-        # calculate_necessary_state_changes code here.
-        return Sequentially(changes=[])
-
-
-@implementer(IDeployer)
-class ApplicationNodeDeployer(object):
-    """
-    Discover and calculate changes for applications running on a node.
-
-    :ivar unicode hostname: The hostname of the node that this is running
-            on.
-    :ivar VolumeService volume_service: The volume manager for this node.
-    :ivar IDockerClient docker_client: The Docker client API to use in
-        deployment operations. Default ``DockerClient``.
-    :ivar INetwork network: The network routing API to use in
-        deployment operations. Default is iptables-based implementation.
-    """
-    def __init__(self, hostname, volume_service, docker_client=None,
-                 network=None):
-        self.hostname = hostname
-        if docker_client is None:
-            docker_client = DockerClient()
-        self.docker_client = docker_client
-        if network is None:
-            network = make_host_network()
-        self.network = network
-        self.volume_service = volume_service
-
-    def discover_local_state(self, local_state):
-        """
-        List all the ``Application``\ s running on this node.
-
-        The given local state is used to figure out if applications have
-        attached volumes that are specific manifestations.
-
-        :return: A ``Deferred`` which fires with a ``NodeState`` instance
-            with information only about
-            ``Application``. ``NodeState.manifestations`` and
-            ``NodeState.paths`` will not be filled in.
-        """
-        # XXX manifestations and path lookup will be moved away, will return
-        # NodeState with manifestations=None and paths=None.
-
         # Add real namespace support in
         # https://clusterhq.atlassian.net/browse/FLOC-737; for now we just
         # strip the namespace since there will only ever be one.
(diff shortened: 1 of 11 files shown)


flocker/node/_deploy.py
```diff
@@ -410,9 +415,30 @@ def run(self, deployer):
...
+    """
+    Discover and calculate changes for peer-to-peer manifestations
+    (e.g. ZFS) on a node.
+
+    :ivar unicode hostname: The hostname of the node that this is running
+            on.
+    :ivar VolumeService volume_service: The volume manager for this node.
+    """
+    def discover_local_state(self, local_state):
+        # XXX Move manifestation discovery code out of ApplicationNodeDeployer,
```
exarkun: including path information, I suppose

flocker/node/_deploy.py
```diff
@@ -816,3 +850,30 @@ def find_dataset_changes(hostname, current_state, desired_state):
...
+
+    Temporary expedient, to be removed in FLOC-1553 or perhaps another
+    sub-task of FLOC-1443.
+    """
+    def __init__(self, hostname, volume_service, docker_client=None,
+                 network=None):
+        self.manifestations_deployer = ...
+        self.applications_deployer = ...
+
+    def discover_local_state(self, cluster_state):
```
exarkun: in the interface, the new argument is `local_state`

flocker/node/_deploy.py
```diff
@@ -80,11 +80,17 @@ class IDeployer(Interface):
     changes to bring local state and desired cluster configuration into
     alignment.
     """
-    def discover_local_state():
+    def discover_local_state(local_state):
```
exarkun: This impacts the caller.  Can you add something about what changes this will require in `_loop.py`?

exarkun:
Thanks.  Seems sensible.  Some inconsistencies, probably fine, but please address those and submit for one more quick look.

flocker/node/_loop.py
```diff
@@ -275,14 +275,15 @@ def output_STORE_INFO(self, context):
...
+        # (i.e. we know nothing).
+        d = DeferredContext(self.deployer.discover_local_state(
+            known_local_state))
 
         def got_local_state(local_state):
             # Current cluster state is likely out of date as regards the local
             # state, so update it accordingly.
-            # XXX FLOC-1542 will need to deal with partial updates, and
-            # make sure there is no duplication with code in
-            # flocker.control._clusterstate.
```
exarkun: Something more necessary here?  Call `cluster_state.update_node` instead of `cluster_state.set`?

Pre-existing issue - the attribute is named `cluster_state` but the documentation for the class says `state` and that it is a `Deployment` but I think it's a `DeploymentState` now.

exarkun:
Great, thanks.  Left one more question inline but I'm pretty sure I know the answer.  Please proceed.

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+        """
+        ``ApplicationNodeDeployer.discover_local_state`` returns ``NodeState``
+        with a a list of running ``Application``\ s; one for each active
+        container.
+        """
+        expected_application_name = u'site-example.com'
+        unit = Unit(name=expected_application_name,
+                    container_name=expected_application_name,
+                    container_image=u"flocker/wordpress:latest",
+                    activation_state=u'active')
```
exarkun: This unit is repeated a lot of times in the following tests (often with minor variations).  Most or all of the duplication could be removed if `Unit` were switched to `pyrsistent` and a template `Unit` created that each test could either use directly or use to create its slightly different desired `Unit`.

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+        fake_docker = FakeDockerClient(units={expected_application_name: unit})
+        application = Application(
+            name=unit.name,
+            image=DockerImage.from_string(unit.container_image)
+        )
```
exarkun: :question:

I started to write a comment about how an `application_from_unit` helper function would aid in removing some of the repetition of `Application` construction in these tests.  That led me to realize the point of most of these tests is exactly to test various edge cases in that function (if it existed).  Perhaps if it did, instead of the logic being embedded in the middle of `discover_node_state`, these tests would be easier to write.

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+        """
+        environment_variables = (
+            ('CUSTOM_ENV_A', 'a value'),
+            ('CUSTOM_ENV_B', 'something else'),
+        )
+        link_environment_variables = (
+            ('APACHE_PORT_80_TCP', 'tcp://example.com:8080'),
+            ('APACHE_PORT_80_TCP_PROTO', 'tcp'),
+            ('APACHE_PORT_80_TCP_ADDR', 'example.com'),
+            ('APACHE_PORT_80_TCP_PORT', '8080'),
```
exarkun: `b".."`

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+
+    def test_discover_application_with_environment(self):
+        """
+        An ``Application`` with ``Environment`` objects is discovered from a
+        ``Unit`` with ``Environment`` objects.
+        """
+        environment_variables = (
+            ('CUSTOM_ENV_A', 'a value'),
+            ('CUSTOM_ENV_B', 'something else'),
+        )
```
exarkun: `b".."`

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+        An ``Application`` with ``Link`` objects is discovered from a ``Unit``
+        with environment variables that correspond to an exposed link.
+        """
+        fake_docker = FakeDockerClient()
+        applications = [
+            Application(
+                name=u'site-example.com',
+                image=DockerImage.from_string(u'clusterhq/wordpress:latest'),
+                links=frozenset([
+                    Link(local_port=80, remote_port=8080, alias='APACHE')
```
exarkun: `u".."`

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+            )
+        ]
+        api = ApplicationNodeDeployer(
+            u'example.com',
+            docker_client=fake_docker,
+            network=self.network
+        )
+        for app in applications:
+            StartApplication(
+                hostname='node1.example.com', application=app
```
exarkun: `u".."` ... and I'm gonna stop commenting on this type of thing.

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+        ]
+        api = ApplicationNodeDeployer(
+            u'example.com',
+            docker_client=fake_docker,
+            network=self.network
+        )
+        d = api.discover_local_state(NodeState(hostname=u'example.com'))
+
+        self.assertItemsEqual(pset(applications),
+                              sorted(self.successResultOf(d).applications))
```
exarkun: are `pset` and `sorted` actually helpful here?  `assertItemsEqual`'s job is to handle those details, isn't it?  Also comparing a set to a list is wonky.

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+            )
+        ]
+        api = ApplicationNodeDeployer(
+            u'example.com',
+            docker_client=fake_docker,
+            network=self.network
+        )
+        d = api.discover_local_state(NodeState(hostname=u'example.com'))
+
+        self.assertEqual(sorted(applications),
```
exarkun: `assertItemsEqual`?

flocker/node/_deploy.py
```diff
@@ -80,11 +80,20 @@ class IDeployer(Interface):
     changes to bring local state and desired cluster configuration into
     alignment.
     """
-    def discover_local_state():
+    hostname = Attribute(
+        "The hostname of the node this deployer is managing.")
```
exarkun: Please add type information to this documentation.

flocker/node/_deploy.py
```diff
@@ -410,36 +418,20 @@ def run(self, deployer):
...
         self.volume_service = volume_service
 
-    def discover_local_state(self):
-        """
-        List all the ``Application``\ s running on this node.
-
-        :returns: A ``Deferred`` which fires with a ``NodeState``
-            instance.
-        """
+    def discover_local_state(self, local_state):
```
exarkun: Docstring

flocker/node/_deploy.py
```diff
@@ -455,14 +447,92 @@ def map_volumes_to_size(volumes):
...
 
-        def applications_from_units(result):
-            units, available_manifestations = result
-            applications = []
+        def got_volumes(available_manifestations):
             manifestations = []  # Manifestation objects we're constructing
             manifestation_paths = {dataset_id: path for (path, (dataset_id, _))
                                    in available_manifestations.items()}
+
+            manifestations += list(
```
exarkun: It was an empty list before this.  Just `manifestations = ...`, perhaps, and lose `manifestations = []` above?

flocker/node/_deploy.py
```diff
@@ -455,14 +447,92 @@ def map_volumes_to_size(volumes):
...
+                manifestations={manifestation.dataset_id: manifestation
+                                for manifestation in manifestations},
+                paths=manifestation_paths,
+            )
+        volumes.addCallback(got_volumes)
+        return volumes
+
+    def calculate_necessary_state_changes(self, *args, **kwargs):
+        # Does nothing in this branch. Follow up will move
+        # calculate_necessary_state_changes code here.
```
exarkun: Add an issue link.

flocker/node/_deploy.py
```diff
@@ -455,14 +447,92 @@ def map_volumes_to_size(volumes):
...
+        """
+        List all the ``Application``\ s running on this node.
+
+        The given local state is used to figure out if applications have
+        attached volumes that are specific manifestations.
+
+        :return: A ``Deferred`` which fires with a ``NodeState`` instance
+            with information only about
+            ``Application``. ``NodeState.manifestations`` and
+            ``NodeState.paths`` will not be filled in.
```
exarkun: Looks oddly mis-wrapped.

flocker/node/_deploy.py
```diff
@@ -658,6 +714,10 @@ def calculate_necessary_state_changes(self, local_state,
         for application_name in applications_to_inspect:
             inspect_desired = desired_applications_dict[application_name]
             inspect_current = current_applications_dict[application_name]
+            # Current state never has metadata, but that's OK:
+            if inspect_desired.volume:
```
exarkun: `... is not None:`

flocker/node/_deploy.py
```diff
@@ -733,7 +793,13 @@ def change_node_state(deployer, desired_configuration,  current_cluster_state):
 
     :return: ``Deferred`` that fires when the necessary changes are done.
     """
-    d = deployer.discover_local_state()
+    nodes = [node for node in current_cluster_state.nodes
+             if node.hostname == deployer.hostname]
+    if nodes:
+        node = [nodes[0]]
```
exarkun: This seems like it must be wrong (missing test case?).  `node = nodes[0]`?

flocker/node/test/test_deploy.py
```diff
@@ -666,8 +669,7 @@ def test_does_not_exist(self):
 # XXX For now we require volume names match application names,
 # see https://github.com/ClusterHQ/flocker/issues/49
 DATASET_ID = unicode(uuid4())
-DATASET = Dataset(dataset_id=DATASET_ID,
-                  metadata=pmap({u"name": APPLICATION_WITH_VOLUME_NAME}))
+DATASET = Dataset(dataset_id=DATASET_ID)
```
exarkun: That metadata wasn't there to cause any particular code paths to be exercised?  And it seems like the comment is out of date now (and perhaps has been for a while - since we switched to UUIDs for dataset ids, I suppose?).  The dataset no longer has a matching name nor a matching "name" metadata item and it seems the `ApplicationNodeDeployer` logic also doesn't care about either of these things (not sure when that changed).  If this is right, delete the comment.

flocker/node/test/test_deploy.py
```diff
@@ -723,7 +729,7 @@ def test_discover_none(self):
             docker_client=fake_docker,
             network=self.network
         )
-        d = api.discover_local_state()
+        d = api.discover_local_state(NodeState(hostname=u"example.com"))
```
exarkun: Lots of incredibly boring repetitive `NodeState` creation.  Refactor this.

flocker/node/test/test_deploy.py
```diff
@@ -1050,7 +1055,6 @@ def test_discover_locally_owned_volume_with_size(self):
                     manifestation=Manifestation(
                         dataset=Dataset(
                             dataset_id=DATASET_ID,
-                            metadata=pmap({"name": unit1.name}),
```
exarkun: This test method's docstring still makes claims about things with the "same name".

flocker/node/test/test_deploy.py
```diff
@@ -1308,6 +1311,572 @@ def test_discover_manifestation_paths(self):
...
+                         sorted(self.successResultOf(d).applications))
+
+    def test_unknown_manifestations(self):
+        """
+        If the given ``NodeState`` indicates ignorance of manifestations, the
+        ``ApplicationNodeDeployer`` doesn't bother doing any discovery and
+        just indicates ignorance of applications.
+
+        XXX file issue for "it's not possible to run the container agent
+        without any dataset agents".
```
exarkun: File that issue, I suppose.

flocker/node/test/test_deploy.py
```diff
@@ -2112,6 +2700,71 @@ def test_volume_resized_before_move(self):
...
+            volume_service, docker_client=docker,
+            network=make_memory_network()
+        )
+
+        changes = api.calculate_necessary_state_changes(
+            self.successResultOf(api.discover_local_state(current_nodes[0])),
+            desired_configuration=desired,
+            current_cluster_state=current,
+        )
+        self.assertEqual(changes, Sequentially(changes=[]))
```
exarkun: :question:

I'm tempted to suggest that a better testing strategy here would be to define the cluster configuration, discover state, calculate changes, **run** the changes, discover state, calculate changes again - and then assert that the changes are null.  This has the benefit of running less fake code.  I'm somewhat afraid that it will turn out to have the major downside of being broken in some random irrelevant way, though.  Try it if you want, I guess.

exarkun:
Thanks.  Please address inline comments and then merge.

exarkun:
(Also green buildbot etc)

flocker/node/test/test_deploy.py
itamarst: I ended up making corresponding shared `Application` instance, so that reduced duplication.

flocker/node/_deploy.py
itamarst: I suspect an acceptance test will catch this. Stupid legacy flocker-changestate stuff.

flocker/node/test/test_deploy.py
itamarst: At some point we switched matching to be based on paths, not name, yeah.

flocker/node/test/test_deploy.py
itamarst: This is the legacy testcase that is just there temporarily, so not going to be maintaining it.

flocker/node/test/test_deploy.py
itamarst: Hm. I'm guessing splitting up calculation logic is going to have sufficient impact on these tests that it's not worth messing with them too much.

PR #1248, opened 2015-03-25
[FLOC-1582] loopback dataset backend - destroy

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1491
Fixes https://clusterhq.atlassian.net/browse/FLOC-1493
Fixes https://clusterhq.atlassian.net/browse/FLOC-1582 (also FLOC-1508 which 1582 seems to be a duplicate of)


flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+    """
+    blockdevice_id = field(type=unicode, mandatory=True)
+    size = field(type=int, mandatory=True)
+    host = field(type=(unicode, type(None)), initial=None)
+    dataset_id = field(type=UUID, mandatory=True)
+
+
+@_logged_statechange
+@with_cmp(["change"])
+class DestroyBlockDeviceDataset(proxyForInterface(IStateChange, "change")):
```
itamarst: I'd much prefer the factory function idiom to this: shorter, less confusing.

flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+        self.volume = volume
+        sequence = Sequentially(changes=[
+            UnmountBlockDevice(volume=volume),
+            DetachVolume(volume=volume),
+            DestroyVolume(volume=volume),
+        ])
+
+        super(DestroyBlockDeviceDataset, self).__init__(sequence)
+
+    @property
```
itamarst: And I'd rather a quick version of FLOC-1591 were done than this. If you *do* stick to this `_action` isn't a great name, it's super confusing. `_eliot_action`?

flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+        with self._action:
+            # IStateChange.run nominally returns a Deferred.  Hook it up to the
+            # action properly.  Do this as part of FLOC-1549 or maybe earlier.
+            return original_run(self, deployer)
+
+    cls.run = run
+    return cls
+
+
+class BlockDeviceVolume(PRecord):
```
itamarst: I guess I didn't notice this in previous reviews, but... does the word "volume" contribute much?

flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+    """
+    A block device that may be attached to a host.
+
+    :ivar unicode blockdevice_id: The unique identifier of the block device.
```
itamarst: Presumably unique per host, not per cluster?

flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+        super(DestroyBlockDeviceDataset, self).__init__(sequence)
+
+    @property
+    def _action(self):
+        return DESTROY_BLOCK_DEVICE_DATASET(_logger, volume=self.volume)
+
+
+def _volume():
+    """
+    Create and return a ``PRecord`` ``field`` for to hold a
```
itamarst: "for to hold"

flocker/node/agents/blockdevice.py
```diff
@@ -112,6 +127,201 @@ class UnattachedVolume(VolumeException):
...
+    @property
+    def _action(self):
+        return DESTROY_VOLUME(_logger, volume=self.volume)
+
+    def run(self, deployer):
+        """
+        Use the deployer's ``IBlockDeviceAPI`` to destroy the volume.
+        """
+        # Make this asynchronous after FLOC-1549, probably as part of
+        # FLOC-1582.
```
itamarst: This is FLOC-1582...

flocker/node/agents/blockdevice.py
```diff
@@ -220,6 +440,21 @@ def attach_volume(blockdevice_id, host):
...
+        Detach ``blockdevice_id`` from whatever host it is attached to.
+
+        :param unicode blockdevice_id: The unique identifier for the block
+            device being detached.
+
+        :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
+            exist.
+        :raises UnattachedVolume: If the supplied ``blockdevice_id`` is
+            not attached to anything.
+        :returns: A ``BlockDeviceVolume`` with a ``host`` attribute set to
```
itamarst: This return result means every backend implementation has to implement the same in-memory generic functionality. Seems better to have the caller that cares do it, once, and make life easier for backend implemetnations.

flocker/node/agents/blockdevice.py
exarkun: That seems reasonable - but maybe this isn't a blocking issue?

flocker/node/agents/blockdevice.py
exarkun: `_eliot_action` seems reasonable to me.  I'd rather no block this on FLOC-1591 though.

flocker/node/agents/blockdevice.py
```diff
@@ -617,4 +871,32 @@ def calculate_necessary_state_changes(self, local_state,
...
+        """
+        :param dict configured_manifestations: The manifestations configured
+            for this node (like ``Node.manifestations``).
+
+        :return: A ``list`` of ``DestroyBlockDeviceDataset`` instances for each
+            volume that needs to be destroyed based on the given
+            configuration and the actual state of volumes (ie which exist
+            and which don't).
+        """
+        volumes = self.block_device_api.list_volumes()
```
itamarst: This seems bad... `calculate_etc` doesn't return Deferred because the presumption is that it's an in-memory operation, purely data structure transformation. I.e. it's presumed that local state, cluster state and cluster configuration are all you need. `DestroyBlockDeviceDataset` can do the dataset_id -> device mapping itself; it's even more inefficient, but we'd want to optimize this later anyway and we may as well go for correctness.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -146,6 +154,235 @@ def test_only_unattached_devices(self):
...
+    Tests for ``BlockDeviceDeployer.calculate_necessary_state_changes``
+    in the cases relating to dataset destruction.
+    """
+    def test_undeleted_dataset_not_deleted(self):
+        """
+        ``BlockDeviceDeployer.calculate_necessary_state_changes`` does not
+        calculate a change to destroy datasets that are not marked as deleted
+        in the configuration.
+        """
+        dataset_id = uuid4()
```
itamarst: Looks like some of this repeats over tests? Can it be moved into utility method?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -146,6 +154,235 @@ def test_only_unattached_devices(self):
...
+            paths={
+                unicode(dataset_id):
+                FilePath(b"/flocker/").child(bytes(dataset_id)),
+            },
+        )
+        cluster_state = Deployment(
+            nodes={local_state.to_node()}
+        )
+
+        local_config = local_state.to_node()
```
itamarst: This is going to fail; merge forward and maybe reuse the `to_node()` test utility I wrote for `test_loop.py`.

flocker/node/agents/blockdevice.py
itamarst: Not blocking, no.

flocker/node/agents/blockdevice.py
itamarst: OK.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -665,6 +902,178 @@ def test_get_device_path_device_repeatable_results(self):
...
+            volume.set(host=None),
+        )
+        self.assertEqual(
+            expected,
+            (set(self.api.list_volumes()), detached)
+        )
+
+        detached_error = fail_mount(device_path)
+
+        # Make an incredibly indirect assertion to try to demonstrate we've
```
itamarst: How about... `fdisk -l`?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -955,24 +1364,293 @@ def mountroot_for_test(test_case):
...
+        Instances of the type provide ``IStateChange``.
         """
-        self.assertTrue(
-            verifyObject(
-                IStateChange,
-                CreateBlockDeviceDataset(
-                    dataset=Dataset(dataset_id=unicode(uuid4())),
-                    mountpoint=FilePath('.')
-                )
+        self.assertTrue(verifyObject(IStateChange, self.state_change()))
```
itamarst: You should be using `flocker.node.test.test_deploy.make_istatechange_tests`.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -955,24 +1364,293 @@ def mountroot_for_test(test_case):
...
+        ))
+        self.assertTrue(a != b)
+
+    def verify_run_log(self, logger):
+        # One action is logged
+        action = assertHasAction(
+            self, logger, DESTROY_BLOCK_DEVICE_DATASET, succeeded=True)
+        all_such_actions = LoggedAction.of_type(
+            logger.messages, DESTROY_BLOCK_DEVICE_DATASET)
+        self.assertEqual([action], all_such_actions)
```
itamarst: How about just `[action] = all_such_action` and skip first line?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -955,24 +1364,293 @@ def mountroot_for_test(test_case):
...
+        volume = api.create_volume(
+            dataset_id=dataset_id, size=REALISTIC_BLOCKDEVICE_SIZE
+        )
+        volume = api.attach_volume(volume.blockdevice_id, node)
+        device = api.get_device_path(volume.blockdevice_id)
+        mountroot = mountroot_for_test(self)
+        mountpoint = mountroot.child(unicode(dataset_id).encode("ascii"))
+        mountpoint.makedirs()
+        check_output([b"mkfs", b"-t", b"ext4", device.path])
+        check_output([b"mount", device.path, mountpoint.path])
```
itamarst: There's no generic mount code you can use yet? I guess it's part of combo statechange for now
.

itamarst:
File an issue for the factory function thing, I guess. Merge forward, address as needed, then merge. Thanks!

flocker/node/agents/blockdevice.py
exarkun: No ... not really.  Maybe nothing.  There are probably a few other names that could be reconsidered now that things have taken a bit more shape.  I'll either do that now or file a follow-up issue if it seems too disruptive.

flocker/node/agents/blockdevice.py
exarkun: It needs to be unique across the cluster.  This is for values like an EBS volume id, not the local, OS-level identifier for the device.  It needs to be cluster-wide unique otherwise you won't know which block device you're trying to attach/detach/destroy.

PR #1251, opened 2015-03-26
[FLOC-1599] API endpoint for changing the maximum_size of a dataset

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1599

Note:  this PR is compared against loopback-destroy-FLOC-1582 because I was originally writing this as part of the design for https://clusterhq.atlassian.net/browse/FLOC-1504 which builds upon the work in the FLOC-1582 branch.

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -454,3 +454,22 @@ class DatasetAPITests(TestCase):
         # Check for 200 response code
         # Wait for the state of that dataset to show the updated size.
         1/0
+
+    def test_dataset_shrink_not_possible(self):
+        """
+        If the dataset has too much data to allow shrinking to the requested size...then what??
+        """
+        # Create a dataset with SMALL_DATASET_SIZE
+        # Somehow add data to the dataset
+        # Request a resize.
+        # ???
+        1/0
+
+    def test_dataset_shrink_not_valid(self):
+        """
+        If the requested maximum_size is smaller than the allowed minimum, then what??
+        """
+        # Create a dataset REALISTIC_BLOCKDEVICE_SIZE
+        # Request a new maximum_size < minimum defined in schema.
+        # Check response code.
+        1/0
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -444,3 +444,13 @@ class DatasetAPITests(TestCase):
         # http://doc-dev.clusterhq.com/advanced/api.html#post--v1-configuration-datasets-%28dataset_id%29
         # Wait for the state of that dataset to show the updated size.
         1/0
+
+    def test_dataset_shrink(self):
+        """
+        The size of a dataset can be decreased.
+        """
+        # Create a dataset with REALISTIC_BLOCKDEVICE_SIZE * 2
+        # Reconfigure that dataset to be REALISTIC_BLOCKDEVICE_SIZE
+        # Check for 200 response code
+        # Wait for the state of that dataset to show the updated size.
+        1/0
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -432,3 +432,15 @@ class DatasetAPITests(TestCase):
             return deleted
         created.addCallback(delete_dataset)
         return created
+
+
+    def test_dataset_grow(self):
+        """
+        The size of a dataset can be increased.
+        """
+        # Create a dataset with REALISTIC_BLOCKDEVICE_SIZE
+        # Reconfigure that dataset to be REALISTIC_BLOCKDEVICE_SIZE * 2
+        # Check for 200 response code
+        # http://doc-dev.clusterhq.com/advanced/api.html#post--v1-configuration-datasets-%28dataset_id%29
+        # Wait for the state of that dataset to show the updated size.
+        1/0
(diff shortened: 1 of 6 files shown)


flocker/acceptance/test_api.py
```diff
@@ -432,3 +432,44 @@ def not_exists():
...
+        """
+        # Create a dataset with REALISTIC_BLOCKDEVICE_SIZE * 2
+        # Reconfigure that dataset to be REALISTIC_BLOCKDEVICE_SIZE
+        # Check for 200 response code
+        # Wait for the state of that dataset to show the updated size.
+        1/0
+
+    def test_dataset_shrink_not_possible(self):
+        """
+        If the dataset has too much data to allow shrinking to the requested size...then what??
```
exarkun: The convergence attempts will fail ... indefinitely ... or until some data is removed.  It will be hard to write a test for this because we don't have a convergence failure reporting mechanism yet (and one isn't planned for 1.0).  I suggest skipping testing this at the API level (we should perhaps test it at the convergence agent level though).

flocker/acceptance/test_api.py
```diff
@@ -432,3 +432,44 @@ def not_exists():
...
+        """
+        # Create a dataset with SMALL_DATASET_SIZE
+        # Somehow add data to the dataset
+        # Request a resize.
+        # ???
+        1/0
+
+    def test_dataset_shrink_not_valid(self):
+        """
+        If the requested maximum_size is smaller than the allowed minimum, then what??
```
exarkun: Elsewhere we've discussed the need to let the backend contribute constraints to the configuration.  So probably makes sense to leave this case aside but make a note of it on https://clusterhq.atlassian.net/browse/FLOC-1579

flocker/control/httpapi.py
```diff
@@ -351,7 +354,7 @@ def saved(ignored):
                       '/v1/endpoints.json#/definitions/configuration_dataset'},
         schema_store=SCHEMAS
     )
-    def update_dataset(self, dataset_id, primary=None):
+    def update_dataset(self, dataset_id, primary=None, maximum_size=None):
```
exarkun: Will this work support removing a maximum size from a ZFS-backed dataset?  I think that's represented as a `maximum_size` of `None` in the current configuration model.

I think it would be okay to defer this work until later.  Resizing ZFS will work without it, it just won't be possible to express the change from a size limit to no size limit - which is probably not an important limitation in the near term.  But file an issue for addressing this later.

flocker/control/test/test_httpapi.py
```diff
@@ -1685,6 +1685,9 @@ class UpdatePrimaryDatasetTestsMixin(APITestsMixin):
     """
     Tests for the dataset modification endpoint at
     ``/configuration/datasets/<dataset_id>``.
+
+    XXX: Make it clear that these tests are only for changes to the ``primary``
+    attribute of a dataset.
```
exarkun: The mixin name reflects this... but maybe that's not sufficiently clear?

flocker/control/test/test_httpapi.py
```diff
@@ -1960,13 +1963,54 @@ def saved(ignored):
         'primary address, but the input schema currently requires the primary '
         'attribute. This will need to be addressed before or as part of '
         'https://clusterhq.atlassian.net/browse/FLOC-1404'
+        # Consider fixing this issue before implementing the resize API changes.
```
exarkun: Probably makes as much sense to do it in this branch as in a FLOC-1404 branch (FLOC-1404 just being the metadata update issue).

run-iblockdevice-acceptance-tests
```diff
@@ -108,6 +108,7 @@ REMOTE_AGENT_PID="${!}"
 
 FLOCKER_ACCEPTANCE_CONTROL_NODE="${FLOCKER_ACCEPTANCE_CONTROL_NODE}" \
 FLOCKER_ACCEPTANCE_AGENT_NODES="${FLOCKER_ACCEPTANCE_AGENT_NODE}" \
+# Run the dataset grow acceptance test
 trial flocker.acceptance.test_api.DatasetAPITests.test_dataset_deletion || echo "TEST FAILURE"
 
 # Kill agent and wait for it to exit
```
exarkun: Oh.  This PR isn't against master.  I see.

We probably should delete this file from the FLOC-1582 branch...

flocker/acceptance/test_api.py
exarkun: Although it's certainly possible to test against the statically defined minimum right now, so maybe that's fine.

flocker/control/httpapi.py
```diff
@@ -336,12 +336,15 @@ def saved(ignored):
 
         * Move a dataset from one node to another by changing the
           ``primary`` attribute.
+        # XXX: Update this comment.
         * In the future, update metadata and maximum size.
```
exarkun: Add a link to the update metadata issue, too.

flocker/control/test/test_httpapi.py
```diff
@@ -1960,13 +1963,54 @@ def saved(ignored):
...
     )
 
+    def test_no_primary(self):
+        """
+        An update request without a primary address is allowed.
```
exarkun: This requirement might be better expressed in the `UpdateSizeDatasetTestsMixin`.  Those tests will presumably issue requests without `primary` in the body.  If this test is written here, what will it be posting?  An empty object?  I suppose we can make that work if we want... but it's not a particular useful operation I think. :smile:

exarkun:
Thanks.

The schema needs to be updated as well (obviously, I suppose).  I don't think that will involve any surprises - we've done lots of schema changes by now.

The rest of this looks good.  Please proceed.  Make sure to create the next PR against master though.

PR #1252, opened 2015-03-26
Use a helper that captures all output instead of only some.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-167

Still a little bit of junk output but this gets rid of a about 99% of it.

### DIFF
--- a/flocker/testtools/__init__.py
+++ b/flocker/testtools/__init__.py
@@ -827,7 +827,7 @@ def run_process(command, *args, **kwargs):
     """
     Run a child process, capturing its stdout and stderr.
 
-    :param list command: An ``argv`` to use to launch the child process.
+    :param list command: An argument list to use to launch the child process.
 
     :raise CalledProcessError: If the child process has a non-zero exit status.
 
--- a/flocker/testtools/__init__.py
+++ b/flocker/testtools/__init__.py
@@ -21,7 +21,7 @@ from unittest import skipIf, skipUnless
 
 from subprocess import PIPE, STDOUT, CalledProcessError, Popen
 
-from pyrsistent import PRecord, field, optional
+from pyrsistent import PRecord, field
 
 from zope.interface import implementer
 from zope.interface.verify import verifyClass, verifyObject
--- a/flocker/testtools/__init__.py
+++ b/flocker/testtools/__init__.py
@@ -1,6 +1,8 @@
 # Copyright Hybrid Logic Ltd.  See LICENSE file for details.
 
-"""Various utilities to help with unit and functional testing."""
+"""
+Various utilities to help with unit and functional testing.
+"""
 
 from __future__ import absolute_import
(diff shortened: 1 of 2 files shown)


flocker/testtools/__init__.py
```diff
@@ -14,17 +16,19 @@
...
 import shutil
-from subprocess import check_call, check_output
 from functools import wraps
-from unittest import skipIf
-from unittest import skipUnless
+from unittest import skipIf, skipUnless
+
+from subprocess import PIPE, STDOUT, CalledProcessError, Popen
+
+from pyrsistent import PRecord, field, optional
```
adamtheturtle: lint error: optional is unused

flocker/testtools/__init__.py
```diff
@@ -592,11 +596,11 @@ def build(self, dockerfile_variables=None):
             b'--tag=%s' % (tag,),
             docker_dir.path
         ]
-        check_call(command)
+        run_process(command)
         # XXX until https://clusterhq.atlassian.net/browse/FLOC-409 is
```
adamtheturtle: This issue is fixed, so this comment should be removed or changed (in a follow-up if you prefer).

flocker/testtools/__init__.py
```diff
@@ -807,3 +811,35 @@ class CustomException(Exception):
...
+    command = field(type=list, mandatory=True)
+    output = field(type=bytes, mandatory=True)
+    status = field(type=int, mandatory=True)
+
+
+def run_process(command, *args, **kwargs):
+    """
+    Run a child process, capturing its stdout and stderr.
+
+    :param list command: An ``argv`` to use to launch the child process.
```
adamtheturtle: :question: This is a little bit confusing to me but might be Python terminology I've just not come across - it reads like `argv` is the type. A quick search's first results are all about `sys.argv` and don't really seem to fit with "an argv".

adamtheturtle:
Thanks @exarkun. This looks good (although I don't know pyrsistent and am just assuming that  nothing is crazy with `_ProcessResult).

I've tested this by:
* "eyeballing" some Buildbot output and seeing that there is significantly less "junk".
* Running an affected test (`flocker.node.functional.test_deploy.DeployerTests.test_cpu_share`) with a modification which made it fail to see that failures are reported as expected.

There are a couple more `check_call`s it might be worth changing, but as you said in the description, there is "Still a little bit of junk output".

I am unable to check the acceptance test output on Buildbot because they are not running at the moment.

One bit of the output which might be easy to fix is the warning about `service_identity`:

```
```

Perhaps it would be best to add `service_identity` as a `dev` dependency.

I think that because this is not a complete solution this should be after some minor changes have been made as you see appropriate, but the parent issue should remain open (and I guess moved out of this sprint) or another issue should be made for the other "junk".

PR #1254, opened 2015-03-26
[FLOC-1504] loopback dataset agent can increase the size of an attached volume

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1504

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -432,3 +432,26 @@ class DatasetAPITests(TestCase):
             return deleted
         created.addCallback(delete_dataset)
         return created
+
+    def test_dataset_grow_(self):
+        """
+        The maximum size of a dataset can be increased.
+
+        A not quite end-to-end test that the size of a dataset can be increased.
+
+        We can't issue the resize request from the API until
+        https://clusterhq.atlassian.net/browse/FLOC-1599 is resolved.
+
+        So manipulate the control service configuration file directly and
+        restart it....or something.
+        """
+        # Create a dataset of REALISTIC_BLOCKDEVICE_SIZE
+        # Wait for it to be created.
+        # Add some data to it somehow.
+        # Stop the dataset agent
+        # Stop the control service
+        # Wipe the configuration file
+        # Start the control service
+        # Create a dataset with the same dataset_id but with maximum_size=REALISTIC_BLOCKDEVICE_SIZE * 2
+        # Wait for the dataset to be resized.
+        # Check that the data (written earlier) is still intact and that the dataset state reports the correct new size.
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -230,6 +230,14 @@ class DestroyBlockDeviceDataset(proxyForInterface(IStateChange, "change")):
         return DESTROY_BLOCK_DEVICE_DATASET(_logger, volume=self.volume)
 
 
+# Implement:
+#     AttachVolume(volume=volume),
+#     MountVolume(volume=volume),
+# in order to carry out the sequence below.
+# Maybe use those in the CreateBlockDeviceDataset state change or paste in some
+# followup issues, or maybe do those issues first??
+
+
 @_logged_statechange
 @with_cmp(["change"])
 class ResizeBlockDeviceDataset(proxyForInterface(IStateChange, "change")):
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -455,6 +455,23 @@ class IBlockDeviceAPI(Interface):
             ``None``.
         """
 
+    # def resize_volume(blockdevice_id, size):
+    #     """
+    #     Resize an unattached ``blockdevice_id``.
+
+    #     :param unicode blockdevice_id: The unique identifier for the block
+    #         device being detached.
+
+    #     :param int size: The required size, in bytes, of the volume.
+
+    #     :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
+    #         exist.
+    #     :raises AlreadyAttached: If the supplied ``blockdevice_id`` is
+    #         already attached.
+    #     :returns: A ``BlockDeviceVolume`` with a ``size`` attribute equal to
+    #         the supplied ``size`` when the volume has been resized.
+    #     """
+
     def list_volumes():
         """
         List all the block devices available via the back end API.
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -899,10 +899,11 @@ class BlockDeviceDeployer(PRecord):
             for manifestation
             in manifestations_to_create
         )
+        # XXX: Maybe list volumes here and pass the list to each helper below.
+        deletes = self._calculate_deletes(configured_manifestations) # volumes=volumes)
+#        resizes = self._calculate_resizes(configured_manifestations, volumes=volumes)
 
-        deletes = self._calculate_deletes(configured_manifestations)
-
-        return InParallel(changes=creates + deletes)
+        return InParallel(changes=creates + deletes) # + resizes)
 
     def _calculate_deletes(self, configured_manifestations):
         """
@@ -914,6 +915,8 @@ class BlockDeviceDeployer(PRecord):
             configuration and the actual state of volumes (ie which exist
             and which don't).
         """
+        # XXX: maybe issue this API call once and supply the list of volumes to
+        # each of the _calculate_* private helper methods.
         volumes = self.block_device_api.list_volumes()
 
         delete_dataset_ids = set(
@@ -928,3 +931,18 @@ class BlockDeviceDeployer(PRecord):
             # Only destroy volumes belonging to deleted datasets.
             if unicode(volume.dataset_id) in delete_dataset_ids
         ]
+
+    # def _calculate_resizes(self, configured_manifestations):
+    #     """
+    #     :param dict configured_manifestations: The manifestations configured
+    #         for this node (like ``Node.manifestations``).
+
+    #     :return: A ``list`` of ``ResizeBlockDeviceDataset`` instances for each
+    #         volume that needs to be resized based on the given
+    #         configuration and the actual state of volumes (ie which have a size
+    #         that is different to the configuration)
+    #     """
+    #     # List volumes using self.api
+    #     # Loop through each comparing actual size with the configured size of
+    #     # the corresponding manifestation.
+    #     # Return ``ResizeBlockDeviceDataset`` for each resize.
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -155,6 +155,11 @@ DESTROY_VOLUME = ActionType(
     u"The volume for a block-device-backed dataset is being destroyed."
 )
 
+# Add new Eliot actions for:
+# RESIZE_BLOCK_DEVICE_DATASET
+# MOUNT_VOLUME
+# ATTACH_VOLUME
+
 
 def _logged_statechange(cls):
     """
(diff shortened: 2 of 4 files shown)


flocker/acceptance/test_api.py
```diff
@@ -432,3 +432,26 @@ def not_exists():
...
+    def test_dataset_grow_(self):
+        """
+        The maximum size of a dataset can be increased.
+
+        A not quite end-to-end test that the size of a dataset can be increased.
+
+        We can't issue the resize request from the API until
+        https://clusterhq.atlassian.net/browse/FLOC-1599 is resolved.
+
+        So manipulate the control service configuration file directly and
```
itamarst: I suggest just pushing the acceptance test into a separate issue that is blocked on 1599, no point wasting time on doing things the hard way when we can do the easier (and more realistic!) way soon.

flocker/node/agents/blockdevice.py
```diff
@@ -230,6 +235,46 @@ def _action(self):
...
+#     AttachVolume(volume=volume),
+#     MountVolume(volume=volume),
+# in order to carry out the sequence below.
+# Maybe use those in the CreateBlockDeviceDataset state change or paste in some
+# followup issues, or maybe do those issues first??
+
+
+@_logged_statechange
+@with_cmp(["change"])
+class ResizeBlockDeviceDataset(proxyForInterface(IStateChange, "change")):
```
itamarst: ???

itamarst: This idiom seems super-verbose and hacky compared to just having a utility function that creates the Sequentially things. See discussion in Zulip.

flocker/node/agents/blockdevice.py
```diff
@@ -230,6 +235,46 @@ def _action(self):
...
+        #     ResizeVolume(volume=volume, size=size),
+        #     AttachVolume(volume=volume),
+        #     MountVolume(volume=volume),
+        # ])
+        # super(ResizeBlockDeviceDataset, self).__init__(sequence)
+        pass
+
+
+    @property
+    def _action(self):
```
itamarst: This is a *super* confusing name for the method. "_eliot_action" maybe?

Or better yet get rid of `_logged_statechange` and just do https://clusterhq.atlassian.net/browse/FLOC-1591 before this. Yes, that won't give the best logging ever, but it'll give a decent baseline and the current approach seems like it's setting you up for lots of boilerplate copy/paste testing.

itamarst:
Just for future reference a 500 line design sketch is rather difficult to review. Will continue this later, or someone else can take it up if they want.

flocker/node/agents/blockdevice.py
exarkun: Agreed on this - though as with FLOC-1582 I don't think fixing this pattern needs to block this work.  However, take a look at the changes that were necessary in FLOC-1582.  The shape of `DeleteBlockDeviceDataset` changed a fair bit to account for some other required changes (not doing asynchronous work in `calculcate_necessary_state_changes`).  The same applies here.

flocker/node/agents/blockdevice.py
exarkun: The helper was changed to use the name `_eliot_action`.

flocker/node/agents/blockdevice.py
```diff
@@ -718,6 +839,17 @@ def detach_volume(self, blockdevice_id):
         volume_path.moveTo(new_path)
         return volume.set(host=None)
 
+    # def resize_volume(self, blockdevice_id, size):
+    #     """
+    #     Increase the size of the loopback backing file whilst maintaining its
+    #     sparseness.
+
+    #     Is that possible?
```
exarkun: Just truncate() it at the new size.  It should remain sparse.

flocker/node/agents/blockdevice.py
```diff
@@ -871,10 +1003,11 @@ def calculate_necessary_state_changes(self, local_state,
             for manifestation
             in manifestations_to_create
         )
+        # XXX: Maybe list volumes here and pass the list to each helper below.
+        deletes = self._calculate_deletes(configured_manifestations) # volumes=volumes)
```
exarkun: Note that `_calculate_deletes` changed to not list volumes.  This is because we want to keep `calculate_necessary_state_changes` as simple as we can.  One way to do that is to only operate on the simple model objects given as inputs.  For now we're accepting some extra complexity that results from this decision in the IStateChange implementations (eg, make the resize operation a no-op if the desired size is already the actual size).  Later, we'll fix this up by making `local_state` and `cluster_state` carry all of the extra information we need to make smarter decisions.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -633,6 +640,48 @@ def test_ignore_deleted_datasets(self):
...
+    """
+    Tests for ``BlockDeviceDeployer.calculate_necessary_state_changes``
+    in the cases relating to resizing a dataset.
+    """
+    def test_maximum_size_unchanged(self):
+        """
+        ``BlockDeviceDeployer.calculate_necessary_state_changes`` does not
+        calculate a change to shrink or grow datasets if the maximum_size of
+        the dataset has not changed.
+        """
```
exarkun: This test will probably not be able to pass in this branch.  I deleted some similar tests from the FLOC-1582 branch in afa1ffe1bf3c1a24bbd4457ac1b69dba8abe195a.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1074,6 +1129,78 @@ def test_attach_destroyed_volume(self):
...
+        # self.assertEqual(exception.args, (blockdevice_id,))
+        1/0
+
+    def test_resize_attached_volume(self):
+        """
+        ``resize_volume`` raises ``AlreadyAttached`` if the supplied
+        ``blockdevice_id`` is for a ``BlockDeviceVolume`` that is attached to a
+        host.
+
+        XXX: Should there be a similar test for an attempt to destroy an attached volume?
```
exarkun: Are actual volume providers unable to resize attached volumes?  Or will they succeed and this is merely something we should never ask them to do.  If the latter, we shouldn't make the interface require that they give us an exception.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1435,6 +1584,8 @@ def test_equal(self):
         """
         Two ``DestroyBlockDeviceDataset`` instances compare as equal if they
         are initialized with the same volume.
+
+        XXX: Move to shared mixin.
```
exarkun: Maybe not worth doing a lot of refactoring for these composed IStateChange implementations yet.  We're going to twiddle the implementation around a bit more before we find the final shape for them, I think.  We want something that's both easier to define and more introspectable than what ended up in the FLOC-1582 branch (which I expect will be much the same as what ends up in this branch).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1748,3 +1934,180 @@ def get_mounts():
...
             yield FilePath(device_path), FilePath(mountpoint), filesystem_type
+
+
+# def _make_resize_dataset():
+#     """
+#     Make a ``ResizeBlockDeviceDataset`` instance for
+#     ``make_state_change_tests``.
+#     """
+#     return ResizeBlockDeviceDataset(
+#         volume=_ARBITRARY_VOLUME,
```
exarkun: I had to change `DestroyBlockDeviceDataset` to accept a `dataset_id`.  I suspect the same will be necessary for `ResizeBlockDeviceDataset`.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1748,3 +1934,180 @@ def get_mounts():
...
+        1/0
+
+    def test_resize_unmounted_volume(self):
+        """
+        ``ResizeBlockDeviceDataset`` can resize an attached but unmounted volume.
+
+        The result is a larger block device with an expanded filesystem
+        containing the data that was their before the resize.
+        """
+        # XXX This situation shouldn't be possible, right?
```
exarkun: It may come up.  It should probably be handled, maybe not as part of this branch though.  The end state should probably be that the filesystem is **mounted** though.  We don't currently have any reason to have a volume attached to a node but the filesystem not mounted (so this is a state to repair, not restore).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1748,3 +1934,180 @@ def get_mounts():
...
+        1/0
+
+    def test_resize_unformatted_volume(self):
+        """
+        ``ResizeBlockDeviceDataset`` can resize an unformatted volume.
+
+        The result is a larger block device which is attached to the host but
+        which does not yet have a filesystem.
+        """
+        # XXX This situation shouldn't be possible, right?
```
exarkun: Same as above, I think.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1748,3 +1934,180 @@ def get_mounts():
...
+        1/0
+
+    def test_resize_unattached_volume(self):
+        """
+        ``ResizeBlockDeviceDataset`` can resize an unattached, unformatted volume.
+
+        The result is a larger block device which is unattached and without filesystem.
+        """
+        # XXX This situation shouldn't be possible, right?
```
exarkun: Hmmm.  As long as it's not attached maybe it doesn't matter if it's the wrong size.  As soon as someone actually wants to use it they'll have to attach it and then it'll get resized.  Probably good enough for now, at least.

exarkun:
Thanks.  Pointed out a few more things.  Unfortunately, the code this was based on has shifted a little bit.  Things will need to be updated to account for that.  Please do that, address comments, and resubmit.

flocker/acceptance/test_api.py
wallrj: The acceptance test has now been merged.

flocker/node/agents/blockdevice.py
wallrj: I updated the design to match the eventual implementation of DestroyBlockDevice

flocker/node/agents/blockdevice.py
wallrj: Also now changed in the design.

flocker/node/agents/blockdevice.py
wallrj: Ok. Noted in the design.

flocker/node/agents/blockdevice.py
wallrj: I'm struggling to see how this is going to work for resizes. I left a few comments in the updated design.

wallrj: Ah, perhaps I do get it. I updated my comments in the new design.

flocker/node/agents/test/test_blockdevice.py
wallrj: I left a comment in the new design about the need to supply the local_state in this case...in which case, perhaps this test can pass.

flocker/node/agents/test/test_blockdevice.py
wallrj: Fair point. I added a comment to the new design.

flocker/node/agents/test/test_blockdevice.py
wallrj: I added a note to the new design.

flocker/node/agents/test/test_blockdevice.py
wallrj: Done.

flocker/node/agents/test/test_blockdevice.py
wallrj: noted in the new design.

flocker/node/agents/test/test_blockdevice.py
wallrj: noted in new design.

wallrj:
Updated design in https://github.com/ClusterHQ/flocker/pull/1302

PR #1255, opened 2015-03-27
[FLOC-1279] Provsion acceptance testing nodes in parallel.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1279

TODO
- Adjust hombrew test.
- Adjust rackspace.


flocker/provision/_effect.py
```diff
@@ -0,0 +1,56 @@
...
+            result = yield effect
+            results.append(result)
+        except:
+            raise SequenceFailed(results=results,
+                                 error=sys.exc_info())
+
+    yield do_return(results)
+
+
+def sequence(_):
```
tomprince: I'm not sure how to differentiate the docstring for this from the one for `Sequence`.

flocker/provision/_install.py
```diff
@@ -5,13 +5,19 @@
 Install flocker on a remote node.
 """
 
-from pipes import quote as shell_quote
 import posixpath
 from textwrap import dedent
 from urlparse import urljoin
-from characteristic import attributes
+from effect import parallel, Func, Effect
```
tomprince: Despite all the lines changed, the only significant changes in this file are:
- Factoring out the code that talks ssh.
- Changing `configure_cluster` and `stop_cluster` to only return effects.

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+
+from ._model import Run, Sudo, Put, Comment, RunRemotely
+
+
+@attributes([
+    "deferred",
+    "address",
+    "username",
+])
+class CommandProtocol(LineOnlyReceiver, object):
```
tomprince: Can't use `pyrsistent`, since `LineOnlyReceiver` isn't immutable.

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+    :ivar Deferred deferred: Deferred to fire when the command finishes
+        If the command finished succesfuly, will fire with ``None``.
+        Otherwise, errbacks with the reason.
+    :ivar bytes username: For logging.
+    :ivar bytes address: For logging.
+    """
+    delimiter = b'\n'
+
+    def connectionMade(self):
+        self.transport.disconnecting = False
```
tomprince: twisted bug

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+    def connectionLost(self, reason):
+        if reason.check(ConnectionDone):
+            self.deferred.callback(None)
+        else:
+            self.deferred.errback(reason)
+
+    def lineReceived(self, line):
+        log.msg(format="%(line)s",
+                system="SSH[%s@%s]" % (self.username, self.address),
+                username=self.username, address=self.address, line=line)
```
tomprince: This should perhaps just take a eliot context to log in. Although, we perhaps want a nice display on stdout (i.e. not JSON for every output line).

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+        Run: perform_run,
+        Sudo: perform_sudo,
+        Put: perform_put,
+        Comment: perform_comment,
+    })
+
+
+def get_connection_helper(address, username):
+    """
+    Get a :class:`twisted.conch.endpoints._ISSHConnectionCreator` to connect to
```
tomprince: Link to twisted bug about making this public.

flocker/provision/_ssh/_crochet.py
```diff
@@ -0,0 +1,28 @@
...
+    perform as perform_with_twisted
+)
+from twisted.internet import reactor
+
+
+def perform(dispatcher, effect):
+    """
+    Perform an effect in a reactor with crochet.
+    """
+    setup()
```
tomprince: This branch is large enough. Changing `admin.acceptance` to be asynchronous can be in a followup.

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+        reactor, address, 22, None, username,
+        keys=keys,
+        password=None,
+        agentEndpoint=agentEndpoint,
+        knownHosts=None, ui=ConsoleUI(lambda: _ReadFile(b"yes")))
+
+
+@deferred_performer
+@inlineCallbacks
+def perform_run_remotely(base_dispatcher, intent):
```
tomprince: This needs some tests.

flocker/provision/_ssh/_crochet.py
```diff
@@ -0,0 +1,28 @@
...
+    return run_in_reactor(perform_with_twisted)(
+        dispatcher,
+        effect,
+    ).wait()
+
+
+dispatcher = ComposedDispatcher([
```
tomprince: I'm not sure where this should live.

flocker/provision/_ssh/__init__.py
```diff
@@ -0,0 +1,17 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+from ._model import (
+    Run, run, run_from_args,
+    Sudo, sudo, sudo_from_args,
+    Put, put,
+    Comment, comment,
+    RunRemotely,
```
tomprince: I avoided importing from `_conch` or `_crochet` since the package builders import this. That should probably be fixed in the dockerfiles for the builds.

admin/acceptance.py
```diff
@@ -161,16 +166,19 @@ def start_nodes(self):
         box_version = vagrant_version(self.package_source.version)
         # Boot the VMs
         check_safe_call(
-            ['vagrant', 'up'],
+            ['vagrant', 'up', '--parallel'],
```
wallrj: I was excited about this change, but when I tested it locally, it didn't seem to make much of a difference. I'm not sure if the virtualbox provider supports parallel provisioning:
 * https://www.mail-archive.com/vagrant-up@googlegroups.com/msg00662.html

If you think it should work, then this might be a change that could be split into a separate issue / branch so that we can see the results in isolation from the other changes in the branch.

admin/build_targets/centos-7/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN ["yum", "install", "--assumeyes", "git", "ruby-devel", "python-devel", "pyth
 # See https://github.com/jordansissel/fpm/issues/611
 RUN ["gem", "install", "fpm"]
 
-RUN ["pip", "install", "twisted==15.0.0", "characteristic==14.1.0", "virtualenv==12.0.5", "effect==0.1a13" , "boto==2.30.0", "requests==2.4.3", "requests-file==1.0", "ipaddr==2.1.11", "eliot==0.6.0"]
+RUN ["pip", "install", "twisted==15.0.0", "characteristic==14.1.0", "virtualenv==12.0.5", "effect==0.1a13" , "boto==2.30.0", "requests==2.4.3", "requests-file==1.0", "ipaddr==2.1.11", "eliot==0.6.0", "pyrsistent == 0.9.1"]
```
wallrj: I couldn't see why Pyrsistent is needed in the package build environment? 

TBH I can't remember why I added all these explicit Python packages to the build environment, couldn't we just create a virtualenv and install the Flocker[release] requirements and run the build-package-entrypoint script inside it?

flocker/provision/_ssh/_crochet.py
wallrj: I'm really confused by all the different `dispatcher` variables.

flocker/provision/_ssh/_conch.py
```diff
@@ -0,0 +1,171 @@
...
+                system="SSH[%s@%s]" % (self.username, self.address),
+                username=self.username, address=self.address, line=line)
+
+
+@sync_performer
+def perform_sudo(dispatcher, intent):
+    """
+    See :py:class:`Sudo`.
+    """
+    return Effect(Run(command='sudo ' + intent.command))
```
wallrj: How does this `sync_performer` interact with the `deferred_performer` that it wraps (in this dispatcher)? 

Does `sync_performer` wait for the deferred to fire before returning?

flocker/provision/_ssh/_crochet.py
wallrj: Not sure I understand this comment. You mean the initialisation and first boot up of cloud nodes?

wallrj:
Thanks @tomprince

Here's my understanding of the proposed changes:
 * Vagrant boxes are now brought up in parallel (although that may not actually work with virtualbox)
 * But for now, cloud nodes are still initialised and booted in series.
 * We use Effect, and different Effect performers for documentation building and for acceptance test provisioning.
   * Although I can't quite understand how the sphinx dispatcher works
 * We use Conch rather than Fabric to run commands on nodes
 * Which requires a Twisted reactor
 * Which needs to be started in a separate thread,
 * which is where Crochet comes in
 * We establish a master conch connection to each node
 * All existing remote node commands are now run over that conch connection
 * By way of `perform_run` function, which is a "deferred performer" which is registered in the dispatcher returned by `get_ssh_dispatcher`
 * The installation of cluster-flocker-node, its dependencies and all the cloud specific preparatory provisioning is performed in parallel on each node
 * Most of the provisioning steps on each node are performed in sequence.
 * The `docker pull` commands are now run in parallel.

Points:
 1. The design branch is already pretty large and was difficult to follow.
    Can the implementation be broken into a series of smaller changes? eg
    1. Parallel vagrant box up (if we think this is a good change)
    2. Switch from flocker.to Effect
    3. Switch from fabric to conch
    4. Parallel node provisioning
    5. Parallel docker pull
 2. I find it difficult to follow how the various dispatchers build upon one another. Maybe the variable name `dispatcher` should be replaced with a more descriptive name.
 3. If you've tested it, it'd be nice if you could share your performance results. How much faster is this in practice?

Overall it seems like a nice plan and a good use of Effect and Crochet.
Please proceed.

flocker/provision/_ssh/_conch.py
wallrj: Link to bug.

flocker/provision/_ssh/_conch.py
wallrj: File a follow up issue or break this work in to smaller branches where the Eliot logging can be added without overwhelming the reviewer.

flocker/provision/_ssh/_conch.py
wallrj: Maybe use `flocker.testtools.ssh.create_ssh_server` for testing.

flocker/provision/_ssh/_crochet.py
```diff
@@ -0,0 +1,28 @@
...
+)
+from twisted.internet import reactor
+
+
+def perform(dispatcher, effect):
+    """
+    Perform an effect in a reactor with crochet.
+    """
+    setup()
+    return run_in_reactor(perform_with_twisted)(
```
wallrj: It's not clear what's going on here. Maybe use keyword arguments and assign the returned function to a descriptive variable before calling it.

flocker/provision/_ssh/_conch.py
wallrj: That's fair enough.

wallrj:
Thought I ought to try running it, but got the following error:

```
(1279)(parallel-provision-FLOC-1279 ? =)[~/.../HybridLogic/flocker]$ time ./admin/run-acceptance-tests --keep --distribution=fedora-20 --provider=aws --config-file=aws.config --branch=0.4.0pre1 --flocker-version=0.4.0pre1 < /dev/null
2015-04-01 13:38:35+0100 [-] Log opened.
2015-04-01 13:38:35+0100 [-] Creating node 0: acceptance-test-richardw-0
/home/richard/.ssh/known_hosts updated.
Original contents retained as /home/richard/.ssh/known_hosts.old
2015-04-01 13:39:35+0100 [-] Creating node 1: acceptance-test-richardw-1
/home/richard/.ssh/known_hosts updated.
Original contents retained as /home/richard/.ssh/known_hosts.old
2015-04-01 13:40:35+0100 [-] --keep specified, not destroying nodes.
2015-04-01 13:40:35+0100 [-] Traceback (most recent call last):
2015-04-01 13:40:35+0100 [-]   File "./admin/run-acceptance-tests", line 13, in <module>
2015-04-01 13:40:35+0100 [-]     main(sys.argv[1:], top_level=TOPLEVEL, base_path=BASEPATH)
2015-04-01 13:40:35+0100 [-]   File "/home/richard/projects/HybridLogic/flocker/admin/acceptance.py", line 411, in main
2015-04-01 13:40:35+0100 [-]     nodes = runner.start_nodes()
2015-04-01 13:40:35+0100 [-]   File "/home/richard/projects/HybridLogic/flocker/admin/acceptance.py", line 247, in start_nodes
2015-04-01 13:40:35+0100 [-]     for node in self.nodes
2015-04-01 13:40:35+0100 [-]   File "/home/richard/projects/HybridLogic/flocker/flocker/provision/_libcloud.py", line 181, in provision
2015-04-01 13:40:35+0100 [-]     variants=variants,
2015-04-01 13:40:35+0100 [-]   File "/home/richard/projects/HybridLogic/flocker/flocker/provision/_aws.py", line 71, in provision_aws
2015-04-01 13:40:35+0100 [-]     ) + task_upgrade_selinux(),
2015-04-01 13:40:35+0100 [-] TypeError: unsupported operand type(s) for +: 'Effect' and 'Effect'

real	2m0.670s
user	0m40.859s
sys	0m0.654s

```

wallrj:
And I realise now why you're thinking about the logging. The output is going to be pretty confused if both nodes are being provisioned in parallel.
I wonder what happens to the `docker pull` progress output?

flocker/provision/_ssh/_crochet.py
tomprince: I'm only using crochet here because this branch was getting to big, and I didn't want to change all of `admin.acceptance` to be asynchronous in this branch.

flocker/provision/_ssh/_conch.py
tomprince: When an effect returns another effect, the returned effect is run, and its result is used as the result of the original effect. (Much like deferreds). So, `perform_sudo` synchronously returns another effect (which is itself asynchronous).

adamtheturtle:
http://build.clusterhq.com/builders/flocker-vagrant-dev-box is failing, most recently with the following:

```
Vagrant can't use the requested machine because it is locked! This
means that another Vagrant process is currently reading or modifying
the machine. Please wait for that Vagrant process to end and try
again. Details about the machine are shown below:

Name: default
Provider: virtualbox
Failed 1: vagrant up
```

If you think that this is anything to do with this branch then fix that.

PR #1256, opened 2015-03-30
[FLOC-1599] API endpoint for changing the maximum_size of a dataset

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1599

This adds a new configure_datasets_primary input schema which requires a primary address to be supplied. This is for use by the `create_dataset` endpoint. While the `update_dataset` endpoint validates, but does not require any of the dataset attributes.

Acceptance tests can be run using the included temporary shell script to build packages and install them on two nodes before performing the test.

That script will be deleted before merging this branch.


flocker/acceptance/test_api.py
```diff
@@ -432,3 +432,60 @@ def not_exists():
...
+            # Check that the configuration response has the expected size.
+            self.assertEqual(new_size, dataset.maximum_size)
+            # Wait for the dataset to have the expected size.
+            return cluster.wait_for_dataset(dataset)
+
+        checking = resizing.addCallback(check_dataset_size)
+
+        return checking
+
+    def test_dataset_shrink(self):
```
itamarst: These tests aren't implemented? You could punt shrinking to another issue if that makes your life easier.

flocker/control/httpapi.py
```diff
@@ -351,7 +319,7 @@ def saved(ignored):
                       '/v1/endpoints.json#/definitions/configuration_dataset'},
         schema_store=SCHEMAS
     )
-    def update_dataset(self, dataset_id, primary=None):
+    def update_dataset(self, dataset_id, primary=None, maximum_size=None):
```
itamarst: This is not good. Our model currently uses `maximum_size=None` to indicate "no particular size set". This is not relevant to block storage backends, but supported by ZFS. And the only reasonable way to expose the JSON API for that is likely `{"maximum_size": null}`. I guess technically one could do `DELETE /v1/configuration/datasets/21321412/maximum_size` but that seems a little weird.

Assuming we don't go with `DELETE`, we now have two meanings for `None`: parameter not passed, and "remove maximum size requirement."

Instead you should have a `_NO_ARGUMENT = object()` on module level and then default value of `maximum_size` can be that, allowing you to distinguish the two cases.

You should then implement support for setting `maximum_size` to None.

itamarst:
A good start, but missing some required functionality:

1. Clearing `maximum_size`. You could argue that maximum size should always be specified but it's kinda nice that ZFS doesn't require it?
2. Acceptance tests for shrinking. Could be moved to separate issue, but then again probably pretty easy to write since ZFS support is already built-in.

flocker/acceptance/test_api.py
wallrj: I got the shrink test working, but I think I'll postpone the "shrink not possible" test for a follow up issue. Perhaps https://clusterhq.atlassian.net/browse/FLOC-1503

wallrj:
Thanks @itamarst 

I think I addressed all your review comments.

 * Added support for removing the configured size limit
 * Added acceptance test for that case.
 * Removed proposed acceptance test for attempting to shrink a dataset smaller than its current data.  

Ready for another review.

docs/advanced/api_examples.yml
```diff
@@ -209,6 +209,30 @@
 
 -
   id:
+    "update dataset size"
+
+  doc: |
+    Update a dataset with a new maximum_size. This will grow or shrink
+    the dataset.
```
exarkun: Should this also say something about filesystems?  I'm not sure.  We aren't super precise in our use of "volumes" vs "datasets" and we hardly ever talk about "filesystems".  Yet I'm pretty I would, as a user of this API, want to know what the state of my filesystem is after a resize operation like this.

flocker/acceptance/test_api.py
```diff
@@ -462,12 +462,13 @@ def in_current():
...
 
-def create_dataset(test_case):
+def create_dataset(test_case, maximum_size=REALISTIC_BLOCKDEVICE_SIZE):
     """
     Create a dataset on a single-node cluster.
 
     :param TestCase test_case: The test the API is running on.
-
+    :param int maximum_size: The size of the dataset to create on the test
+        cluster. Defaults to ``REALISTIC_BLOCKDEVICE_SIZE``.
```
exarkun: Any sensible API documentation generation tool should provide information about the default automatically.

flocker/acceptance/test_api.py
```diff
@@ -581,3 +582,107 @@ def not_exists():
...
+
+    def test_dataset_shrink_not_valid(self):
+        """
+        If the requested maximum_size is smaller than the allowed minimum the
+        response is ``BAD_REQUEST``.
+        """
+        creating = create_dataset(
+            test_case=self, maximum_size=REALISTIC_BLOCKDEVICE_SIZE
+        )
+        new_size = 67108864-1
```
exarkun: space around the operator

flocker/control/httpapi.py
```diff
@@ -172,9 +176,10 @@ def create_dataset_configuration(self, primary, dataset_id=None,
             This is not for easy human use.  For human-friendly identifiers,
             use items in ``metadata``.
 
-        :param int maximum_size: The maximum number of bytes the dataset will
-            be capable of storing.  This may be optional or required depending
-            on the dataset backend.
+        :param maximum_size: Either, the maximum number of bytes the dataset
```
exarkun: No comma after `Either`

flocker/control/httpapi.py
```diff
@@ -172,9 +176,10 @@ def create_dataset_configuration(self, primary, dataset_id=None,
...
+            will be capable of storing or ``null`` make the dataset size
```
exarkun: `... to make ...` I think

exarkun: And `None`, not `null`, because we're in Python land now.

flocker/control/httpapi.py
```diff
@@ -810,6 +787,92 @@ def replace_configuration(self, applications, deployment):
...
+
+    :param Deployment deployment: The deployment containing the dataset to be
+        updated.
+    :param unicode dataset_id: The ID of the dataset to be updated.
+    :param maximum_size: The new size of the dataset or ``None`` to remove the
+        size limit.
+    :returns: An updated ``Deployment``.
+    """
+    manifestation, node = _find_manifestation_and_node(deployment, dataset_id)
+    deployment = deployment.transform(['nodes'], lambda s: s.remove(node))
```
exarkun: Maybe `deployment.set(nodes=deployment.nodes.discard(node))`

flocker/control/httpapi.py
```diff
@@ -810,6 +787,92 @@ def replace_configuration(self, applications, deployment):
...
+    node = node.transform(
+        ['manifestations', dataset_id, 'dataset', 'maximum_size'],
+        maximum_size
+    )
+    return deployment.transform(['nodes'], lambda s: s.add(node))
```
exarkun: Likewise here

flocker/control/schema/endpoints.yml
```diff
@@ -124,12 +124,14 @@ definitions:
...
-      # Temporarily required until volume backends settle down and we know
-      # more about what it means to not have a primary manifestation.
-      - primary
     additionalProperties: false
 
+  configuration_dataset_primary:
+    description: "The configuration for a new dataset with a primary address"
+    allOf:
+      - "$ref": "#/definitions/configuration_dataset"
+      - "required": ["primary"]
```
exarkun: Please use the `- ...` list syntax everywhere rather than mixing multiple list syntaxes.

flocker/control/test/test_httpapi.py
```diff
@@ -2171,42 +2234,143 @@ def saved(ignored):
...
+        """
+        unknown_dataset_id = unicode(uuid4())
+        return self.assertResult(
+            method=b"POST",
+            path=b"/configuration/datasets/%s" % (
+                unknown_dataset_id.encode('ascii'),),
+            request_body={},
+            expected_code=NOT_FOUND,
+            expected_result={u"description": u'Dataset not found.'}
+        )
```
exarkun: This seems a lot like `UpdatePrimaryDatasetTestsMixin.test_unknown_dataset`.  And there doesn't seem to be anything actually related to size updates in this test.

flocker/control/test/test_httpapi.py
```diff
@@ -2171,42 +2234,143 @@ def saved(ignored):
...
+            new_size=REALISTIC_BLOCKDEVICE_SIZE
+        )
+
+    def test_too_small(self):
+        """
+        A dataset must be at least 67108864 bytes.
+        """
+        return self.assert_dataset_resize(
+            original_size=67108864,
+            new_size=67108864-1,
```
exarkun: space around the operator

flocker/control/schema/endpoints.yml
```diff
@@ -124,12 +124,14 @@ definitions:
...
-    required:
-      # Temporarily required until volume backends settle down and we know
-      # more about what it means to not have a primary manifestation.
-      - primary
     additionalProperties: false
 
+  configuration_dataset_primary:
+    description: "The configuration for a new dataset with a primary address"
+    allOf:
+      - "$ref": "#/definitions/configuration_dataset"
```
exarkun: `configuration_dataset` is starting to look more like a "type" than an "endpoint" (if that means anything).  Maybe it should move to the other schema file.  And I think the funny word order, "configuration_dataset" instead of "dataset_configuration" was meant as a reflection of which endpoint this related to - the `/configuration/dataset` endpoint.  So perhaps the schema in the type file should be "dataset_configuration" and the schema that remains here, `configuration_dataset_primary`, should just be called `configuration_dataset` - since it is, after all, the schema that `/configuration/dataset` now uses.

flocker/control/test/test_schemas.py
```diff
@@ -475,97 +475,114 @@
...
-        # All of them can be combined.
-        {u"primary": u"10.0.0.1",
-         u"metadata":
-             dict.fromkeys((unicode(i) for i in range(16)), u"x" * 256),
-         u"maximum_size": 1024 * 1024 * 64,
-         u"dataset_id": u"x" * 36,
-         u"deleted": True},
-    ]
+ConfigurationDatasetsNewSchemaTests = build_schema_test(
+    name="ConfigurationDatasetsPrimarySchemaTests",
```
exarkun: Switch the name here to something better (without "New") and make the identifier and the `name` agree.

run-api-acceptance-tests
```diff
@@ -0,0 +1,134 @@
+#!/bin/sh
```
exarkun: And please don't add this to master. :smile:

exarkun:
Thanks.  Please address inline comments, get buildbot green, and merge.

wallrj:
Thanks for the review @exarkun 

I've created a new dataset_configuration type as suggested.
But I'm not sure if I've done it correctly (the tests pass...)
And the way I've done it seems to have confused our json schema Sphinx plugin.

So perhaps I should revert those changes and file a follow up issue.

Let me know what you think.

wallrj:
I added back the `type: object` element to each of the endpoint schemas, but now the sphinx plugin complains because it expects to find a `properties` list too

```
  File "/home/richard/projects/HybridLogic/flocker/flocker/restapi/docs/publicapi.py", line 120, in fill_in_result
    for property, propSchema in object_schema[u'properties'].iteritems():
KeyError: u'properties'
```

Is there a way to define a `properties` list as a reference to the new dataset_configuration type 
and will the sphinx plugin need updating anyway...can it follow `allOf` style references?

wallrj:
Ok. I found a compromises / hacks and created some follow up issues to properly support schema inheritance.
If buildbot goes green, I'll merge this and move on.

wallrj:
Before merging forward, the API tests ran successfully, locally....

```
flocker.acceptance.test_api
  DatasetAPITests
    test_dataset_grow ...                                                  [OK]
    test_dataset_move ...                                                  [OK]
    test_dataset_remove_size_limit ...                                     [OK]
    test_dataset_shrink ...                                                [OK]
    test_dataset_shrink_not_valid ...                                      [OK]

Ran 7 tests in 33.246s

PASSED (successes=7)
```

And the still pass after merging forward, so I think I'll merge this now.

```
locker.acceptance.test_api
  DatasetAPITests
    test_dataset_grow ...                                                  [OK]
    test_dataset_move ...                                                  [OK]
    test_dataset_remove_size_limit ...                                     [OK]
    test_dataset_shrink ...                                                [OK]
    test_dataset_shrink_not_valid ...                                      [OK]

Ran 7 tests in 33.187s

PASSED (successes=7)

```

PR #1257, opened 2015-03-30
Brew tap floc 1611

adamtheturtle:
Fixes FLOC-1611

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -364,6 +364,8 @@ Release
 
         export VERSION=0.1.2
 
+     Install and test the Homebrew recipe:
+
      .. task:: test_homebrew flocker-${VERSION}
         :prompt: [osx-user]$
 
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -365,7 +365,7 @@ Release
         export VERSION=0.1.2
 
      .. task:: test_homebrew flocker-${VERSION}
-           :prompt: [osx-user]$
+        :prompt: [osx-user]$
 
      If tests fail then the either the recipe on the `master` branch or the package it installs must be modified.
      The release process should not continue until the tests pass.
--- a/docs/indepth/installation.rst
+++ b/docs/indepth/installation.rst
@@ -84,17 +84,10 @@ Make sure Homebrew has no issues:
 
 Fix anything which ``brew doctor`` recommends that you fix by following the instructions it outputs.
 
-Add the ``ClusterHQ/flocker`` tap to Homebrew and install ``flocker``:
+Add the ``ClusterHQ/tap`` tap to Homebrew and install ``flocker``:
 
-.. version-code-block:: console
-
-   alice@mercury:~$ brew tap ClusterHQ/tap
-   ...
-   alice@mercury:~$ brew install flocker-|latest-installable|
-   ...
-   alice@mercury:~$ brew test flocker-|latest-installable|
-   ...
-   alice@mercury:~$
+.. task:: test_homebrew flocker-|latest-installable|
+   :prompt: alice@mercury:~$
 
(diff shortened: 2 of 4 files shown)


docs/indepth/installation.rst
```diff
@@ -84,17 +84,10 @@ Make sure Homebrew has no issues:
...
-
-   alice@mercury:~$ brew tap ClusterHQ/tap
-   ...
-   alice@mercury:~$ brew install flocker-|latest-installable|
-   ...
-   alice@mercury:~$ brew test flocker-|latest-installable|
-   ...
-   alice@mercury:~$
+.. task:: test_homebrew flocker-|latest-installable|
+   :prompt: alice@mercury:~$
```
jongiddy: Not clear to me why the `flocker-|latest-installable|` argument is provided?  It does not appear to be used in the `task_test_homebrew`?

flocker/provision/_sphinx.py
```diff
@@ -66,7 +71,11 @@ def run(self):
         task = getattr(tasks, 'task_%s' % (self.arguments[0],))
         prompt = self.options.get('prompt', '$')
         if len(self.arguments) > 1:
-            task_arguments = self.arguments[1].split()
+            # Some tasks can include the latest installable version as (part
+            # of) an argument. This replaces a placeholder with that version.
+            latest = get_installable_version(version)
+            task_arguments = [item.replace(PLACEHOLDER, latest) for
+                              item in self.arguments[1].split()]
```
jongiddy: This repeats the code in `flocker.docs.version_extensions.VersionCodeBlock`. Maybe make this a separate function.  But again, not seeing why this is necessary?

docs/indepth/installation.rst
adamtheturtle: `flocker-|latest-installable|` is the argument for the `recipe_url` parameter. Perhaps that parameter name should be improved and a docstring added there. This let's the task render as:

```
brew install flocker-0.4.0
```

for example.

flocker/provision/_sphinx.py
adamtheturtle: Explained why it is necessary in a comment above. I will try making it a separate function and then judge if it is nicer that way.

jongiddy: Can you please point to an example where this is done?  Why does the homebrew task get passed the argument, but does not use it?

adamtheturtle: The homebrew task does use the argument:

```python
def task_test_homebrew(recipe_url):
    return [
        Run(command="brew update"),
        Run(command="brew install {url}".format(url=recipe_url)),
        Run(command="brew test {url}".format(url=recipe_url)),
    ]
```

It uses the argument to install and test the recipe - the argument is named `recipe_url` but can really just be the name of the recipe if that repository is tapped. The argument is used in `commands = task_test_homebrew(recipe_url)` in `admin.testbrew.main`. The argument is also used in the new code:

`.. task:: test_homebrew flocker-|latest-installable|`

That task will render the commands from  `task_test_homebrew` as if the `recipe_url` were e.g. `flocker-0.3.3dev8` if 0.3.3dev8 is the latest installable package.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -365,7 +365,7 @@ Release
         export VERSION=0.1.2
```
jongiddy: This creates two sections of code (the export and the task), with no intervening text. It looks strange.

jongiddy: Is it possible to use the latest-installable argument here as well, to allow removal of the export?

adamtheturtle: I will put some intervening text in.

It is not possible - `|latest-installable|` on `master` (which doc-dev is built from, doc-dev is used for the release) uses the latest tag. When the docs are built the tag will be that of the previous release.

jongiddy:
Tidy up the section in the release process documentation to remove the consecutive code extracts, and this is OK to merge (now that I understand what is happening with `latest-installable`

PR #1258, opened 2015-03-30
Switch to the correct directory before publishing docs

adamtheturtle:
Fixes FLOC-1614

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -374,6 +374,7 @@ Release
 
    .. prompt:: bash [vagrant@localhost]$
 
+      cd ~/flocker-${VERSION}
       admin/publish-docs --production
 
 #. Submit the release pull request for review again.


jongiddy:
:question:  If we are doing a documentation release, we would already be in the correct directory, and this command is not needed.  Would it be better to add it as the last command of the Homebrew creation (after the `git push`)?  If so, then the command can be `cd ../flocker-${VERSION}` which is consistent with other `cd` commands - e.g. this makes it easier if the user decides to do all these steps in a sub-directory on the Vagrant machine for some reason.

Otherwise, seems fine, although I haven't run the whole release process to test. OK to merge.

adamtheturtle:
> e.g. this makes it easier if the user decides to do all these steps in a sub-directory on the Vagrant machine for some reason.

Unsupported configuration. Also the `cd` in the hombrew step is not consistent with this.

> If so, then the command can be cd ../flocker-${VERSION} which is consistent with other cd commands

As this comment is marked optional I don't think that this would be better (enough) so I will merge this now.

PR #1259, opened 2015-03-30
Implementation sketch for collecting/tracking/using cluster-wide non-manifest dataset information

exarkun:
Design for https://clusterhq.atlassian.net/browse/FLOC-1513

The general idea here is to add a new attribute to `DeploymentState` describing all datasets that are known to exist but are not manifest on any node.  The node->control service protocol is adjusted to allow nodes to tell the control service about this information.  The control service->node protocol is implicitly changed merely by having a new attribute on `DeploymentState`.

Thus, nodes collect the information and report it to the control service and the control service sends it back out to nodes to be used.  Initially all (IaaS) nodes collect and report this information.  Later we can fix things so fewer of them do (perhaps just one) to avoid the redundant work.

### DIFF
--- a/flocker/control/_clusterstate.py
+++ b/flocker/control/_clusterstate.py
@@ -35,6 +35,12 @@ class ClusterStateService(Service):
         """
         self._deployment_state = self._deployment_state.update_node(node_state)
 
+    # FLOC-1513
+    #
+    # Add a new method, replace_nonmanifest_datasets.
+    # It replaces _deployment_state with a new instance that has a new (not
+    # updated) set of dataset_ids.
+
     def manifestation_path(self, hostname, dataset_id):
         """
         Get the filesystem path of a manifestation on a particular node.
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -507,6 +507,14 @@ class NodeState(PRecord):
     manifestations = pmap_field(unicode, Manifestation, optional=True)
     paths = pmap_field(unicode, FilePath, optional=True)
 
+    # FLOC-1513
+    #
+    # Add an update_cluster_state method that accepts a ``DeploymentState``
+    # instance and just calls ``DeploymentState.update_node(self)``.
+    #
+    # Double-dispatch is in support of letting VisibleClusterState get involved
+    # in the update process too.
+
 
 class DeploymentState(PRecord):
     """
@@ -516,6 +524,28 @@ class DeploymentState(PRecord):
         the state of each cooperating node.
     """
     nodes = pset_field(NodeState)
+
+    # FLOC-1513
+    #
+    # A new mapping from dataset ids to dataset instances.  This represents
+    # information about all datasets that are known to exist but have no
+    # manifestations.  There cannot be any such datasets against a P2P backend.
+    # IaaS agents can discover these datasets by finding unattached volumes.
+    #
+    # This doesn't convey backend-specific information.  Backends are expected
+    # to be able to map a dataset id back onto whatever the backend specific
+    # storage is.  Membership in this map indicates *something* exists.
+    #
+    # The Dataset instances in this pmap are "state" Datasets.  They have a
+    # dataset_id and all the rest of their fields are basically meaningless
+    # (XXX maybe maximum_size should be populated eventually?)  (Huh we should
+    # have a separate type for dataset state vs config).
+    #
+    # nonmanifest_datasets = pmap_field(UUID, Dataset)
+
+    # FLOC-1513
+    #
+    # Add an __invariant__ so keys and values match up.
 
     def update_node(self, node_state):
         """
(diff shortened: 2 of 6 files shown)


flocker/control/_model.py
```diff
@@ -517,6 +525,28 @@ class DeploymentState(PRecord):
...
+    # The Dataset instances in this pmap are "state" Datasets.  They have a
+    # dataset_id and all the rest of their fields are basically meaningless
+    # (XXX maybe maximum_size should be populated eventually?)  (Huh we should
+    # have a separate type for dataset state vs config).
+    #
+    # nonmanifest_datasets = pmap_field(UUID, Dataset)
+
+    # FLOC-1513
+    #
+    # Add an __invariant__ so keys and values match up.
```
itamarst: Utility function? since we already have one of those that's almost the same
.

flocker/control/_protocol.py
```diff
@@ -109,6 +109,11 @@ class NodeStateCommand(Command):
     Used by a convergence agent to update the control service about the
     status of a particular node.
     """
+    # FLOC-1513
+    #
+    # Add another argument, an optional `ListOf(Unicode)` representing
```
itamarst: I propose replacing this with `ListOf(IClusterStateUpdater)`.

flocker/control/_protocol.py
```diff
@@ -134,6 +139,13 @@ def logger(self):
...
 
+    # FLOC-1513
+    #
+    # Accept dataset_ids collection here too.  Pass it on.  The default is
+    # None, distinct from an empty collection.  If the node doesn't pass us the
+    # info, it doesn't know anything about nonmanifest datasets.  If it doesn't
+    # know anything about them, we shouldn't update our local state based on
+    # the (non-)information.
     @NodeStateCommand.responder
     def node_changed(self, eliot_context, node_state):
```
itamarst: Again, replace with list of `ICLusterStateUpdater`.

flocker/control/_clusterstate.py
```diff
@@ -35,6 +35,12 @@ def update_node_state(self, node_state):
         """
         self._deployment_state = self._deployment_state.update_node(node_state)
 
+    # FLOC-1513
```
itamarst: Given list of `IClusterSTateUpdater` instances don't need new method.

flocker/node/agents/blockdevice.py
```diff
@@ -779,6 +779,20 @@ def _manifestation_from_volume(volume):
...
+#
+# It also has a method, update_cluster_state, which both updates the cluster
+# with its ``NodeState`` (just delegate to that object's update_cluster_state
+# method) and with its own ``nonmanifest_datasets`` (use
+# ``DeploymentState.update_nonmanifest_datasets``).
+#
+# Probably introduce an interface for this thing now.  It's what must be
+# returned by ``discover_local_state`` and it must have this
+# ``update_cluster_state`` method.
+
```
itamarst: Replace with `NonManifestDatasets` class which implements `IClusterStateUpdater`. (which may as well reside in `flocker.control._model`) Don't forget to add it to `_model.SERIALIZABLE_CLASSES`.

itamarst:
I suggested an improved scheme:

1. Create new IClusterStateUpdater interface that has a update_cluster_state method, same as the one you're proposing.
2. NodeState is now a IClusterStateUpdater implementer.
3. IDeployer.discover_local_state (maybe renamed?) returns a *set* of IClusterStateUpdater instances.
4. VisibleClusterState can then be replaced with a NonManifestDatasets implementer of IClusterStateUpdater, and the block storage deployer returns one of those as well as a NodeState.

flocker/control/_protocol.py
exarkun: It's AMP.  It can't readily transport arbitrary implementations of an interface.  It could use the `wire_encode` and `wire_decode` to transport one (the only one we care about) specific implementation... Or I could just make it `ListOf(Dataset)`.  `Unicode` is simpler on the wire, though, and there are no plans to use anything from the `Dataset` except the `dataset_id` for now...

itamarst: The point is that `discover_local_state` returns an arbitrary list of `IClusterSTateUpdater` providers, and that's what you need to send over the wire. You shouldn't special case datasets.

exarkun: You can't not special-case them, though.  We have to implement a wire protocol.  We can't (and shouldn't) serialize arbitrary objects.

itamarst: We serialize any object as long as it's on the approved list ... and an instance of the class passed to `SerializableArgument`. And it's not arbitrary objects, it's "IClusterStateUpdater instances that are on the approved list of serializable objects".

So basically you have a variant of `SerializableArgument` which takes a list of objects and checks their interface, and then relies on the approved serializable list after that.

PR #1261, opened 2015-03-30
Fix the mount state misrepresentation perpetuated by BlockDeviceDeployer.discover_state

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1617

The general idea here is to make `BlockDeviceDeployer.discover_local_state` stop claiming it has local manifestations for certain datasets that aren't quite manifest there yet.  Instead of finding attached volumes and claiming a manifestation for each one that matches a dataset configured to be on the node, it will only claim those which also appear to have a filesystem mounted in the right place.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1609,6 +1609,9 @@ def test_run_mkfs_and_mount(self):
         )
 
 
+# FLOC-1617
+#
+# Move this to ``blockdevice.py``.  Write integration tests for it.
```
itamarst: https://pythonhosted.org/psutil/#psutil.disk_partitions

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -122,10 +158,79 @@ def test_no_devices(self):
...
+        and not as a manifestation on the ``NodeState``.
+        """
+        unmounted = self.api.create_volume(
+            dataset_id=uuid4(),
+            size=REALISTIC_BLOCKDEVICE_SIZE,
+        )
+        self.api.attach_volume(
+            unmounted.blockdevice_id, self.expected_hostname
+        )
+        self.assertDiscoveredState(self.deployer, [], [unmounted.dataset_id])
```
wallrj: The signature of the `assertDiscoveredState` isn't obvious. This diff would have been easier to read if we used keyword arguments for `expected_manifestations` and `expected_nonmanifest_datasets` here and in later tests.

flocker/node/agents/blockdevice.py
```diff
@@ -815,14 +846,32 @@ def discover_state(self, node_state):
...
             dataset_id = manifestation.dataset.dataset_id
-            # TODO This assumes things are actually mounted.  Maybe they
-            # aren't.  FLOC-1498 and FLOC-1499 need correct information here.
-            # Probably other folks won't mind it either.
             mountpath = self._mountpath_for_manifestation(manifestation)
-            paths[dataset_id] = mountpath
+
+            # If the expected mount point doesn't actually have a (XXX should
+            # be "the") device mounted where we expected to find this
+            # manifestation, the manifestation doesn't really exist here.
```
wallrj: Elaborate on or remove the XXX part of this comment.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -49,6 +53,37 @@
...
+    :param FilePath device: The path to the file onto which to put the
+        filesystem.  Anything accepted by ``mkfs`` is acceptable (including a
+        regular file instead of a device file).
+    """
+    options = []
+    if not device.isBlockDevice():
+        options.extend([
+            # Make the filesystem even if the destination is not a block
+            # device.
+            b"-F",
```
wallrj: It'd be clearer if we used `--long-options` here and elsewhere in this branch.

wallrj: Actually it looks like there isn't a long option :frowning: but the man page also says

```
       -F     Force mke2fs to create a filesystem, even if the specified device is not a partition on a block special device, or if other  parame‐
              ters  do  not make sense.  In order to force mke2fs to create a filesystem even if the filesystem appears to be in use or is mounted
              (a truly dangerous thing to do), this option must be specified twice.
```

...which makes me wonder why this isn't also needed when we attempt to format an unpartitioned loopback device. Perhaps if you attempted to do this on a "disk like" block device it'd check for a master boot record or something?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -49,6 +53,37 @@
...
+        ])
+    command = [b"mkfs"] + options + [b"-t", b"ext4", device.path]
+    run_process(command)
```
wallrj: `run_process` returns a `ProcessResult` . These new helper functions might be more useful if we return that here.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -49,6 +53,37 @@
...
+def mount(device, mountpoint):
+    """
+    Synchronously mount a filesystem.
+
+    :param FilePath device: The path to the device file containing the
+        filesystem.
+    :param mountpoint device: The path to an existing directory at which to
+        mount the filesystem.
+    """
+    run_process([b"mount", device.path, mountpoint.path])
```
wallrj: And here.

wallrj:
Thanks @exarkun 

* This looks good to me.
* There's one familiar test failure which can be ignored.

Please  address or answer the comments above and then merge.

flocker/node/agents/test/test_blockdevice.py
exarkun: We ended up going via `/dev/loopN` everywhere else, I think.  That seems to be sufficient to avoid the need for `-F`.  I didn't want to do the extra work of setting up and tearing down loopback devices for the new tests I wrote where I just need some other random filesystem mounted in place of the correct volume.

Thinking about this now, maybe I should make this a flag instead of something automatically detected.  Then the new tests that definitely need it can pass the flag and the old tests won't have any chance of accidentally getting this behavior.

flocker/node/agents/test/test_blockdevice.py
exarkun: Could be.  I'll leave this for now - it'll be easy for someone to insert the necessary `return` later if they find a use.

wallrj:
Merged forward and ran the Dataset API acceptance tests against this branch. 

add and delete both passed and move fails (because moving is not yet implemented in the loop agent)

```
flocker.acceptance.test_api
  DatasetAPITests
    test_dataset_move ...                                               [ERROR]
                                              [ERROR]

[ERROR]
Failure: twisted.internet.defer.TimeoutError: <flocker.acceptance.test_api.DatasetAPITests testMethod=test_dataset_move> (test_dataset_move) still running at 120.0 secs

flocker.acceptance.test_api.DatasetAPITests.test_dataset_move
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
DelayedCalls: (set twisted.internet.base.DelayedCall.debug = True to debug)
<DelayedCall 0x4ee5b48 [0.0436689853668s] called=0 cancelled=0 Deferred.callback(None)>

flocker.acceptance.test_api.DatasetAPITests.test_dataset_move
Ran 3 tests in 126.995s

FAILED (errors=2, successes=2)

```

I'll merge this when the builds finish.

wallrj:
Before I merge this I thought it'd be worth merging in the update maximum_size endpoint changes and checking that the create and delete acceptance tests still pass.

```
flocker.acceptance.test_api
  DatasetAPITests
    test_dataset_grow ...                                               [ERROR]
                                              [ERROR]
    test_dataset_move ...                                               [ERROR]
                                              [ERROR]
    test_dataset_remove_size_limit ...                                  [ERROR]
                                 [ERROR]
    test_dataset_shrink ...                                             [ERROR]
                                            [ERROR]
    test_dataset_shrink_not_valid ...                                   [ERROR]
                                  [ERROR]

```

They do and the others are expected to fail.

PR #1263, opened 2015-03-31
Replace Homebrew recipe creation with download from Buildbot

jongiddy:


adamtheturtle:
This seems like a duplicate of FLOC-1150 so close that if you think it can be closed with this.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -350,7 +350,7 @@ Release
         cd
         git clone git@github.com:ClusterHQ/homebrew-tap.git "homebrew-tap-${VERSION}"
         cd homebrew-tap-${VERSION}
-        ../flocker-${VERSION}/admin/make-homebrew-recipe > flocker-${VERSION}.rb
+        wget --output-document=flocker-${VERSION}.rb http://build.clusterhq.com/results/homebrew/${VERSION}/${VERSION}/Flocker${VERSION}.rb
```
adamtheturtle: Buildbot is testing a recipe called `Flocker${VERSION}.rb` and then this is adding it as `flocker-${VERSION}.rb` to the tap.

With Homebrew the filenames are important, they have to match in some way a class name. It just so happens that `Flocker${VERSION}.rb` and `flocker-${VERSION}.rb` have the same expected class name. This would be better if the Buildbot side created and tested `flocker-${VERSION}.rb`.

That would mean that most of this section could be removed - the release engineer does not have to manually test on the Mac Mini.

adamtheturtle:
Thanks @jongiddy , address the comments and re-submit.

tomprince:
We've addressed this another way.

PR #1266, opened 2015-04-01
[FLOC-1636] Certificate authority: sketch of implementation.

itamarst:
Design for https://clusterhq.atlassian.net/browse/FLOC-1636

### DIFF
--- a/flocker/ca/__init__.py
+++ b/flocker/ca/__init__.py
@@ -1,0 +1,5 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+A minimal certificate authority.
+"""
--- a/flocker/ca/_ca.py
+++ b/flocker/ca/_ca.py
@@ -1,0 +1,53 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Low-level logic for a certificate authority.
+
+Uses ECDSA + SHA256? Or if that turns out to be too difficult RSA 4096-bit
++ SHA 256.
+"""
+
+from twisted.internet.ssl import KeyPair
+
+
+def generate_keypair():
+    """
+    Create a new key pair.
+    """
+    #return KeyPair.generate(crypto.TYPE_ECDSA, size=something?)
+    pass
+
+
+class CertificateAuthority(object):
+    """
+    A certificate authority whose configuration is stored in a specified
+    directory.
+    """
+    def __init__(self, path):
+        """
+        :param FilePath path: Directory where private key and certificate are
+            stored.
+        """
+        # try to load private key and certificate, raise exception if that
+        # fails.
+
+    @classmethod
+    def initialize(cls, path, name):
+        """
+        Generate new private/public key pair and self-sign, then store in
+        given directory.
+
+        :param FilePath path: Directory where private key and certificate are
+            stored.
+        :param bytes name: The name of the cluster.
+
+        :return CertificateAuthority: Initialized certificate authority.
+        """
+        # XXX do we want to do something else/more with the name?
+        # More metadata?
+        #dn = DistinguishedName(commonName=name)
+        #keypair = generate_keypair()
+        #request = keypair.requestObject(dn)
+        #certificate = keypair.signRequestObject(dn, request, generateaserial, digestAlgorithm='sha256')
+        #now write out to some files
+        # return cls(path)
--- a/flocker/ca/_script.py
+++ b/flocker/ca/_script.py
@@ -1,0 +1,75 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+The command-line certificate authority tool.
+"""
+
+from twisted.python.usage import Options
+
+
+class InitializeOptions(Options):
+    """
+    Command line for ``flocker-ca initialize``.
+    """
+
+    longdesc = """Create a new certificate authority.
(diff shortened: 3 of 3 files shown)


flocker/ca/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+A minimal certificate authority.
+"""
```
exarkun: Would anything be easier if this were part of the `flocker.cli` package?

Does that even make sense?  Which OS package will this be part of?  `clusterhq-flocker-cli`?

Would this be any better as `flocker ca` instead of `flocker-ca`?

flocker/ca/_ca.py
```diff
@@ -0,0 +1,53 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Low-level logic for a certificate authority.
+
+Uses ECDSA + SHA256? Or if that turns out to be too difficult RSA 4096-bit
++ SHA 256.
```
exarkun: Ooooohhh boy.  I remember this game.  Wait wait is the answer     

```
ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:
DH+AES:ECDH+3DES:DH+3DES:RSA+AESGCM:RSA+AES:RSA+3DES:!aNULL:!MD5:!DSS
```

Oops no that's the cheat code for the sequel.

SHA-256 still seems to be a good thing.  Definitely not SHA-1 (which you didn't suggest, just sayin').  Unfortunately I think you'll find the APIs necessary for ECDSA missing from pyOpenSSL.  There are probably other Python libraries for doing it, though (whether any are any good not, I dunno; I wonder if cryptography offers a good API for this yet).

flocker/ca/_ca.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @classmethod
+    def initialize(cls, path, name):
+        """
+        Generate new private/public key pair and self-sign, then store in
+        given directory.
+
+        :param FilePath path: Directory where private key and certificate are
+            stored.
+        :param bytes name: The name of the cluster.
```
exarkun: If this is to identify the certificate subject, probably better to accept a `DistinguishedName` here (or a `dict` to turn in to one, at least).  The purpose of a single bare "name" string might be easily misunderstood.

Do we even want a `commonName`?  The CA itself may not be an addressable entity in this module (it's only used by our command line tool).  Organization name seems relevant.  Perhaps also organizational unit.  Maybe some location stuff but maybe we don't actually need to be that featureful.

Other considerations - certain extensions would be relevant for this certificate, for example the CA:true constraint (random example code, <http://stackoverflow.com/questions/14972345/creating-self-signed-certificate-using-pyopenssl>).  This is possibly worth a follow-up ticket, maybe not relevant for now though (I'm **somewhat** sure forgetting about these won't cause validation issue...)

flocker/ca/_ca.py
```diff
@@ -0,0 +1,53 @@
...
+
+        :return CertificateAuthority: Initialized certificate authority.
+        """
+        # XXX do we want to do something else/more with the name?
+        # More metadata?
+        #dn = DistinguishedName(commonName=name)
+        #keypair = generate_keypair()
+        #request = keypair.requestObject(dn)
+        #certificate = keypair.signRequestObject(dn, request, generateaserial, digestAlgorithm='sha256')
```
exarkun: We should think about expiry date.  The default for this API is one year.  Having your CA certificate expire is a huge operational drag (most CA certificates have validities measured at least in decades... which may not be a good thing, but something to think about).

flocker/ca/_ca.py
exarkun: (All of this stuff matters a ton for a high-quality CA management tool.  We might not be building that so it might not actually matter at all.  But code often tends to live longer than one plans...)

flocker/ca/_ca.py
```diff
@@ -0,0 +1,53 @@
...
+        #now write out to some files
```
exarkun: Need to be careful writing it out: open the files with the right umask for safety.

flocker/ca/_script.py
```diff
@@ -0,0 +1,75 @@
...
+    def main(self, reactor, options):
+        if options.subCommand is not None:
+            return maybeDeferred(options.subOptions.run)
+        else:
+            return self.opt_help()
+
+
+def flocker_ca_main():
+    return FlockerScriptRunner(
+        CAScript(), CAOptions(), logging=False).main()
```
exarkun: Any particular reason not to want logging for this tool?  Or just expedience?

exarkun:
Thanks.  This needs user-facing documentation, too.  Either include that here or add another sub-task to FLOC-1634.  The tool needs to end up in our packages somehow as well.  Either make sure that's dealt with here or add another sub-task for that as well.

Some other comments inline, just stuff to think about.  Please proceed.

flocker/ca/_ca.py
tomprince: Cryptography has some support for x509 now.

flocker/ca/_script.py
itamarst: Default logging is to stdout, which can be confusing... I'm thinking we'll have `-v` flag that logs to stderr or something, at some point.

PR #1269, opened 2015-04-02
[FLOC-1590] Acceptance move container

dwgebler:
Fixes FLOC-1590

Creates a MongoDB container on node 1 and writes some data. Moves the container with its associated dataset via the API move container endpoint to node 2. Then after moving, shuts down the container on node 2 and launches a new container on node 2 with a different exposed external port, then connects to that container and ensures the data written by the first container is mounted and can be re-read. This approach ensures the container has actually moved; we don't want the acceptance test to pass by accident, which could happen if the first container on node 1 hasn't moved, but we can still read the data because of the proxy layer when connecting to the originally open port on node 2's IP.

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -404,6 +404,8 @@ class ContainerAPITests(TestCase):
                               u"mountpoint": u"/data/db"}],
             }
             created = cluster.create_container(mongodb)
+            created.addCallback(lambda _: self.addCleanup(
+                cluster.remove_container, mongodb[u"name"]))
             created.addCallback(
                 lambda _: get_mongo_client(cluster.nodes[0].address))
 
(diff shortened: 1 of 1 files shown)


flocker/acceptance/test_api.py
```diff
@@ -229,6 +229,34 @@ def create_container(self, properties):
...
+            headers={b"content-type": b"application/json"},
+            persistent=False
+        )
+
+        def check_response_code(response):
+            if response.code != OK:
+                raise ValueError("Unexpected response code:", response.code)
+            return response
+
+        request.addCallback(check_response_code)
```
itamarst: If you merge forward there's a utility function that does this for you.

flocker/acceptance/test_api.py
```diff
@@ -392,6 +420,69 @@ def query_environment(host, port):
...
+                )
+
+                def destroy_and_recreate(_, record):
+                    removed = cluster.remove_container(u"container")
+                    mongodb2 = mongodb.copy()
+                    mongodb2[u"ports"] = [
+                        {u"internal": 27017, u"external": 27018}
+                    ]
+                    mongodb2[u"host"] = cluster.nodes[1].address
+                    removed.addCallback(
```
itamarst: This is relying on the fact that creating a container with a reference to a dataset *doesn't* move the dataset, plus the different port number thing. So maybe add a comment to that effect, otherwise it's not clear why you're bothering doing this.

itamarst:
Looks good, merge forward if necessary, address two minor comments then merge.

PR #1271, opened 2015-04-02
List nodes api floc 1630

adamtheturtle:
Creating this PR for @itamarst 's work because the JIRA issue was awaiting code review.


flocker/acceptance/test_api.py
```diff
@@ -295,10 +310,26 @@ def get_test_cluster(test_case, node_count):
...
-    ))
+        nodes=[]
+    )
+
+    # Wait until nodes are up and running:
+    def nodes_available():
+        d = cluster.current_nodes()
+        d.addCallback(lambda (cluster, nodes): len(nodes) == node_count)
+        return d
+    agents_connected = loop_until(nodes_available)
```
adamtheturtle: :question: A lot of the hassle we have now with the acceptance tests comes from the lack of logging, particularly in loops without their own timeout / exception raising - a failed test is often just a `TimeoutError`. It might be worth adding a log message here. Optional because the issue mentions that this will change soon but do it if it is quick.

adamtheturtle:
As the issue says "Just enough to update the acceptance tests to use this" and the acceptance tests use this, it seems fine to merge after the logging comment is optionally addressed. However,r the acceptance tests failed - but I don't know whether that is related. Merge when you're satisfied that this does not decrease the likelihood of tests passing.

PR #1272, opened 2015-04-02
Getting started release process floc 1592

adamtheturtle:
Fixes FLOC-1592


exarkun:
Thanks.  I'm supposing you've actually followed these steps and they produce the desired result.  If so, great, please merge.

PR #1273, opened 2015-04-02
Accept an invariant parameter in pmap_field.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1650

### DIFF
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -383,6 +383,31 @@ class Node(PRecord):
     )
 
 
+def _get_node(default_factory):
+    """
+    Create a helper function for getting a node from a deployment.
+
+    :param default_factory: A one-argument callable which is called with the
+        requested hostname when no matching node is found in the deployment.
+        The return value is used as the result.
+
+    :return: A two-argument callable which accepts a ``Deployment`` or a
+             ``DeploymentState`` as the first argument and a ``unicode`` string
+             giving a node hostname as the second argument.  It will return a
+             node from the deployment object with a matching hostname or it
+             will return a value from ``default_factory`` if no matching node
+             is found.
+    """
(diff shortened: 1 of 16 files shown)


flocker/control/test/test_model.py
```diff
@@ -721,6 +721,18 @@ class Record(PRecord):
...
+        """
+        The ``invariant`` parameter is passed through to ``field``.
+        """
+        class Record(PRecord):
+            value = pmap_field(
+                int, int,
+                invariant=lambda pmap: (False, "Very strict invariant.")
+            )
+
+        self.assertRaises(InvariantException, Record, value={})
```
itamarst: This doesn't seem quite enough to prove it's using the invariant - maybe use an invariant the passes some things but not others, so you can check both positive and negative invariant use paths?

flocker/control/_model.py
```diff
@@ -316,6 +320,32 @@ def dataset(self):
...
+            if key != getattr(value, attribute):
+                return (
+                    False, "{} is not correct key for {}".format(key, value)
+                )
+        return (True, "")
+    return key_match_invariant
+
+
+# An invariant we use a couple times below in mappings from dataset_id to
+# Dataset instances
```
itamarst: "and Manifestation"

flocker/control/test/test_model.py
```diff
@@ -721,6 +721,18 @@ class Record(PRecord):
                  Record().value2.__class__.__name__) ==
                 ("SomethingAnotherPMap", "IntFloatPMap"))
 
+    def test_additional_keyword_arguments(self):
```
itamarst: `test_invariant`?

itamarst:
Address and merge.

PR #1276, opened 2015-04-02
Deploymentstate modifier floc 1653

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1653

Note this PR is not against master. Make sure to review #1275 first. When all the PRs in this issue are accepted I'll merge them and merge the result into master.

PR #1280, opened 2015-04-03
Nonmanifest datasets floc 1655

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1654

Note this PR is not against master. Make sure to review #1276 first. When all the PRs in this issue are accepted I'll merge them and merge the result into master.

### DIFF
--- a/flocker/control/_model.py
+++ b/flocker/control/_model.py
@@ -586,12 +586,28 @@ class DeploymentState(PRecord):
     """
     A ``DeploymentState`` describes the state of the nodes in the cluster.
 
-    :ivar PSet nodes: A set containing ``NodeState`` instances describing
-        the state of each cooperating node.
+    :ivar PSet nodes: A set containing ``NodeState`` instances describing the
+        state of each cooperating node.
+    :ivar PMap nonmanifest_datasets: A mapping from dataset identifiers (as
+        ``unicode``) to corresponding ``Dataset`` instances.  This mapping
+        describes every ``Dataset`` which is known to exist as part of the
+        cluster but which has no manifestation on any node in the cluster.
+        Such datasets may not be possible with all backends (for example, P2P
+        backends must always store datasets on some cluster node).  This
+        mapping does not convey further backend-specific information; backends
+        are responsible for maintaining or determining additional information
+        themselves given a dataset identifier.  The ``Dataset`` instances which
+        are values in this mapping convey discovered state, not configuration.
+        The fields which are for conveying configuration will not be
+        initialized to meaningful values.
     """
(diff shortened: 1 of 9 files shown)


flocker/control/_model.py
```diff
@@ -586,13 +586,29 @@ class DeploymentState(PRecord):
...
+        ``unicode``) to corresponding ``Dataset`` instances.  This mapping
+        describes every ``Dataset`` which is known to exist as part of the
+        cluster but which has no manifestation on any node in the cluster.
+        Such datasets may not be possible with all backends (for example, P2P
+        backends must always store datasets on some cluster node).  This
+        mapping does not convey further backend-specific information; backends
+        are responsible for maintaining or determining additional information
+        themselves given a dataset identifier.  The ``Dataset`` instances which
+        are values in this mapping convey discovered state, not configuration.
+        The fields which are for conveying configuration will not be
```
itamarst: Link to https://clusterhq.atlassian.net/browse/FLOC-1247

flocker/control/test/test_protocol.py
```diff
@@ -282,13 +298,18 @@ def test_nodestate_updates_node_state(self):
         """
         ``NodeStateCommand`` updates the node state.
         """
+        changes = (NODE_STATE, NONMANIFEST)
         self.successResultOf(
             self.client.callRemote(NodeStateCommand,
-                                   state_changes=(NODE_STATE,),
+                                   state_changes=changes,
                                    eliot_context=TEST_ACTION))
         self.assertEqual(
```
itamarst: :question: Insofar as you're saying both that the same dataset is on the node and that it is not on any node, this test may break someday if we ever add an invariant to enforce that. Probably we won't insofar as state may well be inconsistent due to race conditions, so maybe that's fine.

flocker/node/agents/blockdevice.py
```diff
@@ -803,11 +803,17 @@ def discover_state(self, node_state):
...
-                _manifestation_from_volume(v) for v in volumes
-                if v.host == self.hostname)
-        }
+        manifestations = {}
+        nonmanifest = {}
+
+        for volume in volumes:
+            dataset_id = unicode(volume.dataset_id)
+            if volume.host == self.hostname:
+                manifestations[dataset_id] = _manifestation_from_volume(
```
itamarst: Will unmounted datasets be handled in a separate issue? If so add a comment/link, otherwise maye it should be done here?

itamarst: I guess unmounted is still attached to this node, so really a separate issue.

PR #1281, opened 2015-04-03
Build a new Docker image for this test and use it.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1657

This won't actually work until the image is on Dockerhub.  We should set up an automated build for it (I lack access).

Or else someone should explain why the `docker-images` repository is a better place for this kind of thing and then maybe clusterhq/docker-images#13 is a better idea (and the Dockerfile can be dropped from this PR).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1281)


adamtheturtle:
To confirm, this will work because, as described in https://github.com/ClusterHQ/docker-images/pull/13, nothing else is ever using this, right? Please add a comment to the test to explain why it is special and resubmit. Won't this mean that we aren't testing a case which a user could run into and get the error message that we have been seeing in Buildbot?

> [From the docker-images PR] Hmm. Or possibly this is stupid and this Dockerfile belongs in the Flocker repository. Come to think of it, I'm not sure why docker-images exists at all.

Perhaps keeping the Dockerfiles with the code implies that the Dockerfile as it was in 0.3.0 corresponds to the image one will get when following the documentation for 0.3.0, but instead you get the latest images from the hub. But that doesn't seem particularly valid - particularly when we have examples of similar in Flocker (e.g. the new 404 page). Or perhaps when the RTD SSL proxy was useful, that proxy would be used for all documentation and not just the documentation that the proxy settings lived with. Maybe others have clearer ideas.

Oh, that reminds me, the "Readthedocs Flocker SSL Proxy" is no longer useful and highlights a pattern that when we have things in many hardly-touched repositories which don't go through code review, those repositories rot.

If this does get merged, there should be a follow-up issue to move the files from `docker-images` and perhaps to delete that repository.

Sorry for the incomplete / vague "review" but I'm just not sure what is best here. I'll give this back to you @exarkun but perhaps you'll want to just put it in review for someone else to look at.

adamtheturtle:
This isn't the only test with this problem. If the others (see failures on http://build.clusterhq.com/builders/flocker%2Finstalled-package%2Ffedora-20/builds/1452/steps/trial/logs/stdio) can be fixed similarly please create follow up issues or resolve them here.

exarkun:
> Won't this mean that we aren't testing a case which a user could run into and get the error message that we have been seeing in Buildbot?

Only the test method actually tries to remove the image.  Flocker itself never needs to do this so a user should never be able to encounter the problem being fixed in the test.

> Perhaps keeping the Dockerfiles with the code implies that the Dockerfile as it was in 0.3.0 corresponds to the image one will get when following the documentation for 0.3.0, but instead you get the latest images from the hub. But that doesn't seem particularly valid

That's a really good point.  Perhaps I should make the test pull a certain tag of the image corresponding to the version number of Flocker.  This would probably complicate the release process, though - or is there a way to have Docker hub automatically tag certain release versions of an image?  I didn't notice such a feature in my exploration of the service.

Since this particular image has no interesting behavior and should never have any interesting behavior, I don't think we should need to modify it.  So a compromise solution may be to just leave a comment pointing out that it's a kind of unversioned external resource and warning anyone who ends up needing to change it that they'll have to do something about that.

exarkun:
> This isn't the only test with this problem.

I think it **is** the only test with this particular problem.  The other failures look similar but are about containers being running rather than images being in use.

flocker/node/functional/docker-images/test_pull_image_if_necessary/README
```diff
@@ -0,0 +1,4 @@
+This is a do-nothing image that exists for use by the Flocker test suite.
+It exists only to verify that an image not locally present is pulled automatically.
+It is never used to start a container so it will always be possible to remove it and necessitate a pull
```
adamtheturtle: Add a period.

adamtheturtle:
Thanks, the comments make it very clear now. Create a follow-up issue to move images out of the `docker-images` repository. This can also include an instruction to version the images - it might not be necessary for this particular image but I can certainly imagine this biting us in the future.

If you like, perhaps as part of this issue, edit the README of the `docker-images` repository to state that it is deprecated and new, similar Dockerfiles should be with the relevant code. See https://github.com/ClusterHQ/fedora-packages/blob/master/README.rst for similar.

Merge when you are satisfied.

flocker/node/functional/docker-images/test_pull_image_if_necessary/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM busybox:latest
+MAINTAINER ClusterHQ, Inc. <flocker-users@googlegroups.com>
```
adamtheturtle: Not info@clusterhq.com? From https://www.clusterhq.com > Get in touch

flocker/node/functional/docker-images/README
```diff
@@ -0,0 +1 @@
+This directory contains Docker build contexts (and associated Dockerfiles) necessary to build images that are used by the test suite and required to be available to be pulled.
```
adamtheturtle: After modifying these Dockerfiles something something Dockerhub needs to be done. Please document that or link to relevant documentation. If it is big, that can be a follow up issue (but maybe put notes of how you do it on the issue).

This may also affect in-production documentation / tutorials in the future. But that isn't the case now so should probably not be documented.

wallrj:
Ok, I've figured out how to get Docker Hub to build this image and added some notes in:
 * https://clusterhq.atlassian.net/browse/FLOC-1696

Had to rename the README file so that Docker Hub could use it in its description.

I'll force another build and then merge if green.

wallrj:
Hmm. The problem still seems to be occurring on in this build:
 * http://build.clusterhq.com/builders/flocker%2Finstalled-package%2Ffedora-20/builds/1491
 * http://build.clusterhq.com/builders/flocker%2Finstalled-package%2Ffedora-20/builds/1491/steps/trial/logs/stdio

I checked that it's building the latest revision (93ff54954db74611584c200c51870f7f66f2851c)

```
[ERROR]
  File "/opt/flocker/lib/python2.7/site-packages/twisted/internet/defer.py", line 140, in maybeDeferred
    result = f(*args, **kw)
  File "/opt/flocker/lib/python2.7/site-packages/twisted/trial/_asynctest.py", line 279, in <lambda>
    return lambda: f(*args, **kwargs)
  File "/opt/flocker/lib/python2.7/site-packages/flocker/testtools/__init__.py", line 841, in run_process
    raise CalledProcessError(returncode=status, cmd=command, output=output)
subprocess.CalledProcessError: Command '['docker', 'rmi', u'flockerlocaltests/355822175895']' returned non-zero exit status 1

flocker.node.functional.test_deploy.DeployerTests.test_links
```
...
```
[ERROR]
  File "/opt/flocker/lib/python2.7/site-packages/flocker/node/functional/test_docker.py", line 348, in test_pull_image_if_necessary
    docker.remove_image(image)
  File "/opt/flocker/lib/python2.7/site-packages/docker/client.py", line 903, in remove_image
    self._raise_for_status(res)
  File "/opt/flocker/lib/python2.7/site-packages/docker/client.py", line 94, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
docker.errors.APIError: 409 Client Error: Conflict ("Conflict, cannot delete a76a82ba882f because the container e98b79fa9a3e is using it, use -f to force")

flocker.node.functional.test_docker.NamespacedDockerClientTests.test_pull_image_if_necessary
Ran 1576 tests in 317.136s

FAILED (skips=84, expectedFailures=10, errors=10, successes=1473)
program finished with exit code 1
elapsedTime=321.270958
```

Is it possible that tests are being run in parallel  on a single build slave?

Or perhaps this is a Docker bug after all.

wallrj:
Is it possible that the NamespacedDockerClientTests are running before the docker daemon has actually enacted the container cleanup that was initiated after the same test in `GenericDockerClientTests`.

In which case, perhaps the solution is retry the image removal until docker images reports it gone....or something along those lines.

wallrj:
I thought something like this might reproduce the Docker condition where Docker reports that it has removed a container, but the server state hasn't yet caught up and server refuses to delete the image for that container.

```
while true; do 
    container_id=$(docker run --detach clusterhq/test_pull_image_if_necessary)
    docker stop $container_id
    docker rm $container_id
    # I'd expect this to sometimes fail...but it doesn't
    docker rmi clusterhq/test_pull_image_if_necessary
done
```

https://github.com/docker/docker/pull/11137 may be a fix for this problem.
It looks like ContainerRm used to return success even if the cleanup operations had not succeeded. It will now only behave that way if you specify --force (I think)
So if incomplete container removal is what's preventing our subsequent image remove API call failing, we should see an error when the container removal fails.

exarkun:
Resolved this in a slightly different way.  See #1878.

wallrj:
I deleted the DockerHub automated build: https://hub.docker.com/r/clusterhq/test_pull_image_if_necessary

PR #1282, opened 2015-04-03
[FLOC-1553] Split calculation of container and ZFS logic - design sketch

itamarst:
The basic idea is that you do what you can in each iteration, and because you're doing this repeatedly eventually all the work will be done.

Worries:
* This loses two-phase push since we don't have leases. Don't think there's actual solution.
* flocker-changestate support (adding its own little loop) is getting painful. I can imagine trying to get it to use the actual convergence loop but that still involves some extra work (e.g. there's no AMP, so we'd have to use a fake, and oh god I don't want to write these tests).

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -318,19 +318,6 @@ class ResizeDataset(object):
 
 
 @implementer(IStateChange)
-@attributes(["dataset"])
-class WaitForDataset(object):
-    """
-    Wait for a dataset to exist and be owned locally.
-
-    :ivar Dataset dataset: Dataset to wait for.
-    """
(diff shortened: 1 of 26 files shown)


flocker/node/_deploy.py
```diff
@@ -828,6 +826,8 @@ def change_node_state(deployer, desired_configuration,  current_cluster_state):
     """
     Change the local state to match the given desired state.
 
+    Legacy code used by flocker-changestate.
+
```
exarkun: Maybe a reference to the issue for deleting this.

exarkun:
Thanks, looks good, please proceed.

PR #1283, opened 2015-04-03
[FLOC-1660] Configuration API endpoint

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1660

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -28,7 +28,8 @@ from . import (
     AttachedVolume, Link
 )
 from ._config import (
-    ApplicationMarshaller, FLOCKER_RESTART_POLICY_NAME_TO_POLICY
+    ApplicationMarshaller, FLOCKER_RESTART_POLICY_NAME_TO_POLICY,
+    model_from_configuration, FigConfiguration, FlockerConfiguration
 )
 from .. import __version__
 
@@ -795,6 +796,35 @@ class ConfigurationAPIUserV1(object):
 
         # Didn't find the application:
         raise CONTAINER_NOT_FOUND
+
+    @app.route("/configuration/_compose", methods=['POST'])
+    # Undocumented for now since this is temporary, private endpoint for
+    # flocker-deploy.
+    @structured(
+        inputSchema={
+            '$ref':
+            '/v1/endpoints.json#/definitions/configuration_compose'
+        },
+        outputSchema={},
+        schema_store=SCHEMAS
+    )
+    def replace_configuration(self, applications, deployment):
+        """
+        Replace the existing configuration with one given by flocker-deploy
+        command line tool.
+
+        :param applications: Configuration in Flocker-native or
+            Fig/Compose format.
+
+        :param deployment: Configuration of which applications run on
+            which nodes.
+        """
+        configuration = FigConfiguration(applications)
+        if not configuration.is_valid_format():
+            configuration = FlockerConfiguration(applications)
+        return self.persistence_service.save(model_from_configuration(
+            applications=configuration.applications(),
+            deployment_configuration=deployment))
 
 
 def manifestations_from_deployment(deployment, dataset_id):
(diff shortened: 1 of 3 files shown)


flocker/control/test/test_httpapi.py
```diff
@@ -3082,3 +3085,130 @@ def test_two_containers(self):
...
+class ConfigurationComposeTestsMixin(APITestsMixin):
+    """
+    Tests for the container configuration endpoint at
+    ``/configuration/_compose``.
+    """
+    def configuration_test(self):
+        """
+        POSTing to ``/configuration/_compose`` in Flocker's custom
+        configuration format changes the deployment configuration
+        appropriately.
```
exarkun: What's "appropriately"?

flocker/control/test/test_httpapi.py
```diff
@@ -3082,3 +3085,130 @@ def test_two_containers(self):
...
+        """
+        return self.configuration_test()
+
+    def test_fig_configuration_format(self):
+        """
+        POSTing to ``/configuration/_compose`` in Flocker's custom
```
exarkun: in fig's configuration format, I guess?  I wonder why this isn't done client-side.

flocker/control/httpapi.py
```diff
@@ -796,6 +799,38 @@ def delete_container_configuration(self, name):
...
+            Fig/Compose format.
+
+        :param deployment: Configuration of which applications run on
+            which nodes.
+        """
+        try:
+            configuration = FigConfiguration(applications)
+            if not configuration.is_valid_format():
+                configuration = FlockerConfiguration(applications)
+            return self.persistence_service.save(model_from_configuration(
```
exarkun: You noticed that this changes the behavior of `flocker-deploy` so it completes before the change has been deployed.  Perhaps that change could be avoided with a poll loop here that only provides a result when state and configuration are sufficiently similar?  Maybe we don't have an API that makes comparing a `DeploymentState` to a `Deployment` very easy, yet, though...

On one hand this doesn't really matter a lot because we're definitely going to drop `flocker-deploy` completely before 1.0.  On the other hand, what if I'm wrong and we fail to get rid of it in time? :/

exarkun:
Thanks.  There are some doc build errors and one sort of significant user-facing issue.  Please fix the doc build error and let's have a quick conversation about the user-facing issue.

flocker/control/test/test_httpapi.py
itamarst: Seemed easiest to just have all the parsing logic in one place on the server, mostly.

flocker/control/httpapi.py
```diff
@@ -796,6 +799,42 @@ def delete_container_configuration(self, name):
         # Didn't find the application:
         raise CONTAINER_NOT_FOUND
 
+    @app.route("/configuration/_compose", methods=['POST'])
+    @user_documentation(
+        """
+        Private API endpoint used by flocker-deploy.
```
exarkun: Maybe a little more emphasis than this.  What does "private" mean to the reader of our API documentation, anyway?  "... temporarily used by ...", "This will be removed soon.", that sort of thing.

flocker/control/httpapi.py
exarkun: Our conversation elsewhere concluded that it's a change in quantity rather than quality.  `flocker-deploy` may already complete before your application is running.  This change might make it a little more likely (depending what configuration you're deploying).  This is neither entirely good or bad.  Coupled with a change to `flocker-deploy` to tell the user that they might have to wait a while, it's probably a net positive.  So we'll just go with that.

flocker/control/test/test_httpapi.py
exarkun: Okay.  Just fix the docstring then.

flocker/control/test/test_httpapi.py
```diff
@@ -3082,3 +3085,131 @@ def test_two_containers(self):
...
+                'environment': {'WORDPRESS_ADMIN_PASSWORD': 'admin'},
+                'volumes': ['/var/www/wordpress'],
+                'image': 'sample/wordpress',
+                'ports': ['8080:80'],
+                'links': ['mysql:db'],
+            },
+            'mysql': {
+                'image': 'sample/mysql',
+                'ports': ['3306:3306', '3307:3307'],
+            }
```
exarkun: All these strings are meant to be unicode I think

exarkun:
Great, thanks.  Please address two minor inline comments and merge.

PR #1286, opened 2015-04-07
Client installation instructions should be those just released

adamtheturtle:


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -478,7 +478,7 @@ Post-Release Review Process
 
 #. Verify that the client (``flocker-deploy``) can be installed on all supported platforms:
 
-   Follow the :ref:`Flocker client installation documentation<installing-flocker-cli>`.
+   Follow the Flocker client installation documentation at https://docs.clusterhq.com/en/${VERSION}/indepth/installation.html#installing-flocker-cli.
```
tomprince: Would linking to either `/en/latest/` or `/en/devel` work instead?

adamtheturtle: It would be different for marketing and development releases. I don't have a strong preference either way - if you do, I'll change it.

itamarst:
Fix spell checker errors, merge.

PR #1287, opened 2015-04-07
[FLOC-1065] Deb repositories

adamtheturtle:
https://clusterhq.atlassian.net/browse/FLOC-1065


admin/aws.py
```diff
@@ -290,7 +290,7 @@ def perform_upload_s3_key_recursively(dispatcher, intent):
                 UploadToS3(
                     source_path=intent.source_path,
                     target_bucket=intent.target_bucket,
-                    target_key=intent.target_key,
+                    target_key="%s/%s" % (intent.target_key, file),
```
adamtheturtle: IIRC I didn't do this (or something similar) because AWS key names may end with `/`. Can you confirm that this isn't the case or work around it.

adamtheturtle:
Fix the merge conflicts.

admin/yum.py
```diff
@@ -93,18 +101,42 @@ class CreateRepo(object):
...
+        # The update option means that this is faster when there is existing
+        # metadata but has output starting "Could not find valid repo at:" when
+        # there is not existing valid metadata.
+        check_call([
+            b'createrepo',
+            b'--update',
+            b'--quiet',
+            intent.repository_path.path])
+        return _list_new_metadata(repository_path=intent.repository_path)
+    elif package_type == PackageTypes.DEB:
```
adamtheturtle: This file is called `yum.py`. It should either be renamed or the Ubuntu packaging work should be moved.

admin/test/test_release.py
```diff
@@ -5,8 +5,8 @@
 """
 
 import os
-from unittest import skipUnless, TestCase
-import tempfile
+from twisted.trial.unittest import SynchronousTestCase as TestCase
```
adamtheturtle: Is this "as" just to not have to change many test definitions? Can it just be a find / replace?

admin/yum.py
```diff
@@ -93,18 +101,42 @@ class CreateRepo(object):
...
+        intent.repository_path.child('Release').setContent(
+            "Origin: ClusterHQ\n")
+
+        with intent.repository_path.child(
+                'Packages.gz').open(b"w") as raw_file:
+            with GzipFile(b'Packages.gz', fileobj=raw_file) as gzip_file:
+                gzip_file.write(metadata)
+        return {'Packages.gz', 'Release'}
+    else:
+        raise NotImplementedError("Unknwon package type: %s"
```
adamtheturtle: Typo, should be "Unknown". Also add a `\n` to the output.

admin/yum.py
```diff
@@ -134,23 +167,38 @@ def _perform_create_repository(self, dispatcher, intent):
...
+
+            return {'repodata/repomod.xml', 'repodata/<newhash>-metadata.xml'}
+        elif package_type == PackageTypes.DEB:
+            index = intent.repository_path.child('Packages.gz')
+            index.setContent("Packages.gz for: " + ",".join(packages))
+            intent.repository_path.child('Release').setContent(
+                "Origin: ClusterHQ\n")
+            return {'Packages.gz', 'Release'}
+        else:
+            raise NotImplementedError("Unknwon package type: %s"
```
adamtheturtle: Typo, should be "Unknown". Also add a `\n` to the output.

admin/test/test_release.py
```diff
@@ -726,149 +713,25 @@ def update_repo(self, aws, yum,
...
-
-        self.assertIn(
-            os.path.join(self.target_key, 'repodata', 'repomd.xml'),
-            aws.s3_buckets[self.target_bucket])
-
-    def test_repository_added_to(self):
-        """
-        Calling :func:`update_repo` does not delete packages or metadata which
-        already exist in S3.
+        - Existing package on S3 are preserved in the metadata.
```
adamtheturtle: package > packages

admin/test/test_release.py
```diff
@@ -726,149 +713,25 @@ def update_repo(self, aws, yum,
...
+        - Other packages on the buildserver are not downloaded.
+        - Existing metadata files are left untouched.
```
adamtheturtle: We spoke about this but I want to clarify: The problem with what existed was that there were many tests with the same lengthy setup and an assertion about one aspect of the outcome. This meant that changing something meant changing many tests. So this branch consolidates some of those tests into tests like this one, with a list of criteria in the docstring. Right?

Perhaps a better solution would be to have more test classes with shared `setUp`s, or to use https://testtools.readthedocs.org/en/latest/for-test-authors.html#delayed-assertions. This could mean that test failures are represented in a way which would be easier to debug than a comparison of two large lists.

admin/release.py
```diff
@@ -372,27 +372,38 @@ def update_repo(rpm_directory, target_bucket, target_key, source_repo,
     :param distro_version: The distro_version of the distribution to upload
         packages for.
     """
-    rpm_directory.createDirectory()
+    package_directory.createDirectory()
+
+    from .packaging import Distribution
```
adamtheturtle: I guess / hope that when https://clusterhq.atlassian.net/browse/FLOC-1223 is merged this can be moved to the top of the file.

admin/yum.py
```diff
@@ -93,18 +101,42 @@ class CreateRepo(object):
...
+        # there is not existing valid metadata.
+        check_call([
+            b'createrepo',
+            b'--update',
+            b'--quiet',
+            intent.repository_path.path])
+        return _list_new_metadata(repository_path=intent.repository_path)
+    elif package_type == PackageTypes.DEB:
+        metadata = check_output([
+            b'dpkg-scanpackages',
```
adamtheturtle: The merge conflicts are causing the tests to fail so I can't' be sure but I can't see how this would be available.

adamtheturtle:
Thanks @tomprince , I don't want to do a more thorough review until the tests pass and the potential re-arranging I suggested happens. Please re-submit when the comments are addressed.

admin/yum.py
adamtheturtle: In the Vagrant dev box I ran `pip install -e .[dev,release]` then `dpkg-scanpackages` and got `command not found`.

admin/yum.py
tomprince: This is an exception message, which doesn't typically have `\n`.

admin/yum.py
tomprince: I agree, but I'm not sure if it is worth changing all the code to reflect this.

admin/yum.py
tomprince: Unfortunately, it doesn't appear to be packaged for centos: 
- https://bugzilla.redhat.com/show_bug.cgi?id=1149590
- https://copr-fe.cloud.fedoraproject.org/coprs/xaeth/dpkg/

tomprince: I've submitted it for review without addressing this. I'm going to see if I can get upstream to package this, first.

adamtheturtle:
Was https://github.com/ClusterHQ/flocker/pull/1287#discussion_r28410000 lost as an outdated comment or do you think it isn't worth worrying about? This seems like an ideal use of delayed assertions.

admin/yum.py
adamtheturtle: I think renaming the file is worth it. You can make the judgement whether moving the Ubuntu packaging logic out to another file is or not.

admin/test/test_release.py
```diff
@@ -1334,81 +1200,44 @@ def test_marketing_repositories_created(self):
...
 
-        files_on_s3 = aws.s3_buckets[self.target_bucket].keys()
+def create_fake_repository(test_case, files):
+    """
+    Create files in a directory to mimic a repository of packages.
 
-        self.assertTrue(
-            expected_files.issubset(set(files_on_s3)),
-            "Metadata files for the packages were not created.")
+    :param dict source_repo: Dictionary mapping names of files to create to
```
adamtheturtle: :param test_case:

admin/test/test_release.py
```diff
@@ -896,313 +759,307 @@ def test_repository_added_to(self):
...
         )
 
-        index_path = os.path.join(self.target_key, 'repodata', 'repomd.xml')
-        # The index contains details about the new packages in the repository
-        # which has those new packages. The other repository has no packages
-        # and so the index is different.
-        self.assertNotEqual(
-            aws.s3_buckets[self.target_bucket][index_path],
-            aws.s3_buckets[self.alternative_bucket][index_path])
+    def setUp(self):
```
adamtheturtle: Many of these variables now aren't used. The tests still passed with:

```python
    def setUp(self):
        self.scratch_directory = FilePath(self.mktemp())
        self.scratch_directory.createDirectory()
        self.target_bucket = 'test-target-bucket'
        self.build_server = 'http://test-build-server.example'
        # self.alternative_bucket = 'bucket-with-existing-package'
        # alt_scratch_directory = FilePath(self.mktemp())
        # alt_scratch_directory.createDirectory()
        # self.alternative_package_directory = alt_scratch_directory.child(
        #     b'distro-version-arch')
        # self.operating_systems = [
        #     Distribution(name='fedora', version='20'),
        #     Distribution(name='centos', version='7'),
        #     Distribution(name='ubuntu', version='14.04'),
        # ]
        # self.repo_contents = {
        #     'clusterhq-flocker-cli-0.3.3-0.dev.7.noarch.rpm': 'cli-package',
        #     'clusterhq-flocker-node-0.3.3-0.dev.7.noarch.rpm': 'node-package',
        # }
```

admin/release.py
adamtheturtle: It seems that this can be done here and I think for the other occurrences in this changeset of `from .packaging import`.

admin/yum.py
```diff
@@ -93,18 +101,42 @@ class CreateRepo(object):
...
+        # metadata but has output starting "Could not find valid repo at:" when
+        # there is not existing valid metadata.
+        check_call([
+            b'createrepo',
+            b'--update',
+            b'--quiet',
+            intent.repository_path.path])
+        return _list_new_metadata(repository_path=intent.repository_path)
+    elif package_type == PackageTypes.DEB:
+        metadata = check_output([
```
adamtheturtle: This section should be covered by a test, perhaps similar to `UpdateRepoTests.test_real_yum_utils`.

admin/yum.py
adamtheturtle: Even doing this review and seeing `FakeYum` in `test_fake_deb` is confusing.

adamtheturtle:
Thanks @tomprince , please address the comments and resubmit when the `dpkg-scanpackages` problem is solved.

admin/test/test_release.py
tomprince: There is a trade-off here. We know exactly what we want the final state of the bucket to be. We can either assert that is what we want, or we could write a bunch of assertions that assert the same thing. In the former, it is slightly less clear what specific things are being tested for each part. On the other hand, the latter is more complicated to write and (I think) maintain.

admin/yum.py
tomprince: I had started to make this change, and got part way through and it was getting large, and slightly confusing. It would be much better as a separate follow-up branch.

tomprince:
The dev-box is failing because the zfs epel-testing repository is currently broken. I've asked @ryao to look into it.

adamtheturtle:
Ok thanks @tomprince . Merge when you're happy and Buildbot is to your satisfaction. Create follow-up issues for:
* The confusion of having `yum` names for `deb` logic (and preferably start working on it as a follow-up)
* Switching to an upstream DPKG package when available

and link to those issues from the code. The best live testing will be in https://clusterhq.atlassian.net/browse/FLOC-1103 and as this doesn't affect anything (I hope) until that is done I'm happy for this to be merged.

* The epel-testing problem? Upstream?

adamtheturtle:
Modify the outcomes section and the step which mentions RPMs in the release process.

adamtheturtle:
Add a section in ``packaging.rst`` similar to the current ``Building Omnibus RPMs``.

adamtheturtle:
Change the ``Package Hosting`` section in ``packaging.rst`` to describe the Ubuntu packaging. Sorry, because I've had a few thoughts on what else needs to be done please re-submit when addressed.

vagrant/dev/bootstrap.sh
```diff
@@ -25,3 +25,8 @@ yum install -y \
 	libffi-devel openssl-devel \
 	yum-utils \
 	pypy pypy-devel
+
+
+# dpkg isn't currently available upstream
```
adamtheturtle: This comment is now incorrect, right?

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -17,6 +17,7 @@ By the end of the release process we will have:
 - a Python wheel in the `ClusterHQ package index <http://archive.clusterhq.com>`_,
 - Fedora 20 RPMs for software on the node and client,
 - CentOS 7 RPMs for software on the node and client,
+- Ubuntu 14.04 RPMs for software on the node and client,
```
adamtheturtle: These aren't RPMs.

docs/gettinginvolved/infrastructure/packaging.rst
```diff
@@ -95,8 +96,10 @@ This bucket has the following policy::
 
 A policy can be set by going to a bucket's "Properties", "Permissions", then "Add bucket policy".
 
-`clusterhq-release package`
----------------------------
+``clusterhq-release`` package
+-----------------------------
+
+RPM-based distribution tend to bundle ``yum`` repository definitions in ``*-release`` packages.
```
adamtheturtle: distribution > distributions

adamtheturtle:
Thanks Tom, please address the minor comments and merge when you are happy and satisfied that there are no build failures introduced. I can't keep in my head a state of what is and what is not OK to fail.

vagrant/dev/bootstrap.sh
tomprince: It isn't *quite* available. It is currently a proposed update, and should land in a week or two.

PR #1285, opened 2015-04-06
Refactor istatechange entrypoint floc 1666

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1666

First step towards generic logging for IStateChange execution (FLOC-1591).

PR #1288, opened 2015-04-07
Move run logic floc 1667

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1667

Depends on FLOC-1666 / #1285, review that first.

This is the second step towards FLOC-1591.  It centralizes `IStateChange.run` calls in one place so it is easy to manipulate them (eg by logging them all).

### DIFF
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -75,6 +75,9 @@ def run_state_change(change, deployer):
     return change.run(deployer)
 
 
+# run_state_change doesn't use the IStateChange implementation provided by
+# _InParallel and _Sequentially but those types provide it anyway because
+# certain other areas of the test suite depend on it.
 @implementer(IStateChange)
 class _InParallel(PRecord):
     changes = field(type=PVector, factory=pvector, mandatory=True)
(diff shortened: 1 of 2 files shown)


flocker/node/_change.py
```diff
@@ -62,16 +62,33 @@ def run_state_change(change, deployer):
 
     :return: ``Deferred`` firing when the change is done.
     """
+    if isinstance(change, _InParallel):
+        return gather_deferreds(list(
```
wallrj: I see that `gather_deferreds` was used in the previous implementation too but I forgot that I wrote that and I forgot that it exists in order to log all failures while returning only the first failure.

Perhaps that's no longer necessary if the error logging is going to be done with Eliot,  and closer to the constituent deferred.

wallrj:
Thanks @exarkun 

This looks good.

I tried running the acceptance tests against the followup branch https://github.com/ClusterHQ/flocker/compare/log-istatechange-FLOC-1591 but couldn't get the nodes to co-operate so I'll try again when the that branch is up for review.

Please merge.

flocker/node/_change.py
exarkun: Since it works and it's just a refactoring, I'll probably leave it as-is.  I can imagine revisiting this decision at some future point though.

PR #1270, opened 2015-04-02
[FLOC-1643] Disable SELINUX

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1643

This disables SELINUX for the current  environment and upon reboot, by modifying the /etc/selinux/config file.

I didn't want to force the user to reboot again so I added both commands.

PR #1290, opened 2015-04-07
[FLOC-1672] Disable SELinux

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1672

This has already been reviewed and merged into the 0.4.0pre1 branch in #1270

This is exactly the same change applied to master but with some minor modifications to use `Effect` based command sequences.

See http://doc-dev.clusterhq.com/gettinginvolved/contributing.html#maintenance-branches for maintenance branch process.


wallrj:
I had to merge forward again and resolve some conflicts with the recent switch to Effect in the provisioning modules.

flocker/provision/_install.py
```diff
@@ -156,6 +156,21 @@ def task_install_kernel_devel():
...
+    """
+    Disable SELinux for this session and permanently.
+    XXX: Remove this when we work out suitable SELinux settings.
+    See https://clusterhq.atlassian.net/browse/FLOC-619.
+    """
+    return sequence([
+        run("if selinuxenabled; then setenforce 0; fi"),
+        run("test -e /etc/selinux/config && "
+            "sed --in-place='.preflocker' "
+            "'s/^SELINUX=.*$/SELINUX=disabled/g' "
```
tomprince: We should probably set this to permissive, rather than disabled. This will cause errors to be logged (but not prevented). This will be useful when we decide to properly address selinux support.

tomprince:
Other than the above comment, this looks good to merge.

flocker/provision/_install.py
wallrj: Well I think I'll postpone that change for https://clusterhq.atlassian.net/browse/FLOC-619

PR #1291, opened 2015-04-07
[FLOC-1635] Initial CLI for provisioning.

adamtheturtle:
https://clusterhq.atlassian.net/browse/FLOC-1635

### DIFF
--- a/docs/indepth/installation.rst
+++ b/docs/indepth/installation.rst
@@ -403,3 +403,35 @@ You have also ensured that the ``flocker-deploy`` command line tool is able to c
 Next you may want to perform the steps in :doc:`the tutorial <./tutorial/moving-applications>` , to ensure that your nodes are correctly configured.
 Replace the IP addresses in the ``deployment.yaml`` files with the IP address of your own nodes.
 Keep in mind that the tutorial was designed with local virtual machines in mind, and results in an insecure environment.
+
+
+Experiment on a remote testing cluster
+--------------------------------------
+
+Requires - e.g. Credentials for cloud providers of your choice.
+Maybe permissions you need?
+
+You can quickly start up a cluster to test on.
+Use the flocker-provision tool (see installing flocker-cli).
+
+This installs CentOS 7 on Rackspace nodes.
+
+Options are:
+* Cloud provider? Maybe
+   * For each cloud provider, credential options (e.g. rackspace secret key)
+* Number of nodes
+* Instance details - have defaults which work but user can decide instance size etc.
+
+For example:
+
+flocker-provision create -d rackspace --rackspace-secret=secret --num-agent-nodes=3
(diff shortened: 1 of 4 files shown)


adamtheturtle:
I think that this should be continued with as it matches what has been signed off on by @robhaswell .

docs/indepth/installation.rst
```diff
@@ -403,3 +403,35 @@ You have also ensured that the ``flocker-deploy`` command line tool is able to c
...
+Options are:
+* Cloud provider? Maybe
+   * For each cloud provider, credential options (e.g. rackspace secret key, region)
+* Number of nodes
+* Instance details - have defaults which work but user can decide instance size etc. - follow-up?
+
+For example:
+
+flocker-provision create --driver rackspace --rackspace-secret=secret --num-agent-nodes=3 --rackspace-region=us-west-1
+> Creating... 90%
```
adamtheturtle: Will this progress report be possible? (If not) What helpful output could we give to show that something is going on?

adamtheturtle:
We mentioned potentially having this tool output the YAML files necessary for the tutorial(s) but decided against this because the `flocker-deploy` tool may go away / change significantly soon. We should find out more detail on the plan and see how this tool could be useful for a user following the tutorial - it isn't fun changing all of the IP addresses in the YAML files and console commands.

docs/indepth/installation.rst
```diff
@@ -403,3 +403,35 @@ You have also ensured that the ``flocker-deploy`` command line tool is able to c
 Next you may want to perform the steps in :doc:`the tutorial <./tutorial/moving-applications>` , to ensure that your nodes are correctly configured.
 Replace the IP addresses in the ``deployment.yaml`` files with the IP address of your own nodes.
 Keep in mind that the tutorial was designed with local virtual machines in mind, and results in an insecure environment.
+
+
+Experiment on a remote testing cluster
```
adamtheturtle: Might be best to make it clear from the title how this differs from the above - one might spin up a Rackspace / CentOS 7 machine manually in order to experiment.

Also, the installation page is getting awfully long right now so maybe we should ask @alicesowerby about where this should go.

flocker/provision/script.py
```diff
@@ -0,0 +1,120 @@
...
+from flocker.provision._install import (
+    configure_cluster,
+)
+
+
+class CreateOptions(Options):
+    # TODO longdesc and synopsis
+    # TODO look at docker machine for inspiration
+    optParameters = [
+        ['driver', 'd', 'rackspace', 'choose cloud provider'],
```
adamtheturtle: Maybe not just cloud provider - this might spin up Vagrant VMs?

tomprince:
This is on hold for now.

PR #1284, opened 2015-04-06
[FLOC-1661] flocker-deploy uses endpoint instead of flocker-report/changestate

itamarst:
Depends on #1283.

Fixes https://clusterhq.atlassian.net/browse/FLOC-1563 too.

PR #1292, opened 2015-04-07
[FLOC-1663] Remove flocker-reportstate and friends

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1663

Depends on #1284 being merged first.


exarkun:
There are still mentions of `flocker-reportstate` and `flocker-changestate` in some places (mostly, but not entirely, in the user-facing documentation) which are no longer accurate.

exarkun:
Apart from that, great.  Please removing the remaining references and merge when buildbot is green.

PR #1293, opened 2015-04-07
[FLOC-1648] Don't enable shared folders when building dev box.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1648

This addresses the failure in the issue. There appear to be other issues, but they can be addressed separately.

### DIFF
--- a/vagrant/dev/Vagrantfile
+++ b/vagrant/dev/Vagrantfile
@@ -45,6 +45,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Cleanup
   config.vm.provision :shell, :path => "../cleanup.sh", :privileged => true
 
+  config.vm.synced_folder ".", "/vagrant", disabled: true
+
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Because this is part of the failure it would be clearer in JIRA if there were sub-tasks.

This is annoying because I use this shared folder for development, but less annoying than not being able to build the box.

Put a comment by the new line, and merge, but figure out some way of representing that this has been done in JIRA without accepting the issue (probably sub-tasks). Or, would it help to go back to the old system of maintaining a box with an updated kernel?

tomprince:
> This is annoying because I use this shared folder for development, but less annoying than not being able to build the box.

This only affects whether there is a shared folder during build, not when it is being used.

tomprince:
> figure out some way of representing that this has been done in JIRA without accepting the issue (probably sub-task

I'll just file a separate issue, since the failure documented in this one is resolved by this.

PR #1298, opened 2015-04-09
[FLOC-1695] Adjust documentation / and /en/ redirects to match reality.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1695


adamtheturtle:
Thanks @tomprince. Assuming that these commands were run and things now work as you expect, merge.

PR #1305, opened 2015-04-14
Base branch floc 1706

adamtheturtle:


admin/release.py
```diff
@@ -506,3 +542,136 @@ def publish_rpms_main(args, base_path, top_level):
         raise SystemExit(1)
     finally:
         scratch_directory.remove()
+
+
+def create_release_branch(version, path):
```
itamarst: As a general design principle it would be better to separate the function that calculates the base branch from the function that creates the release branch, because it's good to separate read-only calculations that return something and side-effecty operations that change something. Eiffel calls these `functions` and `procedures` respectively. 

In practice I wouldn't bother changing it at this point, but worth keeping in mind for next time.

admin/test/test_release.py
```diff
@@ -1412,3 +1418,164 @@ def get_dispatcher(self):
...
+              UsageError,
+              options.parseOptions, [])
+
+class CreateReleaseBranchTests(TestCase):
+    """
+    Tests for :func:`create_release_branch`.
+    """
+
+    def setUp(self):
+        repo_directory = FilePath(tempfile.mkdtemp())
```
itamarst: If you use Twisted's `TestCase` or `SynchronousTestCase` you can use `self.mktemp()` to create temporary file paths. They end up in `_trial_temp` in test-specific directories. This means when debugging failing tests it's easy to figure out which files/directories belong to which test. It also means you don't need (or want to) cleanup after creating them unless maybe you created a really huge file.

admin/test/test_release.py
```diff
@@ -1412,3 +1418,164 @@ def get_dispatcher(self):
...
+          options = CreateReleaseBranchOptions()
+          self.assertRaises(
+              UsageError,
+              options.parseOptions, [])
+
+class CreateReleaseBranchTests(TestCase):
+    """
+    Tests for :func:`create_release_branch`.
+    """
+
```
itamarst: You never test that the new branch is created from the calculated base branch.

admin/test/test_release.py
```diff
@@ -1412,3 +1418,164 @@ def get_dispatcher(self):
...
+        """
+        Trying to create a release when a branch already exists for the given
+        version fails.
+        """
+        self.repo.create_head('release/flocker-0.3.0pre1')
+        self.repo.create_tag('0.3.0pre1')
+        self.repo.create_head('release/flocker-0.3.0')
+        self.assertRaises(
+            BranchExists,
+            self.create_release_branch, '0.3.0')
```
itamarst: No test for documentation release.

itamarst: No test for actual release base branch being returned (unless I'm missing it).

flocker/docs/_version.py
```diff
@@ -163,3 +169,37 @@ def is_pre_release(version):
             and parsed_version.weekly_release is None
             and parsed_version.commit_count is None
             and parsed_version.dirty is None)
+
+
+def get_pre_release(version):
```
itamarst: Given a class with methods, why are you adding new functions? Why doesn't the admin script just use `FlockerVersion` instances directly?

admin/release.py
```diff
@@ -506,3 +542,136 @@ def publish_rpms_main(args, base_path, top_level):
...
+    except UsageError as e:
+        sys.stderr.write("%s: %s\n" % (base_path.basename(), e))
+        raise SystemExit(1)
+
+    try:
+        create_release_branch(
+            version=options['flocker-version'],
+            path=FilePath(__file__).path)
+    except NotARelease:
+        sys.stderr.write("%s: Can't create a release branch for non-release.\n"
```
itamarst: Copy/pasting 5 times isn't great. Utility function?

itamarst: Or rather, do a utility function next time, doesn't matter for this.

itamarst:
Great, more automation is better! Address and merge.

admin/release.py
adamtheturtle: Ok

admin/test/test_release.py
adamtheturtle: This is what happens when I don't do TDD properly :disappointed:

flocker/docs/_version.py
adamtheturtle: `_version` is private and so I was thinking that the class was not directly exposed for a reason - perhaps because we will be changing our versioning to fit with PEP440. I was concerned about exposing more of an interface than was necessary for this branch.

adamtheturtle:
Turns out that https://clusterhq.atlassian.net/browse/FLOC-1223 is a blocker unless I want to do another nasty hack for importing git. That is up for review now.

PR #1307, opened 2015-04-14
[FLOC-1357] Log errors in the convergence loop, and then continue to next iteration.

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1357

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -15,7 +15,7 @@ control service, and sends inputs to the ConvergenceLoop state machine.
 
 from zope.interface import implementer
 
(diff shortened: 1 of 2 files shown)


dwgebler:
I can verify that the convergence loop does continue after logging an error and that this behaviour is sufficiently proved in the unit test. I guess it would be nice to see a test that records more than an empty RuntimeError in the log, with some informative traceback, but merge this if you're happy; the only build failures are spurious here and the acceptance criteria that If errors occur then they should be logged with Eliot and the loop should continue regardless have been met.

PR #1308, opened 2015-04-15
Rpmversion floc 1223

adamtheturtle:


flocker/docs/__init__.py
```diff
@@ -5,12 +5,4 @@
...
-from ._version import (
-    get_doc_version, get_installable_version,
-    is_pre_release, is_release, is_weekly_release,
-)
-
-__all__ = [
-    'get_doc_version', 'get_installable_version',
-    'is_pre_release', 'is_release', 'is_weekly_release',
-]
+__all__ = []
```
adamtheturtle: Is this necessary / good?

flocker/common/version.py
```diff
@@ -158,8 +160,74 @@ def is_pre_release(version):
...
+        # tag. Additionally there may be a ``dirty`` suffix which
+        # indicates that there are uncommitted changes in the
+        # working directory.  We probably don't want to release
+        # untagged RPM versions, and this branch should probably
+        # trigger and error or a warning. But for now we'll add
+        # that extra information to the end of release number.
+        # See https://clusterhq.atlassian.net/browse/FLOC-833
+        release.extend(remainder)
+
+    return RPMVersion(version=version, release='.'.join(release))
```
adamtheturtle: This is the only changed line in `make_rpm_version`.

admin/test/test_release.py
```diff
@@ -29,19 +32,20 @@ class MakeRpmVersionTests(TestCase):
     """
     def test_good(self):
```
tomprince: These tests should move.

admin/vagrant.py
```diff
@@ -129,7 +129,8 @@ def build_box(path, name, version, branch, build_server):
     # Generate the enviroment variables used to pass options down to the
     # provisioning scripts via the Vagrantfile
     env = os.environ.copy()
-    rpm_version, rpm_release = make_rpm_version(version)
+    rpm_version = make_rpm_version(version).version
+    rpm_release =  make_rpm_version(version).release
```
tomprince: Extra space.

flocker/common/version.py
```diff
@@ -158,8 +160,74 @@ def is_pre_release(version):
...
+    """
+    Parse the Flocker version generated by versioneer into a
+    :class:`RPMVersion`.
+
+    :param flocker_version: The versioneer style Flocker version string.
+    :return: An ``RPMVersion``.
+    """
+    # E.g. 0.1.2-69-gd2ff20c-dirty
+    # tag+distance+shortid+dirty
+    parts = flocker_version.split('-')
```
tomprince: I think it should be fairly easy to convert this parsing to use `_parse_version` above, and then formating the parts. If it is straightforward, do this now, otherwise, file an issue.

A related suggestion is to make this a method (`os_version`?) on `FlockerVersion`. (Again, this may be a better as a follow-up issue).

flocker/docs/__init__.py
tomprince: If there is nothing here, then we don't need `__all__`.

PR #1310, opened 2015-04-16
[FLOC-1553] Split calculation of container and ZFS logic

itamarst:
Depends on #1292.


flocker/node/_deploy.py
```diff
@@ -766,58 +809,14 @@ def calculate_necessary_state_changes(self, local_state,
...
-            phases.append(InParallel(changes=[
-                HandoffDataset(dataset=handoff.dataset,
-                               hostname=handoff.hostname)
-                for handoff in dataset_changes.going]))
-        # any datasets coming to this node should also be
-        # resized to the appropriate quota max size once they
-        # have been received
-        if dataset_changes.coming:
-            phases.append(InParallel(changes=[
-                WaitForDataset(dataset=dataset)
```
exarkun: Just for your consideration - deletion of this part of the convergence logic, along with all of the support code, might have made for a nice separate PR.  It seems independent and it would have knocked a couple hundred lines off this diff (though, admittedly, pretty much all `-`s so the effect may not have been particularly great).

flocker/node/_deploy.py
```diff
@@ -193,19 +193,18 @@ class StartApplication(object):
...
         volumes = []
         if application.volume is not None:
-            volume = deployer.volume_service.get(
-                _to_volume_name(
-                    application.volume.manifestation.dataset.dataset_id))
+            dataset_id = application.volume.manifestation.dataset_id
             volumes.append(DockerVolume(
                 container_path=application.volume.mountpoint,
-                node_path=volume.get_filesystem().get_path()))
+                node_path=self.node_state.paths[dataset_id]))
```
exarkun: Ooh.  Nice.

flocker/node/_deploy.py
```diff
@@ -454,7 +440,8 @@ def run(self, deployer):
         return gather_deferreds(results)
 
 
-class P2PManifestationDeployer(_OldToNewDeployer):
+@implementer(IDeployer)
+class P2PManifestationDeployer(object):
```
exarkun: Yay.  Thanks for cleaning this up.

flocker/node/_deploy.py
```diff
@@ -642,19 +675,17 @@ def applications_from_units(units):
...
                 paths=None,
-            )
+            )]
         d.addCallback(applications_from_units)
         return d
 
-    def calculate_necessary_state_changes(self, local_state,
-                                          desired_configuration,
-                                          current_cluster_state):
+    def calculate_changes(self, desired_configuration, current_cluster_state):
```
exarkun: The docstring for this method still talks about volumes a little bit more than it should.

flocker/node/_deploy.py
```diff
@@ -681,6 +712,13 @@ def calculate_necessary_state_changes(self, local_state,
 
         :return: A ``IStateChange`` provider.
         """
+        # We are a node-specific IDeployer:
+        current_node_state = current_cluster_state.get_node(self.hostname)
+        if current_node_state.applications is None:
+            # We don't know current application state, so can't calculate
+            # anything...
```
exarkun: Maybe worth mentioning why we might not know about application state, even though we're the application deployer (ie, we don't discover stuff until dataset state has been discovered).

flocker/node/_deploy.py
```diff
@@ -550,10 +583,10 @@ def discover_local_state(self, local_state):
         dataset agent. See
         https://clusterhq.atlassian.net/browse/FLOC-1646.
 
-        :return: A ``Deferred`` which fires with a ``NodeState`` instance
-            with information only about ``Application``.
-            ``NodeState.manifestations`` and ``NodeState.paths`` will not be
-            filled in.
+        :return: A ``Deferred`` which fires with alist containing a
```
exarkun: Typo - `alist`

flocker/node/_deploy.py
```diff
@@ -730,18 +766,25 @@ def calculate_necessary_state_changes(self, local_state,
...
-            StartApplication(application=app, hostname=self.hostname)
+            StartApplication(application=app, node_state=current_node_state)
             for app in desired_node_applications
-            if app.name in start_names
+            if ((app.name in start_names) and
+                # If manifestation isn't available yet, don't start:
+                # XXX in FLOC-1240 non-primaries should be checked.
+                (app.volume is None or
+                 app.volume.manifestation.dataset_id in
+                 current_node_state.manifestations))
```
exarkun: :question: This would probably make a good stand-alone function - `app_volumes_available(app, current_node_state)` or some such.

flocker/node/_deploy.py
```diff
@@ -931,20 +931,28 @@ def __init__(self, hostname, volume_service, docker_client=None,
...
+        d = self.manifestations_deployer.discover_state(local_state)
 
         def got_manifestations_state(manifestations_state):
-            app_discovery = self.applications_deployer.discover_local_state(
+            manifestations_state = manifestations_state[0]
+            app_discovery = self.applications_deployer.discover_state(
                 manifestations_state)
             app_discovery.addCallback(
-                lambda app_state: app_state.set(
+                lambda app_state: app_state[0].set(
```
exarkun: Eww.  So gross.  Good thing this whole class is being deleted soon.  Yay.

flocker/node/_deploy.py
```diff
@@ -497,25 +484,71 @@ def got_volumes(available_manifestations):
...
+        local_state = cluster_state.get_node(self.hostname)
+        # We need to know applications (for now) to see if we should delay
+        # deletion or handoffs.
+        if local_state.applications is None:
+            return Sequentially(changes=[])
+        phases = []
+
+        # For now we delay deletion and handoffs until we know application
+        # isn't using the dataset. Later on we'll use leases to decouple
+        # the application and dataset logic better.
```
exarkun: :question: Do we have an issue filed for leases yet?  It would be nice to link to it from here.

flocker/node/_deploy.py
```diff
@@ -766,58 +809,14 @@ def calculate_necessary_state_changes(self, local_state,
...
-                for dataset in dataset_changes.resizing]))
-
-        # Do an initial push of all volumes that are going to move, so
-        # that the final push which happens during handoff is a quick
-        # incremental push. This should significantly reduces the
-        # application downtime caused by the time it takes to copy
-        # data.
-        if dataset_changes.going:
-            phases.append(InParallel(changes=[
-                PushDataset(dataset=handoff.dataset,
```
exarkun: `PushDataset` is no longer used at all.  We could perhaps delete it and the supporting `VolumeService.push` code (I could imagine wanting to keep it since we'll want that functionality again in the future... but we can always dig it up then?  And maybe we won't want to... since the current API is blocking...).

exarkun: Oh, but not exactly.  `VolumeService.push` is also used by `HandoffDataset`, of course.  So we can't quite delete all of that.

flocker/node/_deploy.py
```diff
@@ -497,25 +484,71 @@ def got_volumes(available_manifestations):
...
+                for dataset in dataset_changes.resizing]))
+
+        going = [handoff for handoff in dataset_changes.going
+                 if handoff.dataset.dataset_id not in in_use_datasets]
+        if going:
+            phases.append(InParallel(changes=[
+                HandoffDataset(dataset=handoff.dataset,
+                               hostname=handoff.hostname)
+                for handoff in going]))
+
```
exarkun: Ooh.  No need to handling `coming` anymore, I see.  The other side pushes stuff here and ... it shows up when it shows up, no need to do anything special.  Neat.

flocker/node/functional/test_deploy.py
```diff
@@ -34,16 +34,23 @@ def change_node_state(deployer, desired_configuration):
...
-
-    def got_changes(changes):
-        cluster_state = DeploymentState()
-        for change in changes:
-            cluster_state = change.update_cluster_state(cluster_state)
-        return deployer.calculate_changes(desired_configuration, cluster_state)
-    d.addCallback(got_changes)
-    d.addCallback(lambda change: change.run(deployer))
-    return d
+    def _converge():
```
exarkun: Probably private enough, considering it's a nested function

flocker/node/functional/test_deploy.py
```diff
@@ -34,16 +34,23 @@ def change_node_state(deployer, desired_configuration):
...
+        def got_changes(changes):
+            cluster_state = DeploymentState()
+            for change in changes:
+                cluster_state = change.update_cluster_state(cluster_state)
+            return deployer.calculate_changes(
+                desired_configuration, cluster_state)
+        d.addCallback(got_changes)
+        d.addCallback(lambda change: change.run(deployer))
+        return d
+    # Repeat a few times until things settle down:
```
exarkun: You missed a nice chance to quote lewis carroll ("what I tell you three times is true") :smile:

exarkun:
Inconveniently, Github won't let me comment on `test_deploy.py` changes.  Therefore, enjoy this primitive review style:

  * in `test_links`, maybe instead of hard-coding a duplicate of the node name, it should use `EMPTY_NODESTATE.hostname`?  I imagine that we'll have to preserve some kind of addressing information on `NodeState` even when we switch to UUIDs for primary identification - otherwise we won't be able to set up links properly?
  * Some test cases are still named `CalculateNecessaryStateChanges` but are testing the new interface, `calculate_changes`, so could reasonably be given shorter names.
  * `test_metadata_does_not_cause_restarts` has a slightly confusing name since "restart" isn't a thing you can do to a volume.

That's all.  Thanks a lot.  Please address these, the minimal inline comments, and merge when you have an as-green-as-possible-oh-why-is-everything-so-terrible build.

flocker/node/_deploy.py
itamarst: Not going to bother unless it's used more than once.

itamarst: (If we want to simplify we can e.g. move ports stuff into its own method).

flocker/node/_deploy.py
itamarst: Hopefully we'll want to resurrect it real soon now? So may as well leave it in.

PR #1311, opened 2015-04-16
Staging docs floc 1712

adamtheturtle:


adamtheturtle:
This brought a problem to my attention which I have documented at https://clusterhq.atlassian.net/browse/FLOC-1719 but I don't think it is necessary to fix for this to be merged.

adamtheturtle:
I think this should also fix https://clusterhq.atlassian.net/browse/FLOC-1394 so would appreciate if a reviewer would close that if they agree.

admin/release.py
```diff
@@ -152,6 +152,8 @@ def publish_docs(flocker_version, doc_version, environment):
     :param bytes flocker_version: The version of flocker to publish the
         documentation for.
     :param bytes doc_version: The version to publish the documentation as.
+        Attempting to publish documentation to a staging environment as a
+        documentation version publishes it as the version being updated.
```
jongiddy: This explanation is a bit confusing. Make it more clear what the difference between `flocker_version` and `doc_version` is, both in production and staging.   A longer explanation in the body of the docstring (after the initial line, before the parameter list) would be useful.

admin/release.py
```diff
@@ -263,8 +265,9 @@ class PublishDocsOptions(Options):
          "The version of flocker from which the documentation was built."],
         ["doc-version", None, None,
          "The version to publish the documentation as.\n"
-         "This will differ from \"flocker-version\" for staging uploads and "
-         "documentation releases."],
+         "This will differ from \"flocker-version\" for staging uploads\n"
+         "Attempting to publish documentation as a documentation version "
+         "publishes it as the version being updated.\n"],
     ]
```
jongiddy: Again, the proliferation of the words "documentation" and "version" makes this hard to understand.

jongiddy:
If you can clarify the docstrings, adding a clearer explanation of the difference between what happens in production and staging doc releases, then I'm OK for this to be merged.

PR #1313, opened 2015-04-17
[FLOC-1637] Control service certificate

dwgebler:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1637

An initial documentation stub for the flocker-ca program exists at branch https://github.com/ClusterHQ/flocker/compare/flocker-ca-documentation-FLOC-1647?expand=1

Docs are not done in this branch because we don't want any flocker-ca docs in master until the authentication layer is completed and usable. The docs branch will continue to be worked on and developed as we continue the authentication layer and flocker-ca commands.

Besides the unit tests for this branch, you can also manually generate a root certificate with ``flocker-ca initialize mycluster`` and then run ``flocker-ca create-control-certificate`` in the same directory. Once your control certificate is generated, run

```
openssl verify -verbose -CAfile cluster.crt control-service.crt
```

to verify the control certificate is signed by the previously generated CA.

And run

```
openssl x509 -text -noout -in control-service.crt
```

to manually verify the generated certificate's expiry date, subject, issuer and public key.


flocker/ca/_ca.py
```diff
@@ -58,6 +60,15 @@ def __eq__(self, other):
     def __ne__(self, other):
         return not self.__eq__(other)
 
+    @classmethod
+    def generate(cls):
+        """
+        Create a new 4096-bit RSA key pair.
```
wallrj: If we aim to contribute this KeyPair class back to Twisted, perhaps we shouldn't add this Flocker specific factory method to it.

It could just as well be added to a new free function called eg `flocker_key_pair` which returns a `FlockerKeyPair` wrapper with the Flocker standard 4096bit RSA keypair.

You could also break `selfSignedCACertificate` into a standalone function leaving only the comparison improvements in the class.

And in that case, you could even rename `FlockerKeyPair` to something a little more generic eg ComparableKeyPair and then add that as a patch to https://twistedmatrix.com/trac/ticket/7847

I'm not sure whether there's a place in twisted for the self signed certificate helper function, but I have found a twisted branch where I add an example script that creates them....perhaps that could use this function instead.
 * https://twistedmatrix.com/trac/ticket/6744
 * https://github.com/twisted/twisted/compare/trunk...ssl-endpoint-description-docs-6744

I'd forgotten all about it to be honest.

flocker/ca/_ca.py
```diff
@@ -106,82 +117,167 @@ def selfSignedCACertificate(self, dn, request, serial, expiry, digest):
         return Certificate(cert)
 
 
-def generate_keypair():
+def load_certificate_from_path(path, key_filename, cert_filename):
```
wallrj: `load_certificate_from_directory` would be a more accurate name.

wallrj: It's nice that this is now a separate function BTW

flocker/ca/_script.py
```diff
@@ -16,8 +16,76 @@
...
+    def run(self):
+        """
+        Check if control service certificate already exist in current
+        directory. If it does, error out. Also check if root key and
+        certificate files (either default or as specified on the command line)
+        exist in the path and error out if they do not. If there are no path
+        errors, create a new control service certificate signed by the root
+        and write it out to the current directory.
+        """
+        d = Deferred()
```
wallrj: If this is all blocking code for the time being, there's no need for a nested function.

Just move the logic into the parent function and then `return succeed(None)` at the end.

flocker/ca/_ca.py
```diff
@@ -216,7 +312,7 @@ def initialize(cls, path, name):
             )
 
         dn = DistinguishedName(commonName=name)
-        keypair = generate_keypair()
```
wallrj: Yeah, I'd prefer `flocker_keypair()` and it should return an instance of ...let's call it `ComparableKeyPair` whose `keypair` attribute is a readonly `@property` which returns `self.keypair` `.original` ?? Not sure about that bit, or if there's an `IKeyPair` you could maybe use `proxyForInterface`.
Chat to @exarkun about what's be most  likely to be accepted back into Twisted.

flocker/ca/_script.py
```diff
@@ -16,8 +16,76 @@
...
+    * certfile: path to the root certificate file. Defaults to ./cluster.crt
+    """
+
+    synopsis = "<authoritypath>"
+
+    def parseArgs(self, *args):
+        self["path"] = FilePath(os.getcwd())
+        self["rootpath"] = FilePath(os.getcwd())
+        if len(args) == 1:
+            self["rootpath"] = FilePath(args[0])
```
wallrj: This doesn't seem to match the synopsis or the documentation above, which talk about authoritypath and certfile.

And if those are really optional, they should be defined as command line options rather than positional arguments.

wallrj:
Thanks @dwgebler

Notes:
 * I ran the script
 * It generated the certificate
 * Openssl verified that the certificate had been signed by the previously generated CA public key
 * The expiry date was set 20 years in the future.
 * Started an SSL twistd web server using the generated key and cert.
   `(1637)[/tmp/1637]$ twistd -n web --port=ssl:10443:privateKey=$PWD/control-service.key:certKey=$PWD/control-service.crt --path $PWD`

Points:
 1. The "common name" of the generated certificate was "control-service"
    But shouldn't it match the hostname or IP address of the control service node?
    Otherwise REST clients will get hostname verification errors.
    (perhaps that doesn't happen when using client certificates)
    Or if not Common Name, perhaps we need a way to specify a cluster domain name and use a wildcard certificate
    or add  all the IP addresses of the cluster nodes in the http://en.wikipedia.org/wiki/SubjectAltName field of the certificate

    1. `Subject: OU=mycluster, CN=control-service`

 2. Some of the error messages need to be more user friendly, I think.
    For example, when the CA cert isn't present, the message is currently:

    ```
    (1637)[/tmp/foo]$ flocker-ca create-control-certificate
    Error: Certificate file /tmp/foo/cluster.crt does not exist.
    ```

    Might be better if the error was "Error: Certificate file /tmp/foo/cluster.crt does not exist. Please initialise the certificate authority before running `whatever-next-step-was-attempted`

 3. I get confused by the word "Certificate" in "FlockerCertificate" "ControlCertificate" and "CertificateAuthority".

    Here's a little sketch of how the code might be re-arranged...for discussion.


```
class FlockerCredentials(PRecord):
    """
    A record of a private key and the corresponding public certificate (the
    public key) which may or may not be signed...but which always should be
    signed by or self-signed in a flocker world.
    """
    keypair = field(...)
    certificate = field(...)

    @classmethod
    def from_file(cls, directory, basename):
        """
        Load a basename.key and basename.crt from directory

        :return: A FlockerCredentials instance
        """

    def to_file(self, directory, basename):
        """
        Dump the keypair and certificate found in ``credentials`` in PEM format to
        ``directory``, using the supplied ``basename``.
        """


def make_certificate_authority_credentials():
    """
    Generate a keypair instance and a self-signed certificate instance and
    assign them to a FlockerCredentials record.

    :return: A FlockerCredentials instance
    """


def make_control_service_credentials(ca_credentials):
    """
    Generate a keypair instance and a certificate signed with the private key
    in the supplied ``ca_credentials``.
    """


def initialize_certificate_authority_credentials(options):
    """
    Raise UsageError if the supplied CA basename files already exist in directory.
    Raise UsageError if the directory can't be written to, etc....
    """
    # ca = make_certificate_authority_credentials(options[...])
    # try:
    #     ca.to_file(options[...])
    # except IOError:
    #     raise UsageError(...)


def initialize_control_service_credentials(options):
    """
    Raise UsageError if the supplied CA credentials cannot be loaded from directory using basename.
    Raise UsageError if the supplied control service credentials already exist.
    """
    # try:
    #     ca = FlockerCredentials.from_file(options[...])
    # except:
    #     raise UsageError('Can not find CA credentials')

    # control_service_credentials = make_control_service_credentials(ca)

    # try:
    #     control_service_credentials.to_file(options[...])
    # except IOError:
    #     raise UsageError(...)

```

Sorry this took so long.

Let's call this a design review.
 * See what you make of the suggestions above
 * Maybe discuss the design  @itamar
 * Submit for a code review.

dwgebler:
Thanks @wallrj 

1. I'll need @exarkun or @itamarst to comment on the identification metadata for the control service certificate. At the moment, besides the authority being the key that signed the control service certificate, the common name of that authority (e.g. `mycluster`) is used as the organizational unit name of the control service certificate, while the common name, as you've pointed out, is just `control-service` - so we have some basic identification that this certificate is for the control service that belongs to `mycluster`.
If we need anything beyond that, such as a hostname or UUID, the JIRA ticket didn't define it.
Node certificates (the next story) do include a UUID in their common name. IP addresses and hostnames might be subject to change in a cluster, so it doesn't necessarily make sense to me to bind the control service certificate to an IP (hostname, maybe...)

2. I have improved error messaging in the latest commits, while also refactoring the loading and writing code to be less "look before you leap" and rely more on catching exceptions rather than checking for potential errors beforehand. New output for the example you gave is:

```
dave@dave-ThinkPad-X230:~/flocker$ flocker-ca create-control-certificate
Error: Unable to load certificate authority file. Please run `flocker-ca initialize` to generate a new certificate authority. No such file or directory /home/dave/flocker/cluster.crt
```

3. I have refactored the code to make it considerably cleaner in a manner similar (but not exactly) in line with your suggestions. We still have (renamed now) `ControlCredential` and `RootCredential` (and later, `NodeCredential` and `UserCredential`) classes with `from_path` and `initialize` class methods, but they are greatly simplified. Probably easier to look through the new structure than for me to explain it :-)

I am of course open to refactoring this further based on any more comments from yourself, Itamar or Jean-Paul, but I think this is considerably more tidy, flexible and extensible now than it was.

flocker/ca/_ca.py
```diff
@@ -51,138 +66,240 @@ def __init__(self, keypair):
...
+        certFile.read(), format=crypto.FILETYPE_PEM)
+    keypair = ComparableKeyPair(
+        keypair=KeyPair.load(keyFile.read(), format=crypto.FILETYPE_PEM)
+    )
+    return (keypair, certificate)
+
+
+class FlockerCredential(PRecord):
+    """
+    Base class for Flocker credentials, comprising a ceritificate and
```
itamarst: "ceritificate"

flocker/ca/_ca.py
```diff
@@ -51,138 +66,240 @@ def __init__(self, keypair):
...
+    public/private key pair.
 
     :ivar FilePath path: A ``FilePath`` representing the absolute path of
-        a directory containing the CRT and KEY files.
+        a directory containing the certificate and key files.
     :ivar Certificate certificate: A signed certificate, populated only by
```
itamarst: I don't understand the "populated only by" bit.

flocker/ca/_ca.py
```diff
@@ -51,138 +66,240 @@ def __init__(self, keypair):
...
-        keyPath = path.child(key_filename)
 
-        if not certPath.isfile():
-            raise PathError(
-                b"Certificate file {path} does not exist.".format(
-                    path=certPath.path)
-            )
+class ControlCredential(FlockerCredential):
+    """
+    A certificate for a control service, signed by a supplied certificate
```
itamarst: And private key too.

flocker/ca/_script.py
```diff
@@ -16,10 +16,74 @@
...
+        directory. If it does, error out. Also check if root key and
+        certificate files (either default or as specified on the command line)
+        exist in the path and error out if they do not. If there are no path
+        errors, create a new control service certificate signed by the root
+        and write it out to the current directory.
+        """
+        self["inputpath"] = FilePath(self["inputpath"])
+        self["outputpath"] = FilePath(self["outputpath"])
+
+        d = Deferred()
```
itamarst: `Deferred` seems unnecessary for much of the code here. If external API requires it then return `succeed()` at the end, and normal blocking logic for the bulk of the code.

flocker/ca/_script.py
```diff
@@ -16,10 +16,74 @@
...
                              FlockerScriptRunner)
 
-from ._ca import (CertificateAuthority, CertificateAlreadyExistsError,
-                  KeyAlreadyExistsError)
+from ._ca import (RootCredential, ControlCredential,
+                  CertificateAlreadyExistsError, KeyAlreadyExistsError,
+                  PathError)
 
 
+@flocker_standard_options
```
itamarst: Don't think you want the standard options on sub-commands, seems like it should only be on the main `Options` class.

itamarst: /me goes to try command-line

itamarst: OK, so, both commands have serious output problems, with our without standard options. Including standard options does result in better output, but:

1. Linebreaks aren't preserved, so output is wonky. I think this is a Twisted "feature", not clear if there's any work to around it but at least we don't to look terrible like we do now.
2. `flocker-ca initialize --help` (and same for this command) has different output than `flocker-ca initialize`; in particular notice `<name>` is missing from the latter.
3. `flocker-ca --help` and `flocker-ca` should have usage `flocker-ca <command> [command-options]`.

flocker/ca/_script.py
```diff
@@ -88,12 +152,14 @@ class CAOptions(Options):
     API clients within a Flocker cluster.
     """
 
-    synopsis = "Usage: flocker-ca <command> [OPTIONS]"
+    synopsis = "Usage: flocker-ca"
```
itamarst: Previous version of this seems better. Why'd you change it?

itamarst:
Besides comments above, this is missing functional tests for the `flocker-ca` interactions of a user. A minimal test would be "generate a CA, then a control service certificate, make sure something reasonable was created" (either via pyOpenSSL/Twisted wrappers or by calling out to `openssl verify`).

itamarst:
Functional tests for likely error paths ("forgot to initialize CA") are probably good idea (but could be done in separate issue).

itamarst:
Address the comments above, and if changes are small enough merge, otherwise resubmit. Thanks!

exarkun:
I'd still like the documentation to be completed before progressing further on the implementation.

flocker/ca/_ca.py
```diff
@@ -51,138 +66,238 @@ def __init__(self, keypair):
...
+    """
+    req = request.original
+    cert = crypto.X509()
+    dn._copyInto(cert.get_issuer())
+    cert.set_subject(req.get_subject())
+    cert.set_pubkey(req.get_pubkey())
+    cert.gmtime_adj_notBefore(0)
+    cert.gmtime_adj_notAfter(expiry)
+    cert.set_serial_number(serial)
+    cert.add_extensions([
```
itamarst: Add some comments explaining these.

flocker/ca/_ca.py
```diff
@@ -51,138 +66,238 @@ def __init__(self, keypair):
...
+        crypto.X509Extension("basicConstraints", True,
+                             "CA:TRUE, pathlen:0"),
+        crypto.X509Extension("keyUsage", True,
+                             "keyCertSign, cRLSign"),
+        crypto.X509Extension("subjectKeyIdentifier", False, "hash",
+                             subject=cert),
+    ])
+    cert.add_extensions([
```
itamarst: And this.

flocker/ca/_ca.py
```diff
@@ -51,138 +66,238 @@ def __init__(self, keypair):
...
 
-        certPath = path.child(certificate_filename)
-        keyPath = path.child(key_filename)
 
-        if not certPath.isfile():
-            raise PathError(
-                b"Certificate file {path} does not exist.".format(
-                    path=certPath.path)
-            )
+class ControlCredential(FlockerCredential):
```
itamarst: :question: Rather than subclassing, it would be better to have a `FlockerCredential` instance as an attribute. In general subclassing should be avoided if at all possible.

itamarst:
Besides the above:
1. `flocker-ca initialize` with no extra arguments still has confusing output.
2. The new output code has no tests. A functional test for help output of `flocker-ca` should suffice.

We may wish to discuss switching to a different command line framework (not immediately but soon). This isn't going to be great for main `flocker` tool, e.g. wrapping to 80 columns isn't quite the right thing to do but we shouldn't be spending much time on trying to fix `Options`.

flocker/ca/test/test_ca.py
```diff
@@ -13,28 +13,299 @@
...
+        cc = ControlCredential.initialize(self.path, self.ca)
+        cert = cc.certificate.original
+        issuer = cert.get_issuer()
+        subject = cert.get_subject()
+        self.assertEqual(
+            issuer.CN,
+            subject.OU
+        )
+
+    def test_certificate_is_signed_by_ca(self):
```
itamarst: Not sure this actually proves what it claims to prove. But functional test does provide more coverage.

flocker/ca/_ca.py
dwgebler: There is a comment explaining these; the link to ttps://www.openssl.org/docs/apps/x509v3_config.html#Basic-Constraints on line 82.

PR #1314, opened 2015-04-17
[FLOC-1487] Create and list OpenStack block volumes

wallrj:
Fixes:
 * https://clusterhq.atlassian.net/browse/FLOC-1487
 * https://clusterhq.atlassian.net/browse/FLOC-1483


flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,114 @@
...
+        http://docs.rackspace.com/cbs/api/v1.0/cbs-devguide/content/POST_createVolume_v1__tenant_id__volumes_volumes.html
+
+        Discussion:
+         * Rackspace (maybe cinder in general) supports a block device type, eg SSD or SATA.
+           I guess we'll hardcode SATA here to start with
+           Maybe expand the API later to make the disk type configurable?
+
+         * pyrax.volume.create expects a size in GB
+           The minimum SATA disk size on Rackspace is 100GB.
+           How do we enforce that?
```
exarkun: https://clusterhq.atlassian.net/browse/FLOC-1579

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,114 @@
...
+           And what (if any) errors should we raise if the user requests something smaller?
```
exarkun: Some useful, informative exception should be raised.  We won't be able to do anything but log it for now.  Later, the HTTP API should prevent bad values from getting this far.  And later still, there will be a way to discover convergence errors via the HTTP API.

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,114 @@
...
+           Is this an OpenStack limit or something specific to Rackspace?
```
exarkun: That's a pretty good question.  Can't set a backend-specific limit if the backend doesn't actually know what the limit is...  I suspect 100G is a Rackspace choice.  I see some indications here and there that Cinder is happy with, eg, 1GB volumes in other configurations.

Perhaps this will have to be set in a cluster-wide configuration file (perhaps alongside the OpenStack credentials information).

flocker/node/script.py
```diff
@@ -174,6 +174,7 @@ def flocker_dataset_agent_main():
     # Later, construction of this object can be moved into
     # DatasetAgentServiceFactory.get_service where various options passed on
     # the command line could alter what is created and how it is initialized.
+    # XXX: Implement the factory proposed above (here or in a blocking / follow up issue)
```
exarkun: We should talk about whether passing command-line arguments (or specifying a config file) on each node is the way to go for node configuration.  We'd previously discussed centralizing the configuration at the control service but maybe we should delay that work until later since it's not strictly necessary and we're likely to run out of time for nice-to-haves.

exarkun: The implication being, I think, that this branch also needs to include a new configuration file (with user-facing documentation) and changes to packaging code so that the agent gets told about that configuration file when it gets started.

Which ... maybe is blocked by packaging work making it possible to start the block device agent *at all* (instead of the ZFS agent)?

Which is probably blocked by the last of Itamar's work to split the ZFS and container agents into two different command line scripts...

exarkun: (And then I start thinking that centralizing the config at the control service is actually the easier path...)

exarkun:
Thanks.  As discussed, leave the configuration and script work for a later issue.  Just focus on the `IBlockDeviceAPI` work for this branch.

wallrj:
Ok. This isn't finished yet, but perhaps it's ready for some feedback.

You can run the tests using something like this:
`SKIP_TODO=TRUE OPENSTACK_API_USER=joe.bloggs.clusterhq OPENSTACK_API_KEY=xyz trial flocker.node.agents.functional.test_cinder`

Which will skip the IBlockDeviceAPI tests that can't yet work.

Still need to do some size conversion and make the authentication less RackSpace specific I suppose.

wallrj:
I've put this up for review, but it can't be merged until the new functional tests can be run automatically by buildbot on a Rackspace node.
 * https://clusterhq.atlassian.net/browse/FLOC-1749

To run the new tests manually, you'll need  the same config file used for running the acceptance tests:
 * http://doc-dev.clusterhq.com/gettinginvolved/acceptance-testing.html

Then  run trial as follows:
```
CLOUD_CONFIG_FILE=$PWD/acceptance.yml 
trial flocker.node.agents.functional
```

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    while True:
+        for listed_volume in volume_manager.list():
+            if listed_volume.id == expected_volume.id:
+                if listed_volume.status == 'available':
+                    return listed_volume
+
+
+@implementer(IBlockDeviceAPI)
+class CinderBlockDeviceAPI(object):
```
wallrj: Perhaps this should be a PRecord. It'd be nice to check the type of the supplied cluster_id and maybe check that the supplied volume_manager provides ICinderVolumeManager.

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,237 @@
...
+def wait_for_volume(volume_manager, expected_volume):
+    """
+    Wait for a ``Volume`` with the same ``id`` as ``expected_volume`` to be
+    listed and to have a ``status`` value of ``available``.
+
+    :param ICinderVolumeManager volume_manager: An API for listing volumes.
+    :param Volume expected_volume: The ``Volume`` to wait for.
+    :returns: The listed ``Volume`` that matches ``new_volume``.
+    """
+    while True:
```
itamarst: Infinite loops are a bad idea in production code... Either file follow up issue or add a timeout here.

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,237 @@
...
+    Create a Cinder API client capable of authenticating with Rackspace and
+    communicating with their Cinder API.
+
+    :param unicode username: A RackSpace API username.
+    :param unicode api_key: A RackSpace API key.
+    :param unicode region: A RackSpace region slug.
+    :return: A ``cinderclient.v1.clien.Client`` instance with a ``volumes``
+        attribute that conforms to ``ICinderVolumeManager``.
+    """
+    username = kwargs.pop('username')
```
itamarst: What's with the `**kwargs`?

flocker/node/agents/cinder.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+CINDER_CLIENT_FACTORIES = {
+    'rackspace': rackspace_cinder_client,
+}
+
+
+def cinder_api(cinder_client, cluster_id):
+    """
+    :param cinder_client: The Cinder API client whose ``volumes`` attribute
```
itamarst: Type?

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -0,0 +1,84 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
itamarst: I don't understand what this is, or how it differs from the other test module.

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -0,0 +1,84 @@
...
+        expected_items = set(expected_metadata.items())
+        actual_items = set(listed_volume.metadata.items())
+        missing_items = expected_items - actual_items
+        self.assertEqual(
+            set(), missing_items,
+            'Metadata {!r} does not contain the expected items {!r}'.format(
+                actual_items, expected_items
+            )
+        )
+    test_create_metadata_is_listed.todo = (
```
itamarst: Presumably we will soon run this against other OpenStack setups?

flocker/node/agents/functional/test_cinder_behaviour.py
itamarst: "Test actual real world behaviour of Cinder implementations, so we know when to update our workarounds"? or something

flocker/node/agents/functional/test_cinder.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+        )
+):
+    """
+    Interface adherence Tests for ``CinderBlockDeviceAPI``.
+    """
+
+
+class CinderBlockDeviceAPIImplementationTests(SynchronousTestCase):
+    """
```
itamarst: Is any post-test cleanup necessary?

flocker/node/agents/test/test_cinder.py
```diff
@@ -0,0 +1,16 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
itamarst: Doesn't seem like this module adds much.

itamarst:
Looks good other than comments above (many of which were answered once I read the code further, but possibly a docstring or two would help.) Merge when buildbot builder is up.

flocker/node/agents/cinder.py
wallrj: Because the `acceptance.yml` contains a `keyname` variable, which isn't useful here, but I wanted to just pass the whole dictionary in:
 * http://doc-dev.clusterhq.com/gettinginvolved/acceptance-testing.html#rackspace

flocker/node/agents/functional/test_cinder_behaviour.py
wallrj: They report that it's a `driver` bug.
I'm not sure if that's something specific to Rackspace but yes, I think I will want to run this against other OpenStacks. The client_from_environment thing should make that reasonably easy to do.

flocker/node/agents/functional/test_cinder.py
wallrj: Volumes created in the tests should be removed by the `TidyCinderClient._cleanup` which is registered as a test cleanup callback.

flocker/node/agents/test/test_cinder.py
wallrj: I removed the module and moved the comments to the functional test module.

wallrj:
So @itamarst ,

You're happy for this to get merged without the Rackspace builder?

It'll make the followup work easier and cut down on merge conflicts.

wallrj:
@itamarst 

Ignore the question above, I think we agreed on Friday that this can  merged despite the lack of automated testing which will be put in place as soon as @tomprince is available to help me set up the Rackspace build slave.

I've added the time limit to wait_for_volume and merged forward.

I'll merge this as soon as I see green(ish) build results.

flocker/node/agents/cinder.py
wallrj: I added a time limit.

wallrj:
OK. None of those build failures seem related to this branch so merging...

PR #1315, opened 2015-04-17
[FLOC-1317] Wait for containers to stop before removing

jongiddy:
Wait for container status to indicate container is not running before attempting to remove container


jongiddy:
For evaluation, the `flocker/installed-package/fedora-20` tests provide the clearest indication of an improvement.  These tests usually fail with multiple container and image removal errors, but pass with this patch.

flocker/node/_docker.py
```diff
@@ -508,6 +508,9 @@ def _remove():
                     break
 
             try:
+                while (self._client.inspect_container(container_name)
+                        ['State']['Running']):
+                    sleep(0.1)
```
exarkun: This fix seems like a work-around for a bug in either the Python Docker client or Docker itself.  

  * There should be a link to an upstream bug report about the misbehavior.
  * We should log iterations in this loop (perhaps including information about the container state - ideally, the complete output of `inspect_container`, I think) so we can see how often this happens, how long we end up having to wait, and so that if someone ever ends up stuck in an infinite loop here, they can see what's going on.
  * I see that Docker logs information about its attempts to stop containers.  It logs `Failed to send SIGTERM to the process, force killing` and `Container %v failed to exit within %d seconds of SIGTERM - using the force`.  Then ... it calls `WaitStop` with an indefinite timeout and ignores the return (which may be an error).  Considering these containers do seem to eventually stop that's probably not the case we're hitting (yet :cry:).
  * Since it's still not clear what's going on here, enabling docker-py logging (which I did in another branch for this issue) seems like a good idea.  Perhaps Docker is actually working fine but docker-py is hiding some other failure from us.  Being able to reproduce the failure on demand would be the best way to determine that.  Since it seems we can't do that, logging the HTTP traffic seems like the next best thing.

exarkun:

flocker/node/_docker.py
exarkun: On further investigation, I noticed two more things:

  * Many builders failed with a number of seemingly new Docker errors when building this branch.  It could be a coincidence or perhaps this change introduced that problem somehow (I can't really see how, but the point of this issue is that Docker interactions are subtle and hard to understand, I think...).  This probably merits further investigation.
  * It seems that containers with a restart policy interact poorly with the *stop* API.  If the container happens to not be running when the API call is processed, it fails with a 404.  Then Docker proceeds to restart the container according to its policy.  Our Docker layer interprets the 404 as meaning "it is safe to proceed to the *remove* step".  This is incorrect given the interaction with restart policies.  We probably need to do something different in the `NOT_FOUND` case in `_remove` in `DockerClient.remove`.

jongiddy: I don't see this as a bug in Docker. It is pretty clear that the stop operation puts the container into a stopping state and then returns to the client before the container is actually stopped.

As for additional failures, I think Monday was a bad day to test. All branches were giving additional errors, mainly to do with database and move failures in the acceptance tests.  That's the problem with having multiple intermittent errors. It can be difficult to isolate them.  However, tests before Monday were working fine for this branch. See http://build.clusterhq.com/builders/flocker%2Finstalled-package%2Ffedora-20/builds/1637 

Dealing with Docker restarts is a separate issue. Although I would expect a Docker stop operation to override any restart policy. Do you have example code to demonstrate this problem? 

I have added further logging to help debugging any future problems, e.g.  [FLOC-1729]

flocker/node/_docker.py
```diff
@@ -480,6 +485,23 @@ def exists(self, unit_name):
         container_name = self._to_container_name(unit_name)
         return deferToThread(self._blocking_exists, container_name)
 
+    def _blocking_container_runs(self, container_name):
```
adamtheturtle: Is it not possible even to patch something to test that this message is logged?

flocker/node/_docker.py
```diff
@@ -508,6 +530,12 @@ def _remove():
                     break
 
             try:
+                # The ``docker.Client.stop`` method puts the container into
+                # a stopping state, but the container cannot be removed
+                # until the container has actually stopped running.
+                while self._blocking_container_runs(container_name):
```
adamtheturtle: Put a comment like the one 10ish lines below saying that this is not covered by tests.

jongiddy: I'd disagree that it is not covered by tests. If you take this code out, you get more test failures. Remember, we're supposed to be testing external behaviour in the tests, not internal details.

There is actually no deterministic behaviour to be tested, due to FLOC-1729. After a `DockerClient.add`, we cannot know what state to expect, and after a `DockerClient.remove`, there is no container to be tested.

@wallrj had some ideas about using a fake Docker Server, but I'm not sure it's worth the effort at this stage.

flocker/node/_docker.py
exarkun: Probably not highest priority concern.  As long as the test suite runs the logging code without unexpected exceptions it's probably okay for the time being.

exarkun:
Thanks.

For what it's worth, I further convinced myself that the solution proposed here is correct - using this manual example (unfortunately I find the current test suite to be not particularly convincing since it's unclear exactly how likely the problem is to occur on any particular run):

```
exarkun@preon:~$ docker run --restart=always busybox:latest /bin/false                                                    
exarkun@preon:~$ docker ps
CONTAINER ID        IMAGE                       COMMAND               ... NAMES
899236f0e157        busybox:buildroot-2014.02   "/bin/false"        ... grave_hoover        
exarkun@preon:~$ docker stop grave_hoover
grave_hoover
exarkun@preon:~$ while docker start grave_hoover && sleep 0.7 && docker stop grave_hoover; do :; done
grave_hoover
grave_hoover
... (a long time passes) ...
Error response from daemon: Cannot stop container grave_hoover: no such process
FATA[0000] Error: failed to stop one or more containers 
exarkun@preon:~$ docker inspect grave_hoover
...
    "State": {
        "Error": "",
        "ExitCode": 1,
        "FinishedAt": "2015-04-21T11:17:03.298123139Z",
        "OOMKilled": false,
        "Paused": false,
        "Pid": 0,
        "Restarting": false,
        "Running": false,
        "StartedAt": "2015-04-21T11:17:03.236443967Z"
    },
...
```

This suggests that the Docker stop API does both fail and succeed in certain cases.  Inspecting the Docker logs suggests it's the (known, I think?) race condition between a process exiting and Docker trying to send the signal.  My comments earlier about the restart flag were probably incorrect and based on an incorrect interpretation of observations (so please disregard them).

Therefore, continuing after the failure and then polling the running state is probably a plausible solution.

The one uncertainty that remains, I think, is what the reported state will be immediately after this condition occurs (as opposed to several seconds later, which is how long it took me to type "docker inspect ..." in the above).  It seems possible some race may still remain which causes the logic of `_blocking_container_runs` to be incorrect.

> As for additional failures, I think Monday was a bad day to test. All branches were giving additional errors, mainly to do with database and move failures in the acceptance tests. That's the problem with having multiple intermittent errors. It can be difficult to isolate them. However, tests before Monday were working fine for this branch.

Fair enough.

> I don't see this as a bug in Docker. It is pretty clear that the stop operation puts the container into a stopping state and then returns to the client before the container is actually stopped.

The bug is that it returns a 404 **and** then later puts it into the stopped state, as far as I can tell.  Even if that's not the case and it just is a little bit slow to put the container into the stopped state, it's at least a documentation bug since it has taken us several months to track this down.

However, since we still don't know exactly the situation in which this occurs (that is, we don't have any hard evidence proving it one way or another), it will be difficult to file an upstream bug.

So please also introduce the HTTP request logging I asked for before in this branch, trigger the bug so that we see HTTP interactions happened, turn that into an upstream bug report (either "the API returns a 404 and also takes an action" or "the API encourages race conditions by returning before the state has changed" or "the documentation gives no indication that the stop API returns before the container has stopped sometimes" or perhaps the already-filed (I think) signal race condition I mentioned above or, who knows, maybe something else that we won't know until we see the logs), link to the bug report, and then re-submit.

itamarst:
Can we merge this forward and resubmit?

exarkun:
I think it needs more than just a merge forward before it can be resubmitted (see my comments above).

itamarst:
Looks like Jon added logging?

jongiddy:
I've redirected the logging to stderr, so it is easy to see the events in Buildbot (since I have not been able to re-create this locally).  I will try to create an issue similar to https://github.com/docker/docker/issues/9379 during my non-pairing time, but I suggest that it would be useful to merge now, to get Buildbot closer to all-green, so we can see when new branches are actually breaking.

jongiddy:
Link added to Docker issue at https://github.com/docker/docker/issues/13088

flocker/node/_docker.py
```diff
@@ -29,6 +30,9 @@
     RestartNever, RestartAlways, RestartOnFailure, pset_field, pvector_field)
 
 
+logger = logging.getLogger(__name__)
```
exarkun: I don't think we've hooked the stdlib logging system up to anything - so these log messages don't go anywhere. :/  I double-checked on buildbot and I didn't see the messages anywhere in `test.log` or on stdout.

`eliot.Message.new(keyword=arguments).write()` is probably the easiest thing to do that will actually send the logs somewhere.  `with eliot.start_action(action_type=stuff, keyword=arguments):` is another pretty easy thing to do if you want parent/child relationships or success/failure results in the logs.

Like the `logger`-based logging you added, this stuff doesn't need test coverage right now.

jongiddy: Since the problem only appears on Buildbot, I had the logging going to stderr, since that was the only way to see the output.  Should I revert the changes to keep that logging?  It adds a lot of noise to the test output, but not sure where else it could go, and still be visible on Buildbot.

jongiddy: See the change from the last two commits (mainly 88f9437) to see what was removed - the nice thing was that this was in the test file, so debugging output did not display during non-test operation.

exarkun: Ah, I see.  Hm.  Ideally this stuff would continue to show up somewhere.  My motivation is that I want some evidence that when we hit this case the work-around being applied here is actually addressing the problem.  Messages in a log file somewhere showing us that we hit it would make good evidence.

As far as stderr being the only place where the log output is visible on buildbot... `test.log` is also available, and I think in general anything that we manage to get into `test.log` will also end up in the application-specific log file under real usage (`/var/log/flocker/whatever.log` for example).

So, I agree that stderr is not a great choice for the "where".  I think this means the ideal is indeed to switch over to Eliot and I think that's probably not so much work that we'll benefit from skipping that work now.  If that seems reasonable, please go ahead with that and then merge.

adamtheturtle:
I could be wrong but check that the following failure is not introduced by this branch before merging:

```
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/trial/util.py", line 296, in _runSequentially
    thing = yield d
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/threadpool.py", line 196, in _worker
    result = context.call(ctx, function, *args, **kwargs)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/node/_docker.py", line 554, in _remove
    while self._blocking_container_runs(container_name):
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/node/_docker.py", line 507, in _blocking_container_runs
    result = self._client.inspect_container(container_name)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/node/functional/test_docker.py", line 844, in error
    raise APIError("", Response())
docker.errors.APIError: <APIError instance at 0x7f090406ff58 with str error:
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/reflect.py", line 443, in safe_str
    return str(o)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/docker/errors.py", line 38, in __str__
    self.response.status_code, self.response.reason)
AttributeError: 'Response' object has no attribute 'reason'
>

flocker.node.functional.test_docker.DockerClientTests.test_list_error_inspecting_container
```

exarkun:
Thanks.  Merge when buildbot shows this has fixed the problem (which won't be green, of course).

PR #1316, opened 2015-04-17
Turn any connection attempt error into a skip reason

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1658

### DIFF
--- a/flocker/node/testtools.py
+++ b/flocker/node/testtools.py
@@ -4,11 +4,11 @@
 Testing utilities for ``flocker.node``.
 """
 
-import errno
 import os
 import pwd
 import socket
-from unittest import skipUnless
+from unittest import skipIf
+from contextlib import closing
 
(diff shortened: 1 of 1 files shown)


flocker/node/testtools.py
```diff
@@ -33,27 +33,25 @@ def docker_accessible():
 
     This may address https://clusterhq.atlassian.net/browse/FLOC-85.
 
-    :return: ``True`` if the current user has permission to connect, else
-        ``False``.
+    :return: A ``unicode`` string describing the reason Docker is not
+        accessible or ``None`` if it appears to be accessible.
     """
```
jongiddy: Python 2.7 `strerror` returns a `str`, not a `unicode`.

jongiddy:
Apart from the incorrect type in the docstring, this is good to be merged.

PR #1318, opened 2015-04-17
[FLOC-1720] Fix intermittent buildbot failures

itamarst:


adamtheturtle:
There are failing tests like:

```
[ERROR]
  File "/srv/buildslave/flocker-acceptance-rackspace-fedora-20-distro-testing/venv/lib/python2.7/site-packages/flocker/acceptance/test_moving_data.py", line 54, in test_moving_data
    u"external", 27018]))
  File "/srv/buildslave/flocker-acceptance-rackspace-fedora-20-distro-testing/venv/lib/python2.7/site-packages/pyrsistent.py", line 934, in transform
    return _transform(self, transformations)
  File "/srv/buildslave/flocker-acceptance-rackspace-fedora-20-distro-testing/venv/lib/python2.7/site-packages/pyrsistent.py", line 2635, in _transform
    for path, command in _chunks(transformations, 2):
exceptions.ValueError: need more than 1 value to unpack

flocker.acceptance.test_moving_data.MovingDataTests.test_moving_data
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
DelayedCalls: (set twisted.internet.base.DelayedCall.debug = True to debug)
<DelayedCall 0x45ace60 [29.9972560406s] called=0 cancelled=0 Client.failIfNotConnected(TimeoutError('',))>

flocker.acceptance.test_moving_data.MovingDataTests.test_moving_data
[ERROR]
Failure: twisted.trial.util.DirtyReactorAggregateError: Reactor was unclean.
Selectables:
<<class 'twisted.internet.tcp.Client'> to ('104.130.30.231', 4523) at 45a3590>

flocker.acceptance.test_moving_data.MovingDataTests.test_moving_data
```

tomprince:
This is due to `transform` taking 2 arguments, not one. I've pushed the trivial fix.

itamarst:
All acceptance tests are green, for once.

adamtheturtle:
Change the docstrings which refer to parts of the documentation to explain the differences between the docs and the tests.

flocker/acceptance/test_environment_variables.py
```diff
@@ -166,19 +170,13 @@ def mysql_can_connect():
...
                         )
-                    except OperationalError as e:
-                        # PyMySQL doesn't provided a structured way to
-                        # get this.
-                        # https://github.com/PyMySQL/PyMySQL/issues/274
-                        if "Connection refused" in str(e):
-                            return False
-                        else:
-                            raise
+                    except Error:
```
adamtheturtle: Apart from the error mentioned in the JIRA issue, are and "Connection refused" are there any other errors which we know we are OK with? This seems like it is more likely to cause a loop which continues until the timeout is hit if there is a problem.

flocker/acceptance/test_environment_variables.py
```diff
@@ -199,7 +197,7 @@ def test_environment_variable_used(self):
             user=b'root',
             passwd=MYSQL_PASSWORD,
         )
-        # No assetion, since _get_mysql_connection will fire with a failure,
+        # No assertion, since _get_mysql_connection will fire with a failure,
```
adamtheturtle: Is this still the case now that there is `except Error: return False`?

flocker/acceptance/test_environment_variables.py
```diff
@@ -247,12 +245,15 @@ def get_mysql_node_2(ignored):
             Move MySQL to ``node_2`` and return a ``Deferred`` which fires
             with a connection to the previously created database on ``node_2``.
             """
+            # Listen on different port so it's clear we're connecting to
```
adamtheturtle: Is this absolutely necessary? Could it not be shown just as well with `assert_expected_deployment`? In fact isn't that the point of having `test_moving_mysql` which now seems redundant?

adamtheturtle:
Thanks @itamarst please address the comments and resubmit.

flocker/acceptance/test_environment_variables.py
itamarst: Being disconnected half way through connection, for example.

flocker/acceptance/test_environment_variables.py
itamarst: Yeah, it'll time out.

flocker/acceptance/test_environment_variables.py
itamarst: `assert_expected_deployment` just checks what the control service *thinks* the current state is. That may not be the same as actual state, which is why we want to test things directly.

flocker/acceptance/test_environment_variables.py
itamarst: This is in fact kinda fragile and brittle, and ideally we'll replace them with tests that don't involve arbitrary complexity like mysql clients. But given we're not replacing them immediately, I think this is probably best we can do.

flocker/acceptance/test_environment_variables.py
adamtheturtle: That's a shame. IIRC it used to use Docker and was at some point switched to use internal Flocker code, which we said we could trust enough.

flocker/acceptance/test_environment_variables.py
adamtheturtle: Is it possible to isolate that particular error and `raise`?

adamtheturtle:
Thanks @itamarst , address the comment about maybe `raise`ing if the details are incorrect and merge.

flocker/acceptance/test_environment_variables.py
exarkun: I've seen the tests fail with 3 or 4 other MySQL errors too. :/

adamtheturtle: Maybe at this point the error can be logged and then we can check the logs if it becomes a problem that we're returning `False` on all errors.

flocker/acceptance/test_environment_variables.py
itamarst: Well, there's *two* things we want to assert:

1. External behaviour from point of view of user (data moved, ports are open, etc.) is as expected. We do this by testing it directly.
2. Flocker's understanding of the state of the world is correct. We do this by looking at state in control service.

Checking what Docker thinks the state of the world is doesn't really address either requirement.

flocker/acceptance/test_environment_variables.py
itamarst: It happens externally, as part of trial.

PR #1319, opened 2015-04-17
[FLOC-1554] Split container agent into its own script and service

itamarst:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1554


admin/package-files/systemd/flocker-agent.service
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description=Flocker Node Agent
+Description=Flocker ZFS Agent
```
wallrj: It'd be clearer if this file was renamed to `flocker-zfs-agent.service` or `flocker-dataset-agent.service` but we can do that when we make the dataset agent backend pluggable.

flocker/node/_deploy.py
```diff
@@ -932,50 +932,3 @@ def find_dataset_changes(hostname, current_state, desired_state):
                    if dataset.deleted)
     return DatasetChanges(going=going, coming=coming, deleting=deleting,
                           creating=creating, resizing=resizing)
-
-
-class P2PNodeDeployer(_OldToNewDeployer):
```
wallrj: Can we also ditch the `_OldToNewDeployer` base class?

flocker/node/functional/test_deploy.py
```diff
@@ -24,6 +24,49 @@
 from ...route import make_memory_network
 
 
+class P2PNodeDeployer(object):
```
wallrj: Why not just update the tests that use this (now deleted) deployer? 
If only to limit required test changes and the size of the branch, then perhaps there should be a linked followup issue to eventually fix those tests.

wallrj:
I started reviewing this and then realised that the node state endpoint was actually from a (now) merged branch.

 * So I've merged this forward.
 * The changes in this branch all look good
 * The changed unit and functional tests work locally

Points:
 * I don't understand why we're keeping the P2PNodeDeployer class as a test helper rather than updating the tests that use it.
 * Maybe update the `docs/advanced/api.rst` documentation that says "start the agent" 

Please address or answer the points above and then merge when buildbot has run the acceptance tests.

flocker/node/functional/test_deploy.py
itamarst: Yeah, expediency basically. I'll file an issue.

PR #1320, opened 2015-04-20
[FLOC-1722] Remove unused code from acceptance tests.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1722


wallrj:
Thanks @tomprince 

* Only thing I notice is that there's still a reference to elasticsearch in setup.py - remove that if you agree it's no longer needed.

Then please merge.

PR #1322, opened 2015-04-20
Automatically log IStateChanges

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1591

This is the last step.  Logging is added to the new `run_state_change` API and existing `IStateChange` implementations are augmented with the necessary Eliot logging information.

### DIFF
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -75,6 +75,9 @@ def run_state_change(change, deployer):
     return change.run(deployer)
 
 
+# run_state_change doesn't use the IStateChange implementation provided by
+# _InParallel and _Sequentially but those types provide it anyway because
+# certain other areas of the test suite depend on it.
 @implementer(IStateChange)
 class _InParallel(PRecord):
     changes = field(type=PVector, factory=pvector, mandatory=True)
@@ -92,6 +95,7 @@ def in_parallel(changes):
     return _InParallel(changes=changes)
 
 
+# See comment above _InParallel.
 @implementer(IStateChange)
 class _Sequentially(PRecord):
     changes = field(type=PVector, factory=pvector, mandatory=True)
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -62,6 +62,16 @@ def run_state_change(change, deployer):
 
     :return: ``Deferred`` firing when the change is done.
     """
+    if isinstance(change, _InParallel):
+        return gather_deferreds(list(
+            subchange.run(deployer) for subchange in change.changes
+        ))
+    if isinstance(change, _Sequentially):
+        d = succeed(None)
+        for subchange in change.changes:
+            d.addCallback(lambda _, change=subchange: change.run(deployer))
+        return d
+
     return change.run(deployer)
 
 
@@ -70,8 +80,7 @@ class _InParallel(PRecord):
     changes = field(type=PVector, factory=pvector, mandatory=True)
 
     def run(self, deployer):
-        return gather_deferreds(
-            [change.run(deployer) for change in self.changes])
+        return run_state_change(self, deployer)
 
 
 def in_parallel(changes):
@@ -88,10 +97,7 @@ class _Sequentially(PRecord):
     changes = field(type=PVector, factory=pvector, mandatory=True)
 
     def run(self, deployer):
-        d = succeed(None)
-        for change in self.changes:
-            d.addCallback(lambda _, change=change: change.run(deployer))
-        return d
+        return run_state_change(self, deployer)
 
 
 def sequentially(changes):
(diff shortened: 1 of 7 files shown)


flocker/node/_change.py
```diff
@@ -77,7 +84,10 @@ def run_state_change(change, deployer):
             )
         return d
 
-    return change.run(deployer)
+    with change.eliot_action.context():
```
itamarst: I wonder if it's worth having a utility function `eliot.twisted.run()` that does that? If so, file an issue.

flocker/node/_deploy.py
```diff
@@ -143,6 +146,16 @@ class StartApplication(object):
     :ivar NodeState node_state: The state of the node the ``Application``
         is running on.
     """
+    application = field(type=Application, mandatory=True)
+    node_state = field(type=NodeState, mandatory=True)
+
+    @property
```
itamarst: Ick. An attribute that is *required* to be side-effect-y isn't great. I suggest `IStateChange.eliot_action` be a method, not an attribute.

flocker/node/_change.py
```diff
@@ -27,6 +29,11 @@ class IStateChange(Interface):
     """
     An operation that changes local state.
     """
+    eliot_action = Attribute(
```
itamarst: The doc isn't accurate; usage indicates this is supposed to return a *new* *started* action. So really this should be a method, not an attribute.

flocker/node/_deploy.py
```diff
@@ -131,9 +131,12 @@ def calculate_changes(self, configuration, cluster_state):
             local_state, configuration, cluster_state)
 
 
+def _eliot_system(part):
+    return u"flocker:p2pdeployer:" + part
```
itamarst: Not really accurate anymore (some of this is container stuff) but I guess we can fix that if/when we move container stuff and p2p stuff into their own modules.

flocker/node/test/test_change.py
```diff
@@ -17,9 +21,41 @@
...
+    ]
+    change = outer_factory(changes=subchanges)
+    run_state_change(change, DEPLOYER)
+    case.assertEqual(
+        (True, DEPLOYER),
+        (inner_action.called, inner_action.deployer)
+    )
 
 
 def _test_nested_change(case, outer_factory, inner_factory):
```
itamarst: Probably want to delete this one?

flocker/node/test/test_deploy.py
```diff
@@ -98,26 +139,51 @@ def test_network_override(self):
         )
 
 
+_DATASET_A = Dataset(dataset_id=unicode(uuid4()))
+_DATASET_B = Dataset(dataset_id=unicode(uuid4()))
+
+
 StartApplicationIStateChangeTests = make_istatechange_tests(
     StartApplication,
-    dict(application=1, node_state=NodeState(hostname="node1.example.com")),
```
itamarst: Hooray.

itamarst:
Address comments above, merge when buildbot is happy.

flocker/node/_change.py
exarkun: Yea, seems reasonable.  It's a lot of boilerplate for "run this async function in the context of this action".

Filed https://github.com/ClusterHQ/eliot/issues/158

flocker/node/_deploy.py
exarkun: I left a comment mentioning that these properties use `start_action` essentially because of laziness on my part.  For that same reason, I'm leaving `eliot_action` as an attribute.  When we have time, someone can define an `ActionType` for each of these (at least, for each that doesn't get deleted!) and since eventually I think Eliot is planning to remove the logger argument required to instantiate an `ActionType`, it will be possible to turn these into normal attributes rather than properties (left a comment to that effect as well).

flocker/node/_change.py
exarkun: Oh.  Uh.  That's not how I thought Eliot worked.

Now I see that `ActionType.__call__` does a `startAction` before returning.  So:

```
X = ActionType(...)
y = X(...)
```

Actually starts the action on the second line.  This basically invalidates everything I said above about getting rid of properties.  Also... blech?  I think?  This seems like a surprising way for it to work.

So now I agree that this should be a method, not a property.  But I still don't feel like fixing it.  Maybe I will in the morning... except then I'll have to get in the three hour build queue... So in the interest of not wasting that time tomorrow, I'm going to merge this as-is (which is what I was going to do anyway, except now I know it's wrong instead of thinking maybe it's right).  Someone can convert it to a method later.

flocker/node/_deploy.py
exarkun: Yea, seems good enough for debugging purposes for now.

flocker/node/test/test_change.py
exarkun: Weee.  Fun times with git merge.

PR #1323, opened 2015-04-21
[FLOC-1717] Remove DigitalOcean support.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1717

We may add support back, after 1.0.


docs/spelling_wordlist.txt
```diff
@@ -17,7 +17,6 @@ dataset
 datasets
 Deis
 devel
-DigitalOcean
```
wallrj: We still mention "DigitalOcean" (now unlinked) in whatsnew.rst, so this probably ought to stay.

wallrj:
Thanks @tomprince 

This looks good.

I checked for unexpected / lingering "digital" "ocean" "pyocean" references and there weren't any.

Only thing I noticed is that  DigitalOcean probably ought to still be in the custom dict, although the  spellchecker  passed on my machine regardless....not sure why / how.

Please merge.

PR #1324, opened 2015-04-21
S3 packages floc 1716

adamtheturtle:
The script can be tested by running:

```
$ admin/publish-artifacts --target=clusterhq-archive-testing --flocker-version=0.4.0
```


admin/test/test_release.py
```diff
@@ -1508,6 +1499,110 @@ def get_dispatcher(self):
...
+             'python/Flocker-0.3.0.tar.gz'])
+
+    def test_setuptools_version_requirement(self):
+        """
+        When setuptools' version is not 3.6, a ValueError is raised.
+        """
+        self.patch(
+            release, 'setuptools_version', '15.1')
+        self.assertRaises(
+            ValueError,
```
adamtheturtle: This raises ``admin.release.IncorrectSetuptoolsVersion``

admin/test/test_release.py
```diff
@@ -1508,6 +1499,110 @@ def get_dispatcher(self):
...
+
+        with open(os.devnull, "w") as discard:
+            sync_perform(
+                ComposedDispatcher(dispatchers),
+                upload_python_packages(
+                    scratch_directory=self.scratch_directory,
+                    target_bucket=self.target_bucket,
+                    top_level=self.top_level,
+                    output=discard,
+                    error=discard,
```
adamtheturtle: `error` should go to `sys.stderr`

adamtheturtle: Actually this goes to `trial`'s output so I haven't done it.

admin/release.py
```diff
@@ -448,16 +470,89 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+    Create an index file for pip.
 
-    :param list args: The arguments passed to the script.
+    :param FilePath scratch_directory: Temporary directory to create index in.
+    :param list packages: List of bytes, filenames of packages to be in the
+        index.
+    """
+    index_file = scratch_directory.child('index')
+    with index_file.open('w') as f:
+        f.write('<!--This is an index for pip-->\n')
```
tomprince: No reason for this to be a HTML comment. If somebody navigates to the page, they'll see this.

admin/release.py
```diff
@@ -468,30 +563,52 @@ def publish_rpms_main(args, base_path, top_level):
     except UsageError as e:
         sys.stderr.write("%s: %s\n" % (base_path.basename(), e))
         raise SystemExit(1)
+    except NotARelease:
+        sys.stderr.write("%s: Can't publish artifacts for a non-release.\n"
+                         % (base_path.basename(),))
+        raise SystemExit(1)
+    except DocumentationRelease:
+        # This should not raise 1 because the release process should continue
+        # for a documentation release without publishing artifacts.
```
tomprince: At a first glance, I'm not convinced this is sensible reasoning.

admin/test/test_release.py
```diff
@@ -1612,6 +1708,120 @@ def test_branch_created_from_base(self):
...
+        expected = dedent(
+            "<!--This is an index for pip-->\n"
+            '<a href="Flocker-0.3.0-py2-none-any.whl">Flocker-0.3.0-py2-none-any.whl</a><br/>\n'  # noqa
+            '<a href="Flocker-0.3.1-py2-none-any.whl">Flocker-0.3.1-py2-none-any.whl</a><br/>\n'  # noqa
+        )
+        self.assertEqual(expected, index.getContent())
+
+    def test_only_wheels_included(self):
+        """
+        The pip index file created only references wheel files.
```
tomprince: This should at least also reference the sdists. But I'm not sure there is a good reason to exclude anything other than `index`.

tomprince: That way, something like

```
pip install --no-use-wheel -f https://clusterhq-archive-testing.s3.amazonaws.com/python/index flocker==0.4.0-657-gdfb9f48-dirty
```

will work.

admin/release.py
```diff
@@ -448,16 +470,89 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+    packages = yield Effect(
+        ListS3Keys(bucket=target_bucket,
+                   prefix='python/'))
+
+    index_path = create_pip_index(
+        scratch_directory=scratch_directory,
+        packages=packages)
+
+    yield Effect(
+        UploadToS3(
```
tomprince: We need to ensure that the content type is `text/html`, otherwise pip will ignore it.

docs/gettingstarted/index.rst
```diff
@@ -99,7 +99,7 @@ Getting started with Flocker
             you@laptop:~$ sudo apt-get update && sudo apt-get install -y gcc python2.7 python-virtualenv python2.7-dev && \
               virtualenv flocker-tutorial && \
               flocker-tutorial/bin/pip install --upgrade pip && \
-              flocker-tutorial/bin/pip install --quiet https://storage.googleapis.com/archive.clusterhq.com/downloads/flocker/Flocker-|latest-installable|-py2-none-any.whl && source flocker-tutorial/bin/activate
+              flocker-tutorial/bin/pip install --quiet https://s3.amazonaws.com/clusterhq-archive/python/Flocker-|latest-installable|-py2-none-any.whl && source flocker-tutorial/bin/activate
```
tomprince: Testing the equivalent URL on `clusterhq-archive-testing` doesn't appear to work. It looks like https://clusterhq-archive-testing.s3.amazonaws.com/ is needed instead.

tomprince:
A few minor comments, then merge (once buildbot looks happy-ish).

docs/gettingstarted/index.rst
adamtheturtle: Ok I will change this here but it seems that similar URLs to this are used in other places. Do you think that it is worth looking into them before the next release?

admin/release.py
adamtheturtle: In the testing archive this is `text/html` but the easiest way to do this is to change the file to be called `index.html`.

setup.py
```diff
@@ -183,7 +187,7 @@
         "release": [
             "gitpython==1.0.0",
             "gsutil",
-            "wheel",
+            "wheel==0.24.0",
```
exarkun: Why this version specifically?  A comment would probably be nice.

adamtheturtle: This is the version at the time of writing and the version used by the tests, which seems to be the reason for most / many of our dependency pinned versions.

admin/release.py
```diff
@@ -448,16 +470,89 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+    """
+    index_file = scratch_directory.child('index.html')
+    with index_file.open('w') as f:
+        f.write('This is an index for pip\n')
+        for package in packages:
+            if package != 'index.html':
+                f.write('<a href={destination}>{title}</a><br/>\n'.format(
+                    destination=quoteattr(package),
+                    title=escape(package)))
+    return index_file
```
exarkun: Argh manual HTML generation?  Argh.  This is not a cool thing. :cry:

adamtheturtle:
This had some merge problems and because I didn't want to `force` and risk losing history I created a new PR at https://github.com/ClusterHQ/flocker/pull/1343.

adamtheturtle:
@exarkun What would you prefer for generating html? Twisted web tempting, https://pypi.python.org/pypi/html/, other?

exarkun:
> @exarkun What would you prefer for generating html? Twisted web templating, https://pypi.python.org/pypi/html/, other?

@adamtheturtle Just about anything that does it correctly instead of with string mashing should be fine.  Considering Twisted is already a dependeny and the html being generated is pretty simple, `twisted.web.template` might be the way to go.

PR #1326, opened 2015-04-21
[FLOC-1726] Flocker-deploy no longer does SSH

itamarst:


exarkun:
Thanks.  Great all around.  Please merge once buildbot is green (or at least not red in a way that suggests this has made things worse, particularly the acceptance tests).

PR #1327, opened 2015-04-21
[FLOC-1727] [FLOC-1631] Node state and configuration have a UUID

itamarst:


flocker/control/_persistence.py
```diff
@@ -37,6 +38,8 @@ def default(self, obj):
         elif isinstance(obj, FilePath):
             return {_CLASS_MARKER: u"FilePath",
                     u"path": obj.path.decode("utf-8")}
+        # elif isinstance(obj, UUID):
+        #    return unicode(obj)
```
exarkun: Probably need something like `_CLASS_MARKER` so it round-trips?

flocker/node/script.py
```diff
@@ -50,14 +52,24 @@ class ZFSAgentOptions(Options):
          "The port on the control service to connect to.", int],
```
exarkun: Don't forget to update the `synopsis`

flocker/node/script.py
```diff
@@ -50,14 +52,24 @@ class ZFSAgentOptions(Options):
...
+    sock.connect_ex((host, port))
+    return sock.getaddr()[1]  # XXX approximately
+
+
+def _get_node_uuid(ip):
+    # Soon we'll extract this from TLS certificate for node.  Until then
+    # we'll just do a temporary hack of some sort.  In particular, by
+    # using a deterministic IP->UUID mapping we can separate out porting
+    # the REST API to new UUID model into a separate issue.
+    pass #return UUID(md5(ip))
```
exarkun: I thought it might be handy if we could do `UUID(node=ip, everything_else=0)` so the mapping could be easily reversible but it seems like the `node` component doesn't map easily to an IP address and there's no easy way to do construct a `UUID` instance from components with the `uuid` module anyway.

exarkun: This strategy will fail if someone has nodes deployed on different private networks with overlapping address ranges and two nodes get assigned the same local address.  But I guess that doesn't matter because this is just a transitional solution.

flocker/node/script.py
```diff
@@ -50,14 +52,24 @@ class ZFSAgentOptions(Options):
...
-        # https://clusterhq.atlassian.net/browse/FLOC-1381 for solution.
-        self["hostname"] = unicode(hostname, "ascii")
+    def parseArgs(self, host):
         self["destination-host"] = unicode(host, "ascii")
 
 
+def _get_external_ip(host, port):
+    sock = socket()
+    sock.connect_ex((host, port))
+    return sock.getaddr()[1]  # XXX approximately
```
exarkun: This is temporary too, right?  Does the overall plan here involve getting rid of the need for clients to determine their own address (replacing it with control service-supplied information on `NodeState`)?

flocker/provision/_install.py
```diff
@@ -434,7 +431,6 @@ def configure_cluster(control_node, agent_nodes):
                     username='root',
                     address=node,
                     commands=task_enable_flocker_agent(
-                        node_name=node,
                         control_node=control_node,
                     ),
                 ),
```
exarkun: Eyes glazing over...

flocker/control/_model.py
```diff
@@ -573,6 +575,7 @@ def __invariant__(self):
                 return (False, '%r is not correct key for %r' % (key, value))
         return (True, "")
 
+    uuid = field(type=UUID, factory=UUID, mandatory=True)
```
exarkun: I kinda hate `factory`.  Code instantiating this type should just pass a `UUID` instance.  Accidentally passing a byte string of the right length shouldn't pass silently.

flocker/control/_model.py
```diff
@@ -552,7 +553,8 @@ class NodeState(PRecord):
     """
     The current state of a node.
 
-    :ivar unicode hostname: The hostname of the node.
+    :ivar UUID uuid: The node's UUID.
+    :ivar unicode hostname: The IP of the node.
```
exarkun: Do we have an issue for using a type from `ipaddr` for this?

flocker/node/_deploy.py
```diff
@@ -58,7 +58,8 @@ class IDeployer(Interface):
     :ivar unicode hostname: The hostname of the node this deployer is
         managing.
     """
-    hostname = Attribute("The hostname for this node.")
+    uuid = Attribute("The UUID of thise node, a ``UUID`` instance.")
+    hostname = Attribute("The public IP address of this node.")
```
exarkun: I don't think I see why the interface needs the `hostname` attribute

itamarst: Possibly not necessary any more, I'll remove it if so.

exarkun:
Thanks.

I have some difficulty with this design since the goal it seems to achieve, for the most part, is "some internals are changed around in some ways".  It seems like it's probably going in the right direction but a really good evaluation involves imagining the design for the rest of FLOC-1381 (which I have not tried to do) and making sure it follows from this work.

Something I did a while ago (only once, as I recall) was to do the entire design as part of the first sub-task.  Then I annotated the design with issue references indicating when they'd actually be implemented.  master ended up with design notes in it by the time that first issue was resolved and the follow-on tasks gradually converted them into implementations.  I don't know how that approach sounds to you but from a review perspective it seems a bit easier (even from a design perspective it seems like it should produce better results, since you actually get to do the whole design and find mistakes that might have been made in the parts that you eventually implement first).

Do with that what you will.

Thinking forward a little bit ... Once the UUID is in the certificate, I'd expect the control service to extract it and merge it into its `NodeState` - instead of trusting the nodes to send the correct value.  I expect the same is the case for the node's public address.  So maybe this design's heavy node-side focus isn't the best step to take?

Address that and inline comments/questions and resubmit.

flocker/control/_persistence.py
itamarst: I was planning on relying on the `field` factory, but could do so if necessary, yeah.

flocker/node/script.py
itamarst: Yeah, hopefully only for a day or two at most. Might be another temporary expedient after that, or maybe we'll be lucky and we can switch straight to getting it out of CA.

flocker/node/script.py
itamarst: Hopefully, yeah. At the moment calculation logic is still using `hostname` though so we still need to get it somehow.

itamarst:
The agent needs to know its node's UUID so that its deployer can do the calculation. So, yes, control service shouldn't trust the result but the agent should still be figuring out its own UUID rather than doing a roundtrip with control service just to get some information it already needs.

itamarst:
A full design might be better... but the problem is there's interim steps I'm taking so that I *don't* have to do full design in one go. E.g. IP->UUID mapping. So I don't really know how to represent that in One Big Branch.

flocker/control/_model.py
itamarst: Yes. Lots and lots of existing code :)

itamarst: (Which I don't want to touch in this branch).

itamarst: Notice that all I'm doing is changing the docstring to be more accurate, no actual code changes will result.

itamarst:
I've updated this way with backwards compat hack that should allow doing this with hopefully minimal changes to existing tests, so I've combined implementation of FLOC-1636 into this branch.

exarkun:
> so I've combined implementation of FLOC-1636 into this branch

Did you mean a different number?  https://clusterhq.atlassian.net/browse/FLOC-1636 looks unrelated.

flocker/control/_model.py
```diff
@@ -547,12 +555,28 @@ def update_cluster_state(cluster_state):
...
+    Deterministically convert IP to UUID.
+
+    This is intended for interim use and backwards compatibility for
+    existing tests. It should not be hit in production code paths.
+
+    :param unicode ip: An IP.
+    """
+    warn(DeprecationWarning, "UUID is required, this is for backwards compat "
+         "with existing tests only. If you see this in production code that's "
+         "a bug")
```
exarkun: Ha ha.  As if you can **ever** see a `DeprecationWarning`.

itamarst:
FLOC-1631, sorry.

exarkun:
Thanks.  This seems reasonable.  Please proceed.

flocker/control/_model.py
```diff
@@ -376,7 +379,19 @@ def __invariant__(self):
...
 
+    def __new__(cls, **kwargs):
+        # PRecord does some crazy stuff, thus _precord_buckets; see
+        # PRecord.__new__.
+        if "uuid" not in kwargs and "_precord_buckets" not in kwargs:
+            warn("UUID is required, this is for backwards compat with existing"
+                 " tests only. If you see this in production code that's "
+                 "a bug.", DeprecationWarning, stacklevel=2)
+            kwargs["uuid"] = ip_to_uuid(kwargs["hostname"])
+        return PRecord.__new__(cls, **kwargs)
```
exarkun: :question: 
Possibly could do this by handling an `InvariantException` from `PRecord.__new__` and then re-trying with the updated keyword arguments.  Possibly better if it isolates us from implementation details like `_precord_buckets` (and avoids a change in pyrsistent breaking our assumptions here).  On the other hand, pinned versions!  **Hooray**.

flocker/control/_model.py
```diff
@@ -463,7 +478,7 @@ def move_application(self, application, target_node):
                 if container.name == application.name:
                     # We only need to perform a move if the node currently
                     # hosting the container is not the node it's moving to.
-                    if node.hostname != target_node.hostname:
+                    if node.uuid != target_node.uuid:
```
exarkun: :question:

A `node.same_node_as(target_node)` API would be a good way to isolate this comparison logic and avoid duplicating it throughout the codebase.

flocker/control/_model.py
exarkun: I just meant an issue in JIRA. :wink:

flocker/control/_model.py
```diff
@@ -573,6 +603,17 @@ def __invariant__(self):
...
 
+    def __new__(cls, **kwargs):
+        # PRecord does some crazy stuff, thus _precord_buckets; see
+        # PRecord.__new__.
+        if "uuid" not in kwargs and "_precord_buckets" not in kwargs:
+            warn("UUID is required, this is for backwards compat with existing"
+                 " tests only. If you see this in production code that's "
+                 "a bug.", DeprecationWarning, stacklevel=2)
+            kwargs["uuid"] = ip_to_uuid(kwargs["hostname"])
+        return PRecord.__new__(cls, **kwargs)
```
exarkun: Ah, this again.  Whether you decide to approach this detection differently or not, please refactor this to avoid the duplication.

flocker/control/test/test_model.py
```diff
@@ -25,6 +25,30 @@
 )
 
 
+class IpToUUIDTests(SynchronousTestCase):
```
exarkun: `IPToUUIDTests` probably

exarkun: (Or `InternetProtocol` if that's too ugly :wink:)

flocker/control/test/test_model.py
```diff
@@ -205,6 +249,24 @@ class NodeStateTests(SynchronousTestCase):
...
+        self.assertIsInstance(node.uuid, UUID)
+
+    def test_uuid(self):
+        """
+        ``NodeState`` can be created with a UUID.
+        """
+        uuid = uuid4()
+        node = NodeState(hostname=u'1.2.3.4', uuid=uuid)
+        self.assertEqual(node.uuid, uuid)
+
```
exarkun: :question: 

It would be nice if the two variations of these two tests, one for `Node` and one for `NodeState`, were shared instead of duplicated.  Less code to maintain, less to clean up later, etc.

flocker/control/test/test_model.py
```diff
@@ -122,7 +146,27 @@ class NodeInitTests(make_with_init_tests(
...
+    Note that hostname will no longer be required and should therefore be
+    removed from these tests in FLOC-1733.
     """
+    def test_no_uuid(self):
+        """
+        If no UUID is given, a UUID is generated from the hostname.
+
+        This is done for backwards compatibility with existing tests, and
+        should be removed enveutally.
```
exarkun: eventually

flocker/control/test/test_model.py
```diff
@@ -205,6 +249,24 @@ class NodeStateTests(SynchronousTestCase):
     """
     Tests for ``NodeState``.
     """
+    def test_no_uuid(self):
+        """
+        If no UUID is given, a UUID is generated from the hostname.
+
+        This is done for backwards compatibility with existing tests, and
+        should be removed enveutally.
```
exarkun: eventually

flocker/node/_deploy.py
```diff
@@ -476,6 +484,7 @@ def got_volumes(available_manifestations):
                 available_manifestations.values())
 
             return [NodeState(
+                # XXX uuid=self.uuid,
```
exarkun: Oops.  I guess this actually works because of the deprecated automatic hostname-to-uuid logic.  Gets used here, gets used in whatever test exercises this code, pow victory.

exarkun: (So add a test for the non-default UUID case, I guess)

flocker/node/_deploy.py
```diff
@@ -594,6 +609,7 @@ def discover_state(self, local_state):
             # convergence agent for datasets will discover the information
             # and then we can proceed.
             return succeed([NodeState(
+                uuid=self.uuid,
```
exarkun: Given the above comment, I wonder if this is properly tested.

flocker/node/_deploy.py
```diff
@@ -674,6 +690,7 @@ def applications_from_units(units):
                 ))
 
             return [NodeState(
+                uuid=self.uuid,
```
exarkun: And this.

flocker/node/agents/blockdevice.py
```diff
@@ -808,6 +808,7 @@ class BlockDeviceDeployer(PRecord):
         mounted.
     """
     hostname = field(type=unicode, mandatory=True)
+    uuid = field(type=UUID, mandatory=True)
```
exarkun: Probably could have a better name.  `node_uuid` or `node_id` maybe?

exarkun: Give it some API documentation in the class docstring, too.

flocker/node/script.py
exarkun: You forgot to update `synopsis`.

flocker/node/script.py
```diff
@@ -50,14 +52,29 @@ class ZFSAgentOptions(Options):
...
+def _get_external_ip(host, port):
+    """
+    Get an external IP address for this node that can in theory connect to
+    the given host and port.
+
+    :param host: A host to connect to.
+    :param port: The port to connect to.
+
+    :return unicode: IP address of external interface on this node.
+    """
```
exarkun: It would be good to refer to an issue for replacing this with control service-side node/address association.

flocker/node/script.py
```diff
@@ -67,8 +84,11 @@ class ZFSAgentScript(object):
     def main(self, reactor, options, volume_service):
         host = options["destination-host"]
         port = options["destination-port"]
-        deployer = P2PManifestationDeployer(
-            options["hostname"].decode("ascii"), volume_service)
+        ip = _get_external_ip(host, port)
+        # Soon we'll extract this from TLS certificate for node.  Until then
+        # we'll just do a temporary hack (probably to be fixed in FLOC-1727).
```
exarkun: This is FLOC-1727, so probably some other issue I guess.

flocker/node/script.py
```diff
@@ -177,10 +191,15 @@ def get_service(self, reactor, options):
...
         :return: The ``AgentLoopService`` instance.
         """
+        host = options["destination-host"]
+        port = options["destination-port"]
+        ip = _get_external_ip(host, port)
         return AgentLoopService(
             reactor=reactor,
-            deployer=self.deployer_factory(hostname=options["hostname"]),
-            host=options["destination-host"], port=options["destination-port"],
+            # Temporary hack, to be fixed in FLOC-1727 probably:
```
exarkun: This is FLOC-1727 so probably some other issue.

flocker/node/_deploy.py
exarkun: Looks like it probably is.  `flocker.node.test.test_deploy.ApplicationNodeDeployerDiscoverNodeConfigurationTests.test_unknown_manifestations` asserts a `NodeState` with the right uuid comes back.

flocker/node/_deploy.py
exarkun: And a bunch of other tests on `ApplicationNodeDeployerDiscoverNodeConfigurationTests` cover this case (mostly incidentally, though).

exarkun:
Thanks.  Address inline comments and then merge when buildbot is red^Wgreen^Wwhatever.

flocker/control/_model.py
itamarst: Honestly I hope this code goes away ASAP, so maybe just worth spending the time in near future to remove the backwards compatibility code if e.g. pyrsistent changes.

flocker/control/_model.py
itamarst: I'm pretty sure this code is going to be different once the next branch comes around (going to remove hostname attribute from `Node`). So if it's still duplicative I'll fix it there.

flocker/control/test/test_model.py
itamarst: Might be changing again, so I'd like to wait until next branch to see what final interactions look like (eventually `Node` oughtn't accept a hostname).

flocker/node/_deploy.py
itamarst: I just got rid backwards compat code in `NodeState`.

itamarst: Never mind, I didn't.

itamarst: I did however previously deliberately port application deployer discovery tests to new model, so I think coverage is sufficient.

PR #1328, opened 2015-04-21
make the order of `changes` passed to `in_parallel` irrelevant

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1714

I almost made `changes` a `PSet` but I concluded that being able to run the same change twice is probably a legitimate use-case (and at the very least less surprising than having two changes collapsed into one).

### DIFF
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -97,6 +97,9 @@ def run_state_change(change, deployer):
 class _InParallel(PRecord):
     changes = field(
         type=PVector,
+        # Sort the changes for the benefit of comparison.  Stick with a vector
+        # (rather than, say, a set) in case someone wants to run the same
+        # change multiple times in parallel.
         factory=lambda changes: pvector(sorted(changes)),
         mandatory=True
     )
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -119,6 +119,7 @@ def in_parallel(changes):
     Failures in one change do not prevent other changes from continuing.
 
     The order in which execution of the changes is started is unspecified.
+    Comparison of the resulting object disregards the ordering of the changes.
     """
     return _InParallel(changes=changes)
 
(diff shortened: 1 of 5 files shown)


flocker/node/test/test_change.py
```diff
@@ -10,22 +10,86 @@
...
+
+        self.assertEqual(
+            (True, False),
+            (first_parallel == second_parallel,
+             first_parallel != second_parallel)
+        )
+
+    def test_duplicates_run(self):
+        """
+        If the same change is passed to ``in_parallel`` twice then it is run
```
itamarst: Last few words in docstring look they can be deleted.

flocker/node/testtools.py
```diff
@@ -105,6 +105,18 @@ def run(self, deployer):
 
 
 @implementer(IDeployer)
+class DummyDeployer(object):
```
itamarst: You should run the standard `IDeployer` test suite against this.

flocker/node/_change.py
```diff
@@ -110,6 +114,8 @@ def in_parallel(changes):
     Run a series of changes in parallel.
 
     Failures in one change do not prevent other changes from continuing.
+
+    The order in which execution of the changes is started is unspecified.
```
itamarst: But perhaps document the comparability aspect, which I assume is the point of the exercise.

flocker/node/_change.py
```diff
@@ -95,7 +95,11 @@ def run_state_change(change, deployer):
 # certain other areas of the test suite depend on it.
 @implementer(IStateChange)
 class _InParallel(PRecord):
-    changes = field(type=PVector, factory=pvector, mandatory=True)
+    changes = field(
+        type=PVector,
+        factory=lambda changes: pvector(sorted(changes)),
```
itamarst: Add comment explaining motivation for sort, and why set is unsuitable.

itamarst:
OK, merge if buildbot is vaguely happy.

PR #1329, opened 2015-04-21
[FLOC-1711] API now allow settings container command line

itamarst:
This will be used to reimplement linking tests in simpler form (but it's a useful feature in general).


flocker/control/schema/types.yml
```diff
@@ -175,6 +175,14 @@ definitions:
       and optional tag, e.g. "clusterhq/redis:latest".
     type: string
 
+  command_line:
+    title: "Command line"
+    description: |
+      Alternate command-line to run using the Docker image's entry point.
+    type: array
+    items:
+      type: string
```
exarkun: Limit on the length of the array and length of the strings in the array makes sense here, I think (though overall maximum length would be really nice, I have no idea how to express that in jsonschema).

Some grepping suggests 4096 * 32 is the overall maximum on Linux.

flocker/control/_model.py
```diff
@@ -20,37 +20,72 @@
...
+
+def pvector_field(item_type, optional=False, initial=()):
+    """
+    Create checked ``PVector`` field.
+
+    :param item_type: The required type for the items in the vector.
+    :param bool optional: If true, ``None`` can be used as a value for
+        this field.
+    :param initial: Initial value to pass to factory if no value is given
+        for the field.
```
exarkun: I suspect this doesn't make much sense for a `pvector` field.  I guess we can leave it and not use it and maybe delete it in five years if it has still never proven useful.

exarkun: Er.  Except of course that's nonsense because I misunderstood what was being made optional.  Please disregard that comment.

flocker/control/test/test_model.py
```diff
@@ -766,6 +775,108 @@ class Record(PRecord):
...
+    def test_factory(self):
+        """
+        ``pvector_field`` has a factory that creates a ``PVector``.
+        """
+        class Record(PRecord):
+            value = pvector_field(int)
+        record = Record(value=[1, 2])
+        assert isinstance(record.value, PVector)
+
+    def test_checked_set(self):
```
exarkun: not a `set`

flocker/control/test/test_model.py
```diff
@@ -766,6 +775,108 @@ class Record(PRecord):
...
+        """
+        ``pvector_field`` results in a set that enforces its type.
```
exarkun: not a `set`

exarkun:
The new test had an error on one of the builders:

```
[ERROR]
Failure: twisted.internet.defer.TimeoutError: <flocker.node.functional.test_docker.NamespacedDockerClientTests testMethod=test_command_line> (test_command_line) still running at 120.0 secs

flocker.node.functional.test_docker.NamespacedDockerClientTests.test_command_line
```

flocker/node/functional/test_docker.py
```diff
@@ -641,6 +644,28 @@ def test_restart_policy_on_failure_maximum_count(self):
...
         return d
 
+    def test_command_line(self):
+        """
+        A container with custom command line is run with those arguments.
+        """
+        directory = FilePath(self.mktemp())
+        directory.makedirs()
+        d = self.start_container(
+            random_name(), image_name=u"busybox",
```
exarkun: A couple things:

  * I'm coming around to suspect that sharing any container images between tests (or between tests and real-world use) may result in more fragile tests.  Better that the test is the only thing that ever touches the image, otherwise who knows what crazy thing will happen inside Docker?  (Of course, if this is a valid line of thinking, there are other non-test implications that I am studiously not thinking about.)
  * `busybox` doesn't have an entrypoint so I don't think this demonstrates that the image's entrypoint is respected - which is the intended behavior, right?

Both of these suggest creating a tiny new image (in the spirit of #1281, sadly not yet merged) and using it in this test could be a good idea.

flocker/node/functional/test_docker.py
```diff
@@ -641,6 +644,28 @@ def test_restart_policy_on_failure_maximum_count(self):
...
+            volumes=[Volume(node_path=directory,
+                            container_path=FilePath(b"/data"))],
+            command_line=[u"sh", u"-c", u"echo 'hello' > /data/output"])
+
+        def started(_):
+            expected = directory.child(b"output")
+            for i in range(100):
+                if expected.exists():
+                    self.assertEqual(expected.getContent(), b"hello\n")
+                    return
```
exarkun: :sob: :sob: :sob: :sob: :sob: :sob:

exarkun: Doing a `loop_until(container is not running)` and then checking for the file might be more robust.  Not that this is related to how I saw this test fail on buildbot (with a timeout error, not a "Files never created" failure as happens below).

exarkun:
Thanks.  Address inline comments to your satisfaction, make sure the new functional test passes reliably, then merge.

flocker/node/functional/test_docker.py
itamarst: We decided that it does, in fact, have an entrypoint (the default one, `/bin/sh`).

itamarst: As far as not reusing tests... we have a bunch of functional testes that happily reuse `busybox`. I don't think sharing images is a problem, I think it's just that the pull-test is a problem.

PR #1332, opened 2015-04-22
FLOC-1575 - Block device attach logic

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1575

Note:

  * The original design called for a two-step attach, mount operation to be performed in this case.  I've now decided not to continue in that direction.  Instead, this branch just produces an attach operation.  A follow-up branch will detect the attached-but-not-mounted state and produce just a mount operation.  This simplifies things.
  * This includes some changes to `MountBlockDevice` from before I made that decision.  I could revert them but they'll just need to be re-made in the branch that does the mount operation so I'd rather just leave them.
  * As promised, `AttachVolume` is now actually asynchronous.
  * `BlockDeviceDeployer` now provides `async_block_device_api` in support for this.  Other operations will be switched over to this as `calculate_changes` is made smarter.
  * Fixed a few other tiny bugs that got in the way here - mostly passing the wrong type of thing to something, eg passing a `Deployment` where a `DeploymentState` was expected.
  * To make the new test pass, which expects exactly one `AttachVolume` operation, I fixed the bug where existing-but-unattached volumes are created (now they are attached instead).


itamarst:
Don't see the shared logic for mount/format etc.? But basically looks fine, please continue.

exarkun:
Scope here excludes filesystem initialization (this code will assume an existing volume it finds has already been initialized - could be wrong in a failure condition but ... later).  This does include mounting, though, hence the `MountBlockDevice` changes in the comments.

flocker/node/agents/blockdevice.py
```diff
@@ -527,23 +553,28 @@ class AttachVolume(PRecord):
...
-            self.volume.blockdevice_id, self.hostname
+        api = deployer.async_block_device_api
+        listing = api.list_volumes()
+        listing.addCallback(
+            _blockdevice_volume_from_datasetid, self.dataset_id
         )
-        return succeed(None)
+
+        def found(volume):
+            ATTACH_VOLUME_DETAILS(volume=volume).write(_logger)
```
wallrj: What if _blockdevice_volume_from_datasetid returns `None` ? `dataset_id` not found among the volumes....for some reason? This log message will blow up with a type error or the attach_volume below will. Which will trigger an errback...which will be handled and logged by...?

wallrj:
There seem to be an awful lot (117) failed tests here:
 * http://build.clusterhq.com/builders/flocker-fedora-20/builds/5865/steps/trial/logs/stdio

Lot's of `requests.exceptions.ReadTimeout: UnixHTTPConnectionPool(host='localhost', port=None): Read timed out. (read timeout=60)` during API requests to Docker.
Not related to this branch, surely, but new to me.

flocker/node/agents/blockdevice.py
```diff
@@ -1160,12 +1185,37 @@ class BlockDeviceDeployer(PRecord):
...
+    def async_block_device_api(self):
+        """
+        Get an ``IBlockDeviceAsyncAPI`` provider which can manipulate volumes
+        for this deployer.
+
+        During real operation, this is a threadpool-based wrapper around the
+        ``IBlockDeviceAPI`` provider.  For testing purposes it can be
+        overridden with a different object entirely (and this large amount of
+        support code for this is necessary because this class is a ``PRecord``
+        subclass).
```
wallrj: Ah, I see. Clever!

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -580,6 +647,57 @@ def test_delete_before_resize(self):
...
+        self.assertEqual(
+            in_parallel(changes=[
+                AttachVolume(
+                    dataset_id=UUID(dataset.dataset_id),
+                    # Attach it to this node.
+                    hostname=deployer.hostname
+                ),
+            ]),
+            changes
+        )
```
wallrj: Yep, that all makes sense.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1875,7 +1984,7 @@ class CreateFilesystemTests(
 class MountBlockDeviceInitTests(
     make_with_init_tests(
         MountBlockDevice,
-        dict(volume=_ARBITRARY_VOLUME, mountpoint=FilePath(b"/foo")),
```
wallrj: OMG. I think I might be responsible for the `make_with_init_tests` function, but it seems like overkill now.
Shall we delete these?
Or at least supply the two dictionaries as keyword arguments so that the reader doesn't have go digging into flocker.testtools.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1974,8 +2083,8 @@ def create(self):
 class MountBlockDeviceTests(
     make_istatechange_tests(
         MountBlockDevice,
-        dict(volume=_ARBITRARY_VOLUME, mountpoint=FilePath(b"/foo")),
-        dict(volume=_ARBITRARY_VOLUME, mountpoint=FilePath(b"/bar")),
+        dict(dataset_id=uuid4(), mountpoint=FilePath(b"/foo")),
```
wallrj: And this one is even more confusing to read because we're supplying the two dictionaries for the benefit of a `make_comparison_tests` whose test case is mixed in to the test case produced by `make_istatechange_tests`.
But even keyword args won't help in this case.

flocker/node/agents/blockdevice.py
```diff
@@ -167,18 +173,30 @@ class UnattachedVolume(VolumeException):
 
 MOUNT_BLOCK_DEVICE = ActionType(
     u"agent:blockdevice:mount",
-    [VOLUME],
+    [DATASET_ID],
     [],
     u"A block-device-backed dataset is being mounted.",
 )
 
+MOUNT_BLOCK_DEVICE_DETAILS = MessageType(
```
wallrj: I wonder how we could eventually add these useful details to the surrounding ACTION?

flocker/node/agents/blockdevice.py
wallrj: And coverage shows that that case never gets tested:
 * file:///home/richard/projects/HybridLogic/flocker/htmlcov/flocker_node_agents_blockdevice.html#n288

Is it possible that a block device might have been destroyed between the calculation of the state changes and the running of the change? It sounds plausible.

flocker/node/agents/blockdevice.py
```diff
@@ -591,6 +622,8 @@ def run(self, deployer):
         return succeed(None)
 
 
+# Get rid of this in favor of calculating each individual operation in
+# BlockDeviceDeployer.calculate_changes.  FLOC-1771
```
wallrj: So eventually, calculate_state_changes will detect eg a block device for a  dataset that isn't formatted and yield a CreateFilesystem change ONLY.
Then in the next convergence loop it'll detect block device for a dataaset that isn't mounted and yield `MountBlockDevice`.

But then won't we need more block device states than `manifest` /  `non-manifest` in the Deployment state? 
...hmm i suppose not 
 * `discover_state` will have to only report mounted filesystems as `manifest` (perhaps it already does)
 * `calculate_changes` then examines the exact state of all non-manifest datasets and yields the correct next state change.

wallrj:
Thanks @exarkun 

I got it eventually. And this branch and proposed followup work makes sense to me.

The buildbot test failures are probably unrelated.

Please answer or address the comments above and merge when you're happy with the build results.

flocker/node/agents/blockdevice.py
exarkun: We may want to represent those other states to the end user anyway.  They might reveal interesting information about the state of their cluster.

PR #1335, opened 2015-04-23
[FLOC-1638] Node certificate

dwgebler:
Fixes FLOC-1638

### DIFF
--- a/flocker/ca/_ca.py
+++ b/flocker/ca/_ca.py
@@ -8,6 +8,8 @@ Uses RSA 4096-bit + SHA 256.
 
 import os
 
+from uuid import uuid4
+
 from OpenSSL import crypto
 from pyrsistent import PRecord, field
 from twisted.internet.ssl import DistinguishedName, KeyPair, Certificate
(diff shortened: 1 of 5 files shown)


flocker/ca/_ca.py
```diff
@@ -229,6 +231,65 @@ def write_credential_files(self, key_filename, certificate_filename):
...
 
 
+class NodeCredential(FlockerCredential):
+    """
+    A certificate for a node agent, signed by a supplied certificate
+    authority.
+
+    :ivar bytes uuid: A unique identifier for the node this certificate
+        identifies, in the form of a version 4 UUID.
+    """
```
jongiddy: Should the docstring contain the fields in the superclass as well?

dwgebler: I wouldn't have thought so, though worth mentioning the follow up API user certificate branch removes this inheritance and uses composition instead, with the docstring specifying that the `credential` field is a `FlockerCredential`.

flocker/ca/_script.py
```diff
@@ -90,6 +90,56 @@ def getSynopsis(self):
...
+    authority (see flocker-ca initialize command for more information).
+    """
+
+    synopsis = "[options]"
+
+    optParameters = [
+        ['inputpath', 'i', os.getcwd(),
+         'Path to directory containing root certificate.'],
+        ['outputpath', 'o', os.getcwd(),
+         'Path to directory to write control service certificate.'],
```
jongiddy: It is unusual to have default arguments that change with context. If I saw the default value once in the usage message, I would expect it to stay as that value every time I ran the command. I think instances of this should display that the default is the current working directory, not the value of the current working directory at this moment.

flocker/ca/test/test_ca.py
```diff
@@ -21,6 +23,297 @@
...
+        crt_path = path.child(cert_file)
+        crt_file = crt_path.open(b'w')
+        crt_file.write(b"dummy")
+        crt_file.close()
+        key_path = path.child(key_file)
+        key_file = key_path.open(b'w')
+        key_file.write(b"dummy")
+        key_file.close()
+        # make file unreadable
+        key_path.chmod(64)
```
jongiddy: More usual to express this in octal (0100). Is there a reason to keep user-executable bit set?

flocker/ca/functional/test_ca.py
```diff
@@ -100,6 +102,24 @@ def test_control_certificate(self):
...
+        """
+        flocker_ca("initialize", "mycluster")
+        status, output = flocker_ca("create-node-certificate")
+        # Find the generated file name with UUID from the output.
+        file_pattern = re.compile("([a-zA-Z0-9\-]*\.crt)")
+        file_name = file_pattern.search(output).groups()[0]
+        self.assertTrue(
+            openssl_verify("cluster.crt", file_name)
+        )
+
```
jongiddy: All but one of the tests in this class create the file `flocker/_trial_temp/cluster.crt`. This causes all but the first test executed to give the message ``Error: Unable to write certificate file. File exists``.  Add a cleanup task, either after each initialize line, or separate out the `help_description` test, and add a `tearDown` method.

flocker/ca/test/test_ca.py
```diff
@@ -21,6 +23,297 @@
...
+        A cert written by ``NodeCredential.initialize`` has an expiry
+        date 20 years from the date of signing.
+        """
+        today = datetime.datetime.now()
+        expected_expiry = today + datetime.timedelta(seconds=EXPIRY_20_YEARS)
+        nc = NodeCredential.initialize(self.path, self.ca)
+        cert = nc.certificate.original
+        asn1 = cert.get_notAfter()
+        expiry_date = datetime.datetime.strptime(asn1, "%Y%m%d%H%M%SZ")
+        self.assertEqual(expiry_date.date(), expected_expiry.date())
```
jongiddy: Just a reminder that this may have problems around midnight

jongiddy: This might be better as a nearly equal expression

jongiddy:
This is OK to merge, if the comments above are addressed.

flocker/ca/test/test_ca.py
exarkun: It seems this didn't get addressed?

dwgebler: Another branch fixes this expiry date issue across all the ``flocker.ca` tests. See #1339

PR #1336, opened 2015-04-23
[FLOC-1639] API user certificate

dwgebler:
Fixes FLOC-1639


flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+    """
+    A certificate for an API user, signed by a supplied certificate
+    authority.
+
+    :ivar FlockerCredential credential: The certificate and key pair
+        credential object.
+    :ivar bytes username: A username.
+    """
+    credential = field(mandatory=True)
+    username = field(mandatory=True, initial=None)
```
exarkun: You can do `type=bytes` or `type=FlockerCredential` in these field definitions to get type enforcement on them (and possibly make the unchecked type documentation in the class docstring superfluous).

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+
+    @classmethod
+    def from_path(cls, path, username):
+        """
+        Load a node certificate from a specified path.
```
exarkun: Needs parameter documentation

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        """
+        key_filename = b"{user}.key".format(user=username)
+        cert_filename = b"{user}.crt".format(user=username)
```
exarkun: :question:

`username + b".key"` works just as well and without the confusing mix of bytes with new-style formatting (which .. works .. but .. it's sort of gross).

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        credential = FlockerCredential(
+            path=path, keypair=keypair, certificate=certificate)
+        return cls(credential=credential, username=username)
+
+    @classmethod
+    def initialize(cls, path, authority, username):
+        """
+        Generate a certificate signed by the supplied root certificate.
+
+        :param FilePath path: Directory where the certificate will be stored.
```
exarkun: Maybe this could have a nicer name - `output_path` or something?

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        :param CertificateAuthority authority: The certificate authority with
+            which this certificate will be signed.
+        :param bytes username: The username to be included in the certificate.
```
exarkun: :question: 

Using `bytes` to represent a username means the encoding is up to the caller.  It would be nice to enforce UTF-8 here.

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        dn = DistinguishedName(
+            commonName=name, organizationalUnitName=organizational_unit
+        )
+        keypair = flocker_keypair()
+        request = keypair.keypair.requestObject(dn)
+        serial = os.urandom(16).encode(b"hex")
+        serial = int(serial, 16)
+        cert = authority.credential.keypair.keypair.signRequestObject(
+            authority.credential.certificate.getSubject(), request,
+            serial, EXPIRY_20_YEARS, 'sha256'
```
exarkun: digest name should be a byte string, `b"sha256"`.

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
             os.umask(original_umask)
 
 
-class NodeCredential(FlockerCredential):
+class UserCredential(PRecord):
```
exarkun: `FlockerCredential` still documents itself as a "Base class".

flocker/ca/_ca.py
```diff
@@ -272,7 +339,7 @@ def initialize(cls, path, authority):
         # The organizational unit is set to the common name of the
         # authority, which in our case is a byte string identifying
         # the cluster.
-        organizational_unit = authority.certificate.getSubject().CN
+        organizational_unit = authority.credential.certificate.getSubject().CN
```
exarkun: Too many `.`s in this expression.  Perhaps `CertificateAuthority` should have a method that helps out with this part of the certificate generation logic.

flocker/ca/_ca.py
```diff
@@ -320,7 +394,7 @@ def initialize(cls, path, authority):
...
```
exarkun: Bonus, that logic could be re-used here instead of duplicating a lot of code.

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        :param bytes username: The username to be included in the certificate.
+        """
+        key_filename = b"{user}.key".format(user=username)
+        cert_filename = b"{user}.crt".format(user=username)
+        # The common name for the node certificate.
+        name = b"user-{user}".format(user=username)
+        # The organizational unit is set to the common name of the
+        # authority, which in our case is a byte string identifying
+        # the cluster.
+        organizational_unit = authority.credential.certificate.getSubject().CN
```
exarkun: And again here.

exarkun: I'm not sure what the benefit of copying the cluster name into the OU of the subject is.  The certificate already includes the issuer name in it, which definitely includes the cluster name already.

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        cert = authority.credential.keypair.keypair.signRequestObject(
+            authority.credential.certificate.getSubject(), request,
+            serial, EXPIRY_20_YEARS, 'sha256'
+        )
+        credential = FlockerCredential(
+            path=path, keypair=keypair, certificate=cert
+        )
+        instance = cls(credential=credential, username=username)
+        instance.credential.write_credential_files(
+            key_filename, cert_filename)
```
exarkun: You could use `credential` to do this (and perhaps before instantiating `cls` so that the instance is always valid instead of maybe referring to files that don't actually exist - as could happen if `write_credential_files` raises an exception because it can't do the write for some reason).

flocker/ca/_script.py
```diff
@@ -90,6 +90,62 @@ def getSynopsis(self):
...
+
+    * name: A username for which the certificate should be created.
+    """
+
+    synopsis = "<name> [options]"
+
+    optParameters = [
+        ['inputpath', 'i', os.getcwd(),
+         'Path to directory containing root certificate.'],
+        ['outputpath', 'o', os.getcwd(),
```
exarkun: :question: 

I think Jon had a concern about variable defaults which sounded kind of reasonable to me.  You could make the default `None` and write words in the description saying the default is the current directory.

flocker/ca/_script.py
```diff
@@ -90,6 +90,62 @@ def getSynopsis(self):
...
+    def run(self):
+        """
+        Check if root key and certificate files (either default or as
+        specified on the command line) exist in the path and error out if
+        they do not. If there are no path errors, create a new node
+        certificate signed by the root and write it out to the current
+        directory.
+        """
+        self["inputpath"] = FilePath(self["inputpath"])
+        self["outputpath"] = FilePath(self["outputpath"])
```
exarkun: :question: 

This is a good thing to do in `postOptions`.  Also you could probably use `FilePath` as the 4th element in `optParameters` items to do this automatically.

flocker/ca/_script.py
```diff
@@ -90,6 +90,62 @@ def getSynopsis(self):
...
+    def parseArgs(self, name):
+        self["name"] = name
+
+    def run(self):
+        """
+        Check if root key and certificate files (either default or as
+        specified on the command line) exist in the path and error out if
+        they do not. If there are no path errors, create a new node
+        certificate signed by the root and write it out to the current
+        directory.
```
exarkun: This is a lot of detail and the main purpose of the method is hidden near the end.  I'd make the main purpose of the method the focus of this docstring and document the error cases with `:raise:` or something similar.

flocker/ca/_script.py
```diff
@@ -90,6 +90,62 @@ def getSynopsis(self):
...
+        try:
+            try:
+                ca = RootCredential.from_path(self["inputpath"])
+                uc = UserCredential.initialize(
+                    self["outputpath"], ca, self["name"])
+                print (
+                    b"Created {user}.crt. You can now give it to your "
+                    "API enduser so they can access the control service "
+                    "API.".format(user=uc.username)
+                )
```
exarkun: `print` is a really bad thing to use in real programs.  Parameterizing the output stream would make this code much more easily testable and would prevent extra junk from being written to the output stream when running the tests.

flocker/ca/functional/test_ca.py
```diff
@@ -133,6 +133,21 @@ def test_node_certificate(self):
...
     @requireCA
+    def test_apiuser_certificate(self):
+        """
+        Test for ``flocker-ca create-api-certificate`` command.
+        Runs ``flocker-ca initialize`` followed by
+        ``flocker-ca create-api-certificate` and calls ``openssl``
+        to verify the generated control certificate and private key is
+        signed by the previously generated certificate authority.
+        """
+        flocker_ca("initialize", "mycluster")
```
exarkun: `flocker_ca` uses `check_output` which leaks bytes onto the parents stdio streams.  It would be better if it used `flocker.testtools.run_process`.

exarkun: Also, these things look like they should be bytes literals.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        pub = uc.credential.certificate.getPublicKey().original
+        pub_asn1 = crypto.dump_privatekey(crypto.FILETYPE_ASN1, pub)
+        priv_asn1 = crypto.dump_privatekey(crypto.FILETYPE_ASN1, priv)
+        pub_der = asn1.DerSequence()
+        pub_der.decode(pub_asn1)
+        priv_der = asn1.DerSequence()
+        priv_der.decode(priv_asn1)
+        pub_modulus = pub_der[1]
+        priv_modulus = priv_der[1]
+        self.assertEqual(pub_modulus, priv_modulus)
```
exarkun: `OpenSSL.Context.check_privatekey` (pyOpenSSL 0.15) will do this comparison for you.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+    def test_keypair_correct_umask(self):
+        """
+        A keypair file written by ``UserCredential.initialize`` has
+        the correct permissions (0600).
+        """
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        key_file = b"{user}.key".format(user=uc.username)
+        keyPath = self.path.child(key_file)
+        st = os.stat(keyPath.path)
+        self.assertEqual(b'0600', oct(st.st_mode & 0777))
```
exarkun: Some of these tests aren't for the key/certificate handling logic exactly.  These tests are very slow (>1s each on average - that's at least 100x too slow) because of the large number of large key pairs they generate.  It would be nice to avoid doing all of this extra work just to make an assertion about the permissions of the file the data gets written to.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        """
+        path = FilePath(self.mktemp())
+        path.makedirs()
+        cert_file = b"{user}.crt".format(user=self.username)
+        e = self.assertRaises(
+            PathError, UserCredential.from_path, path, self.username
+        )
+        expected = (b"Certificate file could not be opened. "
+                    b"No such file or directory "
+                    b"{path}").format(
```
exarkun: Exceptions stringify to the native string type (`"..."`), not necessarily to bytes.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        """
+        A ``PathError`` is raised if the certificate file path given to
+        ``UserCredential.from_path`` cannot be opened for reading.
+        """
+        path = FilePath(self.mktemp())
+        path.makedirs()
+        cert_file = b"{user}.crt".format(user=self.username)
+        key_file = b"{user}.key".format(user=self.username)
+        crt_path = path.child(cert_file)
+        crt_file = crt_path.open(b'w')
```
exarkun: `with crt_path.open("w") as crt_file:` will save you some manual closing here

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        crt_file.write(b"dummy")
+        crt_file.close()
+        # make file unreadable
+        crt_path.chmod(64)
```
exarkun: At least an octal literal here and elsewhere

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        )
+        expected = (
+            b"Private key file could not be opened. "
+            b"Permission denied {path}"
+        ).format(path=key_path.path)
+        self.assertEqual(str(e), expected)
+
+    def test_certificate_subject_username(self):
+        """
+        A cert written by ``UserCredential.initialize`` has the
```
exarkun: `certificate`

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        issuer = cert.get_issuer()
+        subject = cert.get_subject()
+        self.assertEqual(
+            issuer.CN,
+            subject.OU
+        )
+
+    def test_certificate_is_signed_by_ca(self):
+        """
+        A cert written by ``UserCredential.initialize`` is validated
```
exarkun: I don't think "validated" adds anything here.  The test doesn't actually do any validation.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+            issuer.CN,
+            self.ca.credential.certificate.getSubject().CN
+        )
+
+    def test_certificate_expiration(self):
+        """
+        A cert written by ``UserCredential.initialize`` has an expiry
+        date 20 years from the date of signing.
+        """
+        today = datetime.datetime.now()
```
exarkun: I think you said there's a branch that fixes the time math in these tests.  Can you link to it from here?

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        expected_expiry = today + datetime.timedelta(seconds=EXPIRY_20_YEARS)
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
+        asn1 = cert.get_notAfter()
```
exarkun: `asn1` is a confusing variable name for this - <http://en.wikipedia.org/wiki/Abstract_Syntax_Notation_One>.  `cert.get_notAfter()` doesn't return ASN.1 bytes.  It returns a string describing a date.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+
+    def test_certificate_subject_username(self):
+        """
+        A cert written by ``UserCredential.initialize`` has the
+        subject common name "user-{user}" where {user} is the username
+        supplied during the certificate's creation.
+        """
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
+        subject = cert.get_subject()
```
exarkun: Lots of `.`s to get here... `uc.credential.certificate.original.get_subject()`.  Also, this doesn't test what the docstring says which is that a certificate gets "written" somewhere.  This uses the in-memory representation of the certificate.  If you change the test so it reads the written certificate you can use fewer dots and do what the docstring actually describes.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        self.assertEqual(subject.CN, b"user-{user}".format(user=uc.username))
+
+    def test_certificate_ou_matches_ca(self):
+        """
+        A cert written by ``UserCredential.initialize`` has the issuing
+        authority's common name as its organizational unit name.
+        """
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
```
exarkun: Same comment

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+            subject.OU
+        )
+
+    def test_certificate_is_signed_by_ca(self):
+        """
+        A cert written by ``UserCredential.initialize`` is validated
+        as being signed by the certificate authority.
+        """
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
```
exarkun: Same comment

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+
+    def test_certificate_expiration(self):
+        """
+        A cert written by ``UserCredential.initialize`` has an expiry
+        date 20 years from the date of signing.
+        """
+        today = datetime.datetime.now()
+        expected_expiry = today + datetime.timedelta(seconds=EXPIRY_20_YEARS)
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
```
exarkun: Same comment

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        self.assertEqual(expiry_date.date(), expected_expiry.date())
+
+    def test_certificate_is_rsa_4096_sha_256(self):
+        """
+        A cert written by ``UserCredential.initialize`` is an RSA
+        4096 bit, SHA-256 format.
+        """
+        uc = UserCredential.initialize(self.path, self.ca, self.username)
+        cert = uc.credential.certificate.original
+        key = uc.credential.certificate.getPublicKey().original
```
exarkun: Same comment

flocker/ca/test/test_ca.py
```diff
@@ -67,8 +351,8 @@ def test_decoded_certificate_matches_private_key(self):
         be paired with.
         """
         nc = NodeCredential.initialize(self.path, self.ca)
-        priv = nc.keypair.keypair.original
-        pub = nc.certificate.getPublicKey().original
+        priv = nc.credential.keypair.keypair.original
+        pub = nc.credential.certificate.getPublicKey().original
```
exarkun: I'm not sure what "decoded" is meant to convey in this test method name/docstring.  Is it to do with the ASN.1 decoding that happens lower down?  If so, that seems like a sort of uninteresting implementation detail of the test.  I think what's actually important here is that the certificate and private key match each other (ie, make up a key pair).  How this gets proven is best left as implementation details inside the test method.

flocker/ca/_ca.py
```diff
@@ -231,14 +231,81 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+class UserCredential(PRecord):
+    """
+    A certificate for an API user, signed by a supplied certificate
+    authority.
+
+    :ivar FlockerCredential credential: The certificate and key pair
+        credential object.
+    :ivar bytes username: A username.
+    """
+    credential = field(mandatory=True)
```
exarkun: :question: 

It seems like this field just encourages application code to reach way down into the internals to grab random bits.  This isn't great.  If this were a library we were building for serious long term use, I'd want to address this somehow - probably starting by making this attribute private and carefully scrutinizing any code that wants to use it or attributes of the object it refers to.

flocker/ca/test/test_ca.py
```diff
@@ -16,13 +16,296 @@
...
+        ).format(path=crt_path.path)
+        self.assertEqual(str(e), expected)
+
+    @not_root
+    @skip_on_broken_permissions
+    def test_load_error_on_unreadable_key_file(self):
+        """
+        A ``PathError`` is raised if the key file path given to
+        ``UserCredential.from_path`` cannot be opened for reading.
+        """
```
exarkun: This and several of the other error-case tests are basically identical.  Instead of duplicating the implementation for each credential class, it would be better to factor this code so it can be reused.  Take a look at the `make_..._tests` pattern used elsewhere in the code base and try adopting it for these tests.

exarkun:
Thanks.  Please address those points and then re-submit.

flocker/ca/_ca.py
```diff
@@ -95,8 +99,8 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
 
     :param int serial: The certificate serial number.
 
-    :param int expiry: Number of seconds from now until this certificate
-        should expire.
+    :param str expiry: ASN1 formatted datetime string representing the
```
itamarst: Bytes. `str` should only be used to describe a Python-related object, e.g. something that is used with `getattr()`. Otherwise always use `bytes` or `unicode`, since `str` is ambiguous (it means Unicode on Python 3).

exarkun: It's a GeneralizedTime - https://www.ietf.org/rfc/rfc2459, section 4.1.2.5.2

flocker/ca/_ca.py
```diff
@@ -95,8 +99,8 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
...
+        certificate's expiry date, i.e. "%Y%m%d%H%M%SZ"
```
exarkun: This isn't a validate expiration time, though.  It's a format string used to generate such a time.  Put an actual date into this example.

flocker/ca/_ca.py
```diff
@@ -95,8 +99,8 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
...
 
     :param str digest: The digest algorithm to use.
```
exarkun: Also `bytes` I think

flocker/ca/_ca.py
```diff
@@ -95,10 +99,10 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
 
     :param int serial: The certificate serial number.
 
-    :param int expiry: Number of seconds from now until this certificate
-        should expire.
+    :param bytes expiry: ASN1 formatted datetime string representing the
```
exarkun: ASN.1 actually has lots of different date/time formats.  This one is a GeneralizedTime.

exarkun: For future reference, a nice Python-y API would probably just accept a `datetime.datetime` here and do the string formatting itself.  We don't care that much about this as an API, though.

flocker/ca/_ca.py
```diff
@@ -231,14 +267,90 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+    @classmethod
+    def from_path(cls, path, username):
+        """
+        Load a node certificate from a specified path.
+
+        :param FilePath path: Directory where user certificate and key
+            files are stored.
+        :param bytes username: A UTF-8 encoded username.
+        """
+        username = unicode(username, "utf-8")
```
exarkun: Unless you have no idea what type an object is, `foo.decode("utf-8")` is a better way to make it unicode.

flocker/ca/_ca.py
```diff
@@ -106,7 +110,7 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
     cert.set_subject(req.get_subject())
     cert.set_pubkey(req.get_pubkey())
     cert.gmtime_adj_notBefore(0)
-    cert.gmtime_adj_notAfter(expiry)
+    cert.set_notAfter(expiry)
```
exarkun: This should be `cert.set_notBefore(beginning_of_validity_period)` and `cert.set_notAfter(beginning_of_validity_period + validity_interval)`.

`beginning_of_validity_period` should be a parameter accepted by this function.  `validity_interval` can be hard-coded at the 20 year period as defined in the globals.

This makes it very easy to test.  You know exactly what values to expect in these fields because you passed them in (or you passed in the first and the second is a trivial function of the first).  This removes the factor of "what time was it when the tests started to run" from the equation entirely.

It's kind of like how the sys module is parameterized to the script runner.  The actual sys module is complicated and hard to test - so we replace it with something simple and easy to test.

flocker/ca/_ca.py
```diff
@@ -126,6 +130,38 @@ def create_certificate_authority(keypair, dn, request, serial, expiry, digest):
...
+
+    :param bytes digest: The digest algorithm to use.
+    """
+    req = request.original
+    cert = crypto.X509()
+    dn._copyInto(cert.get_issuer())
+    cert.set_subject(req.get_subject())
+    cert.set_pubkey(req.get_pubkey())
+    cert.gmtime_adj_notBefore(0)
+    cert.set_notAfter(expiry)
```
exarkun: Same thing here, of course.

flocker/ca/_ca.py
```diff
@@ -231,14 +267,90 @@ def write_credential_files(self, key_filename, certificate_filename):
...
+        """
+        Load a node certificate from a specified path.
+
+        :param FilePath path: Directory where user certificate and key
+            files are stored.
+        :param bytes username: A UTF-8 encoded username.
+        """
+        username = unicode(username, "utf-8")
+        key_filename = username + b".key"
+        cert_filename = username + b".crt"
```
exarkun: ... `username` is unicode now.  Concatenating bytes only works by accident.

flocker/ca/_script.py
```diff
@@ -90,6 +90,68 @@ def getSynopsis(self):
...
+            self["outputpath"] = os.getcwd()
+
+        self["inputpath"] = FilePath(self["inputpath"])
+        self["outputpath"] = FilePath(self["outputpath"])
+
+        try:
+            try:
+                ca = RootCredential.from_path(self["inputpath"])
+                uc = UserCredential.initialize(
+                    self["outputpath"], ca, self["name"])
```
exarkun: This would be a better place to decode bytes into unicode (and therefore, `initialize` should accept unicode, not bytes).  As is, nothing actually verifies the username is UTF-8.  As is, this will fail with a decoding error as soon as someone uses it in a non-UTF-8 locale with non-ASCII code points.  It will fail inside `initialize` and no one will handle the exception.  Decoding bytes to unicode as near the I/O layer as possible makes the most sense.  That's where you're most likely to know what encoding to use (`stdin.encoding` is a somewhat useful first guess).

flocker/ca/_script.py
```diff
@@ -90,6 +90,68 @@ def getSynopsis(self):
...
+                self._sys_module.stdout.write(
+                    b"Created {user}.crt. You can now give it to your "
+                    "API enduser so they can access the control service "
```
exarkun: You have to make every one of these a bytes literal separately.

flocker/ca/_script.py
```diff
@@ -90,6 +90,68 @@ def getSynopsis(self):
...
+                    "API.".format(user=uc.username)
```
exarkun: Formatting unicode into bytes is not a good idea.  Format it into a unicode template string and the encode the entire thing using an encoding appropriate for the output stream (`stdout.encoding`, probably).

flocker/ca/_script.py
```diff
@@ -131,7 +193,7 @@ def run(self):
             try:
                 ca = RootCredential.from_path(self["inputpath"])
                 nc = NodeCredential.initialize(self["outputpath"], ca)
-                print (
+                self._sys_module.stdout.write(
                     b"Created {uuid}.crt. Copy it over to "
                     "/etc/flocker/node.crt on your node "
                     "machine and make sure to chmod 0600 it.".format(
```
exarkun: Likewise here

flocker/ca/_script.py
```diff
@@ -191,7 +253,7 @@ def run(self):
             try:
                 ca = RootCredential.from_path(self["inputpath"])
                 ControlCredential.initialize(self["outputpath"], ca)
-                print (
+                self._sys_module.stdout.write(
                     b"Created control-service.crt. Copy it over to "
                     "/etc/flocker/control-service.crt on your control service "
                     "machine and make sure to chmod 0600 it."
```
exarkun: And here

flocker/ca/_script.py
```diff
@@ -241,7 +303,7 @@ def run(self):
         try:
             try:
                 RootCredential.initialize(self["path"], self["name"])
-                print (
+                self._sys_module.stdout.write(
                     b"Created cluster.key and cluster.crt. "
                     "Please keep cluster.key secret, as anyone who can access "
                     "it will be able to control your cluster."
```
exarkun: And here

flocker/ca/_ca.py
```diff
@@ -251,12 +363,12 @@ def from_path(cls, path, uuid):
...
-        return cls(
-            path=path, keypair=keypair, certificate=certificate, uuid=uuid
-        )
+        credential = FlockerCredential(
+            path=path, keypair=keypair, certificate=certificate)
+        return cls(credential=credential, uuid=uuid)
 
     @classmethod
-    def initialize(cls, path, authority):
+    def initialize(cls, path, authority, **kwargs):
```
exarkun: I think `uuid=None` and `if uuid is None: uuid = bytes(uuid4())`?

flocker/ca/_ca.py
```diff
@@ -328,20 +448,31 @@ def initialize(cls, path, authority):
...
     """
     A credential representing a self-signed certificate authority.
+    :ivar FlockerCredential credential: The certificate and key pair
+        credential object.
     """
+    credential = field(mandatory=True)
+
+    @property
+    def common_name(self):
+        return self.credential.certificate.getSubject().CN
```
exarkun: Good step

exarkun:
Thanks.  Address inline comments or file issues for them.  Then merge if buildbot is not any redder than it was before.

PR #1337, opened 2015-04-23
[FLOC-1549] Auto-async wrapper for IBlockDeviceAPI  providers.

itamarst:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1549

### DIFF
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -1,0 +1,24 @@
+# FLOC-1549
+
+# from twisted.internet.threads import deferToThreadPool
+
+
+# def _threaded_method(method_name, reactor_name, sync_name, threadpool_name):
+#     def _run_in_thread(self, *args, **kwargs):
+#         reactor = getattr(self, reactor_name)
+#         sync = getattr(self, sync_name)
+#         threadpool = getattr(self, threadpool_name)
+#         original = getattr(sync, method_name)
+#         return deferToThreadPool(reactor, threadpool, original, *args, **kwargs)
+#     return _run_in_thread
+
+
+# def auto_threaded(interface, sync, threadpool):
+#     def _threaded_class_decorator(cls):
+#         for name in interface.names():
+#             cls.__dict__[name] = _threaded_method(
+#                 name, sync, threadpool)
+#             )
+#         return cls
+#     return _threaded_class_decorator
+
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -65,4 +65,3 @@ def auto_threaded(interface, reactor, sync, threadpool):
             )
         return cls
     return _threaded_class_decorator
-
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -1,9 +1,30 @@
-# FLOC-1549
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Some thread-related tools.
+"""
 
 from twisted.internet.threads import deferToThreadPool
 
 
-def _threaded_method(method_name, reactor_name, sync_name, threadpool_name):
+def _threaded_method(sync_name, method_name, reactor_name, threadpool_name):
+    """
+    Create a method that calls another method in a threadpool.
+
+    :param str sync_name: The name of the attribute of ``self`` on which to
+        look up the other method to run.  This is the "sync" object.
+    :param str method_name: The name of the method to look up on the "sync"
+        object.
+    :param str reactor_name: The name of the attribute of ``self`` referencing
+        the reactor to use to get results back to the calling thread.
+    :param str threadpool_name: The name of the attribute of ``self``
+        referencing a ``twisted.python.threadpool.ThreadPool`` instance to use
+        to run the method in a different thread.
+
+    :return: The new thread-using method.  It has the same signature as the
+             original method except it returns a ``Deferred`` that fires with
+             the original method's result.
+    """
     def _run_in_thread(self, *args, **kwargs):
         reactor = getattr(self, reactor_name)
         sync = getattr(self, sync_name)
@@ -40,7 +61,7 @@ def auto_threaded(interface, reactor, sync, threadpool):
     def _threaded_class_decorator(cls):
         for name in interface.names():
             setattr(
-                cls, name, _threaded_method(name, reactor, sync, threadpool)
+                cls, name, _threaded_method(sync, name, reactor, threadpool)
             )
         return cls
     return _threaded_class_decorator
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -69,11 +69,6 @@ def auto_threaded(interface, reactor, sync, threadpool):
 
     def _threaded_class_decorator(cls):
         for name in interface.names():
-            if not isinstance(interface[name], Method):
-                raise TypeError(
-                    "auto_threaded does not support interfaces with "
-                    "non-methods attributes"
-                )
             setattr(
                 cls, name, _threaded_method(sync, name, reactor, threadpool)
             )
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -1,24 +1,47 @@
 # FLOC-1549
 
-# from twisted.internet.threads import deferToThreadPool
+from twisted.internet.threads import deferToThreadPool
 
 
-# def _threaded_method(method_name, reactor_name, sync_name, threadpool_name):
-#     def _run_in_thread(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -371,14 +371,84 @@ def run(self, deployer):
...
         return succeed(None)
 
+# FLOC-1549
+
+# class IBlockDeviceAsyncAPI(Interface):
+#     """
+#     Common operations provided by all block devices backends, exposed via
+#     asynchronous methods.
+#     """
+#     def create_volume(dataset_id, size):
```
itamarst: Maybe we can say "dataset" instead of "volume"?

flocker/common/_thread.py
```diff
@@ -0,0 +1,24 @@
...
+#         sync = getattr(self, sync_name)
+#         threadpool = getattr(self, threadpool_name)
+#         original = getattr(sync, method_name)
+#         return deferToThreadPool(reactor, threadpool, original, *args, **kwargs)
+#     return _run_in_thread
+
+
+# def auto_threaded(interface, sync, threadpool):
+#     def _threaded_class_decorator(cls):
+#         for name in interface.names():
```
itamarst: Don't forget about attributes, I guess.

flocker/node/agents/blockdevice.py
```diff
@@ -371,14 +371,84 @@ def run(self, deployer):
...
+#         """
+
+
+# FLOC-1549
+#
+# @implementer(IBlockDeviceAsyncAPI)
+# @auto_threaded(IBlockDeviceAPI, "_reactor", "_sync", "_threadpool")
+# class _SyncToThreadedAsyncAPIAdapter(PRecord):
+#     """
+#     Adapt any ``IBlockDeviceAPI`` to ``IBlockDeviceAsyncAPI`` by running its
```
itamarst: Document the thread-safety expectations of the wrapped class.

itamarst:
Looks good, please proceed.

flocker/node/agents/blockdevice.py
exarkun: I think this is sufficiently internal that we don't need to worry about confusing naive folks by using "volume" here.  Also, in context, "volume" makes a lot more sense because that's what they're mostly called on the backends this code is for interfacing with.  So I'll stick with volume unless you feel strongly.

flocker/node/agents/blockdevice.py
```diff
@@ -372,13 +374,68 @@ def run(self, deployer):
...
+        See ``BlockDeviceAPI.get_device_path``.
+
+        :returns: A ``Deferred`` that fires with a ``FilePath`` for the device.
+        """
+
+
 class IBlockDeviceAPI(Interface):
     """
-    Common operations provided by all block device backends.
+    Common operations provided by all block device backends, exposed, exposed
```
jongiddy: Remove extra "exposed"

flocker/node/agents/blockdevice.py
```diff
@@ -477,6 +534,18 @@ def get_device_path(blockdevice_id):
         """
 
 
+@implementer(IBlockDeviceAsyncAPI)
+@auto_threaded(IBlockDeviceAPI, "_reactor", "_sync", "_threadpool")
+class _SyncToThreadedAsyncAPIAdapter(PRecord):
+    """
+    Adapt any ``IBlockDeviceAPI`` to ``IBlockDeviceAsyncAPI`` by running its
+    methods threads of a thread pool.
```
jongiddy: "methods in thread" :question:

flocker/common/_thread.py
```diff
@@ -0,0 +1,67 @@
...
+        call methods of the object named by ``sync``.
+
+    :return: The class decorator.
+    """
+    def _threaded_class_decorator(cls):
+        for name in interface.names():
+            setattr(
+                cls, name, _threaded_method(sync, name, reactor, threadpool)
+            )
+        return cls
```
jongiddy: This won't work for interface attributes. Either check whether the name refers to a method (tricky to get right in decorators), or document that attributes are not supported.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1164,6 +1196,51 @@ def setUp(self):
...
+        )
+
+
+def make_iblockdeviceasyncapi_tests(blockdeviceasync_api_factory):
+    """
+    :return: A ``TestCase`` with tests that will be performed on the supplied
+        ``IBlockDeviceAsyncAPI`` provider.  These tests are not exhaustive
+        because we currently assume ``make_iblockdeviceapi_tests`` will be used
+        on the wrapped object.
+    """
```
jongiddy: Is there a possibility of wrapping the synchronous tests with  `auto-threaded` to test the async interface fully?

jongiddy:
Added some comments, but someone with more experience of the block device interface should review it too.

flocker/node/agents/test/test_blockdevice.py
exarkun: Probably not.  Since `auto_threaded` changes the interface of the class it decorates (makes it return `Deferred`s instead of synchronous results) any application code that wants to use the decorated class needs to be aware of this.  The tests are all written against a synchronous interface now.  They'll need to be rewritten to account for the new asynchronous interface.

It's possibly the rewritten result will be usable against both `IBlockDeviceAsyncAPI` and `IBlockDeviceAPI` (if we're a little careful in how we rewrite them - but it's kind of easy to support sync APIs in code written for async APIs - at least, easier than supporting async APIs in code written for sync APIs).  I'm not sure yet if that's the way to go.  It's appealing because having a single set of tests to run against all the implementations is a nice way to prove they all have similar behavior and it's probably less code to maintain.  On the other hand, the naive, synchronous assumptions made in the current tests make the failures a little bit nicer to read and debugging a little simpler.

A bridge to cross when we reach it, I think.

flocker/common/_thread.py
```diff
@@ -0,0 +1,81 @@
...
+    for name in interface.names():
+        if not isinstance(interface[name], Method):
+            raise TypeError(
+                "auto_threaded does not support interfaces with non-methods "
+                "attributes"
+            )
+
+    def _threaded_class_decorator(cls):
+        for name in interface.names():
+            if not isinstance(interface[name], Method):
```
wallrj: Why perform this check twice? If there's a reason, add an explanatory comment.

flocker/common/test/test_thread.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def test_failure_result(self):
+        """
+        When the corresponding method on the provider of the interface raises
+        an exception, the ``Deferred`` fires with a ``Failure`` representing
+        that exception.
+        """
+        self.failureResultOf(
+            self.async_spy.method(None, None, None),
+            TypeError
```
wallrj: `TypeError` seems a bit generic. Could be caused by a mistake in the implementation, right? 
Maybe raise and check for a `Spy` specific exception type?

flocker/common/test/test_thread.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+
+    def test_called_in_threadpool(self):
+        """
+        The corresponding method of the provider of the interface is called in
+        a thread using the threadpool specified to ``auto_threaded``.
+        """
+        self.async_spy.method(self.a, self.b, self.c)
```
wallrj: This looks unfinished. Or is there something subtle that I'm missing?

flocker/common/_thread.py
wallrj: And this second check doesn't have test coverage.

wallrj:
Thanks @exarkun 

I forced a build. And the failures look like the usual suspects.

The code looks great but I suppose it'll be difficult to know whether it meets our requirements until we actually switch our `IStateChange` s to handle deferred responses and supply an `IBlockDeviceAPIAsync` to the deployer.

But that'll be clearer in follow up issues.

Please address or answer the comments above and merge if there are no unexpected buildbot failures.

flocker/common/_thread.py
exarkun: Oops.  No reason.  I meant to move this code but apparently I copied it instead.

flocker/common/test/test_thread.py
exarkun: Nope, got me.  I must have just forgotten to finish this.

exarkun:
Thanks @jongiddy and @wallrj.  All points addressed, merging when buildbot catches up.

exarkun:
Well... BuildBot is up to like a hojillion failures... but I guess that's expected at this point.  So.... Guess I'll merge...

PR #1338, opened 2015-04-23
[FLOC-1741] Choose backend.

adamtheturtle:


flocker/node/script.py
```diff
@@ -179,11 +132,47 @@ def get_service(self, reactor, options):
...
 
+def volume_service_from_config(config):
+    pass
+
+def dataset_deployer_from_options(options):
+    """
+    Returns an IDeployer configured by options.
+    """
+    import yaml
+    config = yaml.safe_load(options['config-file'].getContents())
```
tomprince: We probably want to separate out the parsing of the config-file, from building the services. Particularly since we'll want to do validation of the config-file.

flocker/node/script.py
```diff
@@ -221,6 +200,9 @@ def flocker_container_agent_main():
     This starts a Docker-based container convergence agent.
     """
     service_factory = AgentServiceFactory(
+        # TODO this needs to be something which parses Options
+        # to get a hostname, the only necessary parameter for
+        # ApplicationNodeDeployer
```
tomprince: It occurs to me, that we probably want to get the remote ports to connect to, as well as the hostnames from the configuration file, rather than from command-line arguments (and thus have a configuration file shared between the container and dataset agent).

flocker/node/script.py
```diff
@@ -179,11 +132,47 @@ def get_service(self, reactor, options):
         """
         return AgentLoopService(
             reactor=reactor,
-            deployer=self.deployer_factory(hostname=options["hostname"]),
+            deployer=self.deployer_factory(options=options),
```
tomprince: So this should parse the config-file, and pass the options relevant to the deployer into the factory.

docs/advanced/api.rst
```diff
@@ -27,9 +27,34 @@ To configure firewalld to allow access to the control service REST API, and for
...
+
+.. code-block:: yaml
+
+   dataset:
+      backend: loopback
+      loopback-pool: /var/lib/flocker/loopback
+
+To start the agents on a node, (where ``${CONTROL_NODE}`` is the address of the control node,
+``${NODE_NAME}`` is the name of the node being configured),
+and ``${BACKEND}`` is either ``zfs`` or ``loopback``:
```
exarkun: This maybe isn't exactly API documentation.  Sounds more like a "configuring your cluster nodes" document.

docs/advanced/api.rst
```diff
@@ -27,9 +27,34 @@ To configure firewalld to allow access to the control service REST API, and for
...
 
-.. task:: enable_flocker_agent ${NODE_NAME} ${CONTROL_NODE}
+
+.. code-block:: yaml
+
+   node-name:
+   control-service-endpoint: tcp:control-service-hostname:4524
+   dataset:
+      backend: zfs
+      zfs-pool: flocker
```
exarkun: Use explicit strings in all yaml examples.

docs/advanced/api.rst
```diff
@@ -27,9 +27,34 @@ To configure firewalld to allow access to the control service REST API, and for
...
+
+.. Remove this before doing a release (JIRA-XXX)
+
+or if loopback, dataset key is like:
+
+.. code-block:: yaml
+
+   dataset:
+      backend: loopback
```
exarkun: This configuration file needs to be versioned so we can change it easily.

I'm also not sure `zfs` vs `loopback` reflects the right abstraction.  It's probably "peer-to-peer" vs "block-device" with "zfs" and "loopback" as deeper configuration information supplied to each of those backends.

This might not matter, though.  Particularly if the file is versioned such that we can easily make big changes to it.

flocker/node/script.py
```diff
@@ -119,6 +67,11 @@ class DatasetAgentOptions(_AgentOptions):
 
     synopsis = _AgentOptions.synopsis.format("flocker-dataset-agent")
 
+    optParameters = [
+        ["config-file", "c", FilePath("/etc/flocker/dataset-agent.yml"),
```
exarkun: Using `FilePath` as a default value unfortunately makes the help output look dumb.

flocker/node/script.py
exarkun: :+1:

flocker/node/script.py
exarkun: Yep.  I suggest writing a json schema for it.

docs/advanced/api.rst
tomprince: This is where the existing documentation for "configuring your cluster" lives. It should move elsewhere, but perhaps not in this issue.

exarkun: I think it's reasonable to suggest that this issue not expand this problem by dumping a bunch of "hey the cluster has this configuration file and here's what goes in it" stuff into the API docs.

flocker/node/script.py
```diff
@@ -179,11 +132,47 @@ def get_service(self, reactor, options):
...
+        # Later, construction of this object can be moved into
+        # AgentServiceFactory.get_service where various options passed on
+        # the command line could alter what is created and how it is initialized.
+        loopback_pool = config.get('loopback-pool',
+                                   b"/var/lib/flocker/loopback")
+        api = LoopbackBlockDeviceAPI.from_path(loopback_pool)
+        deployer_factory = partial(
+            BlockDeviceDeployer,
+            block_device_api=api,
+        )
```
exarkun:
```
if x == y:
   ...
elif x == z:
   ...
```

is an anti-pattern.  Do some dynamic dispatch (and we're halfway to pluggable backends).

flocker/node/script.py
```diff
@@ -179,11 +132,47 @@ def get_service(self, reactor, options):
...
-            deployer=self.deployer_factory(hostname=options["hostname"]),
+            deployer=self.deployer_factory(options=options),
             host=options["destination-host"], port=options["destination-port"],
         )
 
 
+def volume_service_from_config(config):
+    pass
+
+def dataset_deployer_from_options(options):
```
exarkun: As mentioned elsewhere, the signature of this method should be different.  Perhaps something more like `dataset_deployer_from_options(backend_name, backend_configuration)`

exarkun:
Thanks.  I suggest more English words and less Python code changes for future designs.  Apart from that and inline comments, seems fine.  Please proceed.

docs/advanced/api.rst
tomprince: re: `zfs` vs `loopback`. I think breaking it up into less specific descriptions only makes sense if the more detailed parts can be autodetected. I don't think that is planned for 1.0, so I think that is currently a unnecessary abstraction.

docs/advanced/api.rst
```diff
@@ -27,9 +27,34 @@ To configure firewalld to allow access to the control service REST API, and for
...
+      loopback-pool: /var/lib/flocker/loopback
+
+To start the agents on a node, (where ``${CONTROL_NODE}`` is the address of the control node,
+``${NODE_NAME}`` is the name of the node being configured),
+and ``${BACKEND}`` is either ``zfs`` or ``loopback``:
+
+
+.. task:: enable_flocker_agent ${NODE_NAME} ${CONTROL_NODE} ${BACKEND}
```
jongiddy: NODE_NAME is no longer required in master

admin/acceptance.py
```diff
@@ -282,6 +282,10 @@ class RunOptions(Options):
          'Version of flocker to install'],
         ['build-server', None, 'http://build.clusterhq.com/',
          'Base URL of build server to download RPMs from'],
+        # TODO this requires a buildbot branch which sets the dataset
+        # backend when run-acceptance-tests is run
+        ['dataset-backend', None, None,
+         'The type of dataset backend to use.'],
```
adamtheturtle: Skip the moving tests if the backend is loopback

PR #1340, opened 2015-04-24
[FLOC-1746] Null environment check

dwgebler:
Fixes FLOC-1746


exarkun:
Using `None` to represent an empty environment instead of using `{}` seems like the wrong choice to make in an API, to me.  `None` requires special handling.  `{}` is just another possible dictionary you might have.

However, it seems we already made this decision a long time ago.  We were already making `Unit.environment` `None` if we didn't find any environment variables.  So the above isn't really a review comment about this PR so much as a comment about Docker's recent poor decisions and our past poor decisions.

One thing that is relevant to this PR, though, is whether `image_environment` is going to come back as `None` at some point and trigger essentially the same bug.  Perhaps an image environment can never be empty, though?

Actually... Thinking about what we know about container environments, which is that they are the combination of the image environment and extra things specified at container creation time, if a container environment is `null` then must it not be the case that its image environment is also null?  If the image had any environment variables set, must they not also be set on the container?

It will be interesting to see what happens on buildbot, I guess.

dwgebler:
> if a container environment is null then must it not be the case that its image environment is also null?

There should never be a case where `container_environment` is not `None` but `image_environment` is `None` unless that's the result of a Docker bug, but now you mention it, I suppose there could be a case where `image_environment` is `None` but `container_environment` isn't. I haven't encountered such a case, but it strikes me as realistic that that could happen.

Perhaps `image_environment = image_data[u"Config"]["Env"] if image_data[u"Config"]["Env"] else list()` is worth adding to this branch.

wallrj:
I checked the Env value of my existing busybox image and then removed and re-pulled latest and checked again. The Env no longer includes a path:

```
$ docker inspect busybox:latest | python -c 'import json; import sys; print json.load(sys.stdin)[0]["Config"]["Env"]'
[u'PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin']

$ docker rmi busybox
Untagged: busybox:latest

$ docker pull busybox:latest
Pulling repository busybox
8c2e06607696: Download complete 
cf2616975b4a: Download complete 
6ce2e90b0bc7: Download complete 
Status: Downloaded newer image for busybox:latest

$ docker inspect busybox:latest | python -c 'import json; import sys; print json.load(sys.stdin)[0]["Config"]["Env"]'
None

```

flocker/node/_docker.py
```diff
@@ -564,11 +564,15 @@ def _list():
                 image_data = self._client.inspect_image(image)
                 unit_environment = []
                 container_environment = data[u"Config"][u"Env"]
```
wallrj: When I run `docker inspect busybox:latest` I get a `Config` and `ContainerConfig` keys...but perhaps that's a peculiarity of `docker inspect` which can inspect both images and containers.

flocker/node/test/test_docker.py
```diff
@@ -170,6 +170,23 @@ def got_list(units):
...
+            self.addCleanup(client.remove, name)
+            d = client.add(name, u"busybox:latest")
+            d.addCallback(lambda _: client.list())
+
+            def got_list(units):
+                unit = [unit for unit in units if unit.name == name][0]
+                self.assertIsNone(unit.environment)
+            d.addCallback(got_list)
+            return d
+
```
wallrj: This test looks identical to the one above. Can't it be shared somehow?

flocker/node/_docker.py
```diff
@@ -564,11 +564,15 @@ def _list():
...
-                image_environment = image_data[u"Config"]["Env"]
-                for environment in container_environment:
-                    if environment not in image_environment:
-                        env_key, env_value = environment.split('=', 1)
-                        unit_environment.append((env_key, env_value))
+                image_environment = (
+                    image_data[u"Config"]["Env"]
+                    if image_data[u"Config"]["Env"] is not None else list()
```
wallrj: When I first looked at this I assumed it was a list comprehension...can you reformat it as a normal
```
if ...:
    ...
else:
    ...
```

...statement.

wallrj:
When I create a new image based on the latest `busybox:latest` it somehow gains a PATH environment variable.
Perhaps that's what happens when you have a base image.

This https://github.com/docker/docker/pull/8827 looks relevant.

`scratch` (since Docker 1.5) is no longer a real image, so if the latest `busybox:latest` image was built with > Docker 1.5 there won't be a base image so the busybox image no-longer gains the PATH environment variable.

flocker/node/_docker.py
exarkun: It was speculated on `#docker` that `ContainerConfig` is stuff inherited by a container that is started with this image.  That makes sense for at least some of the keys present...  It looks like at least for some data the two sections don't contain exactly the same information, but I've only observed a deviation for `Cmd`.  Perhaps it doesn't really matter for this PR, though.

flocker/node/_docker.py
exarkun: Yes, please do (and I think we should stop using `if ... else ...` entirely).

exarkun: Also `[]` is the idiomatic new-empty-list

exarkun:
> $ docker inspect busybox:latest | python -c 'import json; import sys; print json.load(sys.stdin)[0]["Config"]["Env"]'
> None

fyi, there's `jshon`:

```
$ docker inspect busybox | jshon -e 0 -e "Config" -e "Env"
[
 "PATH=\/usr\/local\/sbin:\/usr\/local\/bin:\/usr\/sbin:\/usr\/bin:\/sbin:\/bin"
]
```

flocker/node/functional/test_docker.py
```diff
@@ -367,6 +367,23 @@ def added(_):
...
 
+    def test_null_environment(self):
+        """
+        A container that does not include any environment variables contains
+        an empty ``environment`` in the return ``Unit``.
+        """
+        client = self.make_client()
+        name = random_name()
+        self.addCleanup(client.remove, name)
+        d = client.add(name, u"busybox:latest")
```
exarkun: Using an image we control and can be sure actually has a null environment would make this test less likely to fail in the future when the busybox maintainers design to do something reasonable (like put some environment variables into their image).

flocker/node/functional/test_docker.py
```diff
@@ -367,6 +367,23 @@ def added(_):
...
+        d.addCallback(lambda _: client.list())
+
+        def got_list(units):
+            unit = [unit for unit in units if unit.name == name][0]
```
exarkun: :question: 

`list(generator expression)` is almost always better than `[list comprehension]` if for no other reason than it doesn't leak a name into the local scope (here the `unit` local gets bound once per loop iteration and then it gets bound again at the end to the actual result).

flocker/node/test/test_docker.py
exarkun: Doesn't just look identical.... I think it is identical.  It even has the same name.  Accidental duplication, probably?  Just needs to be deleted?

flocker/node/_docker.py
```diff
@@ -564,11 +564,15 @@ def _list():
...
                 container_environment = data[u"Config"][u"Env"]
-                image_environment = image_data[u"Config"]["Env"]
-                for environment in container_environment:
-                    if environment not in image_environment:
-                        env_key, env_value = environment.split('=', 1)
-                        unit_environment.append((env_key, env_value))
+                if image_data[u"Config"]["Env"] is not None:
+                    image_environment = image_data[u"Config"]["Env"]
+                else:
+                    image_environment = []
```
exarkun: :question: 

All other things equal, it's easier to read code without `not` than with it.  Cases like this, you can easily avoid using `not` by flipping the order of the blocks.

flocker/node/functional/test_docker.py
dwgebler: Have added a custom image we can build at run time in the functional test, in the way we do with a couple of other tests. This is a small busybox-ish image but guaranteed to have no environment variables set.

flocker/node/functional/blank-docker/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM scratch
+MAINTAINER Jérôme Petazzoni <jerome@docker.com>
```
exarkun: Probably not :wink:

flocker/node/functional/blank-docker/Dockerfile
```diff
@@ -0,0 +1,4 @@
...
+ADD rootfs.tar /
```
exarkun: Mmm I'm not excited about this.  Does the container actually need a full copy of the busybox filesystem?  Also, not a fan of binary artifacts checked in to version control.  Whatever we need to build, we should build from source.  If there's non-trivial build overhead, we can think about amortizing it by having Docker hub build it for us automatically.

The only behavior we need from this container is for it to be runnable.  You can run a container from an empty image - though it will fail (because you must run a command and it has no command).  I'm not sure if that failure would be a problem here, maybe not.  Worth a try.  Otherwise 

(Also it's not clear what the license of rootfs.tar is)

flocker/node/functional/test_docker.py
```diff
@@ -367,6 +367,26 @@ def added(_):
         d.addCallback(added)
         return d
 
+    def test_null_environment(self):
+        """
+        A container that does not include any environment variables contains
+        an empty ``environment`` in the return ``Unit``.
+        """
+        docker_dir = FilePath(__file__).sibling('blank-docker')
+        image = DockerImageBuilder(test=self, source_dir=docker_dir)
```
exarkun: Think about having Docker hub build this for us.

exarkun:

dwgebler:
Added https://registry.hub.docker.com/u/clusterhq/buildroot/
This is a custom micro image (c. 1.3MB) based on a Buildroot compilation (see http://buildroot.uclibc.org/). The container is nothing other than the raw `rootfs.tar` generated by running Buildroot's `make`

flocker/node/functional/test_docker.py
```diff
@@ -367,6 +367,24 @@ def added(_):
         d.addCallback(added)
         return d
 
+    def test_null_environment(self):
+        """
+        A container that does not include any environment variables contains
+        an empty ``environment`` in the return ``Unit``.
+        """
+        image_name = u"clusterhq/buildroot:latest"
```
adamtheturtle: In another branch @exarkun has used an image which is specific to a particular test. The relevant Dockerfile lives besides the test. See the discussion about the cost / benefits of this on https://github.com/ClusterHQ/flocker/pull/1281/files and do similar.

dwgebler: The last review comment from @exarkun was to use an image hosted on Docker Hub, see previous discussion on this PR.

exarkun:
Some test errors, seemingly related:

```
docker.errors.APIError: 500 Server Error: Internal Server Error ("Cannot destroy container flocker--NamespacedDockerClientTests-flocker-node-functional-test_docker-NamespacedDockerClientTests-test_null_environment-795064561587--581768466033: Unable to remove filesystem for 751808d16e324d645e9d3b4d04803257171db0e06472d12c014aa39b7f32d681: remove /var/lib/docker/containers/751808d16e324d645e9d3b4d04803257171db0e06472d12c014aa39b7f32d681/secrets: device or resource busy")
```

Most likely due to ... unfortunate features of Docker rather than egregious logic errors on our part... but that probably doesn't mean we get to ignore them.

`flocker.node.functional.test_deploy.DeployerTests.test_environment` hung up several builders:

  * http://build.clusterhq.com/builders/flocker-coverage/builds/5702/steps/trial/logs/stdio
  * http://build.clusterhq.com/builders/flocker-zfs-head/builds/3594/steps/trial/logs/stdio
  * http://build.clusterhq.com/builders/flocker-fedora-20/builds/5815/steps/trial/logs/stdio
  * http://build.clusterhq.com/builders/flocker-ubuntu-14.04/builds/2847/steps/trial/logs/stdio

Considering that's a test for environment variable use and this is a branch changing environment variable handling, these probably do need to be explained (and ideally fixed) before merging.  Please investigate that and share your results.  Thanks.

dwgebler:
The 500 API error is specific to CentOS / RHEL and seems to be a Docker bug, possibly fixed in 1.6:

`Cannot destroy container flocker--NamespacedDockerClientTests-flocker-node-functional-test_docker-NamespacedDockerClientTests-test_null_environment-9153648524--923235904791: Unable to remove filesystem for 9fa5ff7c2ff1f5060e9441097a45f8cfc2074bd2a6be215c0df3c8ec8a80772a: remove /var/lib/docker/containers/9fa5ff7c2ff1f5060e9441097a45f8cfc2074bd2a6be215c0df3c8ec8a80772a/secrets: device or resource busy`

https://bugzilla.redhat.com/show_bug.cgi?id=1209872

>Basically in case of failure we were leaving secrets directory mounted on tmpfs and that's why removal of that directory was failing.

The test that gave that 500 error passes on non-RHEL builders e.g. Fedora.

The hanging thing is happening on other branches, I don't know what's causing that but it's not specific to the changes in this branch.

http://build.clusterhq.com/builders/flocker-zfs-head/builds/3580/steps/trial/logs/stdio

flocker/node/functional/test_docker.py
```diff
@@ -367,6 +367,39 @@ def added(_):
...
+        client = self.make_client()
+        name = random_name()
+        if isinstance(client, NamespacedDockerClient):
+            container_name = client._client._to_container_name(name)
+            client._client._client.create_container(
+                name=container_name, image=image_name)
+        else:
+            container_name = client._to_container_name(name)
+            client._client.create_container(
+                name=container_name, image=image_name)
```
exarkun: Instead of this `isinstance` check, add a new hook to `GenericDockerClientTests`: `create_container(client, name, image)`.  The first branch of this conditional is the implementation of that method for `NamespacedDockerClientTests`.  The second branch of this conditional is the implementation for `GenericDockerClientTests`.

exarkun:
Thanks.  Address final review point and then merge.

PR #1341, opened 2015-04-24
[FLOC-1753] Fix issues with missing images and removal of in-use images

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1753

Also did some other test suite and code improvements while we were at it.


flocker/node/_docker.py
```diff
@@ -561,7 +563,18 @@ def _list():
...
+                try:
+                    image_data = self._client.inspect_image(image)
+                except APIError as e:
+                    if e.response.status_code == NOT_FOUND:
+                        # Image has been deleted, so just fill in some
+                        # stub data so we can return *something*. This
+                        # should happen only for stopped containers so
+                        # some inaccuracy is acceptable.
+                        image_data = {u"Config": {u"Env": []}}
+                    else:
```
dwgebler: Also see https://github.com/ClusterHQ/flocker/pull/1340 - as in that case, as a temporary fix for a problem, I think this is fine, but also strongly suggests we need to re-think our approach in the `list` method and how we're representing default and missing information about containers and states from Docker.

flocker/node/functional/test_docker.py
```diff
@@ -334,11 +366,13 @@ def test_pull_image_if_necessary(self):
         """
         The Docker image is pulled if it is unavailable locally.
         """
-        image = u"busybox:latest"
+        # Use an image that isn't likely to be in use by anything, since
+        # it's old, and isn't used by other tests:
+        image = u"busybox:ubuntu-12.04"
```
dwgebler: :question: "isn't likely to be in use by anything" is a bit flakey; we could have a custom image that can be pulled and exists only for the specific purpose of being used by this test. That might be overkill though, up to you if you want to leave it as out-of-date busybox.

dwgebler:
Your `test_list_with_missing_image` isn't going to pass on Buildbot until #1340 is merged. But unless you want to address the question I left on `pull_image_if_necessary`, please merge once Buildbot is in a state where these two tests are passing and any other build failures that might happen are determined to be unrelated to this branch.

flocker/node/functional/test_docker.py
itamarst: Jean-Paul has better fix in #1281, mostly I just wanted to make it less likely so I could run my own tests better. Main goal was fixing 404s.

exarkun:
Please hold off on merging this.

flocker/node/_docker.py
```diff
@@ -534,6 +534,8 @@ def _list():
                     # have been removed in another thread.
                     if e.response.status_code == NOT_FOUND:
                         continue
+                    else:
+                        raise
```
exarkun: This appears to be untested.

exarkun: Actually, I guess that's probably a consequence of having the majority of the docker functional tests fail early with that `Env` error.  I'll come back to this after that `Env` branch is merged.

itamarst: It's untested, yes; mostly just a bugfix I noticed in passing. In order to test it though I'd need to... trigger an error in Docker? seems difficult to do in practice. Or monkeypatch I guess. If you think it's worth doing that I can add a est.

itamarst:
Merged forward, let's see if Buildbot turns green.

flocker/node/_docker.py
exarkun: I think it's worth testing all of our code paths, yes. :smile:

flocker/node/_docker.py
```diff
@@ -543,14 +543,26 @@ def _list():
...
-                if image_command[u"Config"][u"Cmd"] == command:
+                try:
+                    image_data = self._client.inspect_image(image)
+                except APIError as e:
+                    if e.response.status_code == NOT_FOUND:
+                        # Image has been deleted, so just fill in some
+                        # stub data so we can return *something*. This
+                        # should happen only for stopped containers so
+                        # some inaccuracy is acceptable.
+                        image_data = {u"Config": {u"Env": [], u"Cmd": []}}
```
exarkun: I observed that this can happen even for running containers.  Docker lets you untag an image that is being used by a running container:

```
(flocker)jean-paul@dilaton:~$ docker run -d busybox:latest sleep 60000
097f8f3f629b1b626e3482264eaf179c833f4e5dda95924610c43119f20f8ed7
(flocker)jean-paul@dilaton:~$ docker inspect -f "{{.Config.Image}}" 097f8f3f629b1b626e3482264eaf179c833f4e5dda95924610c43119f20f8ed7
busybox:latest
(flocker)jean-paul@dilaton:~$ docker rmi busybox:latest
Error response from daemon: Conflict, cannot delete 4986bf8c1536 because the running container 097f8f3f629b is using it, stop it and use -f to force
FATA[0000] Error: failed to remove one or more images   
(flocker)jean-paul@dilaton:~$ docker tag busybox:latest busybox:1.2.3.4
(flocker)jean-paul@dilaton:~$ docker rmi busybox:latest
Untagged: busybox:latest
(flocker)jean-paul@dilaton:~$ docker inspect -f "{{.Config.Image}}" 097f8f3f629b1b626e3482264eaf179c833f4e5dda95924610c43119f20f8ed7
busybox:latest
(flocker)jean-paul@dilaton:~$ docker inspect busybox:latest
Error: No such image or container: busybox:latest
[]
(flocker)jean-paul@dilaton:~$ 
```

Note that the container still reports it is using `busybox:latest` even after that tag ceases to exist locally.  Even worse, the tag can reappear (perhaps associated with a different image - for example, because you did a `docker pull` and the `latest` tag moved...) and inspect **still** claims it's running that tag... which now does exist but refers to the wrong image!  (And therefore possibly the wrong environment.)

I can't tell if these terrible cases are the ones we're actually encountering in the test suite because the test failures don't include enough information to actually understand what leads to them (yes, removing some image failed ... because it doesn't exist ... but who knows who created the image or who removed it later or why there's still a container using it).

They seem like issues pretty likely to matter in production use - at the very least it seems like this could lead to spurious application restarts (on top of new images!) .  Please file a follow-up issue for addressing them.

exarkun: Please also add logging here so we have more information available to debug bizarre behavior related to these cases in the future.

flocker/node/functional/test_docker.py
```diff
@@ -205,6 +205,38 @@ def test_dead_is_listed(self):
...
+        happens for non-running containers, who at worst we're going to
+        restart anyway.
+        """
+        path = FilePath(self.mktemp())
+        path.makedirs()
+        path.child(b"Dockerfile.in").setContent("FROM busybox\n")
+        image_name = DockerImageBuilder(test=self, source_dir=path,
+                                        cleanup=False).build()
+        name = random_name()
+        d = self.start_container(unit_name=name, image_name=image_name,
```
exarkun: Does this container ever get cleaned up?

flocker/node/functional/test_docker.py
```diff
@@ -205,6 +205,38 @@ def test_dead_is_listed(self):
                                  expected_states=(u'inactive',))
         return d
 
+    def test_list_with_missing_image(self):
+        """
+        ``DockerClient.list()`` can list containers whose image is missing.
+
+        The resulting output may be inaccurate, but that's OK: this only
```
exarkun: As described elsewhere, it's not actually limited to happening for non-running containers.

flocker/node/functional/test_docker.py
exarkun: I wouldn't mind if you merged that branch forward and resolved the conflicts after this branch is merged.

exarkun: Or you could drop these changes from this branch before merging.

flocker/testtools/__init__.py
```diff
@@ -609,7 +613,11 @@ def build(self, dockerfile_variables=None):
             docker_dir.path
         ]
         run_process(command)
-        self.test.addCleanup(run_process, [b"docker", b"rmi", tag])
+        if self.cleanup:
+            def remove_image():
+                client = DockerClient()
+                client.remove_image(tag, force=True)
```
exarkun: If this removed all containers using this image and then removed the image, the result would probably be more useful.  This would let all users of `DockerImageBuilder` (with `cleanup=True`) ignore container cleanup and have the right thing happen.  It would also avoid leaving dangling broken containers on the system.

File a follow-up issue and refer to it from here if you don't want to implement this behavior now.

flocker/node/_docker.py
```diff
@@ -543,14 +543,26 @@ def _list():
...
+                    else:
+                        raise
```
exarkun: Add test coverage here as well, please.

exarkun:
Any chance you can explain why these images are missing?

itamarst:
The DockerImageBuilder class builds an image, and then does a cleanup where it removes images it created. And apparently sometimes a container using those images gets left behind.

exarkun:
> The DockerImageBuilder class builds an image, and then does a cleanup where it removes images it created. And apparently sometimes a container using those images gets left behind.

In master it does `docker rmi ...` which refuses to remove an image in use by a container.

itamarst:
To give the timeline more clearly:

1. Image created.
2. Container runs.
3. Container stops.
4. Image removed.

At this point let's say test suite crashes. There is now a Docker container on the system that will show up in listing API, but whose image is missing.

flocker/node/functional/test_docker.py
```diff
@@ -205,6 +205,38 @@ def test_dead_is_listed(self):
...
+        happens for non-running containers, who at worst we're going to
+        restart anyway.
+        """
+        path = FilePath(self.mktemp())
+        path.makedirs()
+        path.child(b"Dockerfile.in").setContent("FROM busybox\n")
```
exarkun: This `Dockerfile` appears to be insufficient to actually create a new image.  What it actually does is create a new tag of the busybox image!  This exercises a different case than you intended, I believe (as long as there are multiple tags of an image it seems Docker is always happy to let you remove one of them, even if a running container is using it).  I think you need at least one more statement to create at least one layer so you get a distinct image.

exarkun:
Yep, I get that.  What I'm trying to get at is that this sequence doesn't appear possible:

```
jean-paul@dilaton:~$ mkdir /tmp/buildanimage
jean-paul@dilaton:~$ echo "FROM busybox                                                                    MAINTAINER alice@example.com" > /tmp/buildanimage/Dockerfile
jean-paul@dilaton:~$ docker build -t junk:rmidemo /tmp/buildanimage/
Sending build context to Docker daemon 2.048 kB
Sending build context to Docker daemon 
Step 0 : FROM busybox
 ---> 7287e8cc7f93
Step 1 : MAINTAINER alice@example.com
 ---> Running in 18de03ca840b
 ---> 7694457ddc02
Removing intermediate container 18de03ca840b
Successfully built 7694457ddc02
jean-paul@dilaton:~$ docker run -d junk:rmidemo /bin/true
3766aa3b5b373c26d8bcafdfd5a10a8237116dfd1f02d5f2364f1ed5d32911f5
jean-paul@dilaton:~$ docker inspect -f '{{.State.Running}}' 3766aa3b5b373c26d8bcafdfd5a10a8237116dfd1f02d5f2364f1ed5d32911f5
false
jean-paul@dilaton:~$ docker rmi junk:rmidemo
Error response from daemon: Conflict, cannot delete 7694457ddc02 because the container 3766aa3b5b37 is using it, use -f to force
FATA[0000] Error: failed to remove one or more images   
jean-paul@dilaton:~$ 
```

If the test suite attempted to follow the sequence you outlined, it would have encountered an error removing the image.  It would not have been able to produce a state where the image was deleted but the container still exists (even stopped).

flocker/node/_docker.py
itamarst: I believe we can get the underling Docker image hash from the inspect results, so we could use that instead, solving the tagging issue. I will try to do that.

flocker/node/functional/test_docker.py
itamarst: I think if we inspect based on image hash, rather than tag, it is limited to non-runnong containers. So I'll try to do that.

flocker/node/functional/test_docker.py
itamarst: Yeah, `start_container` has a cleanup.

flocker/node/functional/test_docker.py
itamarst: I feel like this is less likely to fail, so may as well leave it in. And we still have #1281 to actually fix the problem later on.

itamarst:
Maybe older versions of Docker were different?

itamarst:
I have replied and/or addressed in code all comments I think.

itamarst:
Oh, forgot logging.

flocker/node/_docker.py
exarkun: Any luck with that?  Otherwise, add a link to the issue here.

exarkun:
Thanks.  I'm not really happy with the force removals of images.  I'm pretty sure the need for that just means we have other bugs hiding out somewhere.  I think "Maybe older versions of Docker were different?" is a minimally convincing argument for why these changes seem to be necessary.

Do you think this actually fixes the problem or is it just a bandaid?

exarkun:
Basically out of time for this today.  Once the logging changes are in, if buildbot looks good and you're happy with how this looks, go ahead and merge.  Thanks.

itamarst:
It's just better cleanup, basically. So I don't think it's any worse.

wallrj:
Not sure if it's relevant, but I noticed that `docker rmi`  on my laptop uses docker API version 1.17 while we force docker-py to use 1.16 ...I think.

PR #1343, opened 2015-04-28
S3 packages new branch floc 1716

adamtheturtle:
Do not review. This is to avoid git hacks.


admin/release.py
```diff
@@ -460,16 +482,110 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+    """
+    packages = yield Effect(
+        ListS3Keys(bucket=target_bucket,
+                   prefix='python/'))
+
+    index_path = create_pip_index(
+        scratch_directory=scratch_directory,
+        packages=packages)
+
+    yield Effect(
```
adamtheturtle: Why recursively? I'll change this.

admin/release.py
```diff
@@ -460,16 +482,110 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+    :param list packages: List of bytes, filenames of packages to be in the
+        index.
+    """
+    index_file = scratch_directory.child('index.html')
+    with index_file.open('w') as f:
+        # Although this returns a Deferred, it works without the reactor
+        # because there are no Deferreds in the template evaluation.
+        # See this cheat described at
+        # https://twistedmatrix.com/documents/15.0.0/web/howto/twisted-templates.html
+        # TODO: convert the Twisted Deferred into an Effect?
```
adamtheturtle: Don't have TODOs in PRs. If you prefer change to "XXX" with a link to an issue.

admin/release.py
```diff
@@ -460,16 +482,110 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
         )
 
 
-def publish_rpms_main(args, base_path, top_level):
+packages_template = (
```
adamtheturtle: :question: Is this nicer than https://pypi.python.org/pypi/html/, apart from not introducing another dependency?

exarkun: `html` doesn't claim Python 2.7 support on PyPI and the last release is from four years ago.  Also, now that I've looked at the examples, I'm a little skeptical of it (`escape=False`?  Terrible).

admin/release.py
```diff
@@ -326,18 +339,30 @@ def publish_docs_main(args, base_path, top_level):
 
 class UploadOptions(Options):
     """
-    Options for uploading packages.
+    Options for uploading artifacts.
     """
     optParameters = [
         ["flocker-version", None, flocker.__version__,
-         "The version of Flocker to upload packages for."],
+         "The version of Flocker to upload RPMs for."
```
adamtheturtle: Now that the deb repo branch has been merged this is no longer correct - change it back to "packages".

PR #1345, opened 2015-04-28
[FLOC-1770] New version of Eliot

itamarst:


tomprince:
This looks good. The failing tests were on the dev-box, so merge once that builds green (or with an unrelated failure). Also, an issue should be filed to replace uses of `Logger`.

tomprince:
The tests look like they are now failing for other reasons, so I'll merge.

PR #1346, opened 2015-04-28
[FLOC-1733] Use node UUID in flocker-deploy and HTTP API

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1733


flocker/cli/script.py
```diff
@@ -127,8 +127,14 @@ def fail(msg):
...
             if response.code != OK:
                 d = json_content(response)
-                d.addCallback(
-                    lambda error: fail(error[u"description"] + u"\n"))
+
+                def got_error(error):
+                    if isinstance(error, dict):
+                        error = error[u"description"] + u"\n"
+                    else:
+                        error = u"Unknown error: " + unicode(error) + "\n"
```
wallrj: This will be an error raised by json_content, right?
Wouldn't it be more useful to have an error handler one level up that printed the full response (including response code and raw response.text) ?

wallrj: So flocker-deploy continues to post deployment.yml content containing only host IP addresses...

flocker/control/httpapi.py
```diff
@@ -812,11 +818,15 @@ def replace_configuration(self, applications, deployment):
...
         """
+        # XXX we're going to have to deal with UUIDs better,
+        # somehow... look up state of cluster? Change deployment
+        # configuration format?
         try:
             configuration = FigConfiguration(applications)
             if not configuration.is_valid_format():
                 configuration = FlockerConfiguration(applications)
             return self.persistence_service.save(model_from_configuration(
+                self.cluster_state_service.as_deployment(),
```
wallrj: The cluster state now gets passed to `model_from_configuration` so that supplied IP addresses can be matched to node_uuids.
What if agents have not yet reported their state?

flocker/control/httpapi.py
```diff
@@ -812,11 +818,15 @@ def replace_configuration(self, applications, deployment):
         :param deployment: Configuration of which applications run on
             which nodes.
         """
+        # XXX we're going to have to deal with UUIDs better,
+        # somehow... look up state of cluster? Change deployment
+        # configuration format?
```
wallrj: Perhaps you've already answered this question?

flocker/control/httpapi.py
wallrj: Would be easier to read if this was a keyword argument....to make it clear that we're passing a `DeploymentState` not `Deployment` ...despite the method name `as_deployment`.

flocker/control/_config.py
```diff
@@ -1283,7 +1295,7 @@ def model_from_configuration(applications, deployment_configuration):
     :returns: A ``Deployment`` object.
     """
     nodes = deployment_from_configuration(
-        deployment_configuration, applications)
+        deployment_state, deployment_configuration, applications)
     return Deployment(nodes=frozenset(nodes))
```
wallrj: `model_from_configuration` seems a bit pointless now.
Why can't we ditch it and have `deployment_from_configuration` return a `Deployment`? ....which would make more sense, given its name.

flocker/control/_config.py
```diff
@@ -1257,7 +1262,10 @@ def deployment_from_configuration(deployment_configuration, all_applications):
                         hostname=hostname, application_name=name)
                 )
             node_applications.append(application)
-        node = Node(hostname=hostname,
+        if hostname not in node_states:
+            raise ConfigurationError("No known node with address {}.".format(
+                hostname))
```
wallrj: Ah, I see. That's nice.
So this is somewhat related to https://clusterhq.atlassian.net/browse/FLOC-1278 `Check that the primary node a dataset is created on is actually part of the cluster`

itamarst: 1278 should be probably just be thrown out.

flocker/control/_config.py
itamarst: Yeah, but I think it's time to switch to "do things as quickly as possible" and do cleanup later.

flocker/control/_config.py
wallrj: So `Node` has only `uuid` while `NodeState` still has a `hostname`....makes sense.

flocker/control/httpapi.py
itamarst: Yeah, that should just be deleted.

flocker/control/_model.py
```diff
@@ -418,18 +416,16 @@ def __invariant__(self):
                     return (False, '%r manifestation is not on node' % (app,))
         return (True, "")
 
-    def __new__(cls, **kwargs):
```
wallrj: I didn't understand this change and when I tested it, I got this:

```
In [3]: Node(hostname='192.0.2.123')
Out[3]: Node(applications=ApplicationPSet([]), manifestations=UnicodeManifestationPMap({}), uuid=e55fc05f-4842-f369-d48f-87d1c81cf251)

In [4]: Node(hostname='192.0.2.123')
Out[4]: Node(applications=ApplicationPSet([]), manifestations=UnicodeManifestationPMap({}), uuid=e55fc05f-4842-f369-d48f-87d1c81cf251)

In [5]: Node()
AttributeError                            Traceback (most recent call last)
<ipython-input-5-5ae91e5d73cc> in <module>()
----> 1 Node()

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.pyc in __new__(cls, hostname, **kwargs)
    424                  " tests only. If you see this in production code that's "
    425                  "a bug.", DeprecationWarning, stacklevel=2)
--> 426             kwargs["uuid"] = ip_to_uuid(hostname)
    427         return PRecord.__new__(cls, **kwargs)
    428 

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.pyc in ip_to_uuid(ip)
    622     :return UUID: Matching UUID.
    623     """
--> 624     return UUID(bytes=md5(ip.encode("utf-8")).digest())
    625 
    626 

AttributeError: 'NoneType' object has no attribute 'encode'

```

flocker/control/_model.py
```diff
@@ -418,18 +416,16 @@ def __invariant__(self):
...
+    def __new__(cls, hostname=None, **kwargs):
         # PRecord does some crazy stuff, thus _precord_buckets; see
         # PRecord.__new__.
         if "uuid" not in kwargs and "_precord_buckets" not in kwargs:
             warn("UUID is required, this is for backwards compat with existing"
                  " tests only. If you see this in production code that's "
                  "a bug.", DeprecationWarning, stacklevel=2)
```
wallrj: Well can't we just look for these warnings when running the test suite, fix those tests and then remove this. Or are we planning to do that in a followup issue. If so which issue?

wallrj: And can we remove the similar tricks from `NodeState` in this branch? It looks like you've made the necessary changes to always supply a hostname to `NodeState`.

wallrj: In [2]: NodeState()
KeyError                                  Traceback (most recent call last)
<ipython-input-2-c81282e22251> in <module>()
----> 1 NodeState()

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.pyc in __new__(cls, **kwargs)
    659                  " tests only. If you see this in production code that's "
    660                  "a bug.", DeprecationWarning, stacklevel=2)
--> 661             kwargs["uuid"] = ip_to_uuid(kwargs["hostname"])
    662         return PRecord.__new__(cls, **kwargs)
    663 

KeyError: 'hostname'

flocker/control/httpapi.py
wallrj: So the `Deployment` instance is returned and contains `Node`'s whose UUIDs are generated from the supplied IP addresses.
That `Deployment` replaces the existing configuration and the connected agents are notified of a configuration change.

Next the convergence agents on each node need to know the UUID of the node that they're running on so that they can identify which parts of the configuration concern them...and the agents get their UUID from the flocker configuration file whose location is defined in the systemd service file...

wallrj: No, I'm wrong, the agent *used* to get its local hostname from the command line.

And that seems still to be in the documentation for the zfs agent, but not for the container or dataset agents...
```
(1733)(UUID-in-API-FLOC-1733 ? =)[~/.../HybridLogic/flocker]$ flocker-zfs-agent --help
Usage: flocker-zfs-agent [OPTIONS] <local-hostname> <control-service-hostname>
```

Maybe update the help text for `flocker-zfs-agent`.

wallrj: Ok. They work out their own IP address...for now: https://clusterhq.atlassian.net/browse/FLOC-1751

wallrj: And then convert the IP to UUID using the same function as used by `model_from_configuration`...eventually UUID will be extracted from the certificate though:
 * https://clusterhq.atlassian.net/browse/FLOC-1783

flocker/node/script.py
```diff
@@ -198,7 +198,7 @@ def get_service(self, reactor, options):
         ip = _get_external_ip(host, port)
         return AgentLoopService(
             reactor=reactor,
-            # Temporary hack, to be fixed in FLOC-1733 probably:
+            # Temporary hack, to be fixed in FLOC-1783 probably:
             deployer=self.deployer_factory(node_uuid=ip_to_uuid(ip),
```
wallrj: `IDeployer` needs to know the UUID of the node that it's running on....but that bit has already been done.

flocker/node/functional/test_deploy.py
```diff
@@ -32,12 +32,13 @@ class P2PNodeDeployer(object):
     https://clusterhq.atlassian.net/browse/FLOC-1732
     """
     def __init__(self, hostname, volume_service, docker_client=None,
-                 network=None):
+                 network=None, node_uuid=None):
```
wallrj: How many tests are left that *don't* supply a `node_uuid`? It seems like you've updated most of them in this branch.
If so, why not just make this a required parameter and remove the warning from `ApplicationNodeDeployer.__init__` and other IDeployer initialisers?

flocker/node/test/test_deploy.py
```diff
@@ -1159,19 +1161,47 @@ def test_proxy_needs_creating(self):
...
         expected = sequentially(changes=[SetProxies(ports=frozenset([proxy]))])
         self.assertEqual(expected, result)
 
+    def test_no_proxy_if_node_state_unknown(self):
+        """
+        ``ApplicationNodeDeployer.calculate_changes`` does not attempt to
+        create a proxy to a node whose state is unknown, since the
+        destination IP is unavailable.
+        """
+        api = ApplicationNodeDeployer(u'192.168.1.1', node_uuid=uuid4(),
```
wallrj: Would it be a pain to use documentation `192.0.2.x` addresses in this branch:
 * https://tools.ietf.org/html/rfc5737

wallrj: I wonder how the routing will work if a node has multiple IP addresses / routes?
The IP used for communicating with the control service won't necessarily be the same as the IP used for communicating with other nodes.

flocker/node/test/test_deploy.py
```diff
@@ -1159,19 +1161,47 @@ def test_proxy_needs_creating(self):
...
+                                      docker_client=FakeDockerClient(),
+                                      network=make_memory_network())
+        expected_destination_port = 1001
+        port = Port(internal_port=3306,
+                    external_port=expected_destination_port)
+        application = Application(
+            name=b'mysql-hybridcluster',
+            image=DockerImage(repository=u'clusterhq/mysql',
```
wallrj: @exarkun always complains when I use `mysql` in test examples.

flocker/node/test/test_deploy.py
```diff
@@ -1159,19 +1161,47 @@ def test_proxy_needs_creating(self):
...
+                              tag=u'release-14.0'),
+            ports=frozenset([port]),
+        )
+        desired = Deployment(nodes=[Node(uuid=uuid4(),
+                                         applications=[application])])
+        result = api.calculate_changes(
+            desired_configuration=desired, current_cluster_state=EMPTY_STATE)
+        expected = sequentially(changes=[])
+        self.assertEqual(expected, result)
```
wallrj: /me looks for a corresponding change in `ApplicationNodeDeployer`...

flocker/node/_deploy.py
```diff
@@ -742,8 +744,10 @@ def calculate_changes(self, desired_configuration, current_cluster_state):
                     for port in application.ports:
                         # XXX: also need to do DNS resolution. See
                         # https://clusterhq.atlassian.net/browse/FLOC-322
-                        desired_proxies.add(Proxy(ip=node.hostname,
-                                                  port=port.external_port))
+                        if node.uuid in node_states:
+                            desired_proxies.add(Proxy(
+                                ip=node_states[node.uuid].hostname,
+                                port=port.external_port))
```
wallrj: Ok. So is this a new bug or is there an existing issue for this?

And if the node subsequently connects to the control service, the proxy will then be configured.

Does the same thing apply to zfs send and  receive state changes?

wallrj: I looked at `find_dataset_changes`, and it looks like there might be a potential `KeyError` in 
```
    # Look at each dataset that is going to be running elsewhere and is
    # currently running here, and add a DatasetHandoff for it to `going`.
    going = set()
    for dataset_node_uuid, desired in desired_datasets.items():
        if dataset_node_uuid != uuid:
            hostname = uuid_to_hostnames[dataset_node_uuid]
            for dataset in desired:
                if dataset.dataset_id in local_current_dataset_ids:
                    going.add(DatasetHandoff(
                        dataset=dataset, hostname=hostname))

```

...but I can't be sure.

flocker/control/httpapi.py
```diff
@@ -201,6 +201,8 @@ def create_dataset_configuration(self, primary, dataset_id=None,
         if metadata is None:
             metadata = {}
 
+        primary = UUID(hex=primary)
```
wallrj: That's nice. I'll remember to use the `hex` keyword in future.

flocker/control/httpapi.py
```diff
@@ -220,8 +222,7 @@ def create_dataset_configuration(self, primary, dataset_id=None,
         )
         manifestation = Manifestation(dataset=dataset, primary=True)
 
-        primary_node = deployment.get_node(ip_to_uuid(primary),
-                                           hostname=primary)
+        primary_node = deployment.get_node(primary)
```
wallrj: I find it really confusing that `deployment.get_node` should return a new `Node` if the supplied uuid isn't among the current nodes....but it looks like it was always like that and the hostname would only ever have been used if a new Node was created.

flocker/control/httpapi.py
```diff
@@ -397,9 +398,9 @@ def state_datasets(self):
         deployment = self.cluster_state_service.as_deployment()
         datasets = list(datasets_from_deployment(deployment))
         for dataset in datasets:
-            uuid = ip_to_uuid(dataset[u"primary"])
+            node_uuid = UUID(hex=dataset[u"primary"])
             dataset[u"path"] = self.cluster_state_service.manifestation_path(
-                uuid, dataset[u"dataset_id"]).path.decode(
```
wallrj: Maybe it'd be nice to update the parameter name too `manifestation_path(node_uuid...`

docs/advanced/api_examples.yml
```diff
@@ -355,7 +357,7 @@
     "move container"
 
   doc: |
-    Move an existing named container (and any attached datasets) to a new host.
+    Move an existing named container (and any attached datasets) to a new node.
 
   request: |
     POST /v1/configuration/containers/webserver HTTP/1.1
```
wallrj: The "host" below should now be "node_uuid" I think.

wallrj: http://build.clusterhq.com/results/docs/UUID-in-API-FLOC-1733/build-6253/advanced/api.html#post--v1-configuration-containers-%28name%29

wallrj:
Thanks @itamarst 

I haven't kept up with the preceding UUID work so it was interesting to see how it all works.
I think I'm done. Sorry the review took so long though.
Please merge forward, answer or address the comments above and then merge when your happy with the build results.

flocker/cli/script.py
itamarst: Well, it's slightly nicer UI when you have a message rather than just HTTP response... and this is written so I'll just leave it in.

flocker/control/_model.py
itamarst: `Node` doesn't have a hostname anymore... but needs to accept one for backwards compat with existing tests. I will see how many such tests are left.

itamarst: A bunch, so I'll file followup issue to remove that.

flocker/control/_model.py
itamarst: Hostname is required for `NodeState` still. There's enough tests that fail when backwards compat is removed that I will do this in followup issue.

flocker/node/functional/test_deploy.py
itamarst: Sadly, lots.

flocker/node/test/test_deploy.py
itamarst: Eh, it's internal tests, not going to bother (but will try to remember in the future). Re routing... it's a problem, yes. Real solution is using a better routing system.

flocker/node/_deploy.py
itamarst: Yeah I think ZFS logic is buggy in this regard. I'll fix it.

itamarst: This is an issue introduced by fact `Node` no longer has a hostname.

PR #1347, opened 2015-04-28
[FLOC-1485] OpenStack CinderBlockAPI attach

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1485

This required the use of the Nova API for attaching a volume to a server.
The Cinder API is still required in order to add metadata to a volume.

The tests assume Rackspace authentication and attach related tests will only pass on a Rackspace VM.

Tests can be run as follows:

```
sudo /usr/bin/env CLOUD_CONFIG_FILE=$HOME/acceptance.yml $(type -p trial) flocker.node.agents.functional.test_cinder.CinderBlockDeviceAPIInterfaceTests
```

Sample output from a real test run of current HEAD:

```
(env-flocker)[exarkun@richardw-buildslave-floc-1368 develop]$ sudo /usr/bin/env CLOUD_CONFIG_FILE=~richard/acceptance.yml $(type -p trial) -j 4 --testmodule flocker/node/agents/cinder.py
Running 33 tests.
  CinderBlockDeviceAPIInterfaceTests
    test_attach_unattached_volume ...                                      [OK]
    test_compute_instance_id_nonempty ...                                  [OK]
    test_compute_instance_id_unicode ...                                   [OK]
    test_attach_attached_volume ...                                        [OK]
    test_attach_elsewhere_attached_volume ...                              [OK]
    test_foreign_cluster_volume ...                                        [OK]
    test_foreign_volume ...                                                [OK]
    test_attached_volume_listed ...                                        [OK]
    test_get_device_path_device_repeatable_results ...                     [OK]
    test_get_device_path_device ...                                        [OK]
    test_list_attached_and_unattached ...                                  [OK]
    test_resize_destroyed_volume ...                                     [FAIL]
    test_resize_unknown_volume ...                                       [FAIL]
    test_resize_volume_listed ...                                        [FAIL]
flocker.node.agents.functional.test_cinder_behaviour
  VolumesCreateTests
    test_create_metadata_is_listed ...                                     [OK]
  VolumesSetMetadataTests
    test_updated_metadata_is_listed ...                                    [OK]
  CinderBlockDeviceAPIInterfaceTests
    test_detach_volume ...                                                 [OK]
    test_multiple_volumes_attached_to_host ...                             [OK]
    test_reattach_detached_volume ...                                      [OK]

...

Ran 33 tests in 195.037s

FAILED (failures=3, successes=30)
```

(resize functionality is coming in a future PR)


add_github_user
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
```
wallrj: Ignore this. I'll delete it before the branch is merged.

flocker/node/agents/cinder.py
```diff
@@ -98,22 +116,64 @@ def wait_for_volume(volume_manager, expected_volume,
...
+    prefix = 'instance-'
+    name = check_output(['xenstore-read', 'name']).rstrip()
+    return UUID(name[len(prefix):])
+
+
+def _next_device():
+    """
+    Can't just use the dataset name as the block device name
+    inside the node, nor volume.id nor random_name. You can't
+    even leave it blank; auto is not supported.
```
wallrj: This was copied from the original openstack spike: https://github.com/ClusterHQ/flocker/compare/openstack-spike-FLOC-1147

flocker/node/agents/cinder.py
```diff
@@ -131,17 +191,20 @@ def create_volume(self, dataset_id, size):
...
             size=Byte(size).to_GB().value,
             metadata=metadata,
         )
-        created_volume = wait_for_volume(self.volume_manager, requested_volume)
+        created_volume = wait_for_volume(self.cinder_volume_manager, requested_volume)
         # So once the volume has actually been created, we set the metadata
         # again. One day we hope this won't be necessary.
         # See Rackspace support ticket: 150422-ord-0000495'
-        self.volume_manager.set_metadata(created_volume, metadata)
+        self.cinder_volume_manager.set_metadata(created_volume, metadata)
```
wallrj: Only the Cinder API is capable of setting volume metadata, so we need it as well as the Nova API.

flocker/node/agents/cinder.py
```diff
@@ -150,19 +213,51 @@ def list_volumes(self):
...
     def resize_volume(self, blockdevice_id, size):
         pass
 
     def attach_volume(self, blockdevice_id, host):
-        pass
+        """
+        The attaching may have to be done via the nova client :-(
+        See http://www.florentflament.com/blog/openstack-volume-in-use-although-vm-doesnt-exist.html # noqa
+
+        When I attach using the cinder client the volumes become undetachable.
```
wallrj: Yes, this was exactly the case.
Docstring needs updating.

flocker/node/agents/cinder.py
```diff
@@ -150,19 +213,51 @@ def list_volumes(self):
...
+        """
+        unattached_volume = self._get(blockdevice_id)
+        if unattached_volume.host is not None:
+            raise AlreadyAttachedVolume(blockdevice_id)
+
+        device_path = _next_device()
```
wallrj: You have to supply a device path.
Which means that this API *must* be run on the host to which volumes are being attached.
Which makes me wonder if there's any point having the `host` argument of `IBlockDeviceAPI.attach`?

flocker/node/agents/cinder.py
```diff
@@ -150,19 +213,51 @@ def list_volumes(self):
...
+        nova_volume = self.nova_volume_manager.create_server_volume(
+            # Nova API expects an ID string not UUID.
+            server_id=unicode(server_id), 
+            volume_id=unattached_volume.blockdevice_id, 
+            device=device_path
+        )
+        wait_for_volume(
+            volume_manager=self.cinder_volume_manager,
+            expected_volume=nova_volume,
+            expected_status=u'in-use',
```
wallrj: The volume status `in-use` seems to only be set once the volume has been assigned a device inside the host operating system...at least on Rackspace.
So the assertion below is never triggered.
In our spike, we had a loop which waited for the device path to appear instead.
Perhaps that'd be a more reliable test. 
Perhaps we should do both.

flocker/node/agents/cinder.py
```diff
@@ -189,59 +284,70 @@ def _is_cluster_volume(cluster_id, cinder_volume):
...
-def _blockdevicevolume_from_cinder_volume(cinder_volume):
+def _blockdevicevolume_from_cinder_volume(cinder_volume, host_map):
     """
     :param Volume cinder_volume: The ``cinderclient.v1.volumes.Volume`` to
         convert.
     :returns: A ``BlockDeviceVolume`` based on values found in the supplied
         cinder Volume.
     """
+    if cinder_volume.attachments:
+        # There should only be one.
```
wallrj: Cinder and Nova `Volume` instances have an *attachments* attribute which says to me that a volume can be attached to multiple hosts.
But I'm not sure if that's actually possible.
And it certainly shouldn't happen with Flocker managed volumes.

flocker/node/agents/cinder.py
```diff
@@ -189,59 +284,70 @@ def _is_cluster_volume(cluster_id, cinder_volume):
...
         blockdevice_id=unicode(cinder_volume.id),
         size=int(GB(cinder_volume.size).to_Byte().value),
-        host=None,
+        host=flocker_host,
         dataset_id=UUID(cinder_volume.metadata[DATASET_ID_LABEL])
     )
 
 
-def rackspace_cinder_client(**kwargs):
+def rackspace_session(**kwargs):
```
wallrj: The same session can be supplied to both Nova and Cinder clients, so I updated this factory function.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -39,9 +39,20 @@ def cinderblockdeviceapi_for_test(test_case, cluster_id):
         by ``TidyCinderVolumeManager`` to cleanup any lingering volumes that
         are created during the course of ``test_case``
     """
+    local_instance_uuid = _instance_uuid()
+    # Create the nova client last so that its cleanup method is
+    # run first and detaches volumes before the cinder client
+    # cleanup then deletes them.
+    # Yuck!
```
wallrj: I should probably combine these into a single test helper that calls the cleanup operations in the correct order.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -39,9 +39,20 @@ def cinderblockdeviceapi_for_test(test_case, cluster_id):
...
+    cinder_client = tidy_cinder_client_for_test(test_case)
+    nova_client = tidy_nova_client_for_test(test_case)
     return cinder_api(
-        cinder_client=tidy_cinder_client_for_test(test_case),
+        cinder_client=cinder_client,
+        nova_client=nova_client,
         cluster_id=cluster_id,
+        host_map={
+            u'192.0.2.123': local_instance_uuid,
```
wallrj: Fortunately, our current tests only ever attach to a single host.
The attach_elsewhere_attached test fortunately attaches to *this* host first and then some other host, but that fails because of the early check for a `BlockDeviceVolume.host` value in `CinderBlockDeviceAPI.attach`

flocker/node/agents/testtools.py
```diff
@@ -80,14 +84,82 @@ def setUp(self):
...
+        def setUp(self):
+            self.client = client_factory(test_case=self)
+
+    return Tests
+
+
+def client_from_environment(client_factory):
     """
-    Create a ``cinderclient.v1.client.Client`` using credentials from a config
+    Create an openstack API client using credentials from a config
```
wallrj: It actually returns a client_factory, curried  with the   correct `session` and `region` arguments.
That way it can be used for creating both Nova and Cinder clients whose `version` arguments are different.

setup.py
```diff
@@ -162,6 +162,7 @@
         "python-cinderclient==1.1.1",
         "python-keystoneclient-rackspace==0.1.3",
         "bitmath==1.2.3-4",
+        "python-novaclient==2.24.1",
```
exarkun: Might be nice to group/comment the OpenStack dependencies together.

flocker/node/agents/cinder.py
exarkun: This approach has a race condition.  If attach is called twice, both calls will discover the same device name to use with the Nova API call.  Unfortunately the device namespace is shared mutable state.  To avoid having separate attach calls conflict, you probably need to track this shared mutable state from the system in some shared mutable state in Python.  And, ugh, thread safety and locks and so forth.  Sorry.

flocker/node/agents/functional/test_cinder.py
exarkun: That would be reasonable.

flocker/node/agents/testtools.py
```diff
@@ -117,7 +191,21 @@ def tidy_cinder_client_for_test(test_case):
...
+    """
+    Return a ``novaclient.v2.client.Client`` whose ``volumes`` API is a
+    wrapped by a ``TidyNovaVolumeManager`` and register a ``test_case``
+    cleanup callback to detach any volumes that are attached during each test.
+    """
+    client_factory = client_from_environment(client_factory=NovaClient)
+    client = client_factory(version=2)
+    client.volumes = TidyNovaVolumeManager(client.volumes)
+    test_case.addCleanup(client.volumes._cleanup)
+    return client
```
exarkun: Yea I guess doing attach and detach at the same time might have saved some work.

flocker/node/agents/cinder.py
```diff
@@ -98,22 +116,64 @@ def wait_for_volume(volume_manager, expected_volume,
...
+    # But to satisfy the current IBlockDeviceAPI.list API we'll need
+    # to match the OpenStack instance_uuid to the hostname (or soon,
+    # the Flocker Node UUD). How will we do that?
+    # We don't really want this implementation to have to query the
+    # OpenStack servers list and even if it could get that list, how
+    # is it going to match Flocker node UUIDs to OpenStack
+    # instance_uuid.
+    # Perhaps BlockDeviceVolume.host should have the OpenStack
+    # instance_uuid and there should be an extra
+    # IBlockDeviceAPI.is_local_volume method (or something)
```
exarkun: I think we will add a new interface method, probably `IBlockDeviceAPI.what_is_this_nodes_cloudy_identifier`, and make `BlockDeviceVolume` include a `cloudy_identifier` to represent what it's attached to.

This is enough information to let `BlockDeviceDeployer` figure out which volumes are local.

flocker/node/agents/cinder.py
exarkun: Let's hold off until there's a demonstrated need.

exarkun:
Thanks.

As you noted, this goes somewhat beyond design.  Sometimes that's unavoidable though, so no worries - just keep the goal of the design phase in mind (minimize time spent going in the wrong direction).

Apart from the inline comments this seems reasonable.  If you think it will save time, consider doing the detach work here too (seems like if you don't do that you need lots of extra test scaffolding).

Let me know if you want to go over thread safety stuff wrt that race condition.

flocker/node/agents/cinder.py
wallrj: See https://clusterhq.atlassian.net/browse/FLOC-1809

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -0,0 +1,192 @@
...
+            'using the CLOUD_CONFIG_FILE environment variable. '
+            'See: '
+            'https://docs.clusterhq.com/en/latest/gettinginvolved/acceptance-testing.html '  # noqa
+            'for details of the expected format.'
+        )
+
+    with open(config_file_path) as config_file:
+        config = safe_load(config_file.read())
+
+    section = config[provider.name]
```
tomprince: I'd like to reuse the acceptance test configuration for this, which uses `rackspace` as the top-level key here (which doesn't match `openstack` which is the provider name). But I think that can be addressed in a followup ticket, when we provision a rackspace slave to test this on.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -69,15 +62,16 @@ def test_foreign_volume(self):
         Test that ``list_volumes`` lists only those volumes
         belonging to the current Flocker cluster.
         """
-        ec2_client = ec2_client_from_environment()
+        cls, kwargs = get_blockdeviceapi_args(ProviderType.aws)
```
tomprince: If test code is going to be calling `get_blockdeviceapi_args` directly, then that function should probably be raising `SkipTest` directly.

flocker/node/agents/test/blockdevicefactory.py
exarkun: One constraint on us right now, unfortunately, is that our PistonCloud testig environment is ... shared.  I think we need to avoid sending *all* of our API keys to that slave.  I think that means we need to send a subset of the acceptance.yml.  Maybe we could have a quick discussion about our options for this sometime today.

tomprince: Sure, that makes sense. My concern was more that for environments that *already* have the acceptance testing credentials, that the same file (and thus format) be used. That doesn't mean that all environments need all that information. 

This also means that we could change the format to be more supportive of the functional testing case, so long as we adjust the code for the acceptance tests at the same time.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -0,0 +1,192 @@
...
+    Get initializer arguments suitable for use in the instantiation of an
+    ``IBlockDeviceAPI`` implementation compatible with the given provider.
+
+    :param provider: A provider type the ``IBlockDeviceAPI`` is to be
+        compatible with.  A value from ``ProviderType``.
+
+    :raises: ``ConfigMissing`` if a ``CLOUD_CONFIG_FILE`` was not set and the
+        default config file could not be read.
+
+    :return: A ``dict`` that can initialize the matching implementation.
```
tomprince: This appears to return a tuple.

flocker/common/_thread.py
```diff
@@ -60,17 +92,122 @@ def auto_threaded(interface, reactor, sync, threadpool):
...
+                response=e.response.text,
+            ).write()
+            raise
+    return _run_with_logging
+
+
+def auto_openstack_logging(interface, original):
+    """
+    Create a class decorator which will add thread-based asynchronous versions
+    of all of the methods on ``interface``.
```
tomprince: copy-pasta

flocker/node/agents/testtools.py
```diff
@@ -4,60 +4,14 @@
 Test helpers for ``flocker.node.agents``.
```
tomprince: I'm not clear on the distinction between this and `.test.blockdevicefactory`. I wonder if this should be a package with both these as submodules?

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -0,0 +1,192 @@
...
+#
+#
+# vvvvvvvvvvvvvvvvvvvv testing helper that actually belongs in this module
+
+
+def get_blockdeviceapi_with_cleanup(test_case, provider):
+    """
+    Instantiate an ``IBlockDeviceAPI`` implementation appropriate to the given
+    provider and configured to work in the current environment.  Arrange for
+    all volumes created by it to be cleaned up at the end of the current test
```
tomprince: It looks like the cleanup isn't implemented yet.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -17,53 +17,42 @@
...
+
+def cinderblockdeviceapi_for_test(test_case):
     """
     Create a ``CinderBlockDeviceAPI`` instance for use in tests.
 
     :param TestCase test_case: The test being run.
-    :param UUID cluster_id: The Flocker cluster ID for Cinder volumes.
     :returns: A ``CinderBlockDeviceAPI`` instance whose underlying
         ``cinderclient.v1.client.Client`` has a ``volumes`` attribute wrapped
         by ``TidyCinderVolumeManager`` to cleanup any lingering volumes that
```
tomprince: It looks like the implementation details are going to change here.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -0,0 +1,192 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Functionality for creating ``IBlockDeviceAPI`` providers suitable for use in
+the current execution environment.
```
tomprince: It would be easier to review this, if the changes leading to this module were separated out to be reviewed separately.

flocker/node/agents/cinder.py
```diff
@@ -138,49 +175,132 @@ def create_volume(self, dataset_id, size):
...
+                expected_volume=requested_volume,
+            )
+        return _blockdevicevolume_from_cinder_volume(
+            cinder_volume=requested_volume,
         )
-        created_volume = wait_for_volume(self.volume_manager, requested_volume)
-        # So once the volume has actually been created, we set the metadata
-        # again. One day we hope this won't be necessary.
-        # See Rackspace support ticket: 150422-ord-0000495'
-        self.volume_manager.set_metadata(created_volume, metadata)
```
tomprince: This has been fixed upstream.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -17,53 +17,42 @@
...
-@skip_except(
-    supported_tests=[
-        'test_interface',
-        'test_created_is_listed',
-        'test_created_volume_attributes',
-        'test_list_volume_empty',
-        'test_listed_volume_attributes',
-    ]
-)
 class CinderBlockDeviceAPIInterfaceTests(
```
tomprince: It seems like `test_attach_elsewhere_attached_volume` can't obviously be run as-is against a real backend here.

tomprince:
It would be really nice if this were split into (at least) 3 PRs:
- changing the interface between the tests and buildbot
- the logging changes
- the attach changes themselves.

tomprince:
It looks like there is some changes around cleaning up volumes after the test that got missed, as well as a couple of other minor comments.

Overall, the changes look sensible, but there are too many unrelated changes here for me to keep in my head which bits of each have been reviewed.

exarkun:
Split into #1388, #1389, #1390

PR #1348, opened 2015-04-28
[FLOC-1775] Deal with ignorance of containers or datasets

itamarst:


flocker/control/test/test_httpapi.py
```diff
@@ -2769,6 +2769,18 @@ def test_empty(self):
...
         )
 
+    def test_unknown_datasets(self):
+        """
+        When the cluster state is ignorant about datasets on a node, the
+        endpoint does not list information for that node.
+        """
+        self.cluster_state_service.apply_changes([
+            NodeState(hostname=u"192.0.2.101", uuid=uuid4(),
+                      manifestations=None)])
```
exarkun: Perhaps set paths to `None` as well since one of these will never be known without the other.

exarkun:
Thanks.  Code changes look pretty good.  I think this merits some some words in the API documentation.  Seeing some nodes go missing from these lists will be surprising and mysterious and should be explained so people don't think it's just due to random bugs in our software.

Merge when that's done and buildbot is however green it can be expected to be.

PR #1349, opened 2015-04-29
[FLOC-1762] Implement IBlockDeviceAPI.{create_volume, list_volumes,  destroy_volume} for EBS

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1762

Implement IBlockDeviceAPI's {create_volume, list_volumes, destroy_volume} APIs and associated tests for EBS backend.


myechuri:
Pushed a new change to allow reading CLOUD_PROVIDER from env so that both test_ebs.py and test_cinder.py can run in buildbot. Running test_ebs.py in Openstack will be a no-op. Running test_cinder.py in AWS will be a no-op.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+from uuid import UUID
+
+from boto import ec2
+from pyrsistent import PRecord, field
+from zope.interface import implementer
+
+from .blockdevice import IBlockDeviceAPI, BlockDeviceVolume, UnknownVolume
+
+DATASET_ID_LABEL = u'flocker-dataset-id'
+METADATA_VERSION_LABEL = u'flocker-metadata-version'
```
wallrj: That's a good idea!

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+CLUSTER_ID_LABEL = u'flocker-cluster-id'
+
+
+def ec2_client(region, zone, access_key_id, secret_access_key):
+    """
+    Establish connection to EC2 client.
+
+    :param String region: The name of the EC2 region to connect to.
```
wallrj: I think the type here and below should be `str` or `unicode`....`unicode` probably.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+    :param String zone: The zone for the EC2 region to connect to.
```
wallrj: Do you need to specify both region *and* zone?
I don't see that in our acceptance testing config format: http://doc-dev.clusterhq.com/gettinginvolved/acceptance-testing.html#aws
Or is it a block device thing?
Maybe add an explanatory note somewhere.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+                    region,
+                    aws_access_key_id=access_key_id,
+                    aws_secret_access_key=secret_access_key))
+
+
+class _EC2(PRecord):
+    """
+    :ivar string zone: The name of the zone for the connection.
+    :ivar EC2Connection connection: Object representing connection to
+        an EC2 instance.
```
wallrj: Maybe spell the type in full `boto.ec2.connection.EC2Connection` https://github.com/boto/boto/blob/develop/boto/ec2/__init__.py#L56

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    zone = field()
+    connection = field()
```
wallrj: I was going to say mark these as `required` but apparently that's the default: https://github.com/tobgu/pyrsistent#mandatory-fields

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    An EBS implementation of ``IBlockDeviceAPI`` which creates
+    block devices in an EC2 cluster using Boto APIs.
+    """
+    def __init__(self, ec2_client, cluster_id):
+        """
+        Initialize EBS block device API instance.
+
+        :param ``_EC2`` ec2_client: PRecord storing EC2
+            client connection and EC2 instance zone.
```
wallrj: Probably just need to specify the type and say something like "A record of EC2 connection and zone."

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+        :param String cluster_id: UUID of cluster for this
+            API instance.
+        """
+        self.connection = ec2_client.connection
+        self.zone = ec2_client.zone
+        self.cluster_id = cluster_id
+
+    def _blockdevicevolume_from_ebs_volume(self, ebs_volume):
```
wallrj: I've been following the rule-of-thumb that helper functions, like this, can be needn't be instance methods unless they need access to self...which this one doesn't. That way they are more likely to be re-used / refactored for use elsewhere.
On the other hand it's nice to keep all the helpers together. Up to you.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+        return BlockDeviceVolume(
+            blockdevice_id=unicode(ebs_volume.id),
+            size=int(GB(ebs_volume.size).to_Byte().value),
+            host=None,
+            dataset_id=UUID(ebs_volume.tags[DATASET_ID_LABEL])
+        )
+
+    def _wait_for_volume(self, expected_volume,
+                         expected_status=u'available',
+                         time_limit=60):
```
wallrj: Do you think we should create a general `wait_for` function and pass it another function that gets called in a loop until it returns something truthy?
@itamarst already wrote `loop_until` but I think that only works with deferred APIs....let me check that.

wallrj: Hmm. Yeah you can't pass blocking functions to `loop_until`...or you can, but they'll block the reactor.

wallrj: And `loop_until` doesn't have a time limit...it probably should.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        Helper function to wait for up to 60s for given volume
+        to be in 'available' state.
```
wallrj: In the OpenStack branch I also found this useful for waiting for volumes to be `in-use` (attached).
I guess the status labels are different for EBS and perhaps that's another reason for writing a general blocking version of loop_until that accepts a `predicate` function:
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/testtools/__init__.py#L195

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+        as volume tag data.
+
+        :param String dataset_id: Dataset_id for the volume.
+        :param int size: Requested volume size in Bytes.
+
+        :return BlockDeviceVolume volume: Created volume in
+            BlockDeviceVolume format.
+        """
+        requested_volume = self.connection.create_volume(
+            size=int(Byte(size).to_GB().value), zone=self.zone)
```
wallrj: Ah that's where the zone is used.https://github.com/boto/boto/blob/develop/boto/ec2/connection.py#L2276

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # Stamp created volume with Flocker-specific tags.
+        metadata = {
+            METADATA_VERSION_LABEL: '1',
+            CLUSTER_ID_LABEL: unicode(self.cluster_id),
+            DATASET_ID_LABEL: unicode(dataset_id),
+        }
+        self.connection.create_tags([requested_volume.id],
+                                    metadata)
```
wallrj: Huh, so you have to tag the volume separately with EBS too.
Cinder allows metadata to be supplied when you create a volume, but I found it didn't work.
I wonder if we should think about what happens if the volume gets created but the `create_tag` API call fails.
Should we attempt to delete the volume? 
Or should we retry the tag operation?
With cinder you can choose the volume name which would allow you to identify volumes that had been only partially created, but boto doesn't seem to allow you to choose the name.
We should discuss with @exarkun and @itamarst

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # Wait for created volume to reach 'available' state.
+        self._wait_for_volume(requested_volume)
```
wallrj: And here again, I wonder if we should attempt to remove the created volume if we reach the time limit?

wallrj: (I haven't done in the Cinder branch, but I wonder if I should)

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def list_volumes(self):
+        """
+        Return all volumes in {available, in-use} state that belong to
+        this Flocker cluster.
+        """
+        volumes = []
+        for ebs_volume in self.connection.get_all_volumes():
+            if ((self._is_cluster_volume(self.cluster_id, ebs_volume)) and
+               (ebs_volume.status in [u'available', u'in-use'])):
```
wallrj: Why do we care about the status here?
What other status might they be in?
I'd have thought we want to return *all* volumes that appear to belong to the cluster and if their status is unexpected that indicates a bug in our code.
Oh, perhaps there's a "deleting" status...I haven't had to deal with that yet....let me read on.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+    def destroy_volume(self, blockdevice_id):
+        """
+        Destroy EBS volume identified by blockdevice_id.
+
+        :param String blockdevice_id: EBS UUID for volume to be destroyed.
+
+        :raises UnknownVolume: If there does not exist a Flocker cluster
+            volume identified by input blockdevice_id.
+        :raises Exception: If we failed to destroy Flocker cluster volume
+            corresponding to input blockdevice_id.
```
wallrj: This isn't part of the `IBlockDeviceAPI` interface. 
Maybe we should just log this situation.
Another deletion will be attempted by the next convergence loop...I think.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+        """
+        for volume in self.list_volumes():
+            if volume.blockdevice_id == blockdevice_id:
+                ret_val = self.connection.delete_volume(blockdevice_id)
```
wallrj: Under what circumstances might this return False?
I guess it probably happens if you attempt to destroy an attached volume.
But does that mean that `IBlockDeviceAPI.destroy_volume` should raise something like `UnavailableVolume` in this case?
@exarkun, what do you think?

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,198 @@
...
+                if ret_val is False:
+                    raise Exception(
+                        'Failed to delete volume: {!r}'.format(blockdevice_id)
+                    )
+                else:
+                    return
+        raise UnknownVolume(format(blockdevice_id))
```
wallrj: `format` ...I never saw that before! But I think this should probably be `unicode`.

flocker/node/agents/ebs.py
myechuri: We want to only list volumes that are available for us to use. Possible states for an EBS volume to be in are "creating | available | in-use | deleting | deleted | error". 
i ran into 'deleting' status while running a test that deletes a volume and verifies that it is no longer listed in "list_volumes" output. 
Do you think this sounds reasonable? Thanks!

flocker/node/agents/functional/test_cinder.py
```diff
@@ -71,14 +69,17 @@ class CinderBlockDeviceAPIInterfaceTests(
...
+    def test_foreign_cluster_volume(self):
+        """
+        Volumes from other Flocker clusters are not listed.
+        """
+        block_device_api2 = cinderblockdeviceapi_for_test(
+            test_case=self,
+            cluster_id=uuid4(),
+        )
 
+        self.assertVolumesDistinct(block_device_api2)
```
wallrj: Well if we move these tests here, we may as well figure out how to share them between Cinder and EBS test cases...or maybe even make them part of the IBlockDeviceAPI tests and add the cluster ID to the LoopBackBlockDevice API too.

wallrj: We could just add them to a mixin class and mix that into the Cinder and EBS tests for now...and figure our if / how to run them against the LoopbackBlockDeviceAPI later.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -844,6 +844,15 @@ class IBlockDeviceAPITestsMixin(object):
...
     Tests to perform on ``IBlockDeviceAPI`` providers.
     """
+    def tearDown(self):
+        """
+        Destroy volumes created for the cluster used by the test.
+        """
+        volumes = self.api.list_volumes()
+
+        for volume in volumes:
+            self.api.destroy_volume(volume.blockdevice_id)
```
wallrj: I think this approach breaks the LoopBackBlockDeviceAPI tests which have a separate cleanup function that removes all the backing files before this gets run...so the destroy_volume fails.

I don't know. I thought the aim was to be able to run tests for `create_volume` in isolation...without exercising `destroy_volume`...which is why I made `TidyCinderClient` to cleanup test volumes at a lower level.

One advantage of doing it that way is that we can measure the coverage of the just the `create_volume` related tests.

But @itamarst talked about this sort of thing when he was in Bristol last month....that our tests should exercise only our public APIs and supply realistic inputs and outputs...so I'd be interested to know what he thinks about this.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1337,6 +1350,29 @@ def test_resize_destroyed_volume(self):
         )
         self.assertEqual(exception.args, (volume.blockdevice_id,))
 
+    def assertVolumesDistinct(self, block_device_api2):
```
wallrj: Should be `assert_volumes_distinct`

wallrj: And perhaps this would be easier to read if it was a free function...something like

`assert_volumes_distinct(test_case, api1, api2)`

flocker/node/agents/testtools.py
```diff
@@ -80,15 +82,13 @@ def setUp(self):
     return Tests
 
 
-def cinder_client_from_environment():
+def load_client_config_from_environment():
```
wallrj: Ah, yeah. This makes more sense.

flocker/node/agents/ebs.py
itamarst: I suggest filing a follow-up issue for this case (and dealing with it post-1.0).

flocker/node/agents/ebs.py
itamarst: Ditto.

flocker/node/agents/testtools.py
```diff
@@ -104,11 +104,46 @@ def cinder_client_from_environment():
...
+    """
+    Create an EC2 Boto client using credentials from a config
+    file path which may be supplied as an environment variable.
+    See ``load_config`` for details on where config is populated from.
+
+    :returns: An instance of EC2 Boto client
+        using provider specific credentials found in ``CLOUD_CONFIG_FILE``.
+    """
+    config = load_client_config_from_environment()
+    provider_name = os.environ.get('CLOUD_PROVIDER')
```
wallrj: Perhaps this bit could be moved into `load_client_config_from_environment` and supply that function with a provider name.

wallrj:
Thanks @myechuri 

This looks great.

I added one or two comments above.

Please respond to those and make any changes you agree with and resubmit for another review.

I still haven't actually run the tests. I'll have a go tomorrow.

flocker/node/agents/ebs.py
exarkun: Maybe we should make the interface explicit about the fact that we won't try to destroy attached volumes.  If we do, the implementation is free to do whatever it likes since we've violated the interface.

flocker/node/agents/ebs.py
myechuri: Thanks for pointing out! Yaml load does read it as str, changed to str.

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py
myechuri: Thanks for educating! Moved the helper functions out of the class scope.

flocker/node/agents/ebs.py
myechuri: Ah ok. i did look into combining wait_for for EBS and Cinder, but that was getting complicated. If we come across a third usecase, will do the refactoring. Hope that is ok!

flocker/node/agents/ebs.py
myechuri: EBS does have 'in-use' state, but the two current usecases for _wait_for_volume (with default target state of 'available') are in create_volume path, where we do want to wait for it to be in 'available' state before declaring success. Since we are not multi-threaded at the moment, there is no possibility of the volume changing state to in-use before the current create-volume returns. Please correct me if this is not the case?

flocker/node/agents/ebs.py
myechuri: Yup!

flocker/node/agents/ebs.py
myechuri: Does this qualify for a JIRA item, or, is it too fine-grained to warrant JIRA? If not, i'll just make a note to self to revisit past 1.0. Thanks!

flocker/node/agents/ebs.py
myechuri: Since EBS is returning useful information that the destroy actually failed, is there a way for us to make use of this information instead of waiting for next convergence loop? @itamarst @exarkun 
Thanks!

flocker/node/agents/ebs.py
myechuri: EBS volume destroy can fail for several reasons we cannot predict. If we dig down http response body, we might be able to gather exact reason. Therefore, UnavailableVolume is just one of the possible reasons for failure, it might be hard to put together an exhaustive list of all possible reasons.

flocker/node/agents/ebs.py

flocker/node/agents/test/test_blockdevice.py
myechuri: Renamed.

flocker/node/agents/functional/test_cinder.py
myechuri: Reworked test_foreign_cluster_volume() and moved it into test_blockdevice.py. Please let me know what you think? Thanks!

flocker/node/agents/testtools.py
myechuri: Reworked this part. Please let me know what you think? Thanks!

myechuri:
Thanks a lot for the review, Richard! Addressed all review comments except destroy_volume() issue, which is still under discussion, and updated pull request. Please let me know if i am missing anything for next review round. Thanks!

flocker/node/agents/ebs.py
myechuri: Filed https://clusterhq.atlassian.net/browse/FLOC-1792 to track post-1.0 and added comment in ebs.py pointing to the created PR.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+        representing connection to an EC2 instance.
+    """
+    zone = field(mandatory=True)
+    connection = field(mandatory=True)
+
+
+def _blockdevicevolume_from_ebs_volume(ebs_volume):
+    """
+    Helper function to convert Volume information from
+    EBS format to Flocker block device format.
```
wallrj: Add parameter and return type docs.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+        dataset_id=UUID(ebs_volume.tags[DATASET_ID_LABEL])
+    )
+
+
+def _wait_for_volume(expected_volume,
+                     expected_status=u'available',
+                     time_limit=60):
+    """
+    Helper function to wait for up to 60s for given volume
+    to be in 'available' state.
```
wallrj: Add parameter and return type docs.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+                    expected_volume.status, elapsed_time,
+                    time_limit
+                )
+            )
+
+
+def _is_cluster_volume(cluster_id, ebs_volume):
+    """
+    Helper function to check if given volume belongs to
+    given cluster.
```
wallrj: Add parameter and return type docs.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+class EBSBlockDeviceAPI(object):
+    """
+    An EBS implementation of ``IBlockDeviceAPI`` which creates
+    block devices in an EC2 cluster using Boto APIs.
+    """
+    def __init__(self, ec2_client, cluster_id):
+        """
+        Initialize EBS block device API instance.
+
+        :param ``_EC2`` ec2_client: A record of EC2 connection and zone.
```
wallrj: No need for the ```` around the type here.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+        :param String cluster_id: UUID of cluster for this
```
wallrj: String > unicode (or better still `UUID`)

flocker/node/agents/ebs.py
wallrj: So this can be deleted now.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+
+    def create_volume(self, dataset_id, size):
+        """
+        Create a volume on EBS. Store Flocker-specific
+        {metadata version, cluster id, dataset id} for the volume
+        as volume tag data.
+        Open issues: https://clusterhq.atlassian.net/browse/FLOC-1792
+
+        :param String dataset_id: Dataset_id for the volume.
```
wallrj: String > str

But actually another rule-of-thumb is that we don't bother documenting parameters and return types when they're already documented in a zope interface....which is the case here. 
So just leave the implementation specific notes.

flocker/node/agents/ebs.py
wallrj: Great. I added a note to the issue about it also being a problem with the Cinder backend.

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,213 @@
...
+
+def _wait_for_volume(expected_volume,
+                     expected_status=u'available',
+                     time_limit=60):
+    """
+    Helper function to wait for up to 60s for given volume
+    to be in 'available' state.
+    """
+    start_time = time.time()
+    expected_volume.update()
```
wallrj: So `.update()` updates the status of that instance in-place.

flocker/node/agents/ebs.py
wallrj: Ok. So if I understand @exarkun 's suggestion, we should add a new IBlockDeviceAPI test to demonstrate that `delete_volume` doesn't  delete attached volumes.

Add extra logic here to prevent any attempt to delete attached volume.

And finally add Eliot message / action here to report any other failures that come back from the EBS/ / boto `delete_volume` API. (maybe the Eliot logging isn't necessary here...exceptions raised here might be logged by the Eliot logging in the calling deployer....but if we do raise an Exception here, it ought to be a BotoException and it ought to include all the potentially useful information about the failed API request.)

flocker/node/agents/test/test_blockdevice.py
wallrj: If this remains, we'll need to make some changes to the existing loopback block device tests...I think...but let me quickly check...

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1337,6 +1350,29 @@ def test_resize_destroyed_volume(self):
...
+        Test that list_volumes() excludes volumes belonging to
+        other Flocker clusters.
+        """
+
+        flocker_volume1 = self.api.create_volume(
+            dataset_id=uuid4(),
+            size=REALISTIC_BLOCKDEVICE_SIZE,
+        )
+
+        old_cluster_id = self.api.cluster_id
```
wallrj: And the LoopbackBlockDeviceAPI doesn't have a cluster_id attribute so I think this'll break too...sorry it's all so frustrating.

flocker/node/agents/test/test_blockdevice.py
wallrj: Here's what I mean.

```
(1762)(EBS-IBlockDeviceAPI-FLOC-1762 ? =)[~/.../HybridLogic/flocker]$ sudo $(type -p trial) flocker.node.agents.test
flocker.node.agents.test.test_blockdevice
  AttachVolumeInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  AttachVolumeTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run ...                                                           [OK]
  BlockDeviceDeployerAlreadyConvergedCalculateChangesTests
    test_deleted_ignored ...                                          [SKIPPED]
    test_no_changes ...                                                    [OK]
  BlockDeviceDeployerCreationCalculateChangesTests
    test_match_configuration_to_state_of_datasets ...                      [OK]
    test_no_devices_no_local_datasets ...                                  [OK]
    test_no_devices_one_dataset ...                                        [OK]
  BlockDeviceDeployerDestructionCalculateChangesTests
    test_delete_before_resize ...                                          [OK]
    test_deleted_dataset_belongs_to_other_node ...                         [OK]
    test_deleted_dataset_volume_exists ...                                 [OK]
  BlockDeviceDeployerDiscoverStateTests
    test_attached_and_mismounted ...                                       [OK]
    test_attached_unmounted_device ...                                     [OK]
    test_no_devices ...                                                    [OK]
    test_one_device ...                                                    [OK]
    test_only_remote_device ...                                            [OK]
    test_only_unattached_devices ...                                       [OK]
    test_unrelated_mounted ...                                             [OK]
  BlockDeviceDeployerResizeCalculateChangesTests
    test_maximum_size_increased ...                                        [OK]
    test_multiple_resize ...                                               [OK]
  BlockDeviceDeployerTests
    test_calculate_necessary_state_changes ...                             [OK]
    test_discover_state_iclusterstatechange ...                            [OK]
    test_discover_state_list_result ...                                    [OK]
  BlockDeviceInterfaceTests
    test_names ...                                                         [OK]
    test_same_signatures ...                                               [OK]
  CreateBlockDeviceDatasetInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  CreateBlockDeviceDatasetTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run_create ... mke2fs 1.42.12 (29-Aug-2014)
                                                   [OK]
    test_run_mkfs_and_mount ... mke2fs 1.42.12 (29-Aug-2014)
                                           [OK]
  CreateFilesystemInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  CreateFilesystemTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
  DestroyBlockDeviceDatasetInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  DestroyBlockDeviceDatasetTests
    test_dataset_id_must_be_uuid ...                                       [OK]
    test_destroy_nonexistent ...                                           [OK]
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run ...                                                           [OK]
  DestroyVolumeInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  DestroyVolumeTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run ...                                                           [OK]
  DetachVolumeInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  DetachVolumeTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run ...                                                           [OK]
  LoopbackBlockDeviceAPIImplementationTests
    test_create_sparse ...                                                 [OK]
    test_initialise_directories ...                                        [OK]
    test_initialise_directories_attached_exists ...                        [OK]
    test_initialise_directories_unattached_exists ...                      [OK]
    test_list_attached_volumes ...                                         [OK]
    test_list_unattached_volumes ...                                       [OK]
    test_resize_data_preserved ...                                         [OK]
    test_resize_grow_sparse ...                                          [FAIL]
  LoopbackBlockDeviceAPITests
    test_attach_attached_volume ...                                     [ERROR]
    test_attach_elsewhere_attached_volume ...                           [ERROR]
    test_attach_unattached_volume ...                                   [ERROR]
    test_attached_volume_listed ...                                     [ERROR]
    test_detach_volume ...                                              [ERROR]
    test_foreign_cluster_volume ...                                     [ERROR]
    test_get_device_path_device ...                                     [ERROR]
    test_get_device_path_device_repeatable_results ...                  [ERROR]
    test_list_attached_and_unattached ...                               [ERROR]
    test_multiple_volumes_attached_to_host ...                          [ERROR]
    test_reattach_detached_volume ...                                   [ERROR]
    test_resize_destroyed_volume ...                                       [OK]
    test_resize_unknown_volume ...                                         [OK]
    test_resize_volume_listed ...                                          [OK]
  LosetupListTests
    test_parse_empty ...                                                   [OK]
    test_parse_multiple_lines ...                                          [OK]
    test_parse_one_line ...                                                [OK]
    test_remove_deleted_suffix ...                                         [OK]
    test_remove_inode ...                                                  [OK]
  MountBlockDeviceInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  MountBlockDeviceTests
    test_equality ...                                                      [OK]
    test_mountpoint_error ... mke2fs 1.42.12 (29-Aug-2014)
                                             [OK]
    test_mountpoint_exists ... mke2fs 1.42.12 (29-Aug-2014)
                                            [OK]
    test_notequality ...                                                   [OK]
    test_run ... mke2fs 1.42.12 (29-Aug-2014)
                                                          [OK]
  ResizeBlockDeviceDatasetInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  ResizeBlockDeviceDatasetTests
    test_dataset_id_must_be_uuid ...                                       [OK]
    test_dataset_id_required ...                                           [OK]
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_ordering ...                                                      [OK]
    test_run_grow ... mke2fs 1.42.12 (29-Aug-2014)
e2fsck 1.42.12 (29-Aug-2014)
resize2fs 1.42.12 (29-Aug-2014)
                                                     [OK]
    test_size_must_be_int ...                                              [OK]
    test_size_required ...                                                 [OK]
  ResizeFilesystemInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  ResizeFilesystemTests
    test_equality ...                                                      [OK]
    test_grow ... mke2fs 1.42.12 (29-Aug-2014)
e2fsck 1.42.12 (29-Aug-2014)
resize2fs 1.42.12 (29-Aug-2014)
                                                         [OK]
    test_notequality ...                                                   [OK]
  ResizeVolumeInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  ResizeVolumeTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run_grow ...                                                      [OK]
  SyncToThreadedAsyncAPIAdapterTests
  UnmountBlockDeviceInitTests
    test_init ...                                                          [OK]
    test_optional_arguments ...                                            [OK]
    test_optional_defaults ...                                             [OK]
  UnmountBlockDeviceTests
    test_equality ...                                                      [OK]
    test_notequality ...                                                   [OK]
    test_run ...                                                           [OK]

[SKIPPED]
This will pass when the deployer is smart enough to know it should not delete things that do not exist.  FLOC-1756.

flocker.node.agents.test.test_blockdevice.BlockDeviceDeployerAlreadyConvergedCalculateChangesTests.test_deleted_ignored
[FAIL]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 1572, in test_resize_grow_sparse
    (size.actual, size.reported)
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 447, in assertEqual
    % (msg, pformat(first), pformat(second)))
twisted.trial.unittest.FailTest: not equal:
a = (0, 150000000000)
b = (4096, 150000000000)


flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPIImplementationTests.test_resize_grow_sparse
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_attach_attached_volume/vbpUvA/temp/unattached/block-ae47f92a-2868-44e0-87cd-ab74dd84effb'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_attach_attached_volume
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_attach_elsewhere_attached_v/HJpGm8/temp/unattached/block-b01168fa-b328-45a7-a322-92f1d1ba7685'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_attach_elsewhere_attached_volume
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_attach_unattached_volume/9RUPyn/temp/unattached/block-bc6a050d-ee6b-420d-98f2-42926280b2f3'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_attach_unattached_volume
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_attached_volume_listed/NKMYPV/temp/unattached/block-8ef7aa87-9b1a-41ab-af90-1df5c147015a'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_attached_volume_listed
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_detach_volume/pqNMFc/temp/unattached/block-75fae810-d4b1-4e3b-92f8-947459752cf8'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_detach_volume
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 1364, in test_foreign_cluster_volume
    old_cluster_id = self.api.cluster_id
exceptions.AttributeError: 'LoopbackBlockDeviceAPI' object has no attribute 'cluster_id'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_foreign_cluster_volume
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_get_device_path_device/AszLzO/temp/unattached/block-2e4dd101-d9d8-48a4-9a57-577e754f4956'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_get_device_path_device
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_get_device_path_device_repe/nG3vlk/temp/unattached/block-d1a532d9-5029-4323-a838-2ee40bf7d6b8'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_get_device_path_device_repeatable_results
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_list_attached_and_unattache/CpXTFW/temp/unattached/block-ebcb89a0-296d-4754-af57-8f87da869f8e'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_list_attached_and_unattached
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_multiple_volumes_attached_t/x2UKQa/temp/unattached/block-74015beb-0f3a-489b-87f1-df554c57d49d'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_multiple_volumes_attached_to_host
[ERROR]
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/test/test_blockdevice.py", line 854, in tearDown
    self.api.destroy_volume(volume.blockdevice_id)
  File "/home/richard/projects/HybridLogic/flocker/flocker/node/agents/blockdevice.py", line 1011, in destroy_volume
    volume_path.remove()
  File "/home/richard/.virtualenvs/1762/lib/python2.7/site-packages/twisted/python/filepath.py", line 1245, in remove
    os.remove(self.path)
exceptions.OSError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/_trial_temp/flocker.node.agents.test.test_bl/LoopbackBlockDeviceAPITests/test_reattach_detached_volume/ZYsTyj/temp/unattached/block-52967994-a717-4e96-b19c-920fc6463583'

flocker.node.agents.test.test_blockdevice.LoopbackBlockDeviceAPITests.test_reattach_detached_volume
Ran 151 tests in 112.268s

FAILED (skips=1, failures=1, errors=11, successes=138)

```

wallrj:
Thanks @myechuri 

There are some failing tests that to be fixed before this can be merged.

I left a few other comments too.

I've got to go now, but @exarkun may be able to help advise on best way to keep the loopback tests working.

Please answer or address the points above and resubmit when the build results look reasonably green.

flocker/node/agents/ebs.py
exarkun: Not exactly, sorry.

`BlockDeviceDeployer` is the only use of `IBlockDeviceAPI` (right now).  I was trying to point out that `BlockDeviceDeployer` will never call `destroy_volume` on an attached volume.  Also, the specification for `IBlockDeviceAPI.destroy_volume` doesn't say one way or another what is required to happen if it is used on an attached volume.

I was suggesting we change the interface documentation to say it is invalid to call that method on an attached volume.  Then, if `BlockDeviceDeployer` does call it on an attached volume, whatever happens is a bug in `BlockDeviceDeployer` - not in the `IBlockDeviceAPI` implementation.

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py
myechuri: Ah, thanks for letting me know! Removed.

flocker/node/agents/ebs.py
myechuri: Correct.

flocker/node/agents/test/test_blockdevice.py
myechuri: Thanks for pointing out! Reworked this part to SkipTest if cluster_id does not exist. @exarkun , is there a plan for adding cluster_id for LoopbackBlockDeviceAPI? If yes, i can remove SkipTest for test_foreign_cluster_volume when cluster_id is available. Thanks!

myechuri:
Latest set of failures from twisted-trunk seem to be timeouts waiting for docker pull, which is unrelated to my change:
_______
flocker.node.functional.test_docker.GenericDockerClientTests.test_pull_image_if_necessary ... [ERROR]
command timed out: 1200 seconds without output, attempting to kill
process killed by signal 9
_______

myechuri:
The latest test results no longer have cluster_id issue from test_foreign_cluster_volume(). The only issue left is tearDown from IBlockDeviceAPITestsMixin, which i hope will be fixed once the relative path issue is resolved? Requesting a code review to check if there are any other open issues. Thanks!

flocker/node/agents/ebs.py
```diff
@@ -0,0 +1,229 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+An EBS implementation of the ``IBlockDeviceAPI``.
+"""
+
+from bitmath import Byte, GB
```
exarkun: An unwritten rule (:cry: sorry) is that imports are separated into stdlib, third-party, and Flocker - and should appear in that order.

flocker/node/agents/ebs.py
itamarst: Delete this  method.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -91,44 +81,7 @@ def test_foreign_volume(self):
...
-            dataset_id=uuid4(),
-            size=REALISTIC_BLOCKDEVICE_SIZE,
-        )
-
-        self.assertEqual(
-            ([flocker_volume1], [flocker_volume2]),
-            (block_device_api1.list_volumes(),
-             block_device_api2.list_volumes())
-        )
+        cinder_client.connection.delete_volume(requested_volume.id)
```
itamarst: Use `addCleanup` so leaks are less likely.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -0,0 +1,76 @@
...
+        ec2_client = ec2_client_from_environment()
+        requested_volume = ec2_client.connection.create_volume(
+            int(Byte(REALISTIC_BLOCKDEVICE_SIZE).to_GB().value),
+            ec2_client.zone)
+
+        _wait_for_volume(requested_volume)
+
+        self.assertEqual(self.api.list_volumes(), [])
+
+        ec2_client.connection.delete_volume(requested_volume.id)
```
itamarst: Use `addCleanup`

flocker/node/agents/test/test_blockdevice.py
itamarst: Seems like a bug in the loopback implementation; that's a valid public API interaction.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1447,6 +1460,35 @@ def test_resize_destroyed_volume(self):
         )
         self.assertEqual(exception.args, (volume.blockdevice_id,))
 
+    def test_foreign_cluster_volume(self):
```
itamarst: Discussed offline - should move into utility method and then add explicitly to Cinder and EBS test suites.

itamarst:
Address and merge. Thanks!

flocker/node/agents/test/test_blockdevice.py
exarkun: It's not in the implementation.  It's an interaction with the loopback setup code in the test suite.

Does the EBS cleanup have to work this way or can it work more like the loopback cleanup works?  If the former, we may need to change how loopback cleanup is done.  If the latter, I'd just as soon convert this EBS cleanup code to work how the loopback cleanup works.

flocker/node/agents/ebs.py
myechuri: Thanks for pointing out! Itamar mentioned the same as well a few days ago, i missed following convention during a followup edit. Fixed.

flocker/node/agents/functional/test_cinder.py
myechuri: Fixed.

flocker/node/agents/functional/test_ebs.py
myechuri: Thanks for the suggestion! Fixed.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -0,0 +1,100 @@
...
+        ec2_client = ec2_client_from_environment()
+        requested_volume = ec2_client.connection.create_volume(
+            int(Byte(REALISTIC_BLOCKDEVICE_SIZE).to_GB().value),
+            ec2_client.zone)
+
+        _wait_for_volume(requested_volume)
+
+        self.assertEqual(self.api.list_volumes(), [])
+
+        self.addCleanup(ec2_client.connection.delete_volume,
```
adamtheturtle: Add the cleanup as soon as possible after whatever needs to be cleaned up is changed. Otherwise the cleanup won't happen if something fails before it is added, such as the above assertion.

myechuri: Fixed. Thanks!

PR #1352, opened 2015-04-29
[FLOC-1724] Sketch of logging options.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1724


adamtheturtle:
Some build failures (e.g. `lint`) are relevant.

flocker/common/script.py
```diff
@@ -26,6 +29,23 @@
...
+    :cvar DISCARD: Discard all logs.
+    :cvar STDOUT: Log to standard out.
+    :cvar FILE: Log to a file. Defaults to a file in XDG_CACHE_HOME, but
+        overrideable on the command line.
+    """
+    # Used by legacy flocker-volume
+    DISCARD = NamedConstant()
+    # Used by daemons (flocker-control, flocker-*-agent)
+    STDOUT = NamedConstant()
+    # Used by CLI scripts (flocker, flocker-ca, flocker-deploy, flocker-provision)
```
adamtheturtle: Put "e.g." else this will soon be out of date.

flocker/common/script.py
```diff
@@ -160,7 +191,8 @@ def main(self):
         # always do this first before any side-effecty code is run:
         options = self._parse_options(self.sys_module.argv[1:])
 
-        if self.logging:
+        # TODO Factor this out to dispatch on LoggingPolicy
```
adamtheturtle: Do this, or create a follow-up.

adamtheturtle: Actually do it here, it isn't a big task.

flocker/common/script.py
```diff
@@ -26,6 +29,23 @@
 ]
 
 
+class LoggingPolicy(Names):
```
adamtheturtle: This is missing tests.

flocker/common/script.py
```diff
@@ -132,6 +159,10 @@ def __init__(self, script, options, logging=True,
             reactor = global_reactor
         self._reactor = reactor
 
+        if logging is LoggingPolicy.FILE:
+            # This adds log-dir (and verbose?) to the provided options.
```
adamtheturtle: Clarify what this does (no "?").

flocker/common/script.py
```diff
@@ -170,6 +202,12 @@ def main(self):
             # contrast needs to be shutdown because it starts a thread
             # that will keep the process from exiting.
             observer.start()
+        elif self.logging is LoggingPolicy.STDOUT:
+            log_dir = FilePath(self.options['log-dir'])
+            log_file = log_dir.child(
```
adamtheturtle: Write a comment or refactor this so that what it is doing is clear.

flocker/common/script.py
```diff
@@ -53,12 +73,19 @@ def opt_version(self):
...
+    # that is only applied to CLI scripts
     def opt_verbose(self):
         """Turn on verbose logging."""
         self['verbosity'] += 1
     cls.opt_verbose = opt_verbose
     cls.opt_v = opt_verbose
 
+    cls.optParameters = getattr(cls, 'optParameters', [])
+    cls.optParameters.append(
+        ['log-dir', None, user_log_dir("Flocker", "ClusterHQ"),
```
adamtheturtle: I'm not sure what a good default log location is so I'd like someone else to look at this.

adamtheturtle:
Ah shucks. I started doing a code review before seeing that I should be doing a design review.

adamtheturtle:
I'm going to put this back as the comments I had which were like "someone else should look at this" are the relevant ones for design review.

flocker/common/script.py
itamarst: Looks like this was supposed to be `LoggingPolicy.FILE`?

itamarst:
So if I understand this correctly:

1. Each script has a policy: no logging, stdout, or file.
2. File policy logs to a default location which can be overrriden.
3. It's not clear what verbose does.

I propose a slight tweak which I think improves usability. There are three modes:

1. Stdout. No verbose flag.
2. No logging by default. There is a verbose or enable-logging flag which enables logging to a file, and therefore also command-line option to determine where logs go.
3. Logging to file. There is a command-line option to determine where logs go. No verbose flag.

itamarst:
On further thought, since command-line tools aren't long running and won't product much logs, I suggest not bothering with second policy and just having all scripts either log stdout or to file, and drop the verbose flag.

tomprince:
The only reason for having a "no logging" option is for the `flocker-volumes` script which is called remotely by the running system. I didn't think the XDG_USER_DIR of root was a good place for the logs, so wanted to separate it out. Since it is going away, I figured I'd just drop the logs for the time being.

tomprince:
Regarding getting rid of verbose, I agree whole-heartedly.

itamarst:
Re `flocker-volume` do whatever takes the least amount of work. If it logs something to file that's fine.

PR #1355, opened 2015-04-29
More descriptive random names

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1779

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -398,7 +398,7 @@ CMD sh -c "trap \"\" 2; sleep 3"
         image = DockerImageBuilder(test=self, source_dir=docker_dir)
         image_name = image.build()
         client = self.make_client()
-        name = random_name()
+        name = random_name(self)
         self.create_container(client, name, image_name)
         self.addCleanup(client.remove, name)
         d = client.list()
@@ -688,7 +688,7 @@ CMD sh -c "trap \"\" 2; sleep 3"
         A container with custom command line is run with those arguments.
         """
         external_port = find_free_port()[1]
-        name = random_name()
+        name = random_name(self)
         d = self.start_container(
             name, image_name=u"busybox",
             command_line=[u"sh", u"-c", u"""\
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -36,7 +36,7 @@ from ..testtools import if_docker_configured, wait_for_unit_state
 
 
 def namespace_for_test(test_case):
-    return u"ns-" + random_name(test_case).replace(u".", u"-")
+    return u"ns-" + random_name(test_case)
 
 
 class IDockerClientTests(make_idockerclient_tests(
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -220,7 +220,7 @@ class GenericDockerClientTests(TestCase):
             b"FROM busybox\nCMD /bin/true\n")
         image_name = DockerImageBuilder(test=self, source_dir=path,
                                         cleanup=False).build()
-        name = random_name()
+        name = random_name(self)
         d = self.start_container(unit_name=name, image_name=image_name,
                                  expected_states=(u'inactive',))
 
@@ -831,7 +831,7 @@ class DockerClientTests(TestCase):
         :param str method_name: Method of a docker ``Client``.
         :return: ``Deferred`` firing on test success.
         """
-        name = random_name()
+        name = random_name(self)
         client = DockerClient()
         self.addCleanup(client.remove, name)
         d = client.add(name, u"busybox:latest")
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -36,14 +36,14 @@ from ..testtools import if_docker_configured, wait_for_unit_state
 
 
 def namespace_for_test(test_case):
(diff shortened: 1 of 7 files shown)


flocker/node/functional/test_docker.py
```diff
@@ -67,17 +70,13 @@ class GenericDockerClientTests(TestCase):
     Functional tests for ``DockerClient`` and other clients that talk to
     real Docker.
     """
+    clientException = APIError
```
wallrj: I was about to say something about this variable name, but I see it's always been that way.

flocker/node/functional/test_docker.py
```diff
@@ -67,17 +70,13 @@ class GenericDockerClientTests(TestCase):
...
-        self.namespacing_prefix = u"%s-%s-%s--" % (self.__class__.__name__,
-                                                   self.id(), random_name())
-        self.namespacing_prefix = self.namespacing_prefix.replace(u".", u"-")
-
-    clientException = APIError
+        self.namespacing_prefix = namespace_for_test(self)
 
     def make_client(self):
-        # Some of the tests assume container name matches unit name, so we
-        # disable namespacing for these tests.
```
wallrj: This was just an obsolete comment I suppose.

flocker/node/test/test_docker.py
```diff
@@ -172,10 +190,10 @@ def got_list(units):
 
         def test_container_name(self):
             """
-            Each unit also records the name of the container it is running in.
+            Each container also records the container name twice.
```
wallrj: Hehe. I don't think this is an improvement. 
Maybe just link to https://clusterhq.atlassian.net/browse/FLOC-819 with a note that container_name and name are always the same and this test should probably be removed.

wallrj: https://github.com/ClusterHQ/flocker/blob/master/flocker/node/_docker.py#L79-L90

flocker/testtools/__init__.py
```diff
@@ -215,12 +215,16 @@ def loop(result):
     return d
 
 
-def random_name():
-    """Return a short, random name.
+def random_name(case):
+    """
+    Return a short, random name.
```
wallrj: Not as short as it used to be...pretty long in fact.
I had a feeling that I tried to do something like this before and that there was a restriction on the length of Docker image and container names....but perhaps that was a geard restriction.
I'll quickly search.

flocker/node/functional/test_docker.py
```diff
@@ -36,14 +36,14 @@
 
 
 def namespace_for_test(test_case):
-    namespace = u"%s-%s-%s" % (
-        test_case.__class__.__name__, test_case.id(), random_name())
-    namespace = namespace.replace(u".", u"-")
-    return namespace
+    return u"ns-" + random_name(test_case).replace(u".", u"-")
```
wallrj: `random_name` already does the `.` > `-` replacement, I think.

wallrj:
Thanks @exarkun 

Looks good.

The docker container names are definitely more identifiable....if a bit long.

```
(less-random-names-FLOC-1779 ? =)[~/.../HybridLogic/flocker]$ docker ps -a
CONTAINER ID        IMAGE                                                                                                                     COMMAND                CREATED             STATUS                    PORTS               NAMES
2a79d9423c03        flockerlocaltests/flocker_node_functional_test_docker_genericdockerclienttests_test_restart_policy_always-466345:latest   "/usr/local/bin/run"   10 seconds ago      Up 9 seconds                                  ns-flocker_node_functional_test_docker_GenericDockerClientTests_test_restart_policy_always-341425flocker_node_functional_test_docker_GenericDockerClientTests_test_restart_policy_always-368125   
bb46b612a41f        busybox:buildroot-2014.02                                                                                                 "/bin/sh"              46 hours ago        Exited (0) 46 hours ago                       flocker--NamespacedDockerClientTests-flocker-node-functional-test_docker-NamespacedDockerClientTests-test_dead_is_removed-693678932024--533110032983                                              
1d5087fbf5af        busybox:buildroot-2014.02                                                                                                 "/bin/sh"              2 days ago          Exited (0) 2 days ago                         nostalgic_fermi                                                                                                                                                                                   
a7acafdc7952        9b09d93f3f64                                                                                                              "--env=bar=baz sh"     2 days ago                                                        angry_mclean                                                                                                                                                                                
```

I spotted a couple other places that use the testcase id. Maybe use the new random_name there too.

```
(less-random-names-FLOC-1779 ? =)[~/.../HybridLogic/flocker]$ ack '\.id\(\)'
flocker/common/test/test_thread.py
192:        threadpool = ThreadPool(minthreads=1, name=self.id())

flocker/route/functional/test_iptables_create.py
256:        network_namespace = b"%s.%s" % (self.id(), getpid())

flocker/testtools/__init__.py
227:    return u"{}-{}".format(case.id().replace(u".", u"_"), randrange(10 ** 6))

```


There's also another `random_name` function in the ZFS implementation code.

```
flocker/volume/filesystems/zfs.py
38:def random_name():
39:    """Return a random pool name.
43:    return os.urandom(8).encode("hex")

```

...but that doesn't seem to be used.

Anyway. Please answer or address the comments above and merge when you're happy with the build results.

flocker/testtools/__init__.py
exarkun: Docker seems pretty permissive.  I created a container named "x" * 2048 and it was okay.

PR #1357, opened 2015-04-30
[FLOC-1785] CA cluster UUID

dwgebler:
Fixes FLOC-1785

### DIFF
--- a/flocker/ca/_ca.py
+++ b/flocker/ca/_ca.py
@@ -376,7 +376,7 @@ class NodeCredential(PRecord):
         :param CertificateAuthority authority: The certificate authority with
             which this certificate will be signed.
         """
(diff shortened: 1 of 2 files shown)


itamarst:
The downside of this design is that it puts an extra burden on the administrator and whoever is implementing clients.

In order to configure a node you now need to distribute CA certificate, node certificate and node key, as well as having the UUID configured in a config file somewhere. This is because you can't just validate the control service based on being signed by CA and having `control-service` as the common name.
For usage of HTTP API, e.g. via `curl`, this is also the case: you need to distribute yet another piece of information, it's tricker to use `curl`, etc..

I would rather the cluster UUID be encoded in the CA certificate, since we need to distribute that anyway.

itamarst:
Going to redo this the other way.

PR #1361, opened 2015-05-01
[FLOC-1782] CA Script Test

dwgebler:
Fixes FLOC-1782

### DIFF
--- a/flocker/ca/test/test_script.py
+++ b/flocker/ca/test/test_script.py
@@ -38,8 +38,8 @@ class FlockerCAMainTests(TestCase):
         path.makedirs()
 
         cwd = os.getcwd()
-        os.chdir(path.path)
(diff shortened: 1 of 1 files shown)


flocker/ca/test/test_script.py
```diff
@@ -44,6 +45,8 @@ def test_deferred_result(self):
         script = CAScript()
         dummy_reactor = object()
 
+        self.addCleanup(os.chdir, cwd)
```
exarkun: You want to make really sure this cleanup always runs.  If you add it here, there are 4 lines worth of code that could fail that will leave the working directory set to `path.path` - and more than half of them are lines that invoke some of the code under test.

Put this line before line 40 to be sure the correct directory is always restored.

exarkun:
Thanks.  Merge after fixing that.

PR #1362, opened 2015-05-01
Yml startup floc 1790

adamtheturtle:
Where we stray from [the design](https://github.com/ClusterHQ/flocker/pull/1338/files):

* For now the config option is called "control-service-config" instead of "config-options". This is because the "flocker-zfs-agent" script already has a "--config" option. The option name and description matches what this is now but a follow-up branch can change that when what it does changes.

* @exarkun asked for quotes around yml strings. We have done this in the documentation example but in the task we found one way to do this with `yaml.safe_dump` and that was to use `default_style='"'` param. However this looked bad with different types.

* The example yml used to be wrapped by braces but that changed because it looked odd to us.

* We agree that this documentation should be moved but have not done it here. We have asked David to look into this and will likely move these docs before the story is submitted.

Also:

* There is duplication between the agent options but the next branch (as in the design) will reduce the two sets of options to one.

* We have kept the assertion in the `service` files from the design. We have also added a similar assertion to the Ubuntu upstart file. However we aren't sure what this adds over error handling in the node script.

* We think when we want more details, the yaml should contain a dictionary like:

```yaml
control-service: {
  "hostname": "192.168.1.1",
  "port": 4252,
  "protocol": "tcp",
}
```

but this is a direct replacement for the existing parameter and so we have not done this. We think that changes before the next release can probably continue to use version 1.

### DIFF
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -43,6 +43,10 @@ Ubuntu
 
 TODO
 
+TODO? Move most of this document somewhere?
+TODO? Make the service file assert the file exists
+TODO? Make the provision code use the config?
+
 API Details
 ===========
 
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -69,8 +69,6 @@ This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the
 .. task:: enable_flocker_agent ubuntu-14.04 ${CONTROL_NODE}
    :prompt: [root@agent-node]#
 
-TODO port option
-TODO version
 
 API Details
 ===========
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -41,10 +41,9 @@ This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the
 Ubuntu
 ------
 
-TODO
+TODO (including changing upstart file)
 
 TODO? Move most of this document somewhere?
-TODO Make the service file assert the file exists
 
 API Details
 ===========
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -44,8 +44,11 @@ Ubuntu
 TODO
 
 TODO? Move most of this document somewhere?
-TODO? Make the service file assert the file exists
-TODO? Make the provision code use the config?
+TODO? Change the name of the file?
+TODO Make the service file assert the file exists
+TODO Make the provision._install code use the config
+
+
 
 API Details
 ===========
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -67,9 +67,8 @@ This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the
 .. task:: enable_flocker_agent ubuntu-14.04 ${CONTROL_NODE}
    :prompt: [root@agent-node]#
 
-TODO Change upstart file
-
-TODO? Move most of this document somewhere?
+- Quotes around yml - looks bad with option
+- We're not moving this document but agree that it should be moved and have asked David to look into this for before the story is submitted
 
 API Details
 ===========
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -11,8 +11,8 @@ While this API is not yet directly accessible in a standard Flocker setup, the d
 Installation
 ============
 
-Fedora/CentOS
--------------
+Fedora / CentOS
+---------------
 
 To enable the Flocker control service.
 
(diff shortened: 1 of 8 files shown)


docs/advanced/api.rst
```diff
@@ -11,28 +11,65 @@ While this API is not yet directly accessible in a standard Flocker setup, the d
...
+To configure ``UFW`` to allow access to the control service REST API, and for agent connections:
+
+.. task:: open_control_firewall ubuntu-14.04
+   :prompt: [root@control-node]#
+
+For more details on configuring the firewall, see Ubuntu's `UFW documentation <https://help.ubuntu.com/community/UFW>`_.
+
+On AWS, an external firewall is used instead, which will need to be configured similarly.
+
+To start the agents on a node, a configuration file must exist on the node at ``/etc/flocker/agent.yml``.
```
itamarst: This bit is not Ubuntu or Fedora specific, so instead of duplicating I suggest having a single section about the configuration file.

flocker/node/script.py
```diff
@@ -43,17 +46,18 @@ class ZFSAgentOptions(Options):
...
 
-    synopsis = (
-        "Usage: flocker-zfs-agent [OPTIONS] <local-hostname> "
-        "<control-service-hostname>")
+    synopsis = ("Usage: flocker-zfs-agent [OPTIONS]")
 
     optParameters = [
         ["destination-port", "p", 4524,
          "The port on the control service to connect to.", int],
+        ["control-service-config", "c", "/etc/flocker/agent.yml",
```
itamarst: This is not a good option name. `agent-config` seems more future proof.

flocker/node/script.py
```diff
@@ -76,6 +80,46 @@ def _get_external_ip(host, port):
         sock.close()
 
 
+def configuration_from_options(options):
+    """
+    Extract configuration from provided options.
+
+    :param dict options: Dictionary containing command line options.
```
itamarst: :question: I suggest taking the path as a parameter instead.

flocker/node/script.py
```diff
@@ -76,6 +80,46 @@ def _get_external_ip(host, port):
...
+    except IOError:
+        raise ConfigurationError(
+            "Configuration file does not exist at '{}'.".format(
+                control_service_options_file.path))
+
+    options_yaml = yaml.safe_load(options_content)
+
+    configuration = {}
+
+    try:
```
itamarst: We're presumably going to be loading a whole bunch of options this way. So maybe a utility function for "lookup and fail with message if not found."

flocker/node/test/test_script.py
```diff
@@ -254,23 +291,135 @@ def test_custom_port(self):
...
+
+    def setUp(self):
+        self.scratch_directory = FilePath(self.mktemp())
+        self.scratch_directory.makedirs()
+        self.config = self.scratch_directory.child('config.yml')
+        self.options = {
+            'control-service-config': self.config,
+        }
+
+    def test_error_on_file_does_not_exist(self):
```
itamarst: All the error tests are very repetitive. Add a new `assertErrorForYAML` utility method, at least.

flocker/node/script.py
```diff
@@ -76,6 +80,46 @@ def _get_external_ip(host, port):
...
+def configuration_from_options(options):
+    """
+    Extract configuration from provided options.
+
+    :param dict options: Dictionary containing command line options.
+
+    :return dict: Dictionary containing the desired configuration.
+    """
+    control_service_options_file = options[u'control-service-config']
+
```
itamarst: As an alternative to manual validation, consider just writing a JSON Schema schema and validating against that.

itamarst: (Might not be workable for pluggable storage backend validation, so maybe not?)

itamarst: So yeah, never mind.

exarkun: The backend plugins can supply schema additions so I think it works.

I'd really like to see this done with jsonschema.

flocker/provision/_install.py
```diff
@@ -217,23 +218,26 @@ def task_open_control_firewall(distribution):
...
     Configure and enable the flocker agents.
 
     :param bytes control_node: The address of the control agent.
     """
+    put_config_file = put(
+        path='/etc/flocker/agent.yml',
+        content=yaml.safe_dump(
+            {
+                "version": 1,
+                "control-service-endpoint": control_node,
```
itamarst: `control-service-endpoint` should be `control-service-hostname`.

itamarst:
Looks good, address, make sure buildbot is happy, merge.

docs/advanced/api.rst
adamtheturtle: I agree but because this page is going to be removed or at least significantly restructured (perhaps into the installation page) later in this story I don't think that worrying about duplication here is important.

flocker/node/test/test_script.py
adamtheturtle: Done.

PR #1363, opened 2015-05-01
[FLOC-1786] TLS implementation on control service

dwgebler:
Fixes FLOC-1786. Putting this up for code review now for the purposes of gaining feedback on existing code changes and what else should be included in this branch vs follow-up issues. To be addressed in this branch or follow-up issues at a minimum:

- Updating user facing documentation for `flocker-deploy`, which now takes a `--certificate-path` option defaulting to the current directory.

- Updating the provisioning (inc. Vagrantfile for Vagrant acceptance tests) to generate and include the necessary certificates and keys. Acceptance tests can't pass on this branch at the moment, because the nodes have no TLS credentials.

Also there is one outstanding problem with this branch, but I don't know yet why the real HTTP API tests fail on the OS X 10 builder, they pass on everything else.

- http://build.clusterhq.com/builders/flocker-osx-10.10/builds/1871/steps/trial/logs/stdio

^ Interesting one, you'd think the tests aren't cleaning up after themselves, but they should be. I don't know OS X well enough to have an idea of what might be different about that test environment such that these tests are failing.


itamarst:
How will tutorial vagrant image and acceptance machine provisioning be changed to get certs?

dwgebler:
> How will tutorial vagrant image and acceptance machine provisioning be changed to get certs?

Acceptance machine provisioning, not sure - perhaps we have pre-generated root, control and node certificates that we copy when provisioning? Vagrant's a bit easier with a similar approach, I guess have pre-generated, known certificates and keys in PEM format dumped in to files in the Vagrant tutorial boxes via the bootstrap script in the `Vagrantfile`. Flocker will look for its certs in `/etc/flocker` by default.

flocker/control/_protocol.py
```diff
@@ -197,19 +201,39 @@ class ControlAMPService(Service):
     """
     logger = Logger()
 
-    def __init__(self, cluster_state, configuration_service, endpoint):
+    def __init__(self, cluster_state, configuration_service,
+                 endpoint, certificate_path=b"/etc/flocker"):
```
itamarst: Pass in a `FilePath`, not bytes.

flocker/control/httpapi.py
```diff
@@ -1019,7 +1026,8 @@ def api_dataset_from_dataset_and_node(dataset, node_hostname):
     return result
 
 
-def create_api_service(persistence_service, cluster_state_service, endpoint):
+def create_api_service(persistence_service, cluster_state_service, endpoint,
+                       certificate_path=b"/etc/flocker"):
```
itamarst: ``FilePath``.

flocker/control/httpapi.py
```diff
@@ -1031,10 +1039,31 @@ def create_api_service(persistence_service, cluster_state_service, endpoint):
 
     :param endpoint: Twisted endpoint to listen on.
 
+    :param bytes certificate_path: Absolute path to directory containing
+        the cluster root certificate and the control service's certificate
+        and private key.
+
     :return: Service that will listen on the endpoint using HTTP API server.
     """
+    certificate_path = FilePath(certificate_path)
```
itamarst: Duplicate code, so factor this out into utility function.

itamarst:
Besides what's in the sketch, you will also need to update the acceptance tests client, the `test_httpapi` real client, and as mentioned the provisioning code. Looks good, please proceed.

itamarst:
`flocker-deploy` documentation is way out of date anyway (it doesn't talk about control service hostname argument I believe) so for now I suggest not bothering with that. Passing acceptance tests needs to happen, though.

flocker/cli/script.py
```diff
@@ -60,8 +63,12 @@ class DeployOptions(Options):
                 "<control-host> <deployment.yml-path> <application.yml-path>"
                 "{feedback}").format(feedback=FEEDBACK_CLI_TEXT)
 
-    optParameters = [["port", "p", REST_API_PORT,
-                      "The REST API port on the server.", int]]
+    optParameters = [
+        ["port", "p", REST_API_PORT,
+         "The REST API port on the server.", int],
+        ["certificate-path", "c",
+         None, "Certificate path, defaults to current directory."],
```
itamarst: "Certificate path" isn't very meaningful. "Path to directory where TLS certificate and keys can be found. Defaults to current directory." or something?

flocker/control/_protocol.py
```diff
@@ -37,12 +37,37 @@
...
+    TLS connection using the control service certificate found in the
+    supplied path.
+
+    :param FilePath path: Absolute path to directory containing
+        control service certificate and private key files.
+
+    :return CertificateOptions: A context factory pre-loaded with the
+        options of the control service certificate.
+    """
+    if path is None:
```
itamarst: This (i.e. default certificate directory) is library code. Configuration, especially default configuration, should only be in application code, a thin layer on top of the library code. In this case that would probably be `flocker/control/script.py`.

flocker/ca/__init__.py
```diff
@@ -9,7 +9,8 @@
     "ComparableKeyPair", "PathError", "CertificateAlreadyExistsError",
     "KeyAlreadyExistsError", "EXPIRY_20_YEARS",
     "AUTHORITY_CERTIFICATE_FILENAME", "AUTHORITY_KEY_FILENAME",
-    "CONTROL_CERTIFICATE_FILENAME", "CONTROL_KEY_FILENAME"
+    "CONTROL_CERTIFICATE_FILENAME", "CONTROL_KEY_FILENAME",
+    "DEFAULT_CERTIFICATE_PATH"
```
itamarst: This should be in `flocker/control/script.py` - see discussion in another comment.

flocker/control/script.py
```diff
@@ -30,6 +30,10 @@ class ControlOptions(Options):
          "The external API port to listen on."],
         ["agent-port", "a", 'tcp:4524',
          "The port convergence agents will connect to."],
+        ["certificate-path", "c", None,
+         ("Absolute path to directory containing the cluster "
+          "root certificate and control service certificate "
+          "and private key. Defaults to /etc/flocker")],
```
itamarst: `twisted.python.usage` supports defaults, which has benefit of simplifying code slightly but also means you don't need to duplicate "/etc/flocker" in the help string.

flocker/control/httpapi.py
```diff
@@ -1025,7 +1046,38 @@ def api_dataset_from_dataset_and_node(dataset, node_uuid):
     return result
 
 
-def create_api_service(persistence_service, cluster_state_service, endpoint):
+def tls_user_context_factory(path, username=u"client"):
```
itamarst: This is unnecessary; you can just use the MITM-vulnerable `twisted.internet.ssl.ClientContextFactory` for now, or if #1368 gets merged first then use the real validation code.

flocker/cli/script.py
itamarst: And maybe give the parameter a name that suggests it's a directory; `certificates-directory` or something.

flocker/control/test/test_httpapi.py
```diff
@@ -2745,24 +2746,33 @@ def test_returns_service(self):
...
         reactor = MemoryReactor()
         endpoint = TCP4ServerEndpoint(reactor, 6789)
-        verifyObject(IService, create_api_service(None, None, endpoint))
+        verifyObject(IService, create_api_service(
+            None, None, endpoint, path))
 
     def test_listens_endpoint(self):
         """
         ``create_api_service`` returns a service that listens using the given
         endpoint with a HTTP server.
```
itamarst: This isn't tested anymore. On the other hand it's a bit of a silly test... Just fix the docstring i guess.

itamarst:
Scope of branch seems fine, perhaps a bit too big (see comment 3 below). No idea what's up with OS X, sorry... Maybe @exarkun has some ideas.

1. Address comments above.
2. Acceptance tests need to pass. We can do the minimal necessary work at this point and extend to other things later, i.e. make sure the control service has the keys it needs. Or if #1368 gets merged first you can do the whole thing in this branch.
3. You can drop authentication/authorization of any sort if necessary to get tests to pass, since we'll be replacing it with #1368 anyway.

exarkun:
Needs master merged and conflicts resolved, looks like.

Also, what's going on on buildbot?

admin/acceptance.py
```diff
@@ -404,14 +409,22 @@ def main(reactor, args, base_path, top_level):
 
     try:
         nodes = yield runner.start_nodes(reactor)
+
+        ca_directory = FilePath(mkdtemp())
+        print("Generating certificates in: {}".format(ca_directory.path))
+        certificates = Certificates.generate(ca_directory, nodes[0].address,
+                                             len(nodes))
+
```
exarkun: The acceptance tests don't need a new certificate every time, I bet.  It would be nice to re-use existing certificates if they exist since key generation is time consuming.

flocker/acceptance/testtools.py
```diff
@@ -476,12 +479,14 @@ class Cluster(PRecord):
...
-    :param Node control_node: The node running the ``flocker-control``
+    :ivar Node control_node: The node running the ``flocker-control``
         service.
-    :param list nodes: The ``Node`` s in this cluster.
+    :ivar list nodes: The ``Node`` s in this cluster.
+    :ivar treq: A ``treq`` client.
     """
     control_node = field(type=ControlService)
     nodes = field(type=_NodeList)
+    treq = field()
```
exarkun: This looks mandatory

flocker/acceptance/testtools.py
```diff
@@ -771,9 +777,12 @@ def get_test_cluster(node_count=0):
                        "{existing} node(s) are set.".format(
                            necessary=node_count, existing=len(agent_nodes)))
 
+    certificates_path = FilePath(
+        environ["FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH"])
     cluster = Cluster(
         control_node=ControlService(address=control_node),
-        nodes=[]
+        nodes=[],
+        treq=treq_with_authentication(reactor, certificates_path)
```
exarkun: `reactor` should be an argument to `get_test_cluster` instead of a global here.

flocker/ca/functional/test_validation.py
```diff
@@ -285,10 +232,8 @@ class ControlServicePolicyValidationTests(make_validation_tests(
         lambda port, good_ca: ControlServicePolicy(
             ca_certificate=good_ca.root.credential.certificate,
             # This value is irrelevant to the test, but required:
```
exarkun: Which value is irrelevant?

flocker/cli/script.py
```diff
@@ -38,6 +43,29 @@
...
+def treq_with_authentication(reactor, path):
+    """
+    Create a ``treq``-API object that implements the REST API TLS
+    authentication.
+
+    That is, validating the control service as well as presenting a
+    certificate to the control service for authentication.
+
+    :param reactor: The reactor to use.
+    :param FilePath path: Directory where certificates and private key can
```
exarkun: `certificate_path`?   `certificates`?  `path` is pretty generic, sort of like `number`.

itamarst:
Buildbot is, I suspect, having issues due to `/etc/flocker` not existing on non-vagrant images (notice that vagrant acceptance tests passed). I am investigating.

admin/acceptance.py
itamarst: I'd care more if we didn't have those 10 minutes of waiting for docker pull...

itamarst:
omg green.

admin/acceptance.py
```diff
@@ -70,11 +72,13 @@ def run_tests(reactor, nodes, control_node, agent_nodes, trial_args):
         agent, to run API acceptance tests against.
     :param list trial_args: Arguments to pass to trial. If not
         provided, defaults to ``['flocker.acceptance']``.
+    :param FilePath certificates_path: Directory where certificates can be
+        found.
```
exarkun: I think this will be a directory where `flocker-ca --create-control-certificate` has been run?  Maybe mention that here.

admin/acceptance.py
exarkun: Eh, you should care.  "X is slow by Y is slower" is how you get to where we are today.

flocker/acceptance/testtools.py
```diff
@@ -265,15 +267,19 @@ def get_nodes(test_case, num_nodes):
...
 
-    # Remove all existing containers; we make sure to pass in node
-    # hostnames since we still rely on flocker-deploy to distribute SSH
-    # keys for now.
-    clean_deploy = {u"version": 1,
-                    u"nodes": {node: [] for node in reachable_nodes}}
     clean_applications = {u"version": 1,
                           u"applications": {}}
-    flocker_deploy(test_case, clean_deploy, clean_applications)
-    getting = get_test_cluster()
```
exarkun: Huh.  So ... before ... it cleaned up the cluster before it asked for it.  Yea I see how that might be a problem.

I wonder if `get_test_cluster` would be more useful if it just guaranteed it was handing you a pristine cluster.  I assume that what it does now is hand you a cluster full of whatever garbage the previous test left in it.

Worth a follow-up issue, probably.

exarkun: Oh.  Sorry, ignore the second part of that.  This is another helper, not a test method.  I misread.

flocker/acceptance/testtools.py
```diff
@@ -282,10 +288,10 @@ def no_containers(cluster):
     getting.addCallback(lambda cluster:
                         loop_until(lambda: no_containers(cluster)))
 
-    def clean_zfs(_):
+    def clean(_):
         for node in reachable_nodes:
             _clean_node(test_case, node)
-    getting.addCallback(clean_zfs)
+    getting.addCallback(clean)
```
exarkun: This really does only clean up ZFS stuff, though.  We'll need an `IBlockDeviceAPI` cleanup step here eventually.  File an issue?

flocker/acceptance/testtools.py
```diff
@@ -311,9 +318,9 @@ def flocker_deploy(test_case, deployment_config, application_config):
 
     application = temp.child(b"application.yml")
     application.setContent(safe_dump(application_config))
-
-    check_call([b"flocker-deploy", control_node, deployment.path,
-                application.path])
+    check_call([b"flocker-deploy", b"--certificates-directory",
+               certificate_path, control_node, deployment.path,
```
exarkun: This will fail with an obscure error about `None` not being allowed in `check_call` argument list if `FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH` wasn't set in the environment.

flocker/acceptance/testtools.py
```diff
@@ -771,15 +781,20 @@ def get_test_cluster(node_count=0):
                        "{existing} node(s) are set.".format(
                            necessary=node_count, existing=len(agent_nodes)))
 
+    certificates_path = FilePath(
+        environ["FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH"])
```
exarkun: It would be nice to have this kind of environment lookup code in just one place (same applies to the existing code here, eg the code looking up `FLOCKER_ACCEPTANCE_CONTROL_NODE` and duplicating the missing-case handling that's also present in `get_nodes`).

Refactor or file an issue and add a link.

flocker/acceptance/testtools.py
```diff
@@ -771,15 +781,20 @@ def get_test_cluster(node_count=0):
...
+        treq=treq_with_authentication(reactor, certificates_path)
     )
 
     # Wait until nodes are up and running:
     def nodes_available():
         d = cluster.current_nodes()
-        d.addCallback(lambda (cluster, nodes): len(nodes) >= node_count)
+        d.addCallbacks(lambda (cluster, nodes): len(nodes) >= node_count,
+                       # Control service may not be up yet, keep trying:
+                       lambda failure: False)
```
exarkun: Worth logging this failure, probably, particularly since `loop_until` has no timeout.

flocker/ca/_ca.py
```diff
@@ -462,9 +477,10 @@ class ControlCredential(PRecord):
     @classmethod
     def from_path(cls, path, hostname):
         """
-        Load a ``ControlCredential`` from disk.
+        Load a control service certificate and key from the supplied path.
 
-        :param FilePath path: Directory where credentials are stored.
+        :param FilePath path: Directory where user certificate and key
```
exarkun: s/user/control service/

flocker/ca/_validation.py
```diff
@@ -27,16 +27,17 @@ class ControlServicePolicy(PRecord):
...
+    client_credential = field(mandatory=True)
 
     def creatorForNetloc(self, hostname, port):
-        return optionsForClientTLS(u"control-service",
-                                   trustRoot=self.ca_certificate,
-                                   clientCertificate=self.client_certificate)
+        return optionsForClientTLS(
+            u"control-service",
+            trustRoot=self.ca_certificate,
+            clientCertificate=self.client_credential.private_certificate())
```
exarkun: I think that this says "Give me a ... (thing that's isomorphic to a TLS context) ... which requires the peer certificate to have a hostname that validates for `control-service` and has a signature chain from `self.ca_certificate` and present `self.client_credential.private_certificate()` as my key/certificate" - right?

flocker/ca/testtools.py
```diff
@@ -29,3 +41,85 @@ def assert_has_extension(test, credential, name, value):
...
+            self.path.child(b"{}.key".format(uuid)).copyTo(
+                directory.child(b"node.key"))
+
+
+# Cached CredentialSets generated by get_sets():
+_credential_set1 = None
+_credential_set2 = None
+
+
+def get_credential_sets():
```
exarkun: I imagine `get_credential_sets(how_many)` would make some uses of this API nicer.  Probably not a very important thing, though ...

flocker/cli/functional/test_deploy_script.py
```diff
@@ -85,6 +99,7 @@ def _send_configuration(self,
         deployment_config.setContent(safe_dump(deployment_config_yaml))
         return getProcessOutputAndValue(
             b"flocker-deploy", [
+                b"--certificates-directory", self.certificate_path.path,
                 b"--port", unicode(self.port_number).encode("ascii"),
                 b"localhost", deployment_config.path, app_config.path],
             env=environ)
```
exarkun: This method looks a lot like the acceptance test helper `flocker_deploy`.  File an issue for refactoring.

flocker/acceptance/testtools.py
```diff
@@ -28,8 +30,8 @@
 )
 
 from ..control.httpapi import container_configuration_response, REST_API_PORT
-
-from flocker.testtools import loop_until
+from ..cli.script import treq_with_authentication
```
exarkun: I think this import means `treq_with_authentication` needs to go in `flocker.common`.

flocker/control/httpapi.py
```diff
@@ -82,6 +84,20 @@
 _UNDEFINED_MAXIMUM_SIZE = object()
 
 
+class WebClientContextFactory(object):
```
exarkun: This class looks like dead code (and luckily so).

flocker/control/script.py
```diff
@@ -39,17 +49,29 @@ class ControlScript(object):
...
         persistence.setServiceParent(top_service)
         cluster_state = ClusterStateService()
         cluster_state.setServiceParent(top_service)
-        create_api_service(persistence, cluster_state, serverFromString(
-            reactor, options["port"])).setServiceParent(top_service)
+        create_api_service(
+            persistence, cluster_state, serverFromString(
+                reactor, options["port"]),
+            rest_api_context_factory(ca, control_credential)).setServiceParent(
+                top_service)
```
exarkun: Indentation makes it kind of hard to see the `setServiceParent` here

flocker/control/test/test_httpapi.py
```diff
@@ -2628,22 +2629,24 @@ def test_returns_service(self):
         """
         reactor = MemoryReactor()
         endpoint = TCP4ServerEndpoint(reactor, 6789)
-        verifyObject(IService, create_api_service(None, None, endpoint))
+        verifyObject(IService, create_api_service(
+            None, None, endpoint, ClientContextFactory()))
```
exarkun: This call violates the documented interface (`None` is not allowed for those parameters).

flocker/control/test/test_httpapi.py
```diff
@@ -2628,22 +2629,24 @@ def test_returns_service(self):
...
-        service = create_api_service(None, None, endpoint)
+        service = create_api_service(None, None, endpoint,
+                                     ClientContextFactory())
         self.addCleanup(service.stopService)
         service.startService()
         server = reactor.tcpServers[0]
         port = server[0]
         factory = server[1].__class__
-        self.assertEqual((port, factory), (6789, Site))
+        self.assertEqual((port, factory), (6789, TLSMemoryBIOFactory))
```
exarkun: Not that this was a great test before ... but at least it demonstrated HTTP-ness.  It no longer does, though.

flocker/control/test/test_script.py
```diff
@@ -75,19 +77,30 @@ class ControlScriptEffectsTests(SynchronousTestCase):
...
+            b"--port", b"tcp:8001", b"--data-path", self.mktemp(),
+            b"--certificates-directory", self.certificate_path.path
+        ])
         reactor = MemoryCoreReactor()
         ControlScript().main(reactor, options)
         server = reactor.tcpServers[0]
         port = server[0]
         factory = server[1].__class__
-        self.assertEqual((port, factory), (8001, Site))
+        self.assertEqual((port, factory), (8001, TLSMemoryBIOFactory))
```
exarkun: Also no longer really demonstrates that we have an HTTP server that serves up our API (though to be fair it didn't demonstrate anything about our API server before this change, only that it was HTTP-ish).

flocker/control/test/test_script.py
```diff
@@ -95,7 +108,10 @@ def test_no_immediate_stop(self):
         """
         script = ControlScript()
         options = ControlOptions()
-        options.parseOptions([b"--data-path", self.mktemp()])
+        options.parseOptions([
+            b"--data-path", self.mktemp(),
+            b"--certificates-directory", self.certificate_path.path
+        ])
```
exarkun: These various `ControlOptions.parseOptions` calls with the same `b"--certificates-directory", self.certificate_path.path` elements in every call would be a good thing to factor out. to reduce the repetition.

flocker/control/test/test_script.py
```diff
@@ -114,12 +133,15 @@ def test_starts_cluster_state_service(self):
...
+        options.parseOptions([
+            b"--port", b"tcp:8001", b"--data-path", self.mktemp(),
+            b"--certificates-directory", self.certificate_path.path
+        ])
         reactor = MemoryCoreReactor()
         ControlScript().main(reactor, options)
         server = reactor.tcpServers[0]
-        service = server[1].resource._v1_user.cluster_state_service
+        control_resource = server[1].wrappedFactory.resource
+        service = control_resource._v1_user.cluster_state_service
```
exarkun: I think Demeter gets angry even if you spread your `.`s across multiple statements.

Going through `wrappedFactory` seems particularly egregious and may be what tips this over the edge from bad to worth-fixing (not that `resource` is a lot better).  The `_v1_user` trick seems like the lesser evil here.  How about an attribute referring to `cluster_state_service` (since that's what we actually want to get) instead?  Then this is **only** `server[1].resource._cluster_state_service`.

flocker/node/_loop.py
```diff
@@ -404,15 +405,19 @@ class AgentLoopService(object, MultiService):
...
+        :param context_factory: TLS context factory for the AMP client.
+        """
         MultiService.__init__(self)
         convergence_loop = build_convergence_loop_fsm(
             self.reactor, self.deployer
         )
         self.logger = convergence_loop.logger
         self.cluster_status = build_cluster_status_fsm(convergence_loop)
-        self.factory = ReconnectingClientFactory.forProtocol(
-            lambda: AgentAMP(self))
```
exarkun: You could keep a direct reference to this factory around and avoid having to insert `wrappedFactory` in a ton of places.

flocker/node/script.py
```diff
@@ -173,7 +199,10 @@ def main(self, reactor, options, volume_service):
         deployer = P2PManifestationDeployer(ip, volume_service,
                                             node_uuid=node_uuid)
         loop = AgentLoopService(reactor=reactor, deployer=deployer,
-                                host=host, port=port)
+                                host=host, port=port,
+                                context_factory=_context_factory(
+                                    options["agent-config"].parent(),
+                                    host, port))
```
exarkun: I guess `ZFSAgentScript` is going to be removed soon so this duplication isn't a big deal.

flocker/node/test/test_loop.py
```diff
@@ -692,7 +700,8 @@ def test_connected(self):
         is passed to the cluster status FSM.
         """
         service = AgentLoopService(
-            reactor=None, deployer=object(), host=u"example.com", port=1234)
+            reactor=None, deployer=object(), host=u"example.com", port=1234,
+            context_factory=ClientContextFactory())
```
exarkun: Maybe these tests should use the existing `_build_service` helper so this fix doesn't have to be repeated so many times.

flocker/node/test/test_script.py
```diff
@@ -22,12 +22,43 @@
...
+                u"hostname": u"10.0.0.1",
+                u"port": 1234,
+            },
+            u"dataset": {
+                u"backend": u"zfs",
+            },
+            u"version": 1,
+        }))
+    ca_set.copy_to(scratch_directory, node=True)
+    test.non_existent_file = scratch_directory.child('missing-config.yml')
```
exarkun: Yay.  Thanks for this refactoring.

flocker/node/test/test_loop.py
```diff
@@ -643,7 +645,8 @@ def test_initialization(self):
         """
         deployer = object()
         service = AgentLoopService(
-            reactor=None, deployer=deployer, host=u"example.com", port=1234)
+            reactor=None, deployer=deployer, host=u"example.com", port=1234,
+            context_factory=ClientContextFactory())
         cluster_status_fsm_world = service.cluster_status._fsm._world.original
         convergence_loop_fsm_world = (
             cluster_status_fsm_world.convergence_loop_fsm._fsm._world.original)
```
exarkun: is this actually a useful test?

flocker/restapi/testtools.py
```diff
@@ -179,13 +180,21 @@ class RealTests(mixinClass, TestCase):
...
             self.app = fixture(self)
-            self.port = reactor.listenTCP(0, Site(self.app.resource()),
-                                          interface="127.0.0.1")
+            path = FilePath(self.mktemp())
+            path.makedirs()
+            self.port = reactor.listenTCP(
+                0, Site(self.app.resource()),
+                interface=b"127.0.0.1",
+            )
+            self.addCleanup(self.port.factory.stopFactory)
```
exarkun: Ehm.  `stopFactory` is called by the reactor (well, via `doStop`) when the factory is no longer listening on a port.  This shouldn't be necessary.

flocker/provision/_ssh/_conch.py
```diff
@@ -101,7 +101,7 @@ def perform_put(dispatcher, intent):
     """
     See :py:class:`Put`.
     """
-    return Effect(Run(command='printf %s > %s'
+    return Effect(Run(command='printf -- %s > %s'
                               % (shell_quote(intent.content),
                                  shell_quote(intent.path))))
```
exarkun: Does this have a test?

exarkun:
Thanks.  Please address, merge master (again :/), and re-submit.

admin/acceptance.py
itamarst: Also note that in Buildbot I assume we just throw away previous runs, so this doesn't help us there.

flocker/acceptance/testtools.py
itamarst: Merging forward there's an if statement now, so I assume this will be dealt with by the configurable backend branch.

flocker/ca/_validation.py
itamarst: Exactly.

flocker/acceptance/testtools.py
itamarst: I think `flocker.ca` is probably better, but yeah.

flocker/node/test/test_loop.py
itamarst: Probably not.

flocker/provision/_ssh/_conch.py
itamarst: Provisioning fails if you don't make that change, since certificates start with `--`.

itamarst:
Guh, acceptance tests.

admin/acceptance.py
exarkun: True :cry:

flocker/provision/_ca.py
```diff
@@ -0,0 +1,77 @@
...
+        self.cluster = CertAndKey(directory.child(b"cluster.crt"),
+                                  directory.child(b"cluster.key"))
+        # Assume only one control service:
+        self.control = CertAndKey(
+            directory.globChildren(b"control-*.crt")[0],
+            directory.globChildren(b"control-*.key")[0])
+        self.user = CertAndKey(directory.child(b"user.crt"),
+                               directory.child(b"user.key"))
+        nodes = []
+        for child in directory.globChildren(b"????????-????-*.crt"):
```
exarkun: Yuuuuummmmmmmmmmmm

flocker/provision/_install.py
```diff
@@ -149,6 +149,46 @@ def task_disable_selinux(distribution):
...
+
+    :param FilePath ca_cert: Path to CA certificate on local machine.
+    :param FilePath control_cert: Path to control service certificate on
+        local machine.
+    :param FilePath control_key: Path to control service private key
+        local machine.
+    """
+    return sequence([
+        run('mkdir -p /etc/flocker'),
+        run('chmod 0700 /etc/flocker'),
```
exarkun: mnemonics are more typing but easier reading, `chmod u=rwX,g=,o= foo`, probably generally good to prefer that form

:question:

exarkun: I wonder if there's a vulnerability to an attacker to manages to open the directory before it is chmod'd and then uses `openat` to get the private key.

Possibly not an attack vector we're concerned with.  Maybe worth a ticket nevertheless?  Dunno...  Also I should probably be making this a private comment. :cry:

flocker/restapi/testtools.py
```diff
@@ -179,13 +180,20 @@ class RealTests(mixinClass, TestCase):
         """
         def setUp(self):
             self.app = fixture(self)
-            self.port = reactor.listenTCP(0, Site(self.app.resource()),
-                                          interface="127.0.0.1")
+            path = FilePath(self.mktemp())
+            path.makedirs()
```
exarkun: I can't quite see how this is used.

exarkun:
Thanks.  This looks good to me.  Please address inline comments and merge when buildbot is green.

flocker/restapi/testtools.py
itamarst: Hm, left over from some no longer present code.

PR #1364, opened 2015-05-01
Shrink block device volumes

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1503

Initially I thought this would involve improving state discovery to include the size of non-manifest datasets.  I changed my mind, though.  Resizes are only attempted for datasets that have a manifestation and those datasets already have their size discovered.  I left the design notes for non-manifest dataset size discovered in, though, marked `FLOC-1503-x`.  I may go ahead and implement that stuff here anyway, it's barely an extra code and improves the system.  If the reviewer would prefer me not to, though, I'll split those things off into a separate issue.

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -218,8 +218,9 @@ definitions:
     # considered more carefully.
     minimum: 67108864
     # Filesystem initialization in the block device backend requires sizes to
-    # be a multiple of 512.
-    multipleOf: 512
+    # be a multiple of 512 and for ease of implementation we'll tighten that
+    # constraint so it's multiples of 1024 (1 KiB).
+    multipleOf: 1024
 
 
   running:
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -217,8 +217,10 @@ definitions:
     # on what this can be set to for the ZFS backend.  Could probably be
     # considered more carefully.
     minimum: 67108864
-    # This is how you require integers, of course.
-    divisibleBy: 1
+    # Filesystem initialization in the block device backend requires sizes to
+    # be a multiple of 512.
+    multipleOf: 512
+
 
   running:
     title: "Running"
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -413,12 +413,14 @@ class CreateFilesystem(PRecord):
 
 def _valid_size(size):
     """
-    Pyrsistent invariant for filesystem size, which must be a multiple of 512
+    Pyrsistent invariant for filesystem size, which must be a multiple of 1024
     bytes.
     """
-    if size % 512 == 0:
+    if size % 1024 == 0:
         return (True, "")
-    return (False, "Filesystem size must be multiple of 512, not %d" % (size,))
+    return (
+        False, "Filesystem size must be multiple of 1024, not %d" % (size,)
+    )
 
 
 @implementer(IStateChange)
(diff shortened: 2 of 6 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -487,6 +487,17 @@ def run(self, deployer):
         )
         unmount = UnmountBlockDevice(dataset_id=self.dataset_id)
         detach = DetachVolume(dataset_id=self.dataset_id)
+
+        # FLOC-1503
+        #
+        # If the volume is shrinking, arrange the changes like
+        #
+        # unmount, detach, resize volume, attach, resize filesystem, mount
```
itamarst: That seems wrong, don't you need to resize the filesystem first?

itamarst:
Rejecting since the problem I see is either a complete misunderstanding on my part or a fundamental problem.

flocker/node/agents/blockdevice.py
exarkun: Oops.  I reversed the cases.

wallrj:
Thanks @exarkun 

You've just talked me through the proposed changes here and they make sense.

Please proceed and file follow up issues and add links for 
 * Resizing non-manifest datasets
 * Having calculate_changes emit the constituent resize state changes rather than the current compound state change.

flocker/control/schema/types.yml
```diff
@@ -217,8 +217,10 @@ definitions:
     # on what this can be set to for the ZFS backend.  Could probably be
     # considered more carefully.
     minimum: 67108864
-    # This is how you require integers, of course.
-    divisibleBy: 1
+    # Filesystem initialization in the block device backend requires sizes to
+    # be a multiple of 512.
+    multipleOf: 512
```
itamarst: I suggest making this 1024, insofar as you've just added code that presumes KiB.

flocker/node/agents/blockdevice.py
```diff
@@ -487,19 +513,29 @@ def run(self, deployer):
...
-                ]
-            ),
-            deployer,
-        )
+
+        resize_filesystem = ResizeFilesystem(volume=volume, size=self.size)
+        resize_volume = ResizeVolume(volume=volume, size=self.size)
+        if self.size < volume.size:
+            changes = [
+                unmount,
```
itamarst: Is there a follow-up issue for what happens if this fails half-way through?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2787,65 +2846,148 @@ class ResizeFilesystemInitTests(
...
+                volume=volume, filesystem=filesystem,
+            ),
+            deployer
+        ))
+
+        if new_size > original_size:
+            api.detach_volume(volume.blockdevice_id)
+            api.resize_volume(volume.blockdevice_id, new_size)
+            api.attach_volume(volume.blockdevice_id, host)
+
```
itamarst: `ResizeBlockDeviceDataset` detaches before resizing filesystem when shrinking, so this doesn't quite match actual interaction pattern.

itamarst:
Looks good: address, merge forward, merge if buildbot is green.

flocker/node/agents/test/test_blockdevice.py
exarkun: You can only resize a filesystem when the volume is attached.  In the shrink case, `ResizeBlockDeviceDataset` does the filesystem resize and **then** detaches the volume for the volume resize part of the operation.

PR #1367, opened 2015-05-04
[FLOC-1785] Store cluster UUID in the certificate authority's organizational unit

itamarst:


flocker/ca/test/test_ca.py
```diff
@@ -216,13 +216,14 @@ def test_load_error_on_unreadable_key_file(self):
         def test_certificate_ou_matches_ca(self):
             """
             A certificate written by ``UserCredential.initialize`` has the
```
tomprince: It appears this test is for both `UserCredential` and `NodeCredential` (as are the rest of the tests here. This should be fixed either here, or in a followup issue.

PR #1368, opened 2015-05-04
[FLOC-1805] Validating TLS context factories for control service, AMP clients and REST API clients

itamarst:
These will be hooked up in follow-up issue.


flocker/ca/_ca.py
```diff
@@ -4,6 +4,25 @@
 Low-level logic for a certificate authority.
 
 Uses RSA 4096-bit + SHA 256.
+
+We have three categories of certificates:
+
+1. Control service. Needs to be validated over HTTP by 3rd party clients.
+2. Node agents. Needs to encode a node UUID.
+3. API clients. Used by HTTP API clients to authenticate.
```
exarkun: Each item in this list comes with an explanation of a different area.  It would be nice to see information for all three areas for all three items.

flocker/ca/_ca.py
```diff
@@ -4,6 +4,25 @@
 Low-level logic for a certificate authority.
 
 Uses RSA 4096-bit + SHA 256.
```
exarkun: Not related to this branch, but this seems like overly specific information to offer here.

flocker/ca/_script.py
```diff
@@ -226,9 +226,11 @@ class ControlCertificateOptions(PrettyOptions):
     (defaults to current working directory).
     """
 
-    synopsis = "[options]"
+    synopsis = "[options] --hostname <hostname>"
```
exarkun: `[options]` comes before `--hostname <hostname>`.  Also this tool takes other `--options` that aren't mentioned here...  And `--hostname` is optional?  Or mandatory?

flocker/ca/_validation.py
```diff
@@ -0,0 +1,53 @@
+"""
+TLS context factories used to validate the various kinds of
+certificates generated by the CA.
+"""
+
+@implementer(IPolicyForHTTPS)
```
exarkun: Oh shit.

exarkun:
Thanks?  The design doesn't cover testing - I wonder what the strategy for that will be?

itamarst:
Try connecting with all the things, make sure they reject stuff from wrong CA, reject correct CA but wrong type, accept correct CA with correct type.

exarkun:
Okay, looks good, please proceed.  Thanks.

flocker/ca/_ca.py
```diff
@@ -453,7 +477,11 @@ def initialize(cls, path, authority, begin=None):
             which this certificate will be signed.
         :param datetime begin: The datetime from which the generated
             certificate should be valid.
+        :param bytes hostname: The hostname of the node where the control
+            service will be running.
         """
+        # XXX FLOC-1805 also take hostname.
```
exarkun: This is FLOC-1805

flocker/ca/_script.py
```diff
@@ -246,6 +254,7 @@ def run(self):
         errors, create a new control service certificate signed by the root
         and write it out to the current directory.
         """
+        # XXX FLOC-1805 pass through the hostname
```
exarkun: Ditto

flocker/ca/_validation.py
```diff
@@ -0,0 +1,108 @@
...
+"""
+
+from OpenSSL.SSL import VERIFY_PEER, VERIFY_FAIL_IF_NO_PEER_CERT
+
+from zope.interface import implementer
+
+from twisted.web.iweb import IPolicyForHTTPS
+from twisted.internet.ssl import (
+    optionsForClientTLS, CertificateOptions,
+    )
```
exarkun: Overindented `)` compared to the rest of our code

flocker/ca/functional/test_ca.py
```diff
@@ -82,13 +81,6 @@ def setUp(self):
         """
         flocker_ca(b"initialize", b"mycluster")
 
-    def tearDown(self):
-        """
-        Delete the previously created root certificate.
-        """
-        os.remove(b"cluster.crt")
-        os.remove(b"cluster.key")
-
```
exarkun: Thanks

flocker/ca/test/test_ca.py
```diff
@@ -326,9 +333,42 @@ def test_certificate_subject_node_uuid(self):
...
+        value.
+    """
+    expected = crypto.X509Extension(name, False, value)
+    x509 = credential.certificate.original
+    for i in range(x509.get_extension_count()):
+        extension = x509.get_extension(i)
+        if extension.get_short_name() == name:
+            test.assertEqual(extension.get_data(), expected.get_data())
+            return
+    test.fail("Couldn't find extension {}.".format(name))
```
exarkun: Cool.  Can you put this into a testing library instead of a test module, though?

flocker/ca/test/test_ca.py
```diff
@@ -341,6 +381,21 @@ def test_certificate_subject_control_service(self):
...
+        """
+        The written certificate has a subjectAltName containing the given
+        hostname.
+        """
+        assert_has_extension(self, self.credential.credential,
+                             b"subjectAltName", b"DNS:control.example.com")
+
+    def test_extendedKeyUsage(self):
+        """
+        The written certificate has extendedKeyUsage set to "serverAuth".
```
exarkun: These two test method docstrings duplicate the prevailing convention in this test suite of saying the assertion is about a certificate that is "written" somewhere but the tests just use an in-memory object.

flocker/ca/test/test_ca.py
```diff
@@ -309,6 +309,13 @@ def test_certificate_subject_username(self):
         self.assertEqual(subject.CN, u"user-{user}".format(
             user=self.credential.username))
 
+    def test_extendedKeyUsage(self):
+        """
+        The written certificate has extendedKeyUsage set to "serverAuth".
```
exarkun: This docstring does as well.

flocker/ca/test/test_ca.py
```diff
@@ -309,6 +309,13 @@ def test_certificate_subject_username(self):
...
+        """
+        assert_has_extension(self, self.credential.credential,
+                             b"extendedKeyUsage", b"clientAuth")
```
exarkun: This doesn't assert the usage that the docstring talks about.

flocker/ca/test/test_ca.py
```diff
@@ -326,9 +333,42 @@ def test_certificate_subject_node_uuid(self):
         self.assertEqual(subject.CN, b"node-{uuid}".format(
             uuid=self.credential.uuid))
 
+    def test_extendedKeyUsage(self):
+        """
+        The written certificate has extendedKeyUsage set to "serverAuth".
```
exarkun: More "written certificate" language here.

flocker/ca/test/test_ca.py
```diff
@@ -326,9 +333,42 @@ def test_certificate_subject_node_uuid(self):
...
+        """
+        assert_has_extension(self, self.credential.credential,
+                             b"extendedKeyUsage", b"clientAuth")
```
exarkun: Also asserting something contrary to the docstring.

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+    :return: Tuple, a pair of ``CredentialSet`` instances.
+    """
+    a = CredentialSet.create()
+    b = CredentialSet.create()
+    global get_sets
+
+    def get_sets():
+        return a, b
+    return get_sets()
```
exarkun: Possibly somewhat over-clever.

:question:

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+
+
+class CredentialSet(PRecord):
+    """
+    A full set of credentials for a CA.
+
+    :param RootCredential root: The CA root credential.
+    :param ControlCredential control: A control service credential.
+    :param UserCredential user: A user credential.
+    :param NodeCredential node: A CA root credentials.
```
exarkun: `ivar`s

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+    user = field()
+    node = field()
+
+    @staticmethod
+    def create():
+        """
+        :return: A new ``CredentialSet``.
+        """
+        directory = FilePath(mkdtemp())
+        atexit.register(rmtree, directory.path)
```
exarkun: Why not `TestCase.mktemp` and no cleanup as we mostly seem to do?

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+         given a port number on localhost and a ``CertificateSet``. For
+         server context factories the port number can be ignored.
+
+    :param str good_certificate_name: Name of certificate (an attribute of
+        ``CredentialSet``) that should validate successfully.
+
+    :param bool validator_is_client: Whether or not the context factory
+         being tested is a client.
+
+    :return TestCase: Tests for given validator.
```
exarkun: I think `:return type:` is not actually valid markup.  And the return type here is `type` anyway. :smile:

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+
+        def assert_validates(self, credential):
+            """
+            Asserts that a TLS handshake is successfully established between a
+            client using the validation logic and a server based on the
+            given credential.
+
+            :param credential: The high-level credential to use.
+
+            :return ``Deferred``: Fires on success.
```
exarkun: markup mumble mumble

:question:

exarkun: Does it do something on failure?

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+                client_context_factory = peer_context_factory
+
+            server_endpoint = SSL4ServerEndpoint(reactor, port,
+                                                 server_context_factory,
+                                                 interface='127.0.0.1')
+            d = server_endpoint.listen(
+                ServerFactory.forProtocol(SendingProtocol))
+            d.addCallback(lambda port: self.addCleanup(port.stopListening))
+            validating_endpoint = SSL4ClientEndpoint(
+                reactor, b"127.0.0.1", port, client_context_factory)
```
exarkun: Documented type is `str` - whatever that means.

:question:

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+            credential.
+
+            :param FlockerCredential credential: The private key/certificate
+                to use for the server.
+
+            :return ``Deferred``: Fires on success (i.e. if no TLS handshake is
+                established).
+            """
+            return self.assertFailure(self.assert_validates(credential),
+                                      SSLError)
```
exarkun: I wouldn't mind seeing these two `assert_` methods as free functions instead.

:question:

exarkun: It would probably be somewhat good to try to make an assertion about what the error was that leads to the `SSLError`.  I admit that might be difficult given the form in which OpenSSL/pyOpenSSL present failures.  The security-sensitive nature of this code probably merits an attempt, though, so we can be sure that when one of these negative-case tests passes it's because of validation and not because some other stupid bugs.

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+
+        def test_same_ca_another_wrong_type(self):
+            """
+            If the expected certificate is generated by the same CA but is of
+            the wrong type the validator will reject it.
+
+            This is different wrong type than ``test_same_ca_wrong_type``.
+            """
+            return self.assert_does_not_validate(
+                getattr(self.another_ca, another_bad_name))
```
exarkun: Another case would be a certificate that purports to be signed by the correct CA but has a signature that does not validate.  This is probably an important case to have work so it would be nice to have a test - but at least it's OpenSSL's responsibility to make it work, not ours (so we'll have lots of company the day the CVE for this comes out).  I don't see an obvious way to construct such a certificate using the APIs in pyOpenSSL, so maybe it's harder than it's worth right now?  Or maybe it would be possible with a hack like getting a validly signed cert and then modifying it (eg change the subject which will result in the existing signature being invalidated)...

:question:

flocker/ca/test/test_ca.py
itamarst: They do actually get written out, but yes.

flocker/ca/functional/test_validation.py
itamarst: I actually had the post-modification case cause validation failures, so I proved once that it fails... so probably fairly easy to reproduce.

flocker/ca/functional/test_validation.py
itamarst: I'll see what I can do.

flocker/ca/_ca.py
```diff
@@ -3,7 +3,25 @@
...
+None of these should be able to impersonate the others. We therefore use
+the following Distinguised Name scheme:
+
+1. Control service: common name is "control-service", subjectAltName is
+   administrator-specifiable DNS hostname, to support standard HTTPS
+   client authentication, and extendedKeyUsage is set to "serverAuth".
+2. Node agents: common name is "node-<uuid>", and extendedKeyUsage is set
+   to "clientAuth" (technically clientAuth means HTTPS, but whatevs).
+3. API clients: Common name is set to "user-<username>", and
+   extendedKeyUsage is set to "clientAuth".
```
exarkun: The correct thing for us to do is probably create a new extension type and use that.  This is probably a lot of work and apart from being less abusive of the x509v3 standard (with has interop consequences) I'm not sure it actually makes the software better (rather, if it makes it better, I'm not sure in what ways).  It's perhaps worth a note about follow-up work, though?

:question:

flocker/ca/functional/test_validation.py
itamarst: Trial catches it and fails? I guess given this code's importance I should be more explicit.

flocker/ca/functional/test_validation.py
itamarst: 1. It's test setup code, so I feel less bad about it...
2. ... and it makes things much faster on my (very new) machine.

flocker/ca/_ca.py
```diff
@@ -344,8 +365,9 @@ def initialize(cls, output_path, authority, username, begin=None):
         cert = sign_certificate_request(
             authority.credential.keypair.keypair,
             authority.credential.certificate.getSubject(), request,
-            serial, EXPIRY_20_YEARS, b'sha256', start=begin
-        )
+            serial, EXPIRY_20_YEARS, b'sha256', start=begin,
+            additional_extensions=[crypto.X509Extension(
+                b"extendedKeyUsage", False, b"clientAuth")])
```
exarkun: I wish flake8 was opinionated about closing `)` formatting.

:question:

flocker/ca/functional/test_validation.py
itamarst: Cause it's shared between tests. Really what we need is a fixture framework.

flocker/ca/_ca.py
```diff
@@ -432,19 +455,20 @@ class ControlCredential(PRecord):
     :ivar FlockerCredential credential: The certificate and key pair
         credential object.
     """
-    credential = field(mandatory=True)
+    credential = field(mandatory=True, type=FlockerCredential)
 
     @classmethod
-    def from_path(cls, path):
+    def from_path(cls, path, hostname):
         keypair, certificate = load_certificate_from_path(
```
exarkun: A docstring for this method would be nice.

flocker/ca/_validation.py
```diff
@@ -0,0 +1,108 @@
...
+        :param Certificate ca_certificate: The certificate authority's
+            certificate.
+
+        :param ControlCredential control_credential: The control service's
+            credentials.
+
+        :param bytes prefix: The required prefix on certificate common names.
+        """
+        key = control_credential.credential.keypair.keypair.original
+        certificate = control_credential.credential.certificate.original
```
exarkun: Blarghy blargy blarg blarh blaaaa...

How about giving `ControlCredential` a method which returns a suitably constructed `CertificateOptions` and this code can call it instead of reaching way down into `ControlCredential`'s bowels for these pieces?

Trying to resolve the meaning of these two lines, I also noticed that `FlockerCredential.keypair` is documented as a `FlockerKeyPair` - which is not actually defined anywhere.

flocker/ca/functional/test_validation.py
exarkun: `a` and `b` (with better names :smile: ) could be the globals.  I think that might be less confusing to maintainers than a function that replaces itself.

flocker/ca/functional/test_validation.py
exarkun: Ah, yes, of course.

flocker/ca/functional/test_validation.py
```diff
@@ -0,0 +1,274 @@
...
+        self._buffer = b""
+
+    def dataReceived(self, data):
+        self._buffer += data
+        if self._buffer == EXPECTED_STRING:
+            self.result.callback(None)
+            self.result = None
+
+    def connectionLost(self, reason):
+        if self.result:
```
exarkun: `... is not None:`

exarkun:
Thanks.  Looks good.  Also, "Coverage for flocker/ca/_validation : 100%".  Good work.

Please address inline comments to your satisfaction and then merge.

PR #1370, opened 2015-05-05
[FLOC-1803] Rely on Docker restart policies; no longer restart stopped containers ourselves.

itamarst:
Question for reviewer: perhaps our default policy should be restart always, instead of restart never?

*Update:* Rob has stated this is the desired default.

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -757,23 +757,22 @@ class ApplicationNodeDeployer(object):
         if desired_open_ports != set(self.network.enumerate_open_ports()):
             phases.append(OpenPorts(ports=desired_open_ports))
 
-        current_node_applications = set(
+        running_applications = set(
             app for app in current_node_state.applications if app.running)
         all_applications = current_node_state.applications
 
         # Compare the applications being changed by name only.  Other
         # configuration changes aren't important at this point.
-        current_state = {app.name for app in current_node_applications}
+        running = {app.name for app in running_applications}
         desired_local_state = {app.name for app in
                                desired_node_applications}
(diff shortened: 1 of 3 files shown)


flocker/node/_deploy.py
```diff
@@ -757,23 +757,22 @@ def calculate_changes(self, desired_configuration, current_cluster_state):
...
             app.name for app
-            in all_applications.difference(current_node_applications)}
+            in all_applications.difference(running_applications)}
 
         # Don't start applications that exist on this node but aren't
-        # running; instead they should be restarted:
-        start_names = desired_local_state.difference(
-            current_state | not_running)
+        # running; Docker is in charge of restarts:
+        start_names = desired_local_state.difference(running | not_running)
```
tomprince: Is there any reason to have both `running` and `not_running` here? It looks like this is the only usage of them.

flocker/node/_deploy.py
```diff
@@ -757,23 +757,22 @@ def calculate_changes(self, desired_configuration, current_cluster_state):
...
+        running_applications = set(
             app for app in current_node_state.applications if app.running)
         all_applications = current_node_state.applications
 
         # Compare the applications being changed by name only.  Other
         # configuration changes aren't important at this point.
-        current_state = {app.name for app in current_node_applications}
+        running = {app.name for app in running_applications}
         desired_local_state = {app.name for app in
                                desired_node_applications}
```
tomprince: :question: There is inconsistent naming of things having sets of names, vs `Applications` here, which makes reviewing the code slightly harder.

PR #1360, opened 2015-05-01
Replace gsutil with awsutil

adamtheturtle:
This can be reviewed but should not be merged yet because there is one outstanding use of `gsutil` in the release process. The issue (currently awaiting review) to remove this is https://clusterhq.atlassian.net/browse/FLOC-1718.

There are other uses of gsutil in the docs but some are going away, none already told you to use Flocker's setup.py, and most are very rarely run.

PR #1373, opened 2015-05-06
Revert "Replace gsutil with awsutil"

adamtheturtle:
Reverts ClusterHQ/flocker#1360


exarkun:
Why?

adamtheturtle:
I put explanation of this on FLOC-1738 but probably should have put it here too. That PR was not meant to be merged before a blocking issue was merged - I realised that seconds after merging #1360 so reverted it and put a note on the issue to revert this when ready.

PR #1374, opened 2015-05-06
[FLOC-1823] Don't leak build path into dpkg repository metadata.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1823

### DIFF
--- a/admin/test/test_release.py
+++ b/admin/test/test_release.py
@@ -5,10 +5,14 @@ Tests for ``admin.release``.
 """
 
 import os
-from unittest import skipUnless
-from setuptools import __version__ as setuptools_version
+
+from gzip import GzipFile
+from StringIO import StringIO
 import tempfile
 from textwrap import dedent
+from unittest import skipUnless
+
+from setuptools import __version__ as setuptools_version
 
 from effect import sync_perform, ComposedDispatcher, base_dispatcher
 from git import Repo
@@ -1130,6 +1134,9 @@ class UpdateRepoTests(SynchronousTestCase):
         """
         Calling :func:`update_repo` with real dpkg utilities creates a
         repository in S3.
+
+        The filenames in the respository metadata does not have the build directory
+        in them.
         """
         source_repo = FilePath(self.mktemp())
         source_repo.createDirectory()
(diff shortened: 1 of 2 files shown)


admin/test/test_release.py
```diff
@@ -1177,6 +1184,13 @@ def get_dispatcher(self):
         # given RPMs, not that any metadata files are copied.
         self.assertEqual(expected_files, set(files_on_s3.keys()))
 
+        # The repository is build in self.packages_directory
+        # Ensure that that does not leak into the metadata.
+        packages_gz = files_on_s3[os.path.join(self.target_key, 'Packages.gz')]
+        with GzipFile(fileobj=StringIO(packages_gz), mode="r") as f:
+            packages_metadata = f.read()
+        self.assertNotIn(self.package_directory.path, packages_metadata)
+
```
adamtheturtle: Put some explanation of why this test has multiple assertions - perhaps because the set up is particularly slow (but then https://testtools.readthedocs.org/en/latest/for-test-authors.html#delayed-assertions).

admin/yum.py
```diff
@@ -123,7 +123,8 @@ def perform_create_repository(dispatcher, intent):
         metadata = check_output([
             b'dpkg-scanpackages',
             b'--multiversion',
-            intent.repository_path.path])
```
adamtheturtle: :question: Put a comment before each option to explain what it does - I've found this useful elsewhere in our source.

adamtheturtle:
Thanks @tomprince this looks good. I will do some testing on S3 tomorrow and then if it works I will accept this.

admin/test/test_release.py
```diff
@@ -1177,6 +1184,13 @@ def get_dispatcher(self):
         # given RPMs, not that any metadata files are copied.
         self.assertEqual(expected_files, set(files_on_s3.keys()))
 
+        # The repository is build in self.packages_directory
+        # Ensure that that does not leak into the metadata.
+        packages_gz = files_on_s3[os.path.join(self.target_key, 'Packages.gz')]
```
jongiddy: 'build' -> 'built'
Reference the issue that prompted this test: FLOC-1823 :question:

admin/test/test_release.py
```diff
@@ -1130,6 +1134,9 @@ def test_real_dpkg_utils(self):
         """
         Calling :func:`update_repo` with real dpkg utilities creates a
         repository in S3.
+
+        The filenames in the respository metadata does not have the build directory
```
jongiddy: 'respository' -> 'repository'
'does not have' -> 'do not have'
Also this line is > 80 characters

jongiddy:
This can be merged, once the minor comments have been addressed.

admin/test/test_release.py
tomprince: I judged the cost (in both time and code-complexity) in refactoring the code to separate the two assertions isn't worth it for this support code.

PR #1377, opened 2015-05-07
Control service yml floc 1801

adamtheturtle:
Unlike the design at https://github.com/ClusterHQ/flocker/pull/1338 this does not specify TCP. It is possible to have a "protocol" key and just make sure that it is TCP but I think that it is better to just leave it and add an option later if necessary.

As per the previous PR, there is some documentation duplication and the documentation is not in the best place. However David Oliver is looking into restructuring the documentation and will look at this story in particular before submission so I don't think that should be solved here.

### DIFF
--- a/docs/advanced/api.rst
+++ b/docs/advanced/api.rst
@@ -32,11 +32,15 @@ On AWS, an external firewall is used instead, which will need to be configured s
 To start the agents on a node, a configuration file must exist on the node at ``/etc/flocker/agent.yml``.
 This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the control node:
 
+# TODO document what the port variable is, here and in Ubuntu section
+# TODO maybe make port an option to
+
 .. code-block:: yaml
 
    "version": 1
    "control-service":
       "hostname": "${CONTROL_NODE}"
+      "port": 4524
 
 .. task:: enable_flocker_agent fedora-20 ${CONTROL_NODE}
    :prompt: [root@agent-node]#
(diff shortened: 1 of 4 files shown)


flocker/provision/_install.py
```diff
@@ -229,10 +229,11 @@ def task_enable_flocker_agent(distribution, control_node):
...
             {
                 "version": 1,
-                "control-service-hostname": control_node,
+                "control-service": {
+                    "hostname": control_node,
+                    "port": 4524,
+                },
             },
-            # Don't wrap the whole thing in braces
-            default_flow_style=False,
```
adamtheturtle: This didn't work well with the new dictionary style unfortunately.

docs/advanced/api.rst
```diff
@@ -30,12 +30,15 @@ For more details on configuring the firewall, see Fedora's `firewalld documentat
 On AWS, an external firewall is used instead, which will need to be configured similarly.
 
 To start the agents on a node, a configuration file must exist on the node at ``/etc/flocker/agent.yml``.
-This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the control node:
+This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the control node.
+The ``port`` variable is the port on the control node to connect to:
```
itamarst: Since 99% of the time it's going to be 4524, it would be nice if port was optional. Can be follow up issue since it's not critical, just nice to have.

itamarst: (i.e. we can do it if we have time before 1.0, but don't have to)

itamarst:
TCP option is not required, we're just going with TLS across the board. Merge when you're happy with Buildbot.

docs/advanced/api.rst
adamtheturtle: This was quick so I did it.

PR #1378, opened 2015-05-07
Change virtualbox settings

adamtheturtle:
As mentioned on the issue, I have done this issue because the slow network speed is annoying when testing packaging changes.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -16,6 +16,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.box = "clusterhq/flocker-dev"
   config.vm.box_version = "> 0.3.2.1714"
 
+  # The following reduces the difference in network speed between the guest
+  # and the host. See https://github.com/mitchellh/vagrant/issues/1807.
+  config.vm.provider :virtualbox do |vb|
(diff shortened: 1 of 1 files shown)


itamarst:
I note the vagrant builder is unhappy - is that unrelated?

adamtheturtle:
@itamarst Yes, there is WIP to fix that. See https://clusterhq.atlassian.net/browse/FLOC-1810. I have built this manually and tested it.

adamtheturtle:
Oops, I forgot to say:

I tested the speed with https://github.com/sivel/speedtest-cli.
I ran this a few times on each system and while the results below are not averages, the results were quite consistent.

## Locally

Latency: 10.058 ms
Download: 29.80 Mbit/s
Upload: 36.12 Mbit/s

## On the Vagrant box without the proposed changes:

Latency: 20.079 ms
Download: 6.11 Mbit/s Mbit/s
Upload: 1.95 Mbit/s

## On the Vagrant box with the proposed changes:

Latency: 14.044 ms
Download: 12.22 Mbit/s
Upload: 31.21 Mbit/s

In the issue I mentioned the `config.vm.network "public_network"` possibility, but the Vagrant docs say that this is due to change and it doesn't seem suitable. Also you don't want to choose which interface to bridge to.

itamarst:
OK, merge.

PR #1379, opened 2015-05-07
[FLOC-1269] Design sketch of expiring cluster state updates

itamarst:


flocker/control/_clusterstate.py
```diff
@@ -21,8 +21,19 @@ class ClusterStateService(Service):
...
     """
-    def __init__(self):
+    def __init__(self, reactor):
         self._deployment_state = DeploymentState()
+        # self._information_wipers = pmap() # map ``IClusterStateWipe.key`` to its instance.
+        # self._information_wipers_added = pmap() # map ``IClusterStateWipe.key`` to timestamp
+
+    def startService(self):
+        # Start LoopingCall that applies all wipers whose added time is
+        # more than ten seconds in the past.
```
exarkun: Maybe make this a MultiService with a child TimerService and leave start/stop glue up to MultiService

exarkun: File a follow-up issue for replacing "ten seconds" with ... something.

flocker/control/_clusterstate.py
```diff
@@ -58,3 +69,10 @@ def apply_changes(self, changes):
             self._deployment_state = change.update_cluster_state(
                 self._deployment_state
             )
+        # for change in changes:
+        #     wiper = change.get_information_wipe()
+        #     key = wiper.key()
+        #     self._information_wipers = self._information_wipers.set(
+        #         key, wiper)
+        #     self._information_wipers_added = self._information_wipers.set(
+        #         key, time())
```
exarkun: `reactor.seconds()` I guess?

exarkun: If you feel like it, `_information_wipers_added` could probably be a pvector of (timestamp, key) that gets appended to.  The cleanup step can just process a prefix of it, then, instead of visiting the entire collection.  If not, file a follow-up ticket.

exarkun:
Thanks.  This looks nice.  Please proceed.

PR #1382, opened 2015-05-07
[FLOC-1755] Don't mess with in use datasets in block device deployer

itamarst:


flocker/node/_deploy.py
```diff
@@ -435,6 +435,41 @@ def run(self, deployer):
         return gather_deferreds(results)
 
 
+class NotInUseDatasets(object):
+    """
+    Filter out datasets that are in use by applications.
+
+    For now we delay things like deletion we know applications
+    aren't using the dataset. Later on we'll use leases to decouple
```
exarkun: Missing a word here, perhaps?  I don't quite follow this sentence.

flocker/node/_deploy.py
```diff
@@ -435,6 +435,41 @@ def run(self, deployer):
...
+    the application and dataset logic better.
+    """
+    def __init__(self, node_state):
```
exarkun: Maybe a class method, `from_node_state(node_state)`

flocker/node/agents/blockdevice.py
```diff
@@ -1511,6 +1513,16 @@ def _mountpath_for_dataset_id(self, dataset_id):
     def calculate_changes(self, configuration, cluster_state):
         this_node_config = configuration.get_node(
             self.node_uuid, hostname=self.hostname)
+        local_state = cluster_state.get_node(self.node_uuid,
+                                             hostname=self.hostname)
+
+        # We need to know applications (for now) to see if we should delay
+        # deletion or handoffs. Eventually this will rely on leases instead.
+        if local_state.applications is None:
+            return in_parallel(changes=[])
```
exarkun: Was there another issue filed for this part?  It's pretty related, so maybe not... But it'd be nice to make sure it gets closed at the same time if so.

exarkun: I guess there should be a link to an issue about leases here, too.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -588,6 +590,24 @@ class ScenarioMixin(object):
         }
     )
 
+    @staticmethod
+    def add_application_with_volume(node_state):
```
exarkun: A free function, instead?

exarkun:
Thanks.  This looks good.  Address minor inline comments and merge if the buildbot failures are unrelated.

flocker/node/_deploy.py
itamarst: Well, the implementation is going to change anyway when we switch to leases, so having a specific `__init__`? Eh. If this ever becomes more generally useful maybe.

flocker/node/agents/blockdevice.py
itamarst: I don't think there was another issue, no.

PR #1388, opened 2015-05-08
Introduce the blockdevicefactory module, primarily the get_blockdeviceapi helper function.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1836

Taken from #1347

PR #1389, opened 2015-05-08
OpenStack logging tools

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1838

Depends on #1388

[Relative diff](https://github.com/ClusterHQ/flocker/compare/blockdevice-factory-FLOC-1836...openstack-logging-FLOC-1838)

### DIFF
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -9,14 +9,46 @@ from zope.interface.interface import Method
 from twisted.internet.threads import deferToThreadPool
 
 
-def _threaded_method(sync_name, method_name, reactor_name, threadpool_name):
+# TODO: Add tests and documentation for this, make it public (somewhere else).
+def _interface_decorator(decorator_name, interface, method_decorator,
+                         *args, **kwargs):
+    """
+    Create a class decorator which applies a method decorator to each method of
+    an interface.
+
+    :param str decorator_name: A human-meaningful name for the class decorator
+        that will be returned by this function.
+    :param zope.interface.InterfaceClass interface: The interface from which to
+        take methods.
+    :param method_decorator: A callable which will decorate a method from the
+        interface.  It will be called with the name of the method as the first
+        argument and any additional positional and keyword arguments passed to
+        ``_interface_decorator``.
+
+    :return: The class decorator.
+    """
+    for method_name in interface.names():
+        if not isinstance(interface[method_name], Method):
+            raise TypeError(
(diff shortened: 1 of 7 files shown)


flocker/common/test/test_openstack.py
```diff
@@ -0,0 +1,142 @@
...
+                u"url": result.url,
+                u"method": result.method,
+            },
+        )
+
+    @capture_logging(lambda self, logger: None)
+    def test_keystone_client_exception(self, logger):
+        """
+        ``keystoneclient.openstack.common.apiclient.exceptions.BadRequest`` is
+        treated similarly to ``novaclient.exceptions.ClientException``.
```
tomprince: Either refer explicitly to the test whose behaviour this test is going analogous to, or explicitly explain the expected behaviour; it isn't clear that people will be reading the test docstrings in order.

flocker/common/_thread.py
```diff
@@ -9,14 +9,46 @@
 from twisted.internet.threads import deferToThreadPool
 
 
-def _threaded_method(sync_name, method_name, reactor_name, threadpool_name):
+# TODO: Add tests and documentation for this, make it public (somewhere else).
```
tomprince: Link to an issue.

tomprince:
After #1388 is merged, the above two minor comments are addressed, and buildbot looks statisfactory, merge.

tomprince:
Actually, looking at where this is used, is there reason `auto_openstack_logging` doesn't just return a class that does the appropriate thing? I don't see any obvious use for the decorated class to be anything other than a class with a single field and nothing else.

exarkun:
> Actually, looking at where this is used, is there reason auto_openstack_logging doesn't just return a class that does the appropriate thing? I don't see any obvious use for the decorated class to be anything other than a class with a single field and nothing else.

I think you're right.  I don't want to spend time refactoring this now, though.  The awkward interface is worth being able to re-use the guts of the implementation for now.  I'd be happy to see this improved later, though.

tomprince:
You could just do
```python
def auto_openstack_logging(cls, interface):
    @_interface_decorator(...)
    class Wrapper(PRecord):
        _original = field(mandatory=true)
    return wrapper
```
or something slightly more complicated with `type` to get nicer names. That would allow you to re-use the guts without major refactoring.

On further thought, I guess maybe every use `_interface_decorator` could be refactored the same way, so maybe it doesn't make sense to do here.

PR #1391, opened 2015-05-08
[FLOC-1824] Use Nova API calls to get compute_instance_id

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1824

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -107,3 +107,8 @@ class CinderBlockDeviceAPIInterfaceTests(
         self.addCleanup(blockdevice_api2.destroy_volume,
                         flocker_volume.blockdevice_id)
         self.assert_foreign_volume(flocker_volume)
+
+    # Maybe add an implementation specific test here to verify that
+    # the ID returned by compute_instance_id() can be retreived using
+    # the nova server get API and that it has our IP addesses
+    # http://developer.openstack.org/api-ref-compute-v2.html#getServer
--- a/flocker/node/test/test_script.py
+++ b/flocker/node/test/test_script.py
@@ -90,6 +90,8 @@ def get_all_ips():
     """
     Find all IPs for this machine.
 
+    XXX: Move this to flocker.common.???
+
     :return: ``list`` of IP addresses (``bytes``).
     """
     ips = []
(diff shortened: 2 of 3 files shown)


flocker/node/agents/functional/test_cinder.py
```diff
@@ -107,3 +107,8 @@ def test_foreign_cluster_volume(self):
         self.addCleanup(blockdevice_api2.destroy_volume,
                         flocker_volume.blockdevice_id)
         self.assert_foreign_volume(flocker_volume)
+
+    # Maybe add an implementation specific test here to verify that
+    # the ID returned by compute_instance_id() can be retreived using
+    # the nova server get API and that it has our IP addesses
+    # http://developer.openstack.org/api-ref-compute-v2.html#getServer
```
exarkun: I think our existing test for `get_device_path` might cover this.  It requires that after attaching to our `compute_instance_id` the reported device path is an actual block device on the system where the tests are running.  That's a pretty good indication that we've gotten the right compute instance id.

flocker/node/test/test_script.py
```diff
@@ -90,6 +90,8 @@ def get_all_ips():
     """
     Find all IPs for this machine.
 
+    XXX: Move this to flocker.common.???
+
```
exarkun: Hm.  This is really `get_all_ipv4s`.  Maybe put a real `get_all_ips` (ipv4 and ipv6) in `flocker.common` and we can think about transitioning everything to that and away from this.

You're planning to include ipv6 addresses in the matching, right?

exarkun:
Thanks.  Looks pretty good.  One other thing to think about ... is `xenstore-read` better than IP matching on systems where `xenstore-read` exists?

flocker/node/agents/functional/test_cinder.py
wallrj: True. This isn't needed then.

flocker/node/test/test_script.py
wallrj: Yep, I'll modify it to get *all* IPs...assuming the API returns IPv6 addresses.

wallrj:
`xenstore-read` is faster  and more precise but requires root.
`/var/lib/cloud/data/instance-id` is as precise, but available to non-root users.

We could implement a function that tried these in order, falling back to API calls but, as discussed, it'll add complexity and I'll leave that for a follow up issue.

PR #1392, opened 2015-05-08
Wait until `host` can resolve `clusterhq.com` before really provisioning.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1842

(Quite possibly this is not helpful, need to see what happens on buildbot.)


tomprince:
I don't this is the issue. (I can reproduce the issue in a local VM that has a working resolver.

PR #1393, opened 2015-05-09
Ebs attach detach get_device_path floc 1813

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1813

Draft of attach_volume, detach_volume implementations for EBS.


flocker/node/agents/ebs.py
```diff
@@ -152,6 +158,58 @@ def compute_instance_id(self):
...
+            if volume.blockdevice_id == blockdevice_id:
+                return volume
+        raise UnknownVolume(blockdevice_id)
+
+    def _get_ebs_volume(self, blockdevice_id):
+        """
+        Lookup EBS Volume information for a given blockdevice_id.
+
+        :param unicode blockdevice_id: ID of a blockdevice that needs lookup.
+        :returns boto.ec2.volume.Volume for the input id.
```
adamtheturtle: Document in these methods when `None` is returned.

myechuri: Thanks for pointing out! Fixed.

flocker/node/agents/ebs.py
```diff
@@ -195,11 +300,71 @@ def resize_volume(self, blockdevice_id, size):
...
+        :raises UnknownVolume: If there does not exist a BlockDeviceVolume
+            corresponding to the input blockdevice_id.
+        :raises AlreadyAttachedVolume: If the input volume is already attached
+            to a device.
+        """
+        volume = self._get(blockdevice_id)
+        if volume.attached_to is not None:
+            raise AlreadyAttachedVolume(blockdevice_id)
+
+        self.lock.acquire()
```
exarkun: I think you can use `with` with this lock object.  It's a "context manager".  This is syntactic sugar for doing the `release` in the equivalent of a `finally` block at the end of a try/finally.  So this makes the code a bit clearer (you get an indented block for all of the code run while the lock is held) and safer (if an exception is raised the lock is released for you) and shorter (you can delete the explicit release because `with` will do it for you).

flocker/node/agents/ebs.py
```diff
@@ -195,11 +300,71 @@ def resize_volume(self, blockdevice_id, size):
...
+        blockdevices = glob.glob('/sys/block/*')
```
exarkun: Something like `FilePath(b"/sys/block").children()` would be better here.  `*` is a simple enough glob that you don't really need globbing and `twisted.python.filepath.FilePath` should typically be preferred for path-related operations (instead of bare strings which are more error prone).

flocker/node/agents/ebs.py
```diff
@@ -152,6 +201,62 @@ def compute_instance_id(self):
...
+        :param unicode instance_id: EC2 instance ID.
+
+        :returns unicode file_name: available device name for attaching
+            EBS volume.
+        :returns ``None`` if suitable EBS device names on this EC2
+            instance are currently occupied.
+        """
+        volumes = self.connection.get_all_volumes()
+        devices = [v.attach_data.device for v in volumes
+                   if v.attach_data.instance_id == instance_id]
```
exarkun: I'm curious/confused about the use of `attach_data.device` here compared to the metadata/tag strategy used elsewhere.

Is `attach_data.device` reliable?  If so, why not use it in `get_device_path`?  If not, will this function make bad recommendations about what device name to try to use next (and if Amazon doesn't promise to honor the requested device, does it matter what name we ask for)?

flocker/node/agents/ebs.py
```diff
@@ -152,6 +201,62 @@ def compute_instance_id(self):
...
+        for prefix in ['f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p']:
```
exarkun: Strings are iterable.  I'm not sure if this would be better or worse as `for prefix in "fghijklmnopq":`.  It'd be shorter, anyway. :wink:   One reason to change might be that these should probably all be explicit unicode or byte string literals and `b"fghijklmnopq"` is a lot easier than `b"f", b"g", ...`.

Speaking of, we generally treat filesystem paths as bytes (because we don't know the filesystem encoding so treating them as unicode is harder).  There are some exceptions and often it doesn't matter (as in this case where the paths are always pure ASCII which is easy to convert between unicode and bytes).  `FilePath` doesn't support `unicode`, though, so that's a good reason to stick with bytes.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +117,44 @@ def _wait_for_volume(expected_volume,
...
+        to create a new block device.
+
+    :returns unicode device: Name of the new block device
+        we are interested in.
+
+    :raises Exception: When new EBS block device did not
+        manifest in OS within time limit.
+
+    """
+    patterns = ['sd.*', 'xvd.*']
```
exarkun: These could be prefixes instead of patterns and then you could use `device.startswith(...)` instead of using regexp matching.  Less powerful - but still sufficiently powerful.  And it's nice to avoid regexps whenever possible.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +117,44 @@ def _wait_for_volume(expected_volume,
...
+    start_time = time.time()
+    elapsed_time = time.time() - start_time
+    while elapsed_time < time_limit:
+        for device in list(set(glob.glob('/sys/block/*')) - set(base)):
```
exarkun: More glob stuff here that should likely use `FilePath` instead

flocker/node/agents/ebs.py
```diff
@@ -152,6 +201,62 @@ def compute_instance_id(self):
...
+        Lookup EBS Volume information for a given blockdevice_id.
+
+        :param unicode blockdevice_id: ID of a blockdevice that needs lookup.
+
+        :returns boto.ec2.volume.Volume for the input id. ``None`` if
+            no boto.ec2.volume.Volume was found for the given id.
+        """
+        for volume in self.connection.get_all_volumes():
+            if volume.id == blockdevice_id:
+                # Sync volume for uptodate metadata
```
exarkun: The volume just came back from a `get_all_volumes` call.  Does it already have out-of-date metadata?  Seems like it would be nice if boto gave us an object that was up-to-date at the time of the `get_all_volumes` call.

flocker/node/agents/ebs.py
```diff
@@ -152,6 +201,62 @@ def compute_instance_id(self):
...
+        :returns unicode file_name: available device name for attaching
+            EBS volume.
+        :returns ``None`` if suitable EBS device names on this EC2
+            instance are currently occupied.
+        """
+        volumes = self.connection.get_all_volumes()
+        devices = [v.attach_data.device for v in volumes
+                   if v.attach_data.instance_id == instance_id]
+        for prefix in ['f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p']:
+            file_name = u'/dev/sd%c' % prefix
```
exarkun: Always interpolate a tuple with `%`.  For example, ` u'/dev/sd%c' % (prefix,)`.  Or use the new style of formatting (particularly, or maybe only, with unicode strings): ` u'/dev/sd{}'.format(prefix)`.  Or there's simple string concatenation - `u'/dev/sd' + prefix`.  (Which also makes it a bit more clear that this is actually a suffix :smile: )

flocker/node/agents/ebs.py
```diff
@@ -60,10 +69,11 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
 
     :return: Input volume in BlockDeviceVolume format.
     """
+    ebs_volume.update()
     return BlockDeviceVolume(
         blockdevice_id=unicode(ebs_volume.id),
         size=int(GB(ebs_volume.size).to_Byte().value),
-        attached_to=None,
+        attached_to=ebs_volume.attach_data.instance_id,
```
exarkun: I guess `attach_data` is set to something (not `None`) even when the volume isn't attached - or some tests would be failing when `create_volume` calls this function with an unattached volume.  Or maybe those tests would fail if I tried to run them on an EC2 instance.  Since we don't have buildbot set up to run these yet, I'll just ask - did you run the full ebs functional test suite and see it all pass?

exarkun:
Thanks.  This looks promising so far.  One other question Richard and I had was what happens to an EC2 instance with attached EBS volumes when the EC2 instance is rebooted (and does it matter how it is rebooted - soft reboot from the OS or something harder initiated by some EC2 API)?

If volumes stay attached, do we have to rely on the tags we put on them to figure out what device files they're associated with?  What about that `attach_data` attribute volumes have?

Please address the inline comments and share anything you might know about the reboot cases (which we may not actually need to **handle** yet - or we can at least defer 'til later) and re-submit.  Thanks again.

wallrj:
I wonder if you can get at the virtual serial number of the virtual disk?

Or if there's some other unique label assigned to the block devices.

On Rackspace I see:

```
[~]$ cat /sys/block/xvd*/device/nodename 
device/vbd/768
device/vbd/51792
```
 * https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-bus-xen-backend


Of course, none of this relates to EBS, but thought I'd paste the links somewhere before I close the tabs:
 * https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/5/html-single/Online_Storage_Reconfiguration_Guide/#persistent_naming
 * http://en.wikipedia.org/wiki/World_Wide_Name
 * https://www.kernel.org/doc/Documentation/ABI/stable/sysfs-bus-xen-backend
 * http://stackoverflow.com/questions/4193514/how-to-get-hard-disk-serial-number-using-python
 * https://wiki.archlinux.org/index.php/Persistent_block_device_naming#by-id_and_by-path

flocker/node/agents/ebs.py
myechuri: Thanks for the suggestion and reasoning! Changed to `with`.

flocker/node/agents/ebs.py
myechuri: Fixed. Thanks for explaining!

flocker/node/agents/ebs.py
myechuri: Sorry for not being clear in the comments. `attach_data.device` points to the device name you chose during attach_volume. The tag i added carries the eventual device name picked by OS's device driver. For example `attach_volume(..., '/dev/sdg',...)` might end up attaching the volume to `/dev/xvdh`. All EC2 instance APIs, including `attach_data.device` will show `/dev/sdg`. The tag i add for the volume will show `/dev/xvdh`. `get_device_path` needs `/dev/xvdh` because that is what the OS recognizes, so, we need to use that for creating file systems, mounting, etc. `/dev/sdg` is pretty much useless for our consumption.

flocker/node/agents/ebs.py
myechuri: Definitely cleaner. Thank you so much for the reasoning!

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py
myechuri: i actually added the `volume.update()` calls in ebs.py due to failing tests. Recent attach/detach information is not updated right away in EBS volume info returned from Boto unfortunately.

flocker/node/agents/ebs.py
myechuri: Thanks a lot! Change to `u'/dev/sd' + suffix`.

flocker/node/agents/ebs.py
myechuri: Yes, ran all tests except `*resize*` and `test_detach_volume`. `test_detach_volume` has a mount failure that i need to look into. Rest all tests, listed in `test_ebs.py` in this pull request pass.

myechuri:
```
what happens to an EC2 instance with attached EBS volumes when the EC2 instance is rebooted (and does it matter how it is rebooted - soft reboot from the OS or something harder initiated by some EC2 API)?
```

There is a delete_on_termination attribute associated with a volume. By default, it is set to `False`, which works for us since we want the volumes to persist across node reboots. This was my reasoning for not overriding default. Please let me know if it doesn't sounds reasonable? Thanks!

myechuri:
```
I wonder if you can get at the virtual serial number of the virtual disk?
Or if there's some other unique label assigned to the block devices.
```
Yes, there exists device nodename on EC2 (Ubuntu) instance as well:
```
ubuntu@ip-172-31-25-23:~/flocker/flocker/node/agents$ cat /sys/block/xvdj/device/nodename 
device/vbd/51856
ubuntu@ip-172-31-25-23:~/flocker/flocker/node/agents$ 
```
But there is no way to map `/dev/sdj` to `/dev/xvdj`.

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,68 @@ def _wait_for_volume(expected_volume,
...
+    suffix matches the input size.
+
+    :param unicode device: Name of the block device to check for size.
+    :param int size: Size, in SI metric bytes, of device we are interested in.
+
+    :returns True if a block device with given name has given size.
+        False otherwise.
+    """
+    device_name = b"/dev/" + device
+    command = [b"lsblk", b"-nb", b"-o", b"SIZE", device_name]
```
exarkun: Can you switch this to use long options instead?  Where possible, we prefer long options.  Where tools only support short options, we usually include some comments explaining them so the code is a little bit more readable.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,68 @@ def _wait_for_volume(expected_volume,
...
+    Helper function to check if the size of block device with given
+    suffix matches the input size.
+
+    :param unicode device: Name of the block device to check for size.
+    :param int size: Size, in SI metric bytes, of device we are interested in.
+
+    :returns True if a block device with given name has given size.
+        False otherwise.
+    """
+    device_name = b"/dev/" + device
```
exarkun: Concatenating `bytes` and `unicode` here.  Should be something like `b"/dev/" + device.encode("ascii")` instead.

exarkun: (which is exactly what Python will implicitly do for you in this case but we **try** not to rely on implicit encode/decode anywhere)

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,68 @@ def _wait_for_volume(expected_volume,
...
+    command = [b"lsblk", b"-nb", b"-o", b"SIZE", device_name]
+    device_size = int(check_output(command).strip().decode("ascii"))
+
+    # OS reports device size in powers of 2, where has EBS
+    # talks powers of 10. Hence, convert before comparing.
+    if size/(1000**3) == device_size/(1024**3):
```
exarkun: `size` and `device_size` are both in bytes.  I would have expected `size == device_size` to be the reasonable comparison to make here.

If `size/(1000**3) == device_size/(1024**3)` determines the sizes match, isn't it because either:

  * this uses integer division so many different sizes are effectively rounded to the same number of GB or GiB (eg, `3 * 10 ** 9 / 2 ** 30 == 3` even though there's a clear unit conversion error there)
  * or the Amazon volumes are being created with the wrong size which the unit mismatch here exactly masks (looking at http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.create_volume I think this might be the case, it says "GiB" but I think the `create_volume` implementation passes in a value as if it were meant to be GB)

exarkun: Having seen your use of `lsblk` above and realizing that the existing test suite didn't catch the GB/GiB mismatch, we should probably have a couple new tests in the test suite generated by `make_iblockdevice_tests`.  One (most importantly) which creates a volume and then attaches it and uses `lsblk` to verify it has the right size.  And another similar test for resize (but since we're probably dropping resize from 1.0 that can be lower priority).  This would provide an outside confirmation that sizes were being handled correctly instead of relying on the object being tested to tell us.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,68 @@ def _wait_for_volume(expected_volume,
...
+    """
+    Helper function to wait for up to 60s for new
+    EBS block device (`/dev/sd*` or `/dev/xvd*`) to
+    manifest in the OS.
+
+    :param list base: List of baseline block devices
+        that existed before execution of operation that expects
+        to create a new block device.
+
+    :returns unicode device: Name of the new block device
```
exarkun: `:returns unicode device:` won't be valid markup.  I'm still not sure whether `:returns unicode:` is or not (I think possibly you have to write `:returns: ...` and separately `:rtype: ...` though I've started writing `:returns unicode:` with the optimistic viewpoint that we'll just fix the tool chain to support that form, since it's easier to write).

exarkun: Parameter docs for `size` are missing

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,68 @@ def _wait_for_volume(expected_volume,
...
+        we are interested in.
+
+    :raises Exception: When new EBS block device did not
+        manifest in OS within time limit.
```
exarkun: We should probably not actually raise `Exception` itself anywhere.  Feel free to define any new `Exception` subclasses you like and raise them instead.  Since exception handling in Python is **largely* type-based (that is, you write things like `except SomeExceptionType: ...`) and since the matching respects inheritance (that is, `except SomeExceptionType: ...` also handles all subclasses of `SomeExceptionType`) it is beneficial to have a very rich collection of exception types to allow you to be precise in specifying which case you're actually trying to handle.

flocker/node/agents/ebs.py
myechuri: Sorry, my bad. Switched to fully qualified path name, and added comment.

flocker/node/agents/ebs.py
myechuri: Thanks for catching it! Fixed.

myechuri:
```
I get the following output for a partitioned device:

$ sudo lsblk -nb -o SIZE /dev/xvda
171798691840
171792149504
So you should probably only parse the first line.
```
Ah, thats the root disk. Although we just created the volume without partitions in our workflow, getting the first line is definitely better. Thanks for pointing out! Fixed.

myechuri:
```
size and device_size are both in bytes. I would have expected size == device_size to be the reasonable comparison to make here.

If size/(1000**3) == device_size/(1024**3) determines the sizes match, isn't it because either:

this uses integer division so many different sizes are effectively rounded to the same number of GB or GiB (eg, 3 * 10 ** 9 / 2 ** 30 == 3 even though there's a clear unit conversion error there)
or the Amazon volumes are being created with the wrong size which the unit mismatch here exactly masks (looking at http://boto.readthedocs.org/en/latest/ref/ec2.html#boto.ec2.connection.EC2Connection.create_volume I think this might be the case, it says "GiB" but I think the create_volume implementation passes in a value as if it were meant to be GB)
```
Thanks for raising this issue! Here is what is happening:
- Boto and EBS are consistent in size unit: Boto expects GiB, EBS provides and displays size in GiB
- Our create_volume API does the right thing by expecting size in bytes
- Problem is with our `_blockdevicevolume_from_ebs_volume()` which interpret's EBS volume size unit as `GB` instead of `GiB`. Fixed that. Might need similar change for OpenStack as well.

exarkun:
> There is a delete_on_termination attribute associated with a volume. By default, it is set to False, which works for us since we want the volumes to persist across node reboots. This was my reasoning for not overriding default. Please let me know if it doesn't sounds reasonable? Thanks!

Okay, that makes sense.  I had something else in mind but I wasn't clear enough in my question: do volumes remain attached to EC2 instances after the EC2 instance reboots?  If so, do they remain attached to the same device files?

We'll need to find out the answer to this same question for OpenStack, I think (because I don't know the answer for that platform, anyway).

flocker/node/agents/ebs.py
myechuri:
```
:returns: ...
:rytpe: ..
```
seems to be preferred. Changed to the format.
Added docstring for `size`, thanks for catching it!

flocker/node/agents/ebs.py
myechuri: Added a new `Exception` class, then realized it would be difficult to test it. If we do not throw the exception, the subsequent `get_device_path` call will do the right thing since `device_name` label is missing from the EBS volume, indicating unattached device. Hence, removed the exception. Thanks for raising the concern!

myechuri:
> do volumes remain attached to EC2 instances after the EC2 instance reboots?

Yes.

> If so, do they remain attached to the same device files?

Yes.

Validated by manual testing. Would be good to add reboot tests to stress test suite when we get there.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+        False otherwise.
+    """
+    device_name = b"/dev/" + device.encode("ascii")
+
+    # Retrieve size of device as OS sees it using `lsblk`.
+    # Requires util-linux-ng package on CentOS, and
+    # util-linux on Ubuntu.
+    # Required package is installed by default
+    # on Ubuntu 14.04 and CentOS 7.
+    command = [b"/bin/lsblk", b"-nb", b"-o", b"SIZE", device_name]
```
exarkun: `--noheader` would be nicer than `-n` here.  Likewise for `--bytes` and `--output`.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+
+    # Get the base device size, which is the first line in
+    # `lsblk` output. Ignore partition sizes.
+    command_output = check_output(command).split(b'\n')[0]
+    device_size = int(command_output.strip().decode("ascii"))
+
+    return size == device_size
```
exarkun: Yay, nice simple comparison. :smile:

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+    # Retrieve size of device as OS sees it using `lsblk`.
+    # Requires util-linux-ng package on CentOS, and
+    # util-linux on Ubuntu.
+    # Required package is installed by default
+    # on Ubuntu 14.04 and CentOS 7.
+    command = [b"/bin/lsblk", b"-nb", b"-o", b"SIZE", device_name]
+
+    # Get the base device size, which is the first line in
+    # `lsblk` output. Ignore partition sizes.
+    command_output = check_output(command).split(b'\n')[0]
```
exarkun: This could fail if `lsblk` if the given device either doesn't exist or isn't associated with an actual block device - `lsblk` exits with status 1 and `check_output` converts this to an exception.  Probably not critical that we fix this now but please file an issue and link to it from here.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+    Helper function to wait for up to 60s for new
+    EBS block device (`/dev/sd*` or `/dev/xvd*`) to
+    manifest in the OS.
+
+    :param list base: List of baseline block devices
+        that existed before execution of operation that expects
+        to create a new block device.
+    :param int size: Size of the block device we are expected
+        to manifest in the OS.
+
```
exarkun: `time_limit` should have param docs here

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+    """
+    start_time = time.time()
+    elapsed_time = time.time() - start_time
+    while elapsed_time < time_limit:
+        for device in list(set(FilePath(b"/sys/block").children()) -
+                           set(base)):
+            device_name = FilePath.basename(device)
+            if (device_name.startswith((b"sd", b"xvd")) and
+                    _check_blockdevice_size(device_name, size)):
+                new_device = u'/dev/' + device_name
```
exarkun: `device_name` is `bytes` (since it came from `FilePath.basename`) - decode it as `ascii` here, probably.

flocker/node/agents/ebs.py
```diff
@@ -107,6 +115,63 @@ def _wait_for_volume(expected_volume,
...
+    manifest in the OS.
+
+    :param list base: List of baseline block devices
+        that existed before execution of operation that expects
+        to create a new block device.
+    :param int size: Size of the block device we are expected
+        to manifest in the OS.
+
+    :returns: formatted string name of the new block device.
+    :rtype: unicode
```
exarkun: If this times out, `None` is returned.  Probably should document this - or make it an exception if that happens and document what exception gets raised.

flocker/node/agents/ebs.py
```diff
@@ -152,6 +218,62 @@ def compute_instance_id(self):
...
+        Lookup BlockDeviceVolume for given blockdevice_id.
+
+        :param unicode blockdevice_id: ID of a blockdevice that needs lookup.
+
+        :returns BlockDeviceVolume for the given input id.
+        """
+        for volume in self.list_volumes():
+            if volume.blockdevice_id == blockdevice_id:
+                return volume
+        raise UnknownVolume(blockdevice_id)
```
exarkun: You could probably use `flocker.node.agents.blockdevice.get_blockdevice_volume` in place of this if you merge in master now.

flocker/node/agents/ebs.py
exarkun: It might be better not to ever rely on getting the device we ask for.  Just ask for `/dev/sda` every time.  The instance's block device driver will do what it feels like doing and the other code you've already written will figure out what device was actually assigned.  This way we're sure we exercise the "figure out the correct device" code because it runs and must work on every attach operation instead of being fallback code for certain configurations we may not be testing on buildbot.

The caveat, of course, is that maybe AWS will get angry if ask for `/dev/sda` every single time - I haven't looked into this.

flocker/node/agents/ebs.py
```diff
@@ -152,6 +218,62 @@ def compute_instance_id(self):
...
+            instance are currently occupied.
+        """
+        volumes = self.connection.get_all_volumes()
+        devices = [v.attach_data.device for v in volumes
+                   if v.attach_data.instance_id == instance_id]
+        for suffix in b"fghijklmonp":
+            file_name = u'/dev/sd' + suffix
+            if file_name not in devices:
+                return file_name
+        return None
```
exarkun: @wallrj pointed out that a side-effect of this approach is that it counts how many devices we're actually allowed to attach to an EC2 instance.  There's definitely an interesting failure case related to this but `IBlockDeviceAPI.attach_volume` doesn't currently specify what's supposed to happen in that case.  Can you file an issue for fixing the interface so that it specifies what exception should be raised when no more devices can be attached?  And probably link to that issue from the EBS `attach_volume` implementation.

flocker/node/agents/ebs.py
```diff
@@ -195,11 +317,74 @@ def resize_volume(self, blockdevice_id, size):
...
+        """
+        volume = self._get(blockdevice_id)
+        if volume.attached_to is not None:
+            raise AlreadyAttachedVolume(blockdevice_id)
+
+        with self.lock:
+            # begin lock scope
+
+            blockdevices = FilePath(b"/sys/block").children()
+            device = self._next_device(attach_to)
```
exarkun: Related to the comment in `_next_device`, the current failure mode for "no more devices can be attached" seems to be that `None` is returned and then passed in to the `self.connection.attach_volume` call (one line below this) which I'm guessing is an error because the AWS API docs say the device is a required argument?

Since the interface doesn't specify what to do in this error case I don't think anything needs to be fixed here but maybe this is a good place to link to the issue about making the interface specify what to do.

flocker/node/agents/ebs.py
```diff
@@ -160,7 +282,7 @@ def create_volume(self, dataset_id, size):
         Open issues: https://clusterhq.atlassian.net/browse/FLOC-1792
         """
         requested_volume = self.connection.create_volume(
-            size=int(Byte(size).to_GB().value), zone=self.zone)
+            size=int(Byte(size).to_GiB().value), zone=self.zone)
```
exarkun: Truncating here will under-allocate volumes and probably also cause the convergence algorithm to never converge.  I think the same is true of the OpenStack implementation of this method.  It doesn't explicitly truncate to int but I strongly suspect the Cinder library we're using does it for us (because the Cinder HTTP API documentation says size must be an integer).

So in both cases if someone asks for a number of bytes that doesn't divide evenly into the unit supported by the backend, they'll get a slightly too-small volume and the convergence agent will then infinitely loop trying to resize it to be the right size (which of course will fail since no one has implemented resize yet :smile: ).

FLOC-1874 seems particularly relevant here.  So maybe just add a link to it here.

exarkun: Oh, and maybe add a link to that issue in the corresponding Cinder code too, if you don't mind.

flocker/node/agents/ebs.py
```diff
@@ -152,6 +218,62 @@ def compute_instance_id(self):
...
+    def _get_ebs_volume(self, blockdevice_id):
+        """
+        Lookup EBS Volume information for a given blockdevice_id.
+
+        :param unicode blockdevice_id: ID of a blockdevice that needs lookup.
+
+        :returns boto.ec2.volume.Volume for the input id. ``None`` if
+            no boto.ec2.volume.Volume was found for the given id.
+        """
+        for volume in self.connection.get_all_volumes():
```
exarkun: A minor optimization here might be `get_all_volumes(volume_ids=[blockdevice_id])`.

:question: (optional to address since performance is not our main concern and there's no benchmark demonstrating this actually helps)

flocker/node/agents/ebs.py
```diff
@@ -224,4 +409,24 @@ def destroy_volume(self, blockdevice_id):
...
+            not attached to a host.
+        """
+        volume = self._get(blockdevice_id)
+        if volume.attached_to is None:
+            raise UnattachedVolume(blockdevice_id)
+
+        ebs_volume = self._get_ebs_volume(blockdevice_id)
+        device = ebs_volume.tags[ATTACHED_DEVICE_LABEL]
+        if device is None:
+            raise UnattachedVolume(blockdevice_id)
```
exarkun: There's already a check for the volume being unattached above.  So this case never gets executed by the test suite.  I think it may not work, either, because `ebs_volume.tags[ATTACHED_DEVICE_LABEL]` probably raises a `KeyError` if that tag doesn't exist (eg because it was deleted by an earlier call to `detach_volume`).  I don't think this case is impossible - someone could modify the tags in between the `self._get(...)` call and the `self._get_ebs_volume(...)` call - but our test suite doesn't exercise it.

If you can think of a way to test this off the top of your head, add a test that demonstrates this works (or that demonstrates it needs to handle `KeyError` and then fix the implementation).  If it's hard to test, I suggest removing this second bit of code handling the unattached case (only this second bit, the first bit above is good) and filing an issue for addressing the more obscure failure case later (and link to it from here).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1168,6 +1168,22 @@ class IBlockDeviceAPITestsMixin(object):
...
+        :param int size: Expected size of input volume.
+        """
+        volume_size = volume.size
+        device_path = self.api.get_device_path(volume.blockdevice_id).path
+        device = device_path.encode("ascii")
+
+        command = [b"/bin/lsblk", b"-nb", b"-o", b"SIZE", device]
+        command_output = check_output(command).split(b'\n')[0]
+        device_size = int(command_output.strip().decode("ascii"))
+        self.assertEquals(volume_size, device_size)
```
exarkun: `TestCase` provides like a million aliases for all of its methods.  `assertEqual` is the preferred spelling of this one, though (and I think some of the alternate spellings are even deprecated).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1217,9 +1233,12 @@ def test_listed_volume_attributes(self):
...
         )
         [listed_volume] = self.api.list_volumes()
+        volume_size_GiB = int(Byte(listed_volume.size).to_GiB().value)
+        create_size_GiB = int(Byte(REALISTIC_BLOCKDEVICE_SIZE).to_GiB().value)
+
         self.assertEqual(
-            (expected_dataset_id, REALISTIC_BLOCKDEVICE_SIZE),
-            (listed_volume.dataset_id, listed_volume.size)
+            (expected_dataset_id, create_size_GiB),
+            (listed_volume.dataset_id, volume_size_GiB)
```
exarkun: A link to FLOC-1874 would probably be appropriate here.  I'd really like these tests to make a more precise assertion than one truncated down to GiB.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1232,9 +1251,12 @@ def test_created_volume_attributes(self):
             dataset_id=expected_dataset_id,
             size=REALISTIC_BLOCKDEVICE_SIZE
         )
+
+        volume_size_GiB = int(Byte(new_volume.size).to_GiB().value)
+        create_size_GiB = int(Byte(REALISTIC_BLOCKDEVICE_SIZE).to_GiB().value)
```
exarkun: And here

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1249,6 +1271,23 @@ def test_attach_unknown_volume(self):
...
+
+        new_volume = self.api.create_volume(
+            dataset_id=dataset_id,
+            size=REALISTIC_BLOCKDEVICE_SIZE
+        )
+        attached_volume = self.api.attach_volume(
+            new_volume.blockdevice_id, attach_to=self.this_node,
+        )
+
+        self._assert_volume_size(attached_volume, REALISTIC_BLOCKDEVICE_SIZE)
```
exarkun: Since REALISTIC_BLOCKDEVICE_SIZE isn't an even multiple of 1GiB, how does this pass on EC2?

flocker/node/agents/ebs.py
myechuri: Fixed in both ebs.py and test_blockdevice.py.

flocker/node/agents/ebs.py
myechuri: We only get to running `lsblk` in `_check_blockdevice_size()` after we noticed a new block device in `/sys/block` in `_wait_for_new_device()`. So, both error conditions listed above cannot be reached. 
There are other possible error cases, like `lsblk` not being available on the system, etc, will file a JIRA issue for that (https://clusterhq.atlassian.net/browse/FLOC-1886) and referred in comment.

flocker/node/agents/ebs.py
myechuri: Sorry, missed it. Fixed.

flocker/node/agents/ebs.py
exarkun: I think there's still a tiny race condition where you find a device in `/sys/block` and then it disappears from the system before the code gets around to running `lsblk` (only a millisecond or less later, probably, but run the code long enough and it'll probably eventually happen).  Definitely not a high priority bug.

exarkun:
Thanks!  This looks great.  Please go ahead and merge after addressing this latest round of feedback. :smile:

flocker/node/agents/ebs.py
myechuri: Thanks! Fixed.

flocker/node/agents/ebs.py
myechuri: i wrote up a new Exception yesterday, then realised that there is not much value in it. If `new_device` is `None`, the EBS volume's attachment tag will be empty, indicating the volume as unattached. Which phase of attach failed isn't of much use to subsequent `get_device_path` calls. So, undid the new Exception.

Added more comments and made `new_device is not None` check more explicit in the tagging phase. Thanks!

flocker/node/agents/ebs.py
myechuri: Thanks! Fixed.

flocker/node/agents/ebs.py
myechuri: Valid! Filed https://clusterhq.atlassian.net/browse/FLOC-1887 and linked in comment and linked to FLOC-1813. Thanks!

flocker/node/agents/ebs.py
myechuri: Changed to:

            device = self._next_device(attach_to)

            if device is None:
                # TODO: Handle lack of free devices in ``/dev/sd[f-p]`` range
                # (see https://clusterhq.atlassian.net/browse/FLOC-1887).
                # No point in attempting an ``attach_volume``, so, return.
                return

            self.connection.attach_volume(blockdevice_id, attach_to, device)

flocker/node/agents/ebs.py
myechuri: Added link to FLOC-1874 for EBS and Cinder. Added a note to FLOC-1874 as well. Thanks!

flocker/node/agents/ebs.py
adamtheturtle: The code is full of "XXX" and not "TODO" for issues. This means that I can search for "TODO" while I use it in my branch. Could we stick to that? (It isn't a policy but a personal preference of mine).

flocker/node/agents/ebs.py
myechuri: Thanks! Fixed. Definitely better to supply volume_ids.

flocker/node/agents/ebs.py
myechuri: Changed TODO in flocker/node/agents/ to XXX. Thanks.

flocker/node/agents/ebs.py
myechuri: Good catch! Added test to demonstrate need to handle KeyError:
[ERROR]
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/flocker/node/agents/functional/test_ebs.py", line 133, in test_attached_volume_missing_device_tag
    device = self.api.get_device_path(volume.blockdevice_id)
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/flocker/node/agents/ebs.py", line 442, in get_device_path
    device = ebs_volume.tags[ATTACHED_DEVICE_LABEL]
exceptions.KeyError: u'attached-device-name'

flocker.node.agents.functional.test_ebs.EBSBlockDeviceAPIInterfaceTests.test_attached_volume_missing_device_tag

flocker/node/agents/test/test_blockdevice.py
myechuri: Thanks! Fixed.

flocker/node/agents/test/test_blockdevice.py
myechuri: Added a note referring to FLOC-1874. Agree that GiB comparison is not ideal: tried KiB, MiB as well, but there was still a mismatch. Since realistic device sizes are mostly going to be in GB/GiB, error in test precision might be worth tolerating until FLOC-1874 in favor of value gained from these tests?

flocker/node/agents/test/test_blockdevice.py
myechuri: Added comment. Thanks.

flocker/node/agents/test/test_blockdevice.py
myechuri: Thanks for pointing this out! 
There are 2 kinds of comparisons i wanted to do in this test:
1) Device size reported by `lsblk` matches volume size reported in BlockDeviceVolume.
2) Device size reported by `lsblk` matches input volume size at create_volume time.

(1) was implemented and the test passed. i missed doing (2). Added that part, along with FLOC-1874 disclaimer. Thanks for catching!

wallrj:
I went searching to find out how Amazon Linux does the symlink-to-requested device path trick. I didn't find an answer to that, but I did find this article which may suggest that cloud-init on Ubuntu-12.04 on Eucalyptus also does that trick:
 * https://eucalyptus.atlassian.net/browse/EUCA-9251

And among the comments there's an xml file `/var/run/libvirt/qemu/i-7c7fd842.xml ` which seems to have been taken from inside an Eucalyptus guest machine and which contains the real and requested device paths.

I don't know if AWS also uses Qemu? But thought it might be of interest in case the "new device of expected size" detection turns out not to be accurate enough.

There's some discussion here about how the sysfs properties (including serial number) of a virtual block device might be set from the qemu command line:
 * comments.gmane.org/gmane.comp.emulators.qemu/64827

...but can't tell if that's ever been implemented.

myechuri:
Thanks for the interesting links, @wallrj !

> I don't know if AWS also uses Qemu? But thought it might be of interest in case the "new device of expected size" detection turns out not to be accurate enough.

Depends on the kind of provisioned EC2 instance: AWS gives a choice of HVM and PV. Within PV spectum, degree of virtualization cannot be specified at instance provision time. So, Qemu might/might not be used based on the instance type.

PR #1394, opened 2015-05-11
[FLOC-1848] Remove resize dataset from API

dwgebler:
Fixes:
 * https://clusterhq.atlassian.net/browse/FLOC-1848

Removes the maximum_size change capabilities from the update dataset API endpoint.

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -271,3 +271,19 @@ definitions:
       maximum_size:
         '$ref': '#/definitions/maximum_size'
     additionalProperties: false
+
+  dataset_configuration_update:
+    title: "Dataset Configuration"
+    description: "The configuration for a particular dataset."
+    type: object
+    properties:
+      primary:
+        '$ref': '#/definitions/primary'
+      dataset_id:
+        '$ref': '#/definitions/dataset_id'
+      deleted:
+        '$ref': '#/definitions/deleted'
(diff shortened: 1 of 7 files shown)


wallrj:
Thanks @dwgebler 

Not a full review, but:

 * I was going to say that it'd be nice to not duplicate the properties in the types.yml file, but I know that it's difficult to merge / inherit json schemas and have them rendered nicely....so leave that.
 * There are some remaining (failing) acceptance tests that need to be deleted though.

http://build.clusterhq.com/builders/flocker%2Facceptance%2Frackspace%2Ffedora-20/builds/1318/steps/trial/logs/stdio

Please remove / fix those and resubmit.

flocker/control/schema/types.yml
```diff
@@ -271,3 +271,19 @@ definitions:
       maximum_size:
         '$ref': '#/definitions/maximum_size'
     additionalProperties: false
+
+  dataset_configuration_update:
```
itamarst: This schema allows things the Python method doesn't actually accept, i.e. deleted, dataste_id, metadata.

itamarst: (dataset_id comes out of the URL).

itamarst:
Looks good, address and merge. And maybe file issue for resurrecting this code, with a link to the check-in so it's easy to revert.

PR #1294, opened 2015-04-08
[FLOC-1675] Switch run-acceptance-tsts to use spawnProcess.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1675

PR #1395, opened 2015-05-11
[FLOC-1849] Show stderr from admin.runner too.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1849

This doesn't address the underlying issue (which is we don't interact with vagrant and virtualbox in a way that cleans up old images properly). But it does address the regression in #1294, where we stopped showing stderr from vagrant.

(See 
http://build.clusterhq.com/builders/flocker%2Facceptance%2Fvagrant%2Ffedora-20/builds/2059
where it demonstrates the output).

### DIFF
--- a/admin/runner.py
+++ b/admin/runner.py
@@ -49,7 +49,7 @@ class CommandProtocol(ProcessProtocol, object):
         Otherwise, errbacks with the reason.
     :ivar file-like output: For logging.
 
-    :ivar defaultdict _fds: Mapping from fds to `_LineParsers`.
+    :ivar defaultdict _fds: Mapping from file descriptors to `_LineParsers`.
     """
     def __init__(self):
         self._fds = defaultdict(lambda: _LineParser(output=self.output))
--- a/admin/runner.py
+++ b/admin/runner.py
@@ -1,25 +1,46 @@
 # Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# -*- test-case-name: admin.test.test_runner -*-
 """
 Tools for running commands.
 """
 import sys
 import os
+from collections import defaultdict
 
 from characteristic import attributes
 
-from twisted.internet.error import ConnectionDone
-from twisted.internet.endpoints import ProcessEndpoint, connectProtocol
+from twisted.internet.error import ProcessDone
 from twisted.internet.defer import Deferred
+from twisted.internet.protocol import ProcessProtocol
 
 from twisted.protocols.basic import LineOnlyReceiver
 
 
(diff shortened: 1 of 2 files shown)


admin/runner.py
```diff
@@ -27,21 +48,21 @@ class CommandProtocol(LineOnlyReceiver, object):
...
+        self._fds = defaultdict(lambda: _LineParser(output=self.output))
 
-    def connectionMade(self):
-        self.transport.disconnecting = False
+    def childDataReceived(self, childFD, data):
+        self._fds[childFD].dataReceived(data)
 
-    def connectionLost(self, reason):
-        if reason.check(ConnectionDone):
+    def processEnded(self, reason):
```
tomprince: This should probably actually be `processExited`.

admin/test/test_runner.py
```diff
@@ -105,6 +105,20 @@ def test_unregisters_killer_failure(self):
...
+        the reason.
+        """
+
+        reactor = ProcessCoreReactor()
+        d = run(reactor, ['command', 'and', 'args'])
+        [process] = reactor.processes
+
+        expected_failure = Failure(ProcessDone(0))
+        process.processProtocol.processEnded(expected_failure)
+        self.successResultOf(d),
```
adamtheturtle: This should be compared to the reason.

admin/runner.py
```diff
@@ -27,21 +48,21 @@ class CommandProtocol(LineOnlyReceiver, object):
         If the command finished successfully, will fire with ``None``.
         Otherwise, errbacks with the reason.
     :ivar file-like output: For logging.
+
+    :ivar defaultdict _fds: Mapping from fds to `_LineParsers`.
```
adamtheturtle: :question: It seems to be more standard to write "file descriptors" - at least it'd be easier for me to read.

admin/test/test_runner.py
adamtheturtle: Actually, the reason for success? That means the docstring should be fixed and this comma should be removed.

admin/runner.py
adamtheturtle: Whichever it ends up as can you comment please - I have read the definitions of both and am not clear on why it would be either.

adamtheturtle:
This appears to work and doesn't seem to break anything. I've looked into some of the Twisted things used but I don't completely understand it. I've asked for some comments but I don't think that it is top priority that this is the shiniest code so merge when you and Buildbot are happy, or resubmit if there are any major changes.

PR #1396, opened 2015-05-12
[FLOC-1766] Skip tests in setup if flocker-ca not installed

dwgebler:
Fixes FLOC-1766

### DIFF
--- a/flocker/ca/functional/test_ca.py
+++ b/flocker/ca/functional/test_ca.py
@@ -75,6 +75,7 @@ class FlockerCATests(make_script_tests(EXECUTABLE)):
     """
     Tests for ``flocker-ca`` script.
     """
+    @requireCA
     def setUp(self):
         """
         Create a root certificate for the test.
--- a/flocker/ca/functional/test_ca.py
+++ b/flocker/ca/functional/test_ca.py
@@ -5,8 +5,8 @@ Functional tests for ``flocker-ca`` CLI.
 """
 
 import re
-
 from subprocess import CalledProcessError
+from unittest import skipUnless
 
 from twisted.python.procutils import which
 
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks Dave.

The tests have their own decorators which can now be removed.

It is not changed in this branch so this is optional (but I think it is nicer to improve code as we go) but the decorator itself is more complex than others in our code.

```python
def requireCA(test):
    """
    Simple test decorator to check if both flocker-ca and OpenSSL are
    installed and skip if either isn't.
    """
    def inner(testcase, *args, **kwargs):
        if not which(EXECUTABLE):
            return testcase.skipTest(EXECUTABLE + " not installed")
        if not which(b"openssl"):
            return testcase.skipTest("openssl not installed")
        return test(testcase, *args, **kwargs)
    return inner
```

vs something like (untested):

```python
from unittest import skipUnless

require_flocker_ca = skipUnless(which(EXECUTABLE), "message contents")
require_openssl = skipUnless(which(b"openssl"), "Message contents")

@require_flocker_ca
@require_openssl
def setUp(self)...
```

Please address and merge after you are happy that:
* The tests are not failing anywhere on BuildBot (force the dev box builder because IIRC it is not run on branches automatically but it was failing these tests)
* The tests are passing everywhere they should - that is, no builder which this should be tested on is skipping the tests

Or resubmit if there are any changes not mentioned above.

adamtheturtle:
> Please address and merge after you are happy that:
* The tests are passing everywhere they should - that is, no builder which this should be tested on is skipping the tests

I say this partially because the build hasn't finished yet but also because I'm not certain which builders need to run / pass these tests. Perhaps it would be useful to list them here.

PR #1397, opened 2015-05-12
[FLOC-1379] Lots more useful logging in the convergence loop

itamarst:


tomprince:
The changes here look good. Merge once buildbot is happy.

PR #1398, opened 2015-05-12
[FLOC-1269] Expire out-of-date cluster state from the control service

itamarst:
Still haven't manually tested this, but seems like it's in decent state for a review at least.


flocker/control/_clusterstate.py
```diff
@@ -20,9 +23,32 @@ class ClusterStateService(Service):
...
+        self._clock = reactor
+
+    def _wipe_expired(self):
+        """
+        Clear any expired state from memory.
+        """
+        current_time = self._clock.seconds()
+        evolver = self._information_wipers.evolver()
+        for key, (wiper, timestamp) in self._information_wipers.items():
+            if current_time - 10 >= timestamp:
```
exarkun: Make a symbolic constant for `10`

exarkun: Also, design review asked for a follow-up ticket for doing something smarter than hard-coding ten seconds.

exarkun: And for follow-up ticket for using an ordered sequence instead of a mapping.

flocker/control/_model.py
```diff
@@ -625,6 +625,47 @@ def update_cluster_state(cluster_state):
...
+        """
+        Return a key describing what information will be wiped.
+
+        Providers that wipe the same information should return the same
+        key, and providers that wipe different information should return
+        differing keys.
+
+        Different ``IClusterStateWipe`` implementors are presumed to
+        cover different information, so there is no need for the key to
+        express that differentation.
```
exarkun: Implied here is that the type of the object that provides the interface is part of the interface.  Probably worth stating this explicitly.

flocker/control/_clusterstate.py
itamarst: Ordered sequence is problematic given you need to replace stuff by key when an update happens.

itamarst: Sorry about the other issue, will do.

flocker/control/_model.py
```diff
@@ -689,6 +730,57 @@ def __new__(cls, **kwargs):
     def update_cluster_state(self, cluster_state):
         return cluster_state.update_node(self)
 
+    # Attributes that may be set to None to indicate ignorance:
+    _POTENTIALLY_IGNORANT_ATTRIBUTES = ["used_ports", "applications",
+                                        "manifestations", "paths",
+                                        "devices"]
```
exarkun: Too bad these can't be actual field instances.  But it seems like field instances are really inconvenient to use as anything exception descriptors.

flocker/control/_model.py
```diff
@@ -689,6 +730,57 @@ def __new__(cls, **kwargs):
...
+
+    def _completely_ignorant(self):
+        """
+        Return whether the node is completely ignorant of anything.
+        """
+        return all(getattr(self, attr) is None
+                   for attr in self._POTENTIALLY_IGNORANT_ATTRIBUTES)
```
exarkun: Only one caller and it negates the result.  Invert this and the one usage is easier to read.

:question:

flocker/control/_model.py
```diff
@@ -689,6 +730,57 @@ def __new__(cls, **kwargs):
...
+
+    def get_information_wipe(self):
+        attributes = [attr for attr in
```
exarkun: Docstring

flocker/control/_model.py
```diff
@@ -689,6 +730,57 @@ def __new__(cls, **kwargs):
...
+
+    :ivar UUID node_uuid: The UUID of the node being wiped.
+    :ivar PSet attributes: Names of ``NodeState`` attributes to wipe.
+    """
+    node_uuid = field(mandatory=True, type=UUID)
+    attributes = pset_field(str)
+
+    def update_cluster_state(self, cluster_state):
+        nodes = {n for n in cluster_state.nodes
+                 if n.uuid == self.node_uuid}
```
exarkun: Hm too bad `cluster_state.get_node` doesn't support this.  File a follow-up issue for making it accept a default to return instead of creating one?

exarkun: Although on second thought, I'm not entirely sure in which case this code would fail to find a node.

flocker/control/test/test_clusterstate.py
```diff
@@ -136,3 +140,43 @@ def test_manifestation_path(self):
...
+        service.apply_changes([app_node])
+        self.clock.advance(1)
+        service.apply_changes([app_node_2])
+        self.clock.advance(9)
+        ten_second_state = service.as_deployment()
+        self.clock.advance(1)
+        eleven_second_state = service.as_deployment()
+        self.assertEqual(
+            [ten_second_state, eleven_second_state],
+            [DeploymentState(nodes=[app_node_2]), DeploymentState()])
```
exarkun: How about a test for some state being overwritten by a later update and then not being wiped prematurely?  eg, `applications=[APP1]` at t=0, `applications=[APP2]` at t=1, check state at t=10

flocker/control/test/test_model.py
```diff
@@ -1187,3 +1188,132 @@ def test_both(self):
...
+        """
+        self.assertNotEqual(self.APP_WIPE.key(), self.DATASET_WIPE.key())
+
+    def test_key_same_by_attribute_contents(self):
+        """
+        The ``IClusterStateWipe`` has the same key if it is wiping same
+        attributes on same node.
+        """
+        different_apps_node = self.NODE_FROM_APP_AGENT.set(
+            "applications", {APP2}).set("used_ports", {4, 5})
```
exarkun: There's also `.set(applications={APP2}, used_ports={4, 5})` fwiw

:question:

exarkun:
Thanks.  Looks good.  Address inline comments and then merge.

PR #1401, opened 2015-05-13
[FLOC-1878] Allow PMaps to be serialized from/to JSON even if their keys aren't strings

itamarst:


tomprince:
This looks good, and I've manually tested it in another branch.

PR #1402, opened 2015-05-13
Add dev env documentation for Ubuntu. Change default Vagrant to CentOS.

myechuri:
[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1402)

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -51,7 +51,7 @@ Development Environment
 =======================
 
 * To run the complete test suite you will need `ZFS`_ and `Docker`_ installed.
-  The recommended way to get an environment with these installed is to use the included ``Vagrantfile`` which will create a pre-configured Fedora 20 virtual machine.
+  The recommended way to get an environment with these installed is to use the included ``Vagrantfile`` which will create a pre-configured CentOS 7 virtual machine.
   Vagrant 1.6.2 or later is required.
   Once you have Vagrant installed (see the `Vagrant documentation <http://docs.vagrantup.com/>`_) you can run the following to get going:
 
@@ -59,6 +59,13 @@ Development Environment
 
      $ vagrant up
      $ vagrant ssh
+
+* If you prefer developing on Ubuntu, you will need libssql-dev and libffi-dev packages to run tests. If missing these packages, please install by running:
+
+  .. code-block:: console
+
+     $ sudo apt-get install libssl-dev
+     $ sudo apt-get install libffi-dev
 
 * You will need Python 2.7 and a recent version of PyPy installed on your development machine.
 * If you don't already have ``tox`` on your development machine, you can install it and other development dependencies (ideally in a ``virtualenv``) by doing:
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -60,7 +60,7 @@ Development Environment
      $ vagrant up
      $ vagrant ssh
 
-* If you prefer developing on Ubuntu, you will need libssql-dev and libffi-dev packages to run tests. If missing these packages, please install by running:
+* If you prefer developing on Ubuntu, you will need libssl-dev and libffi-dev packages to run tests. If missing these packages, please install by running:
 
   .. code-block:: console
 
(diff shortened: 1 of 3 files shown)


docs/spelling_wordlist.txt
```diff
@@ -18,6 +18,7 @@ dataset
 datasets
 Debian
 Deis
+dev
```
adamtheturtle: As per http://sphinxcontrib-spelling.readthedocs.org/en/latest/customize.html this should be "a list of words known to be spelled correctly but that do not appear in the language dictionary". Of course, we don't have to stick to that (and you can see in this file that we haven't always) but I prefer to wrap the offending "libssl-dev" in double backquotes, as it is kinda-sorta a code sample.

docs/spelling_wordlist.txt
```diff
@@ -41,6 +42,8 @@ installable
 internet
 invalidations
 Kubernetes
+libffi
```
adamtheturtle: Ditto

CONTRIBUTING.rst
```diff
@@ -60,6 +60,13 @@ Development Environment
      $ vagrant up
      $ vagrant ssh
 
+* If you prefer developing on Ubuntu, you will need libssl-dev and libffi-dev packages to run tests. If missing these packages, please install by running:
+
+  .. code-block:: console
```
adamtheturtle: switch to `.. prompt:: bash $` and then the commands can be copied but their prompts won't be.

CONTRIBUTING.rst
```diff
@@ -60,6 +60,13 @@ Development Environment
      $ vagrant up
      $ vagrant ssh
 
+* If you prefer developing on Ubuntu, you will need libssl-dev and libffi-dev packages to run tests. If missing these packages, please install by running:
```
adamtheturtle: We should probably have a style guide (starting with https://clusterhq.atlassian.net/browse/FLOC-1216?) because we have these half-rules which have built up but aren't documented and aren't always kept to. If you look at http://rhodesmill.org/brandon/2012/one-sentence-per-line/ as an inspiration it might be helpful - we tend to use one sentence per line for some of the reasons described there, such as easier diff-ing.

CONTRIBUTING.rst
```diff
@@ -51,7 +51,7 @@ Development Environment
 =======================
```
adamtheturtle: The way I read this section is:

1. You need ZFS and Docker to run *all* the tests.
2. A quick way to do this is Vagrant.
3. On Ubuntu you also need these other packages.
4. You need these other things

There are two things confusing about this:
* Are the things in (4) - Python and `tox` Ubuntu only? Are they to be installed on the development machine even if using Vagrant?
* The first section is basically saying "use any platform including CentOS in Vagrant" then the second one is saying "this is for Ubuntu only" so I think I could use CentOS just fine without extra packages.

These can be solved by having a clear introduction of "The way to get an environment with all dependencies is to use Vagrant, to do this follow XXX", then a new heading for developing locally.

And relatedly:
* Do we need PyPy really? Buildbot doesn't have a PyPy builder (not new in this PR).
* To run all the tests we need more than just the packages mentioned here - ZFS, Docker, `libssl-dev`, `libffi-dev`, right? On CentOS we have all the packages in `flocker/vagrant/dev/bootstrap.sh` (though some of them are just for the release process) like `rpmdevtools`. I think that it is best to mention that some tests cannot be run on Ubuntu with this setup.

Perhaps some of the above is overly pedantic and we can assume a higher level of awareness from developers.

adamtheturtle:
Thanks @myechuri - this is certainly a section which is worth improving, and explaining how we actually do things (instead of the quickest way to get set up - Vagrant and `tox`) is really useful. Please resubmit when you have addressed the comments.

CONTRIBUTING.rst
adamtheturtle: Though actually this particular document might be a weird one - maybe we care about it being viewed in GitHub and `.. prompt` may be worse there.

jml:

wallrj:
This has been mostly addressed  in a separate PR: #2730

PR #1403, opened 2015-05-14
Backend section floc 1881

adamtheturtle:
This adds a few more lines to copy/paste for the user but adds no functionality.
It was split out of FLOC-1791.
There is some duplication but the agents (and their tests) will be merged in FLOC-1791.

### DIFF
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -83,14 +83,11 @@ def _get_external_ip(host, port):
 
 def validate_configuration(configuration):
     """
-    # TODO change
+    Validate a provided configuration.
 
-    Extract configuration from provided options.
+    :param dict configuration: A desired configuration for an agent.
 
-    :param FilePath path: Path to a file containing specified options for an
-        agent.
-
-    :return dict: Dictionary containing the desired configuration.
+    :raises: jsonschema.ValidationError if the configuration is invalid.
     """
     schema = {
         "$schema": "http://json-schema.org/draft-04/schema#",
@@ -166,6 +163,7 @@ class ZFSAgentScript(object):
                 "Configuration file does not exist at '{}'.".format(
                     agent_config.path))
 
+        # TODO test that this is called
         validate_configuration(configuration=configuration)
 
         host = configuration['control-service']['hostname']
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -11,8 +11,7 @@ from os import getpid
 
 import yaml
 
-from jsonschema import FormatChecker, validate
-from jsonschema.exceptions import ValidationError
+from jsonschema import FormatChecker, Draft4Validator
 
 from pyrsistent import PRecord, field
 
@@ -82,8 +81,10 @@ def _get_external_ip(host, port):
         sock.close()
 
 
-def agent_config_from_file(path):
+def validate_configuration(configuration):
     """
+    # TODO change
+
     Extract configuration from provided options.
 
     :param FilePath path: Path to a file containing specified options for an
@@ -91,18 +92,16 @@ def agent_config_from_file(path):
 
     :return dict: Dictionary containing the desired configuration.
     """
-    try:
-        options = yaml.safe_load(path.getContent())
-    except IOError:
-        raise ConfigurationError(
-            "Configuration file does not exist at '{}'.".format(path.path))
-
     schema = {
         "$schema": "http://json-schema.org/draft-04/schema#",
         "type": "object",
-        "required": ["version", "control-service"],
+        "required": ["version", "control-service", "dataset"],
         "properties": {
-            "version": {"type": "number"},
+            "version": {
+                "type": "number",
+                "maximum": 1,
+                "minimum": 1,
+            },
             "control-service": {
                 "type": "object",
                 "required": ["hostname"],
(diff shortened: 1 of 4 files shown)


flocker/node/test/test_script.py
```diff
@@ -304,180 +469,170 @@ def test_custom_config_file(self):
...
+        """
+        No exception is raised when validating a valid configuration with a
+        ZFS backend.
         """
-        if configuration is not None:
-            self.config.setContent(yaml.safe_dump(configuration))
+        self.configuration['dataset'] = {
+            u"backend": u"loopback",
+            u"loopback-pool": u"custom-pool",
+        }
```
tomprince: :question: I'm not sure if this style of testing is valuable. You need to look in too different places to figure out what each test is testing. It would perhaps make sense to refactor this to use `build_schema_test` from `flocker.restapi.testools` here.

flocker/node/test/test_script.py
```diff
@@ -168,6 +309,30 @@ def deployer_factory(node_uuid, hostname):
...
+        """
+        ``AgentServiceFactory.get_service`` raises a ``ConfigurationError`` if
+        the given configuration file does not exist.
+        """
+        deployer = object()
+        reactor = MemoryCoreReactor()
+        options = DatasetAgentOptions()
+        options.parseOptions([b"--agent-config", self.non_existent_file.path])
+
+        def factory(**kw):
```
tomprince: This function seems used enough that it worth factoring out.

flocker/node/script.py
```diff
@@ -113,32 +109,42 @@ def agent_config_from_file(path):
...
+                "type": "object",
+                "oneOf": [
+                    {
+                        "required": ["backend"],
+                        "properties": {
+                            "backend": {
+                                "type": "string",
+                                "pattern": "zfs",
+                            },
+                            "zfs-pool": {
```
tomprince: I don't think we need to prefix options here. It doesn't make sense for a configuration file to have configuration for multiple backends at once in it.

tomprince:
Two minor points, and then merge.

flocker/node/script.py
adamtheturtle: They can't - it is "oneOf" - that reminds me that there is a missing test. I did it like this to match the `zfs-pool: flocker` in the accepted design in https://github.com/ClusterHQ/flocker/pull/1338.

flocker/node/test/test_script.py
adamtheturtle: I'll try it and then decide what I prefer if that's OK. `build_schema_test` seems more suited to schemas which are one and maybe two lines long, it might get unwieldy.

flocker/node/test/test_script.py
```diff
@@ -304,180 +469,170 @@ def test_custom_config_file(self):
...
 
-        :param Exception exception: The exception type which
-            :func:`agent_config_from_file` should fail with.
-        :param dict configuration: The contents of the agent configuration
-            file. If ``None`` then the file will not exist.
-        :param bytes message: The expected exception message.
+    def test_valid_loopback_configuration(self):
+        """
+        No exception is raised when validating a valid configuration with a
+        ZFS backend.
```
adamtheturtle: Incorrect docstring.

flocker/node/script.py
adamtheturtle: (no missing test, dictionaries can't have two identical keys).

flocker/node/test/test_script.py
adamtheturtle: I tried this but it was just one long test - the current schema isn't set up to work the way `build_schema_test` works so we started testing a list of valid and invalid configs, and then got to the problem where you see only one error if the test fails and now think that the trade-offs are not worth it, at least now while people are waiting on this story.

flocker/node/script.py
```diff
@@ -148,9 +154,18 @@ class ZFSAgentScript(object):
...
-        configuration = agent_config_from_file(path=options[u'agent-config'])
+        try:
+            agent_config = options[u'agent-config']
+            configuration = yaml.safe_load(agent_config.getContent())
+        except IOError:
+            raise ConfigurationError(
+                "Configuration file does not exist at '{}'.".format(
+                    agent_config.path))
+
+        validate_configuration(configuration=configuration)
```
exarkun: It would be good if all of this code were still in a separate function for loading and validating the configuration.  As an extra bonus, that function could also serve as a central point to supply defaults (rather than forcing all code using the configuration to supply per-item defaults in an ad hoc way).  Please file an issue and link to it from here.

flocker/node/script.py
```diff
@@ -259,9 +274,18 @@ def get_service(self, reactor, options):
...
         :return: The ``AgentLoopService`` instance.
         """
-        configuration = agent_config_from_file(path=options[u'agent-config'])
+        try:
+            agent_config = options[u'agent-config']
+            configuration = yaml.safe_load(agent_config.getContent())
+        except IOError:
+            raise ConfigurationError(
+                "Configuration file does not exist at '{}'.".format(
+                    agent_config.path))
```
exarkun: This is an anti-pattern: handle an exception with multiple possible causes and convert it into a more general exception that doesn't differentiate between those causes (with a twist: claim it is one specific cause without knowing for sure).

Please file an issue for fixing this and link to it from here.

flocker/node/script.py
```diff
@@ -148,9 +154,18 @@ class ZFSAgentScript(object):
...
     """
     def main(self, reactor, options, volume_service):
-        configuration = agent_config_from_file(path=options[u'agent-config'])
+        try:
+            agent_config = options[u'agent-config']
+            configuration = yaml.safe_load(agent_config.getContent())
+        except IOError:
+            raise ConfigurationError(
+                "Configuration file does not exist at '{}'.".format(
+                    agent_config.path))
```
exarkun: Same comment about exception-handling anti-pattern here.

Also, this looks section looks like an exact copy of the code from `AgentServiceFactory`.  Another reason to put it in a function, sounds like.  Please make a note of this in that issue and link to it from here as well.

flocker/node/script.py
adamtheturtle: See description:

> There is some duplication but the agents (and their tests) will be merged in FLOC-1791.

PR #1405, opened 2015-05-14
[FLOC-1783] Node and cluster uuids are loaded from node TLS certificate

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1783


flocker/ca/_ca.py
```diff
@@ -445,9 +461,18 @@ def initialize(cls, path, authority, begin=None, uuid=None):
             path=path, keypair=keypair, certificate=cert)
         credential.write_credential_files(
             key_filename, cert_filename)
-        instance = cls(credential=credential, uuid=uuid)
+        instance = cls(credential=credential)
         return instance
 
+    @property
+    def uuid(self):
+        # We need to strip the "node-" prefix:
```
tomprince: If we are going to strip the prefix, somewhere we should verify that that is in fact waht we are stripping.

tomprince:
This looks good, except one minor comment, but I'm going to reject this, since the parent branch hasn't landed.

itamarst:
Shall I just merge once base branch has been merged, or resubmit?

tomprince:
Please resubmit.

flocker/ca/functional/test_validation.py
```diff
@@ -49,14 +49,12 @@ def __init__(self):
...
-        if self._buffer == EXPECTED_STRING:
-            self.result.callback("handshake!")
-            self.result = None
 
     def connectionLost(self, reason):
-        if self.result is not None:
+        if self._buffer == EXPECTED_STRING:
+            self.result.callback("handshake!")
+        else:
             self.result.errback(reason)
```
exarkun: Probably would be useful to say something about `self._buffer` as well here (though I expect the typical reason this test would fail is TLS handshake failure ... which is probably mostly represented by `reason`).

:question:

flocker/node/script.py
```diff
@@ -81,7 +82,7 @@ def _context_factory(path, host, port):
     node_credential = NodeCredential.from_path(path, b"node")
     policy = ControlServicePolicy(
         ca_certificate=ca, client_credential=node_credential.credential)
-    return policy.creatorForNetloc(host, port)
+    return (policy.creatorForNetloc(host, port), node_credential)
```
exarkun: Ick.  Tuple return type.  Even `namedtuple` would be better (so callers could use attribute lookup if they don't want to turn into a maintenance burden later).

flocker/ca/_ca.py
```diff
@@ -400,9 +400,13 @@ class NodeCredential(PRecord):
         credential object.
     :ivar bytes uuid: A unique identifier for the node this certificate
         identifies, in the form of a version 4 UUID.
+    :ivar bytes cluster_uuid: A unique identifier for the cluster this
+        certificate identifies, in the form of a version 4 UUID.
```
exarkun: It might make slightly more sense for both of these to be `UUID` instead of `bytes` since user-code code elsewhere seems to go and construct a `UUID` from the values.

flocker/node/test/test_script.py
```diff
@@ -58,13 +58,14 @@ def setup_config(test):
...
     ca_set.copy_to(scratch_directory, node=True)
+    test.ca_set = ca_set
     test.non_existent_file = scratch_directory.child('missing-config.yml')
 
 deployer = object()
 
 
 def deployer_factory_stub(**kw):
-    if set(kw.keys()) != {"node_uuid", "hostname"}:
+    if set(kw.keys()) != {"node_uuid", "cluster_uuid", "hostname"}:
```
exarkun: Would be nice to formalize this somehow.  File an issue for an explicit interface for this thing and for fixing this stub to be at least minimally verified, fake-wise?

flocker/node/test/test_script.py
```diff
@@ -299,7 +331,7 @@ def test_deployer_factory_called_with_ip(self):
         """
         spied = []
 
-        def deployer_factory(node_uuid, hostname):
+        def deployer_factory(node_uuid, hostname, cluster_uuid):
```
exarkun: And this one

exarkun:
Thanks.  Looks good.  Address inline comments and merge when buildbot is green.

flocker/ca/functional/test_validation.py
itamarst: Yeah, given how small the string is you can pretty much expect all of it or disconnection before any user data is exchanged.

PR #1406, opened 2015-05-14
add configurable block_size for different driver backends 

wallnerryan:
This enables configuring the REALISTIC_BLOCKDEVICE_SIZE that is used in test_blockdevice. Related to FLOC-1874 this allows configuring the block_size that may work with some drivers and not others for implementing IBlockDeviceAPI.

This changes make_iblockdeviceapi_tests() to accept a block_size attribute.

Signed-off-by: Ryan Wallner <ryan.wallner@emc.com>


wallnerryan:
This came up with a conversation with @itamarst, this may need tests to make sure block_size is none or of a value, let me know if PR makes sense, what needs to be changed etc. thx.

itamarst:
Hi,

Could you merge forward and include the two callers in the patch? Thanks!

wallnerryan:
which two callers are you referencing?

wallnerryan:
merged forward, fixed a few more in IBlockDeviceAPITestsMixin to use self.block_size

wallnerryan:
```make_iblockdeviceapi_tests``` is called in ```flocker/node/agents/functional/test_cinder.py```, ```flocker/node/agents/functional/test_ebs.py``` , and ```flocker/node/agents/test/test_blockdevice.py```. The ```block_size``` is an optional parameter so these will still work as expected, they will just inherit ```block_size``` from ```REALISTIC_BLOCKDEVICE_SIZE```

vaibhavkhanduja:
In master I see REALISTIC_BLOCKDEVICE_SIZE = int(GB(100).to_Byte().value) is changed to 
REALISTIC_BLOCKDEVICE_SIZE = int(GiB(100).to_Byte().value)

this should probably solve the problem ...

thanks

wallnerryan:
Thanks @vaibhavkhanduja , although it still may be nice to have it configurable. For instance, having large 100GiB volumes are small vagrant testbed could cause problems when running tests (configured my env down to 64GiB). Leaving the PR here until CHQ closes or suggests other wise.

wallrj:
I did something similar in https://github.com/ClusterHQ/flocker/pull/1448 for https://clusterhq.atlassian.net/browse/FLOC-1946 

Let me know if that works for you and if so we'll close this PR.

Thanks again for the contribution and sorry for not commenting sooner.

wallnerryan:
Thanks @wallrj this is what was needed, better than it actually checks the device on the host. Merged in and tested https://github.com/ClusterHQ/flocker-emc/commit/7792d4e601a3ab54095fcbb25c49e0daae24fe86

PR #1407, opened 2015-05-14
[FLOC-1889] Make acceptance tests work on loopback block-device backend.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1889

This currently switches the backend used by default to to the loopback one. Once FLOC-1741 has landed, this will become configurable.


flocker/acceptance/test_deployment.py
```diff
@@ -53,134 +53,6 @@ class DeploymentTests(TestCase):
     """
     @require_flocker_cli
     @require_cluster(num_nodes=2)
-    def test_application_volume_quotas_changed(self, cluster):
```
tomprince: Follow up to https://clusterhq.atlassian.net/browse/FLOC-1848

flocker/acceptance/test_environment_variables.py
```diff
@@ -1,280 +0,0 @@
...
-                dl = loop_until(mysql_can_connect)
-                return dl
-            waiting_for_container.addCallback(mysql_connect)
-            return waiting_for_container
-
-        waiting_for_cluster.addCallback(got_cluster)
-        return waiting_for_cluster
-
-    @require_pymysql
-    def test_environment_variable_used(self):
```
tomprince: https://clusterhq.atlassian.net/browse/FLOC-1888
and 
https://github.com/docker-library/mysql/issues/71

This tests is really the only novel one, and 
https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/test_api.py#L84-L135 tests the same thing. (We lose the test for this code path in `flocker-deploy`, but that is probably acceptable)

flocker/acceptance/testtools.py
```diff
@@ -203,6 +203,61 @@ def _clean_node(test_case, node):
...
+    :param reason: The reason the backend isn't supported.
+    """
+    def decorator(test_method):
+        """
+        :param test_method: The test method that will be called when the
+            cluster is available and which will be supplied with the
+            ``cluster``keyword argument.
+        """
+        def call_test_method_with_cluster(cluster, test_case, args, kwargs):
+            kwargs['cluster'] = cluster
```
exarkun: Cruft?

flocker/control/_persistence.py
```diff
@@ -30,6 +30,8 @@ def default(self, obj):
         if isinstance(obj, PRecord):
             result = dict(obj)
             result[_CLASS_MARKER] = obj.__class__.__name__
+            if result[_CLASS_MARKER] == 'NodeState':
+                result['devices'] = {}
```
exarkun: This looks like something that merits a unit test?  Maybe even its own issue.

flocker/acceptance/testtools.py
```diff
@@ -203,6 +203,61 @@ def _clean_node(test_case, node):
...
+    :param test_case: The ``TestCase`` running this unit test.
+
+    :raise SkipTest: if the backend is specified.
+    """
+    backend = environ.get("FLOCKER_ACCEPTANCE_VOLUME_BACKEND")
+    if backend is None:
+        raise SkipTest(
+            "Set acceptance testing volume backend using the " +
+            "FLOCKER_ACCEPTANCE_VOLUME_BACKEND environment variable.")
+    return backend
```
exarkun: Let's make non-strings the default for representing things like that.  Doesn't have to be `twisted.python.constants` but should at least be something in that ballpark.

flocker/acceptance/testtools.py
```diff
@@ -203,6 +203,61 @@ def _clean_node(test_case, node):
...
+                        reason=reason,
+                        supported=', '.join(supported),
+                    )
+                )
+            return test_method(test_case, *args, **kwargs)
+        return wrapper
+    return decorator
+
+require_moving_backend = require_backend(
+    supported=[b"zfs", b"openstack", b"ebs"],
```
exarkun: This is going to hamper third-party testing.  On the one hand, I'd really like to not have to think about that right now.  On the other hand, I don't know if we'll be able to get away with that.  Rather than agonize over that, maybe it'd be easier just to suggest inverting this and blacklisting `loopback` instead of whitelisting these three.

admin/acceptance.py
```diff
@@ -91,6 +91,7 @@ def check_result(f):
             FLOCKER_ACCEPTANCE_CONTROL_NODE=control_node.address,
             FLOCKER_ACCEPTANCE_AGENT_NODES=':'.join(
                 node.address for node in agent_nodes),
+            FLOCKER_ACCEPTANCE_VOLUME_BACKEND='loopback',
```
exarkun: What's the impact of this on our ability to run acceptance tests against the ZFS backend on BuildBot (in the Vagrant configuration, at least)?

exarkun:
Thanks.  This basically looks good.  I'm concerned about losing testing of ZFS, though.  I think you'll be able to preserve that by making BuildBot set this new environment variable to `zfs` on some builders, though?  If that's the case then this is probably good to merge.

exarkun:
(As long as buildbot is happy, too.  Lots of red builders in the report above at the moment but I lack the bandwidth to look at all of them.)

flocker/control/_persistence.py
tomprince: This was a workaround for FLOC-1878.

admin/acceptance.py
tomprince: This will get rid of it in the short term, but hopefully we'll get it back by the end of the sprint.

tomprince: Note, this only informs the test what backend is being used, not which is running. The changes to systemd/upstart files mean that the loopback backend will be running.

tomprince:
There were some failures with postgres, as well. But since #1413 has landed, it is no longer straightforward to test the loopback agent. Once we can select the backend, we can address any remaining issues.

PR #1408, opened 2015-05-14
[FLOC-1769] Split out ZFS installation instructions.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1769

~~Depends on #1407 ([relative diff](https://github.com/ClusterHQ/flocker/compare/loop-back-testing-FLOC-1889...zfs-instructions-FLOC-1769))~~

I've started to drop support for Fedora 20 here (by not documenting it explicitly). The non-vagrant fedora-20 acceptance tests thus fail here. I'll address this on buildbot before merging.


flocker/provision/_install.py
```diff
@@ -282,6 +275,57 @@ def task_create_flocker_pool_file():
     ])
 
 
+def task_install_zfs(distribution, variants=set()):
+    commands = []
+    if distribution == 'ubuntu-14.04':
```
itamarst: Please split this into separate functions, instead of giant if/else.

itamarst:
Fundamentally this seems sound, but with buildbot all red it's hard to tell if it's actually correct. So resubmit when configuration is available.

flocker/provision/_install.py
tomprince: There is probably a better organization of this code (this file is getting a little unweildly). But given the rest of the code follow this style, I don' think there is currently a benefit in splitting this into three functions. I'm open to suggestions (But I think it is probably worth a separate issue)

adamtheturtle: That separate issue is probably https://clusterhq.atlassian.net/browse/FLOC-1670.

tomprince:
The `fedora-20` acceptance failures are expected.

admin/acceptance.py
```diff
@@ -269,6 +271,12 @@ def start_nodes(self, reactor):
                            variants=self.variants)
             for node in self.nodes
         ])
+        if self.volume_backend == VolumeBackend.zfs:
```
tomprince: I don't want to do dispatch on this yet, until we see what other backends might need.

tomprince:
@davidoliverSP2 I've update the installation instruction here. Could you have a look at the changes?

admin/packaging.py
```diff
@@ -597,7 +593,7 @@ def __str__(self):
         'non-conffile-in-etc /etc/init/flocker-control.conf',
 
         # Cryptography hazmat bindings
-        'package-installs-python-pycache-dir opt/flocker/lib/python2.7/site-packages/cryptography/hazmat/bindings/__pycache__/',
+        'package-installs-python-pycache-dir opt/flocker/lib/python2.7/site-packages/cryptography/hazmat/bindings/__pycache__/',  # noqa
```
exarkun: I sorta hate flake8.

exarkun:
> The fedora-20 acceptance failures are expected.

I guess that builder will be disabled soon?  Is there a build.clusterhq.com PR to review?

Apart from that ... looks good.  The vagrant tutorial box supposedly built successfully, that's good.  Hm.  But the fedora-20 acceptance failure (ie, it doesn't run at all) means the box isn't really tested, I think?  That's not so good.  Ideally we'd need a CentOS 7 vagrant tutorial box so we could have a CentOS 7 vagrant acceptance builder to verify it...

What's the plan for this?

tomprince:
> I guess that builder will be disabled soon? Is there a build.clusterhq.com PR to review?

Yes. I wasn't planning on bothering with a PR to delete those two lines.

> the fedora-20 acceptance failure (ie, it doesn't run at all)

That is `acceptance/rackspace/fedora-20`, not `acceptance/vagrant/fedora-20`. The later has a single acceptance test failure (I'm not sure what is going on there).

exarkun:
> Yes. I wasn't planning on bothering with a PR to delete those two lines.

Okay, great.  That's fine.

> That is acceptance/rackspace/fedora-20, not acceptance/vagrant/fedora-20. The later has a single acceptance test failure (I'm not sure what is going on there).

Aahh right.  Okay, great.  This looks good to me, then.  Address inline comments and then merge.

exarkun:
> Address inline comments

Force of habit, I don't think I made any actionable comments.

PR #1409, opened 2015-05-14
Get the region from the environment.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1890

Goes along with https://github.com/ClusterHQ/build.clusterhq.com/pull/81 which sets the environment variable this depends on.

This is currently staged on http://23.253.99.92/boxes-flocker

### DIFF
--- a/flocker/node/agents/test/blockdevicefactory.py
+++ b/flocker/node/agents/test/blockdevicefactory.py
@@ -159,12 +159,12 @@ def _openstack_auth_from_config(**config):
     return plugin_class(**plugin_kwargs)
 
 
-def _openstack(region, **config):
+def _openstack(**config):
     """
     Create Cinder and Nova volume managers suitable for use in the creation of
-    a ``CinderBlockDeviceAPI``.
+    a ``CinderBlockDeviceAPI``.  They will be configured to use the region
+    where the server that is running this code is running.
 
-    :param bytes region: The name of the region to which to connect.
     :param config: Any additional configuration (possibly provider-specific)
         necessary to authenticate a session for use with the CinderClient and
         NovaClient.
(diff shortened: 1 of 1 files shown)


flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -172,6 +172,14 @@ def _openstack(region, **config):
     :return: A ``dict`` giving initializer arguments for
         ``CinderBlockDeviceAPI``.
     """
+    # The execution context should have set up this environment variable,
+    # probably by inspecting some cloud-y state to discover where this code is
+    # running.  Since the execution context is probably a stupid shell script,
+    # fix the casing of the region name here (keystone is very sensitive to
+    # case) instead of forcing me to figure out how to upper case things in
+    # bash (I already learned a piece of shell syntax today, once is all I can
```
tomprince: :question: The personal comments probably don't belong. 

:crying_cat_face:

tomprince:
Looking at the staged builder, it looks like there are some failing tests. They appear to be resizes tests. Please fix these tests failures before merging.

tomprince:
This can be merged after ClusterHQ/build.clusterhq.com#81  lands. (You can use `failing_builder` to avoid generating spurious failures on other branches).

wallrj:
For our PistonCloud, I need to be able to omit the region.

Consider changing the implementation to be something like:
```
+    region = environ.get('FLOCKER_FUNCTIONAL_TEST_OPENSTACK_REGION')
+    if region is not None:
+        region = region.upper()
```

...I'm omitting the region because of what seems to be a bug in the keystone library / server, but since there's only one region there anyway, this might be easier than, for example, having a special case `PistonCloud` factory function.

wallrj:
Also note that the case changing stuff is a hack to accommodate the fact that the acceptance.yml file in lastpass contains a lower case region slug.
We should probably just change the case there and check whether the acceptance tests (which use libcloud Rackspace driver) continue to work....or did we already consider / test that? I can't remember.

Perhaps the libcloud driver does a case insensitive comparison when comparing the supplied region slug to those listed by the nova API.

exarkun:
Making the region optional sounds reasonable.  I think I'm going to punt that down the road to you, though, @wallrj, unless you have any serious objections.  This code doesn't have test coverage (it is test coverage) so I'd rather not make these changes without a configuration where I can actually see if they work - and I think that configuration is the pistoncloud work you're doing.

exarkun:
> Also note that the case changing stuff is a hack to accommodate the fact that the acceptance.yml file in lastpass contains a lower case region slug.

Sufficiently localized hack that I don't want to worry about it right now.

PR #1412, opened 2015-05-15
Add API headers FLOC 1867

davidoliverSP2:
https://clusterhq.atlassian.net/browse/FLOC-1867

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -104,9 +104,12 @@ class ConfigurationAPIUserV1(object):
         self.cluster_state_service = cluster_state_service
 
     @app.route("/version", methods=['GET'])
-    @user_documentation("""
+    @user_documentation(
+        """
         Get the version of Flocker being run.
-        """, examples=[u"get version"])
+        """,
+        header=u"Get Flocker version"
+        examples=[u"get version"])
     @structured(
         inputSchema={},
         outputSchema={'$ref': '/v1/endpoints.json#/definitions/versions'},
@@ -123,6 +126,7 @@ class ConfigurationAPIUserV1(object):
         """
         Get the cluster's dataset configuration.
         """,
+        header=u"Get the cluster's dataset configuration"
         examples=[u"get configured datasets"],
     )
     @structured(
@@ -147,6 +151,7 @@ class ConfigurationAPIUserV1(object):
         """
         Create a new dataset.
         """,
+        header=u"Create new dataset"
         examples=[
             u"create dataset",
             u"create dataset with dataset_id",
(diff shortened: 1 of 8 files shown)


flocker/restapi/_infrastructure.py
```diff
@@ -223,7 +223,7 @@ def got_result(result):
     return deco
 
 
-def user_documentation(doc, examples=None):
+def user_documentation(doc, examples=None, header=None):
```
tomprince: If it is a problem that there isn't a header, then it probably makes sent to require it to be present, rather than having a default value for it.

flocker/restapi/docs/publicapi.py
```diff
@@ -183,6 +183,8 @@ def _introspectRoute(route, exampleByIdentifier, schema_store):
     """
     result = {}
 
+    result['header'] = route.attributes.get('header', 'No header!')
```
tomprince: Another option would be to raise [`SphinxError`](http://sphinx-doc.org/extdev/appapi.html#sphinx.errors.SphinxError) here.

tomprince:
It would be slightly easier to review if moving the documentation for configuring the API was separate from the rest of the changes, since they don't really depend on one another.

flocker/control/httpapi.py
```diff
@@ -362,13 +365,14 @@ def saved(ignored):
         return saving
 
     @app.route("/state/datasets", methods=['GET'])
-    @user_documentation("""
-        Get current cluster datasets.
```
tomprince: One though I had, is, that we could just turn this into a header here, i.e.
```python
@user_documentation("""
    Get current cluster datasets
    ============================

    ....
```
But given that I think we want to enforce headers being there, then it is easier to do what you did.

flocker/restapi/docs/publicapi.py
```diff
@@ -363,6 +365,10 @@ def makeRst(prefix, app, exampleByIdentifier, schema_store):
     for route in sorted(getRoutes(app)):
         data = _introspectRoute(route, exampleByIdentifier, schema_store)
         for method in route.methods:
+            print data['header']
```
tomprince: This looks like debugging output.

tomprince:
Two minor comments to address, and then merge when buildbot is happy.

flocker/restapi/_infrastructure.py
davidoliverSP2: unfortunately there is code that expects the 2nd parameter to be "examples", so header has to be 3rd. Therefore it needs to have this default value.

flocker/restapi/docs/publicapi.py
davidoliverSP2: Added a sphinxerror for when no headers are provided on the API calls.

flocker/restapi/docs/publicapi.py
davidoliverSP2: noted - removed the print statement.

flocker/restapi/_infrastructure.py
tomprince: You could just change that code to use keyword arguments.

PR #1414, opened 2015-05-15
[FLOC-1841] Pervasive ignorance in node state

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1841
Depends on #1398.


flocker/control/_config.py
```diff
@@ -1301,6 +1301,8 @@ def model_from_configuration(deployment_state, applications,
 
 def current_from_configuration(current_configuration):
     """
+    UNUSED, remove when convenient.
+
```
exarkun: Any reason now isn't convenient?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -982,12 +983,13 @@ def test_no_devices_one_dataset(self):
...
                     uuid=uuid,
-                    hostname=node,
                     manifestations={dataset_id: manifestation},
                 )
             }
         )
-        state = DeploymentState(nodes=[])
+        state = DeploymentState(nodes=[NodeState(
+            uuid=uuid, hostname=node, applications=[], manifestations={},
+            devices={}, paths={})])
```
exarkun: This is still sort of invalid, right?  This `NodeState` represents `applications` but not `used_ports` which isn't something actually possible in real use.

We have `flocker.node.testtools.EMPTY` and `EMPTY_STATE`.  I wonder if we should have similar things for `Node` and `NodeState`.  `EMPTY_NODE_STATE`, `EMPTY_NODE_APP_STATE`, `EMPTY_NODE_DATASET_STATE`.  Centralize proper initialization of all these attributes.  File as a follow-up issue if you want.

Another thing that might be nice is more invariant enforcement on `NodeState`.  Knownness of `used_ports` and `applications` matching could be enforced; likewise for knownness of `manifestations`, `devices`, `paths`.  Another follow-up issue, probably.

flocker/node/test/test_deploy.py
```diff
@@ -619,7 +619,9 @@ def test_does_not_exist(self):
     name=APP_NAME2,
     image=DockerImage.from_string(UNIT_FOR_APP2.container_image)
 )
-EMPTY_NODESTATE = NodeState(hostname=u"example.com")
+EMPTY_NODESTATE = NodeState(hostname=u"example.com",
+                            manifestations={}, applications=[],
+                            used_ports=[], devices={}, paths={})
```
exarkun: Kind of a nit but grouping application attributes together and grouping dataset attributes together wouldn't hurt.

exarkun: Also - pass a uuid here.  And hey this looks kinda like `EMPTY_NODE_STATE` so you could just move it to `flocker.node.testtools` and use it from both here and in `test_no_devices_one_dataset` above.

flocker/node/test/test_deploy.py
```diff
@@ -631,7 +633,9 @@ def setUp(self):
         self.network = make_memory_network()
         self.node_uuid = uuid4()
         self.EMPTY_NODESTATE = NodeState(hostname=u"example.com",
-                                         uuid=self.node_uuid)
+                                         uuid=self.node_uuid,
+                                         manifestations={}, applications=[],
+                                         used_ports=[], devices={}, paths={})
```
exarkun: Here as well

flocker/node/test/test_deploy.py
```diff
@@ -1168,8 +1166,11 @@ def test_proxy_needs_creating(self):
 
         desired = Deployment(nodes=nodes)
         current = DeploymentState(nodes=[
+            NodeState(
+                uuid=api.node_uuid, hostname=api.hostname, applications=[]),
```
exarkun: Another known `applications`/unknown `used_ports`

flocker/node/test/test_deploy.py
```diff
@@ -1168,8 +1166,11 @@ def test_proxy_needs_creating(self):
...
             NodeState(uuid=destination_node_uuid,
-                      hostname=expected_destination_host)])
+                      hostname=expected_destination_host, manifestations={},
```
exarkun: known `manifestations`/ unknown `devices` and `paths

flocker/node/test/test_deploy.py
```diff
@@ -1242,19 +1244,23 @@ def test_open_port_needs_creating(self):
...
             Node(
-                hostname=api.hostname,
+                uuid=api.node_uuid,
                 applications=[application]
             )
         ]
 
+        node_state = NodeState(
+            hostname=api.hostname, uuid=api.node_uuid,
+            applications=[])
```
exarkun: More knowledge/ignorance mismatch

flocker/node/test/test_deploy.py
```diff
@@ -1312,18 +1319,22 @@ def test_application_needs_starting(self):
 
         nodes = frozenset([
             Node(
-                hostname=u'example.com',
+                uuid=api.node_uuid,
```
exarkun: This and the other similar changes to avoid hitting the deprecated codepaths are good ... but would be a lot easier to review in a separate PR (particularly one that makes non-test changes).

flocker/node/test/test_deploy.py
```diff
@@ -1312,18 +1319,22 @@ def test_application_needs_starting(self):
...
                 applications=frozenset([application])
             )
         ])
 
+        node_state = NodeState(
+            hostname=api.hostname, uuid=api.node_uuid,
+            applications=[])
```
exarkun: That mismatch again

flocker/node/test/test_deploy.py
```diff
@@ -1764,6 +1775,7 @@ def test_dataset_deleted(self):
             hostname=u"10.1.1.1",
             manifestations={MANIFESTATION.dataset_id:
                             MANIFESTATION},
+            applications=[],
```
exarkun: And here (on both app and data sides)

flocker/node/test/test_deploy.py
```diff
@@ -1857,7 +1869,7 @@ def test_no_handoff_if_in_use(self):
             applications={APPLICATION_WITH_VOLUME},
         )
         another_node_state = NodeState(
-            hostname=u"node2.example.com",
+            hostname=u"node2.example.com", manifestations={},
```
exarkun: And here

flocker/node/test/test_deploy.py
```diff
@@ -1902,17 +1914,19 @@ def test_no_handoff_if_destination_unknown(self):
...
-        node but is now running on another node must be handed off.
+        ``P2PManifestationDeployer.calculate_changes`` specifies that a volume
+        was previously running on this node but is now running on another
+        node must be handed off.
         """
         node_state = NodeState(
             hostname=u"node1.example.com",
             manifestations={MANIFESTATION.dataset_id:
                             MANIFESTATION},
+            applications=[],
```
exarkun: And here

flocker/node/test/test_deploy.py
```diff
@@ -1902,17 +1914,19 @@ def test_no_handoff_if_destination_unknown(self):
...
         )
         another_node_state = NodeState(
             hostname=u"node2.example.com",
+            manifestations={},
```
exarkun: And here

exarkun: And a handful more times below in the obvious places

flocker/node/testtools.py
```diff
@@ -226,5 +226,5 @@ def to_node(node_state):
     :return Node: Equivalent node.
     """
     return Node(uuid=node_state.uuid, hostname=node_state.hostname,
-                applications=node_state.applications,
-                manifestations=node_state.manifestations)
+                applications=node_state.applications or [],
+                manifestations=node_state.manifestations or {})
```
exarkun: Mmm this is interesting.  Maybe this change isn't required now that the various test data has been updated to supply empty applications/manifestations where that's what it actually means?  If it's not required it would probably be better to drop it (and perhaps even turn these cases into explicit failure modes with nice exceptions).  I get a vague sense that `to_node` behaving this way may be a good way to hide some bugs.


(I tried running the test suite with a change to make this raise an exception instead of supply empty defaults but the test suite still doesn't get past `test_pull_image_if_necessary` for me :cry: ).

exarkun:
Thanks.  This looks pretty good.  Fix up the model inconsistencies so the tests are operating on realistic data, address other inline comments, and then merge.

flocker/node/test/test_deploy.py
itamarst: Unfortunately until we remove the hack where UUID is calculated from hostname I don't want to unify stuff too much, since probably different pieces of code have different expectations.

flocker/node/agents/test/test_blockdevice.py
itamarst: I think I'll just do the invariant in this branch, since it'll reduce the chances of this branch introducing subtle bugs.

flocker/node/test/test_deploy.py
itamarst: Also passing in uuid will break a bunch of tests that assuming just having `hostname="example.com"` is sufficient.

flocker/node/testtools.py
itamarst: Unfortunately it's still perfectly valid to have missing applications or missing manifestations. Lots of tests require both because we have the stupid workaround for lack of leasing, but that isn't in place in many places.

PR #1417, opened 2015-05-18
Fix doc tests floc 1913

jongiddy:
This undoes the constraint introduced in PR #1412 that `header` is a required field in `user_documentation` decorator. Several tests do not include the `user_documentation` decorator at all, and it is possibly reasonable for this to happen for real operations ("private" operations, for example).


exarkun:
Code-wise, this seems fine to me.  Doc-feature-wise, it seems like it would be nice if endpoints definitely had to be documented or else explicitly marked as undocumented (eg the private endpoint).  But I guess we don't force the presence of other pieces of doc data so there's no good reason to start imposing that requirement here.

Go ahead and merge once buildbot is green and if @davidoliverSP2 is happy with the change.

PR #1421, opened 2015-05-18
[FLOC-1824] Get the local instance ID by comparing IP server addresses

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1824

The interesting build results are in the `expected_failures` section and not listed below. Here's a direct link:
 * http://build.clusterhq.com/builders/flocker%2Ffunctional%2Frackspace%2Fcentos-7%2Fstorage-driver/builds/1


flocker/node/agents/cinder.py
```diff
@@ -169,36 +180,60 @@ def wait_for_volume(volume_manager, expected_volume,
...
         self.cinder_volume_manager = cinder_volume_manager
         self.nova_volume_manager = nova_volume_manager
+        self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
 
     def compute_instance_id(self):
         """
-        Look up the Xen instance ID for this node.
+        Find the Nova API server with the same non-loopback IP
+        addresses as this node.
```
wallrj: Well, in fact we now look for the API server addresses that are a subset of the locally discovered IPv4 and IPv6 addresses.
Probably should update this docstring.

flocker/node/agents/cinder.py
```diff
@@ -169,36 +180,60 @@ def wait_for_volume(volume_manager, expected_volume,
...
+        Return the instance ID of that server.
         """
-        # See http://wiki.christophchamp.com/index.php/Xenstore
-        # $ sudo xenstore-read name
-        # instance-6ddfb6c0-d264-4e77-846a-aa67e4fe89df
-        prefix = u"instance-"
-        command = [b"xenstore-read", b"name"]
-        return check_output(command).strip().decode("ascii")[len(prefix):]
+        local_ips = get_all_ips()
+        for server in self.nova_server_manager.list(detailed=True):
```
wallrj: The `detailed=True` argument is the default. And I didn't include it in `INovaServerManager` so I should probably remove it here.

flocker/node/agents/cinder.py
```diff
@@ -305,8 +340,8 @@ def detach_volume(self, blockdevice_id):
 
         # This'll blow up if the volume is deleted from elsewhere.  FLOC-1882.
         wait_for_volume(
-            volume_manager=self.nova_volume_manager,
-            expected_volume=nova_volume,
+            volume_manager=self.cinder_volume_manager,
```
wallrj: Switching to `cinder_api` instead of instantiating `CinderBlockAPI` revealed that `wait_for_volume` was calling `list` which isn't declared in the `INovaVolumeManager` interface.
So I now use the cinder client to get the volume and to wait for it's status to change.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -79,16 +79,16 @@ def test_foreign_volume(self):
...
             size=Byte(REALISTIC_BLOCKDEVICE_SIZE).to_GB().value
         )
         self.addCleanup(
-            cinder_volumes.delete,
+            cinder_client.volumes.delete,
             requested_volume.id,
         )
         wait_for_volume(
-            volume_manager=cinder_volumes,
+            volume_manager=cinder_client.volumes,
```
wallrj: More changes related to the `cinder_api` factory function...see below.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -179,7 +178,9 @@ def _openstack(**config):
     # case) instead of forcing me to figure out how to upper case things in
     # bash (I already learned a piece of shell syntax today, once is all I can
     # take).
-    region = environ['FLOCKER_FUNCTIONAL_TEST_OPENSTACK_REGION'].upper()
+    region = environ.get('FLOCKER_FUNCTIONAL_TEST_OPENSTACK_REGION')
```
wallrj: There's a bug in the Keystone Server / client API in our PistonCloud test environment which means that it doesn't respond well to region slugs.
This allows me to run the tests in the PistonCloud by omitting the region slug from the environment.

wallrj: I should probably create an issue for that.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -220,9 +221,9 @@ def _aws(**config):
 # Map provider labels to IBlockDeviceAPI factory and a corresponding argument
 # factory.
 _BLOCKDEVICE_TYPES = {
-    ProviderType.openstack: (CinderBlockDeviceAPI, _openstack),
+    ProviderType.openstack: (cinder_api, _openstack),
```
wallrj: This meant that none of the `LoggingNova???` wrappers were being used.

flocker/common/_net.py
```diff
@@ -0,0 +1,26 @@
...
+"""
+
+import netifaces
+
+
+def get_all_ips():
+    """
+    Find all IPs for this machine.
+
+    :return: ``set`` of IP addresses (``bytes``).
```
exarkun: An `ipaddr` type would make sense here (instead of `bytes`).

flocker/node/agents/cinder.py
```diff
@@ -169,36 +180,60 @@ def wait_for_volume(volume_manager, expected_volume,
             )
 
 
+def _extract_nova_server_addresses(addresses):
+    """
+    :param dict addresses: A ``dict`` mapping OpenStack network names
+        to lists of address dictionaries in that network assigned to a
+        server.
+    :return: A ``set`` of all the IPv4 and IPv6 addresses from the
+        ``addresses`` attribute of a ``Server``.
```
exarkun: `ipaddr` here too (currently bytes but the docstring doesn't say this)

flocker/node/agents/cinder.py
```diff
@@ -169,36 +180,60 @@ def wait_for_volume(volume_manager, expected_volume,
...
+        self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
 
     def compute_instance_id(self):
         """
-        Look up the Xen instance ID for this node.
+        Find the Nova API server with the same non-loopback IP
+        addresses as this node.
+        That server is assumed to be this node.
+        Return the instance ID of that server.
```
exarkun: Convert this to proper markup.  Also, what if no match is found?  The error case is probably worth handling explicitly.  We may also want to log enough information here that we could debug a failure caused by someone having a network configuration we didn't anticipate.

flocker/node/agents/cinder.py
```diff
@@ -291,8 +326,8 @@ def attach_volume(self, blockdevice_id, attach_to):
     def detach_volume(self, blockdevice_id):
         our_id = self.compute_instance_id()
         try:
-            nova_volume = self.nova_volume_manager.get(blockdevice_id)
-        except NovaNotFound:
+            cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
+        except CinderNotFound:
```
exarkun: Hm.  Does this make any real difference?

If we stick with this, we no longer use `get` from the nova volume manager so maybe we should drop that from the interface.

flocker/node/agents/cinder.py
exarkun: Ah, makes sense.

flocker/node/agents/cinder.py
exarkun: (comment just below explains this change)

flocker/node/agents/test/blockdevicefactory.py
exarkun: These changes should work on Rackspace too.  I'd confine this PR to the basic address-driven heuristic.  Then make fixes as necessary after there's a PistonCloud builder set up.

exarkun:
Thanks.  Looks good.  Please address inline comments (including your own :smile: ) and then merge.

flocker/node/agents/test/blockdevicefactory.py
wallrj: The PistonCloud builder now exists....so I think I'll leave this.

wallrj:
I addressed the comments and forced a new build on my (unmerged) pistoncloud buildslave:
 http://23.253.99.92/builders/flocker%2Ffunctional%2Fpistoncloud%2Fcentos-7%2Fstorage-driver/builds/6

I'll merge tomorrow if all goes well.

PR #1422, opened 2015-05-18
[FLOC-1835] Add Boto debug config option.

myechuri:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1835

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -18,7 +18,7 @@ from boto import ec2
 from boto import config
 from boto.ec2.connection import EC2Connection
 from boto.utils import get_instance_metadata
-from boto.exception import BotoClientError, BotoServerError
+from boto.exception import EC2ResponseError
 from twisted.python.filepath import FilePath
 from eliot import Field, MessageType
 
@@ -32,9 +32,18 @@ METADATA_VERSION_LABEL = u'flocker-metadata-version'
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
 BOTO_NUM_RETRIES = u'10'
+
+# Set Boto debug level to `1`, requesting basic debug messages from Boto
+# to be printed.
+# See https://code.google.com/p/boto/wiki/BotoConfig#Boto
+# for available debug levels.
+# Warning: BotoConfig is a Boto global option, so, all Boto clients from
+# from this process will have their log levels modified. See
+# https://clusterhq.atlassian.net/browse/FLOC-1962
+# to track evaluation of impact of log level change.
 BOTO_DEBUG_LEVEL = u'1'
 
-# Scaffolding for logging Boto client and server exceptions
+# Begin: Scaffolding for logging Boto client and server exceptions
 # via Eliot.
 CODE = Field.for_types("code", [int], u"The HTTP response code.")
 MESSAGE = Field.for_types(
@@ -50,27 +59,17 @@ URL = Field.for_types("url", [bytes, unicode], u"The request URL.")
 METHOD = Field.for_types("method", [bytes, unicode], u"The request method.")
 
 RESPONSE = Field.for_types("response", [bytes, unicode], u"The response body.")
-BOTO_CLIENT_ERROR = MessageType(
-    u"boto:boto_client_error", [
-        CODE,
-        RESPONSE,
-        MESSAGE,
-        DETAILS,
-        REQUEST_ID,
-        URL,
-        METHOD,
-    ],
-)
-BOTO_SERVER_ERROR = MessageType(
-    u"boto:boto_server_error", [
+
+# Log boto.exception.BotoEC2ResponseError, covering all errors from AWS:
+# server operation rate limit exceeded, invalid server request parameters, etc.
+BOTO_EC2RESPONSE_ERROR = MessageType(
+    u"boto:boto_ec2response_error", [
         CODE,
         MESSAGE,
-        DETAILS,
         REQUEST_ID,
-        URL,
-        METHOD,
     ],
 )
+# End: Scaffolding for logging Boto errors.
 
 
 def ec2_client(region, zone, access_key_id, secret_access_key):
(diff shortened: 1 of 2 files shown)


flocker/node/agents/ebs.py
```diff
@@ -29,6 +32,45 @@
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
 BOTO_NUM_RETRIES = u'10'
+BOTO_DEBUG_LEVEL = u'1'
+
+# Scaffolding for logging Boto client and server exceptions
+# via Eliot.
+CODE = Field.for_types("code", [int], u"The HTTP response code.")
+MESSAGE = Field.for_types(
+    "message", [bytes, unicode],
```
wallrj: Does the type really vary?

flocker/node/agents/ebs.py
```diff
@@ -29,6 +32,45 @@
...
+    u"A human-readable error message given by the response.",
+)
+DETAILS = Field.for_types("details", [dict], u"Extra details about the error.")
```
wallrj: Will the `dict` of details always be serialisable to json? I should read on..

wallrj: Oh, I see this is exactly how @exarkun did it for OpenStack....I guess he knows what he's doing....so fine.

flocker/node/agents/ebs.py
```diff
@@ -29,6 +32,45 @@
...
+BOTO_CLIENT_ERROR = MessageType(
+    u"boto:boto_client_error", [
+        CODE,
+        RESPONSE,
+        MESSAGE,
+        DETAILS,
+        REQUEST_ID,
+        URL,
+        METHOD,
+    ],
```
wallrj: :question: It'd be helpful to add a description, explaining when / where / how this error might be expected to occur.
 * http://eliot.readthedocs.org/en/0.7.1/types.html#message-types

flocker/node/agents/ebs.py
```diff
@@ -29,6 +32,45 @@
...
+)
+BOTO_SERVER_ERROR = MessageType(
+    u"boto:boto_server_error", [
+        CODE,
+        MESSAGE,
+        DETAILS,
+        REQUEST_ID,
+        URL,
+        METHOD,
+    ],
```
wallrj: Here too.

flocker/node/agents/ebs.py
```diff
@@ -58,11 +100,95 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
     config.set('Boto', 'num_retries', BOTO_NUM_RETRIES)
     config.set('Boto', 'metadata_service_num_attempts', BOTO_NUM_RETRIES)
 
+    # Set Boto debug level to BOTO_DEBUG_LEVEL:
+    # "1": basic debug messages
+    config.set('Boto', 'debug', BOTO_DEBUG_LEVEL)
```
wallrj: :question: Are we setting global config options here? Or just for the scope of this connection? It looks global, which seems a bit drastic? 
I guess it doesn't matter too much but if some other library in the agent process was also using boto, it'd be affected by this too, right? Create a followup issue if you agree that this is an issue and if there's a better way.

flocker/node/agents/ebs.py
myechuri: Yes, i copied @exarkun's message format, filtered out things not need for EBS (interface decorator  parts).
My interpretation of two types (bytes, unicode) were "string" type flavors. Please correct me if that is not the case.

flocker/node/agents/ebs.py
```diff
@@ -29,6 +32,45 @@
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
 BOTO_NUM_RETRIES = u'10'
+BOTO_DEBUG_LEVEL = u'1'
```
wallrj: :question: Maybe add a comment and a link explaining what this will level means:
 * https://code.google.com/p/boto/wiki/BotoConfig#Boto

flocker/node/agents/ebs.py
myechuri: Good idea! Fixed.

flocker/node/agents/ebs.py

wallrj:
Thanks @myechuri 

This all looks good to me.

I guess I prefer the zope interface approach that we used in the OpenStack version, rather than wrapping every connection method, but it should do the job. 
We quite liked that the `Interface` ensured that we knew exactly how we were calling the cinder and nova APIs.

I had some questions about the Field types being `unicode` or `bytes` but i see the same is true in the OpenStack logging branch, so fine.

Some of the docstrings use ` rather than `` around python and code references. 

 And finally, I can't see the results of the logging because I don't think any of the failing AWS builders have attempted to test this branch.

Please address or answer the inline comments above, ensure that the logged messages are actually visible in the failing buildbot results and then merge when you're happy with the build results.

flocker/node/agents/ebs.py
myechuri: Agree on the impact on other modules in the process. Filed https://clusterhq.atlassian.net/browse/FLOC-1962 and linked to it from ebs.py.
Debug level "1" might be useful in troubleshooting operations from other modules as well, but as you pointed out, would be good to quantify impact. Thanks for noticing!

flocker/node/agents/ebs.py

myechuri:
> And finally, I can't see the results of the logging because I don't think any of the failing AWS builders have attempted to test this branch.

The test changes in test_ebs.py were not included in the build results, my bad. Let me do another push to get new test results. Thanks for pointing out!

myechuri:
> I guess I prefer the zope interface approach that we used in the OpenStack version, rather than wrapping every connection method, but it should do the job. 
We quite liked that the Interface ensured that we knew exactly how we were calling the cinder and nova APIs.

i started off initially with interface approach (since i was following OpenStack pattern), but felt that the levels of indirection were too many. i now understand the reason for the indirection (what you mentioned above). If that is the preferred way, happy to move in that direction in the follow up branch where i will work on tracing IBlockDeviceAPI parameter values. Thanks!

myechuri:
Verified latest run's test log has the Eliot output (http://build.clusterhq.com/builders/flocker%2Ffunctional%2Faws%2Fcentos-7%2Fstorage-driver/builds/17/steps/trial/logs/test.log):

```
2015-05-22 22:37:26+0000 [-] --> flocker.node.agents.functional.test_ebs.EBSBlockDeviceAPIInterfaceTests.test_boto_ec2response_error <--
2015-05-22 22:37:27+0000 [-] ELIOT: {'code': u'MissingParameter',
	 'message': u'The request must contain the parameter size/snapshot',
	 u'message_type': u'boto:boto_ec2response_error',
	 'request_id': u'eb9cbfcb-a5ac-4a0f-b3bc-60f343361df8',
	 u'task_level': [1],
	 u'task_uuid': u'61141c50-0473-4144-a9f5-17387f495ff0',
	 u'timestamp': 1432334247.007681}
2015-05-22 22:37:27+0000 [-] ELIOT: {'code': u'InvalidZone.NotFound',
	 'message': u"The zone 'invalid_zone' does not exist.",
	 u'message_type': u'boto:boto_ec2response_error',
	 'request_id': u'381b94d2-90a8-4f76-a24f-dc758ef9d19a',
	 u'task_level': [2],
	 u'task_uuid': u'61141c50-0473-4144-a9f5-17387f495ff0',
	 u'timestamp': 1432334247.024903}
```

PR #1423, opened 2015-05-19
Twisted 152 floc 1918

adamtheturtle:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1918

### DIFF
--- a/flocker/common/functional/test_script.py
+++ b/flocker/common/functional/test_script.py
@@ -135,12 +135,10 @@ FlockerScriptRunner({}(), Options()).main()
         def got_messages(messages):
             assertContainsFields(self, messages[1],
                                  {u"message_type": u"twisted:log",
-                                  u"message": u"stdout!",
-                                  u"error": False})
+                                  u"message": u"stdout!"})
             assertContainsFields(self, messages[2],
                                  {u"message_type": u"twisted:log",
-                                  u"message": u"stderr!",
-                                  u"error": True})
+                                  u"message": u"stderr!"})
         d.addCallback(got_messages)
         return d
 
--- a/admin/build_targets/centos-7/Dockerfile
+++ b/admin/build_targets/centos-7/Dockerfile
@@ -12,7 +12,7 @@ RUN ["yum", "install", "--assumeyes", "git", "ruby-devel", "python-devel", "libf
 # See https://github.com/jordansissel/fpm/issues/611
 RUN ["gem", "install", "fpm"]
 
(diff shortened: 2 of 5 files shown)


flocker/common/functional/test_script.py
```diff
@@ -135,12 +135,10 @@ def test_stdout_stderr(self):
         def got_messages(messages):
             assertContainsFields(self, messages[1],
                                  {u"message_type": u"twisted:log",
-                                  u"message": u"stdout!",
-                                  u"error": False})
+                                  u"message": u"stdout!"})
             assertContainsFields(self, messages[2],
```
adamtheturtle: This is the bit I'm not sure about but I'm not sure if this is important. If it is maybe I shouldn't continue with this as it was only meant to be a quick issue to get the test output benefits.

Without this:

```
[FAIL]
  File "/srv/buildslave/flocker-ubuntu-14_04/venv/local/lib/python2.7/site-packages/flocker/common/functional/test_script.py", line 143, in got_messages
    u"error": True})
  File "/srv/buildslave/flocker-ubuntu-14_04/venv/local/lib/python2.7/site-packages/eliot/testing.py", line 44, in assertContainsFields
    test.assertEqual(messageSubset, fields)
  File "/srv/buildslave/flocker-ubuntu-14_04/venv/local/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 437, in assertEqual
    super(_Assertions, self).assertEqual(first, second, msg)
  File "/usr/lib/python2.7/unittest/case.py", line 515, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/lib/python2.7/unittest/case.py", line 837, in assertDictEqual
    self.fail(self._formatMessage(msg, standardMsg))
twisted.trial.unittest.FailTest: {u'message': u'stderr!', u'message_type': u'twisted:log', u'error': False} != {u'message': u'stderr!', u'message_type': u'twisted:log', u'error': True}
- {u'error': False, u'message': u'stderr!', u'message_type': u'twisted:log'}
?            ^^^^

+ {u'error': True, u'message': u'stderr!', u'message_type': u'twisted:log'}
?            ^^^


flocker.common.functional.test_script.FlockerScriptRunnerTests.test_stdout_stderr
```

exarkun: I wonder why the existing builder didn't catch this problem?

exarkun: Filed this upstream, anyway, https://twistedmatrix.com/trac/ticket/7903

exarkun: This probably isn't a difference that really matters to us.  The change in Twisted is a regression, though, so we should probably expect a 15.2.1 that re-introduces the previous behavior.  If that happens we won't want to have changed our tests.

PR #1424, opened 2015-05-19
[FLOC-1601] API acceptance tests clean up

dwgebler:
Fixes FLOC-1601

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -608,6 +608,29 @@ class Cluster(PRecord):
         return request
 
     @log_method
+    def reset_state(self):
+        """
+        Reset the state of a cluster by sending an empty application and
+        deployment config to the _compose API endpoint.
+        """
+        properties = {
+            "applications": {},
+            "deployment": {"version": 1}
+        }
+        for node in self.nodes:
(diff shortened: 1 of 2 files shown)


flocker/acceptance/testtools.py
```diff
@@ -608,6 +608,28 @@ def delete_dataset(self, dataset_id):
...
+            properties["deployment"]["nodes"][node.address] = []
+        request = post(
+            self.base_url + b"/configuration/_compose",
+            data=dumps(properties),
+            headers={b"content-type": b"application/json"},
+            persistent=False
+        )
+        request.addCallback(check_and_decode_json, OK)
+        request.addCallback(lambda response: (self, response))
+        return request
```
exarkun: There's similar cleanup code in `get_nodes` already.  `get_nodes` also does filesystem-level cleanup (it destroys ZFS filesystems in the `flocker` pool).  It seems like using `get_nodes` instead of using `get_test_cluster` would be a better way to achieve the desired cleanup.

exarkun:
Thanks.  Please see if you can switch over to `get_nodes` to avoid duplicating the cleanup logic.  Then resubmit.

PR #1425, opened 2015-05-19
[FLOC-1601] API acceptance tests clean up

dwgebler:
Fixes FLOC-1601


dwgebler:
@exarkun my only concern with using `get_nodes` is that it uses and therefore requires `flocker-deploy` CLI and does the old thing of SSHing in to nodes to run `zfs destroy` - arguably the API tests should not require Flocker CLI, since the tests themselves are not concerned with it. Perhaps `get_nodes` should be changed to use the API instead of calling `flocker-deploy` CLI?

itamarst:
1. The SSH/ZFS calls should be replaced with calls to the API to list all datasets and delete them; that way it'll work with any backend. flocker-deploy could be replaced with equivalent "list all containers and then delete them" API calls as well. This could be done here, or you could file a follow-up issue.
2. Do all the other test modules already do cleanup on startup? If not they should be fixed too.

flocker/acceptance/test_api.py
```diff
@@ -32,13 +32,25 @@ def verify_socket(host, port):
...
+class APITestCase(TestCase):
+    """
+    Base test case that cleans the cluster state before each test.
+    """
+    @require_flocker_cli
+    def setUp(self):
+        """
+        Reset the cluster to an empty deployment.
+        """
+        return get_nodes(self, num_nodes=2)
```
itamarst: The fact `get_nodes` cleans up nodes is not very clear... maybe rename `get_nodes` to `get_clean_nodes`?

itamarst:
Address comments and merge when buildbot is happy.

dwgebler:
I am resubmitting this instead of merging; the changes are pretty substantial.

flocker/acceptance/test_api.py
```diff
@@ -340,14 +339,16 @@ class DatasetAPITests(TestCase):
...
+    def test_dataset_creation(self, cluster):
         """
         A dataset can be created on a specific node.
         """
         return create_dataset(self)
 
+    @require_cluster(2)
     @require_moving_backend
-    def test_dataset_move(self):
+    def test_dataset_move(self, cluster):
```
exarkun: Now that `cluster` is passed here, does the test need to start with a call to `get_test_cluster`?  It seems redundant.

flocker/acceptance/test_api.py
```diff
@@ -340,14 +339,16 @@ class DatasetAPITests(TestCase):
     """
     Tests for the dataset API.
     """
-    def test_dataset_creation(self):
+    @require_cluster(1)
+    def test_dataset_creation(self, cluster):
         """
         A dataset can be created on a specific node.
         """
         return create_dataset(self)
```
exarkun: `create_dataset` calls `get_test_cluster` too.  Maybe the dataset creation part of `create_dataset` should be factored into a function that takes a cluster as an argument and then all of the tests that have a `cluster` thanks to `require_cluster` can just pass it in to that new function.

flocker/acceptance/testtools.py
```diff
@@ -324,32 +325,66 @@ def get_nodes(test_case, num_nodes):
...
+
+        def check_datasets(cluster):
+            d = loop_until(lambda: no_datasets(cluster))
+            d.addCallback(lambda _: cluster)
+            return d
+
+        get_datasets.addCallback(check_datasets)
+        return get_datasets
+
+    getting.addCallback(clean_datasets)
```
exarkun: Same pattern repeated here for datasets and containers.  This could be avoided with a little helper that takes `cluster.configured_{datasets,containers}`, `cluster.{delete_dataset,remove_container}`, `cluster.{dataset_state,current_containers}` (particularly if the weird tuple return type of `current_containers` gets fixed).

This probably makes sense as a top-level function in the module, too, rather than a nested helper.

exarkun:

adamtheturtle:
Some of the test failures seem relevant, e.g. http://build.clusterhq.com/builders/flocker%2Facceptance%2Fvagrant%2Ffedora-20%2Fzfs/builds/2399/steps/trial/logs/stdio:

```
[ERROR]
  File "/home/buildslave/fedora-vagrant/flocker-acceptance-vagrant-fedora-20-zfs/venv/lib/python2.7/site-packages/flocker/acceptance/testtools.py", line 482, in got_results
    cluster, existing_containers = results
exceptions.ValueError: need more than 0 values to unpack

flocker.acceptance.test_deployment.DeploymentTests.test_deploy
flocker.acceptance.test_ports.PortsTests.test_deployment_with_ports
flocker.acceptance.test_postgres.PostgresTests.test_deploy
flocker.acceptance.test_postgres.PostgresTests.test_moving_postgres
[ERROR]
  File "/home/buildslave/fedora-vagrant/flocker-acceptance-vagrant-fedora-20-zfs/venv/lib/python2.7/site-packages/flocker/acceptance/testtools.py", line 482, in got_results
    cluster, existing_containers = results
exceptions.ValueError: need more than 1 value to unpack

flocker.acceptance.test_moving_applications.MovingApplicationTests.test_moving_application
```

dwgebler:
> Some of the test failures seem relevant, e.g.

Oh yeah, my bad, I know what that is. I'll push another commit shortly to fix.

adamtheturtle:
It doesn't seem that the helper suggested at https://github.com/ClusterHQ/flocker/pull/1425#discussion_r30739672 has been created.

adamtheturtle:
Thanks, address and resubmit - if it is me reviewing this I'll be looking only for passing tests and the helper @exarkun suggested.

dwgebler:
> It doesn't seem that the helper suggested at #1425 (comment) has been created.

The dataset cleaning code checks for whether or not a configured dataset is deleted, so a generic helper that works on both containers and datasets isn't really going to work out. No such flag exists on configured containers and if you have to explicitly check which you're dealing with, the helper is a bit pointless, so that's why I'm leaving that.

adamtheturtle:
Perhaps you could use a `.get` with a default to work around that: https://docs.python.org/2/library/stdtypes.html#dict.get. Or share the loops for waiting until there are 0 containers / datasets.

PR #1426, opened 2015-05-19
Dispatch backend floc 1791

adamtheturtle:
This should be Just Enough (TM) that other issues can be started which builds upon this. Other subtasks of the parent issue are linked here and will make this nice.


adamtheturtle:
I have submitted this without waiting for any builds to finish (the acceptance tests are important here) because I know other people are blocked on this.

flocker/node/script.py
```diff
@@ -320,18 +302,84 @@ def flocker_dataset_agent_main():
...
+    :return: A callable which can be called with a node UUID and hostname to
+        create a dataset deployer.
+    """
+    backend_to_deployer_factory = {
+        'zfs': zfs_dataset_deployer,
+        'loopback': loopback_dataset_deployer,
+    }
+    backend = dataset_configuration['backend']
+    deployer_factory = backend_to_deployer_factory[backend]
+    return deployer_factory(
```
itamarst: I think it's important you pass through the configuration (or a sub-key of the configuration); otherwise this doesn't really help backends that need further configuration.

itamarst:
Seems like a good start, though of course with lots more to do. Address and merge (and if I get 1783 merged before you do you'll need to merge forward).

flocker/node/script.py
adamtheturtle: Further configuration and everything necessary for it will happen in https://clusterhq.atlassian.net/browse/FLOC-1925 so I'll merge without this and get working on that.

adamtheturtle:
I think all of the failures are unrelated and I have mentioned the follow up issue which covers the only concern, so merging.

PR #1427, opened 2015-05-19
[FLOC-1914] Switch from GB to GiB.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1914

Adjust to the reality that OpenStack Cinder APIs operate in terms of GiB, not GB (as documentation suggests).

Here's a completely green run of the OpenStack functional tests on Rackspace (from this branch): http://build.clusterhq.com/builders/flocker%2Ffunctional%2Frackspace%2Fcentos-7%2Fstorage-driver/builds/54


myechuri:
Looks good, thank you for making this change!

PR #1429, opened 2015-05-19
Consume FLOCKER_FUNCTIONAL_TEST env variable from block device tests [FLOC-1931]

myechuri:


flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -245,11 +245,20 @@ def get_blockdeviceapi_with_cleanup(test_case, provider):
     :param provider: A provider type the ``IBlockDeviceAPI`` is to be
         compatible with.  A value from ``ProviderType``.
 
-    :raises: ``SkipTest`` if a ``FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE``
+    :raises: ``SkipTest`` if:
```
exarkun: maybe add "either" here so it's clear either condition is sufficient for a skip

PR #1430, opened 2015-05-20
Acceptance keep floc 1935

adamtheturtle:


admin/acceptance.py
```diff
@@ -449,11 +449,14 @@ def main(reactor, args, base_path, top_level):
             make_dispatcher(reactor),
             configure_cluster(control_node=nodes[0], agent_nodes=nodes,
                               certificates=certificates))
+
+        control_node = nodes[0]
```
exarkun: How do we know the control service is running on this node and not another one?

exarkun: Or is this prescriptive rather than descriptive?

adamtheturtle: Just above this:

```python
configure_cluster(control_node=nodes[0], agent_nodes=nodes, ...
```

I think that this variable should be moved above that and used there too.

exarkun: Ah.  `nodes[0]` is passed to `configure_cluster` two lines above.  Please move this assignment up and use `control_nodes` as the value to pass to `configure_cluster` to make it more clear we're setting it up this way.

exarkun: :grin:

exarkun:
Thanks.  Address that comment and then merge.

PR #1431, opened 2015-05-20
[FLOC-1607] Encode subjectAltName correctly when given hostname is actually an IP

itamarst:
And add dependency on `service_identity`.

### DIFF
--- a/flocker/ca/test/test_ca.py
+++ b/flocker/ca/test/test_ca.py
@@ -352,13 +352,23 @@ class ControlCredentialTests(
         self.assertEqual(
             subject.CN, b"control-service")
 
-    def test_subjectAltName(self):
+    def test_subjectAltName_dns(self):
         """
         The generated certificate has a subjectAltName containing the given
-        hostname.
+        hostname as a DNS record.
         """
         assert_has_extension(self, self.credential.credential,
                              b"subjectAltName", b"DNS:control.example.com")
+
+    def test_subjectAltName_ipv4(self):
+        """
+        The generated certificate has a subjectAltName containing the given
+        IPv4 address as a IP record.
+        """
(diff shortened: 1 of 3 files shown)


flocker/ca/test/test_ca.py
```diff
@@ -352,14 +352,24 @@ def test_certificate_subject_control_service(self):
...
         The generated certificate has a subjectAltName containing the given
-        hostname.
+        hostname as a DNS record.
         """
         assert_has_extension(self, self.credential.credential,
                              b"subjectAltName", b"DNS:control.example.com")
 
+    def test_subjectAltName_ipv4(self):
+        """
+        The generated certificate has a subjectAltName containing the given
```
adamtheturtle: Reading the test docstring for this and `test_subjectAltName_ipv4`, they can't both be true. Put "if ..." at the end of each.

setup.py
```diff
@@ -90,6 +90,10 @@
     "pytz",
     "characteristic >= 14.1.0",
     "Twisted == 15.1.0",
+    # TLS support libraries for Twisted:
+    "service_identity == 14.0.0",
```
adamtheturtle: Is it more important with these dependencies than others that they are kept up to date (just because they relate to security)?

adamtheturtle:
Thanks @itamarst, I'm accepting this with the assumption that with just the additional dependencies and without the additional code, Buildbot fails (like my local machine failed yesterday).

setup.py
itamarst: In general they ought to be kept up to date, yes, although we don't actually rely on service_identity or idna for security.

itamarst:
It does fail, yeah. I also filed an issue for adding a test for the underlying use case that this would break (3rd party HTTPS clients).

PR #1433, opened 2015-05-21
[FLOC-1943] Validate openstack credentials

adamtheturtle:
Design for: https://clusterhq.atlassian.net/browse/FLOC-1943

We plan to support a third dataset backend option, 'openstack' (after 'zfs' and 'loopback').
This requires:
* Documentation about what is supported,
* Documentation about how to create a valid agent.yml,
* Validation of the currently supported options.

Openstack backends require an auth plugin. For each auth plugin, there are various plugin-specific options.
We currently support two plugins: "rackspace" and "password".
The "password" plugin is used (at least) for PistonCloud.

Supported configs would be something like the following (but `yaml`):

```python
config_rackspace = {
    'control-service': {
        'hostname': '172.16.255.240',
        'port': 4524,
    },
    'dataset': {
        'backend': 'openstack',
        'tenant_id': 'my_id', # required if openstack
        'auth_url': 'https://example.com/rackspace/', # required if openstack, must be a uri
        'auth_plugin': 'rackspace': # required if openstack
        'username': 'adam', # required if rackspace plugin
        'region': 'my_region', # required if rackspace plugin
        'api_key': 'my_username', # required if rackspace
    },
    'version': 1
}

config_piston = {
    'control-service': {
        'hostname': '172.16.255.240',
        'port': 4524,
    },
    'dataset': {
        'backend': 'openstack',
        'tenant_id': 'my_id', # required if openstack
        'auth_url': 'https://example.com/piston/', # required if openstack, must be a uri
        'auth_plugin': 'password', # required if openstack
        'username': 'my_username', # required if 'password' plugin
        'password': 'my_password', # required if 'password' plugin
    },
    'version': 1
}
```

### DIFF
--- a/docs/indepth/installation.rst
+++ b/docs/indepth/installation.rst
@@ -395,6 +395,9 @@ For more details on configuring the firewall, see Fedora's `FirewallD documentat
 
 On AWS, an external firewall is used instead, which will need to be configured similarly.
 
+.. This (and the Ubuntu section below) should link to a new section about starting agents on a node.
+That new section can have subsections for each supported backend.
+
 To start the agents on a node, a configuration file must exist on the node at ``/etc/flocker/agent.yml``.
 This should be as follows, replacing ``${CONTROL_NODE}`` with the address of the control node.
 The optional ``port`` variable is the port on the control node to connect to:
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -131,6 +131,9 @@ def validate_configuration(configuration):
             "dataset": {
                 "type": "object",
                 "oneOf": [
+                    # Add further "oneOf" option for openstack, which itself
+                    # has "oneOf" options for each auth plugin.
+                    # We may use references to split this part of the schema up.
                     {
                         "required": ["backend"],
                         "properties": {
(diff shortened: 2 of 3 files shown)


exarkun:
Turns out we don't support PistonCloud.  We support "a Red Hat thing".

itamarst:
This design is problematic. We're going to have dynamically pluggable storage backends, which means we can't have one monolithic validation function for them. I suggest just loading config and passing it to backend, and it's backend responsibility to validate. Or we could have a validation hook as part of the plugin interface. but we really can't have one giant schema, even if it's split up across files. This schema needs to just have wildcard for the backend stuff.

exarkun:
> I suggest just loading config and passing it to backend, and it's backend responsibility to validate.

Just as easy (or easier, since we have four backends) is to ask the backend for a `config_schema` attribute and do the validation for it.

The suggestion about the wildcard in the overall schema sounds good to me.

adamtheturtle:
Ok, ignoring the proposed implementation, what do you think of the proposed supported configs?

adamtheturtle:
^^ @exarkun or @itamarst

PR #1434, opened 2015-05-21
[FLOC-1794] New introduction documentation layout

davidoliverSP2:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1794

This doc re-structure should be quite basic as there is currently only one topic for "Introduction to Flocker". A new folder called /introduction has been created, and the existing introduction.rst topic has been moved into it, and renamed index.rst. Additionally, the /images folder has also been moved into /introduction.

When new content for the "Introduction to Flocker" section is written (for example, "What's new in this release") it will go into the /introduction folder.

Buildbot has a few failures, which I reviewed with @jongiddy - please have a look to ensure these are known failures, and not caused by this branch.


wallrj:
@davidoliverSP2 you've explained that the reason for moving the images folder was that sphinx didn't seem able to resolve relative image links. I think there must be a way.
But I also see that the gettingstarted guide has an images sub directory. So maybe that's our preferred layout.
Anyway, the change seems fine.
There don't seem to be any (new) broken links.
The new introduction/index page renders exactly the same as it did before.

I'll merge the branch.

PR #1435, opened 2015-05-21
[FLOC-1895] Flocker dataset agent floc 1895

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1895


itamarst:
Thanks!

PR #1436, opened 2015-05-21
[FLOC-1874] Create block devices with interval sizes

wallrj:
Design for: https://clusterhq.atlassian.net/browse/FLOC-1874

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -777,6 +777,12 @@ class CreateBlockDeviceDataset(PRecord):
         :returns: An already fired ``Deferred`` with result ``None``.
         """
         api = deployer.block_device_api
+
+        # FLOC-1874 - Round maximum_size up to the nearest
+        # api.allocation_unit().  It might be better to do this in
+        # BlockDeviceDeployer but it's a ton easier to do it here (it's hard to
+        # do it there because calculate_changes shouldn't really make calls
+        # onto the api object).
         volume = api.create_volume(
             dataset_id=UUID(self.dataset.dataset_id),
             size=self.dataset.maximum_size,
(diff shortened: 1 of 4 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -1490,6 +1517,12 @@ def discover_state(self, node_state):
             NodeState(
                 uuid=self.node_uuid,
                 hostname=self.hostname,
+                # FLOC-1874 The sizes of the datasets for these manifestations
+                # will be the size of the actual storage-driver-supplied
+                # volume.  This may not be the size requested in the
+                # configuration.  File a follow-up issue for fixing this if it
+                # matters.  Fixing this requires having access to the
```
itamarst: Insofar as:

1. `calculate_changes` knows about fact that underlying storage size may not quite match, and
2. Lying about underlying storage size is quite difficult in some backends,

I think reporting accurately is the thing to do.

itamarst: I.e. I don't think there's need for follow-up.

itamarst:
Sounds good, please proceed.

itamarst:
You should also replace `REALISTIC_BLOCK_SIZE` or whatever it is in tests with a dynamic value based on `allocation_unit()`.

flocker/node/agents/blockdevice.py
exarkun: At minimum, expanding our model to have "this is how much the storage is" and "this is the requested size that we converged on" seems like a good thing.  It will be rather confusing to users when they ask for 7GiB and they get 8GiB and the system doesn't try to converge further.

itamarst: Copying stuff from config is kinda bogus. How about including "minimum unit of allocation" in state output?

PR #1437, opened 2015-05-21
[FLOC-1947] Log from run-acceptance-tests.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1947


itamarst:
That last commit was supposed to be on another branch, sorry. Doesn't do any harm though, and don't want to revert it so it doesn't cause merge problems later.

admin/acceptance.py
```diff
@@ -456,6 +469,16 @@ def main(reactor, args, base_path, top_level):
 
     runner = options.runner
 
+    from flocker.common.script import eliot_logging_service
+    log_file = open("%s.log" % base_path.basename(), "a")
```
itamarst: I'd rather have the name hard-coded rather than unexpectedly changing when someone renames the file, given that Buildbot is hardcodng it.

flocker/common/script.py
```diff
@@ -78,17 +87,18 @@ def main(reactor, options):
...
-class EliotObserver(object):
+class EliotObserver(Service):
     """
     A Twisted log observer that logs to Eliot.
     """
-    def __init__(self, publisher=twisted_log):
+    def __init__(self, publisher=twisted_log, set_stdout=True):
         """
         :param publisher: A ``LogPublisher`` to capture logs from, or if no
             argument is given the default Twisted log system.
```
itamarst: Document `set_stdout`.

itamarst:
Address minor comments, merge when buildbot is happy.

admin/acceptance.py
tomprince: Buildbot is also hard-coding the binary.

PR #1439, opened 2015-05-22
[FLOC-1949] Choose dataset backend in acceptance test runner.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1949

This will need to be refined to support cloud backends (passing credentials).

I also wonder if having `DatasetBackend` is worthwhile, if we expect people writing out-of-tree backends to use our acceptance tests.

There are some failures in the postgres tests that I haven't investigated yet.

This (temporarily) removes the ZFS acceptance tests, until after https://github.com/ClusterHQ/build.clusterhq.com/pull/91 lands

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -308,7 +308,7 @@ class RunOptions(Options):
          'The target provider to test against. '
          'One of {}.'.format(', '.join(PROVIDERS))],
         ['dataset-backend', None, 'loopback',
-         'The dataset backend to testt against. '
+         'The dataset backend to test against. '
          'One of {}'.format(', '.join(backend.name for backend
                                       in DatasetBackend.iterconstants()))],
         ['config-file', None, None,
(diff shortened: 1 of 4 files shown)


admin/acceptance.py
```diff
@@ -306,6 +307,10 @@ class RunOptions(Options):
         ['provider', None, 'vagrant',
          'The target provider to test against. '
          'One of {}.'.format(', '.join(PROVIDERS))],
+        ['dataset-backend', None, 'loopback',
+         'The dataset backend to testt against. '
```
jongiddy: testt -> test

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -63,6 +67,7 @@ Vagrant
 -------
 
 A configuration file is not required for the vagrant provider.
+The default dataset backend is the ZFS backend.
```
jongiddy: It looks like the default is `loopback` in `acceptance.py`

jongiddy:
To avoid acceptance tests failing in current Buildbot, please change the default to ZFS. I will reject for now, and wait for (at least some) passing acceptance tests.

PR #1443, opened 2015-05-22
[FLOC-1960] Don't explode with a TypeError in BlockDeviceDeployer.discover_state

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1960

Avoids producing bogus values from `LoopbackBlockDeviceAPI.get_device_path` and adds handling and logging for bogus values to `BlockDeviceDeployer.discover_state`.


flocker/node/agents/blockdevice.py
```diff
@@ -1339,7 +1364,13 @@ def get_device_path(self, blockdevice_id):
              volume.blockdevice_id.encode("ascii")]
         )
         # May be None if the file hasn't been used for a loop device.
-        return _device_for_path(volume_path)
+        path = _device_for_path(volume_path)
+        if path is None:
+            # It was supposed to be attached but it has no loopback device.  So
```
itamarst: Shouldn't calling `get_device_path` on some random block id that has not device be a bug in the caller? In which case this should raise exception.

itamarst: Unless... race conditions I guess? Something was listed, but then detached before get_device_path was called?

itamarst: Even if the general case of "I can't get_device_path for this path" can happen, i still don't see why loopback implemetnation should pretend this is valid thing to happen, rather than raising exception.

itamarst:
Since I don't understand the motivation for not just raising an exception, please explain or change, then resubmit.

itamarst:
And since new pyrsistent sounded like it was useful in debugging, please add it here or file an issue.

exarkun:
Expanded the comment above that section of loopback code so that, hopefully, it justifies the re-attach behavior more clearly.  Regarding pyrsistent upgrade, https://clusterhq.atlassian.net/browse/FLOC-1959 (already done).

flocker/node/agents/blockdevice.py
```diff
@@ -1441,21 +1476,32 @@ def discover_state(self, node_state):
...
         manifestations = {}
         nonmanifest = {}
-        devices = {
-            volume.dataset_id: api.get_device_path(volume.blockdevice_id)
-            for volume
-            in volumes
-            if volume.attached_to == compute_instance_id
-        }
 
+        def check_device_path(dataset_id, path):
```
itamarst: Maybe `is_existing_block_device` would be clearer?

itamarst:
Thanks, one minor comment then merge.

PR #1445, opened 2015-05-22
[FLOC-1963] Fix bug where we mutated shared state, causing segfaults

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1963


flocker/ca/test/test_validation.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+    def test_amp_new_context_each_time(self):
+        """
+        Each call to the AMP server's context factory ``getContext`` returns a
+        new instance, to prevent issues with global shared state.
+        """
+        ca_set, _ = get_credential_sets()
+        context_factory = amp_server_context_factory(
+            ca_set.root.credential.certificate, ca_set.control)
+        self.assertNotIdentical(context_factory.getContext(),
```
exarkun: I think `assertNotIs`

flocker/ca/test/test_validation.py
```diff
@@ -0,0 +1,39 @@
...
+    def test_rest_new_context_each_time(self):
+        """
+        Each call to the REST API server's context factory ``getContext``
+        returns a new instance, to prevent issues with global shared
+        state.
+        """
+        ca_set, _ = get_credential_sets()
+        context_factory = rest_api_context_factory(
+            ca_set.root.credential.certificate, ca_set.control)
+        self.assertNotIdentical(context_factory.getContext(),
```
exarkun: and here

exarkun:
Thanks. Looks pretty good.

There **could** be a direct unit test for the observed problem.  Get the context, create an in-memory connection with it, start the handshake, get the context again (mutating the verify callback), finish the  handshake.

That's probably a bunch more work, though, and removing the sharing seems like it gets rid of just about all of the possibility for the underlying problem to exist.  And we will continue to have coverage (albeit less directly, harder to debug coverage) for this in the acceptance test suite (or something like it).

Please address inline comments and merge when buildbot is green.

PR #1446, opened 2015-05-23
[FLOC-1729] Ensure the container has reached State/Running in DockerClient.add

runcom:
https://clusterhq.atlassian.net/browse/FLOC-1729

This is my first contribution here so sorry in advance for any mistake :)
I opened this to fix that nasty problem where docker container start api returns before the container is actually running (we know this from docker/docker project itself)
I plan to add a regression test here as soon as this patch is ok for you

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1446)

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -461,6 +461,11 @@ class DockerClient(object):
                 sleep(0.001)
                 continue
             self._client.start(container_name)
+            # Ensure the container has reached State/Running True before going on.
+            # see https://clusterhq.atlassian.net/browse/FLOC-1729
+            while not self._blocking_container_runs(container_name):
+                sleep(0.001)
+
         d = deferToThread(_add)
 
         def _extract_error(failure):
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -160,6 +160,20 @@ class GenericDockerClientTests(TestCase):
         name = random_name(self)
         return self.start_container(name)
 
+    def test_add_starts_container_ensure_running(self):
+        """
+        ``DockerClient.add`` starts the container and ensure it's in running state
+        """
+        name = random_name(self)
+        d = self.start_container(name)
+
+        def started_and_running(_):
+            docker = Client()
+            data = docker.inspect_container(self.namespacing_prefix + name)
+            self.assertTrue(data[u"State"][u"Running"])
+        d.addCallback(started_and_running)
+        return d
+
     def test_correct_image_used(self):
         """
         ``DockerClient.add`` creates a container with the specified image.


lukemarsden:
Hey @runcom - thanks so much for tackling this issue!

Could you please try adding unit and integration test coverage for the change? Then one of our engineers will be able to review the patch. If you need help with this, feel free to ask for help in `#clusterhq` on freenode (I see you've already joined the channel). :smile:

runcom:
@lukemarsden sure I'm gonna add tests asap! (and yes I joined #clusterhq to seek some guidance for my first PR here :))

runcom:
@lukemarsden I've added an integration test that ensure after add the container is in Running state. Not sure about doing a unit one but need help on it anyway

itamarst:
Hi, thanks for the patch!

What is the motivation for this? I.e. what problems might you or other users have if we don't make this change and leave things as is?

runcom:
@itamarst I don't have the full sight about issues coming from returning a false "started/running" state from DockerClient.add, I just think this have to be done to prevent race conditions (in docker integration tests for instance we have this problem and we always polling docke inspect for State Runinng True)

lukemarsden:
Thanks for adding the test! Does the test fail before the code change and succeed afterwards? I suspect that because it's solving a race condition, we might need to figure out a way to make a container startup take a long time to ensure that the test is really solving the problem. Maybe starting a container with an anonymous volume with a large file in it would help.

Btw, I think you have a typo `AssertEqual` => `assertEqual`. Have you been able to run the tests? Looks like http://doc-dev.clusterhq.com/gettinginvolved/contributing.html?highlight=running%20tests#running-tests should help with this in case you haven't yet.

runcom:
@lukemarsden yup I see my typo now that I'm doing what you are suggesting to expose the race in the test (sorry but I'm kinda new to python :D)

lukemarsden:
No problem, thanks!

BTW @itamarst, https://clusterhq.atlassian.net/browse/FLOC-1729 was raised previously by Jon, I think that's probably sufficient justification for accepting this change? WDYT?

runcom:
@lukemarsden I can repro the slow transition to State Running True with this Dockerfile:
```dockerfile
FROM busybox
VOLUME /test
```
after building the image I do the following commands:
```
fallocate -l 10G bigfile
docker create -v /home/antonio/test/bigfile:/test/bigfile testimage top
```
then just looking at container inspect you'll see the slow transition to State Running True even if starting the container returns immediatly

So now, do you have any idea how can I translate this to a test in your functional test suite? :) (seeking help sigh sigh)

exarkun:
Hi runcom,

Thanks for your efforts on this.

Making the container startup take a long time is one approach to making sure the test fails without your fix.  However, there are a couple drawbacks to this approach:

  * It's still non-deterministic.  The outcome will depend on how the kernel decides to schedule different processes on the system.  The test may spuriously pass even without your fix.  This can be alleviated by making sure the container startup really does take a **long** time, but ...
  * Having a slow test suite kind of sucks.  So it's hard to get excited about the prospect of intentionally introducing slow tests. :smile: 

I wonder what other approaches might be possible either for testing this or implementing the fix for it.

One idea that comes to mind (which may be totally bogus, please don't take the fact that I'm mentioning it to mean that I am confident it is the best approach or even that it will work at all, it just seems like an idea worth exploring) is that the Docker event stream might tell us exactly when the container has started and that the event stream may be a Docker interface that is more easily faked than the inspect interface.

I would explore the following:

  * Does the API described at https://docs.docker.com/reference/api/docker_remote_api_v1.18/#monitor-dockers-events emit an event that corresponds to the state change that matters to the code being monitored?  It looks like there is a `start` event that might be just the right thing (but the documentation isn't very detailed and I don't see information about what each event actually corresponds to; it seems the reader is meant to make inferences based on the names of events or else do further research elsewhere - the source code, perhaps?).
  * If so, does `docker.Client` give us access to the event stream in a usable way?
  * If so, can we have a fake of `docker.Client` that can handle the operations required by `add` (the Flocker method being modified), including producing a realistic event stream for the startup?  (Importantly, can we reasonably test that this fake behaves just like the real `docker.Client` in this situation?)
  * If so, then we can probably give that fake the ability to delay delivering the `start` event.  This means we can write a test that uses the fake, delays the `start` event, makes an assertion that the `Deferred` returned by `add` *has no result*, delivers the `start` event, and finally makes an assertion that the `Deferred` now does have the correct result.  This proves that `add` waits exactly long enough to provide the behavior that Flocker is interested in, it does so without introducing an arbitrarily long delay into the test suite, and it does so in a way that deterministically tests the behavior (instead of relying on kernel process scheduling and disk I/O load and so forth).

I know that's a big wall of text I just threw at you.  Considering what Flocker *already* does elsewhere in our interactions with `docker.Client`, you might wonder why we should go to all this extra trouble.  The answer is that all of those existing interactions are similarly problematic and it would be wonderful if we could eventually replace them with something more like what I've described here.  The extra effort of doing things this way is exactly the reason we took the shortcut of just introducing some sleep loops without rigorous testing - but that just means we have the expedient solution rather than the correct one. :smile: 

Flocker is still finding its balance regarding expediency versus correctness (particularly in our current run up to 1.0) and regarding the standards we hold contributions to - from inside or outside the company.  What I *hope* is that you'll read this and think it sounds like an interesting testing challenge and dive right in to it. :smile:   However, if you think I'm crazy (or confusing or even wrong) then I'd be delighted to discuss this or any other ideas for resolving this issue with you.  So please consider this just another part of the discussion, not an iron mandate for how to proceed.

Thanks again for your interest and your efforts so far!

runcom:
@exarkun really many thanks for this brainstorm!
My really first approach to this test was indeed to fake the docker client and mock api calls to delay so I'm able to mock this behavior.
I'm aware of docker events streaming the start event but I'm pretty sure it suffers the same problem (meaning it will emit start event *before* the state is running). I'm saying this becouse it's a known bug in Docker and we're trying hard to fix this (for instance, a container is made Runnning even before tcp port are forwarded completely, meaning there's a race here too) :/

itamarst:
Hi,

We will be phasing out the mode in which Flocker controls Docker just as soon as our integrations with other tools are feature-compatible. E.g. we're working on a Docker volumes plugin (https://docs.clusterhq.com/en/1.0.3/labs/index.html#flocker-docker-plugin-with-cli-and-gui) where Docker calls Flocker, instead of vice versa.

As such we're not focusing much more effort on the code that calls Docker directly, so I'm going to close this and the issue it refers to. Thanks for submitting it though, we do appreciate your submission. If you email thanks@clusterhq.com we'd be happy to send you a t-shirt.

PR #1448, opened 2015-05-25
[FLOC-1946] Allocate BlockDeviceVolumes with size intervals

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1946

I had to introduce two new things:
 * ``IBlockDeviceAPI.allocation_unit`` which is the interval impose by the AWS or Cinder API (both 1GiB), this allows us to get rid of all the rounding in the tests.
 * ``over_allocation`` size which is the amount that the redhat-openstack Cinder storage drivers over allocate. This is only used in the tests to verify that the devices that get attached to a host have a predictable size.

I originally assumed that the redhat-openstack platform was doing rounding up the requested size to some storage specific allocation_unit but it appears to just add 4GiB more than the requested GiB requested size.
I don't know why it does that.

I modified the LoopBackBlockDeviceAPI to encode the ``allocated_size`` into the backing file name. This isn't strictly necessary; I made the change while I was thinking of allocation_unit as something imposed by the Cinder backend storage driver rather than the Cinder API.
But having done it I could now modify LoopBackBlockDeviceAPI to simulate the "redhat-openstack" situation where a requested size of `1.5 GiB` is rounded up by the API to `2GiB` and then further rounded up by the Cinder storage driver to `4GiB`.

I haven't removed the resize tests yet because it seems like it'd be fairly straight forward to implement resize for the Loopback API....in fact I was able to get the grow test passing fairly easily and think the shrink test just needs a slightly larger margin of error in the test.

Of particular interest are the test results for builders which are currently marked as expected failures:
 * http://build.clusterhq.com/boxes-flocker?branch=blockdevice-size-intervals-FLOC-1946&num_builds=10

With this branch, some storage-driver builders should all pass.


itamarst:
There are failures in the resize tests. I suggest just deleting them - they'll be in git history, we can always revive them when necessary.

exarkun:
> There are failures in the resize tests. I suggest just deleting them - they'll be in git history, we can always revive them when necessary.

It'll be easier to just mark them as skip, I think.  Otherwise we end up with a half-deleted feature (because no one ever remembers to delete **all** the pieces).

flocker/node/agents/blockdevice.py
```diff
@@ -743,6 +743,25 @@ def run(self, deployer):
...
+def allocated_size(allocation_unit, requested_size):
+    """
+    Round ``requested_size`` up to the nearest ``allocation_unit``.
+
+    :param int allocation_unit: The interval in ``bytes`` to which
+        ``requested_size`` will be rounded up.
+    :param int requested_size: The size in ``bytes`` that is required.
+    :return: The ``allocated_size`` in ``bytes``.
+    """
+    previous_interval_size = (
```
itamarst: Maybe convert all inputs to bytes, just to be on the safe side?

flocker/node/agents/blockdevice.py
```diff
@@ -881,6 +912,16 @@ class IBlockDeviceAPI(Interface):
     Note: This is an early sketch of the interface and it'll be refined as we
     real blockdevice providers are implemented.
     """
+    def allocation_unit():
+        """
+        The size in bytes up to which the requested block device size will be
+        rounded by the underlying storage layer.  (XXX Don't imply we'll
```
itamarst: Still has XXX.

flocker/node/agents/blockdevice.py
```diff
@@ -1198,13 +1266,14 @@ def create_volume(self, dataset_id, size):
         See ``IBlockDeviceAPI.create_volume`` for parameter and return type
         documentation.
         """
+        actual_size = allocated_size(self.allocation_unit(), size)
```
itamarst: I don't think you should be doing rounding in loopback implementation. Instead, you should throw exception if requested size is not multiple of allocation unit.

At least some backends *cannot* do rounding, e.g. ScaleIO. We therefore want to ensure that we never call create_volume() with something that isn't multiple of `allocation_unit()`. So I think loopback should implement the more restrictive behaviour so it can catch bugs in the deployer.

flocker/node/agents/blockdevice.py
```diff
@@ -1286,18 +1357,22 @@ def resize_volume(self, blockdevice_id, size):
         This implementation is limited to being able to resize volumes only if
         they are unattached.
         """
-        backing_path = self._unattached_directory.child(
-            blockdevice_id.encode("ascii")
-        )
+        volume = get_blockdevice_volume(self, blockdevice_id)
+        filename = _backing_file_name(volume)
+        backing_path = self._unattached_directory.child(filename)
+        new_size = allocated_size(self.allocation_unit(), size)
```
itamarst: Same comment as above; I don't think loopback implemetnation should do rounding itself.

flocker/node/agents/blockdevice.py
```diff
@@ -1490,6 +1569,12 @@ def discover_state(self, node_state):
             NodeState(
                 uuid=self.node_uuid,
                 hostname=self.hostname,
+                # FLOC-1874 The sizes of the datasets for these manifestations
```
itamarst: This comment needs to be addressed, i.e. file a follow-up issue and link to it if you'd like. I don't think that this follow-up is necessary, personally - NodeState is reporting accurately.

flocker/node/agents/blockdevice.py
```diff
@@ -1585,6 +1670,11 @@ def calculate_changes(self, configuration, cluster_state):
             local_state.devices, local_state.paths, configured_manifestations,
         ))
         deletes = self._calculate_deletes(configured_manifestations)
+
+        # FLOC-1874 - The API doesn't support resizing.  Drop this calculation
```
itamarst: Another un-addressed comment. Probably resizing should be deleted (ideally in single commit so it's easy to resurrect) since there's no point trying to make failing tests pass.

flocker/node/agents/cinder.py
```diff
@@ -249,6 +249,22 @@ def __init__(self,
...
+        """
+        Return a 1GiB fixed allocation_unit for now; one which should
+        work on all currently supported and soon-to-be supported
+        platforms.
+
+        This is the minimum allocation unit described by the OpenStack
+        Cinder API documentation.
+         * http://developer.openstack.org/api-ref-blockstorage-v2.html#createVolume # noqa
+
+        XXX: This should be loaded from configuration because
```
itamarst: 1. XXX needs addressing, at minimum by filing follow up issue.
2. Why would that violate API?

flocker/node/agents/cinder.py
```diff
@@ -283,14 +299,9 @@ def create_volume(self, dataset_id, size):
         }
         action_type = u"blockdevice:cinder:create_volume"
         with start_action(action_type=action_type):
-            # There could be difference between user-requested and
-            # Cinder-created volume sizes due to several reasons:
-            # 1) Round off from converting user-supplied 'size' to 'GiB' int.
-            # 2) Cinder-specific size constraints.
-            # XXX: Address size mistach (see
-            # (https://clusterhq.atlassian.net/browse/FLOC-1874).
+            allocated_bytes = allocated_size(self.allocation_unit(), size)
```
itamarst: The deployer is doing this, so I don't think IBlockDeviceAPIs should be required to (or need to!).

flocker/node/agents/ebs.py
```diff
@@ -294,14 +306,9 @@ def create_volume(self, dataset_id, size):
         as volume tag data.
         Open issues: https://clusterhq.atlassian.net/browse/FLOC-1792
         """
-        # There could be difference between user-requested and
-        # created volume sizes due to several reasons:
-        # 1) Round off from converting user-supplied 'size' to 'GiB' int.
-        # 2) EBS-specific size constraints.
-        # XXX: Address size mistach (see
-        # (https://clusterhq.atlassian.net/browse/FLOC-1874).
+        allocated_bytes = allocated_size(self.allocation_unit(), size)
```
itamarst: Same comment as above; shouldn't be doing this here.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -441,10 +444,12 @@ def test_one_device(self):
         and the volume's filesystem is mounted in the right place.
         """
         dataset_id = uuid4()
+        requested_size = REALISTIC_BLOCKDEVICE_SIZE
```
itamarst: Given `allocation_unit()` I think `REALISTIC_BLOCKDEVICE_SIZE` should die, since it shouldn't have to try to support N backends, and be replaced with a dynamically calculated value based on `allocation_unit()`.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1119,6 +1124,7 @@ def test_detach_manifestation(self):
         )
 
 
+# FLOC-1874 - Skip or remove all these tests.
```
itamarst: Comment should be addressed. ELsewhere I suggested deleting this; fixing tests so they pass is the other alternative.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1398,22 +1421,43 @@ def test_attach_unknown_volume(self):
...
+        self._verify_volume_size(
+            requested_size=requested_size,
+            expected_volume_size=requested_size
         )
-        attached_volume = self.api.attach_volume(
-            new_volume.blockdevice_id, attach_to=self.this_node,
+
+    def test_size_round_up(self):
+        """
+        Validate that block devices created by the API, allocate size
```
itamarst: This puts the burden of rounding on `IBlockDeviceAPI`. I think that's wrong, it means duplicating functionality in all the backends. Not to mention that the deployer (or really, state changes) already does it, there's no need for `IBlockDeviceAPI` implementations to do any rounding.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2822,6 +2898,35 @@ def test_run_create(self):
 
         self.assertEqual(expected_volume, volume)
 
+    # FLOC-1874 A test like test_run_create that requires rounding to happen.
```
itamarst: Can delete this comment.

itamarst:
Thanks! Please address and resubmit.

flocker/node/agents/blockdevice.py
wallrj: Ok, I convert all inputs to ``int``. Is that what you meant?

flocker/node/agents/blockdevice.py
wallrj: Removed and improved.

flocker/node/agents/blockdevice.py
wallrj: Ok. I removed the rounding in LoopbackBlockDeviceAPI and raise a ``ValueError`` if the supplied size isn't divisble.
That broke a bunch of other tests, which I've now fixed by replacing REALISTIC_BLOCKDEVICE_SIZE with ``allocation_unit()``.
Also lowered the default ``LoopbackBlockDeviceAPI.allocation_unit()`` to 64MiB in line with our minimum API size which is based on ZFS contraints....I think.

itamarst:
Note that 64MiB is minimum *size*, not interval, for ZFS.

flocker/node/agents/blockdevice.py
wallrj: Done. And raise ValueError here too.

flocker/node/agents/blockdevice.py
wallrj: I removed the comment and added another noting that we don't report the size of non-manifest datasets.

wallrj: https://clusterhq.atlassian.net/browse/FLOC-1983

flocker/node/agents/blockdevice.py
wallrj: I won't delete it. The loopback tests should still pass. But I have explicitly skipped the resize tests for AWS (like we do for OpenStack).

flocker/node/agents/cinder.py
wallrj: This was written when I thought allocation_unit represented the allocation size of the Cinder driver and it seemed wrong that this should ever return different size  than that documented in the OpenStack API docs.

wallrj: I updated the comment.

flocker/node/agents/cinder.py
wallrj: Done

flocker/node/agents/ebs.py
wallrj: Done.

flocker/node/agents/test/test_blockdevice.py
wallrj: I don't think I can face changing them all in this branch...and you wouldn't want to review the changes. I'll create a followup issue if you agree.

flocker/node/agents/test/test_blockdevice.py
wallrj: I'll fix them.

flocker/node/agents/test/test_blockdevice.py
wallrj: I removed that test.

flocker/node/agents/test/test_blockdevice.py
wallrj: Done.

wallrj:
Ok. I think I addressed all the review comments.
Build results will be here:
 * http://build.clusterhq.com/boxes-flocker?branch=blockdevice-size-intervals-FLOC-1946

The branch has grown rather large.

I could revert the file naming changes to LoopBackBlockDeviceAPI. Those aren't really necessary except if we wanted to simulated APIs that create larger block devices than they report.

flocker/node/agents/blockdevice.py
```diff
@@ -1298,6 +1394,9 @@ def resize_volume(self, blockdevice_id, size):
                 backing_file.truncate(size)
             finally:
                 backing_file.close()
+        resized_volume = volume.set('size', size)
```
itamarst: Did the old implementation not have move logic?

itamarst: Oh, filename encodes size now, right.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1270,36 +1276,54 @@ class IBlockDeviceAPITestsMixin(object):
     """
     this_node = None
 
-    def _assert_volume_size(self, volume, size):
+    def _verify_volume_size(self, requested_size, expected_volume_size):
         """
-        Assert that volume size is consistent across EBS, OS, IBlockDeviceAPI.
+        Assert that all implementations of
```
itamarst: "all limplementations" is misleading, it's really one implementation at a time.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1341,26 +1365,26 @@ def test_created_is_listed(self):
...
     def test_listed_volume_attributes(self):
         """
-        ``list_volumes`` returns ``BlockDeviceVolume`` s that have the same
-        dataset_id and size as was passed to ``create_volume``.
+        ``list_volumes`` returns ``BlockDeviceVolume`` s that have the
+        same dataset_id and (maybe over-allocated size) as was passed
+        to ``create_volume``.
         """
         expected_dataset_id = uuid4()
+        requested_size = REALISTIC_BLOCKDEVICE_SIZE
```
itamarst: That's... confusing. Is there some expectation of the relationship between REALISTIC_BLOCKDEVICE_SIZE and allocation_unit()? seems like you can just use allocation_unit? E.g. int(allocation_unit() * 1.5) or something?

itamarst: Or actually, just use allocation_unit() here and new test for allocation where requested size is not multiple of allocation_unit.

itamarst: (Deleted some bogus comments).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2822,6 +2910,33 @@ def test_run_create(self):
...
+        """
+        ``CreateBlockDeviceDataset.run`` rounds up the size if the
+        requested size is less than ``allocation_unit``.
+        """
+        dataset_id = uuid4()
+        allocation_unit = int(MiB(10).to_Byte().value)
+
+        # A realistic size which is divisible by the allocation_unit
+        expected_size = (
+            allocation_unit * (REALISTIC_BLOCKDEVICE_SIZE // allocation_unit)
```
itamarst: Why not just "expected_size = 11"?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2822,6 +2910,33 @@ def test_run_create(self):
...
+        )
+        (volume,
+         device_path,
+         expected_mountpoint,
+         compute_instance_id) = self._create_blockdevice_dataset(
+            dataset_id=dataset_id,
+            # Request a size which will force over allocation.
+            maximum_size=expected_size - 1,
```
itamarst: And then you don't need "- 1" here either.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3274,3 +3390,132 @@ def test_shrink(self):
...
         """
         self._resize_test(0.5)
+
+
+class AllocatedSizeTypeTests(SynchronousTestCase):
+    """
+    Tests for type coercion of parameters supplied to
+    ``allocated_size``.
+    """
+    def test_allocation_unit_bytes(self):
```
itamarst: This really shouldn't be part of the interface. Floats are the only thing that is worth worrying about. So if it happens to work that's OK, but I don't think it's worth testing.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3274,3 +3390,132 @@ def test_shrink(self):
...
+            allocated_size(
+                allocation_unit=10,
+                requested_size=b"1",
+            ),
+            int,
+        )
+
+    def test_requested_size_float(self):
+        """
+        ``allocated_size`` raises ``TypeError`` if the supplied
```
itamarst: Docstring is wrong.

itamarst:
Looking good; address comments, make buildbot happy, merge.

flocker/node/agents/test/test_blockdevice.py
wallrj: done.

flocker/node/agents/test/test_blockdevice.py
wallrj: I can't use ``allocation_unit`` because of the minimum size requirement on Rackspace, but I can request ``allocated_size(allocation_unit, REALISTIC_....)` instead and add that extra test you suggest.
Then a followup issue to supply ``MINIMUM_VOLUME_SIZE`` to the test case factory function...#1406 in fact...and replace `REALISTIC_BLOCKDEVICE_SIZE` throughout.
(sorry @wallnerryan for not contributing to the conversation on that PR)

wallnerryan: no worries @wallrj don't know if its still useful or not

wallrj: I added a `minimum_allocatable_size` variable to the test case factory and set it to `1GiB` on AWS and OpenStack and `100GiB` on Rackspace based on the supplied CLOUD_PROVIDER environment variable.

flocker/node/agents/test/test_blockdevice.py
wallrj: I simplified it.

flocker/node/agents/test/test_blockdevice.py
wallrj: done.

flocker/node/agents/test/test_blockdevice.py
wallrj: Done.

PR #1450, opened 2015-05-26
[FLOC-1956] Require maximum_size when creating datasets 

wallrj:
Design / partial implementation for https://clusterhq.atlassian.net/browse/FLOC-1956

 * There are a bunch of existing tests that'll need updating.
 * We could add an `allocation_unit` validation step to the schema...which only accepts multiples of 1GiB
 * We could also set the minimum maximum_size to 100GiB to avoid any confusion for rackspace users, where the SATA minimum size is 100GiB.


itamarst:
We should not validate particular values for maximum size, given we've implemented rounding logic. We do need to validate that it's not `null`, which is I believe a valid value in the schema?

itamarst:
We should also not set minimum size to 100GiB. Lots of systems can do smaller.

exarkun:
> We should also not set minimum size to 100GiB. Lots of systems can do smaller.

Not much smaller.  I think Rackspace minimum size is 75GiB.  If we want to support sizes that are actually kind of smallish we need to parameterize this per-something.

itamarst:
Not certain, but seems like ScaleIO supports as small as 8 GiB.

itamarst:
and "You can create EBS Magnetic volumes from 1 GiB to 1 TiB in size"

exarkun:
> and "You can create EBS Magnetic volumes from 1 GiB to 1 TiB in size" (or more for SSD, but minimum is same AFAICT).

This is for ... ?

exarkun:
> > and "You can create EBS Magnetic volumes from 1 GiB to 1 TiB in size" (or more for SSD, but minimum is same AFAICT).
> This is for ... ?

Sorry, nevermind, reading comprehension failure.

The point is that a common minimum value has to work on all our platforms.  That includes Rackspace.

itamarst:
Richard suggested setting a minimum allowed size for datasets in the schema. I am pointing out that relying on rackspace minimums of 100 GiB is problematic for users of other backends.

exarkun:
> Richard suggested setting a minimum allowed size for datasets in the schema. I am pointing out that relying on rackspace minimums of 100 GiB is problematic for users of other backends.

Yes, and setting it lower than 100 GiB is problematic for users of Rackspace.

itamarst:
My guess is EBS has way more users than Rackspace...

exarkun:
> My guess is EBS has way more users than Rackspace...

So let's be explicit.  The suggestion here is "Break Rackspace so that EBS works better."  Then we can talk about whether that's the trade-off we want to make.

itamarst:
Or, hm. We could make the allocation logic in FLOC-1946 slightly smarter so it's not just about units of allocation but also rounding up to a minimum? But I guess that's probably enough work it won't make it into 1.0. Do you want to ask product?

exarkun:
Maybe rounding up to a minimum isn't really very hard?  Can we just tell `IBlockDeviceAPI` implementors they have to do this (and then do it in the Cinder implementation ... one extra `min(..., size)` call and one extra test that's probably mostly the same as some existing test)?

Or ... Can we tweak the schema on-the-fly when the backend is selected?

itamarst:
Hm. I had a bunch of comments on the allocation_unit branch about how I don't think we should duplicate logic in the storage backends if we can do it once in the deployer/statechange logic. Adding a `minimum_allocation_unit` and a `min()` in `CreateBlockDeviceVolume` doesn't seem like a lot of work, it's true.

wallrj:
We're going in a different direction. 
`BlockDeviceDeployer` will instead allocate a default size if maximum_size was null in the API request.

PR #1453, opened 2015-05-26
[FLOC-1984] Fix bug where pvector blew up docker-py

itamarst:
This will also be covered by end-to-end acceptance test (the reason I discovered it) in FLOC-1976. In master there are no acceptance tests using `command_line` which is why this wasn't caught.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -431,10 +431,16 @@ class DockerClient(object):
                 port_bindings=port_bindings,
                 restart_policy=restart_policy_dict,
             )
+            # We're likely to get e.g. pvector, so make sure we're passing
+            # in something JSON serializable:
+            command_line_values = command_line
+            if command_line_values is not None:
+                command_line_values = list(command_line_values)
+
             self._client.create_container(
                 name=container_name,
                 image=image_name,
-                command=command_line,
+                command=command_line_values,
                 environment=environment,
                 ports=[p.internal_port for p in ports],
                 mem_limit=mem_limit,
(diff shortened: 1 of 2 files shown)


flocker/node/_docker.py
```diff
@@ -431,10 +431,16 @@ def _create():
                 port_bindings=port_bindings,
                 restart_policy=restart_policy_dict,
             )
+            # We're likely to get e.g. pvector, so make sure we're passing
+            # in something JSON serializable:
```
adamtheturtle: Not just likely, but this is what the docstring says that this `ivar` will be.

:question: Explain why this must be something JSON serialisable - looking at `docker-py` it is because a JSON formatted `str` is used as a config file.

flocker/node/functional/test_docker.py
```diff
@@ -726,13 +728,15 @@ def test_command_line(self):
         name = random_name(self)
         d = self.start_container(
             name, image_name=u"busybox",
-            command_line=[u"sh", u"-c", u"""\
+            # Pass in pvector since this likely to be what caller actually
+            # passes in:
+            command_line=pvector([u"sh", u"-c", u"""\
```
adamtheturtle: :question: If I'm reading the code correctly, this must be "a sequence of ``unicode``, or ``None``" so sequence could be list etc. Either we should have tests for PVectors and lists or we should only support one I think.

adamtheturtle:
Thanks @itamarst address and merge if Buildbot is green or resubmit if there are significant changes.

adamtheturtle:
Actually, I'd prefer if you waited to merge this until there is an acceptance test which needs this about to be merged - the one in FLOC-1976 being accepted basically.

PR #1454, opened 2015-05-26
[FLOC-1950] EBS IBlockDeviceAPI logging

myechuri:
Add logging support for IBlockDeviceAPI calls to AWS.

Fixes: https://clusterhq.atlassian.net/browse/FLOC-1950


flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
```
wallrj: Add a short module docstring explaining what's in here.

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
...
+from eliot import Field, ActionType, MessageType
+
+# Begin: Helper datastructures to log all
+# IBlockDeviceAPI calls from AWS storage driver using Eliot.
+
+# An OPERATION is a list of:
+# IBlockDeviceAPI name, positional arguments, keyword arguments.
+OPERATION = Field.forTypes(
+    u"operation", [list],
```
wallrj: Is ``list`` a valid ``Field`` type?
If the supplied list might contain other complex types, then Eliot will need to know how to serialise those.
I'm thinking of a situation where a caller has supplied the wrong argument types to ``attach_volume``. In which case the list might be  `['attach_volume', BlockDeviceVolume()...]`. How will Eliot know how to serialise `BlockDeviceVolume` ? Does it just call `repr`? Or will it blow up when our `stdout` log observer attempts to serialise that object?
@itamarst will know.

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
...
+    u"The IBlockDeviceAPI command being executed, \
+    along with positional and keyword arguments.")
+
+# ActionType used by AWS storage driver.
+AWS_ACTION = ActionType(
+    u"flocker:node:agents:blockdevice:aws",
+    [OPERATION],
+    [],
+    u"An IBlockDeviceAPI command is executing using AWS storage driver.")
```
wallrj: :question: "command" > "method" ...maybe.

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
...
+AWS_MESSAGE = Field.for_types(
+    "aws_message", [bytes, unicode],
+    u"A human-readable error message given by the response.",
+)
+AWS_REQUEST_ID = Field.for_types(
+    "aws_request_id", [bytes, unicode],
+    u"The unique identifier assigned by the server for this request.",
+)
+
+# Log boto.exception.BotoEC2ResponseError, covering all errors from AWS:
```
wallrj: Does boto raise `BotoEC2ResponseError` for errors interacting with `EBS`?
Or is `BotoEBSResponseError` a thing?

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
...
+        AWS_MESSAGE,
+        AWS_REQUEST_ID,
+    ],
+)
+
+DEVICES = Field.forTypes(
+    u"devices", [list],
+    u"The list of devices currently in use by the compute instance.")
+NO_AVAILABLE_DEVICE = MessageType(
+    u"flocker:node:agents:blockdevice:aws:devicesearch",
```
wallrj: I think this label should contain `no_available_device` if that's the error condition that this `Message` indicates.

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,66 @@
...
+    [DEVICES],
+)
+
+SIZE = Field.forTypes(
+    u"size", [int],
+    u"The size, in bytes, of new device we are expecting to manifest \
```
wallrj: I'm not sure if it's written anywhere, but we don't use `\` for splitting long lines.
Use 
```
u"The size, in bytes, of new device we are expecting to manifest "
u"in the OS."
```
...instead.

flocker/node/agents/ebs.py
```diff
@@ -128,15 +106,18 @@ def _run_with_logging(self, *args, **kwargs):
...
+        # Trace IBlockDeviceAPI ``method`` as Eliot Action.
+        with AWS_ACTION(operation=[method_name, args, kwargs]):
+            try:
+                return method(*args, **kwargs)
+            except EC2ResponseError as e:
+                BOTO_EC2RESPONSE_ERROR(
+                    aws_code=e.code,
+                    aws_message=e.message,
+                    aws_request_id=e.request_id,
+                ).write()
```
wallrj: I have to admit that I still don't understand Eliot logging best practice but since the method is now run in the context of an Eliot Action, I think that Eliot will log the `AWS_ACTION` as failed and include the traceback automatically. So I don't think there's any need to write this Eliot Message too.
Right @itamarst ?

flocker/node/agents/ebs.py
```diff
@@ -300,6 +281,7 @@ def _wait_for_new_device(base, size, time_limit=60):
                 return new_device
         time.sleep(0.1)
         elapsed_time = time.time() - start_time
+    NO_NEW_DEVICE_IN_OS(devices=base, size=size, time_limit=time_limit).write()
```
wallrj: If a new device with the expected size does not appear within 60s then it'd be useful to log the actual list of devices and their sizes, I think. ie a new device has shown up, but with a slightly unexpected size.

flocker/node/agents/ebs.py
```diff
@@ -393,6 +375,10 @@ def _next_device(self, instance_id):
             file_name = u'/dev/sd' + suffix
             if file_name not in devices:
                 return file_name
+
+        # Could not find any suitable device that is available
+        # for attachment. Log to Eliot before giving up.
+        NO_AVAILABLE_DEVICE(devices=devices).write()
```
wallrj: I suppose this'll be redundant if / when we implement https://clusterhq.atlassian.net/browse/FLOC-1887 "Allow users to know when they run out of device names to attach EBS volume" and the suggested new ``CannotAttach`` exception proposed in that issue.

Perhaps we should just fix that issue instead.

However, the log message is useful in the meantime and it'd be even more useful (I think) if it included the blockdevice_id and the real allocated device path...the one found by _wait_for_new_device (above) which is different, right?

flocker/node/agents/ebs.py
wallrj: Yeah, here's an example of the resulting duplication:
 * http://build.clusterhq.com/builders/flocker%2Ffunctional%2Faws%2Fcentos-7%2Fstorage-driver/builds/72/steps/trial/logs/test.log
```
2015-05-27 00:24:26+0000 [-] ELIOT: {'aws_code': u'VolumeInUse',
	 'aws_message': u'Volume vol-a346f1b3 is currently attached to i-115ccfe6',
	 'aws_request_id': u'7c9da9e9-ed15-4c3a-878c-4f29a209cfc7',
	 u'message_type': u'boto:boto_ec2response_error',
	 u'task_level': [2],
	 u'task_uuid': u'0a1abfe9-7a66-43d6-949c-6c744fd3f1e0',
	 u'timestamp': 1432686266.464014}
2015-05-27 00:24:26+0000 [-] ELIOT: {u'action_status': u'failed',
	 u'action_type': u'flocker:node:agents:blockdevice:aws',
	 u'exception': u'boto.exception.EC2ResponseError',
	 u'reason': u'EC2ResponseError: 400 Bad Request\n<?xml version="1.0" encoding="UTF-8"?>\n<Response><Errors><Error><Code>VolumeInUse</Code><Message>Volume vol-a346f1b3 is currently attached to i-115ccfe6</Message></Error></Errors><RequestID>7c9da9e9-ed15-4c3a-878c-4f29a209cfc7</RequestID></Response>',
	 u'task_level': [3],
	 u'task_uuid': u'0a1abfe9-7a66-43d6-949c-6c744fd3f1e0',
	 u'timestamp': 1432686266.464498}
```

All the information logged in the message is already logged by the failed action...wrapped in XML tags...but the info is there already.

wallrj: It'd be interesting to see if the new log messages allow us to diagnose why that call to `delete_volume` fails.
`Volume vol-a346f1b3 is currently attached to i-115ccfe6`
Why is that volume still attached?
Is it because we haven't yet attempted to detach it?
No, I can see a "successful" detach operation logged prior to calling delete, so perhaps we need to make the detach method more robust.

```
2015-05-27 00:24:17+0000 [-] ELIOT: {u'action_status': u'started',
	 u'action_type': u'flocker:node:agents:blockdevice:aws',
	 'operation': ['detach_volume', (u'vol-a346f1b3',), {}],
	 u'task_level': [1],
	 u'task_uuid': u'3be1d055-b5a4-48b8-ac58-4326b8e79555',
	 u'timestamp': 1432686257.168922}
2015-05-27 00:24:17+0000 [-] ELIOT: {u'action_status': u'succeeded',
	 u'action_type': u'flocker:node:agents:blockdevice:aws',
	 u'task_level': [2],
	 u'task_uuid': u'3be1d055-b5a4-48b8-ac58-4326b8e79555',
	 u'timestamp': 1432686257.328639}

```

wallrj:
Thanks @myechuri 

The new logging looks useful and does seem to help  diagnose failures (see inline comments)

* I'm not sure whether some of the Messages are needed.
* I'm think we should discuss with @itamarst whether we're making the best use of actions here
* I wonder if the logged actions could be applied generally in a `proxyForInterface(IBlockDeviceAPI)` which is a twisted thing.
 * I wonder if we should add an Eliot action to the test methods and to the cleanup functions so that we know where who is calling the failing `IBlockDeviceAPI` operations.

Please:
 * address or answer the inline comments above, 
 * discuss the logging strategu with @itamarst and 
 * resubmit for a final (quick) review.

flocker/node/agents/ebs.py
```diff
@@ -20,12 +20,15 @@
...
-from eliot import Field, MessageType
 
 from .blockdevice import (
     IBlockDeviceAPI, BlockDeviceVolume, UnknownVolume, AlreadyAttachedVolume,
     UnattachedVolume, get_blockdevice_volume
 )
+from ._logging import (
+    AWS_ACTION, BOTO_EC2RESPONSE_ERROR, NO_AVAILABLE_DEVICE,
+    NO_NEW_DEVICE_IN_OS
+)
```
wallrj: What's your plan for the new `_logging` module?
Will we move all our Eliot definitions here? Including the ones already defined in `blockdevice` and `cinder`?
Which types, messages and actions can be shared between the implementations?
And `_logging` implies that its contents are private. Are there any which should be made public in `flocker.node.agents.__init__` ?

flocker/node/agents/_logging.py
myechuri: Fixed.

flocker/node/agents/_logging.py
myechuri: Fixed.

flocker/node/agents/_logging.py
myechuri: The error is ``boo.exception.EC2ResponseError``, thanks for noticing the typo. Fixed.
EC2ResponseError covers all errors in response from EC2 to Boto.

flocker/node/agents/_logging.py
myechuri: Fixed. Fixed for NO_NEW_DEVICE_IN_OS as well.

flocker/node/agents/_logging.py
myechuri: Thanks for pointing out! Fixed in other parts of the file as well.

flocker/node/agents/_logging.py
itamarst: It will blow up (in a nice don't-break-actual-business-logic way). There's no implicit `repr()` of random objects.

myechuri: > Is list a valid Field type?

Yes, it is: https://github.com/ClusterHQ/eliot/blob/master/eliot/_validation.py#L124

flocker/node/agents/ebs.py
myechuri: > All the information logged in the message is already logged by the failed action...wrapped in XML tags...but the info is there already.

True. 2 reasons why having ``BOTO_EC2RESPONSE_ERROR`` is useful:
1) The error fields are nicely presented for future processing.
2) Easier to unit test using ``test_boto_ec2response_error`` in ``test_ebs.py``.

Please let me know if the above two reasons for having explicit ``BOTO_EC2RESPONSE_ERROR`` do not sound reasonable? Thanks!

myechuri: > It'd be interesting to see if the new log messages allow us to diagnose why that call to delete_volume fails.
Volume vol-a346f1b3 is currently attached to i-115ccfe6
> Why is that volume still attached?
> Is it because we haven't yet attempted to detach it?
> No, I can see a "successful" detach operation logged prior to calling delete, so perhaps we need to make the detach method more robust.

Yes, the log message pointed to the cause of the failed test. What happened was:

1) Detach succeeded in initiating detach.
2) We timed out (after 180 seconds) waiting for the the volume to transition from ``in-use`` to ``available``. Detach took longer than 180 seconds.
3) Detach operation returned success.
4) Test moved on to attempting destroy while detach was still in progress.

myechuri: Am fixing the detach bug in this branch since it is easy to verify.

flocker/node/agents/_logging.py
myechuri: > I'm thinking of a situation where a caller has supplied the wrong argument types to attach_volume. In which case the list might be ['attach_volume', BlockDeviceVolume()...]. How will Eliot know how to serialise BlockDeviceVolume ? Does it just call repr? Or will it blow up when our stdout log observer attempts to serialise that object?

The other alternate is to have separate MessageTypes for each of IBlockDeviceAPI operations. 
My initial thought was that this would cause bloated log definitions in _logging.py.

flocker/node/agents/ebs.py
myechuri: Good idea! Logged.

flocker/node/agents/ebs.py
myechuri: > it'd be even more useful (I think) if it included the blockdevice_id and the real allocated device path...the one found by _wait_for_new_device (above) which is different, right?

Agree. But then, we will have to pass ``blockdevice_id`` to ``_next_device`` just for logging. i am slightly inclined to not do that because this is going to go away once FLOC-1887 is in, as you pointed out. If you feel strongly about it, happy to make the change. Thanks!

> the one found by _wait_for_new_device (above) which is different, right?
``_wait_for_new_device`` is after attach happened, so, there will be a new allocated device path.
At ``_next_device``, we haven't yet attempted an attach.

flocker/node/agents/ebs.py
myechuri: i want to consolidate all node/agents/* logging actions in _logging, if this pattern looked ok to you. i took route/_logging.py as an example for this file. If this pattern looks fine, will consolidate ``block device`` and ``cinder`` logs here as a followup.

myechuri:
> I wonder if the logged actions could be applied generally in a proxyForInterface(IBlockDeviceAPI) which is a twisted thing.

Thanks for the pointer! ``proxyForInterface`` looks interesting! Will explore unifying logging across storage drivers in a followup branch.
Meanwhile, added first level support for OpenStack logging in this branch. Please let me know if it looks ok. Thanks!

myechuri:
> I wonder if we should add an Eliot action to the test methods and to the cleanup functions so that we know where / who is calling the failing IBlockDeviceAPI operations.

Good idea! Since this branch is already bloated up and includes a couple of unrelated fixes, will take up test logging in a follow up. Thanks for pointing out!

flocker/node/agents/ebs.py
myechuri: Tried first attempt at OpenStack ActionType in this branch. The straightforward log change caused a lot of buildbot test failures, which made me realize that adding OpenStack logging will require some code refactoring. Since this branch is getting big, decided to revert OpenStack changes from the branch and do it in a follow up. Please let me know if this does not sound reasonable. Thanks!

flocker/node/agents/_logging.py
```diff
@@ -0,0 +1,78 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+# Helper module to provide macros for logging support
+# for storage drivers (AWS, Cinder).
```
wallrj: Put this in a module docstring ie

```
"""
Helper module to provide macros for logging support
for storage drivers (AWS, Cinder).
"""
```

wallrj:
Thanks @myechuri 

The new logging is great. Can't wait to have the same in the cinder module.

Not so sure about removing the time out. I think that probably ought to stay.

Let's discuss with everyone and merge when we've reached a conclusion.

wallrj:
Oh and branch needs merging forward.

Merge when green.

flocker/node/agents/_logging.py
myechuri: Oops, sorry, fixed.

PR #1455, opened 2015-05-26
[FLOC-1987] Improve test cleanup after test_ebs.py

myechuri:
1) Retry failed test cleanup attempts.
2) Improve a few test cases (noticed due to failed tests in my local env).


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -66,6 +68,18 @@
 from ....common.test.test_thread import NonThreadPool, NonReactor
 
 LOOPBACK_BLOCKDEVICE_SIZE = int(MB(64).to_Byte().value)
+CLEANUP_RETRY_LIMIT = 10
+
+# Helper structures to Eliot output leftover volumes after a test run.
+VOLUMES = Field.for_types(
```
itamarst: Since this is test code only you could (in the future) just use `start_action` directly; typed messages are good for things like production code where we want to write unit tests to validate that logging really works, and do serialization etc.. For logging *by* tests it's not needed.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -193,14 +207,26 @@ def create_blockdevicedeployer(
 def detach_destroy_volumes(api):
     """
     Detach and destroy all volumes known to this API.
+    If we failed to detach a volume for any reason,
```
itamarst: Presumably failures might happen as exceptions? In which case the loop won't happen. So maybe add try/except that logs errors.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1473,6 +1499,8 @@ def test_attach_volume_validate_size(self):
             new_volume.blockdevice_id, attach_to=self.this_node,
         )
 
+        self.assertNotEqual(attached_volume, None)
```
itamarst: Why are these assertions necessary (as opposed to any other random assertion)? Won't returned `None` cause later assertions to fail?

flocker/node/agents/test/test_blockdevice.py
myechuri: In addition to failures happening as exceptions (not yet seen in test runs), there could be yet-to-be-reasoned-out failures as well:
Some of the tests run in a pull request experienced the following set of actions:

1) Detach volume - success
2) Destroy volume: Failed because volume was ``in-use`` state.

Will modify the loop to try-except that logs and retries. Thanks!

flocker/node/agents/test/test_blockdevice.py
myechuri: Thanks a lot for the explanation! Will modify this branch since this is a small change.

flocker/node/agents/test/test_blockdevice.py
myechuri: i was trying to save cpu cycles attempting a future operation using ``attached_volume`` when it is guaranteed to not succeed.

itamarst: I wouldn't bother. CPU cycles are not a bottleneck in our tests. In general better to just have one assertion that tests what we care about.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -193,14 +196,34 @@ def create_blockdevicedeployer(
...
+        while retry < CLEANUP_RETRY_LIMIT and len(volumes) > 0:
+            for volume in volumes:
+                try:
+                    if volume.attached_to is not None:
+                        api.detach_volume(volume.blockdevice_id)
+                    api.destroy_volume(volume.blockdevice_id)
+                except Exception as e:
+                    Message.new(u"agent:blockdevice:failedcleanup:volume",
+                                blockdevice_id=volume.blockdevice_id,
+                                error=str(e).write(Logger()))
```
itamarst: No need to pass in `Logger` instance.

itamarst: I suggest `eliot.write_traceback`, since it'll include more information.

itamarst:
Remove `assertNotEqual` you added, and other minor comment, merge forward, merge when buildbot is happy.

flocker/node/agents/test/test_blockdevice.py
myechuri: Ok, will revert back. Thanks.

myechuri:
Thanks for the review, @itamarst ! 
Removed ``assertNotEqual``, moved to ``write_traceback``, removed ``Logger``, and merged forward.

PR #1456, opened 2015-05-27
[FLOC-1988] flocker-ca create-control-certificate output

dwgebler:
Fixes FLOC-1988 https://clusterhq.atlassian.net/browse/FLOC-1988


wallrj:
May as well fix the other missing trailing newlines here too. Eg:
```
$ flocker-ca initialize mycluster
Created cluster.key and cluster.crt. Please keep cluster.key secret, as anyone who can access it will be able to control your cluster.(1988)(flocker-ca-output-FLOC-1988 ? =)[~/.../HybridLogic/flocker]$
```

https://clusterhq.atlassian.net/browse/FLOC-1930

wallrj:
I notice that `flocker-ca create-control-certificate` synopsis doesn't document the expected positional arguments.

```
$ flocker-ca create-control-certificate
Usage: flocker-ca <command> [options]
Options:
  -i, --inputpath=   Path to directory containing root certificate. Defaults to
                     current working directory.
  -o, --outputpath=  Path to directory to write control service certificate.
                     Defaults to current working directory.
      --version      Print the program's version and exit.
      --help         Display this help and exit.
  -v, --verbose      Turn on verbose logging.
flocker-ca is used to create TLS certificates.

The certificates are used to identify the control service, nodes and
API clients within a Flocker cluster.
ERROR: Wrong number of arguments.
```

Fix it here or create a followup issue.

wallrj:
The new output looks ok

```
$ flocker-ca create-control-certificate control.example.com
Created control-control.example.com.crt and control-control.example.com.key.
Copy these files to the directory /etc/flocker on your control service machine. Rename the files to control-service.crt and control-service.key and set the correct permissions by running chmod 0600 on both files.

```

But maybe it'd be more readable with some added newlines

flocker/ca/_script.py
```diff
@@ -269,11 +269,23 @@ def run(self):
...
+                    b"Created {certificate} and {key}.\n"
+                    b"Copy these files to the directory /etc/flocker on your "
+                    b"control service machine. Rename the files to "
+                    b"control-service.crt and control-service.key and "
+                    b"set the correct permissions by running chmod 0600 on "
+                    b"both files."
+                    b"\n".format(
+                        certificate=output_certificate_filename,
+                        key=output_key_filename
+                    )
```
wallrj: It works in Python2.7, but I don't think we're supposed to use `format` on `byte` strings. 
In Python3 byte strings don't have that method.
But these strings can probably be `unicode` instead and then encoded to `utf-8` before dumping them to `stdout`

wallrj:
Thanks @dwgebler 

Please:
 * address or answer the inline comments above
 * Fix other missing newlines and close https://clusterhq.atlassian.net/browse/FLOC-1930
 * Create followup issue for fixing and testing synopsis of flocker-ca subcommands (I think our other functional script tests already check that the supplied command is present in the synopsis)

Then merge.

PR #1458, opened 2015-05-27
[FLOC-1941] Gather logs from cluster nodes when running acceptance tests

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-1941

Depends on #1437.


admin/acceptance.py
```diff
@@ -452,6 +453,51 @@ def eliot_output(message):
...
+    :param file output_file: File to write to.
+    """
+    ran = run(reactor, [
+        "ssh",
+        b"-C",  # compress traffic
+        b"-q",  # suppress warnings
+        b"-l", 'root',
+        # We're ok with unknown hosts; we'll be switching away from
+        # SSH by the time Flocker is production-ready and security is
+        # a concern.
```
exarkun: Probably this comment isn't relevant here.

admin/acceptance.py
```diff
@@ -452,6 +453,51 @@ def eliot_output(message):
...
+        b"-o", b"ControlMaster=no",
+        # Some systems (notably Ubuntu) enable GSSAPI authentication which
+        # involves a slow DNS operation before failing and moving on to a
+        # working mechanism.  The expectation is that key-based auth will
+        # be in use so just jump straight to that.  An alternate solution,
+        # explicitly disabling GSSAPI, has cross-version platform and
+        # cross-version difficulties (the options aren't always recognized
+        # and result in an immediate failure).  As mentioned above, we'll
+        # switch away from SSH soon.
+        b"-o", b"PreferredAuthentications=publickey",
```
exarkun: Not sure when we'll switch away from SSH for this (though it would certainly be nice to have a real log aggregation system).

admin/acceptance.py
```diff
@@ -452,6 +453,51 @@ def eliot_output(message):
...
+        host,
+        ' '.join(map(shell_quote, [
+            'journalctl',
+            '--lines', '0',
+            '--follow',
+            # Only bother with units we care about:
+            '-u', 'docker',
+            '-u', 'flocker-control',
+            '-u', 'flocker-dataset-agent',
+            '-u', 'flocker-container-agent',
```
exarkun: Bytes literals for all these

admin/acceptance.py
```diff
@@ -483,6 +529,10 @@ def main(reactor, args, base_path, top_level):
 
     try:
         nodes = yield runner.start_nodes(reactor)
+        if options['distribution'] in ('fedora-20', 'centos-7'):
+            remote_logs_file = open("remote_logs.log", "a")
+            for node in nodes:
+                capture_journal(reactor, node.address, remote_logs_file)
```
exarkun: How do these connections end?  I guess they live for the entire process lifetime and then die when the reactor stops...

exarkun:
Thanks.  This is pretty gross.  I really want the functionality, though.  So I think you should merge it.  Are the necessary buildbot changes done somewhere?

exarkun:
Or ... if you just wrote to the Eliot log instead of a new log file, the buildbot changes wouldn't be necessary?

itamarst:
Writing to Eliot log file means JSON inside of JSON, which will be really hard to read. There's a trivial one line change to buildbot config that will be done in follow-up issue.

admin/acceptance.py
itamarst: Yeah, lovely innit.

admin/acceptance.py
itamarst: Tom just copy/pasted some code for his initial experiment, I think, and I forgot to edit away comments.

PR #1459, opened 2015-05-27
[FLOC-1976] [FLOC-2022] Non-root volumes are writeable: an  acceptance test

itamarst:
This includes both adding an acceptance test (FLOC-1976), and fixing the resulting failures in the block device backend (FLOC-2022).


flocker/node/functional/test_docker.py
```diff
@@ -726,13 +728,15 @@ def test_command_line(self):
...
+            command_line=pvector([u"sh", u"-c", u"""\
 echo -n '#!/bin/sh
 echo -n "HTTP/1.1 200 OK\r\n\r\nhi"
 ' > /tmp/script.sh;
 chmod +x /tmp/script.sh;
 nc -ll -p 8080 -e /tmp/script.sh
-"""],
+"""]),
             ports=[PortMap(internal_port=8080,
                            external_port=external_port)])
```
exarkun: These changes seem related to some other issue.

exarkun: Sorry, that was a little vague.  What I meant was "Issue per logical change"

flocker/acceptance/test_api.py
```diff
@@ -286,6 +288,60 @@ def in_current():
         creating.addCallback(created)
         return creating
 
+    @require_cluster(1)
+    def test_non_root_container_can_access_dataset(self, cluster):
+        """
+        A container running as a user that is not root can still write to a
```
exarkun: I'd stop `still`

flocker/acceptance/test_api.py
```diff
@@ -286,6 +288,60 @@ def in_current():
...
+                    u"su", u"-", u"nobody", u"-c", u"sh", u"-c",
+                    # Write something to volume we attached, and then
+                    # expose what we wrote as a web server:
+                    u"""\
+echo -n '#!/bin/sh
+echo -n "HTTP/1.1 200 OK\r\n\r\nhi"
+' > /data/script.sh;
+chmod +x /data/script.sh;
+nc -ll -p 8080 -e /data/script.sh
+                    """]}
```
exarkun: Ow, the quoting, my brain.

flocker/acceptance/test_api.py
```diff
@@ -286,6 +288,60 @@ def in_current():
...
+
+        def query(host):
+            req = get(
+                "http://{host}:{port}".format(host=host, port=port),
+                persistent=False
+            ).addCallback(content)
+            return req
+
+        def checked(cluster):
+            host = cluster.nodes[0].address
```
exarkun: Factor `cluster.nodes[0]` duplication out

flocker/acceptance/test_api.py
exarkun: er... drop.

flocker/node/functional/test_docker.py
```diff
@@ -726,13 +728,15 @@ def test_command_line(self):
         name = random_name(self)
         d = self.start_container(
             name, image_name=u"busybox",
-            command_line=[u"sh", u"-c", u"""\
+            # Pass in pvector since this likely to be what caller actually
+            # passes in:
+            command_line=pvector([u"sh", u"-c", u"""\
```
exarkun: Probably should have a test for at least two types, maybe pvector and list (or a test using an entirely custom sequence implementation ...).

exarkun:
So ... looks like it did indeed fail.

```
Failure: twisted.web._newclient.ResponseNeverReceived: [<twisted.python.failure.Failure <class 'twisted.internet.error.ConnectionDone'>>]
```

It would be pretty great if it failed in a more meaningful way.  Something like `/data/script.sh: Permission denied` would be fabulous.  Perhaps switching away from shell would make it easier to capture and then report the write error so the test fails with that information instead of with whatever random error you get when you make an HTTP request to an nc server that's been told to do something impossible.

exarkun:
And then I expect this should come with a fix for the failure, too?

exarkun:
Thanks.  Please address those points (would be great if you could split the Docker fix out into a separate branch but I probably won't reject the PR if you don't ... because I'm a bad reviewer I guess) and re-submit.

flocker/node/functional/test_docker.py
itamarst: I copied some code from there, didn't intend to modify it. Bit confused how it worked before, or alternately how it works now.

itamarst:
The fix for the block device-specific failures is a separate issue, which I will now add to the sprint now (FLOC-2022).

exarkun:
> The fix for the block device-specific failures is a separate issue, which I will now add to the sprint now (FLOC-2022).

Okay but don't introduce new failing tests into master. :smile:

itamarst:
GAH. The docker changes were merged in a separate branch already so I dunno why they showed up in the diff here. If I'm lucky it's github diff caching. If I'm unlucky something is messed up in the repository itself.

itamarst:
I could fix FLOC-2022 in this branch too, I suppose?

exarkun:
> I could fix FLOC-2022 in this branch too, I suppose?

It seems like that would be the normal workflow, would it not be?  Add a failing test, fix the implementation, party.

itamarst:
ok

wallrj:
There's a merge conflict. I'll try and merge forward.

flocker/acceptance/test_api.py
```diff
@@ -305,6 +305,69 @@ def in_current():
         creating.addCallback(created)
         return creating
 
+    def assert_busybox_http(self, host, port):
+        """
+        Assert that a HTTP serving a response with body ``b"hi"`` is running
+        at given host and port.
```
wallrj: Seems like something's missing from this docstring.

flocker/acceptance/test_api.py
```diff
@@ -305,6 +305,69 @@ def in_current():
...
+
+        :param host: Host to connect to.
+        :param port: Port to connect to.
```
wallrj: Document the types.

flocker/acceptance/test_api.py
```diff
@@ -305,6 +305,69 @@ def in_current():
...
+                # Run as non-root user:
+                u"su", u"-", u"nobody", u"-c", u"sh", u"-c",
+                # Write something to volume we attached, and then
+                # expose what we wrote as a web server:
+                u"""\
+echo -n '#!/bin/sh
+echo -n "HTTP/1.1 200 OK\r\n\r\nhi"
+' > /data/script.sh;
+chmod +x /data/script.sh;
+nc -ll -p 8080 -e /data/script.sh
```
wallrj: What's ``-ll`` ? 
http://www.busybox.net/downloads/BusyBox.html#nc

flocker/node/agents/blockdevice.py
```diff
@@ -852,7 +875,7 @@ def run(self, deployer):
         check_output(["mkfs", "-t", "ext4", device.path])
 
         # This duplicates MountBlockDevice now.
-        self.mountpoint.makedirs()
+        _create_mount_directory(self.mountpoint)
```
wallrj: Does this work? 
I thought you had to set the permissions on the mounted filesystem. 
ie the directory permissions of the mount root need to be stored in the mounted filesystem.

wallrj:
Sorry, I got distracted half way through the review.

I haven't checked, but I think you need to set the mountpoint permissions after mounting the filesystem. 

Merge this if I'm wrong, otherwise add a test to verify this and then merge.

itamarst:
That sounds... extremely plausible. Let me go check buildbot...

flocker/node/agents/blockdevice.py
```diff
@@ -574,6 +575,29 @@ def run(self, deployer):
...
+        mountroot = mountpoint.parent()
+        if not mountroot.exists():
+            mountroot.makedirs()
+        mountroot.chmod(S_IRWXU)
+        if not mountpoint.exists():
+            mountpoint.makedirs()
+        mountpoint.chmod(S_IRWXU | S_IRWXG | S_IRWXO)
+        mountpoint.restat()
+    finally:
+        umask(original_umask)
```
exarkun: Flocker is rather moderately multi-threaded.  Can we get away with this? :/

flocker/acceptance/test_api.py
itamarst: For some reason that's less info than man page inside the docker container. I added comment with reference.

itamarst:
Yeah, looks like it's failing.

flocker/node/agents/blockdevice.py
itamarst: let me look at my copy of The Linux Programming Interface, and see what I find out!

itamarst: So apparently there is a way to make `umask` not cross-thread, by creating special kind of thread, but that doesn't help us. In any case, though, apparently I don't actually need `umask`? Possibly I thought I did cause of confusion caused by `FilePath` stat caching...

itamarst:
OK, looks like this makes the test pass on block device backend.

It does *not* make the two failing postgres tests pass, so I guess I'll file an issue for them.

Any interest in doing another review pass, or shall I just merge? I made a bunch of changes.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2845,6 +2845,19 @@ def test_mountpoint_error(self):
...
+        The mountpoint is world-writeable (since containers can run as any
+        user), and its parent is only accessible as current user (for
+        security).
+        """
+        mountroot = mountroot_for_test(self)
+        mountpoint = mountroot.child(b"mount-test")
+        self._run_success_test(mountpoint)
+        self.assertEqual((mountroot.getPermissions().shorthand(),
+                          mountpoint.getPermissions().shorthand()),
+                         ('rwx------', 'rwxrwxrwx'))
```
wallrj: :question: I can see how this will fix the problem, but those permissions (`rwxrwxrwx`) seem a bit too open when I see them represented that way.

 * User `root` owns the mount point directory.
 * A process in a container running as user `nobody` can create a file or a sub-directory and 
 * the default permissions on those will be determined by the umask of the process.
 * But *any user* on the node who knows the path to the mount point can delete all the files inside?

But that is prevented because the mount root directory permissions do not include user / group execute. Right?

Yeah, I see it now.

```
[~/tmp/foo]$ ls -la
total 0
drwx------. 1 richard richard   6 Jan 20 19:15 .
drwxrwxr-x. 1 richard richard 992 May 13 14:36 ..
drwxrwxr-x. 1 richard richard   6 Jan 20 19:15 bar
[~/tmp/foo]$ cd ../
[~/tmp]$ sudo chown root:root foo
[sudo] password for richard: 
[~/tmp]$ cd foo/
bash: cd: foo/: Permission denied
[~/tmp]$ cd foo/bar
bash: cd: foo/bar: Permission denied
[~/tmp]$ ls -l foo/bar/
ls: cannot access foo/bar/: Permission denied
[~/tmp]$ rm -rf foo/bar
rm: cannot remove ‘foo/bar’: Permission denied
[~/tmp]$ rm -rf foo/bar/baz
rm: cannot remove ‘foo/bar/baz’: Permission denied

```


So all those files and sub-directories are owned by user `nobody` on node1 who has UID 99

But what if `nobody` on node2 has a different UID?

I suppose that was always the case though.
Worth a followup issue?
Worth marking this bit of code for a future security audit?
Your call.

wallrj:
Thanks @itamarst 

This looks good to me. Please merge.

flocker/node/agents/test/test_blockdevice.py
wallrj: I also meant to mention "Sticky Bit":
 * http://en.wikipedia.org/wiki/Sticky_bit

But not sure if that would make things any more secure or whether it would add complication.

itamarst:
The UIDs come out of the container, not the node. So if the container's image changes such that it runs as different UID that's a problem, yes... but not something we can solve.

PR #1460, opened 2015-05-27
[FLOC-1734] Acceptance test for linking

itamarst:


flocker/acceptance/test_api.py
```diff
@@ -286,6 +288,70 @@ def in_current():
         creating.addCallback(created)
         return creating
 
+    @require_cluster(2)
+    def test_linking(self, cluster):
+        """
+        A container one machine can be linked to another.
```
myechuri: Typo around ``one``. Thanks!

myechuri:
Interesting! Simpler and cleaner than the log processing validation i was going for. Thanks!

itamarst:
@myechuri was that a review, or just some comments?

flocker/acceptance/test_api.py
```diff
@@ -286,6 +288,70 @@ def in_current():
         creating.addCallback(created)
         return creating
 
+    @require_cluster(2)
+    def test_linking(self, cluster):
```
adamtheturtle: Enough of this is shared with https://github.com/ClusterHQ/flocker/pull/1459/files that some of the comments on that are (and future ones likely will be) applicable.

adamtheturtle:
Thanks @itamarst this looks like a sensible linking test but is quite hard to read (though perhaps not for @myechuri ).  It seems like a place for inline callbacks.

Don't merge until #1459 has been accepted and all of the comments on that (particularly to do with readability) which apply to this have been addresed.

Perhaps one solution would be to change the layout to be something like:

```python
def create_container_and_add_callback_to_delete_it(container):
  ...

container_which_posts_hi = {
  u"name": random_name(self),
  ...
}

container_which_netcats = {
  u"name": random_name(self),
  ...
}


created = create_container_and_add_callback_to_delete_it(container_which_posts_hi)

created = created.addCallback(lambda: create_container_and_add_callback_to_delete_it(container_which_netcats))

created.addCallback(checked)
```


I'll leave it to you to decide how much to do.
Resubmit if the changes are significant, else merge.

itamarst:
There were no comments about readability on #1459... but then that's Jean-Paul. There were comments about debugability, though, so if I come up with framework for that over there then maybe I can reuse it here.

adamtheturtle:
> Ow, the quoting, my brain.

and

> Factor cluster.nodes[0] duplication out

are what I was referring to.

itamarst:
Clarity is better, I don't think it's worth spending more time on polish.

PR #1461, opened 2015-05-27
[FLOC-1894] Clean up server-side connections

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1894

### DIFF
--- a/flocker/ca/functional/test_validation.py
+++ b/flocker/ca/functional/test_validation.py
@@ -13,7 +13,7 @@ from twisted.internet.endpoints import (
     SSL4ServerEndpoint, connectProtocol, SSL4ClientEndpoint,
     )
 from twisted.internet import reactor
-from twisted.internet.defer import Deferred
+from twisted.internet.defer import Deferred, gatherResults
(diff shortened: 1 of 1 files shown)


flocker/ca/functional/test_validation.py
```diff
@@ -74,6 +81,22 @@ def getContext(self):
         return ctx
 
 
+class WaitForDisconnectsFactory(ServerFactory):
+    """
+    A factory for use with ``SendingProtocol`` that makes it possible to wait
```
adamtheturtle: Does this set us up for timeouts if something does not disconnect? If so I guess we won't have clear error messages telling us what went wrong.

adamtheturtle:
This passes the tests, which is great. I'm only going to assume that the diagnosis was correct, and given that, this change looks good.

Optionally add error reporting where I suggested it.

flocker/ca/functional/test_validation.py
exarkun: We will have at least as much information as we have now, a dirty reactor warning pointing at the connection that's still connected.

PR #1463, opened 2015-05-27
[FLOC-2002] Config gsg floc 2002

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2002


docs/gettingstarted/index.rst
```diff
@@ -210,39 +210,31 @@ Getting started with Flocker
       You should have the Flocker client installed on your laptop and flocker-node installed on some servers: either VMs on your laptop, or real instances on cloud infrastructure.
       Now you can try our simple tutorial: a Python web application and a Redis server.
 
-      .. code-block:: console
+      .. container:: hidden
```
exarkun: What does this do?

adamtheturtle: There is a comment within this block which says "Create the files to be downloaded with curl, but don't show download links for them"

adamtheturtle: Perhaps that comment should be moved to above the block.

docs/gettingstarted/index.rst
```diff
@@ -210,39 +210,31 @@ Getting started with Flocker
...
-           ports:
-            - "80:80"
-         redis:
-           image: redis
-           ports:
-            - "6379:6379"
-           volumes: ["/data"]
+         you@laptop:~$ curl -O https://docs.clusterhq.com/en/|latest-installable|/_downloads/fig.yml
+         you@laptop:~$ curl -O https://docs.clusterhq.com/en/|latest-installable|/_downloads/deployment-node1.yml
+         you@laptop:~$ curl -O https://docs.clusterhq.com/en/|latest-installable|/_downloads/deployment-node2.yml
```
exarkun: I think replacing one `git clone ...` command with three `curl ...` commands will probably annoy some people (even though this is just copy/paste stuff).

I think you could fix this with shell shenanigans, eg `curl ... /_downloads/{fig,deployment-node1,deployment-node2}.yml`.  Or we could host a tar archive and `curl ... | tar xf -` - though I don't really want to spend time on the setup that would require.  Maybe just checking an uncompressed tar archive into the repo would be an okay interim solution (with an issue filed for follow-up to replace it with unarchived files and a doc generation step that creates the tar archive).

Thoughts?

exarkun: Also, it occurs to me that whatever we ultimately do we should probably combine the `Vagrantfile` download into it as well - then the user has one download step instead of two.

docs/gettingstarted/index.rst
```diff
@@ -210,39 +210,31 @@ Getting started with Flocker
...
 
-      .. code-block:: yaml
-
-         "version": 1
-         "nodes":
-           "172.16.255.250": ["web", "redis"]
-           "172.16.255.251": []
-
+      .. literalinclude:: deployment-node1.yml
+         :language: yaml
```
exarkun: Yaaay.

docs/gettingstarted/index.rst
```diff
@@ -273,12 +265,8 @@ Getting started with Flocker
...
       --------------------
 
-      .. code-block:: yaml
-
-         "version": 1
-         "nodes":
-           "172.16.255.250": ["web"]
-           "172.16.255.251": ["redis"]
+      .. literalinclude:: deployment-node2.yml
+         :language: yaml
```
exarkun: Yaaay.

docs/gettingstarted/index.rst
exarkun: I mean what does `.. container:: hidden` do?

adamtheturtle: It puts the contents into an HTML div with the class `hidden`.

docs/gettingstarted/index.rst
adamtheturtle: I don't like `tar` from an implementation point of view because I like having one `fig.yml` which can be shown in a `literalinclude` and downloaded. I'll try some alternatives.

adamtheturtle: Note that it is now three `curl` commands but was previously not just a `clone` but a `cd`.

docs/gettingstarted/index.rst
exarkun: Is that better than just using a rst comment without a hidden container?

exarkun:
Thanks.  Basically looks good.  I'll let you try to figure out something better than three curl commands (but don't spend a ton of time on it).  Merge when you're happy.

exarkun:
And once merged, please put something in the flocker-quickstart repository indicating it is obsolete.

adamtheturtle:
Thanks, I'll also use https://github.com/ClusterHQ/flocker-quickstart/pull/2.

lukemarsden:
Flagging @alicesowerby to make sure she's aware of this change.

docs/gettingstarted/index.rst
adamtheturtle: If I'm understanding you correctly you want there to be just a comment like:

```
.. :download:`fig.yml`
```

to replace 

```
      .. container:: hidden

         .. Create the files to be downloaded with curl, but don't show download links for them

         :download:`fig.yml`
```

The download directive in the former doesn't actually create a file to download (in `_downloads`) but the latter does.

adamtheturtle:
I have tried using `&& \`s like the install commands, but I think that this is worse because of the highlighting issue (https://clusterhq.atlassian.net/browse/FLOC-1318)

I tried

```
curl -O https://docs.clusterhq.com/en/0.4.1dev1/_downloads/fig.yml -O https://docs.clusterhq.com/en/0.4.1dev1/_downloads/deployment-node1.yml -O https://docs.clusterhq.com/en/0.4.1dev1/_downloads/deployment-node2.yml
```

but that is just unreadable.

I like what is there now because it is one extra copy / paste, has no rendering issues and you can easily know what is going on. If it is a real UX issue we can follow up later but I don't think it is worth investigating now.

lukemarsden:
@adamtheturtle fwiw, I preferred @exarkun 's `curl ... /_downloads/{fig,deployment-node1,deployment-node2}.yml` suggestion

PR #1464, opened 2015-05-27
Tested osx floc 2004

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2004

Also happens to fix https://clusterhq.atlassian.net/browse/FLOC-1551.

This removes the &&s, but part of the advantage of that is that you can copy in one go - the prompt directive allows this to be done too by not copying the prompt.

### DIFF
--- a/docs/gettingstarted/index.rst
+++ b/docs/gettingstarted/index.rst
@@ -81,11 +81,8 @@ Getting started with Flocker
 
          Install the Flocker client on your Mac (requires Homebrew):
 
-         .. version-code-block:: console
-
-            you@laptop:~$ brew update && \
-              brew tap clusterhq/flocker && \
-              brew install flocker-|latest-installable|
+         .. task:: test_homebrew flocker-|latest-installable|
+            :prompt: you@laptop:~$
 
       .. noscript-content::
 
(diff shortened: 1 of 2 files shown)


docs/gettingstarted/index.rst
```diff
@@ -81,11 +81,8 @@ Getting started with Flocker
 
          Install the Flocker client on your Mac (requires Homebrew):
 
-         .. version-code-block:: console
-
-            you@laptop:~$ brew update && \
-              brew tap clusterhq/flocker && \
-              brew install flocker-|latest-installable|
+         .. task:: test_homebrew flocker-|latest-installable|
+            :prompt: you@laptop:~$
```
exarkun: I guess we're happy that the order of `brew tap ...` and `brew update` is changing (it's the other way around in `task_test_homebrew` and we're happy to have the use run the `brew test ...` command as well?  Considering that's what we're testing I think it seems like a great change, just want to be sure.

adamtheturtle: I'm happy with it, and as you say this is tested so if it stops working we can swap them or change things.

exarkun:
> This removes the &&s, but part of the advantage of that is that you can copy in one go

An advantage of the `&&` style is that if a command fails the pipeline stops.  I think this will no longer be the case.  I'm not sure it really matters, though.  Maybe later on we can have a task-y thing that lets you re-combine steps with `&&` (please don't file a follow-up issue for this now though).

So - Thanks.  This looks good (apart from the fact that everything has to be fixed twice in the getting started guide... but that's pre-existing...).  So please merge.

PR #1465, opened 2015-05-28
[FLOC-1622] Manual cli test floc 1622

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1622

PR #1467, opened 2015-05-28
[FLOC-1978] [FLOC-1325] Cli install floc 1978

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1978

Also happens to fix https://clusterhq.atlassian.net/browse/FLOC-1325

This has been tested by following the getting started and tutorial instructions on a container started with:

```
docker run -i -t fedora:20 /bin/bash
```

and the tutorial instructions for "Other Linux Distributions" (http://doc-dev.clusterhq.com/indepth/installation.html#other-linux-distributions) with:

```
$ docker run -i -t ubuntu /bin/bash
```

These did not work before.
These work now.
The instructions for how to test these are proposed in https://github.com/ClusterHQ/flocker/pull/1465.

I have also made the tangentially related change to download the install script in Fedora instead of duplicating it - an attempt to decrease the maintenance burden. This can be reverted if you think that it should not be done (here).


docs/gettingstarted/index.rst
```diff
@@ -105,10 +105,10 @@ Getting started with Flocker
 
          .. version-code-block:: console
 
-            you@laptop:~$ sudo yum install -y @buildsys-build python python-devel python-virtualenv && \
-              virtualenv flocker-tutorial && \
-              flocker-tutorial/bin/pip install --upgrade pip && \
-              flocker-tutorial/bin/pip install --quiet https://clusterhq-archive.s3.amazonaws.com/python/Flocker-|latest-installable|-py2-none-any.whl && source flocker-tutorial/bin/activate
+            you@laptop:~$ sudo yum install -y @buildsys-build python python-devel python-virtualenv libffi-devel openssl-devel && \
+              curl -O https://docs.clusterhq.com/en/|latest-installable|/_downloads/linux-install.sh && \
+              sh linux-install.sh && \
```
exarkun: Cool kids execute straight from the 'net, `curl https://.../linux-install.sh | sh`

adamtheturtle: I'm not that cool, errors piping to shell etc.

docs/indepth/installation.rst
```diff
@@ -35,13 +35,14 @@ On Fedora 20 you can install these by running:
 
 .. code-block:: console
 
-   alice@mercury:~$ sudo yum install @buildsys-build python python-devel python-virtualenv
+   alice@mercury:~$ sudo yum install @buildsys-build python python-devel python-virtualenv libffi-devel openssl-devel
```
exarkun: Aahhh so many copies of this package list.  Can we refactor this somehow?

docs/gettingstarted/index.rst
adamtheturtle: https://www.seancassidy.me/dont-pipe-to-your-shell.html

docs/indepth/installation.rst
adamtheturtle: At least for the two in this document we could make a `task` if you think that is worthwhile.
That would also be necessary if we are ever going to test this (which we now have a framework for).

exarkun: File an issue for getting rid of this duplication.

docs/indepth/installation.rst
```diff
@@ -35,13 +35,14 @@ On Fedora 20 you can install these by running:
...
 
 On Ubuntu or Debian you can run:
 
 .. code-block:: console
 
-   alice@mercury:~$ sudo apt-get install gcc python2.7 python-virtualenv python2.7-dev
+   alice@mercury:~$ sudo apt-get update
+   alice@mercury:~$ sudo apt-get install gcc libssl-dev libffi-dev python2.7 python-virtualenv python2.7-dev
```
exarkun: And this.

docs/gettingstarted/index.rst
exarkun: As written this will still run `sh` on some errors spat out by an HTTP server.  The essential point isn't the use of `|` it's the blind use of a shell on some bytes dumped into a file from the internet...

Although now I realize **this** section of the docs still uses `&&`.  Too bad we're inconsistent about that.  Anyway, not a blocking issue.

exarkun:
Thanks.  File follow-up issues and then merge.

adamtheturtle:
Filed https://clusterhq.atlassian.net/browse/FLOC-2023

PR #1469, opened 2015-05-28
[FLOC-1994] Fix bug where using non-upper-case links caused Flocker to restart containers

itamarst:


flocker/node/test/test_deploy.py
```diff
@@ -493,20 +493,19 @@ def test_link_environment(self):
...
             {
-                u'DASH_ALIAS_PORT_80_UDP': u'udp://the-host:8080',
-                u'DASH_ALIAS_PORT_80_UDP_PROTO': u'udp',
-                u'DASH_ALIAS_PORT_80_UDP_ADDR': u'the-host',
-                u'DASH_ALIAS_PORT_80_UDP_PORT': u'8080',
+                u'DASHALIAS_PORT_80_UDP': u'udp://the-host:8080',
+                u'DASHALIAS_PORT_80_UDP_PROTO': u'udp',
+                u'DASHALIAS_PORT_80_UDP_ADDR': u'the-host',
+                u'DASHALIAS_PORT_80_UDP_PORT': u'8080',
             })
```
exarkun: Mmmhh no, not really, not anymore. :wink:

itamarst: Wait why is that udp?

exarkun: I guess `_link_environment` doesn't care about the protocol you tell it to use...  It would probably make more sense to exercise TCP here since that's what we want to work... But I don't think it actually matters.

flocker/control/_config.py
```diff
@@ -685,11 +685,19 @@ def _link_applications(self):
...
+                            Link(local_port=local_port,
+                                 remote_port=remote_port,
+                                 alias=link_definition['alias'])
+                        )
+                    except InvariantException:
+                        raise ConfigurationError(
+                            ("Application '{application}' has a config "
+                             "error. 'links' alias names must be "
+                             "alphanumeric.").format(
+                                 application=application_name)
```
exarkun: Could this fail because of an invariant exception from `local_port` or `remote_port` and make the new exception misleading?

exarkun: Also `alphanumeric` here annoyingly duplicates the declared schema elsewhere (and so it can become inconsistent with what's being enforced).

flocker/control/_config.py
```diff
@@ -1014,9 +1022,15 @@ def _parse_link_configuration(self, application_name, config):
...
-                                  remote_port=remote_port,
-                                  alias=alias))
+                try:
+                    links.append(Link(local_port=local_port,
+                                      remote_port=remote_port,
+                                      alias=alias))
+                except InvariantException:
+                    # We've already verified types of Link parameters, so
+                    # that just leaves the alias naming limits as the
+                    # source of error.
```
exarkun: The port values could be out of range.  Not something we enforce in the `Link` model now but it could be later.  Can we get some useful information out of the raised `InvariantException`?

flocker/control/_config.py
itamarst: We validate lintness so basically it's "out of range".

flocker/control/_model.py
```diff
@@ -210,7 +216,10 @@ class Link(PRecord):
     """
     local_port = field(mandatory=True, type=int)
     remote_port = field(mandatory=True, type=int)
-    alias = field(mandatory=True)
+    alias = field(mandatory=True, factory=lambda s: s.lower(),
+                  invariant=lambda s: (
+                      (re.match(u"[a-zA-Z0-9]*$", s) is not None,
```
exarkun: `s.isalnum()`

flocker/control/_config.py
itamarst: It seemed a little tricky. My *hope* is that we just throw this code away. Reality is maybe we won't for a while... but if we add new invariants presumably we'll have tests for them which will catch the problem? I really don't want to spend too much time thinking about this particular pile of code.

flocker/control/_model.py
exarkun: Slight difference in semantics ... `u""` matches `u"[a-zA-Z0-9]*$"` but `u"".isalnum()` is `False`.  That's probably a change for the better, though.

flocker/control/_config.py
itamarst: I'll see what InvariantException provides. If something useful, I can use it. If not, I'd rather just leave this as is.

flocker/control/test/test_config.py
```diff
@@ -658,6 +658,35 @@ def test_invalid_fig_config_unknown_link(self):
...
+                'image': 'sample/postgres',
+                'ports': ['54320:5432'],
+                'volumes': ['/var/lib/postgres'],
+                'links': ['wordpress:wordpress-bad'],
+            },
+            'wordpress': {
+                'image': 'sample/wordpress',
+                'ports': ['8080:8080'],
+            }
+        }
```
exarkun: unicode string literals

flocker/control/_config.py
exarkun: Given difficulty of extracting useful information from `InvariantException`, maybe just logging the exception that got raised is sufficient (... er except do logs from this code go anywhere? :/)

flocker/control/_config.py
exarkun: Logging here too, unless logs don't go anywhere in which case I'm just going to close my eyes and sing.

flocker/node/functional/test_deploy.py
```diff
@@ -80,6 +80,8 @@ def change_node_state(deployer, desired_configuration):
         calculate changes.
     :param Deployment desired_configuration: The intended configuration of all
         nodes.
+    :param iterations: Number of times to run discover/calculate/apply
+        loop.
```
exarkun: This looks unrelated?  Is it github screwing with us again?

exarkun: Or maybe it's just unimplemented.  Back this part out?

exarkun:
Thanks.  Looks pretty good.  Please address inline comments, remove partly-implemented `change_node_state` stuff (don't forget the new caller that passes `iterations=1`), and merge.

itamarst:
I think I just forgot to substitute `iterations` for `3` in that for loop.

itamarst:
but I guess it's not necessary so yeah.

PR #1470, opened 2015-05-28
[FLOC-2029] Bump docker-py and requests versions.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2029

I'm hoping the tests pass with this.

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -138,15 +138,6 @@ class GenericDockerClientTests(TestCase):
 
         return d
 
-    def test_default_base_url(self):
-        """
-        ``DockerClient`` instantiated with a default base URL for a socket
-        connection has a client HTTP url after the connection is made.
-        """
-        client = DockerClient()
-        self.assertEqual(client._client.base_url,
-                         u'http+docker://localunixsocket')
-
     def test_custom_base_url_tcp_http(self):
         """
         ``DockerClient`` instantiated with a custom base URL for a TCP
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -145,7 +145,7 @@ class GenericDockerClientTests(TestCase):
         """
         client = DockerClient()
         self.assertEqual(client._client.base_url,
-                         u'http+unix://var/run/docker.sock')
+                         u'http+docker://localunixsocket')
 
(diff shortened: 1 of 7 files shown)


flocker/node/functional/test_docker.py
```diff
@@ -138,15 +138,6 @@ def start_container(self, unit_name,
...
 
-    def test_default_base_url(self):
-        """
-        ``DockerClient`` instantiated with a default base URL for a socket
-        connection has a client HTTP url after the connection is made.
-        """
-        client = DockerClient()
-        self.assertEqual(client._client.base_url,
-                         u'http+unix://var/run/docker.sock')
-
```
tomprince: I'm just deleting the test, since we don't really care if docker-py does with `base_url` if we don't specify one.

adamtheturtle:
I have forced builds on the acceptance test runners where the nodes failed to build.

adamtheturtle:
The `base_url` change seems unrelated? Is it, or is it that the new dependencies change what is done and you thought it easier to change the way that we test connecting to Docker than to modify the tests?

adamtheturtle:
I think that we should pin `python-keystoneclient` because we don't want its changes causing problems for us in the future. Maybe that could be the whole change here.

flocker/node/functional/test_docker.py
adamtheturtle: To confirm - we only checked this because the testtools relied on it to check if it is possible to connect to Docker. This branch changes that check so the URL is not relied on anywhere. If that is correct, this is fine.

adamtheturtle:
Thanks @tomprince address and merge when you and Buildbot are happy.

exarkun:
This seems like a pretty indirect way of dealing with the problem - and, as @adamtheturtle pointed out, possibly without actually solving the problem in a way that will still work tomorrow.  Why not **just** pin the python-keystoneclient version?

adamtheturtle:
There is a branch to do the pinning as suggested above at https://github.com/ClusterHQ/flocker/pull/1472.

itamarst:
We are using a pretty old version of Docker-py; newer version has lots of bugfixes. So we should revisit this soon after 1.0 even if it's not necessary now. Also quite possible we need newer version to support next release of Docker, but we'll see what I figure out for Docker HEAD.

PR #1471, opened 2015-05-28
[FLOC-2006] Factor in configure_cluster.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2006

The test failures will hopefully be fixed by #1470.


admin/acceptance.py
```diff
@@ -194,12 +193,30 @@ def start_nodes(self, reactor):
...
         for node in self.NODE_ADDRESSES:
             yield remove_known_host(reactor, node)
-        returnValue([
+
+        certificates_path = FilePath(mkdtemp())
+        print("Generating certificates in: {}".format(certificates_path.path))
+        certificates = Certificates.generate(
+            certificates_path,
+            self.NODE_ADDRESSES[0],
+            len(self.NODE_ADDRESSES))
```
tomprince: This code is duplicated, but short-lived (see #1466).

admin/acceptance.py
```diff
@@ -62,23 +65,24 @@ def remove_known_host(reactor, hostname):
     return run(reactor, ['ssh-keygen', '-R', hostname])
 
 
-def run_tests(reactor, nodes, control_node, agent_nodes, dataset_backend,
-              trial_args, certificates_path):
+def get_trial_environment(cluster):
```
adamtheturtle: Docstring

admin/acceptance.py
```diff
@@ -572,23 +590,20 @@ def main(reactor, args, base_path, top_level):
...
-            }
-
-            for environment_variable in environment_variables:
-                print "export {name}={value};".format(
-                    name=environment_variable,
-                    value=environment_variables[environment_variable],
-                )
+            try:
+                cluster
+            except NameError:
```
itamarst: Aieeeeee

itamarst:
Was going to review when acceptance tests pass, but they're failing... in useful way looks like? So address and resubmit (and coordinate with Dave's work).

tomprince:
The failing tests are also failing on master.

admin/acceptance.py
itamarst: Using NameError or AttributeError like this is a recipe for bugs, since it can't distinguish typos from expected variable. Instead you can do:

```
cluster = None
try:
    cluster = yield runner.start_cluster()
    ....
finally:
    if cluster is not None: ...
```

itamarst:
Address, and then merge when both vagrant and non-vagrant builds are happy; I can't actually see a passing vagrant build right now because of Unrelated Reasons.

tomprince:
Merging with vagrant failure, since the builder has never succeeded.

PR #1473, opened 2015-05-29
[FLOC-1925] Storage driver configuration

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-1925


flocker/node/script.py
```diff
@@ -333,7 +334,13 @@ def get_configuration(options):
...
     configuration = yaml.safe_load(agent_config.getContent())
 
-    validate_configuration(configuration=configuration)
+    # FLOC-1925 remove this.
+    # It doesn't really work at the moment:
+    # https://github.com/ClusterHQ/flocker/pull/1440#issuecomment-105502142
+    # And if we want to support pluggable backends the config validation
+    # has to be pluggable too.
+    # JSON Schema style validation can be done
+    # validate_configuration(configuration=configuration)
```
exarkun: Instead of deleting this, I'd suggest just making it more lenient.  Let it enforce the parts that aren't about the storage driver.  Make the storage driver requirements wide open (a `dataset` section with anything at all in it).

flocker/node/script.py
```diff
@@ -396,6 +403,55 @@ class DeployerType(Names):
...
+    )
+    nova_client = NovaClient(
+        session=session, region_name=region, version=2
+    )
+
+    return cinder_api(
+        cinder_client=cinder_client,
+        nova_client=nova_client,
+        cluster_id=config['cluster_id']
+    )
```
exarkun: This function looks familiar.  Can you refactor this and `_openstack` from `blockdevicefactory.py` to remove the duplication?

flocker/node/script.py
```diff
@@ -396,6 +403,55 @@ class DeployerType(Names):
...
+    """
+    return EBSBlockDeviceAPI(
+        ec2_client=ec2_client(
+            region=config['region'],
+            zone=config['zone'],
+            access_key_id=config['access_key_id'],
+            secret_access_key=config['secret_access_key'],
+        ),
+        cluster_id=config['cluster_id']
+    )
```
exarkun: Ditto for this regarding `_aws` in `blockdevicefactory.py`.

admin/acceptance.py
```diff
@@ -547,12 +547,19 @@ def main(reactor, args, base_path, top_level):
 
         control_node = nodes[0]
         dataset_backend = options.dataset_backend
+        # FLOC-1925 get the backend config and credentials from the
+        # supplied config.
```
exarkun: From `options["config-file"]`?  Do we have an example/other docs for the structure of that file?

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -27,6 +27,7 @@ The :program:`admin/run-acceptance-tests` script has several options:
 .. option:: --dataset-backend <dataset-backend>
 
    Specifies what dataset backend to use for the cluster.
+   # FLOC-1925 Add a note about available backends.
```
exarkun: I'm not totally happy that we have duplicate CLI parameter documentation at all.  Maybe instead of adding more duplication here, just add a note that `run-acceptance-tests --help` is the canonical source for usage documentation.

agent.yml.sample.rackspace
```diff
@@ -0,0 +1,11 @@
...
+    hostname: 'localhost'
+    port: 4524
+dataset:
+    backend: "openstack"
+    region: "LON"
+    auth_plugin: "rackspace"
+    username: "joe.bloggs.rackspace"
+    api_key: "aaa-bbb-ccc-ddd"
+    auth_url: "https://identity.api.rackspacecloud.com/v2.0"
+version: 1
```
exarkun: Probably not the final location for these two sample files, right?  They look like they belong in the documentation somewhere.  Some minimal level of automated testing, maybe just a smoke test that some kind of object comes back from `AgentService.get_api` for these configurations (without raising an exception) would be good, too.

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -119,6 +120,7 @@ The configuration file for the Rackspace provider looks like:
      username: <rackspace username>
      key: <access key>
      keyname: <ssh-key-name>
+     # FLOC-1925 Add auth_url, auth_plugin, provider
```
exarkun: These extra pieces weren't necessary before because the acceptance test runner just ran the ZFS backend on a Rackspace VM, right?  And that didn't require using the OpenStack API - just libcloud with username and access key.  But now we want to run against the Cinder storage driver.  Does this imply that `run-acceptance-tests` will only support testing the "native" storage driver on each cloud provider?

exarkun: If so, that's fine for now.  Cinder is the only storage driver that could work on Rackspace right now (apart from ZFS which doesn't count).  Maybe later we'll want to have something more like the `dataset` section from `agent.yml` in this file as well to separate "how to get a machine to test on" from "how to get a storage driver to test against".

docs/using/installing/index.rst
```diff
@@ -497,6 +497,7 @@ The ZFS ``pool`` variable should match the pool name created in the ZFS section.
 
 .. The following is put in to demonstrate how to format alternative backends.
    Once OpenStack or EBS is added, the loopback device can be removed, as it is only for testing.
+   FLOC-1925 Replace the documentation below
```
exarkun: Replace what with what?

docs/using/installing/index.rst
```diff
@@ -506,6 +507,34 @@ For a ``loopback`` device, change the ``dataset`` clause to:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: yaml
+
+   dataset:
+       backend: "aws"
+       region: "us-west-1"
+       zone: "us-west-1a"
+       access_key_id: "foo"
+       secret_access_key: "bar"
```
exarkun: Any way we can test this?  Can it be some kind of include of data that gets used by that smoke test I mentioned elsewhere?  Or ... doctest ...

exarkun: (Sphinx has doctest support ... but probably not good support for testing yaml data instead of Python code ... so nevermind that probably.)

flocker/node/agents/cinder.py
```diff
@@ -223,6 +223,9 @@ def _extract_nova_server_addresses(addresses):
     return all_addresses
 
 
+# FLOC-1925 Maybe make this a PRecord with a unicode typed field for
+# cluster_id. Would catch errors such as:
+# https://github.com/ClusterHQ/flocker/commit/bc5d4c61317241c9779d4bbe07c744c6dd297ac5
```
exarkun: Ideally the tests would catch that error.  A unicode-enforcing field would still let you pass u"cluster-1" everywhere but if you actually do that the code is badly broken.

exarkun: So I'd say do this change as a follow-up and add a proper automated test as well as a unicode-only field.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -166,6 +166,9 @@ def _openstack(**config):
     a ``CinderBlockDeviceAPI``.  They will be configured to use the region
     where the server that is running this code is running.
 
+    # FLOC-1925 Make this a public helper function
+    # Move the REGION from environment parts elsewhere.
+
```
exarkun: So .. yes.  This is the refactoring I suggested earlier in the review I guess.  glad we're thinking the same thing. :smile: 

This should be public in a different module though.  `test` packages don't really get to have public interfaces.

exarkun: I suggest `flocker.node.agents.cinder`.

exarkun: For minimal disruption, the region-from-environment stuff should probably continue to live in exactly this function but all the rest of the logic should be delegated to the public API that can be shared with the production code-path for config loading.

exarkun:
Thanks.  This looks like it will work.  Go ahead.

PR #1476, opened 2015-05-29
Replace hardcoded 14.04 in Ubuntu node instructions

adamtheturtle:
This has been tested with an Ubuntu 14.04 container, as the output of ``lsb_release --release --short`` is ``14.04``. When the repository doesn't exist (if we change ``$(lsb_release --release --short)`` to ``foo``) then the error is much clearer than the datetime issue shown on the parent:

``sudo apt-get update`` gives:

```
W: Failed to fetch https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/foo/amd64/Packages  HttpError403

E: Some index files failed to download. They have been ignored, or old ones used instead.
```

at the end. Then:

```
root@3fb2482ad88f:/# sudo apt-get -y --force-yes install clusterhq-flocker-cli
Reading package lists... Done
Building dependency tree
Reading state information... Done
E: Unable to locate package clusterhq-flocker-cli
```

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -45,7 +45,7 @@ CLUSTERHQ_REPO = {
                     ),
     # FLOC-1828 TODO - use ubuntu rather than ubuntu-testing
     'ubuntu-14.04': 'https://{archive_bucket}.s3.amazonaws.com/ubuntu-testing/'
-                    '$(lsb_release --release --short)/$(ARCH)'.format(
+                    '$(lsb_release --release --short)/\\$(ARCH)'.format(
                         archive_bucket=ARCHIVE_BUCKET
                     ),
 }
@@ -156,10 +156,8 @@ def install_cli_commands_ubuntu(distribution, package_source):
             "apt-get", "-y", "install", "apt-transport-https",
             "software-properties-common"]),
         # Add ClusterHQ repo for installation of Flocker packages.
-        sudo_from_args([
-            'add-apt-repository', '-y',
-            'deb {} /'.format(CLUSTERHQ_REPO[distribution])
-            ])
+        sudo(command='add-apt-repository -y "deb {} /"'.format(
+            CLUSTERHQ_REPO[distribution])),
         ]
 
     if use_development_branch:
@@ -616,11 +614,9 @@ def task_install_flocker(
             run_from_args([
                 "add-apt-repository", "-y", "ppa:james-page/docker"]),
             # Add ClusterHQ repo for installation of Flocker packages.
-            run_from_args([
-                'add-apt-repository', '-y',
-                'deb {} /'.format(CLUSTERHQ_REPO[distribution])
-                ])
-            ]
+            run(command='add-apt-repository -y "deb {} /"'.format(
+                CLUSTERHQ_REPO[distribution])),
+        ]
 
         if use_development_branch:
             # Add BuildBot repo for testing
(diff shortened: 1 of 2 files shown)


itamarst:
There are some relevant failing tests.

flocker/provision/_install.py
```diff
@@ -44,9 +44,9 @@
                     archive_bucket=ARCHIVE_BUCKET,
                     ),
     # FLOC-1828 TODO - use ubuntu rather than ubuntu-testing
-    'ubuntu-14.04': 'https://{archive_bucket}.s3.amazonaws.com/'
-                'ubuntu-testing/14.04/$(ARCH)'.format(
-                    archive_bucket=ARCHIVE_BUCKET
+    'ubuntu-14.04': 'https://{archive_bucket}.s3.amazonaws.com/ubuntu-testing/'
+                    '$(lsb_release --release --short)/\\$(ARCH)'.format(
```
itamarst: The `$ARCH` thing confuses me.

1. Why is it there at all?
2. Adding the extra backslashes seems wrong, presuming it actually needs to be there.

itamarst:
Address, make buildbot happy, merge.

flocker/provision/_install.py
adamtheturtle: I didn't write `$(ARCH)` in originally but I think that it is because we build packages for the `amd64` architecture only and put packages in a directory named for that.

jongiddy: This value is evaluated as a shell command. The ` `$(lsb...)` part must be substituted as part of that command evaluation.  The `$(ARCH)` part must be left unevaluated, hence the backslash escapes (one to make shell ignore the `$` as a substitution marker, and then doubled to make Python ignore the `\` as an escape marker).

The output of this value then goes into `/etc/apt/sources.list` which does its own substitution on `$(ARCH)` during a subsequent `apt-get update`.

flocker/provision/_install.py
```diff
@@ -44,9 +44,17 @@
...
-                    archive_bucket=ARCHIVE_BUCKET
+    # This could hardcode the version number instead of using ``lsb_release``
+    # but that allows instructions to be shared between versions, and for
+    # earlier error reporting if you try to install on a separate version.
+    # The $(ARCH) part must be left unevaluated, hence the backslash escapes
+    # (one to make shell ignore the $ as a substitution marker, and then
+    # doubled to make Python ignore the \ as an escape marker).
+    # The output of this value then goes into /etc/apt/sources.list which does
+    # its own substitution on $(ARCH) during a subsequent apt-get update
+    'ubuntu-14.04': 'https://{archive_bucket}.s3.amazonaws.com/ubuntu-testing/'
```
exarkun: Shell quoting should probably done at the place where the value is used with a shell, not here.

flocker/provision/_install.py
```diff
@@ -44,9 +44,17 @@
...
+                    '$(lsb_release --release --short)/\\$(ARCH)'.format(
```
exarkun: Ahah.  But of course this makes things so much more complicated.  So this is a string in which **some** pieces are meant to be interpreted by a shell and some pieces are meant to be interpreted by ... a different shell.

Um.  Yea, I dunno.  Probably the right thing to do is replace `$(lsb_release --release --short)` with `{architecture}` and do Python-side determination and substitution of that value instead of asking the shell to do it.  Then the whole value can be shell-escaped and passed to `apt-add-repository` (or whichever).  Possibly that's not worth the effort right now though.

adamtheturtle: That isn't architecture, that is version. The point of this is that when @binocarlos copied the instructions with 14.04 into his shell on 14.10 (thinking he was on 14.04) he ran into the `datetime` issue mentioned on the parent issue. This gives an error earlier in that case, as per the description on this issue.

adamtheturtle:
I think that the errors are unrelated, the old related errors are resolved and Jon has answered the question and I've put it in a comment.

PR #1477, opened 2015-05-29
[FLOC-2036] Updates to Install documentation based on Lukes feedback

davidoliverSP2:
https://docs.google.com/document/d/1A52Jv9CXHMJJ2__skq4E4vuZs0Ng4-J0Rc0-1isj1Hc/edit

I've removed the image from this file, as the one used in this PR is still due to come back from fffunction as a vectored image. See FLOC-2077

Fixes https://clusterhq.atlassian.net/browse/FLOC-2036


lukemarsden:
LGTM, please merge!

docs/using/installing/index.rst
```diff
@@ -4,17 +4,25 @@
...
+The following diagram illustrates the architecture of the cluster you are about to install:
 
-There is also a ``clusterhq-flocker-node`` package which is installed on each node in the cluster.
-It contains the services that need to run on each node.
+.. image:: images/installintro.png
+   :width: 75%
+   :alt: Diagram illustrating the Flocker CLI and control service and agents.
 
-.. note:: The ``clusterhq-flocker-node`` package is pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
+The CLI is installed on your local machine via the ``flocker-cli`` package which provides command line tools to control the cluster. This package also includes the ``flocker-ca`` tool, which you must use to generate certificates for all the Flocker components.
```
jongiddy: Another thing we didn't tell you, and it took me ages to break the habit, is that in .rst files we put each sentence on a separate line.  This doesn't affect the final output, but makes future changes easier to review with `git diff`.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,25 @@
...
+
+The Flocker agents are installed on any number of hosts (or nodes) in the cluster where your containers will run.
+The agent software is included in the ``clusterhq-flocker-node`` package.
+
+There is also a Flocker control service which you must install on one of the agent hosts, or on a separate machine. The control service is also included in the ``clusterhq-flocker-node`` package, but is activated separately later in these installation instructions.
```
jongiddy: Separate here as well

docs/using/installing/index.rst
```diff
@@ -340,13 +348,22 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
 .. code-block:: console
 
    $ flocker-ca create-control-certificate example.org
    Created control-example.org.crt. Copy it over to /etc/flocker/control-service.crt on your control service machine and make sure to chmod 0600 it.
    
+Before copying these files you will need to SSH into the node and create a /etc/flocker directory:
+
+.. code-block:: console
+
+   root@mercury:~/$ mkdir -p /etc/flocker
```
jongiddy: Generally, we're using a prompt `[root@node]#` for logging into remote nodes.

Also, we are mostly preferring the `prompt` directive over the `code-block` directive.

I'd also add the `chmod` from the subsequent text, to help ensure the security practice is at least partly followed.

so my directive would be:
```rst
.. prompt:: bash [root@node]#
   mkdir -p /etc/flocker
   chmod 700 /etc/flocker
```

docs/using/installing/index.rst
```diff
@@ -340,13 +348,22 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
+
 You will need to copy both ``control-example.org.crt`` and ``control-example.org.key`` over to the node that is running your control service, to the directory ``/etc/flocker/`` and rename the files to ``control-service.crt`` and ``control-service.key`` respectively.
 You should also copy the cluster's public certificate, the `cluster.crt` file.
```
jongiddy: Add double-backquotes (as used for `control-service.crt`) around `cluster.rst`. (I realise it isn't part of this PR, but it's a minor fix.)

docs/using/installing/index.rst
```diff
@@ -379,6 +396,8 @@ Copy the generated files to ``/etc/flocker/`` on the target node and name them `
 Perform the same ``chmod 600`` commands on ``node.key`` as you did for the control service in the instructions above.
 The ``/etc/flocker/`` directory should be set to ``chmod 700``.
 
+You should now have ``cluster.crt``, ``node.crt``, and ``node.key`` on each of your agent nodes, and ``control-service.crt`` and ``control-service.key`` on your control node.
```
jongiddy: Add `cluster.crt` for control node as well, as it is needed in the case that the control node is not an agent node (and is merely redundant if the control node is an agent node).

jongiddy:
Thanks, this is OK to merge once the comments have been addressed, and with a final Buildbot test check.

adamtheturtle:
Before merging, let's think about how these images will be dealt with in the future. This may have been done already and I don't know about it.

The documentation is written in a way which attempts to be maintainable. This means that we acknowledge that someone (maybe future David or future Adam) will come along and want to fix a problem, or add a feature. Maybe a logo will change, for example. Storing PNGs without their source, or a link to their source, is not maintainable. It means that a new contributor without knowledge of where the image came from will not be able to e.g. rename the Flocker Control Service. We will either have to redraw the whole image or hunt down the source via someone's memory of where it is stored.

I'm not sure what the ideal solution to this is but I think that it warrants thought or discussion.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
 ==================
 
-As a user of Flocker you will need to install the ``flocker-cli`` package which provides command line tools to control the cluster.
-This should be installed on a machine with SSH credentials to control the cluster nodes
-(e.g., if you use our Vagrant setup then the machine which is running Vagrant).
+The following diagram illustrates the architecture of the cluster you are about to install:
 
-There is also a ``clusterhq-flocker-node`` package which is installed on each node in the cluster.
-It contains the services that need to run on each node.
+.. image:: images/installintro.png
```
adamtheturtle: I don't think that this is a descriptive name for the diagram, it describes where the diagram is. Also I have a personal preference for `some_kind-of delimiterBetween_Words`.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
+   :width: 75%
+   :alt: Diagram illustrating the Flocker CLI and control service and agents.
```
adamtheturtle: This is not suitable alt text. Alternative text should provide the reader with the information provided by the image. There are many conflicting views on exactly what is best practice for alt text, but the RNIB link to http://webaim.org/techniques/alttext/#context for example. Try to put yourself in the position of someone who cannot see the image - that person would preferably have all of the information that a fully sighted person has at the end of reading this document. If everything is explained in the text already then there may be no need for alt text.

adamtheturtle:
Looking at the built documentation, at http://build.clusterhq.com/results/docs/install-ubunturackspace-FLOC-2036/build-7339/using/installing/index.html the first thing I see on that page is the diagram. It is confusing and unclear to me:

1. I have a Data Center? I assume that this means that I will be installing Flocker in a data center.
2. It looks like "Flocker Containers API or CLI" is a heading. I then got confused because I happen to know Flocker's architecture and wanted to know where my laptop was. An important thing is that the CLI is running n my laptop. But an API isn't running anywhere - it is accessed from my laptop.
3. What are C1 and C2 and Cn. I assume containers, and I can have as many as I like so Cn represents Container 100 for example.
4. What are V2 V2 and Vm. I assume volumes. But why m and not n like containers? Or are n and m just placeholders for any number and but they are different letters to show that they are different numbers. I am thinking too hard to understand this - and I already understand the architecture (ish :smile:).
5. What are S1, S2, Sm? m implies, following the logic above, that there are as many Ss as there are Vs.
6. The lines everywhere are confusing. It looks like I can get from the Flocker Agent to the shared storage backend by going straight down through the DAEMON, C, V, and S. Or I can go around.

Also it has some text which is uncomfortably small for my liking.

Sorry to contradict the above but I think that this needs more work and another review. Perhaps split it up so that the accepted stuff can be merged.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
 Installing Flocker
 ==================
 
-As a user of Flocker you will need to install the ``flocker-cli`` package which provides command line tools to control the cluster.
-This should be installed on a machine with SSH credentials to control the cluster nodes
-(e.g., if you use our Vagrant setup then the machine which is running Vagrant).
+The following diagram illustrates the architecture of the cluster you are about to install:
```
adamtheturtle: I'm also installing a CLI, not just a cluster, right?

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
+The CLI is installed on your local machine via the ``flocker-cli`` package which provides command line tools to control the cluster. 
+This package also includes the ``flocker-ca`` tool, which you must use to generate certificates for all the Flocker components.
+
+The Flocker agents are installed on any number of hosts (or nodes) in the cluster where your containers will run.
+The agent software is included in the ``clusterhq-flocker-node`` package.
+
+There is also a Flocker control service which you must install on one of the agent hosts, or on a separate machine. 
+The control service is also included in the ``clusterhq-flocker-node`` package, but is activated separately later in these installation instructions.
+
+.. note:: The agents and control service are pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
```
adamtheturtle: Not new but `tutvagrant` is not a good name. We are not that limited in characters and can use things like `vagrant-tutorial`. Elsewhere we have been using `-` between words.

docs/using/installing/index.rst
jongiddy: I think it's OK to consider the CLI part of the cluster for this sentence.  It's sometimes better to be succinct than perfectly precise.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
 
-There is also a ``clusterhq-flocker-node`` package which is installed on each node in the cluster.
-It contains the services that need to run on each node.
+.. image:: images/installintro.png
+   :width: 75%
+   :alt: Diagram illustrating the Flocker CLI and control service and agents.
 
-.. note:: The ``clusterhq-flocker-node`` package is pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
+The CLI is installed on your local machine via the ``flocker-cli`` package which provides command line tools to control the cluster. 
+This package also includes the ``flocker-ca`` tool, which you must use to generate certificates for all the Flocker components.
```
adamtheturtle: Personally I like to avoid "you" in documents. I guess that this is a style thing that you will decide upon. But "you must" is unpleasant to read. Perhaps in the context of "to do X you must first do Y" but my reaction to this is "MUST I?".

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
+
+The Flocker agents are installed on any number of hosts (or nodes) in the cluster where your containers will run.
```
adamtheturtle: Let's choose one of hosts and nodes and stick to it - the mental model is complex enough already without having multiple words for the same thing.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,27 @@
...
+The agent software is included in the ``clusterhq-flocker-node`` package.
+
+There is also a Flocker control service which you must install on one of the agent hosts, or on a separate machine. 
+The control service is also included in the ``clusterhq-flocker-node`` package, but is activated separately later in these installation instructions.
+
+.. note:: The agents and control service are pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
 
 .. note:: If you're interested in developing Flocker (as opposed to simply using it) see :ref:`contribute`.
 
+This document will describe how to install the CLI locally and install the agents and control service on Vagrant or cloud infrastructure.
```
adamtheturtle: Not quite right I don't think. I think that it will describe how to install the agents and control service on cloud infrastructure, but I think that it will also show you how to get Vagrant nodes started which already have these services running. I'm not 100% sure on this though.

docs/using/installing/index.rst
```diff
@@ -219,11 +229,11 @@ Using Rackspace
 Another way to get a Flocker cluster running is to use Rackspace.
 You'll probably want to setup at least two nodes.
 
-#. Create a new Cloud Server running Fedora 20
+#. Create a new Cloud Server
 
    * Visit https://mycloud.rackspace.com
    * Click "Create Server".
-   * Choose the Fedora 20 Linux distribution as your image.
+   * Choose a supported Linux distribution as your image (for example, Ubuntu 14.04 or CentOS 7).
```
adamtheturtle: How do I know what else is OK? As a reader I am only going to pick one of those two, or I am going to go hunting for a full list.

adamtheturtle: Perhaps an TOC like index for links to the headings below.

docs/using/installing/index.rst
```diff
@@ -219,11 +229,11 @@ Using Rackspace
 Another way to get a Flocker cluster running is to use Rackspace.
 You'll probably want to setup at least two nodes.
 
-#. Create a new Cloud Server running Fedora 20
+#. Create a new Cloud Server
```
adamtheturtle: Why is "Cloud Server" title case (and not "new"). I don't mind as long as we are consistent within and between titles.

docs/using/installing/index.rst
adamtheturtle: In my opinion not in this case, and particularly not when it directly contradicts the diagram which it describes - that shows Flocker Cluster separate to the CLI.

lukemarsden:
@adamtheturtle We know the diagram is not optimal, but David made a note to talk to Alice about improving it with our design agency (based on ad-hoc feedback we got from showing the new instructions to Kai). Regarding maintainability/editability of diagrams, that's an ongoing issue which @davidoliverSP2 will need to be involved in.

In general I feel that this PR as it stands is a great incremental improvement to the installation docs, so I would be in favour of filing follow-up issues to the specific points raised and moving on. I'll reply to the specifics now.

docs/using/installing/index.rst
```diff
@@ -340,13 +350,22 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
+If you have a domain name for your cluster, replace ``example.org`` with the hostname of your control service node.
+You can also use an IP address here.
+For TLS verification purposes this hostname must match the one you will give to REST API clients.
 
 .. code-block:: console
 
    $ flocker-ca create-control-certificate example.org
    Created control-example.org.crt. Copy it over to /etc/flocker/control-service.crt on your control service machine and make sure to chmod 0600 it.
    
+Before copying these files you will need to SSH into the node and create a /etc/flocker directory:
```
adamtheturtle: Sphinx has a role for this - `:file:`/etc/flocker`. We haven't been great at using it but the other docs at least use double backticks.

docs/using/installing/index.rst
lukemarsden: This sort of tone-of-voice issue is up to @davidoliverSP2 to have a position on. :)

docs/using/installing/index.rst
lukemarsden: The image refers to hosts and the rest of the doc refers to nodes, hence this imperfect association of the two. When the image is redone perhaps it could refer to nodes and then this sentence replaced by "nodes" only @davidoliverSP2.

docs/using/installing/index.rst
adamtheturtle: Not all feedback must be acted on - I still think that it is useful to mention opinion.

docs/using/installing/index.rst
lukemarsden: Seems reasonable.

docs/using/installing/index.rst
lukemarsden: Yes, we mulled over this -- until we have an autogenerated list of supported platforms this will have to do.

adamtheturtle: Elsewhere when we want to improve things later we use a comment with "XXX This is what to improve now FLOC-XYZ" where FLOC-XYZ is a follow up issue. It helps. Also, sometimes it is helpful for a reviewer if the compromises made are put as comments on the PR. Else I'm saying "X would be nice" and reading "we thought of X, but didn't want to spend the time" - when not spending the time is a perfectly reasonable decision to make, but it'd be nice for me not to have to do the same thinking as the implementer.

docs/using/installing/index.rst
adamtheturtle: Or, as mentioned above, the `:file:` role.

docs/using/installing/index.rst
adamtheturtle: As I know you use chrome @davidoliverSP2 - test your prompt changes in Firefox to avoid confusion as there is a known bug. Prompt has a number of advantages, but mainly that the prompt is not copied even when selected, so you can drag over multiple lines and copy multiple commands.

In the context of these docs, `mercury` is `alice`'s local computer, to differentiate from `node`.

docs/using/installing/index.rst
jongiddy: In the text, use "nodes (hosts)". This indicates the preferred term, indicates that a single node generally maps to a (physical or virtual) host, and overcomes the mismatch in the diagram. When the diagram is modified, the text can then stay the same.

adamtheturtle: Link to a follow-up issue if this is preferred.

exarkun:
> In general I feel that this PR as it stands is a great incremental improvement to the installation docs, so I would be in favour of filing follow-up issues to the specific points raised and moving on. I'll reply to the specifics now.

I'm really opposed to the inclusion of a PNG in our source tree.  Where is the source material for this image?

lukemarsden:
> I'm strongly opposed to the inclusion of a PNG in our source tree (iow: don't merge this). Where is the source material for this image?

I'll leave this for @davidoliverSP2 to address on Monday.

jongiddy:
Previously, I used a Google Code Subversion repo for large binary files, that were retrieved as needed using `wget`, but that specific option is no longer available.

Any suggestions on what to do with PNG's and similar files such that we can version control them, and link them easily to the Git repo, but not bloat the Git repo?

adamtheturtle:
@jongiddy Git has a feature for this, https://git-lfs.github.com. I haven't worked on it.

lukemarsden:
@jongiddy I think @exarkun's concern was about editability of the images, rather than the size of them. Including relatively small images in our source tree is probably fine, but indeed they need to be editable (Alice has already raised this issue with our designers), and we need to have some sort of defined process (or at least a plan) for managing them.

exarkun:
> Any suggestions on what to do with PNG's and similar files such that we can version control them, and link them easily to the Git repo, but not bloat the Git repo?

My primary concern is that whoever put this image together did not create it as a PNG.  They created it using a featureful piece of graphics editing software in a way that created a ton of metadata which is all discarded by the PNG format, rendering the image 100% immutable.  That makes this diagram unmaintainable.  The correct thing to do is check in an SVG (one or two other formats may be feasible) and generate raster image formats from that source material as necessary.  This has the happy consequence of keeping the repository smaller but the repository is also many dozens of MB so ... eh.

adamtheturtle:
Do we not need `something.png` and `something@2x.png`? `@3x.png`?

docs/using/installing/index.rst
davidoliverSP2: I've made a note of this in the style guidelines document I've started: https://docs.google.com/spreadsheets/d/11E38SvKX7ujJ95qs7rvzgh3vIVmPyiTOSsvjR8KWMoY/edit#gid=0
I've been used to using shorthand for ref tags, but happy to implement a clear standard. I'll go through the whole doc set post v1.0 and clean up the ref tags.

adamtheturtle: @davidoliverSP2 Great to have a style guide. We used to have one for our old product's documentation. It was by no means comprehensive, but I'll send it to you in a private message, I think it is a good place to start.

In Flocker's Python code, we have style preferences but many of them are enforced as you know by the flake8 lint builder. This means that a build won't pass unless certain style rules are adhered to. I have proposed in the past (on JIRA I thought but I can't find it) that we use a tool similar to flake8 but for restructured text. I haven't yet tried any myself but https://pypi.python.org/pypi/restructuredtext_lint and https://pypi.python.org/pypi/doc8 exist. Having automatic linting means faster and less tedious review cycles.

docs/using/installing/index.rst
```diff
@@ -340,16 +350,26 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
-You should also copy the cluster's public certificate, the `cluster.crt` file.
-On the server, the ``/etc/flocker`` directory and private key file should be set to secure permissions via ``chmod``:
+Before copying these files you will need to SSH into the node and create a :file:`/etc/flocker` directory:
+
+.. prompt:: bash [root@node]#
+
+   mkdir -p /etc/flocker
+   chmod 700 /etc/flocker
+
+You will need to copy both ``control-example.org.crt`` and ``control-example.org.key`` over to the node that is running your control service, to the :file:`/etc/flocker` directory and rename the files to ``control-service.crt`` and ``control-service.key`` respectively.
```
adamtheturtle: These can also be use the `:file:` role - this is not the case everywhere in our docs but I think that it should be (at least for new additions).

docs/using/installing/index.rst
```diff
@@ -340,16 +350,26 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
+You should also copy the cluster's public certificate, the ``cluster.crt`` file.
+On the server, the :file:`/etc/flocker` directory and private key file should be set to secure permissions via ``chmod``:
```
adamtheturtle: `chmod` can use the `:command:` role - http://sphinx-doc.org/markup/inline.html#role-command. Again, we haven't been great at using roles but we should get better.

docs/using/installing/index.rst
davidoliverSP2: For consistency across the docs, we'll continue to address the user directly (there isn't enough time to go through the docs to change this) I'm happy to have a docs "tone" discussion post v1 when I would have more time to implement any changes. A good point to raise though, cheers

exarkun:
> Do we not need something.png and something@2x.png? @3x.png?

I'm not 100% sure I know what you mean.  Is that a naming convention for image files on the web to do with scaling?  So, a version of the png at one size, at twice the size, and at three times the size?

I don't know - do we?

adamtheturtle:
@exarkun It is necessary I think for high DPI displays with scaling.

> So, a version of the png at one size, at twice the size, and at three times the size?

Yes.

> I don't know - do we?

Some of the images in our repo have "@2x" and I think (almost) all on the website do too. "@3x" is pretty new I think. @lukemarsden ?

exarkun:
And we're discussing this because of the implications for repository size?

lukemarsden:
fwiw, I'm happy with having just one high res PNG version of an image and relying on browsers to downscale nicely (they do these days) when a relative width is specified. PNG is quite good at compressing upscaled images so it doesn't affect users on slow connections too badly.

As long as it looks okay on a retina display (@davidoliverSP2 happens to have one, so he can verify) I'm okay with just having a single HiDPI image like we do presently. If you like we can document our minimum dpi requirements for docs images as a follow-up issue.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,21 @@
 Installing Flocker
 ==================
 
-As a user of Flocker you will need to install the ``flocker-cli`` package which provides command line tools to control the cluster.
-This should be installed on a machine with SSH credentials to control the cluster nodes
-(e.g., if you use our Vagrant setup then the machine which is running Vagrant).
+The Flocker CLI is installed on your local machine via the ``flocker-cli`` package which provides command line tools to control the cluster.
```
adamtheturtle: I think that this introduces confusion between `flocker-cli` and "The Flocker CLI". As I mentioned in the glossary I think we should stick to having them be synonymous. Particularly because I don't `brew install flocker-cli` - so a Mac user will never see `flocker-cli` on their machine.

adamtheturtle:
Now that the image isn't used in the docs use `git rm docs/using/installing/images/installintro.png` to remove it from this branch's additions.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,21 @@
...
 
-As a user of Flocker you will need to install the ``flocker-cli`` package which provides command line tools to control the cluster.
-This should be installed on a machine with SSH credentials to control the cluster nodes
-(e.g., if you use our Vagrant setup then the machine which is running Vagrant).
+The Flocker CLI is installed on your local machine via the ``flocker-cli`` package which provides command line tools to control the cluster. 
+This package also includes the ``flocker-ca`` tool, which you must use to generate certificates for all the Flocker components.
 
-There is also a ``clusterhq-flocker-node`` package which is installed on each node in the cluster.
-It contains the services that need to run on each node.
+The Flocker agents are installed on any number of nodes (or hosts) in the cluster where your containers will run.
```
adamtheturtle: This clarification was suggested by @jongiddy at https://github.com/ClusterHQ/flocker/pull/1477#discussion_r31351034 as a workaround to the image which has now been removed. Stick with hosts or nodes instead of both. I prefer nodes.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,21 @@
 Installing Flocker
 ==================
```
adamtheturtle: Before we explain how each bit is installed, perhaps keep the intention of the diagram which was to give an overview. In the getting started guide this is done with:

```
      Flocker Client
      --------------

      +--------------------------------------------------------------------------+
      | Runs on your laptop                                                      |
      +--------------------------------------------------------------------------+
      | Uses application and deployment configuration files                      |
      +--------------------------------------------------------------------------+
      | Deploys containers to a cluster of servers                               |
      +--------------------------------------------------------------------------+

      Flocker Node
      ------------

      +--------------------------------------------------------------------------+
      | Runs on each server in a cluster                                         |
      +--------------------------------------------------------------------------+
      | Links, ports and volumes work across hosts                               |
      +--------------------------------------------------------------------------+
      | After deployment, containers can move around                             |
      +--------------------------------------------------------------------------+
```

docs/using/installing/index.rst
```diff
@@ -4,17 +4,21 @@
...
 
-.. note:: The ``clusterhq-flocker-node`` package is pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
+There is also a Flocker control service which you must install on one of the agent hosts, or on a separate machine. 
+The control service is also included in the ``clusterhq-flocker-node`` package, but is activated separately later in these installation instructions.
+
+.. note:: The agents and control service are pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
 
 .. note:: If you're interested in developing Flocker (as opposed to simply using it) see :ref:`contribute`.
 
+This document will describe how to install the CLI locally and install the agents and control service on cloud infrastructure. It also describes how to Vagrant nodes started which already have these services running.
```
adamtheturtle: New sentence, new line. Make it clear that the cloud infrastructure vs Vagrant is a choice of one, not that you can do both.

docs/using/installing/index.rst
```diff
@@ -36,7 +39,7 @@ Before you install ``flocker-cli`` you will need a compiler, Python 2.7, and the
 On Fedora 20 you can install these by running:
 
 .. code-block:: console
-
+
```
adamtheturtle: Get rid of this whitespace - perhaps an editor can do this throughout for you.

docs/using/installing/index.rst
adamtheturtle: Please make and link to the follow up as per the previous comment.

docs/using/installing/index.rst
```diff
@@ -348,9 +356,9 @@ Using an IP address is not recommended as it may break some HTTPS clients.
 
    $ flocker-ca create-control-certificate example.org
 
-You will need to copy both ``control-example.org.crt`` and ``control-example.org.key`` over to the node that is running your control service, to the directory ``/etc/flocker/`` and rename the files to ``control-service.crt`` and ``control-service.key`` respectively.
+You will need to copy both ``control-example.org.crt`` and ``control-example.org.key`` over to the node that is running your control service, to the directory :file:`/etc/flocker` and rename the files to ``control-service.crt`` and ``control-service.key`` respectively.
```
adamtheturtle: These `.crt` and `.key`s are also `:file:`s.

adamtheturtle:
Great, thanks @davidoliverSP2 . Please address the minor comments then I'll have a quick check over before merging.

docs/using/installing/index.rst
```diff
@@ -4,17 +4,24 @@
...
+
+There is also a Flocker control service which you must install on one of the agent hosts, or on a separate machine. 
+The control service is also included in the ``clusterhq-flocker-node`` package, but is activated separately later in these installation instructions.
+
+.. note:: The agents and control service are pre-installed by the :ref:`Vagrant configuration in the tutorial <tutvagrant>`.
 
 .. note:: If you're interested in developing Flocker (as opposed to simply using it) see :ref:`contribute`.
 
+This document will describe how to install the CLI locally and install the agents and control service on cloud infrastructure.
+It also describes how to Vagrant nodes started which already have these services running.
```
adamtheturtle: Missing a word "how to Vagrant nodes started".

docs/using/installing/index.rst
```diff
@@ -219,15 +225,15 @@ Using Rackspace
 Another way to get a Flocker cluster running is to use Rackspace.
 You'll probably want to setup at least two nodes.
 
-#. Create a new Cloud Server running Fedora 20
+#. Create a new cloud server:
 
    * Visit https://mycloud.rackspace.com
    * Click "Create Server".
-   * Choose the Fedora 20 Linux distribution as your image.
+   * Choose a supported Linux distribution as your image (For available distributions, see the list of installation instructions in Step 3).
```
adamtheturtle: "(For ...)" is a continuation of the sentence so should not be a capital letter as it does not proceed a full stop. I'd say just have a full stop.

adamtheturtle:
Thanks @davidoliverSP2 I think that this is a good PR in that the improvements are of high quality and importantly it doesn't have a large scope. Splitting up tasks into small sub-tasks helps the author and reviewer. I'd like to test this by hand but don't have time right now. If you have done the steps as they stand now on Rackspace then I think that is good enough. Else, wait until a reviewer has time to do this.

Perhaps @tomprince has an opinion about whether these instructions (copying files, `chmod`ing things) should be (tested) `task`s.

If this has been manually tested by you and works, and Tom is satisfied with the task-less nature of this then this can be merged when the minor comments are addressed. Please check through all the so-called "outdated" comments. If there is anything missed then address them and resubmit.

wallrj:
Needs merging forward.

wallrj:
@davidoliverSP2 There are some remaining references to configuring the AWS firewall to allow SSH access to all the nodes and between the nodes.

> If you wish to customize the instance's security settings make sure to permit SSH access both from the intended client machine (for example, your laptop) and from any other instances on which you plan to install ``clusterhq-flocker-node``.

It might be worth changing this to say that if users want to customize the AWS firewall settings, they must remember to leave open the flocker-control TCP ports 4523 (REST API) and 4524 (agent connections) on the node running the control service.
> Allow SSH access for the root user, then log out.

> Log back into the instances as user "root", e.g.:

None of those are strictly necessary any more.
I mean the user will probably want SSH access in order to install and configure Flocker node but the  Flocker CLI tools do not need (root) SSH access to the nodes.
And the nodes only need (root) SSH access to each other when using ZFS.

Might be worth removing / de-emphasising / or expanding on those parts of the docs.

wallrj:
In the Rackspace section we say:

> Choose a supported Linux distribution as your image. For available distributions, see the list of installation instructions in Step 3.

Perhaps we should spell out "supported Linux distributions" right at the top of this document and link to it here. Or create a follow up issue for that.

wallrj:
At the end of the Fedora20 (soon to be removed) and  Centos-7 instructions, we say

> Finally, you will need to run the flocker-ca tool that is installed as part of the CLI package. This tool generates TLS certificates that are used to identify and authenticate the components of your cluster when they communicate, which you will need to copy over to your nodes. Please see the cluster authentication instructions.

But we don't say this at the end of the Ubuntu 14.04 instructions.

I think we should to make it clear to Ubuntu installers what the next step is.

wallrj:
Sorry to hijack this PR, but another point I'd like to make is that I find the "Cluster Authentication Layer Configuration" section very difficult to read and difficult to follow.
(try reading it out loud)

 * http://build.clusterhq.com/results/docs/install-ubunturackspace-FLOC-2036/build-7649/using/installing/index.html#cluster-authentication-layer-configuration

I'd rather see it was a numbered list of steps.

Maybe create a follow up issue for improving that section.

wallrj:
Another point which is probably the subject of a follow up issue.

"ZFS Backend Configuration" 

 * is misnamed; it actually describes how to install ZFSonLinux on various systems. The ZFS backend configuration instructions come later in the document under the title "ZFS Peer-to-Peer Backend Configuration (ALPHA)"
 * is currently too high up in the document.  ZFS is not a supported backend. "To enable the Flocker control service on Fedora / CentOS" should come next, after generating and installing certificates.

http://build.clusterhq.com/results/docs/install-ubunturackspace-FLOC-2036/build-7649/using/installing/index.html#zfs-backend-configuration

wallrj:
> You have now installed clusterhq-flocker-node and created a ZFS pool for it.

Isn't a suitable summary line any more.

docs/using/installing/index.rst
davidoliverSP2: FLOC-2169 created to respond to @wallrj comments, not already addressed.

wallrj:
Thanks @davidoliverSP2 

This looks good to me.

Please resolve the latest merge conflict, taking care to retain all the Fedora20 removal work, check the buildbot built docs and then merge.

PR #1478, opened 2015-05-29
[FLOC-1944] 3rd party HTTPS client tests for our REST API TLS server

itamarst:


flocker/ca/functional/test_validation.py
```diff
@@ -318,3 +330,62 @@ class RESTAPIContextFactoryValidationTests(
...
+    def test_requests_ip(self):
+        """
+        The ``requests`` library can communicate with the certificate used by
+        the REST API when it is accessed via an IP address.
+        """
+        return self.assert_requests_validates(b"127.0.0.1", "control")
+    test_requests_ip.skip = (
+        "This is a actually an expected failure, see "
+        "https://bugs.python.org/msg138405, "
+        "but the failing test is resulting in a deadlock somewhere.")
```
exarkun: This test should go into an issue in the tracker instead (ever-growing pile of skips crowding the end of trial output is a trial misfeature).

flocker/ca/functional/test_validation.py
```diff
@@ -318,3 +330,62 @@ class RESTAPIContextFactoryValidationTests(
...
+    """
+    Tests for the ``requests`` library's ability to talk to REST API TLS
+    server.
+    """
+    def assert_requests_validates(self, hostname, cert_name):
+        """
+        Assert that ``requests`` succeeds in doing a TLS query to our server.
+
+        :param hostname: Hostname to connect to.
+        :param cert_name: Which attribute of the ``ControlSet`` to use.
```
exarkun: types missing in parameter documentation

flocker/ca/functional/test_validation.py
```diff
@@ -318,3 +330,62 @@ class RESTAPIContextFactoryValidationTests(
...
+
+        def req():
+            return requests.get(
+                "https://{}:{}/".format(hostname, port),
+                headers={"connection": "close"},
+                stream=False,
+                cert=(certs.child(b"user.crt").path,
+                      certs.child(b"user.key").path),
+                verify=certs.child(b"cluster.crt").path).content
+        pool = ThreadPool()
```
exarkun: Pass `min` (and `max`?) here.  Default is to create 5 threads and this test only ever needs one.

docs/using/installing/index.rst
```diff
@@ -340,13 +340,15 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
 
 You are now able to generate authentication certificates for the control service and each of your nodes.
 To generate the control service certificate, run the following command from the same directory containing your authority certificate generated in the previous step.
-Replace ``example.org`` with the hostname of your control service node; this hostname should match the one you will give to REST API clients.
+Replace ``example.org`` with the hostname of your control service node; this hostname should match the hostname you will give to REST API clients.
```
exarkun: This is pre-existing ... but ... Are we really calling this the "REST API"?  It's an HTTP API.

docs/using/installing/index.rst
```diff
@@ -340,13 +340,15 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
+It should be a valid DNS name that HTTPS clients can resolve, since they will use it as part of TLS validation.
```
exarkun: And to talk about "REST API clients" in one breath and "HTTPS clients" in the next seems unnecessarily confusing to me.

exarkun: Also, extraneous comma.

docs/using/installing/index.rst
```diff
@@ -340,13 +340,15 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
...
+Using an IP address is not recommended as it may break some HTTPS clients.
 
 .. code-block:: console
 
    $ flocker-ca create-control-certificate example.org
    Created control-example.org.crt. Copy it over to /etc/flocker/control-service.crt on your control service machine and make sure to chmod 0600 it.
-   
+
 You will need to copy both ``control-example.org.crt`` and ``control-example.org.key`` over to the node that is running your control service, to the directory ``/etc/flocker/`` and rename the files to ``control-service.crt`` and ``control-service.key`` respectively.
```
exarkun: Yay, duplicate/conflicting documentation - incomplete/incorrect version in the command output, complete version in the docs no one will ever read (because they'll see the wrong information in the command output, get frustrated, and go away).

flocker/ca/functional/test_validation.py
```diff
@@ -158,15 +178,7 @@ def _handshake(self, credential):
...
-            server_endpoint = SSL4ServerEndpoint(reactor, port,
-                                                 server_context_factory,
-                                                 interface='127.0.0.1')
-            server_factory = WaitForDisconnectsFactory.forProtocol(
-                SendingProtocol
-            )
-            self.addCleanup(lambda: server_factory.wait_for_disconnects())
-            d = server_endpoint.listen(server_factory)
-            d.addCallback(lambda port: self.addCleanup(port.stopListening))
+            start_tls_server(self, port, server_context_factory)
```
exarkun: I guess this assumes that server endpoints synchronously listen... which isn't quite what the interface guarantees... though the TLS endpoint happens to do this.  Pre-existing, anyway.  File an issue if you don't fix it here.

flocker/ca/functional/test_validation.py
```diff
@@ -318,3 +330,62 @@ class RESTAPIContextFactoryValidationTests(
...
+                stream=False,
+                cert=(certs.child(b"user.crt").path,
+                      certs.child(b"user.key").path),
+                verify=certs.child(b"cluster.crt").path).content
+        pool = ThreadPool()
+        pool.start()
+        self.addCleanup(pool.stop)
+        d = deferToThreadPool(reactor, pool, req)
+        d.addCallback(self.assertEqual, EXPECTED_STRING)
+        return d
```
exarkun: It would be nice to have the most possible information when this test fails.  If I comment out the `verify` argument in the `requests.get` call above then I do see a nice (well, probably as nice as feasible) traceback ending with an `SSLError`.  If I comment out the `cert` section, though, then the TLS error is on the server side and doesn't seem to get logged at all.  An `SSLError` still shows up on the client but it's not nearly as informative as the first one.

Should we err on the side of logging too much about TLS errors in our server setup?  And should we do the same thing for the actual HTTPS server (slightly sad this test can't/doesn't share the relevant code with the actual HTTPS server...)?  Can we even do that without hacking through klein and Twisted Web? :/

I would expect that people trying out their own clusters and getting stuck on TLS problems might appreciate as much information as possible (or, as usual, we might appreciate it when they come asking us for help).  Making test failures easier to debug seems like a good thing.  Making real-world failures easier to debug seems like a _really_ good thing.

exarkun:
Thanks.  Looks good.  Please address and merge.

docs/using/installing/index.rst
itamarst: Actual output from the tool is better now (Dave merged a thing earlier).

flocker/ca/functional/test_validation.py
itamarst: We can't really do much to help users validate their custom 3rd party clients except suggest logging (and provide good docs and maybe client libraries).

I will improve logging on server, at least (we can do that in the context factory verify callback I think).

itamarst: Or rather, file an issue.

docs/using/installing/index.rst
```diff
@@ -340,13 +340,14 @@ The file ``cluster.key`` should be kept only by the cluster administrator; it do
 
 You are now able to generate authentication certificates for the control service and each of your nodes.
 To generate the control service certificate, run the following command from the same directory containing your authority certificate generated in the previous step.
-Replace ``example.org`` with the hostname of your control service node; this hostname should match the one you will give to REST API clients.
+Replace ``example.org`` with the hostname of your control service node; this hostname should match the hostname you will give to HTTP API clients.
+It should be a valid DNS name that HTTPS clients can resolve since they will use it as part of TLS validation.
+Using an IP address is not recommended as it may break some HTTPS clients.
```
lukemarsden: I'm concerned about this from a product perspective, because it makes owning a domain name and having administrative access to adding A records to it a prerequisite for trying out Flocker. Which clients have problems with this?

Also note that this is a merge conflict with https://github.com/ClusterHQ/flocker/pull/1477/files#diff-87e3ab44da633c377fd2a47b19dd220aR358

lukemarsden: Maybe we can encourage users in this situation to use http://xip.io/

exarkun: You can put a hostname in your `/etc/hosts` file.  Also, the cluster works fine with IP addresses in the certificate.  And so does reasonable HTTPS client software.  Just not **all** HTTPS client software.  Also, even with broken HTTPS client software, you can always turn off validation.

lukemarsden: OK, so let's document this in a way that is friendly to users who don't have access to a DNS provider or know what `/etc/hosts` is.

exarkun: https://clusterhq.atlassian.net/browse/FLOC-2048

PR #1479, opened 2015-05-29
[FLOC-2027] OpenStack logging

myechuri:
This branch combines FLOC-2027 and FLOC-1847 (https://clusterhq.atlassian.net/browse/FLOC-1847) since FLOC-1847 is a prereq for FLOC-2027.

Fixes:
 * https://clusterhq.atlassian.net/browse/FLOC-2027
 * https://clusterhq.atlassian.net/browse/FLOC-1847


myechuri:
flocker.common.test.test_interface
  InterfaceDecoratorTests
    test_decorated_exception ...                                           [OK]
    test_decorated_method ...                                              [OK]
    test_raise ...                                                         [OK]
    test_return ...                                                        [OK]

Ran 4 tests in 0.539s

PASSED (successes=4)

flocker/common/interface.py
```diff
@@ -0,0 +1,40 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
```
exarkun: Should be `_interface.py` and public parts re-exposed in `__init__.py` (because that's an idiom we seem to have adopted... so we should be consistent unless we decide to abandon it entirely).

flocker/common/interface.py
```diff
@@ -0,0 +1,40 @@
...
+
+"""
+Some interface-related tools.
+"""
+
+from zope.interface.interface import Method
+
+
+# TODO: Add tests and documentation for this, make it public (somewhere else).
+# https://clusterhq.atlassian.net/browse/FLOC-1847
```
exarkun: Looks like this is being done?  This comment can go.

flocker/common/interface.py
```diff
@@ -0,0 +1,40 @@
...
+            raise TypeError(
+                "{} does not support interfaces with non-methods "
+                "attributes".format(decorator_name)
+            )
+
+    def class_decorator(cls):
+        for name in interface.names():
+            setattr(cls, name, method_decorator(name, *args, **kwargs))
+        return cls
+    return class_decorator
```
exarkun: The previous implementation in `_openstack.py` should be removed.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+from twisted.trial.unittest import SynchronousTestCase
+
+from eliot.testing import (
+    LoggedMessage, assertContainsFields, capture_logging
+)
+from eliot import Field, MessageType
+
+from zope.interface import Interface, implementer
+
+from ..interface import interface_decorator
```
exarkun: After `interface` becomes `_interface` this should be changed to use the canonical, public name instead of importing from the private module.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+# Eliot structures for testing ``interface_decorator``.
+METHOD = Field.for_types(
+    u"method", [unicode],
+    u"The name of the decorated method.")
+TEST_MESSAGE = MessageType(u"flocker:common:test:interface:message", [METHOD],)
+TEST_EXCEPTION = MessageType(u"flocker:common:test:interface:exception",
+                             [METHOD],)
```
exarkun: I like trailing commas when they simplify diffs as they do when you decide to add another argument to:

```
foo(
    bar,
    baz,
)
```

But these trailing commas won't simplify any diffs, they just look weird. :smile:

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def raise_method(self):
+        raise self._result
+
+
+def _test_logged_method(method_name, original_name):
+    """
+    Decorator that logs a message to ``Eliot`` logger in two flows:
+    1) Log before calling given ``method_name``.
+    2) Log if ``method_name`` resulted in an ``Exception``.
```
exarkun: reStructuredText has a syntax for numbered lists - http://docutils.sourceforge.net/docs/user/rst/quickref.html#enumerated-lists - it might make sense to use it.  Also, a popular convention for docstrings is that the first sentence (ideally one line, but that's harder) is a summary of what the thing does.  Then a following paragraph offers more details.  It's kind of a nice way to arrange things but it's not mandatory.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+        self._dummy = dummy
+
+
+class InterfaceDecoratorTests(SynchronousTestCase):
+    """
+    Tests for ``interface_decorator``.
+    """
+    def test_return(self):
+        """
+        Decorated method returns the value returned by the original method.
```
exarkun: "A decorated ..." or "The decorated ..."

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        result = object()
+        logging_dummy = LoggingDummy(Dummy(result))
+        self.assertIs(result, logging_dummy.return_method())
+
+    @capture_logging(lambda self, logger: None)
+    def test_decorated_method(self, logger):
+        """
+        Decorated method logs expected text to ``Eliot``.
```
exarkun: Probably just "Eliot".  "\`\`" is for "inline literals"... which is a little vague but it probably doesn't cover human-friendly project names.  If this were instead "A decorated method logs expected text using ``eliot``." then I think the backticks would make perfect sense - since then the docstring is referring to the Python package by that name and that seems more like a thing you might consider a "literal".  Another way to put that rule of thumb might be that things in double backticks should maybe at least seem like you might be able to evaluate them as Python statements/expressions.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,158 @@
...
+
+        logged = LoggedMessage.of_type(
+            logger.messages, TEST_MESSAGE,
+        )[0]
+
+        assertContainsFields(
+            self, logged.message, {
+                "method": u"return_method",
+            },
+        )
```
exarkun: I think an intended use of the Eliot testing helpers (like `capture_logging`) is that you write just one test method (like `test_return`) and then you put interesting assertions like this one (roughly lines 120 to 128) into the function you pass to `capture_logging`).  Then you don't have to duplicate your application logic in order to exercise its logging behavior.  You just write one test and one logging assertion function and you get both pieces applied to one test run.

This is slightly different from the idea that you can write one test method and then put several `self.assertFoo(...)` calls at the end because any failures from the Eliot logging assertion function (the one you pass to `capture_logging`) are recorded **in addition to** the "normal" result of the test whereas in the case of multiple assertions of the normal type, the first one that fails ends execution of the test method and you don't get to find out what the rest would have done.

flocker/common/thread.py
```diff
@@ -0,0 +1,67 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
```
exarkun: Also should be private - `_thread.py`

flocker/common/thread.py
```diff
@@ -0,0 +1,67 @@
...
+
+"""
+Some thread-related tools.
+"""
+
+from twisted.internet.threads import deferToThreadPool
+
+from interface import interface_decorator
```
exarkun: This should be an "explicit relative import" instead of an "implicit relative import" (or, if Flocker didn't use explicit relative imports all over the place, it should have been an absolute import instead of an implicit relative import - but we do so ...).

Relative imports are ridiculously error prone.  They're a horrible misfeature of the language.  Just one of the terrible side-effects of doing this is that afterwards there are actually two copies of the interface module loaded.  One of them is named `flocker.common.interface` and another is called `interface`.  And they each define objects that look the same but they're actually distinct objects in distinct modules and this can lead to all kinds of nastiness...

So `from .interface import interface_decorator`.  And then, ideally, after `interface` becomes `_interface` and its function gets re-exposed in the `flocker.common` module, `from . import interface_decorator` - but you won't actually be able to do this since it will create a circular dependency between `flocker/common/__init__.py` and `flocker/common/thread.py`.  Very sad.  This is the main reason I'm not sure I like the idiom we use of having a bunch of private modules and then a `__init__.py` that imports a bunch of stuff from them.

flocker/node/agents/_logging.py
```diff
@@ -7,12 +7,8 @@
...
 from eliot import Field, ActionType, MessageType
 
-# Begin: Helper datastructures to log all
-# IBlockDeviceAPI calls from AWS storage driver using Eliot.
-# - Log all IBlockDeviceAPI calls as Eliot ``ActionType``.
-# - Log ``EC2ResponseError`` from AWS to Boto.
-# - Log failure of attached device manifestation
-# - Log OS out of available devices for attaching volume.
+# Begin: Common structures used by all (AWS, OpenStack)
+# storage drivers.
```
exarkun: I'm not a huge fan of this style of comment block.  People usually don't pay close enough attention when they're maintaining the code later on and the comments end up out-of-date with the implementation (ie, wrong).

If there's a bunch of related code and the grouping seems important enough to highlight then it's probably worth moving all of that code into a new module.  New modules aren't quite free but they're most likely cheap enough that it's not worth thinking about their cost.  This also forces you to think about a name to describe the collection and like to believe that thinking about names helps produce better abstractions by forcing you to understand what you're actually doing.

flocker/node/agents/_logging.py
```diff
@@ -94,3 +99,70 @@
     [VOLUME_ID, STATUS, TARGET_STATUS, WAIT_TIME],
     u"Waiting for a volume to reach target status.",)
 # End: Helper datastructures used by AWS storage driver.
+
+# Begin: Helper datastructures used by OpenStack storage drivers
+
+CODE = Field.for_types("code", [int], u"The HTTP response code.")
+MESSAGE = Field.for_types(
```
exarkun: Looks like this module has become a bit inconsistent about whether it uses `Field.for_types` or `Field.forTypes`.  Eliot is kind of weird in the way it tries to cater to both naming conventions.  We should stick to using just one, though.  And since we use `under_scores` most everywhere, I expect `Field.for_types` is the one to use.  So maybe update the `forTypes` uses near this code to be `for_types` instead.

flocker/node/agents/_logging.py
```diff
@@ -94,3 +99,70 @@
...
+    "message", [bytes, unicode],
+    u"A human-readable error message given by the response.",
```
exarkun: Human readable error messages are unicode because unicode is text and text is what humans read. :smile:

flocker/node/agents/_logging.py
```diff
@@ -94,3 +99,70 @@
...
+)
+DETAILS = Field.for_types("details", [dict], u"Extra details about the error.")
+REQUEST_ID = Field.for_types(
+    "request_id", [bytes, unicode],
+    u"The unique identifier assigned by the server for this request.",
```
exarkun: Does this need to support both types?  There should usually be a specific (good) reason for allowing bytes and unicode to mix - and it should be explained in the code that allows the mixing.  And then all variations of mixing need to be tested.  Python makes it very easy to shoot yourself in the foot in this area by providing implicit conversion between **some** byte strings to unicode and **some** unicode strings to bytes.  This implicit conversion makes it easy to conclude that your string usage is correct only to learn later on that it's totally broken (because someone decided to run your code in a French locale or some such).

flocker/node/agents/_logging.py
```diff
@@ -94,3 +99,70 @@
...
+)
+URL = Field.for_types("url", [bytes, unicode], u"The request URL.")
+METHOD = Field.for_types("method", [bytes, unicode], u"The request method.")
```
exarkun: Ditto

exarkun: Actually... Some of this looks familiar.  We have some HTTP request-related Eliot logging stuff over in `flocker/restapi/_logging.py`.

It would be nice to re-use this stuff instead of re-defining the fields.  I'm not exactly sure how best to do this.  Maybe this merits another module in `flocker.common`?  I'm not sure.  Maybe @itamarst will weigh in.

flocker/node/agents/_logging.py
```diff
@@ -94,3 +99,70 @@
...
+
+COMPUTE_INSTANCE_ID_NOT_FOUND = MessageType(
+    u"blockdevice:cinder:compute_instance_id:not_found",
+    [LOCAL_IPS, API_IPS],
+    u"Unable to determine the instance ID of this node.",
+)
+
+# ActionType used by OpenStack storage driver.
+OPENSTACK_ACTION = ActionType(
+    u"flocker:node:agents:blockdevice:openstack",
```
exarkun: I'm a bit mixed up by the different message and action types defined here...  `openstack:keystone_http_error`, `blockdevice:cinder:compute_instance_id:not_found`, `flocker:node:agents:blockdevice:openstack`... It looks like three competing naming conventions.  Is this intentional?  If so it probably warrants some explanation in the code.

I know that constructing the names for this field of Eliot isn't really something we've established a good convention for.  I wish we had (and as we add more logging I know that we're going to regret not having one ever more acutely).  I think a lot of the code does try to be at least locally consistent, though - so using a single convention throughout this module would probably be reasonable.  Please defend the values you chose if you chose them for a particular reason, though. :smile:

flocker/node/agents/cinder.py
```diff
@@ -9,26 +9,33 @@
...
+from novaclient.exceptions import ClientException as NovaClientException
 
 from twisted.python.filepath import FilePath
 
 from zope.interface import implementer, Interface
 
 from ...common import (
-    auto_openstack_logging, get_all_ips, ipaddress_from_string
+    interface_decorator, get_all_ips, ipaddress_from_string
+    # auto_openstack_logging, get_all_ips, ipaddress_from_string
```
exarkun: I expect this can go?

flocker/node/agents/cinder.py
```diff
@@ -38,23 +45,73 @@
...
+    :param str original_name: The name of the attribute of self where the
+        wrapped object can be found.
+
+    :return: A function which will call the method of the wrapped object and do
+        the extra exception logging.
+    """
+    def _run_with_logging(self, *args, **kwargs):
+        original = getattr(self, original_name)
+        method = getattr(original, method_name)
+        with OPENSTACK_ACTION(operation=[method_name, args, kwargs]):
```
exarkun: Hm.  This seems possibly problematic.  The Python OpenStack APIs probably don't guarantee that all the arguments they take will be simple, JSON-serializable objects.  I'm not sure this logging will be terribly useful if half of the attempts at writing a log message encounter a serialization error.

flocker/common/__init__.py
```diff
@@ -6,18 +6,14 @@
 
 __all__ = [
     'INode', 'FakeNode', 'ProcessNode', 'gather_deferreds',
-    'auto_threaded', 'auto_openstack_logging',
+    'auto_threaded', 'interface_decorator',
     'get_all_ips', 'ipaddress_from_string',
 ]
 
-import platform
+# import platform
```
exarkun: This can go?

flocker/common/interface.py
```diff
@@ -0,0 +1,40 @@
...
+    :param str decorator_name: A human-meaningful name for the class decorator
+        that will be returned by this function.
+    :param zope.interface.InterfaceClass interface: The interface from which to
+        take methods.
+    :param method_decorator: A callable which will decorate a method from the
+        interface.  It will be called with the name of the method as the first
+        argument and any additional positional and keyword arguments passed to
+        ``_interface_decorator``.
+
+    :return: The class decorator.
```
exarkun: An example would probably be helpful as part of this function's documentation.

flocker/node/agents/cinder.py
```diff
@@ -38,23 +45,73 @@
...
+        calls available.
+
+    :return: The class decorator.
+    """
+    return interface_decorator(
+        "auto_openstack_logging",
+        interface,
+        _openstack_logged_method,
+        original,
+    )
```
exarkun: This still has a slightly funny interface.

The application code that uses this looks like:

```python
@auto_openstack_logging(ICinderVolumeManager, "_cinder_volumes")
class _LoggingCinderVolumeManager(PRecord):
    _cinder_volumes = field(mandatory=True)

...

    logging_cinder = _LoggingCinderVolumeManager(
        _cinder_volumes=cinder_client.volumes
    )
```

And the application code that uses the threaded helper based on the same underlying stuff is:

```python
@implementer(IBlockDeviceAsyncAPI)
@auto_threaded(IBlockDeviceAPI, "_reactor", "_sync", "_threadpool")
class _SyncToThreadedAsyncAPIAdapter(PRecord):
    """
    Adapt any ``IBlockDeviceAPI`` to ``IBlockDeviceAsyncAPI`` by running its
    methods in threads of a thread pool.
    """
    _reactor = field()
    _sync = field()
    _threadpool = field()
...
            return _SyncToThreadedAsyncAPIAdapter(
                _sync=self.block_device_api,
                _reactor=reactor,
                _threadpool=reactor.getThreadPool(),
            )
```

I think these examples demonstrate the API isn't quite right.  The thing being decorated is nothing but a place to hang some state.  So perhaps instead it would be better if these things were used on instances, perhaps like:

```python
logging_cinder = auto_openstack_logging(ICinderVolumeManager, cinder_volumes=cinder_client.volumes)
threaded_api = auto_threaded(
    IBlockDeviceAPI, reactor=reactor, api=self.block_device_api, threadpool=reactor.getThreadPool()
)
```

This cuts down on the boilerplate quite a bit and I think it also makes the functions easier to understand.

exarkun:
Thanks!  Great start.  Please address comments and re-submit.

flocker/common/interface.py
myechuri: Fixed.

flocker/common/interface.py
myechuri: Fixed.

flocker/common/interface.py
myechuri: Kept it around as a backup in case i was on the wrong path. Removed.

flocker/common/test/test_interface.py
myechuri: Fixed.

flocker/common/test/test_interface.py
myechuri: Thanks for the reasoning! Fixed.

flocker/common/test/test_interface.py
myechuri: Fixed.

flocker/common/test/test_interface.py
myechuri: Thanks for the informative link!
Fixed as suggested.

flocker/common/test/test_interface.py
myechuri: Thanks for the explanation! Replaced all ``Eliot`` occurrences with Eliot.

flocker/common/test/test_interface.py
myechuri: This is definitely better than what i had! Followed example from test_blockdevice.py, and consolidated the test functions. Thanks for the detailed explanation!

flocker/common/thread.py
myechuri: Fixed.

flocker/common/thread.py
myechuri: Thanks for the detailed explanation! Fixed.

flocker/node/agents/_logging.py
myechuri: Thanks for making me evaluate the logs. The log structures are not yet large in number to deserve a separate module for each storage driver. Trimmed down the documentation.

flocker/node/agents/_logging.py
myechuri: Fixed.

flocker/node/agents/_logging.py
myechuri: Yikes! Sorry! Fixed.

flocker/node/agents/cinder.py
myechuri: Yes, fixed. Left out cleanup in this draft to allow undo in case i was going in the wrong direction.

flocker/common/__init__.py
myechuri: Fixed.

flocker/node/agents/_logging.py
myechuri: i copied this chunk of code from control/_openstack.py.
In fact, @wallrj pointed out during Boto logging branch review that having both bytes and unicode for message type might not be correct.
My interpretation of having two types was ``gettext`` used by Nova APIs to populate message strings. Since the return type is bytes in Python 2.* and unicode in Python 3.0, we would need both types on our end. - This was my reasoning for what i saw. Please correct me if that is not the case? Thanks!

flocker/node/agents/_logging.py
myechuri: First pass through flocker/restapi/_logging.py did not reveal re-usable components.
But, agree with the idea that we could consolidate logging in ``flocker.common``. 
Created https://clusterhq.atlassian.net/browse/FLOC-2053 and assigned to self to track.

flocker/node/agents/_logging.py
myechuri: Agree with having a convention.  There was no particular reason for choosing these names: copied the code blocks as is from _openstack.py.
Changed all names to ``flocker:node:agents:blockdevice:*``.

flocker/node/agents/cinder.py
myechuri: These arguments are ``IBlockDeviceAPI`` arguments, which are in our control.
You are right that some of them might not be JSON-seriazable (like ``BlockDeviceVolume``, as pointed out by @wallrj in EBS logging branch review).
Filed  https://clusterhq.atlassian.net/browse/FLOC-2054 to track this issue.

flocker/common/interface.py
myechuri: Added. Thanks!

myechuri:
Ready for semantic code review, if anyone has time before FLOC-2075 (https://clusterhq.atlassian.net/browse/FLOC-2075) is fixed to allow testing.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,133 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Tests for ``flocker.common.interface_decorator``.
```
wallrj: :question: Nit picking, but this test module will have "Tests  for ``flocker.common._interface``" not just this new function.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,133 @@
...
+        Raise something.
+        """
+
+
+@implementer(IDummy)
+class Dummy(object):
+    """
+    Dummy class that implements ``IDummy`` interface.
+    Implements two methods: one to return an object,
+    and the second to raise an ``Exception``.
```
wallrj: Looks like your right margin might be too narrow. 
We all try and wrap at 79 columns ....and it reminds me that I need to fix this in my own editor.

flocker/common/test/test_interface.py
```diff
@@ -0,0 +1,133 @@
...
+from twisted.trial.unittest import SynchronousTestCase
+
+from eliot.testing import (
+    assertHasMessage, capture_logging
+)
+from eliot import Field, MessageType
+
+from zope.interface import Interface, implementer
+
+from ..common import interface_decorator
```
wallrj: Don't need the `common` here...just `from .. import interface_decorator` i think.

```
[ERROR]
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/trial/runner.py", line 497, in loadPackage
    module = modinfo.load()
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/modules.py", line 383, in load
    return self.pathEntry.pythonPath.moduleLoader(self.name)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/reflect.py", line 303, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/reflect.py", line 250, in _importAndCheckStack
    reraise(excValue, excTraceback)
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/common/test/test_interface.py", line 16, in <module>
    from ..common import interface_decorator
exceptions.ImportError: No module named common

flocker.common.test.test_interface
```

flocker/node/agents/cinder.py
```diff
@@ -38,23 +42,76 @@
...
+                raise
+    return _run_with_logging
+
+
+def auto_openstack_logging(interface, original):
+    """
+    Create a class decorator which will add OpenStack-specific exception
+    logging versions versions of all of the methods on ``interface``.
+    Specifically, some Nova and Cinder client exceptions will have all of their
+    details logged any time they are raised.
```
wallrj: I got confused by this docstring. It's no longer a class decorator, right?

wallrj:
Thanks @myechuri 

I think this looks good.

I like the new instance wrapping version better than the class decorator versions. 
(that's the key change here, right?)

I tried running the tests, but there are some relative import problems and the eliot decorated test methods expect a logger parameter.

So fix up the tests and lets see if it works.

flocker/common/test/test_interface.py
myechuri: Fixed.

flocker/common/test/test_interface.py
myechuri: Thanks! Fixed.

flocker/common/test/test_interface.py
myechuri: Thanks! Fixed.

wallrj:
Not sure why, but when I run the tests locally I get errors because the eliot log testing decorators supply a `logger` parameter to the decorated tests. 

I'm using ``eliot==0.7.1`` as specified in ``setup.py``.

Perhaps you have a newer version?

And I don't understand why the tests aren't also failing on buildbot.

Any ideas? If there's a simple explanation, I'll mark this accepted and we can merge it.

```
(2027)(openstack-eliot-FLOC-2027 ? =)[~/.../HybridLogic/flocker]$ trial flocker.common.test.test_interface
flocker.common.test.test_interface
  InterfaceDecoratorTests
    test_raise ...                                                      [ERROR]
                                                     [ERROR]
    test_return ...                                                     [ERROR]
                                                    [ERROR]

[ERROR]
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 276, in wrapper
    return function(self, *args, **kwargs)
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 308, in wrapper
    return function(self, logger)
exceptions.TypeError: test_raise() takes exactly 1 argument (2 given)

flocker.common.test.test_interface.InterfaceDecoratorTests.test_raise
[ERROR]
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 1368, in _runCleanups
    f(*args, **kwargs)
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 274, in <lambda>
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 341, in assertHasMessage
    testCase.assertTrue(messages, "No messages of type %s" % (messageType,))
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 403, in assertTrue
twisted.trial.unittest.FailTest: No messages of type <eliot._validation.MessageType object at 0x21b21d0>

flocker.common.test.test_interface.InterfaceDecoratorTests.test_raise
[ERROR]
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 276, in wrapper
    return function(self, *args, **kwargs)
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 308, in wrapper
    return function(self, logger)
exceptions.TypeError: test_return() takes exactly 1 argument (2 given)

flocker.common.test.test_interface.InterfaceDecoratorTests.test_return
[ERROR]
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 1368, in _runCleanups
    f(*args, **kwargs)
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 274, in <lambda>
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/eliot/testing.py", line 341, in assertHasMessage
    testCase.assertTrue(messages, "No messages of type %s" % (messageType,))
  File "/home/richard/.virtualenvs/2027/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 403, in assertTrue
twisted.trial.unittest.FailTest: No messages of type <eliot._validation.MessageType object at 0x21b2110>

flocker.common.test.test_interface.InterfaceDecoratorTests.test_return
Ran 2 tests in 0.689s

FAILED (errors=4)

```

wallrj:
Oh, I do see these errors on the http://build.clusterhq.com/builders/flocker%2Finstalled-package%2Fvagrant%2Fcentos-7/builds/229/steps/trial/logs/stdio build 

And I was looking at http://build.clusterhq.com/builders/flocker-centos-7/builds/4410/steps/trial/logs/stdio which didn't have these in the summary because the build was killed.

OK. I'll reject this until we get the new tests passing on buildbot and all the previous PR comments have been addressed or answered.

Sorry for delaying things.

myechuri:
Latest buildbot runs have interface tests passing:
http://build.clusterhq.com/builders/flocker-centos-7/builds/4462/steps/trial/logs/stdio

Functional storage driver tests for Rackspace passed as well:
http://build.clusterhq.com/builders/flocker%2Ffunctional%2Frackspace%2Fcentos-7%2Fstorage-driver/builds/746

All review comments are now addressed. Thanks!

wallrj:
Thanks @myechuri 

I was going to say that test_interface doesn't include coverage of the TypeError for non-methods:

screenshot from 2015-06-10 17 09 51

...but the build results:
 * http://build.clusterhq.com/results/complete-coverage/openstack-eliot-FLOC-2027/build-7025/flocker_common__interface.html#n47

...show that that is covered by something else...I wonder what?

But anyway, I think this is fine. 

Please merge and I'll look forward to having extra logs to help debugging issues like FLOC-2192

myechuri:
> http://build.clusterhq.com/results/complete-coverage/openstack-eliot-FLOC-2027/build-7025/flocker_common__interface.html#n47
...show that that is covered by something else...I wonder what?

``auto_threaded`` also covers this code path. Coverage might be coming from there.
You are right, test_interface should include test for that. Let me add that before merging by end of day.
Thanks for the review!

PR #1474, opened 2015-05-29
[FLOC-1828] Ubuntu archive floc 1828

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1828

PR #1481, opened 2015-06-01
[FLOC-2033] Publish 15.04 packages floc 2033

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2033

Some of this is @exarkun's work from a branch against the parent issue.
The major difference between this and the related work on ``ubuntu-15.04-FLOC-1989`` is that instead of factoring out the creation of repo contents like https://github.com/ClusterHQ/flocker/compare/list-distros-FLOC-2040...ubuntu-15.04-FLOC-1989#diff-324e36d03bad565e5a834100130c7235R1229, many of the tests which this new function helped are now deleted.

These tests are now covered by the tests for https://github.com/ClusterHQ/flocker/pull/1474/files#diff-78dcbb17d48bb3df61e1d451dd15333bR210.

**outdated comment (PR is merged)**:

I think that this can be reviewed but should not be merged until that PR and some other issues are done to support this.


admin/release.py
```diff
@@ -513,6 +517,7 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
         {'distro': 'fedora', 'version': '20', 'arch': 'x86_64'},
         {'distro': 'centos', 'version': '7', 'arch': 'x86_64'},
         {'distro': 'ubuntu', 'version': '14.04', 'arch': 'amd64'},
+        {'distro': 'ubuntu', 'version': '15.04', 'arch': 'amd64'},
```
tomprince: It would be nice if this list weren't hard-coded (like is done elsewhere).

adamtheturtle: The architecture is now taken from an existing mapping in `packaging.py`.

admin/release.py
```diff
@@ -406,7 +406,12 @@ def publish_homebrew_recipe(homebrew_repo_url, version, source_bucket,
 
     homebrew_repo.index.add([recipe])
     homebrew_repo.index.commit('Add recipe for Flocker version ' + version)
+
+    # Sometimes this raises an index error, and it seems to be a race
+    # condition. There should probably be a loop until push succeeds or
+    # whatever condition is necessary for it to succeed is met. FLOC-2043.
```
jongiddy: More useful to display the return value of `push` if this error occurs., particularly as GitPython docs are vague as to actual return value ("If the operation fails completely, the length of the returned IterableList will be null."):
```py
push_info = homebrew_repo.remotes.origin.push(homebrew_repo.head)
assert hasattr(push_info, '__len__') and len(push_info) == 1, push_info
push_info = push_info[0]
```

adamtheturtle: I have linked to the above from the new issue, FLOC-2043 but won't modify it here as it is an unrelated change.

admin/release.py
```diff
@@ -510,23 +514,28 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
+        {'distro': 'fedora', 'version': '20'},
+        {'distro': 'centos', 'version': '7'},
+        {'distro': 'ubuntu', 'version': '14.04'},
+        {'distro': 'ubuntu', 'version': '15.04'},
     ]
 
     for operating_system in operating_systems:
+        for package_type, distribution_names in PACKAGE_TYPE_MAP.items():
+            if operating_system['distro'] in distribution_names:
+                architecture = ARCH['native'][package_type]
```
jongiddy: This change will make it more difficult to support additional architectures later.  I'd suggest adding `arch` back into `operating_systems`.

Instead of naming package-dependent values, add `'arch': 'native'` for each OS. Later, we can do a global replace of `native` into `amd64`. Then we can add `i386` alternatives pretty simply.

This line is then `architecture = ARCH[operating_system['arch']][package_type]`

jongiddy:
Seems OK, but I have not been able to test it to upload packages. Any suggestions on how to create the packages and upload them to, say, `clusterhq-archive-testing`?

adamtheturtle:
@jongiddy `admin/publish-artifacts --help` change the `target` and the `homebrew-tap`. Or modify the code to only run this bit of the artifacts publishing.

admin/test/test_release.py
```diff
@@ -1241,151 +1243,50 @@ def test_development_repositories_created(self):
...
         }
 
-        self.upload_rpms(
+        self.upload_packages(
             aws=aws,
             yum=FakeYum(),
             scratch_directory=self.scratch_directory,
             target_bucket=self.target_bucket,
-            version='0.3.3dev7',
+            version='0.3.3dev1',
```
exarkun: Anything in particular wrong with `7`?

adamtheturtle: No, I didn't actively change it but the diff is a result of me mashing up tests I think.

admin/test/test_release.py
```diff
@@ -1241,151 +1243,50 @@ def test_development_repositories_created(self):
...
+            'ubuntu-testing/14.04/amd64/clusterhq-flocker-node_0.3.3-0.dev.1_all.deb',  # noqa
+            'ubuntu-testing/14.04/amd64/clusterhq-python-flocker_0.3.3-0.dev.1_amd64.deb',  # noqa
             'ubuntu-testing/14.04/amd64/Packages.gz',
             'ubuntu-testing/14.04/amd64/Release',
-        }
-
-        files_on_s3 = aws.s3_buckets[self.target_bucket].keys()
-        self.assertEqual(expected_files, set(files_on_s3))
-
-    def test_development_repositories_created_for_pre_release(self):
```
exarkun: I think this test and the following test are being deleted because they're tedious and repetitive and the functionality that varies is in `make_rpm_version`, not `upload_packages` (the function this case claims to be for).

I think it's probably still worth testing `upload_packages` in combination with the pre- and regular release versions.  If the only thing that's varying is the part of the filename string that contains the tail of the version number, maybe it would be straightforward to factor a test helper out of `test_repositories_created` and then call it from the two other test methods - each time with a different kind of Flocker version number?

At one level, this is repetitive, redundant testing.  It's just a string that's varying, after all.  On the other hand, the different kinds of releases actually involve different logic and the string is encoding a bunch of semantics.  I could imagine `upload_packages` breaking because it tries to extract some of those semantics from the string and getting confused by the different types of versions we have.  So having test coverage for all three is appealing.

admin/release.py
```diff
@@ -559,23 +565,28 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
         target_distro_suffix = ""
 
     operating_systems = [
-        {'distro': 'fedora', 'version': '20', 'arch': 'x86_64'},
-        {'distro': 'centos', 'version': '7', 'arch': 'x86_64'},
-        {'distro': 'ubuntu', 'version': '14.04', 'arch': 'amd64'},
+        {'distro': 'fedora', 'version': '20', 'arch': 'native'},
+        {'distro': 'centos', 'version': '7', 'arch': 'native'},
+        {'distro': 'ubuntu', 'version': '14.04', 'arch': 'native'},
+        {'distro': 'ubuntu', 'version': '15.04', 'arch': 'native'},
```
exarkun: If they're all "native" do we need this last item?

adamtheturtle: It wasn't there and then @jongiddy suggested it in a review comment https://github.com/ClusterHQ/flocker/pull/1481#discussion_r31619597.

admin/release.py
```diff
@@ -559,23 +565,28 @@ def upload_rpms(scratch_directory, target_bucket, version, build_server):
...
     ]
 
     for operating_system in operating_systems:
+        for package_type, distribution_names in PACKAGE_TYPE_MAP.items():
+            if operating_system['distro'] in distribution_names:
+                architecture = ARCH[operating_system['arch']][package_type]
```
exarkun: This loop would probably make a good stand-alone function.  Probably in `admin/packaging.py` (using these magic strings, "native" and "all", as an API between modules is meh).  With error reporting, probably, instead of fall-off-the-end.

admin/test/test_release.py
adamtheturtle: It is not just that it is redundant because it is "just a string" but because the function which creates string, `flocker.common.version.get_package_key_suffix` is tested and the tests here show (enough, in my opinion) that this function is being used. Do you still think that having test coverage for all three is beneficial?

exarkun: Or maybe that's the same mistake as I made when working on this branch earlier.  But I think I was trying to factor out the repository path string and make distro, version, etc parameters to that.  That's a little different from what I'm suggesting here.

admin/release.py
exarkun: Possibly a good idea but also possibly premature generalization.  Fine to leave as-is, I guess.  We'll fix it if we have to fix it (which is the case either way).

admin/test/test_release.py
exarkun: > the tests here show (enough, in my opinion) that this function is being used. Do you still think that having test coverage for all three is beneficial?

My concern is that strings are a crap data structure.  I'm somewhat unconfident that none of the code in the `upload_packages` function cares about the exact format of the version string.  If any of it does care about that format and fails to account for one of the formats (dev, pre, final) then it'll fail while doing some parsing.

exarkun: Ehh.  Actually I may have misunderstood the point of this test.  Let's back up.

admin/test/test_release.py
```diff
@@ -1241,151 +1243,50 @@ def test_development_repositories_created(self):
...
+            'centos-testing/7/x86_64/clusterhq-flocker-cli-0.3.3-0.dev.1.noarch.rpm',  # noqa
+            'centos-testing/7/x86_64/clusterhq-flocker-node-0.3.3-0.dev.1.noarch.rpm',  # noqa
+            'centos-testing/7/x86_64/clusterhq-python-flocker-0.3.3-0.dev.1.x86_64.rpm',  # noqa
+            'centos-testing/7/x86_64/repodata/repomod.xml',  # noqa
+            'centos-testing/7/x86_64/repodata/<newhash>-metadata.xml',  # noqa
+            'ubuntu-testing/14.04/amd64/clusterhq-flocker-cli_0.3.3-0.dev.1_all.deb',  # noqa
+            'ubuntu-testing/14.04/amd64/clusterhq-flocker-node_0.3.3-0.dev.1_all.deb',  # noqa
+            'ubuntu-testing/14.04/amd64/clusterhq-python-flocker_0.3.3-0.dev.1_amd64.deb',  # noqa
             'ubuntu-testing/14.04/amd64/Packages.gz',
             'ubuntu-testing/14.04/amd64/Release',
```
exarkun: From comments elsewhere, it sounds like the point of this test is actually just to assert that every file created is in a "<distro>-testing" directory.  If so, I suggest throwing away most of the data here and just asserting the thing I just wrote.  Something like `assertEqual(set(), {x for x in files if not top_dir(x) not in distro_dirs})`.

exarkun: And then you should probably re-add one more test for the non-testing case otherwise the simpler implementation of `upload_packages` that just crams a "-testing" onto the distro name will also pass the test suite. :smile:

exarkun: Also slightly confusing that `get_package_key_suffix` isn't actually in this branch, just master.

adamtheturtle: Yes, though I did explain in the description of the PR that it was a WIP and merging this branch should be blocked waiting for that to be merged.

> From comments elsewhere, it sounds like the point of this test is actually just to assert that every file created is in a "-testing" directory.

I'll test my theory that it also tests (and is the only test which tests) that `update_repo` gets the right params (OS names, versions).

exarkun:
Thanks.  Modulo minor inline comments, this looks good to me.  I tried testing it though:

```
$ ./admin/publish-artifacts --target s3://jp-testing-clusterhqxxx-archive --flocker-version 0.4.0dev3
...
flocker.provision._effect.SequenceFailed: <SequenceFailed(results=[], exc_info=(<class 'socket.gaierror'>, gaierror(-5, 'No address associated with hostname'), <traceback object at 0x7f7fe7164248>))>
$ ./admin/publish-artifacts --target jp-testing-clusterhqxxx-archive --flocker-version 0.4.0dev3
...
flocker.provision._effect.SequenceFailed: <SequenceFailed(results=[], exc_info=(<class 'requests.exceptions.HTTPError'>, HTTPError('404 Client Error: Not Found',), <traceback object at 0x7f6ea3f659e0>))>
```

Probably user error but it's not clear what I'm doing wrong.

adamtheturtle:
It would be nice to add the failing URL to the exception there. I suspect that it is the missing Fedora packages. I will address these comments soon and may bug you if what I think is a good solution for the repository test isn't what you suggested.

exarkun:
> It would be nice to add the failing URL to the exception there. I suspect that it is the missing Fedora packages. I will address these comments soon and may bug you if what I think is a good solution for the repository test isn't what you suggested.

Ah.  Okay.  I stuck a print in and it is, `http://build.clusterhq.com/results/omnibus/0.4.0dev3/fedora-20/clusterhq-python-flocker-0.4.0-0.dev.3.x86_64.rpm`.  So not an issue with this branch.  Please address other comments and merge (or talk to me about testing some more).  Thanks.

admin/test/test_release.py
adamtheturtle: So this test tests that `update_repo` is called with the correct arguments. It could really just be a spy but I think that this is less fragile. I will re-add a test to prove that '-testing' isn't just hard coded. I think that assertion is too minimal.

adamtheturtle:
This has had enough changes that I think I shouldn't merge without another review:
* The discussed test has mainly stayed the same, but has a better description now. I think that it is a useful test that `update_repo` is given the expected arguments, and is less fragile than a spy. The function just runs  `update_repo` and nothing else so I think that is a good test.
* Less passing around distribution names and version strings. More ``Distribution``s.
* No longer hard code the dictionary of distributions, instead use the existing function to get available distributions from the build_targets directory structure.
* The above required that I add a Dockerfile for Ubuntu 15.04

jongiddy:
I looked over the diff of your recent changes, and they seem fine. I don't think there's a way to test it further apart from actually doing a dev release. (Maybe we need a "not even a dev release" process?

So, I'm happy for it to be merged, so we can test it soon.

adamtheturtle:
> Maybe we need a "not even a dev release" process?

Wait until 1.0 to hear my suggestions :smile:

PR #1482, opened 2015-06-01
[FLOC-2003] Command line introspection was wrong, resulting in unnecessary application restart

itamarst:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2003

### DIFF
--- a/flocker/node/functional/test_deploy.py
+++ b/flocker/node/functional/test_deploy.py
@@ -6,7 +6,7 @@ Functional tests for ``flocker.node._deploy``.
 
 from uuid import uuid4
 
-from pyrsistent import pmap
+from pyrsistent import pmap, pvector
 
 from twisted.trial.unittest import TestCase
 from twisted.python.filepath import FilePath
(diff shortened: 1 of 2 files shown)


flocker/node/functional/test_deploy.py
```diff
@@ -280,13 +277,44 @@ def test_links_lowercase(self):
...
+        """
+        Checking the command-line status results in same command-line we
+        passed in.
+        """
+        command_line = pvector([u"nc", u"-l", u"-p", u"8080"])
+        d = self._start_container_for_introspection(command_line=command_line)
+        d.addCallback(
+            lambda results: self.assertIn(
+                command_line,
+                [app.command_line for app in results[0].applications]))
```
wallrj: :question: There should only be *one* application, right? So maybe I'd unpack the applications into `[application]` and use `assertEqual`.

wallrj:
Thanks @itamarst 

It makes sense. 

 * calculate_changes is saying restart this existing container because the the discovered containers appear to have a different command line than the configured command line, right?

I wondered if it'd be worth adding a calculate changes test to demonstrate that an existing Application  with a command line doesn't get constantly restarted. But it'd be tedious to have to similar tests for all such cases...unless there are already such tests.

wallrj:
Answer or address the comments above and merge when you're satisfied with the build results.

flocker/node/functional/test_deploy.py
itamarst: I think I was worried about previous test runs leaving junk around.

PR #1485, opened 2015-06-02
[FLOC-2056] Less agent.yml validation

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2056

This removes some of the validation for zfs and loopback backends.  This is the easier option compared to adding proper validation for aws, openstack, and third-party backends (at least for now).  In general this was good validation and we should re-introduce it later in a way that's more friendly to extensions.


dwgebler:
Thanks. the build failures are unrelated, good to merge.

PR #1486, opened 2015-06-02
[FLOC-2057] Improved acceptance tests HTTP failure logging

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2057

This gives special treatment to a particular failure Twisted's web client can produce and makes sure to format as much information as we can get from it into the logs.

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -936,21 +936,14 @@ def get_test_cluster(reactor, node_count=0):
         ).write()
 
         def failed_query(failure):
-            try:
-                # Try to detect _WrapperException.
-                reasons = failure.value.reasons
-            except AttributeError:
+            reasons = getattr(failure.value, 'reasons')
+            if reasons is None:
                 # Guess it was something else.  Do some simpler logging.
-                system = "acceptance:is_available_error"
-                Message.new(
-                    message_type=system,
-                    reason=unicode(failure),
-                    exception=unicode(failure.__class__)
-                ).write()
+                write_failure(failure, logger=None)
             else:
                 # It is one of those.  Log all of the stuff from inside it.
                 for reason in reasons:
-                    write_failure(reason, None, system=system)
+                    write_failure(reason, logger=None)
             return False
         d = cluster.current_nodes()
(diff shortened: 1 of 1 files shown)


flocker/acceptance/testtools.py
```diff
@@ -936,9 +936,21 @@ def nodes_available():
         ).write()
 
         def failed_query(failure):
-            Message.new(message_type="acceptance:is_available_error",
-                        reason=unicode(failure),
-                        exception=unicode(failure.__class__)).write()
+            try:
+                # Try to detect _WrapperException.
+                reasons = failure.value.reasons
+            except AttributeError:
```
itamarst: I'd rather have `getattr(failure.value, "reasons", None) is None` or even `isinstance`.

flocker/acceptance/testtools.py
```diff
@@ -936,9 +936,21 @@ def nodes_available():
...
+                # Guess it was something else.  Do some simpler logging.
+                system = "acceptance:is_available_error"
+                Message.new(
```
itamarst: :question: `eliot.writeFailure` will include a traceback, if you think that'd be helpful.

flocker/acceptance/testtools.py
```diff
@@ -936,9 +936,21 @@ def nodes_available():
...
+                    message_type=system,
+                    reason=unicode(failure),
+                    exception=unicode(failure.__class__)
+                ).write()
+            else:
+                # It is one of those.  Log all of the stuff from inside it.
+                for reason in reasons:
+                    write_failure(reason, None, system=system)
```
itamarst: I expect this to blow up with a `NameError`.

itamarst: Latest version of Eliot doesn't require `system` argument anyway for `write_failure`, so you can just omit it.

itamarst:
Address, merge when buildbot is actually green.

flocker/acceptance/testtools.py
wallrj: Done (I think).
Is `logger`  still a required ?

PR #1487, opened 2015-06-02
[FLOC-1788] Docker pull timeout

jongiddy:
Does not do anything with streaming, as mentioned in JIRA issue. This PR just prevents requests/urllib3 from using an infinite timeout, swapping it for a 10 minute timeout.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -290,6 +290,9 @@ class TimeoutClient(Client):
     """
     A subclass of docker.Client that sets any infinite timeouts to the
     provided ``long_timeout`` value.
+
+    This class is hopefully a temporary fix until docker-py with
+    PR/issue TBD merged is released (working on PR, or will submit issue)
     """
 
     def __init__(self, *args, **kw):
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -293,6 +293,8 @@ class TimeoutClient(Client):
 
     This class is a temporary fix until docker-py is released with
     PR #625 or similar. See https://github.com/docker/docker-py/pull/625
+
+    See Flocker JIRA Issue FLOC-2082
     """
 
     def __init__(self, *args, **kw):
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -291,8 +291,8 @@ class TimeoutClient(Client):
     A subclass of docker.Client that sets any infinite timeouts to the
     provided ``long_timeout`` value.
 
-    This class is hopefully a temporary fix until docker-py with
-    PR/issue TBD merged is released (working on PR, or will submit issue)
+    This class is a temporary fix until docker-py is released with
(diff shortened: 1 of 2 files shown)


flocker/node/functional/test_docker.py
```diff
@@ -403,6 +403,68 @@ def test_pull_image_if_necessary(self):
...
+        # Both are subclasses of IOError, so use that for now
+        # https://github.com/kennethreitz/requests/issues/2620
+        self.assertFailure(d, IOError)
+        return d
+
+    def test_pull_timeout_pull(self):
+        """
+        Image pull timeout does not affect subsequent pulls.
+        """
+        # Use an image that isn't likely to be in use by anything, since
```
adamtheturtle: Instead, use an image specifically for this test (similar to https://github.com/ClusterHQ/flocker/pull/1281) - this has only a bit of overhead but is lower risk, and can be a tiny image and doesn't need this comment.

jongiddy: Is it OK to use the same image, assuming that each functional test will never run concurrently?

But this PR hasn't been merged, and the issue looks like it is off the 1.0 track.  How about I add a JIRA issue to update this code when FLOC-1657 gets done?

flocker/node/_docker.py
```diff
@@ -298,10 +323,12 @@ class DockerClient(object):
     :ivar unicode namespace: A namespace prefix to add to container names
         so we don't clobber other applications interacting with Docker.
     """
-    def __init__(self, namespace=BASE_NAMESPACE,
-                 base_url=BASE_DOCKER_API_URL):
+    def __init__(
+            self, namespace=BASE_NAMESPACE, base_url=BASE_DOCKER_API_URL,
+            long_timeout=600):
```
jongiddy: self.comment: add docstring param for the `long_timeout` parameter.

jongiddy: Done.

flocker/node/_docker.py
```diff
@@ -286,6 +286,31 @@ def list(self):
 BASE_DOCKER_API_URL = u'unix://var/run/docker.sock'
 
 
+class TimeoutClient(Client):
+    """
+    A subclass of docker.Client that sets any infinite timeouts to the
+    provided ``long_timeout`` value.
+
+    This class is a temporary fix until docker-py is released with
+    PR #625 or similar. See https://github.com/docker/docker-py/pull/625
```
itamarst: File an issue for upgrading our docker-py dependency when that happens.

flocker/node/functional/test_docker.py
itamarst: Once we upgrade to docker-py when it has this feature we'll just rip these tests out anyway. Seems fine.

itamarst:
Address and merge.

flocker/node/_docker.py
jongiddy: Done.

PR #1490, opened 2015-06-02
Ubuntu cli 15.04 floc 2034

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2034

> This replaces 14.04 CLI instructions with 15.04 CLI instructions, instead of adding 15.04. Note that the instructions are the same but aren't currently guaranteed to be by the code / tests. Do we care about 14.04 in the getting started guide? On the installation page? We could just say "Ubuntu" and it'd fail for the other versions. Or "Ubuntu 14.04 / 15.04" and have a test which confirms that they are the same.

I've just changed this so that the instructions are duplicated for both versions. The text is short enough that this duplication shouldn't matter, and we can revisit later.


admin/client.py
```diff
@@ -297,6 +284,16 @@ def main(reactor, args, base_path, top_level):
...
+    from flocker.common.script import eliot_logging_service
+    log_file = open("%s.log" % base_path.basename(), "a")
+    log_writer = eliot_logging_service(
+        log_file=log_file,
+        reactor=reactor,
+        capture_stdout=False)
+    log_writer.startService()
+    reactor.addSystemEventTrigger(
+        'before', 'shutdown', log_writer.stopService)
+
```
tomprince: I needed these logs to debug the conch failure.

docs/gettingstarted/index.rst
```diff
@@ -84,6 +84,14 @@ Getting started with Flocker
          .. task:: test_homebrew flocker-|latest-installable|
             :prompt: you@laptop:~$
 
+         Ubuntu 15.04
```
tomprince: Adding this makes the tabs run over onto two lines here, but this is changing in #1506 anyway.

adamtheturtle:
This breaks the look of the installation part of the GSG (Safari 8.0.6).

screenshot 2015-06-07 01 47 13

docs/gettingstarted/index.rst
adamtheturtle: My understanding is that as that isn't a blocker to FLOC-2008 it might not be in 1.0. Perhaps @rogthedodge can clarify. If that's the case let's not make this worse.

flocker/provision/_install.py
```diff
@@ -505,7 +520,7 @@ def task_open_control_firewall(distribution):
     """
     if distribution in ('centos-7', 'fedora-20'):
         open_firewall = open_firewalld
-    elif distribution == 'ubuntu-14.04':
+    elif distribution in ('ubuntu-14.04', 'ubuntu-15.04'):
```
adamtheturtle: Some of these are for node instructions so should be reverted.

tomprince: No reason to revert them. It wouldn't surprise me if this makes ubuntu 15.04 just work. (But really, these should probably just check the distro portion, and not the version.)

adamtheturtle: I'm not sure about that. What if Ubuntu 16.04 has a breaking change? And some things will have to change I think because 15.04 uses SystemD instead of Upstart by default.

The reason I think that they should be reverted is because it leaves the code in the state of part-supporting 15.04. Better to do it all in one go. It might be more OK to leave this if we had very short term plans to support 15.04 on the node.

adamtheturtle:
From discussion with @ferrantim - this is worth having a release note for (formerly What's New).

docs/gettingstarted/index.rst
adamtheturtle: This is no longer an issue now that Fedora instructions have been removed

itamarst:
I agree with Adam that we should revert unrelated changes to provisioning code (server-side stuff).

itamarst:
Besides minor comment above, this is OK to merge if the ubuntu-15.04 builder actually passes. So far it has not. I've triggered another build of the latest revision, hopefully it'll pass soon. If it still fails you will need to figure out why :grin:

itamarst:
http://build.clusterhq.com/builders/flocker%2Fclient%2Frackspace%2Fubuntu-15.04

adamtheturtle:
It was failing because there hadn't been a release. Now there has been, I hope it'll pass - it should pass or fail later. This has been manually tested.

PR #1491, opened 2015-06-02
[FLOC-2012] Handle unknown other node state somewhat better in P2PManifestationDeployer

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2012

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -886,7 +886,9 @@ def find_dataset_changes(uuid, current_state, desired_state):
                         for node in desired_state.nodes}
     current_datasets = {node.uuid:
                         set(manifestation.dataset for manifestation
-                            in node.manifestations.values())
+                            # We pretend ignorance is equivalent to no
+                            # datasets; this is wrong. See FLOC-2060.
+                            in (node.manifestations or {}).values())
                         for node in current_state.nodes}
     local_desired_datasets = desired_datasets.get(uuid, set())
     local_desired_dataset_ids = set(dataset.dataset_id for dataset in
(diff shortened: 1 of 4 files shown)


flocker/node/_deploy.py
```diff
@@ -930,15 +932,11 @@ def find_dataset_changes(uuid, current_state, desired_state):
                     going.add(DatasetHandoff(
                         dataset=dataset, hostname=hostname))
 
-    # Look at each dataset that is going to be hosted on this node.  If it
-    # was running somewhere else, we want that dataset to be in `coming`.
-    coming_dataset_ids = local_desired_dataset_ids.intersection(
-        remote_current_dataset_ids)
-    coming = set(dataset for dataset in local_desired_datasets
-                 if dataset.dataset_id in coming_dataset_ids)
-
```
wallrj: :question: This seems unrelated. Is it just house keeping? I guess so, since you didn't have to make any corresponding test modifications.

flocker/node/test/test_deploy.py
```diff
@@ -2196,6 +2196,38 @@ def test_unknown_applications(self):
...
+        The fact that a different node is ignorant about its manifestations
+        does not prevent calculating changes necessary for the current
+        node.
+        """
+        node_state = NodeState(
+            hostname=u"10.1.1.1",
+            uuid=uuid4(),
+            manifestations={MANIFESTATION.dataset_id:
+                            MANIFESTATION},
+            devices={}, paths={},
```
wallrj: I didn't think this should be allowed...
```
# Dataset attributes that must all be non-None if one is non-None:
_DATASET_ATTRIBUTES = {"manifestations", "paths", "devices"}
```

flocker/node/test/test_deploy.py
```diff
@@ -2196,6 +2196,38 @@ def test_unknown_applications(self):
...
+            Node(hostname=api.hostname, uuid=api.node_uuid,
+                 manifestations=node_state.manifestations.transform(
+                     (DATASET_ID, "dataset", "deleted"), True))])
+
+        changes = api.calculate_changes(desired, current)
+        expected = sequentially(changes=[
+            in_parallel(changes=[DeleteDataset(dataset=DATASET.set(
+                "deleted", True))])
+            ])
+        self.assertEqual(expected, changes)
```
wallrj: Any way to use `assert_calculated_changes` here too?

wallrj: Ah, I see it hard codes `BlockDeviceDeployer` so never mind.

wallrj:
Thanks @itamarst 

I left a few comments and questions above but it all makes sense. Please merge when you're satisfied with the build results.

flocker/node/_deploy.py
itamarst: Mostly, yeah. Although originally I thought deleting that code would suffice, until I reached the creation code, and figured I may as well leave that improvement in.

flocker/node/test/test_deploy.py
itamarst: ? none of them are `None`, I don't see what should break.

wallrj: Oh, right, empty dicts. I misread. But the invariant should be stricter about this, right? If there are manifestations there must be devices and paths, otherwise they're non-manifest datasets.

itamarst: Not necessarily. ZFS doesn't have devices. For now this is fine, I think, improving manifests will happen some other time.

PR #1492, opened 2015-06-02
[FLOC-2086] Cache BlockDeviceVolumes in AWS storage driver.

myechuri:


myechuri:
Ready for semantic code review, if anyone has time before FLOC-2075 (https://clusterhq.atlassian.net/browse/FLOC-2075) is fixed to allow testing.

myechuri:
Every testcase in ``test_ebs.py`` noticed savings from cache:
```
2015-06-04 22:20:47+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 22:20:54+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:26:18+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 22:31:44+0000 [-] ELIOT: {'backend_calls_saved': 6,
2015-06-04 22:31:47+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:37:13+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 22:37:48+0000 [-] ELIOT: {'backend_calls_saved': 5,
2015-06-04 22:37:50+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 22:37:52+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 22:37:53+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 22:38:00+0000 [-] ELIOT: {'backend_calls_saved': 3,
2015-06-04 22:38:06+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:38:14+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:38:21+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:38:31+0000 [-] ELIOT: {'backend_calls_saved': 3,
2015-06-04 22:38:38+0000 [-] ELIOT: {'backend_calls_saved': 3,
2015-06-04 22:38:41+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:44:26+0000 [-] ELIOT: {'backend_calls_saved': 8,
2015-06-04 22:46:06+0000 [-] ELIOT: {'backend_calls_saved': 6,
2015-06-04 22:46:48+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:46:49+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:46:57+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:52:23+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 22:57:53+0000 [-] ELIOT: {'backend_calls_saved': 8,
2015-06-04 22:57:59+0000 [-] ELIOT: {'backend_calls_saved': 3,
2015-06-04 22:58:01+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 22:58:01+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 23:03:30+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 23:03:33+0000 [-] ELIOT: {'backend_calls_saved': 2,
2015-06-04 23:03:43+0000 [-] ELIOT: {'backend_calls_saved': 3,
2015-06-04 23:05:43+0000 [-] ELIOT: {'backend_calls_saved': 7,
2015-06-04 23:11:06+0000 [-] ELIOT: {'backend_calls_saved': 5,
2015-06-04 23:11:08+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 23:11:09+0000 [-] ELIOT: {'backend_calls_saved': 1,
2015-06-04 23:11:10+0000 [-] ELIOT: {'backend_calls_saved': 1,

```

myechuri:
Cache test results:
```
flocker.node.agents.test.test_blockdevice
  BlockDeviceVolumeCacheTests
    test_hit_percentage ...                                                [OK]
    test_hit_percentage_empty_cache ...                                    [OK]
    test_insert ...                                                        [OK]
    test_list_keys ...                                                     [OK]
    test_list_volumes ...                                                  [OK]
    test_lookup ...                                                        [OK]
    test_remove ...                                                        [OK]
    test_update ...                                                        [OK]
    test_update_nonexisting_volume ...                                     [OK]

Ran 9 tests in 0.007s

PASSED (successes=9)

```

exarkun:
* Branch seems like a mix of boto reconfiguration and caching
  * Would be nice to have these things in separate PRs
  * At minimum, PR description should mention which issues the PR is intended to resolve.

* Statistics / Logging

  * This is great.  We need lots of stuff like this.  My only comment is that I
    wonder if it makes most sense to compute the statistics in process or just
    log all the events and have a log post-processor do that computing.  Maybe
    @itamarst has an opinion.

* Caching implementation

  * Putting the cache on to the `IBlockDeviceAPI` implementation means that
    each implementation has to duplicate the effort.  That's three
    implementations for ClusterHQ to update.  Putting the cache into the layer
    "above" `IBlockDeviceAPI` (ie, into the code that uses this interface, the
    "user-code" of this interface) would cut this down to just one
    implementation to create and maintain.  It might also address some other
    issues I'll describe below.

  * This needs new test coverage.  There are two kinds of tests needed:

    * Tests that the new code achieves the desired behavior - ie, fewer calls
      are made to the underlying system.  These will be tricky tests to write
      given the current implementation because the current implementation
      varies what EC2 calls get made using boto.  Measuring boto calls will be
      hard because (as far as I know) boto doesn't offer this as a feature and
      because actually making boto calls involves interacting with EC2 (which,
      as we've learned, is highly error prone).  If the caching logic were in
      user-code then it would be varying what `IBlockDeviceAPI` calls got made.
      Since `IBlockDeviceAPI` is our interface, we'll probably have an easier
      time doing testing at this layer.  For example, we have
      `LoopbackBlockDeviceAPI` - explicitly developed as a testing aid.  Tests
      for caching logic in user-code and run against `LoopbackBlockDeviceAPI`
      could either work by:

        * Augment `LoopbackBlockDeviceAPI` to count API calls and then use
          these counts to verify desired caching.
        * Do something which caches some state, interfere with the
          `LoopbackBlockDeviceAPI` directly (for example, delete the storage
          file for one of its volumes) and then verify that user-code can't
          actually see this change because it's respecting the cache instead of
          making another API call.  (However, this implies what I think is a
          misfeature of this caching layer that I'll describe below).

    * Tests for existing behavior (that one might argue should have already
      been written; it's mostly impossible to write *all* possible tests in
      advance - nevertheless we probably wrote too few) that might have been
      broken by the new caching logic. For example, since we use
      `IBlockDeviceAPI` on every Flocker dataset agent node, the following
      sequence is possible:

      * Configuration change so dataset x is manifest on node a
      * node a does create_volume(x), now has [x] cached
      * Configuration change so dataset y is manifest on node b
      * node b does create_volume(y), now has [y] cached
      * Configuration change so that dataset x is manifest on node b (ie, move x from a to b)
      * node b sees configuration change, consults list_volumes(), gets [y] as a result
      * node b does create_volume(x), system explodes

  * Mixing the caching implementation together with the code that actually
    talks to the underlying system (EBS, Cinder, etc) complicates the code for
    talking to the underlying system.  It also complicates the tests for that
    code since it now has to account for caching, possibly do manual cache
    invalidation at certain points, and bugs in the caching system might cause
    failures in the tests for code that's just supposed to talk to the
    underlying system.  Keeping the cache logic separate seems like a pretty
    worthwhile goal.  It will allow the existing tests to continue to be valid
    for the existing `IBlockDeviceAPI` implementations.  It will allow tests to
    be written for the caching logic independent of the details of any
    particular `IBlockDeviceAPI` (or at least based on just one of them,
    probably `LoopbackBlockDeviceAPI` - designed for use in tests).

  * Currently the user-code for `IBlockDeviceAPI` is `BlockDeviceDeployer` (and
    the pile of `IStateChange` implementations it creates).  Caching logic
    could go directly into those classes but that would introduce many
    complications similar to the ones discussed here.  Alternatively, caching
    could go into a new layer between `BlockDeviceDeployer` and
    `IBlockDeviceAPI`.  And ideally, that new layer would actually be just
    another `IBlockDeviceAPI` implementation.  The various caching logic would
    go into methods of that new implementation and whenever that new
    implementation found it didn't have something cached, that it needed to
    consult the underlying system, it would just make a method call onto
    another `IBlockDeviceAPI` implementation with which it was initialized.
    That would go talk to EBS or Cinder or whoever.  `BlockDeviceDeployer` can
    then wrap a "real" `IBlockDeviceAPI` implementation with this caching
    implementation (or, even better, just be handed an object already so
    wrapped so zero BlockDeviceDeployer changes are required - also preserving
    the possibility of creating one of these without the caching layer for the
    purposes of performance comparisons or ruling out the caching layer as a
    source of some bug, etc - eg
    `BlockDeviceDeployer(block_device_api=CacheThing(cinder), ...)`) and
    remain otherwise unchanged.

  * As reflected in the node a/node b scenario above, the cache is perhaps too
    aggressive.  Or maybe the invalidation is not aggressive enough. :smile: I
    think that the only place items are removed from the cache is the
    implementation of `destroy_volume` and the `list_volume` cache is
    effectively never invalidated.  I think this means any changes made by
    another node are invisible to code using this cache.  It will only reflect
    state changes made locally.  This would be okay if we only had one node
    talking to the underlying system (as we eventually plan to do) but it
    probably breaks most of the `calculate_changes` logic we have now.  I'm not
    sure what the best fix for this is but here's a suggestion for a possible
    fix:

      * Try to solve caching for just one operation at a time.  For example,
        `CreateFilesystem` looks up the OS device path for the blockdevice_id
        it's supposed to be creating a filesystem on.  Implementing just the
        part of the caching layer that makes it possible to avoid an API call
        to the underlying system for that case and make a PR for that.
        Caching/invalidation is probably subtle enough that it makes sense to
        have the shortest branches possible to minimize bug count.  Small PRs
        probably also increase the chance of at least getting some improvements
        in to 1.0 even if we don't get them all in.  I wouldn't call the change
        here "too big" but I suspect it's possible to make progress in smaller
        chunks.

      * Consider having both a short-lived, per-iteration cache in addition to
        a longer-lived cache.  For example, if `calculate_changes` decides it
        needs to create one new dataset, the code in master will:

        * discover_state
          * compute_instance_id()
          * list_volumes()
          * get_device_path() for each attached volume
          * get_device_path() for each attached volume
        * CreateBlockDeviceDataset
          * create_volume()
          * compute_instance_id()
          * attach_volume()
          * get_device_path()
          * list_volumes()
          * get_device_path()

        compute_instance_id and get_device_path seem like good candidates for a
        long-lived cache.

        compute_instance_id isn't allowed to change so we can just call it once
        and be done with it forever.  This removes one API call per create (and
        per-other-stuff too).

        We can expect get_device_path to only change after an `attach_device`
        or `detach_device` call so we can cache its result for a particular
        blockdevice_id until one of those operations affects that
        blockdevice_id.  This gets rid of (number of attached volumes) + 2 *
        (number of creates) API calls.

        We can expect that list_volumes might change at any time since another
        node might create or destroy a volume.  Caching it only for one
        iteration of the convergence loop means all code in one loop will
        operate on the same data, though.  It might be out-of-date but that's
        why we keep iterating.  If it is out of date, it'll be up-to-date on
        the next iteration.  This gets rid of one `list_volumes` call per
        create (and in the vase of EBS, this actually gets rid of N + 1 calls
        to EBS since list implies update-per-existing-volume).

        I think the analysis will be similar for other operations.  This won't
        be quite as good as the savings as you've implemented here but that
        might be a necessary decrease in savings to get the desired behavior.

* Model Changes

  * `is_cached_copy` seems to be unused.  Since we're exposing
    `BlockDeviceVolume` to third-parties (possibly a mistake, too late to do
    anything about it now) we probably shouldn't add it.  I can definitely
    imagine it being useful to know whether data came from the cache or not.
    An alternate approach to providing this information might be to have a
    `CachedBlockDeviceVolume` type with all the same fields as
    `BlockDeviceVolume` (perhaps via inheritance).  I think pyrsistent makes it
    trivial to turn one of these into the other - seems like
    `CachedBlockDeviceVolume.create(block_device_volume)` does it (given my
    extremely minimal investigation).  This exposes cached state in the type
    (easily inspected) and avoids making changes to `BlockDeviceVolume`.  Along
    with moving the caching to user-code this should help avoid needing to make
    any/many changes to the `IBlockDeviceAPI` implementations to support this
    caching.

  * Richard and I went back and forth on whether to have the `attached_device`
    attribute or not.  I can't remember why we decided not to at this point.  I
    see that it's probably the only (or at least the only convenient) way to
    communicate this information from the EBS implementation to the cache
    without forcing an extra, redundant EBS API call.  Maybe that's sufficient
    reason to keep it.  If device path caching can work without the new
    attribute, though, I think that'd be better (since it has less impact on
    each `IBlockDeviceAPI` implementer).

PR #1493, opened 2015-06-02
[FLOC-2071] Always include NonManifestDatasets in discovery

itamarst:


flocker/node/agents/blockdevice.py
```diff
@@ -1673,10 +1673,9 @@ def is_existing_block_device(dataset_id, path):
                 applications=None,
                 used_ports=None,
             ),
+            NonManifestDatasets(datasets=nonmanifest),
         )
 
-        if nonmanifest:
-            state += (NonManifestDatasets(datasets=nonmanifest),)
```
exarkun: Hm.  I think this breaks real-world use.  Now ... why do I think that.  Well, I wrote this originally and at the time I guess I thought the `if nonmanifest:` guard was important... But that's not really a why...

:thinks about it on the way to the office:

exarkun: Okay.  Can't think of a good reason it would be important to not report these if the value is empty.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -337,18 +338,15 @@ def assert_discovered_state(case,
...
+    # FLOC-1806 - Make this actually be a dictionary (callers pass a list
+    # instead) and construct the ``NonManifestDatasets`` with the
+    # ``Dataset`` instances that are present as values.
+    expected += (
+        NonManifestDatasets(datasets={
+            unicode(dataset_id):
+            Dataset(dataset_id=unicode(dataset_id))
+            for dataset_id in expected_nonmanifest_datasets
+        }),)
     case.assertEqual(expected, state)
```
exarkun: I guess this helper would have been less fragile if it had been defined in terms of a `DeploymentState` instead of a sequence of changes.  Not actionable, just noting.

exarkun:
Thanks.  Looks good.  Please merge.

PR #1494, opened 2015-06-03
[FLOC-2051] Datasets should be empty by default

itamarst:
Passing criteria - the postgres tests should pass on the loopback builder: http://build.clusterhq.com/builders?branch=empty-datasets-FLOC-2051


itamarst:
Seems to have broken loopback; I'm investigating.

itamarst:
Depends on #1496 now.

PR #1495, opened 2015-06-03
[FLOC-2041] Force mke2fs so that it formats non-partitioned devices

itamarst:
@vaibhavkhanduja could you test if it fixes your problems, and add reply here? Thanks!


vaibhavkhanduja:
My branch is quite behind master as I am waiting for plugin work to finish ... I made manual changes to my code adding -F as additional parameter which works well ....I also realized to rename my application so as to get brand new volume from array ...(with the same name it gives same volume, somehow do not mkfs issue with it, I still though have find this in code)

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2800,7 +2802,8 @@ class MountBlockDeviceTests(
     )
 ):
     """
-    Tests for ``MountBlockDevice``\ 's ``IStateChange`` implementation.
+    Tests for ``MountBlockDevice``\ 's ``IStateChange`` implementation, as
+    well as ``CreateFilesystem`` testing.
```
exarkun: :frowning:

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2570,12 +2570,14 @@ def multistep_change_log(parent, children):
...
-        ]
-        self.assertEqual(children_actions, parent_action.children)
+        children_actions = {
+            child_action.action_type:
+            LoggedAction.of_type(logger.messages, child_action)
+            for child_action in children
+        }
+        for child in parent_action.children:
+            action_type = child.startMessage["action_type"]
+            self.assertIn(child, children_actions.get(action_type, []))
```
exarkun: Why this?  I'm kind of interested in the order of log events.  e.g. if a mount happens before an attach ... that's messed up.

exarkun:
Thanks.  Apart from the testing change relating to log messages, this looks good to me.  It sounds like it probably addresses @vaibhavkhanduja's problem too.

Buildbot seems uncharacteristically pleased as well.

Can you avoid losing the ordering assertion in the logging helper?  Actually, what I'd really like is a clear explanation of why that change is correct.  Looking at how the logged messages have changed, what I see is that for the resize tests, the final mount operation seems to have switched over to an entirely different parent of the action hierarchy.  I don't understand how that happened considering nothing in this branch appears to touch the resize implementation.

Please submit with explanation for that test change.  Or just merge if you fix whatever problem introduced the need for it and therefore can revert the test change.

flocker/node/agents/test/test_blockdevice.py
itamarst: That's how I found it.

flocker/node/agents/test/test_blockdevice.py
itamarst: It does still verify order. The issue is just that because the CreateVolume now logs sub-actions you get multiple mounts logged in the test. So you assert requested children happened in correct order but don't blow if child occurred earlier in test, when it's logs for pre-test-assertion setup.

itamarst:
Ah, guess I'm not asserting order. I'll fix that.

flocker/node/agents/blockdevice.py
```diff
@@ -869,12 +878,12 @@ def run(self, deployer):
         )
         device = api.get_device_path(volume.blockdevice_id)
 
-        # This duplicates CreateFilesystem now.
-        check_output(["mkfs", "-t", "ext4", device.path])
+        create = CreateFilesystem(volume=volume, filesystem=u"ext4")
```
itamarst: This and next change are why logging changed, I believe: **new** logging was added.

exarkun: Without the logging assertion change, the only failures I see are in `flocker.node.agents.test.test_blockdevice.ResizeBlockDeviceDatasetTests` which it seems shouldn't be affected by `CreateBlockDeviceDataset` changes.

itamarst: `_run_resize_test` creates a `CreateBlockDeviceDataset` instance.

exarkun: Details of the failure:

```
>>> print len(a)
6
>>> print len(b)                                                                                                                                                                               
6
>>> print a[:5] == b[:5]
True
>>> print a[5] == b[5]                                                                                                                                                                         
False
>>> pprint((lambda m: (m.startMessage, m.endMessage, m.children))(a[5]))
({u'action_status': u'started',
  u'action_type': u'agent:blockdevice:mount',
  'dataset_id': UUID('c5de3757-44a7-48b2-8a07-3c7e07d32159'),
  u'task_level': [3, 1],
  u'task_uuid': u'34c6a4ba-13f5-40ab-b334-aab68fa2a950',
  u'timestamp': 1433439727.046077},
 {u'action_status': u'succeeded',
  u'action_type': u'agent:blockdevice:mount',
  u'task_level': [3, 3],
  u'task_uuid': u'34c6a4ba-13f5-40ab-b334-aab68fa2a950',
  u'timestamp': 1433439727.061461},
 [LoggedMessage(message={u'timestamp': 1433439727.052725, 'volume': BlockDeviceVolume(blockdevice_id=u'block-c5de3757-44a7-48b2-8a07-3c7e07d32159', dataset_id=UUID('c5de3757-44a7-48b2-8a07-3c7e07d32159'), attached_to=u'flocker_node_agents_test_test_blockdevice_ResizeBlockDeviceDatasetTests_test_run_shrink-909280', size=16777216), u'task_uuid': u'34c6a4ba-13f5-40ab-b334-aab68fa2a950', 'block_device_path': FilePath('/dev/loop0'), u'message_type': u'agent:blockdevice:mount:details', u'task_level': [3, 2]})])
>>> pprint((lambda m: (m.startMessage, m.endMessage, m.children))(b[5]))
({u'action_status': u'started',
  u'action_type': u'agent:blockdevice:mount',
  'dataset_id': UUID('c5de3757-44a7-48b2-8a07-3c7e07d32159'),
  u'task_level': [7, 1],
  u'task_uuid': u'ae0f9110-31f1-4e9c-a5ac-2127426a0110',
  u'timestamp': 1433439727.239218},
 {u'action_status': u'succeeded',
  u'action_type': u'agent:blockdevice:mount',
  u'task_level': [7, 3],
  u'task_uuid': u'ae0f9110-31f1-4e9c-a5ac-2127426a0110',
  u'timestamp': 1433439727.254057},
 [LoggedMessage(message={u'timestamp': 1433439727.248761, 'volume': BlockDeviceVolume(blockdevice_id=u'block-c5de3757-44a7-48b2-8a07-3c7e07d32159', dataset_id=UUID('c5de3757-44a7-48b2-8a07-3c7e07d32159'), attached_to=u'flocker_node_agents_test_test_blockdevice_ResizeBlockDeviceDatasetTests_test_run_shrink-909280', size=8388608), u'task_uuid': u'ae0f9110-31f1-4e9c-a5ac-2127426a0110', 'block_device_path': FilePath('/dev/loop0'), u'message_type': u'agent:blockdevice:mount:details', u'task_level': [7, 2]})])
>>> 
```

exarkun: It seems like the same operation being logged but it has a different `task_level` and a different `task_uuid`.

exarkun: Ahhhg.  And a `Dataset` with a different `maximum_size`.

itamarst: Yeah, that's... FLOC-2084? There's a PR up for that to rip out resize logic.

itamarst: I still think the logging assertion change is fine: all it does is not blow up if some other code happens to log a related action. Which it does, legitimately, in this case.

flocker/node/agents/blockdevice.py
```diff
@@ -869,12 +878,12 @@ def run(self, deployer):
...
+        d = run_state_change(create, deployer)
 
         mount = MountBlockDevice(dataset_id=UUID(hex=self.dataset.dataset_id),
                                  mountpoint=self.mountpoint)
-        d = mount.run(deployer)
+        d.addCallback(lambda _: run_state_change(mount, deployer))
```
exarkun: Ah.  Because this mount operation wasn't being logged before but now it is.  And the logging assertion is getting confused and finding this unrelated mount event in the logs and treating it like it's the one it is looking for.

I guess that means `multistep_change_log` is broken.  It's finding children of actions other than the ones it is supposed to be finding.

exarkun: A quick fix might be to move the monkey-patching of the logger object to after the `CreateBlockDeviceDataset` is done so the extra mount isn't around to confuse things.  Please also file a bug report about `multistep_change_log`, though.

exarkun: Or just revert the `multistep_change_log` change and don't merge this until after FLOC-2084 is merged.

flocker/node/agents/blockdevice.py
exarkun: I don't think the change is valid.  The action discovery code is simply buggy.  It was passing by accident before because there weren't other actions being logged to confuse it.  Now there are other actions to confuse it.  It's failing because it finds an unrelated mount operation in the log.  If the resize code somehow stopped logging its mount operation, the new version of the assertion would still pass and that would be incorrect.

So I think it would be better to revert the assertion change until we want to spend time correctly fixing that helper.

exarkun:
Thanks.  Trivial conflict in `test_blockdevice.py` now, sadly.  Resolve that, revert logging change, and merge after #1510 (so the now-broken resize tests will have been deleted).

PR #1496, opened 2015-06-03
[FLOC-2089] Don't hardcode ZFS backend in libcloud runner

itamarst:


itamarst:
This was fixed differently in FLOC-1925, so just going to close this for now.

PR #1497, opened 2015-06-03
[FLOC-1925] Supply dataset backend configuration to acceptance testing nodes

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1925

Makes it possible to configure dataset backends.
Adds EBS and Cinder backends to flocker-dataset-agent as well.
Plus numerous other unfortunate but necessary changes that don't belong here.

PR #1498, opened 2015-06-03
[FLOC-2051] Datasets should be empty by default

itamarst:
http://build.clusterhq.com/builders?branch=empty-datasets-FLOC-2051 loopback builder needs to be green.

Builds on #1496.


flocker/node/agents/blockdevice.py
```diff
@@ -498,7 +498,13 @@ def run(self, deployer):
         #     e2fsck refuses to run non-interactively, though.
         #
         # See FLOC-1814
-        check_output([b"e2fsck", b"-f", b"-y", device.path])
+
+        # The first call will exit with non-zero exit code as it needs to
+        # recreate lost+found. The second call ought to succeed.
+        command_line = [b"e2fsck", b"-f", b"-y", device.path]
+        call(command_line)
```
wallrj: This will write to stdout / stderr. Does that matter? Are we interested in the output? Will the output corrupt the logs sent to journald?

wallrj:
Thanks @itamarst 

This looks really nice and I learned about `mklost+found`.

And I *really* like the follow up issue that allows us to store metadata in the volume root.

Please merge when you're happy with the build results and after reviewing and merging #1497 (https://clusterhq.atlassian.net/browse/FLOC-1925) to avoid causing further pain to @exarkun .

PR #1499, opened 2015-06-03
fix issues found during release process

Azulinho:


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -85,6 +83,13 @@ Preparing For a Release
 
    This uses SSH agent forwarding so that you can push changes to GitHub using the keys from your workstation.
 
+   Add your SSH key to the ``sshd`` agent.
+
+   .. prompt:: bash $
+
+      [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent)
+      ssh-add $HOME/.ssh/id_rsa
```
adamtheturtle: Do this only if the file exists, and mention explicitly that it should be changed to a key which is linked to your GitHub account.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -118,7 +123,8 @@ Preparing For a Release
 
    - The NEWS date format is YYYY-MM-DD.
    - The NEWS file should also be updated for each pre-release and Weekly Development Release, however there should be only one NEWS entry for each Major Marketing Release and Minor Marketing Release.
-   - This means that in doing a release, you may have to change the NEWS heading from a previous Weekly Development Release or pre-release.
+     This means that in doing a release, you may have to remove the previous development release or pre-release header, merging the changes from that previous release into the current release.
+     This applies to both Weekly Development Releases or pre-releases.
```
adamtheturtle: "This applies to both Weekly Development Releases or pre-releases." is redundant.

adamtheturtle:
Great, thanks. Please address, and merge when the docs builder is green.

docs/gettinginvolved/infrastructure/release-process.rst
Azulinho: ssh -A won't work if ssh-agent is not running and the key added to the agent. So it's not really optional here.

the github comment is spot on

PR #1500, opened 2015-06-04
Always keep acceptance cluster if --keep specified.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2070


dwgebler:
We don't have tests for this script :( I had to make a similar temporary change locally when I was running the acceptance runner repeatedly for another branch. Build failures seem unrelated, please merge.

adamtheturtle:
Usually I prefer `if x` instead of `if not x`. Doesn't matter though, merging as per @dwgebler 's review.

PR #1501, opened 2015-06-04
[FLOC-2099] Missing parameter in acceptance.py

dwgebler:
Fixes FLOC-2099


adamtheturtle:
@dwgebler Thanks, merge if the relevant builders pass. Else resubmit if you need to make changes.

dwgebler:
The TLS branch that separately fixes this problem by not requiring Vagrant's runner to use `configured_cluster_for_nodes` got merged first, so this is no longer necessary.

PR #1502, opened 2015-06-04
Add patched prompt directive that works on Safari and Chrome

jongiddy:
Upstream patch submitted as https://github.com/sbrunner/sphinx-prompt/pull/3

This temporarily uses the code from that patch in the Flocker code.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -34,10 +34,8 @@ extensions = [
     'sphinx.ext.ifconfig',
     'flocker.provision._sphinx',
     'flocker.docs.version_extensions',
-    # Replace sphinx-prompt with patched version.  Patched version from
-    # https://github.com/sbrunner/sphinx-prompt/pull/3
-    # See FLOC-2102
-    # 'sphinx-prompt',
+    # Replace sphinx-prompt with patched version. See FLOC-2102. Patched
+    # version from https://github.com/sbrunner/sphinx-prompt/pull/3
     'flocker.docs.prompt_patched',
     'sphinxcontrib.httpdomain',
     'flocker.restapi.docs.publicapi',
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -34,7 +34,11 @@
     'sphinx.ext.ifconfig',
     'flocker.provision._sphinx',
     'flocker.docs.version_extensions',
-    'sphinx-prompt',
+    # Replace sphinx-prompt with patched version.  Patched version from
+    # https://github.com/sbrunner/sphinx-prompt/pull/3
+    # See FLOC-2102
+    # 'sphinx-prompt',
```
adamtheturtle: Get rid of commented out code, the issue is enough.

flocker/docs/prompt_patched.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/python
+# coding: utf-8
+
+
+from docutils import nodes
+from docutils.parsers import rst
+
+from pygments import highlight
```
adamtheturtle: I think that if we import something directly it should have a version-pinned dependency in setup.py

flocker/docs/prompt_patched.py
```diff
@@ -0,0 +1,164 @@
+#!/usr/bin/python
+# coding: utf-8
+
```
adamtheturtle: Docstring with comment and link to issue like in conf.py

adamtheturtle:
Thanks @jongiddy.

Tested by viewing http://build.clusterhq.com/results/docs/sphinx-prompt-FLOC-1548/build-7547/using/installing/index.html on OS X 10.10.3 in Safari 8.0.6, Firefox 38.01 and Chrome 43. Unfortunately it showed up some bad prompts! I'll deal with that separately.

Hopefully the patch will be accepted soon. Address the minor comments and merge when you and the docs builder are happy.

adamtheturtle:
@jongiddy As this has been merged into master upstream maybe we should just wait for a release? Depends on the timeframe.

jongiddy:
Sounds like a few days at least. It's so easy to undo, I'll merge once tests have passed.

PR #1504, opened 2015-06-04
Swap tabs and noscript content, to make prompt styles work

jongiddy:
A quick patch to make Getting Started Guide work with modified `sphinx-prompt`. Swaps the order of the `noscript` section with the normal tabbed section.

In `master`, the prompts are missing from the OS X and Ubuntu tabs since #1502  was merged. In this branch, the prompts should be back.

### DIFF
--- a/docs/gettingstarted/index.rst
+++ b/docs/gettingstarted/index.rst
@@ -74,7 +74,7 @@ Getting started with Flocker
 
          Local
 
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks, add a comment in this file and where ``noscript-content`` is defined about the bad intersection between that and sphinx prompt. Then merge.

PR #1507, opened 2015-06-04
[FLOC-2103] Catch duplicate names in the deployment configuration

itamarst:


flocker/control/_config.py
```diff
@@ -1268,6 +1269,11 @@ def deployment_from_configuration(deployment_state, deployment_configuration,
             )
         node_applications = []
         for name in application_names:
+            if name in seen_applications:
+                raise ConfigurationError(
+                    "Application '{name}' appears more than once.".format(
+                        name=name))
+            seen_applications.add(name)
```
exarkun: I wonder if this would make more sense as a `Deployment` invariant.  If we had implemented the duplication application check in the `POST /configuration/containers` endpoint in terms of the `Deployment` invariant then this bug probably never would have arising.

This looks like it solved the problem, though, so probably best to file a follow-up issue to unify these two implementations of the same check as an invariant and go with this for now.

exarkun:
Thanks.  File follow-up and merge when buildbot is green.

PR #1510, opened 2015-06-04
[FLOC-2084] Rip out unused resize logic, since it keeps introducing bugs/maintenance burden

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2084


exarkun:
Thanks.

Other resize logic I see still in the branch:

  * All of the code in the ZFS backend.  Since this is totally different code paths that probably do actually work, leaving it makes sense.
  * The eliot field and action types that only relate to resize operations, 
  * The last `IStateChange` implementation in support of resize, `ResizeVolume`.
  * The `IBlockDeviceAPI` and `IBlockDeviceAsyncAPI` interfaces' `resize_volume` method.
  * The implementations of that method on `LoopbackBlockDeviceAPI`, `CinderBlockDeviceAPI`, and `EBSBlockDeviceAPI`.
  * Comments about ResizeFilesystemTests at the top of `test_blockdevice.py`
  * Tests for the above mentioned things - including backend-specific functional tests that are currently marked as skipped.

I could almost imagine leaving `IBlockDeviceAPI.resize_volume` and everything "below" that - but not quite.  I don't think it makes sense to suggest that people should implement this considering we don't know if we'll actually be able to use it for anything or if we need an entirely different approach for resize.

Thanks.  Resubmit with those additional deletions.

wallrj:
The thing I don't understand is that the DatasetAPI move tests work for me.

```
flocker.acceptance.test_api
  DatasetAPITests
    test_dataset_move ...                                                  [OK]

Ran 3 tests in 137.511s

PASSED (successes=3)
--keep specified, not destroying nodes.
To run acceptance tests against these nodes, set the following environment variables: 
export FLOCKER_ACCEPTANCE_VOLUME_BACKEND=openstack;
export FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH=/tmp/tmphj9qtw;
export FLOCKER_ACCEPTANCE_AGENT_NODES=10.208.32.32:10.208.32.49;
export FLOCKER_ACCEPTANCE_CONTROL_NODE=10.208.32.32;

```

exarkun:
What is surprising, in the context of removing support for resize, about the acceptance tests for move working?

exarkun:
Ah, I suppose what's surprising is that this was referenced as a fix for a bug encountered which a user characterized as move-related.  But our tests don't exercise the particular situation the user encountered - so the test suite passes.

wallrj:
Thanks @itamarst 

Sorry this took so long to review.

I assumed that I'd be able to recreate the test failure without realising the test failed out side of normal test conditions.
Then spent ages getting the acceptance tests to run.

So fine. I can't find any remaining resize references outside of flocker.volume and I see that we're leaving that in place for the time being.

The test failures look unrelated and I'm going to manually run the acceptance tests on rackspace with the openstack dataset backend (I don't think we have a working builder for that).

But please merge anyway.

PR #1511, opened 2015-06-05
[FLOC-2096] Support specifying hostname explictly.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2096

### DIFF
--- a/docs/using/installing/index.rst
+++ b/docs/using/installing/index.rst
@@ -498,6 +498,9 @@ The optional ``port`` variable is the port on the control node to connect to.
 This value must agree with the configuration for the control service telling it on what port to listen.
 Omit the ``port`` from both configurations and the services will automatically agree.
 
+If node has a public address that is different than its local address (for example, on an EC2 instance),
+then the file should include a ``hostname`` item with a public address of the node.
+
 The file must also include a ``dataset`` item.
 This selects and configures a dataset backend.
 All nodes must be configured to use the same dataset backend.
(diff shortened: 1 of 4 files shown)


flocker/provision/_install.py
```diff
@@ -992,6 +995,7 @@ def configure_cluster(cluster, dataset_backend_configuration):
                             dataset_backend_configuration=(
                                 dataset_backend_configuration
                             ),
+                            public_ip=node.address if node._node.driver.name == 'Amazon EC2' else None,
```
tomprince: This should probably be changed. Perhaps it should just be passed unconditionally? (My first inclination was to not do that, so we test the case where it isn't specified).

flocker/node/test/test_script.py
```diff
@@ -558,6 +618,34 @@ def deployer_factory(node_uuid, hostname, cluster_uuid):
...
+        def deployer_factory(node_uuid, hostname, cluster_uuid):
+            spied.append(hostname)
+            return object()
+
+        reactor = MemoryCoreReactor()
+        options = DatasetAgentOptions()
+
+        config = yaml.safe_load(self.config.getContent())
+        config["hostname"] = "hostname.local"
+        self.config.setContent(yaml.safe_dump(config))
```
tomprince: This is a bit ugly, but I think the entire `flocker.node.script` needs to be refactored so there is less duplication. And this then probably won't need to be reading a file.

docs/using/installing/index.rst
```diff
@@ -498,6 +498,9 @@ The optional ``port`` variable is the port on the control node to connect to.
 This value must agree with the configuration for the control service telling it on what port to listen.
 Omit the ``port`` from both configurations and the services will automatically agree.
 
+If node has a public address that is different than its local address (for example, on an EC2 instance),
+then the file should include a ``hostname`` item with a public address of the node.
```
itamarst: That is wrong. We explicitly want our linking layer to run over the internal IP, because of:

1. Security (default VPC stuff in modern EC2 accounts).
2. Bandwidth costs (internal network traffic is free).

PR #1512, opened 2015-06-05
[FLOC-2083] Release 0.4.1dev3

adamtheturtle:
It was agreed to ignore failing tests which are not release-specific. That means that the following test failures should be ignored in review:

* Failed tests on ``flocker/acceptance/vagrant/centos-7/zfs`` because of a key error
* Known `flocker.provision.test.test_ssh_conch` tests

If you'd like when reviewing the link redirects here, it might be a good opportunity to review the automation of that: https://clusterhq.atlassian.net/browse/FLOC-1701.

Redirects have been manually tested and do work, but the authors page has moved so the instructions and tool need to be updated.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-Flocker 0.4.1dev1 (2015-05-25)
+Flocker 0.4.1dev3 (2015-06-05)
 ==============================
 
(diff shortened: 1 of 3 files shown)


jongiddy:
LGTM. Please proceed to the Release.

adamtheturtle:
Closing as the release failed.

PR #1514, opened 2015-06-05
[FLOC-2093] Allow REST API to be split into sections.

tomprince:


admin/acceptance.py
```diff
@@ -154,7 +154,6 @@ def stop_cluster(reactor):
 
     # dict giving configuration for the dataset backend the nodes use - eg
     # {"pool": "flocker"}
-    'dataset_backend_configuration',
```
itamarst: Diffs in this module seem unrelated?

itamarst: (So I'm not reviewing them.)

docs/reference/api.rst
```diff
@@ -26,14 +26,38 @@ For more information, see the :ref:`API authentication guide <authenticate>`.
...
+---------
+
+.. autoklein:: flocker.control.httpapi.ConfigurationAPIUserV1
+    :schema_store_fqpn: flocker.control.httpapi.SCHEMAS
+    :prefix: /v1
+    :examples_path: api_examples.yml
+    :section: container
+
+
+.. XXX: Improvements to the API (split in sections, and create collapse directive), both require Engineering effort:
```
itamarst: Seems like splitting was already done?

flocker/control/httpapi.py
```diff
@@ -790,7 +802,16 @@ def list_current_nodes(self):
                 self.cluster_state_service.as_deployment().nodes]
 
     @app.route("/configuration/_compose", methods=['POST'])
-    @private_api
+    @user_documentation(
+        u"""
+        Private API endpoint used by flocker-deploy.
+
+        Please do not use it as it may be removed in the near future.
```
itamarst: I think I'd still rather use the `@private_api` decorator, but if you stick to this at least update docstring appropriately.

itamarst:
1. A bunch of unrelated changes; they should be removed (or merged forward or whatever it is).
2. Can't tell if tests pass or not; make sure they do.

Otherwise looks good; merge when issues above are addressed.

PR #1516, opened 2015-06-05
[FLOC-2086] Draft for adding NodeSnapshot in BlockDeviceDeployer.

myechuri:


itamarst:
This design impacts all deployers, including those that are very different in needs and structure (application, peer-to-peer storage), so I'm not sure it's worth proceeding with as is. I'm going to try to implement some of Jean-Paul's suggestions from #1492 and see how that goes.

PR #1517, opened 2015-06-06
[FLOC-2119] Temporary workaround for requests timeout

itamarst:
This was two bugs: some sort of race condition involving ThreadPool, and what I believe is a bug fixed in newer versions of `requests` (or rather `urllib3` - https://github.com/shazow/urllib3/pull/520). I was never able to reproduce this bug locally, so I don't think it's worth wasting more time fighting it since it's not a bug in our code AFAICT.

Even if we do want to fix this before 1.0 I still think it's worth skipping for now so it doesn't bottleneck all our other branches with red builders.


flocker/ca/functional/test_validation.py
```diff
@@ -385,3 +389,6 @@ def test_requests_dns(self):
         the REST API when it is accessed via a domain name.
         """
         return self.assert_requests_validates(b"localhost", "control_dns")
+    test_requests_dns.skip = (
```
adamtheturtle: :question: Can you merge the WIP on that issue and run it to see if this  proposed solution is valid?

adamtheturtle:
The failing builders aren't those which are affected by the issue and I don't think that they could be caused by this change.

flocker/ca/functional/test_validation.py
adamtheturtle: I don't see why this is better than just removing the test, and creating an issue to create it again.

flocker/ca/functional/test_validation.py
```diff
@@ -385,3 +389,6 @@ def test_requests_dns(self):
...
+        "This fails in some environments, probably due to using old version of"
+        " requests. Re-enable in FLOC-2035 when we upgrade requests.")
```
adamtheturtle: If we are going to stick with this, put a comment on that issue.

adamtheturtle:
Thanks @itamarst . I think that if we have (manual) testing of the API endpoint examples then we can go without this test for now. Please remove the test and make sure that there is an issue which explicitly will add it back (maybe blocked by the requests-py issue) and merge, or explain why you think it is better and to have the test and resubmit.

adamtheturtle:
And check that you also don't think that the failing builders are caused by this.

itamarst:
I left the test being skipped since there were two problems, and I wanted to include a fix for one of them. I can just file an issue pointing to that commit revision though.

adamtheturtle:
I think that is better than having code which doesn't run in the repo (it is kind of like commented out code).

PR #1519, opened 2015-06-06
[FLOC-2130] Process lifetime cache for IBlockDeviceAPI

itamarst:
With goal of reducing unnecessary calls to storage APIs, in particular on EC2 where we're hit rate limits.


flocker/node/agents/blockdevice.py
```diff
@@ -1997,3 +1998,46 @@ def _calculate_resizes(self, configured_manifestations, local_state):
...
+    :ivar _device_paths: Mapping from blockdevice ids to cached device path.
+    """
+    def __init__(self, api):
+        self._api = api
+        self._instance_id = None
+        self._device_paths = {}
+
+    def compute_instance_id(self):
+        """
+        Always return initial result, since this shouldn't change until a
```
exarkun: extraneous comma

exarkun:
The cache looks good.  The invalidation looks good.

The functional tests don't use this cache, though, so they're not going to benefit at all from this (they also don't all pass if I change them to use it; and ignoring that, they don't really seem to benefit much from it anyway, I suppose because they're mostly just making `IBlockDeviceAPI` calls, not invoking `BlockDeviceDeployer` which is where most of the repetition is, and constantly switching to new Python instances and new volumes).

The acceptance tests should benefit from this ... six hours later ... well it's too hard to actually run the acceptance tests (at least if you're only willing to try for a single day).  So putting this aside for now.

exarkun:
Okay ... Some more manual hacking and I managed to get the acceptance tests to run.  Observation demonstrates this improves `flocker.acceptance.test_api`.  Still waiting for a complete run of the acceptance test suite to complete for comparison.

Here's the limited results:

```
ubuntu@ip-172-31-28-36:~/flocker$ grep Params with.log | wc -l
744
ubuntu@ip-172-31-28-36:~/flocker$ grep Params without.log | wc -l
839
```

itamarst:
I feel like the IBlockDeviceAPI design has some issues that cause too many API calls within it (see https://clusterhq.atlassian.net/browse/FLOC-2132). Since we're not going to implement changes to that pre-1.0 we might want to consider some caching within the EBS backend if we're still seeing rate issues, rather than the per-iteration listing cache.

exarkun:
Full run complete.  Real, significant improvements there - on the same order of magnitude as in the shorter run:

```
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Param with.log  | wc -l
1104
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Param without.log  | wc -l
1223
```

exarkun:
Tons of acceptance test failures on buildbot.  Probably mostly to do with EC2/private IP address problems though.  The API tests all pass.  Resolve merge conflict and merge.  Thanks.

itamarst:
Ah, I guess you just didn't hit Accept button in JIRA.

PR #1520, opened 2015-06-06
[FLOC-2133] Remove unnecessary EBS calls

itamarst:


flocker/node/agents/ebs.py
```diff
@@ -371,16 +370,22 @@ def _get_ebs_volume(self, blockdevice_id):
 
         :param unicode blockdevice_id: ID of a blockdevice that needs lookup.
 
-        :returns boto.ec2.volume.Volume for the input id. ``None`` if
-            no boto.ec2.volume.Volume was found for the given id.
+        :returns boto.ec2.volume.Volume for the input id.
```
exarkun: This markup seems invalid.

flocker/node/agents/ebs.py
```diff
@@ -371,16 +370,22 @@ def _get_ebs_volume(self, blockdevice_id):
...
         """
-        for volume in self.connection.get_all_volumes(
-                volume_ids=[blockdevice_id]):
-            if volume.id == blockdevice_id:
-                # Sync volume for uptodate metadata
-                volume.update()
-                return volume
-        return None
+        try:
+            for volume in self.connection.get_all_volumes(
```
exarkun: Pull the `get_all_volumes` call out so the exception handling is around it, alone?  Or can `volume.id` raise this exception?  Or `__iter__` or `__next__`?

flocker/node/agents/ebs.py
```diff
@@ -371,16 +370,22 @@ def _get_ebs_volume(self, blockdevice_id):
...
+                    volume_ids=[blockdevice_id]):
+                if volume.id == blockdevice_id:
+                    return volume
+        except EC2ResponseError as e:
+            if e.error_code == "InvalidName.NotFound":
```
exarkun: Ugh, classy.

Maybe worth a link to https://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html#CommonErrors since it took me a few minutes to find.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1620,6 +1620,16 @@ def test_created_volume_attributes(self):
...
 
+    def unknown_blockdevice_id(self):
+        """
+        Return a legitimate unknown block device id.
+
+        Override in subclasses as needed.
+
+        :return: By default a newly generated UUID.
+        """
+        return unicode(uuid4())
```
exarkun: Maybe better not to have a default implementation at all.  It's easy to overlook and might grant the appearance of correctness if malformed identifiers trigger different behavior from well-formed identifiers not associated with a volume.  `raise NotImplementedError(...)` would probably be good.

flocker/node/agents/ebs.py
```diff
@@ -174,7 +174,6 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
 
     :return: Input volume in BlockDeviceVolume format.
     """
-    ebs_volume.update()
```
exarkun: This makes sense to me but I suspect it will break some other code.

`create_volume` does `requested_volume = self.connection.create_volume(...)` and later `return _blockdevicevolume_from_ebs_volume(requested_volume)`.  `_blockdevicevolume_from_ebs_volume` uses `ebs_volume.tags[DATASET_ID_LABEL]`.  `create_volume` only sets that tag after the volume is created.  So I'd expect `requested_volume` to be missing the information.  Perhaps boto is extra-mutate-y and makes the tag available anyway somehow, I'm not sure.    BuildBot hasn't caught up yet so I'm not sure what the test suite has to say about this.

exarkun: If this is an issue, it seems that `_wait_for_volume` could probably return the most up-to-date ebs volume object and code that previously used a kind-of out-of-date object could use that recently-updated one instead.

exarkun:
Thanks.  Basically looks good to me, with caveats:

  * It'd be nice to see how much this helps.  Would it be reasonable to only merge this after the branch that adds boto logging to let us see requests being made and then compare those logs with and without this change to see how much things improve in the test suite?  If so, please do that and leave a note here with the results.  If not, is there another way to measure how much better this makes things?
  * It'd love to see some passing (or at least completing) builds on BuildBot. :/  The `requests` / acceptance test thing seems to be getting in the way of some builders.  Maybe only of builders irrelevant to this change, though?  If you agree that's the case, okay.
  * The implementation does seem to be lacking at least some small thing, though, because some relevant tests do fail on buildbot (unless I am looking at an old build or something) - eg http://build.clusterhq.com/builders/flocker%2Ffunctional%2Faws%2Fcentos-7%2Fstorage-driver/builds/487/steps/trial/logs/problems has flocker.node.agents.functional.test_ebs.EBSBlockDeviceAPIInterfaceTests.test_detach_unknown_volume
 failing with "<Response><Errors><Error><Code>InvalidVolume.NotFound</Code><Message>The volume 'vol-00000000' does not exist.</Message></Error></Errors><RequestID>a296bc28-4547-4dc9-9a85-153e0ede68d3</RequestID></Response>".  That looks like a case that the code is trying to handle so maybe it is an old build...

exarkun:
Please merge after addressing those points.

flocker/node/agents/ebs.py
itamarst: `_wait_for_volume` does an `update()`, which is mutate-y, so should be fine.

exarkun: Argh hooray.

exarkun:
I fixed the boto logging branch (~3 hours) and did some tests with this branch and that branch merged together (~3 minutes) and produced these results:

For a single (arbitrarily selected) test method:

```
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Params with-fix/test.log  | wc -l
28
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Params without-fix/test.log  | wc -l
41
```

For the full functional test suite:


```
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Params with-fix/test.log  | wc -l                                                                 
569
(flocker.env)ubuntu@ip-172-31-28-36:~/flocker$ grep Params without-fix/test.log  | wc -l                                                              782
```

Number of "Params" lines is basically a proxy for number of AWS API calls.  This is for a single test method.

Perhaps not very interesting is that the functional test suite without the fix took 392s and with the fix took 381s (a difference which is probably less than the noise).  These tests conducted in a separate region free of interference from BuildBot so I expect this just means that a single test suite run doesn't come close to the AWS rate limits (or if it does, only just barely) therefore reducing the number of API calls doesn't materially affect runtime.

Given these numbers (and with the rest of the review points addressed) I'd be happy to have this branch merged now.  Thanks again.

exarkun:
Tons of red on BuildBot.  Some of the failures have to do with handling of unknown volume ids.  I don't think they're introduced by this PR but it's hard to tell.  And given the rest of the problems, it's basically impossible to get a coherent set of build results for master and this branch in order to do a comparison right now.

Given more time, I'd do a bunch of manual runs of the relevant parts of the test suite on master and on this branch to compare the results.  Lacking more time (as we do), I'm just merging it.  Comment above demonstrates how it does reduce AWS API calls.  With this and a few more similar changes perhaps we'll start getting builds with meaningful failures instead of meaningless ones.

PR #1518, opened 2015-06-06
[FLOC-2113] Support lack of storage backend configuration for Vagrant runner

itamarst:
Possibly a hacky solution and we should've just changed config somewhere, but seemed expedient.

PR #1522, opened 2015-06-07
[FLOC-2113] flocker/acceptance/vagrant/centos-7/zfs fails with a KeyError

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2113

This makes the storage drivers section of the acceptance testing configuration file optional as it was intended to be.  Since the Vagrant runner doesn't accept any configuration, this fixes the issue (which seems to be that the slave is deployed with an old version of the configuration file).

Obsoletes #1518

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -539,7 +539,9 @@ class RunOptions(Options):
         Get the configuration corresponding to storage driver chosen by the
         command line options.
         """
-        return self['config']['storage-drivers'][self['dataset-backend']]
+        drivers = self['config'].get('storage-drivers', {})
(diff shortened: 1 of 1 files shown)


tomprince:
Merge when buildbot is happy.

tomprince:
Note the slave has a custom acceptance test configuration [here](https://github.com/ClusterHQ/build.clusterhq.com/blob/master/slave/vagrant/fabfile.py#L35-L37).

exarkun:
Succinct.

exarkun:
BuildBot is so, so red.  But in a way that indicates this change has accomplished the intended goal.  Merging.

PR #1523, opened 2015-06-08
[FLOC-1789] Change docker image name

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1789

### DIFF
--- a/docs/gettingstarted/fig.yml
+++ b/docs/gettingstarted/fig.yml
@@ -5,7 +5,7 @@ web:
   ports:
    - "80:80"
 redis:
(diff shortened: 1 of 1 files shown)


dwgebler:
Thanks @adamtheturtle this is correct as far as this image name goes, it is now just `redis` - can you confirm that you've tested this new app and deployment config as per the GSG? And that this is the only place in our docs we refer to `dockerfile/` image names?

dwgebler:
I have tested this and followed the GSG successfully. Good to merge.

PR #1524, opened 2015-06-08
[FLOC-1903] Refactor acceptance variables handling.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-1903

This is preparation for [FLOC-2093](https://clusterhq.atlassian.net/browse/FLOC-2093).

The original control flow was incoherent.
- `flocker_deply`, `get_clean_nodes` and `get_test_cluster` all parsed the environment variables describing the cluster.
- `flocker_deploy` parsed the variables to get the control node and certificates. I changed this to a method of `Cluster, and changed the callers to use `require_cluster`. Incidentally, this allowed me to remove a call to `get_clean_nodes` from each of them.
- `get_clean_nodes` parsed the variables, checked that `ssh` was available (which is no longer accessed in the tests),  called `get_test_cluster` to cluster to clean the nodes, and then returned a list of addresses. It was used in some of the flocker-deploy tests. I changed it to be a method of `Cluster` that just cleans the nodes. The one remaining caller already was getting a cluster and discarded the return, and this change made it do that only once.
- `assert_expected_deployment` called `get_test_cluster` to get a cluster, but all the callers already had access to a cluster, so I changed this to be a method.


flocker/acceptance/testtools.py
```diff
@@ -121,94 +118,6 @@ def create_attached_volume(dataset_id, mountpoint, maximum_size=None,
     )
 
 
-class SSHCommandFailed(Exception):
```
tomprince: This is dead code.

flocker/acceptance/testtools.py
```diff
@@ -121,94 +118,6 @@ def create_attached_volume(dataset_id, mountpoint, maximum_size=None,
...
-    #   $ flocker-deploy volume-deployment.yml volume-application.yml
-    #
-    # http://doc-dev.clusterhq.com/advanced/cleanup.html#removing-zfs-volumes
-    # A tool or flocker-deploy option to purge the state of a node does
-    # not yet exist. See https://clusterhq.atlassian.net/browse/FLOC-682
-    try:
-        run_SSH(22, 'root', node, [b"zfs"] + [b"destroy"] + [b"-r"] +
-                [b"flocker"], None)
-    except SSHCommandFailed:
-        pass
```
exarkun: I am quite happy to see these go.  Thanks.

flocker/acceptance/testtools.py
```diff
@@ -890,8 +591,130 @@ def current_nodes(self):
...
+
+        def configuration_matches_state():
+            d = self.current_containers()
+            d.addCallback(got_results)
+            return d
+
+        return loop_until(configuration_matches_state)
+
+
+def _get_test_cluster(reactor, node_count=0):
```
exarkun: Only one caller of this now and it passes, `node_count`, so the default could probably go.

exarkun:
Thanks.  This looks a ton better than it was.  There's also a ton of acceptance test failures, but I think all because the new `flocker_deploy` method accidentally uses `self.certificate_path` instead of `self.certificates_path`.  Please fix that and merge if it produces a build without acceptance test regressions.

PR #1528, opened 2015-06-08
Update top navigation in docs inline with new site

msh100:
Fixes FLOC-2145


itamarst:
The links don't quite match up with ones on preview site, but that's just a preview for now so maybe doesn't match final links? Fix that if necessary, otherwise looks fine. Merge when ready.

msh100:
Thanks Itamar, the links in the document provided are correct. If they don't align with the new site, then that wont pass website review

PR #1531, opened 2015-06-08
[FLOC-2153] Add sleep before checking for result, to reduce polling

itamarst:
I went with 5 seconds, in hope that sometimes it'll be faster and 10 seconds is a long time.


myechuri:
Looks good, thanks!

exarkun:
It would have been nice to see a comparison of AWS API call count before and after this change.

PR #1532, opened 2015-06-08
[FLOC-2143] Remove host attribute from /state/containers

itamarst:
Probably not *strictly* necessary, but I feel like it'll encourage people to interact badly with the API.

### DIFF
--- a/flocker/control/test/test_httpapi.py
+++ b/flocker/control/test/test_httpapi.py
@@ -3048,7 +3048,6 @@ class ContainerStateTestsMixin(APITestsMixin):
         ])
         expected_dict = dict(
             name=u"myapp",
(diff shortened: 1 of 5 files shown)


flocker/control/test/test_httpapi.py
```diff
@@ -3048,7 +3048,6 @@ def test_one_container(self):
         ])
         expected_dict = dict(
             name=u"myapp",
-            host=expected_hostname,
```
exarkun: Yay for minuses and stuff.  Booo for so much duplication in our test suite. :cry:

exarkun:
Thanks.  Is good (except, you know, all the standard red).  Merging.

PR #1533, opened 2015-06-08
[FLOC-2149] Update to the AWS install section of the Installation doc

davidoliverSP2:
Robs feedback comments:
https://docs.google.com/document/d/12LZHegqXSCLaEB2OPd4Dcg5rHRuU1rK1Xz53qyXwdb8/edit#heading=h.obzsgevhppff
and
https://docs.google.com/document/d/1aTIrDN6YGq6I7Qc1MwFKxnooO9yz7qZ_mFtNo8bOFfg/edit#heading=h.5crjnr15nsco
(there is some overlap on these drive docs)

### DIFF
--- a/docs/using/installing/index.rst
+++ b/docs/using/installing/index.rst
@@ -177,9 +177,8 @@ Using Amazon Web Services
    * Tag instance.
    * Configure security group.
       
-     * If you wish to customize the instance's security settings, make sure to permit SSH access both from the intended client machine (for example, your laptop) and from any other instances on which you plan to install ``clusterhq-flocker-node``.
-     * When you add a custom TCP rule, you will need to provide a port range of 4523-4524.
-     * This enables Flocker agents to communicate with the control service and for external access to the API.
+     * If you wish to customize the instance's security settings, make sure to permit SSH access from the administrators machine (for example, your laptop).
+     * To enable Flocker agents to communicate with the control service and for external access to the API, add a custom TCP security rule enabling access to ports 4523-4524.
      * Keep in mind that (quite reasonably) the default security settings firewall off all ports other than SSH.
      * For example, if you run the tutorial you won't be able to access MongoDB over the Internet, nor will other nodes in the cluster.
      * You can choose to expose these ports but keep in mind the consequences of exposing unsecured services to the Internet.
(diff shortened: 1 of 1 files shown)


docs/using/installing/index.rst
```diff
@@ -168,19 +168,26 @@ Using Amazon Web Services
...
+     * When you add a custom TCP rule, you will need to provide a port range of 4523-4524.
+     * This enables Flocker agents to communicate with the control service and for external access to the API.
+     * Keep in mind that (quite reasonably) the default security settings firewall off all ports other than SSH.
+     * For example, if you run the tutorial you won't be able to access MongoDB over the Internet, nor will other nodes in the cluster.
+     * You can choose to expose these ports but keep in mind the consequences of exposing unsecured services to the Internet.
+     * Links between nodes will also use public ports but you can configure the AWS VPC to allow network connections between nodes and disallow them from the Internet.
+
+   * Review to ensure your instances have sufficient storage and your security groups have the required ports.
+
+   Launch when you are happy to proceed.
```
itamarst: "ready" instead of "happy"?

docs/using/installing/index.rst
```diff
@@ -168,19 +168,26 @@ Using Amazon Web Services
...
+#. Configure the instance.
+   Complete the configuration wizard; in general the default configuration should suffice.   
+
+   * Choose instance type. We recommend at least the ``m3.large`` instance size.
+   * Configure instance details. You will need to configure a minimum of 2 instances.
+   * Add storage. It is important to note that the default storage of an AWS image can be too small to store popular Docker images, so we recommend choosing at least 16GB to avoid potential disk space problems.
+   * Tag instance.
+   * Configure security group.
+      
+     * If you wish to customize the instance's security settings, make sure to permit SSH access both from the intended client machine (for example, your laptop) and from any other instances on which you plan to install ``clusterhq-flocker-node``.
```
itamarst: This is... inaccurate and ambiguous.

1. There are two things one might mean by "client" - if you mean "flocker-deploy" then SSH access is *not* needed. If you mean "machine admin is using to configure the new instance" then probably saying something like that is more accurate than "client machine".
2. Other instances only need SSH access if they're using ZFS backend... and ZFS on AWS isn't really a supported configuration. 
3. "When you add a custom TCP rule" doesn't quite make sense. "You should add a custom TCP security rule enabling access to ports 4523-4524" perhaps? This is not a nice-to-have, it's a necessary step.

itamarst:
Address comments above, merge when buildbot doc builder is happy.

PR #1534, opened 2015-06-08
[FLOC-2161] Longer wipe grace period

itamarst:


itamarst:
I expect some unit tests to fail... I think they hard code assumption of ten seconds.

itamarst:
If so, fix them so they use the constant rather than hard-coded numbers, then merge when green.

PR #1535, opened 2015-06-08
[FLOC-2150] Wait for delete to complete before returning from destroy_volume.

myechuri:
flocker.node.agents.functional.test_ebs
  EBSBlockDeviceAPIInterfaceTests

Ran 1 tests in 6.590s

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -228,16 +228,11 @@ def _wait_for_volume(volume,
         try:
             volume.update()
         except EC2ResponseError as e:
-            # If the volume is being deleted, upon successful deletion,
-            # we will fail to see the volume, as indicated by
-            # ``EC2ResponseError`` code ``InvalidVolume.NotFound``.
+            # If AWS cannot find the volume, raise ``UnknownVolume``.
             # (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
             # for error details).
-            if (transient_status == u'deleting' and
-                    e.code == u'InvalidVolume.NotFound'):
-                return
-            else:
-                raise
+            if e.code == u'InvalidVolume.NotFound':
+                raise UnknownVolume(volume.id)
         if volume.status == end_status:
             return
         elif volume.status not in [start_status, transient_status]:
@@ -597,11 +592,13 @@ class EBSBlockDeviceAPI(object):
         ebs_volume = self._get_ebs_volume(blockdevice_id)
         destroy_result = self.connection.delete_volume(blockdevice_id)
         if destroy_result:
-            _wait_for_volume(ebs_volume,
-                             start_status=u'available',
-                             transient_status=u'deleting',
-                             end_status='')
-            return
+            try:
+                _wait_for_volume(ebs_volume,
+                                 start_status=u'available',
+                                 transient_status=u'deleting',
+                                 end_status='')
+            except UnknownVolume:
+                return
         else:
             raise Exception(
                 'Failed to delete volume: {!r}'.format(blockdevice_id)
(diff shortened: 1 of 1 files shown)


flocker/node/agents/ebs.py
```diff
@@ -590,6 +602,11 @@ def destroy_volume(self, blockdevice_id):
                         'Failed to delete volume: {!r}'.format(blockdevice_id)
                     )
                 else:
+                    ebs_volume = self._get_ebs_volume(blockdevice_id)
+                    _wait_for_volume(ebs_volume,
+                                     start_status=u'available',
+                                     transient_status=u'deleting',
+                                     end_status='')
```
exarkun: Not directly related to this branch but `if ret_val is False:` is a bad expression.  That should be `if not ret_val:`.  And since there's an `if` and an `else` block you could make it `if ret_val:` and flip the order of the blocks for easier reading.  And `ret_val` is a pretty generic variable name.  Perhaps rename it something like `destroy_result`.

exarkun: I wonder why `destroy_volume` uses `self.list_volumes()` instead of using `self._get_ebs_volume` right from the start.

flocker/node/agents/ebs.py
```diff
@@ -590,6 +602,11 @@ def destroy_volume(self, blockdevice_id):
                         'Failed to delete volume: {!r}'.format(blockdevice_id)
                     )
                 else:
+                    ebs_volume = self._get_ebs_volume(blockdevice_id)
```
exarkun: Since we just deleted the volume, it seems plausible (though significantly unlikely) that this could fail with `UnknownVolume`.  It just depends whether EC2 manages to destroy the volume faster than we manage to make the `connection.get_all_volumes` call.

I have no realistic suggestions for how to test this case but it seems like it could be avoided by making the `_get_ebs_volume` call before the `connection.delete_volume` call.

myechuri: Thanks for pointing out! Calling ``_get_ebs_volume`` before ``destroy_volume`` is definitely simpler. Fixed.

flocker/node/agents/ebs.py
myechuri: > I wonder why destroy_volume uses self.list_volumes() instead of using self._get_ebs_volume right from the start.

No good reason: i wrote destroy_volume before turning _get_ebs_volume into macro. Thanks for pointing out optimization!

flocker/node/agents/ebs.py
```diff
@@ -225,7 +225,19 @@ def _wait_for_volume(volume,
...
-        volume.update()
+        try:
+            volume.update()
+        except EC2ResponseError as e:
+            # If the volume is being deleted, upon successful deletion,
+            # we will fail to see the volume, as indicated by
+            # ``EC2ResponseError`` code ``InvalidVolume.NotFound``.
+            # (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
+            # for error details).
+            if (transient_status == u'deleting' and
```
itamarst: Having this special case deleting seems... unnecessary? How about just raising `UnknownVolume` if you ever get that code, and then you can just do `try: _wait_for(...) except UnknownVolume: return` in `destroy_volume()`.

itamarst:
It would be nice to have a test that directly demonstrates the expected behavior change (though green builder on thing that has been red forever does demonstrate that indirectly). However, once https://clusterhq.atlassian.net/browse/FLOC-2165 is done I expect the "deleted volumes aren't listed" test will actually cover this change, so I think that's OK.

So: address minor design improvement above, then merge if functional builder is green.

flocker/node/agents/ebs.py
myechuri: Cleaner option! Fixed.

myechuri:
> It would be nice to have a test that directly demonstrates the expected behavior change 

``test_attach_destroyed_volume`` consistently failed locally without the fix.
With the fix, it passed.

PR #1536, opened 2015-06-08
[FLOC-2136] Reliable agent IP detection

itamarst:
[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1536)

### DIFF
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -91,6 +91,8 @@ def _get_external_ip(host, port):
     Get an external IP address for this node that can in theory connect to
     the given host and port.
 
+    Failures are retried until a successful connect.
+
     See https://clusterhq.atlassian.net/browse/FLOC-1751 for a possibly
     better solution.
 
@@ -99,15 +101,20 @@ def _get_external_ip(host, port):
 
     :return unicode: IP address of external interface on this node.
     """
-    with LOG_GET_EXTERNAL_IP(host=unicode(host), port=port) as ctx:
-        sock = socket()
+    while True:
         try:
-            sock.connect((host, port))
-            result = unicode(sock.getsockname()[0], "ascii")
-            ctx.addSuccessFields(local_ip=result)
-            return result
-        finally:
-            sock.close()
+            with LOG_GET_EXTERNAL_IP(host=unicode(host), port=port) as ctx:
+                sock = socket()
+                try:
+                    sock.connect((host, port))
+                    result = unicode(sock.getsockname()[0], "ascii")
+                    ctx.addSuccessFields(local_ip=result)
+                    return result
+                finally:
+                    sock.close()
+        except:
+            # Error is logged by LOG_GET_EXTERNAL_IP.
+            pass
 
 
 class _TLSContext(PRecord):
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -6,6 +6,8 @@ The command-line ``flocker-*-agent`` tools.
 """
 
 from socket import socket
+from contextlib import closing
+from time import sleep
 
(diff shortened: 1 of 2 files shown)


flocker/node/test/test_script.py
```diff
@@ -915,3 +920,46 @@ class ContainerAgentOptionsTests(
...
+                     assertHasAction(test, logger, LOG_GET_EXTERNAL_IP, True,
+                                     {u"host": u"localhost",
+                                      u"port": test.destination_port},
+                                     {u"local_ip": u"127.0.0.1"}))
+    def test_successful_get_external_ip(self, logger):
+        """
+        A successful external IP lookup returns the local interface's IP.
+        """
+        self.assertEqual(u"127.0.0.1",
+                         _get_external_ip(u"localhost", self.destination_port))
```
exarkun: Makes some assumptions about the system's network stack/configuration. :/  Really hard to avoid doing that without just faking the `socket` module, though ... :/

Still, I'm not looking forward to when we have to start ignoring failures from this test due to stupid network behaviors/misbehaviors.  So maybe faking the socket module is actually the better course?  Hard to decide which way is less bad.  I think fake socket might be since the fake can eventually be verified and then all the gross real socket/system interaction goes in one well-defined place instead of spread across the rest of the project.

flocker/node/test/test_script.py
```diff
@@ -915,3 +920,46 @@ class ContainerAgentOptionsTests(
...
+        A failed external IP lookup is retried (and the error logged).
+        """
+        original_connect = socket.socket.connect
+
+        def fail_once(*args, **kwargs):
+            socket.socket.connect = original_connect
+            raise RuntimeError()
+        self.patch(socket.socket, "connect", fail_once)
+        self.assertEqual(u"127.0.0.1",
+                         _get_external_ip(u"localhost", self.destination_port))
```
exarkun: Pfft.  Maybe someday I'll actually get into the habit of reading to the bottom of the page before making comments.

exarkun: So if faking is good enough here then I think it's definitely good enough for the previous test.

flocker/node/script.py
```diff
@@ -78,24 +80,41 @@ def deployer_factory(cluster_uuid, **kwargs):
...
+        try:
+            with LOG_GET_EXTERNAL_IP(host=unicode(host), port=port) as ctx:
+                sock = socket()
+                try:
+                    sock.connect((host, port))
+                    result = unicode(sock.getsockname()[0], "ascii")
+                    ctx.addSuccessFields(local_ip=result)
+                    return result
+                finally:
+                    sock.close()
```
exarkun: There's also `with closing(socket()) as sock:` if you want to give this code that new-car smell.

flocker/node/script.py
```diff
@@ -78,24 +80,41 @@ def deployer_factory(cluster_uuid, **kwargs):
...
+    while True:
+        try:
+            with LOG_GET_EXTERNAL_IP(host=unicode(host), port=port) as ctx:
+                sock = socket()
+                try:
+                    sock.connect((host, port))
+                    result = unicode(sock.getsockname()[0], "ascii")
+                    ctx.addSuccessFields(local_ip=result)
+                    return result
+                finally:
```
exarkun: Probably not super nice to busy-loop here.  Add a sleep. :disappointed:

exarkun:
Thanks.  Address those points and then, if the two new tests are passing on buildbot, merge.

itamarst:
I think I broke the unit test suite (it's trying to connect forevers), I'll fix that tomorrow.

PR #1537, opened 2015-06-08
[FLOC-2160] Report non-manifest datasets in /state/datasets

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2160

### DIFF
--- a/docs/reference/api_examples.yml
+++ b/docs/reference/api_examples.yml
@@ -255,7 +255,7 @@
 
   response: |
     HTTP/1.1 200 OK
-
+    # FLOC-2160 Add a non-manifest dataset here.
     [{"dataset_id": "47440eff-e933-4de0-b56c-d3469b61421f",
       "primary": "%(NODE_0)s",
       "maximum_size": 1073741824,
--- a/docs/reference/api_examples.yml
+++ b/docs/reference/api_examples.yml
@@ -250,8 +250,8 @@
   doc: |
     Get a list of all datasets in a deployment.
     This includes manifest (attached) and non-manifest (unattached) datasets.
-    Some examples of non-manifest datasets are datasets that are being created or
-    destroyed and datasets that are moving between nodes.
+    For example, a dataset on a block device that has been created, but is not
+    attached to any machine, is a non-manifest dataset.
     The ``primary`` and ``path`` values will not be present for non-manifest
     datasets.
 
--- a/docs/reference/api_examples.yml
+++ b/docs/reference/api_examples.yml
@@ -250,8 +250,10 @@
   doc: |
     Get a list of all datasets in a deployment.
     This includes manifest (attached) and non-manifest (unattached) datasets.
-    A dataset is non-manifest when it is moving between nodes.
-    The ``primary`` and ``path`` values of non-manifest datasets will be ``null``.
+    Some examples of non-manifest datasets are datasets that are being created or
+    destroyed and datasets that are moving between nodes.
+    The ``primary`` and ``path`` values will not be present for non-manifest
+    datasets.
 
   request: |
     GET /v1/state/datasets HTTP/1.1
--- a/docs/reference/api_examples.yml
+++ b/docs/reference/api_examples.yml
@@ -264,9 +264,7 @@
       "maximum_size": 1073741824,
       "path": "/flocker/somearbitrarypath"},
      {"dataset_id": "886ed03a-5606-453a-94a9-a1cbaf35164c",
-      "primary": None,
-      "maximum_size": 1073741824,
-      "path": None}]
+      "maximum_size": 1073741824}]
 
 -
   id:
--- a/docs/reference/api_examples.yml
+++ b/docs/reference/api_examples.yml
@@ -249,17 +249,24 @@
 
   doc: |
     Get a list of all datasets in a deployment.
+    This includes manifest (attached) and non-manifest (unattached) datasets.
+    A dataset is non-manifest when it is moving between nodes.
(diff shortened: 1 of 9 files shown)


myechuri:
We might need a test_blockdevice.py test for to cover all listed states in ``list_volumes``? 
Thanks!

flocker/control/httpapi.py
```diff
@@ -938,10 +942,16 @@ def datasets_from_deployment(deployment):
...
     :return: Iterable returning all datasets.
     """
+    # FLOC-2160 This needs to report all datasets including non-manifest
+    # datasets.  And happily there's a new ``DeploymentState.all_datasets()``
+    # method which does this for us...
     for node in deployment.nodes:
         if node.manifestations is None:
             continue
         for manifestation in node.manifestations.values():
+            # FLOC-2160 Whether it's important to only report "primary"
```
itamarst: Yeah, non-primary is whole other can of worms.

docs/reference/api_examples.yml
```diff
@@ -249,6 +249,9 @@
 
   doc: |
     Get a list of all datasets in a deployment.
+    This includes manifest (attached) and non-manifest (unattached) datasets.
+    A dataset is non-manifest when it is moving between nodes.
```
exarkun: That's just one example of when it's non-manifest but this reads like it's comprehensive.

flocker/control/schema/endpoints.yml
```diff
@@ -182,6 +182,8 @@ definitions:
         path:
           '$ref': 'types.json#/definitions/node_path'
       required:
+      # FLOC-2160 - primary and path will no longer be required
+      # They will be ``None`` in non-manifest datasets
```
exarkun: They're still required, then.  They're just `None` instead of a string.

docs/reference/api_examples.yml
```diff
@@ -249,6 +249,9 @@
...
+    The ``primary`` and ``path`` values of non-manifest datasets will be ``null``.
```
exarkun: We don't include `maximum_size` in the response at all if it's not set on the dataset.  It would make sense to be consistent with that rather than have some behaviors where clients have to check for presence and others where they have to check for None.

flocker/control/schema/endpoints.yml
exarkun: Given comments above about how these values shouldn't be `None`, these should indeed not be required.

docs/reference/api_examples.yml
```diff
@@ -249,6 +249,11 @@
 
   doc: |
     Get a list of all datasets in a deployment.
+    This includes manifest (attached) and non-manifest (unattached) datasets.
+    Some examples of non-manifest datasets are datasets that are being created or
```
itamarst: This is still confusing, I think. How about "For example, a block device volume that has been created but is not attached to any machine." or something along those lines?

flocker/control/_model.py
```diff
@@ -874,16 +874,17 @@ def update_node(self, node_state):
 
     def all_datasets(self):
         """
-        :returns: A generator of all the manifest and non-manifest datasets in
-            the ``DeploymentState``.
+        :returns: A generator of all the primary manifest datasets and
```
itamarst: Actually it generates tuples...

flocker/control/httpapi.py
```diff
@@ -385,20 +385,38 @@ def saved(ignored):
     )
     def state_datasets(self):
         """
-        Return the current primary datasets in the cluster.
+        Return all primary manifest datasets and all non-manifest datasets in
+        the cluster.
 
         :return: A ``list`` containing all datasets in the cluster.
         """
+        # XXX This duplicates code in datasets_from_deployment, but that
```
itamarst: Either file follow up issue or just leave explanation.

flocker/control/test/test_model.py
```diff
@@ -1206,10 +1206,34 @@ def test_nonmanifest_datasets_keys_are_their_ids(self):
 
     def test_all_datasets(self):
```
itamarst: The fact it omits replicas is a temporary expedient, not a goal. I suggest having separate test for that which can then be deleted later on.

itamarst:
Address comments, merge when buildbot is happy. Thanks!

wallrj:
@myechuri wrote:
> We might need a test_blockdevice.py test for to cover all listed states in list_volumes? 

That seems like something for FLOC-2163 (#1555)?

docs/reference/api_examples.yml
wallrj: Done....but it kind of implies that you can create non-manifest datasets....which perhaps we will allow in the future....I remember discussing a use-case with @exarkun which went something like "I want to prepare a batch of datasets in advance to avoid the delay of formatting them all when applications need to use them".

wallrj: I created https://clusterhq.atlassian.net/browse/FLOC-2206

flocker/control/_model.py
wallrj: Done.

flocker/control/httpapi.py
wallrj: Done. https://clusterhq.atlassian.net/browse/FLOC-2207

flocker/control/test/test_model.py
wallrj: Done.

wallrj:
Ok. The test failures seem unrelated or "the usual suspects" so I'll merge.

PR #1539, opened 2015-06-08
[FLOC-2158] Remove Fedora 20 install instructions from the GSG

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2158

This also fixes https://clusterhq.atlassian.net/browse/FLOC-1309 by removing the last untested OS in the GSG.


jongiddy:
OK to merge.

PR #1541, opened 2015-06-09
[FLOC-2141] Add versioned prompt directive

jongiddy:
This adds a versioned prompt directive and applies it to the OS X client instructions, to resolve FLOC-1612.

Needs a follow-up to make other documentation more consistent (FLOC-1388).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1541)


adamtheturtle:
I'd like to know from @tomprince what his preference would be over doing this at all.

adamtheturtle:
I think that if this solution satisfies @tomprince in principal (or maybe he'd like a follow-up) then it is good to merge modulo some minor comments.

flocker/docs/version_extensions.py
```diff
@@ -111,7 +112,31 @@ def run(self):
...
+    Similar to PromptDirective but replaces a placeholder with the
+    latest installable version of Flocker.
+
+    Usage example:
+
+    .. version-prompt:: bash $
+
+       $ brew install flocker-|latest-installable|
+    """
+    def run(self):
```
adamtheturtle: This is repeated code from the VersionCodeBlock. If this code can be neatly shared then please do that.

jongiddy: Ideally, if we decide to convert all code-blocks to prompts, to address FLOC-1388, the code-block version could be deleted. Hence, I think refactoring to share code is not useful at this stage.

adamtheturtle: Link to that issue in the `VersionCodeBlock`.

itamarst:
This doesn't seem to be on critical path to 1.0, nor does 1612. Please merge, but I don't think we have time to follow up on this before 1.0.

itamarst:
Actually, no tests. So, please explain why this needs to be finished now, or just put it off until later.

jongiddy:
This is a fairly old PR, but @davidoliverSP2 has asked for it to be done, as he needs it for the next stage of docs updates.

To address the various concerns expressed here and in the JIRA issue:
- There is now a test for the versioned prompt directive.
- The `version-code-block` directive has been removed, hence no net change in repetition.

I did look at whether it would be possible to create a `versioned-content` directive, to process the content separately to the code formatting, as suggested by @tomprince in the JIRA issue. However, neither the `code-block` nor the `prompt` directives parse their content as RST. Instead they treat their internal content as literal code to be parsed as a programming language.

It would be possible to create a `versioned-content` directive that wrapped the `prompt` directive, but that seems flaky. It may or may not work, depending on whether the `prompt` directive thinks the pipe character should be colorized or not. (We use `|latest-installable|` as the version placemarker).

flocker/docs/test/test_version_extensions.py
```diff
@@ -0,0 +1,32 @@
...
+
+from twisted.python.filepath import FilePath
+from twisted.trial.unittest import SynchronousTestCase
+
+from flocker import __version__ as version
+from flocker.common.version import get_installable_version
+from flocker.testtools import run_process
+
+
+class VersionExtensionsTest(SynchronousTestCase):
```
adamtheturtle: Docstring for class and test

adamtheturtle:
Thanks, re-run tests until they pass ( :cry: ) and then merge.

adamtheturtle:
Or you might be interested in https://pypi.python.org/pypi/sphinx-testing/.

PR #1542, opened 2015-06-09
[FLOC-2159] Warning in pip instructions

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2159

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -223,9 +223,6 @@ Preparing For a Release
 
    Wait for an accepted code review before continuing.
 
-   .. warning:: Add a note to the pull request description explaining that the branch should not be merged until the release process is complete.
-
-
 .. _pre-tag-review:
 
 Pre-tag Review Process
@@ -250,9 +247,12 @@ So it is important to check that the code in the release branch is working befor
    If there are no problems spotted, comment on the Pull Request that the release engineer can continue by following :ref:`the Release section <release>` (do not merge the pull request).
    Otherwise, add comments to the Pull Request for any problems, and comment that they must be resolved before repeating this review process.
 
+#. Merge the release pull request.
+   Do not delete the release branch because it may be used as a base branch for future releases.
+
 #.  Reject the JIRA issue.
 
-    This is necessary because the release branch will need another review.
+    This is necessary because the release process is not "Done" even though no further changes will be made on the branch.
 
 .. _release:
 
@@ -302,7 +302,7 @@ Release
 
       ~/flocker-${VERSION}/admin/test-redirects --production
 
-#. Copy the AWS configuration to your local home directory:
+#. (Optional) Copy the AWS configuration to your local home directory:
 
    If the AWS configuration is on your workstation it will not have to be recreated next time you do a release.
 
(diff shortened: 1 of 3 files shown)


adamtheturtle:
I think that this should fix https://clusterhq.atlassian.net/browse/FLOC-1039 because this makes the one remaining set of instructions have a warning and not really be "supported" - we can't test all "Other Linux Distributions".

docs/releasenotes/index.rst
```diff
@@ -4,8 +4,6 @@ Release Notes
 
 .. note:: If you already have a tutorial environment from a previous release see :ref:`upgrading-vagrant-environment`.
 
-.. note:: If you already have an old version of ``flocker-cli`` installed, delete the install script and directory before installing the newer version.
```
adamtheturtle: These instructions now describe that they may need so tweaks, and they are just guidelines so I don't think that this is necessary.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -312,24 +312,6 @@ Release
       Connection to 127.0.0.1 closed.
       $ vagrant scp default:/home/vagrant/.aws ~/
 
-#. Submit the release pull request for review again.
-
-Post-Release Review Process
```
adamtheturtle: We don't have any QA process remaining if we stop testing these instructions (I think we should) so this section is removed and the merge happens earlier. This will stop us having the constant problem of having to copy and paste already-reviewed release notes and NEWS from one failed release's PR to a new PR.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -223,9 +223,6 @@ Preparing For a Release
 
    Wait for an accepted code review before continuing.
 
-   .. warning:: Add a note to the pull request description explaining that the branch should not be merged until the release process is complete.
-
-
```
jongiddy: Why are these changes here? They do not appear to relate to the issue.

docs/using/installing/index.rst
```diff
@@ -34,14 +34,20 @@ On Ubuntu, the Flocker CLI can be installed from the ClusterHQ repository:
...
 
+.. warning::
+
+   These are guidelines for installing Flocker on a Linux distribution which we do not provide native packages for.
+   These guidelines may require some tweaks, depending on the details of the Linux distribution in use.
+
 Before you install ``flocker-cli`` you will need a compiler, Python 2.7, and the ``virtualenv`` Python utility installed.
-On Fedora 20 you can install these by running:
+
+To install these with the ``yum`` package manager, run:
```
jongiddy: To make this work in general with yum systems, replace @buildsys-build with gcc. If a particular OS requires additional dependencies, let the user figure them out.

jongiddy:
Please address the comments and then merge. Note, FLOC-2116 will be resolved by changing @buildsys to gcc (in the sense that neither Fedora nor CentOS/RHEL should fail to do these steps - whether the client works is another matter that we're explicitly not caring about right now).

Note, these instructions would be half the size if the script was replaced with direct commands and virtualenv used properly - FLOC-2142

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: They are in that manual testing of this part of the process was the last remaining work on the post-release review, so I removed it and moved the merge earlier. This is explained at https://github.com/ClusterHQ/flocker/pull/1542#discussion_r31993933.

jongiddy: OK - I'm not much of a reader.

adamtheturtle:
I tried on a Fedora 20 container with the `gcc` change so I will merge when the docs have built.

> Note, these instructions would be half the size if the script was replaced with direct commands and virtualenv used properly - FLOC-2142

Noted

PR #1543, opened 2015-06-09
Make tag instance instructions not ambiguous any more.

robhaswell:
The AWS instructions mention "Tag instance." as one of the steps, but do not give further instruction.

Clarify that there is nothing specific for the administrator to do here.


robhaswell:
I will add to this PR.

PR #1547, opened 2015-06-09
universe repo needs to be enabled for python-virtualenv, git was missing

Azulinho:
The Getting Started Guide needs python-virtualenv which is available on the 'universe' repository.
This repository is not enabled by default on a fresh Ubuntu install.

Additionality it would be 'nice' to install git as part of the instalation steps, as it is used for the 'git clone' operations later on in the guide.

git is not installed by default (how can anyone live without git ?)


exarkun:
I thought we were removing the `git clone` instructions?

exarkun:
fwiw no need to duplicate the content of the JIRA issue.  A link to the issue being fixed is handy and provides that information.

exarkun:
Seems like we did, now we use curl: <http://doc-dev.clusterhq.com/gettingstarted/index.html#step-2-deploying-a-demo-app>

Maybe there's another section?

Azulinho:
not relevant as per new 1.0 docs

PR #1548, opened 2015-06-09
[FLOC-2100] Rename the Using/Configuring Flocker topics

davidoliverSP2:


docs/using/config/index.rst
```diff
@@ -1,6 +1,6 @@
-===================
-Configuring Flocker
-===================
+==================
+Setting up Flocker
```
itamarst: This isn't super accurate either. "Setting up Flocker" sounds just like "Install Flocker". This isn't about setting *flocker* up, it's about setting up datasets and applications *using* Flocker.

docs/using/config/configuration.rst
```diff
@@ -1,8 +1,8 @@
 .. _configuration:
 
-===================
-Configuring Flocker
-===================
+===============================
+Specifying your Flocker Cluster
```
itamarst: You're not specifing a flocker cluster. You're running some applications and creating some datasets/volumes.

The goal of the user here is not to do something to flocker, it's to run a database or whatever. Flocker is just a tool.

itamarst: To put it another way, "Flocker Cluster" is ambiguious - is it the set of servers that Flocker consists of (flocker-control, flocker-dataset-agent, etc.) or the set of servers or datasets the user runs utilizing Flocker. In this case it's the latter, but the title doesn't express that.

itamarst:
I think this needs a different set of phrases.

itamarst:
Great, much better. Thank you!

PR #1549, opened 2015-06-09
[FLOC-2124] Rm fedora20 packaging

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2124

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1549)


adamtheturtle:
Remaining mentions of Fedora:
* Old NEWS / What's New
* Links to Fedora docs about RPMs which are still relevant
* A testing example
* Descriptions in "Legacy" sections
* Instructions for MongoDB install with yum

I think that these are all fine.

docs/gettinginvolved/infrastructure/packaging.rst
```diff
@@ -145,6 +140,8 @@ go to the relevant directory in `admin/release-packaging` and run:
 Legacy
 ------
 
+Fedora packages were published to Amazon S3 until version 1.0.0.
```
exarkun: maybe make it clear that this excludes 1.0.0

flocker/provision/_install.py
```diff
@@ -902,8 +871,6 @@ def provision(distribution, package_source, variants):
     """
     commands = []
 
-    if Variants.DISTRO_TESTING in variants:
```
exarkun: `DISTRO_TESTING` is still defined as a variant but it looks like it's basically no longer supported.  This might be confusing.

I suggest leaving this, leaving `task_enable_updates_testing`, but make the task always raise the unsupported distribution exception.  Hopefully we'll be able to do some testing-distribution testing for other distributions eventually.

exarkun:
Thanks.  Please address those points and then merge when buildbot is sufficiently non-red.

PR #1550, opened 2015-06-09
[FLOC-2154] Restart Flocker services on failure

jongiddy:


admin/package-files/systemd/flocker-container-agent.service
```diff
@@ -3,6 +3,7 @@ Description=Flocker Container Agent
 
 [Service]
 ExecStart=/usr/sbin/flocker-container-agent
+Restart=on-failure
```
itamarst: My impression from reading docs is that this will *not* restart if it exited cleanly with exit code 0. So perhaps we want `Restart=always` just to be safe.

itamarst:
Change to always restart, merge when relevant buildbot builder pass (the acceptance test ones I guess).

jongiddy:
Acceptance tests are failing due to Rackspace issues. Howerer, I have started a control/agent node, and killed all 3 processes to test that they are restarted.

PR #1551, opened 2015-06-09
[FLOC-2096] Provide private -> public IP mappings on AWS.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2096


tomprince:
I saw some acceptance tests pass locally. Hopefully they will on buildbot too. (Not all of the tests, though).

admin/acceptance.py
```diff
@@ -79,11 +80,15 @@ def get_trial_environment(cluster):
...
-        'FLOCKER_ACCEPTANCE_AGENT_NODES':
-            ':'.join(node.address for node in cluster.agent_nodes),
+        'FLOCKER_ACCEPTANCE_NUM_AGENT_NODES': str(len(cluster.agent_nodes)),
         'FLOCKER_ACCEPTANCE_VOLUME_BACKEND': cluster.dataset_backend.name,
         'FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH':
             cluster.certificates_path.path,
+        'FLOCKER_ACCEPTANCE_HOST_MAPPING': json.dumps({
+            node.private_address: node.address
+            for node in cluster.agent_nodes
+            if node.private_address is not None
```
itamarst: Doesn't this break OpenStack acceptance tests?

flocker/acceptance/testtools.py
```diff
@@ -215,9 +216,11 @@ class Node(PRecord):
     """
     A record of a cluster node.
 
-    :ivar bytes address: The IPv4 address of the node.
+    :ivar bytes address: The public address of the node.
+    :ivar bytes hostname: The address of the node, as reported by the API.
```
itamarst: This is a rather confusing set of attributes... Maybe `reported_address` instead of `hostname`?

flocker/acceptance/testtools.py
```diff
@@ -777,10 +785,17 @@ def failed_query(failure):
     # happen know these in advance, but in FLOC-1631 node identification
     # will switch to UUIDs instead.
     agents_connected.addCallback(lambda _: cluster.current_nodes())
+
+    def node_from_dict(node):
+        hostname = node["host"]
+        address = host_mapping.get(hostname, hostname)
```
itamarst: Confusing. Maybe if `host_mapping` was `private_to_public`, and then `public_address = private_to_public.get(hostname, hostname)`.

admin/acceptance.py
itamarst: Ah, no, because the mapping isn't necessary given cluster API for case where there's no private/public mapping, OK.

flocker/acceptance/testtools.py
itamarst: And `public_address` instead of `address`.

itamarst:
I gather `agent.yml` is using public IP now for address of control service? If so file follow-up issue to fix that, since we're testing unrealistic configuration (real cluster would point at private AWS IP, or hostname that resolves appropriately more likely).

itamarst:
Seems like some testing code still needs updating, e.g. `assert_expected_deployment` should be using `node.hostname` (or really `node.reported_address` since the former is confusing).

itamarst:
1. Change attribute names to be less confusing.
2. Address any other comments above, e.g. there's some tests that will pass I think given another fix.
3. Merge when you get a completely passing build on AWS, rackspace and vagrant (modulo the test_move_data issues in FLOC-2050 I guess).

itamarst:
Or at least builds on the three environments where it's sufficiently clear you didn't introduce any new failures.

flocker/acceptance/testtools.py
tomprince: I picked `hostname` to match the `host` in the API.

itamarst:
1. Vagrant acceptance tests ran.
2. Rackspace acceptance tests ran.

Just need running AWS acceptance tests and this can be merged - I restarted one.

PR #1552, opened 2015-06-09
[FLOC-2135] Only invalidate agent-reported state if the agent really goes away

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2135

This includes:

  * Changes to both sides of the AMP connection to send "ping" commands periodically.  This forces TCP-level disconnection notification in a timely manner.
  * Changes to use AMP command dispatch as a signal of "activity" and considers this sufficient evidence that the agent is still operating properly.
  * Changes to use "activity" (or inactivity) as the trigger for wiping state instead of using a hard-coded timeout on the age of the state itself.

It's probable this resolves FLOC-1896 as well.  The requirements there just say "do better".  This is probably better.

### DIFF
--- a/flocker/control/_clusterstate.py
+++ b/flocker/control/_clusterstate.py
@@ -6,6 +6,8 @@ Combine and retrieve current cluster state.
 
 from datetime import datetime, timedelta
 
+from twisted.python.versions import Version
+from twisted.python.deprecate import deprecated
 from twisted.application.service import MultiService
 from twisted.application.internet import TimerService
 
@@ -17,6 +19,7 @@ from ._model import DeploymentState, ChangeSource
 # Allowed inactivity period before updates are expired
 EXPIRATION_TIME = timedelta(seconds=120)
 
+v1_0 = Version("flocker", 1, 0, 0)
 
 class _WiperAndSource(PRecord):
     """
@@ -121,6 +124,7 @@ class ClusterStateService(MultiService):
                 key, _WiperAndSource(wiper=wiper, source=source)
             )
 
+    @deprecated(v1_0, "ClusterStateService.apply_changes_from_source")
     def apply_changes(self, changes):
         """
         Compatibility layer.  See and use ``apply_changes_from_source``.
--- a/flocker/control/_clusterstate.py
+++ b/flocker/control/_clusterstate.py
@@ -4,18 +4,33 @@
 Combine and retrieve current cluster state.
 """
 
-from datetime import timedelta
+from datetime import datetime, timedelta
 
 from twisted.application.service import MultiService
 from twisted.application.internet import TimerService
 
-from pyrsistent import pmap
+from pyrsistent import PRecord, field, pmap
 
-from ._model import DeploymentState
+from ._model import DeploymentState, ChangeSource
 
 
 # Allowed inactivity period before updates are expired
 EXPIRATION_TIME = timedelta(seconds=120)
+
+
+class _WiperAndSource(PRecord):
+    """
+    :ivar IClusterStateWipe wiper: A change wiper.
+    :ivar IClusterStateSource source: Where the change wiper came from.
+    """
+    wiper = field()
+    source = field()
+
+    def last_activity(self):
+        return self.source.last_activity()
+
+    def update_cluster_state(self, deployment_state):
+        return self.wiper.update_cluster_state(deployment_state)
 
 
 class ClusterStateService(MultiService):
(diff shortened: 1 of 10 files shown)


flocker/control/_clusterstate.py
```diff
@@ -4,16 +4,33 @@
...
+class _WiperAndSource(PRecord):
+    """
+    :ivar IClusterStateWipe wiper: A change wiper.
+    :ivar IClusterStateSource source: Where the change wiper came from.
+    """
+    wiper = field()
+    source = field()
+
+    def last_activity(self):
+        return self.source.last_activity()
```
itamarst: Docstring.

flocker/control/_clusterstate.py
```diff
@@ -4,16 +4,33 @@
...
 
-# Seconds until updates are expired:
-EXPIRATION_TIME = 120
+    def update_cluster_state(self, deployment_state):
```
itamarst: Docstring.

flocker/control/_clusterstate.py
```diff
@@ -90,4 +112,13 @@ def apply_changes(self, changes):
...
             self._information_wipers = self._information_wipers.set(
-                key, (wiper, self._clock.seconds()))
+                key, _WiperAndSource(wiper=wiper, source=source)
+            )
+
+    def apply_changes(self, changes):
+        """
+        Compatibility layer.  See and use ``apply_changes_from_source``.
+        """
+        source = ChangeSource()
```
itamarst: Add deprecation warning.

flocker/control/test/test_clusterstate.py
```diff
@@ -156,17 +160,46 @@ def test_expiration(self):
         (in seconds) old are wiped.
         """
         service = self.service()
-        app_node = NodeState(hostname=u"10.0.0.1", uuid=uuid4(),
-                             manifestations=None, devices=None, paths=None,
-                             applications=[APP1], used_ports=[])
-        service.apply_changes([app_node])
-        self.clock.advance(EXPIRATION_TIME - 1)
+        service.apply_changes([self.WITH_APPS])
```
itamarst: `apply_changes_with_source` is now the recommended public API to use, and in fact what does get used. But only one of the tests uses it. I would suggest all the tests use it and maybe have one bonus test for the backwards compat layer.

flocker/control/test/test_clusterstate.py
```diff
@@ -156,17 +160,46 @@ def test_expiration(self):
...
+        advance_rest(self.clock)
+        before_wipe_state = service.as_deployment()
+        advance_some(self.clock)
```
itamarst: :question: You swapped rest and some. Not that it really matters.

flocker/control/test/test_protocol.py
```diff
@@ -95,6 +101,14 @@ def massage_error(error):
          applications=frozenset([APP1, APP2]))]))
 MANIFESTATION = Manifestation(dataset=Dataset(dataset_id=unicode(uuid4())),
                               primary=True)
+
+# A very simple piece of node state that makes for nice-looking, easily-read
+# test failures.  It arbitrarily supplies only ports because integers have a
+# very simple representation.
+SIMPLE_NODE_STATE = NodeState(
+    hostname=u"192.0.2.17", applications=[], used_ports=[1],
```
itamarst: uuid

itamarst:
Failing tests need addressing.

flocker/control/_model.py
```diff
@@ -683,6 +684,44 @@ def key():
...
+        """
+        Set the time of the last activity.
+
+        :param float since_epoch: Number of seconds since the epoch at which
+            point the activity occurred.
+        """
+        self._last_activity = since_epoch
+
+    def last_activity(self):
+        return datetime.utcfromtimestamp(self._last_activity)
```
itamarst: There is no `utctimestamp`, there is only `timestamp(..., pytz.UTC)`.

itamarst:
Thanks! Address, and merge when buildbot is happy.

flocker/control/test/test_clusterstate.py
exarkun: I left the backwards compat layer so I wouldn't have to mess with all of those tests.  :/

flocker/control/_model.py
exarkun: Er?  There is, though.  I'm not sure what you mean.

flocker/control/test/test_clusterstate.py
exarkun: Filed https://clusterhq.atlassian.net/browse/FLOC-2202 for just updating all the tests and then removing the old method.

flocker/control/_model.py
exarkun: `utcfromtimestamp` returns a naive datetime.  That's the problem.  Naive datetime objects are the worst.  We should never ever have them.  Filing a follow-up issue to fix this.

PR #1554, opened 2015-06-09
[FLOC-2198] Attempt to retry SSH key installation on AWS

itamarst:
Not actually tested in anyway, so chances are this won't actually pass buildbot at all.


itamarst:
Ok, I added a sleep to force the problem to happen, and the fix worked:
http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fcentos-7%2Faws/builds/222

tomprince:
There is probably a more principled way to do this, so file an issue to improve this post-1.0. For example, have an intent that does this, and have the intent initialize the time, rather than mutating something in the closure.

PR #1555, opened 2015-06-09
[FLOC-2165] EBS: ``list_volumes`` should list volumes in all states.

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2165

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -475,13 +475,11 @@ class EBSBlockDeviceAPI(object):
 
     def list_volumes(self):
         """
-        Return all volumes in {available, in-use} state that belong to
-        this Flocker cluster.
+        Return all volumes that belong to this Flocker cluster.
         """
         volumes = []
         for ebs_volume in self.connection.get_all_volumes():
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks @myechuri 

This makes sense, but do you think we've got time to add a test that demonstrates the behaviour.

It could be an implementation specific test method added to ``EBSBlockDeviceAPIInterfaceTests`` which patches ``boto.get_all_volumes`` to return real(istic) boto volumes in all the documented boto volume states.

Having said that, we don't have such a test for the Cinder API...yet. Is that a good enough excuse to not test this change (@exarkun)?

Do the test if you agree and if it's reasonably easy. Discuss it on Zulip if you don't think it's worth it and then merge when you've reached a decision and when the builds pass.

exarkun:
Deleting code without adding tests is frequently an okay thing to do.

exarkun:
(And adding a verified fake for part of Boto is probably a looooot of work.)

wallrj:
Ok. So I may as well merge this now so that it's included in the next dev / pre release.

PR #1556, opened 2015-06-10
[FLOC-2191] Fix content type for Vagrant metadata file

jongiddy:


adamtheturtle:
I think that this is a better solution than the dictionary I mentioned in the ticket. A proper review would involve publishing to S3 and checking files and their content types. I don't think that doing this well now is a priority - I think that the JSON metadata in the dictionary is a lower risk feature that we can ship and test today quickly. To be safer than just that, perhaps add similar content type dictionary checking in the FakeAWS performer to make sure that our test_release.py code at least doesn't upload anything with an unrecognised file name. In my opinion the best thing to do would be to keep this around until after 1.0 and add a couple of lines in a new PR to submit now.

jongiddy:
The mapping solution requires you to copy the JSON file within S3. When does that get done in the current release process? An even simpler solution that will work is to add a step to the release process to set the content type correctly.

jongiddy:
Added a functional test to check file has correct content-type

adamtheturtle:
> The mapping solution requires you to copy the JSON file within S3. When does that get done in the current release process?

```python
                Effect(
                    CopyS3Keys(
                        source_bucket=DEV_ARCHIVE_BUCKET,
                        source_prefix=vagrant_prefix,
                        destination_bucket=options['target'],
                        destination_prefix=vagrant_prefix,
                        keys=[box_name],
                    )
                ),
```

in `publish_artifacts_main`

jongiddy:
That happens before the JSON file is uploaded

adamtheturtle:
Yes, then when it is copied the `CopyS3Keys` uses the dictionary to set the content type.

admin/aws.py
```diff
@@ -339,6 +339,7 @@ def perform_upload_s3_key_recursively(dispatcher, intent):
     "target_bucket",
     "target_key",
     "file",
+    Attribute("content_type", default_value=None),
```
exarkun: This should go into the docstring immediately below

admin/aws.py
```diff
@@ -397,6 +401,11 @@ class FakeAWS(object):
     """
     def __init__(self):
         self.cloudfront_invalidations = []
+        for bucket in self.s3_buckets.values():
+            for key, contents in bucket.items():
+                if isinstance(contents, basestring):
```
exarkun: `basestring` always makes me wary.  Does the real API really work correctly with either bytes or unicode?  Or does it work with some unicode some of the time by accident?  I suspect this should really only allow bytes here but if both bytes and unicode are meant to work, I'd suggest going with `(bytes, unicode)` to be explicit about it.

Alternatively, the values of this dictionary should be a PRecord (or characteristic) thing instead of a tuple, and then this could just be `not isinstance(contents, TypedContents)` instead.

Explicitly rejecting unicode here if unicode isn't how the real AWS APIs work might be valuable, though.

admin/aws.py
```diff
@@ -471,7 +480,45 @@ def _perform_upload_s3_key(self, dispatcher, intent):
...
         bucket = self.s3_buckets[intent.target_bucket]
         with intent.file.open() as source_file:
-            bucket[intent.target_key] = source_file.read()
+            bucket[intent.target_key] = source_file.read(), intent.content_type
+
+    def get_bucket_keys(self, bucket):
+        """
+        Retrieve the keys from a bucket in S3.
+
+        :param str bucket: name of the S3 bucket.
```
exarkun: This and following - `bytes` or `unicode` (almost certainly `bytes`).  `str` makes it too easy to forget about the difference between the two types.  It's probably only correct as the type of things like Python identifiers (eg, the 2nd argument of `getattr` or the elements of the return value of `dir`).

admin/functional/test_aws.py
```diff
@@ -0,0 +1,53 @@
...
+
+if_aws = skipIf(not _can_connect, "Requires boto AWS credentials")
+
+
+class AWSTest(SynchronousTestCase):
+
+    @if_aws
+    def test_upload_content_type(self):
+        """
+        Test that a content type can be set for an uploaded file.
```
exarkun: Drop the "Test that" prefix.  Just describe desired behavior.

admin/functional/test_aws.py
```diff
@@ -0,0 +1,53 @@
...
+        """
+        filename = random_name(self)
+        tmpdir = self.mktemp()
+        os.mkdir(tmpdir)
+        self.addCleanup(shutil.rmtree, tmpdir, ignore_errors=True)
```
exarkun: trial will clean up temporary directories from `self.mktemp` automatically on the next test run.  Better to let stuff like this stay around in case it ends up being useful for debugging.

admin/functional/test_aws.py
```diff
@@ -0,0 +1,53 @@
...
+
+class AWSTest(SynchronousTestCase):
+
+    @if_aws
+    def test_upload_content_type(self):
+        """
+        Test that a content type can be set for an uploaded file.
+        """
+        filename = random_name(self)
+        tmpdir = self.mktemp()
```
exarkun: Make a `FilePath` of this here and you can use great `FilePath` methods for the rest of the test. :smile:

admin/aws.py
exarkun: It might also make sense to add an alternate constructor which requires the PRecord thing rather than overloading `__init__` like this.

admin/aws.py
```diff
@@ -471,7 +480,45 @@ def _perform_upload_s3_key(self, dispatcher, intent):
...
+
+    def get_content_type(self, bucket, key):
+        """
+        Retrieve the content of a file in S3.
+
+        :param str bucket: name of the S3 bucket.
+        :param str key: key for the S3 file.
+        :returns: the contents of the file.
+        """
+        return self.s3_buckets[bucket][key][1]
```
exarkun: I think that `FakeAWS` is supposed to be a fake of an object which only has one method on its interface, `get_dispatcher`.  Adding these public methods makes this code a little more confusing because it makes it seem like these are methods we expect to be available from the real implementation (perhaps from Boto or something).

A quick fix would be to make these private like the rest of the methods.  And/Or add documentation clarifying what `FakeAWS` is a fake of, exactly.  Longer term, it would probably make sense to separate the useful helper functions like `get_content_type` away from the ... Effect dispatcher implementation (I think that's what this is?) so the latter can be passed in to implementation code for testing and the former held back for use by the tests themselves.

admin/test/test_release.py
```diff
@@ -903,6 +903,8 @@ def test_fake_rpm(self):
                          '<oldhash>-metadata.xml'):
                 'metadata for: existing_package.rpm',
         }
+        # Copy before passing to FakeAWS
+        expected_keys = existing_s3_keys.copy()
```
exarkun: Change `FakeAWS` so that this isn't necessary.  ie, `FakeAWS` should be doing the copy instead of mutating the argument passed to it.

admin/test/test_release.py
```diff
@@ -1414,26 +1417,25 @@ class UploadOptionsTests(SynchronousTestCase):
...
-              options.parseOptions,
-              ['--flocker-version', '0.3.0+doc1'])
+        """
+        Trying to upload artifacts for a documentation version fails.
+        """
+        options = UploadOptions()
+        self.assertRaises(
+            DocumentationRelease,
+            options.parseOptions,
+            ['--flocker-version', '0.3.0+doc1'])
```
exarkun: erk.  Thanks.

exarkun:
Thanks.  Address inline comments, file follow-up issue, and merge.

admin/aws.py
jongiddy: Removed all these changes in favour of a subclass of `unicode` that has an additional `content_type` attribute. This allows all existing tests to work as before.

admin/test/test_release.py
jongiddy: As this was a fix for an existing bug in the tests, unrelated to the changes done for this PR, I decided not to change the implementation code.

admin/aws.py
jongiddy: Done.

admin/test/test_release.py
jongiddy: My editor is set to "super-annoying" for PEP8 violations.

PR #1557, opened 2015-06-10
[FLOC-2016] rename fig.yml in gettingstarted repo

davidoliverSP2:
[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1557)


adamtheturtle:
Thanks @davidoliverSP2 . I don't like how this fits in with the rest of the docs in that they describe this configuration file as "fig-compatible". I'm assuming that you have tested this. Please either:
* Modify other docs which mention Fig like the one below to describe docker compose and resubmit
* Merge and start work on a follow-up now
* Merge and create a follow up for after 1.0

I'll let you choose what is best for the docs.

```
.. _fig-compatible-config:

Fig-compatible Application Configuration
----------------------------------------

As an alternative to Flocker's configuration syntax, you may also use `Fig`_'s configuration syntax to define applications.
```

adamtheturtle:
I know you use Textmate - one way to search for those is Edit > Find > Find in Project, then use the search box as below (to rule out "configure, configuration") from search terms.

screenshot 2015-06-10 11 40 35

PR #1558, opened 2015-06-10
Don't start flocker services before network is up

Azulinho:
Addresses https://clusterhq.atlassian.net/browse/FLOC-2136


exarkun:
#1536 also contains a proposed fix for this issue.  Itamar pointed out that systemd maintainers strongly urge people not to depend on the network service.  Also this doesn't address the problem on Ubuntu.

PR #1560, opened 2015-06-10
[FLOC-2194] add /etc/flocker instructions

davidoliverSP2:


robhaswell:
Text content is fine, however the prompt is rendered incorrectly in Chrome:

screen shot 2015-06-10 at 13 34 35

Please fix the double-prompt, check it is gone in the build (ask if you know how to do this) and then merge.

Thank you for this fix.

robhaswell:
I'm itchy about the inconsistent presentation but otherwise I'm happy to :+1: this as you said you will fix that another time.

PR #1561, opened 2015-06-10
[FLOC-2210] No agent yaml in install docs

dwgebler:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2210

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1561)


davidoliverSP2:
It looks like the Ubuntu section still has the EOF section in the codeblock: http://build.clusterhq.com/results/docs/no-agent-yaml-install-docs-FLOC-2210/build-7893/using/installing/index.html#enabling-the-flocker-control-service-on-ubuntu

dwgebler:
> It looks like the Ubuntu section still has the EOF section in the codeblock

Different section, this issue was about the `agent.yml` file in the enabling the Flocker agents section:

http://build.clusterhq.com/results/docs/no-agent-yaml-install-docs-FLOC-2210/build-7893/using/installing/index.html#ubuntu

davidoliverSP2:
ah ok, cool

flocker/provision/_install.py
```diff
@@ -517,48 +516,48 @@ def task_open_control_firewall(distribution):
...
+                    },
+                    "dataset": dataset_backend_configuration,
                 },
-                "dataset": dataset_backend_configuration,
-            },
-        ),
-    )
+            ),
+        )
+        initial_sequence.append(put_config_file)
```
adamtheturtle: `intial_sequence` should be renamed to something descriptive. `put_config_file` or similar perhaps.

flocker/provision/_install.py
```diff
@@ -517,48 +516,48 @@ def task_open_control_firewall(distribution):
...
 
 def task_enable_flocker_agent(distribution, control_node,
-                              dataset_backend=DatasetBackend.zfs,
-                              dataset_backend_configuration=dict(
-                                  pool=u'flocker'
-                              )):
+                              dataset_backend=None,
+                              dataset_backend_configuration=None):
     """
     Configure and enable the flocker agents.
```
adamtheturtle: This only "Configure"s the agents in certain circumstances, sometimes it assumes that they are configured already. Update the docstring accordingly.

flocker/provision/_install.py
```diff
@@ -517,48 +516,48 @@ def task_open_control_firewall(distribution):
 
 
 def task_enable_flocker_agent(distribution, control_node,
```
adamtheturtle: Now our tests are testing something different to the docs. I think that this would perhaps be split up into two separate tasks:
* One which creates an `agent.yml`.
* One which starts the agents.

This would allow:
* All the functionality to be in the docs and tested.
* The docs to use the former task in the ZFS section above where these commands are rendered, in place perhaps of the example YAML.
* The docs to use the former task with different dataset arguments (maybe not now) for each other backend to allow easy testing of the docs.

I like having the `cat <<` command just because it stops someone having to write that themselves.

adamtheturtle:
Thanks @dwgebler please address and resubmit. Or don't - I'm not sure if this is on the critical path to 1.0 or was just added to the sprint because it was expected to not use engineering time. @robhaswell ?

robhaswell:
Definitely critical path. Thanks!

itamarst:
Note to reviewer: apparently we want this in 1.0.1, but this is against master.

dwgebler:
I have a separate PR now to merge this in to 1.0 release branch so it can get in to 1.0.1

https://github.com/ClusterHQ/flocker/pull/1643

dwgebler:
Also https://github.com/ClusterHQ/flocker/pull/1644 is this rebased against master.

PR #1563, opened 2015-06-10
[FLOC-2192] Don't attempt to destroy nonmanifest datasets 

wallrj:
Design / fix for https://clusterhq.atlassian.net/browse/FLOC-2192

There's a test in this branch which demonstrates at least one of the problems we're seeing in the logs attached to FLOC-2192

```
Jun 09 16:41:29 acceptance-test-buildbot-0 flocker-dataset-agent[12516]: {"timestamp": 1433868088.771906, "task_uuid": "3701d26f-b458-407e-914c-5a544f
0d42ad", "message_type": "flocker:agent:converge:actions", "calculated_actions": "_InParallel(changes=pvector([DestroyBlockDeviceDataset(dataset_id=UU
ID('9ea3afb6-c3ca-4bef-8f0a-774a58fc005d')), MountBlockDevice(mountpoint=FilePath('/flocker/9ea3afb6-c3ca-4bef-8f0a-774a58fc005d'), dataset_id=UUID('9
ea3afb6-c3ca-4bef-8f0a-774a58fc005d'))]))", "task_level": [2, 4]}
```

The problem is that ``DestroyBlockDeviceDataset`` is a compound state change which attempts to perform:
 * unmount
 * detach
 * destroy
 ...in series.

Therefore the target dataset must be mounted for this to succeed.

Some ideas:
 1. Change the order of state changes so that ``MountBlockDevice`` always comes before ``DestroyBlockDeviceDataset`` and return an ``_InSeries`` state change collection....but that's probably too slow.

 2. Change the ``_calculate_deletes`` helper to only calculate deletes for fully manifest datasets. (by fully manifest, I mean that the dataset is attached and mounted). I could base this on the presence of a path key matching the deleted dataset ID but actually ``discover_state`` already moves datasets to ``non_manifest_datasets`` if the dataset is attached but not mounted.

 3. Change the ``_calculate_mounts``, ``_calculate_attaches``, ``_calculate_unmounts``, ``_calculate_detaches`` to only consider extant (not deleted) datasets in the configuration.

 4. And probably best of all would be to get rid of all the compound state changes and calculate all the intermediate steps separately.

After discussing with @itamarst I went for No2.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1563)

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -894,6 +894,28 @@ class BlockDeviceDeployerDestructionCalculateChangesTests(
             in_parallel(changes=[]),
         )
 
+    def test_deleted_dataset_volume_unmounted(self):
+        """
+        ``DestroyBlockDeviceDataset`` is a compound state change that first
+        attempts to unmount the block device.
+        Therefore do not calculate deletion for blockdevices that are attached
+        but unmounted.
+        """
+        local_state = self.ONE_DATASET_STATE
+        local_config = to_node(local_state).transform(
+            ["manifestations", unicode(self.DATASET_ID), "dataset", "deleted"],
+            True
+        )
+        # Remove the mount path for the manifestation.
+        local_state = local_state.transform(
+            ['paths', unicode(self.DATASET_ID)],
+            discard
+        )
+        assert_calculated_changes(
+            self, local_state, local_config, set(),
+            in_parallel(changes=[]),
+        )
+
 
 class BlockDeviceDeployerAttachCalculateChangesTests(
         SynchronousTestCase, ScenarioMixin
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -915,10 +915,19 @@ class BlockDeviceDeployerDestructionCalculateChangesTests(
             ['paths', unicode(self.DATASET_ID)],
             discard
         )
-
+        # Local state shows that there is a device for the (now) non-manifest
+        # dataset. i.e it is attached.
+        self.assertEqual([self.DATASET_ID], local_state.devices.keys())
         assert_calculated_changes(
-            self, local_state, local_config, set(),
-            in_parallel(
+            case=self,
+            node_state=local_state,
+            node_config=local_config,
+            # The unmounted dataset has been added back to the non-manifest
+            # datasets by discover_state.
+            nonmanifest_datasets=[
+                self.MANIFESTATION.dataset
+            ],
+            expected_changes=in_parallel(
                 changes=[
                     MountBlockDevice(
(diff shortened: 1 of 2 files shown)


wallrj:
I pushed the branch with the fix disabled and then again with it enabled and the build results will eventually be here:
 * http://build.clusterhq.com/boxes-flocker?branch=destroy-when-mounted-FLOC-2192&num_builds=2

It'll be interesting to see if the rackspace + rackspace acceptance tests pass on the second build.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -894,6 +894,42 @@ def test_no_delete_if_in_use(self):
...
+            in_parallel(
+                changes=[
+                    MountBlockDevice(
+                        mountpoint=FilePath('/flocker/').child(
+                            unicode(self.DATASET_ID)
+                        ),
+                        dataset_id=self.DATASET_ID
+                    )
+                ]
+            ),
```
exarkun: My guess would be that this test doesn't pass because the dataset doesn't exist anywhere in the state so a create will be calculated instead.  Put the dataset into the non-manifest state to avoid a create - and then you should get an attach.

exarkun:
Thanks.  Great diagnosis, nice looking test (modulo minor comment inline).  Issue is marked for design review but I think you have more or less the entire fix done here... So clicking "accept" on the issue to accept the design.  If the change I suggested makes sense and you can get the relevant tests to pass on buildbot, go ahead and merge without a further review.

exarkun:
Eh, @itamarst clicked some buttons in jira while I wasn't looking.  Anyway, what I said above, accepted.

flocker/node/agents/test/test_blockdevice.py
wallrj: I know what you mean but as it stands, the list of creates is generated from a filtered list of configured datasets where `deleted=False`.
Only then does it check the cluster state for existing manifest and non-manifest datasets.

A Mount rather than an Attach is calculated because we're supplying state where there's a device file but not a mount path for a dataset.

itamarst:
Is this mergeable?

wallrj:
I was planning to run the acceptance tests manually with my own credentials since buildbot's rackspace limits have been reached.

But maybe we should just merge it anyway.

exarkun:
Have you seen the test pass anywhere?

wallrj:
Ok. So unit test passes on
 * Ubuntu 14.04: http://build.clusterhq.com/builders/flocker-ubuntu-14.04/builds/4423
 * Centos-7: http://build.clusterhq.com/builders/flocker-centos-7/builds/4586

And the DatasetAPI deletion tests still pass with the Loopback backend:
 * http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fcentos-7%2Floopback/builds/387

Which we think is good enough.

PR #1564, opened 2015-06-10
[FLOC-2197] Try to explain that firewalld may not be installed.

itamarst:


docs/using/installing/index.rst
```diff
@@ -481,8 +481,10 @@ Enabling the Flocker control service on CentOS 7
 .. task:: enable_flocker_control centos-7
    :prompt: [root@control-node]#
 
-The control service needs to accessible remotely.
-To configure FirewallD to allow access to the control service HTTP API, and for agent connections:
+The control service needs to be accessible remotely.
+You will therefore need to configure FirewallD to allow access to the control service HTTP API, and for agent connections.
```
exarkun: Too many words.  "You need to ..."

exarkun: Extraneous comma here (it was extraneous before too, but as long as you're changing the line...)

docs/using/installing/index.rst
```diff
@@ -481,8 +481,10 @@ Enabling the Flocker control service on CentOS 7
...
+Note that on some environments, in particular AWS, the ``firewalld`` package is not installed and the ``firewall-cmd`` program will not be found.
+If that is the case then just skip these commands, otherwise run:
```
exarkun: `... these commands.  Otherwise, run:`

exarkun:
Thanks.  Looks good apart from trivial comments noted inline.  Please fix and merge.

PR #1565, opened 2015-06-10
[FLOC-2224] Don't expect all datasets to have a path

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2224


wallrj:
Thanks @itamarst 

Yeah, this will have been introduced by #1537.

I guess missed this because of all the expected failures in the acceptance tests at the time.

Please merge if the eventual acceptance test results are free of path KeyErrors .

PR #1567, opened 2015-06-10
[FLOC-1979] Client certificate instructions that work on OS X

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1979

The tabs appear to **not** render correctly in the branch as-is.  If someone knows how to fix that, I'd love to know (or just check it into the branch).  If not, then I guess we'll drop the tabs and just have one section after the other.

I also added a link to this section of the docs from the other section about certificates since the other section leaves the user hanging.

I also didn't add any instructions about the pkcs12 export password.  Possibly I should have?

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1567)

### DIFF
--- a/docs/_themes/clusterhq/layout_docs.html
+++ b/docs/_themes/clusterhq/layout_docs.html
@@ -24,9 +24,12 @@
     <script type="text/javascript" src="https://fast.fonts.net/jsapi/b20e6cd5-c1a4-4878-9d69-da3806cb86b8.js"></script>
     <link href='https://fonts.googleapis.com/css?family=Lato&subset=latin,latin-ext' rel='stylesheet' type='text/css'>
 
+    <!--
+         Enable Bootstrap tabs in the documentation. There's probably a better
+         way to get these dependencies loaded.  FLOC-2248.
+    -->
     <script src="//ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
     <script>window.jQuery || document.write('<script src="{{ pathto('_static/js/vendor/jquery-1.11.1.min.js', 1) }}"><\/script>')</script>
-
     <script src="{{ pathto('_static/js/vendor/bootstrap.min.js', 1)}}"></script>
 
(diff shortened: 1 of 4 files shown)


davidoliverSP2:
Cheers for picking this up JP

@alicesowerby Could you take a look at http://build.clusterhq.com/results/docs/curl-on-osx-FLOC-1979/build-7940/reference/authentication.html#li-os-x, and let us know what you think about having tabs in the documentation from a UX point of view?

My view is that this would be an excellent way of reducing the amount of scrolling and searching, but its something we'd need roll out across the whole site. As it stands, this would be the only tabbed section (that I know of) in the docs, and I'm also not sure we've enough time to get it working correctly.

exarkun:
Possibly I shouldn't have spent the time, but I tracked down the cause of the tabs not working and fixed it.  You can see them operating properly on the most recent docs build, http://build.clusterhq.com/results/docs/curl-on-osx-FLOC-1979/build-7985/reference/authentication.html

> As it stands, this would be the only tabbed section (that I know of) in the docs, and I'm also not sure we've enough time to get it working correctly.

The old (current) getting started guide uses them, of course. :wink: 

Please let me know if we should go ahead with this or revert to a one-after-the-other style for this section.

alicesowerby:
As I commented previously... When we have a defined problem to solve and that problem has been prioritised I am happy to do the UX/interaction design to solve it. AFAIK this has not been prioritised as a problem at this point in time and therefore I support Dave's implied position of leaving this until we can do it properly and in a considered manner. 

FWIW it's probably a good solution to a probable problem :)

davidoliverSP2:
I would recommend reverting this to one-after-the-other style for this section, as this matches the continuity of the rest of the docs.

I really like this feature though, and have created FLOC 2251 to keep track, and post v1 @alicesowerby and I can do a full UX design, and it would then be rolled out across the site.

Thanks JP

exarkun:
Unhappy that the decision is to do more work for probably a worse result - but okay, here I go.

PR #1568, opened 2015-06-10
[FLOC-1953] EBS: retry attach_volume 

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-1953

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1568)

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -38,8 +38,7 @@ DATASET_ID_LABEL = u'flocker-dataset-id'
 METADATA_VERSION_LABEL = u'flocker-metadata-version'
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
-BOTO_NUM_RETRIES = u'5'
-# BOTO_NUM_RETRIES = u'20'
+BOTO_NUM_RETRIES = u'20'
 VOLUME_STATE_CHANGE_TIMEOUT = 300
 MAX_RETRIES = 3
 
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -13,7 +13,7 @@ from uuid import UUID
 
 from bitmath import Byte, GiB
 
-from pyrsistent import PRecord, field
+from pyrsistent import PRecord, field, pset
 from zope.interface import implementer
 from boto import ec2
 from boto import config
@@ -38,8 +38,10 @@ DATASET_ID_LABEL = u'flocker-dataset-id'
 METADATA_VERSION_LABEL = u'flocker-metadata-version'
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
-BOTO_NUM_RETRIES = u'20'
+BOTO_NUM_RETRIES = u'5'
+# BOTO_NUM_RETRIES = u'20'
 VOLUME_STATE_CHANGE_TIMEOUT = 300
+MAX_RETRIES = 3
 
 
 class EliotLogHandler(logging.Handler):
@@ -419,7 +421,7 @@ class EBSBlockDeviceAPI(object):
                 return volume
         raise UnknownVolume(blockdevice_id)
 
-    def _next_device(self, instance_id):
+    def _next_device(self, instance_id, ignore_devices):
         """
         Get the next available EBS device name for a given EC2 instance.
         Algorithm:
@@ -440,8 +442,9 @@ class EBSBlockDeviceAPI(object):
             instance are currently occupied.
         """
         volumes = self.connection.get_all_volumes()
-        devices = [v.attach_data.device for v in volumes
-                   if v.attach_data.instance_id == instance_id]
+        devices = pset({v.attach_data.device for v in volumes
+                       if v.attach_data.instance_id == instance_id})
+        devices.add(ignore_devices)
         IN_USE_DEVICES(devices=devices).write()
 
         for suffix in b"fghijklmonp":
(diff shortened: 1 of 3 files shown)


flocker/node/agents/ebs.py
```diff
@@ -511,30 +515,49 @@ def attach_volume(self, blockdevice_id, attach_to):
...
+                    # picked (http://docs.aws.amazon.com/AWSEC2/latest/
+                    # UserGuide/device_naming.html), wait for new block device
+                    # to be available to the OS, and interpret it as ours.
+                    # Wait under lock scope to reduce false positives.
+                    new_device = _wait_for_new_device(blockdevices,
+                                                      volume.size)
+                    attached = True
+
+                    # end lock scope
+            except EC2ResponseError as e:
```
wallrj: This Exception can only come from the `self.connection.attach_volume` call, so move the `try: except` around that call.

wallrj:
Has anyone got any idea what's causing this?

I searched zulip and elsewhere but couldn't find anything.

The strategy in master seems to be:

```
def attach_volume:
    # Acquire lock (there can be no further attach_volume until this one
    # finishes)
    lock

    # Look for device names that are used by locally attached volumes. Compare
    # with a set of all allowed device names and return the first unused name.
    next_free_api_device = find_next_available_ebs_device()

    # Attempt to attach
    boto.attach(volume, next_free_api_device)

    # wait for *any* new device name to appear matching `/sys/block/sd?` or
    # `/sys/block/xvd?
    actual_device = wait_for_new_device(expected_size)

    # store the actual device in the ebs volume metadata.
    update_metadata

    # Other attachments can proceed.
    lock_release
```

So how does the error occur?

 * Is there more than one EBSBlockDeviceAPI being tested on the build slave?
   I thought that a build slave could only run one build at a time.

 * Is some other process attempting to attach EBS volumes to the build slave? 

 * Does the boto cache the VolumeAttachment information? 
   http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_VolumeAttachment.html
   In which case we might not be getting the complete list of devices from the API   


And generally,

 * The `try: except` should be kept close to the `boto.attach` call.

 * it seems to me that we only need to retry first two steps.
   ```
    next_free_api_device = find_next_available_ebs_device()

    # Attempt to attach
    boto.attach(volume, next_free_api_device)
   ```

 * or we could just log the failure and allow the agent to do the retrying...it
   would eventually succeed, right?

 * Any ideas how we might test this? 
   I guess this is the same discussion as we had in https://github.com/ClusterHQ/flocker/pull/1555#issuecomment-110686600
   And the answer is that it's probably too much work at this time.

Please resubmit for a final review after discussing or addressing the points above.

myechuri:
Apologies for not listing the problem and testing details.

The sequence of events leading to the bug are:
- Test attempts to attach volume1
- Our driver (``ebs.py``) finds empty device (say ``/dev/sbf``) to attach to.
- AWS attach call for volume1 completes transitioning from ``available`` to ``in-use``.
- We (``ebs.py``) decide volume is attached, declare success.

- Test attempts to attach volume2
- Our driver (``ebs.py``) asks AWS for all volumes and their attachment info.
- Volume1's attachment info has not been updated with ``/dev/sdf`` yet.
- ``ebs.py`` decides ``/dev/sdf`` is available and decides to attach volume2 to it.
- AWS ``attach_volume`` throws exception indicating ``/dev/sdf`` is already in-use, as far as OS is concerned, which is true.

_____________

How i reproduced:

2 ways:

1) In ``test_multiple_volumes_attached_to_host``, i increased number of volumes being attached to 10. And in ``ebs.py``, i reduced number of boto retries to 5. This was the right combination to reproduce the bug 100% of runs. 

2) After (1) succeeded in reproducing, i did the following change in ``ebs.py``:
while looking for an available free device in ``_next_device``, i changed:
```
     for suffix in b"fghijklmonp":
            if file_name not in devices:
                return file_name
```
to
```
     for suffix in b"fghijklmonp":
            return file_name
```

This guaranteed picking a used device. This also helped in verifying the fix, by stepping through the retry iterations.

i decided not to include the test changes in (1) because they are more of stress tests than functional.

flocker/node/agents/ebs.py
```diff
@@ -40,6 +40,7 @@
 ATTACHED_DEVICE_LABEL = u'attached-device-name'
 BOTO_NUM_RETRIES = u'20'
 VOLUME_STATE_CHANGE_TIMEOUT = 300
+MAX_RETRIES = 3
```
itamarst: This name is ambiguous. Max retries of what?

flocker/node/agents/ebs.py
```diff
@@ -419,7 +420,7 @@ def _get_ebs_volume(self, blockdevice_id):
                 return volume
         raise UnknownVolume(blockdevice_id)
 
-    def _next_device(self, instance_id):
+    def _next_device(self, instance_id, devices_in_use):
         """
         Get the next available EBS device name for a given EC2 instance.
```
itamarst: Document devices_in_use.

flocker/node/agents/ebs.py
```diff
@@ -440,8 +441,11 @@ def _next_device(self, instance_id):
             instance are currently occupied.
         """
         volumes = self.connection.get_all_volumes()
-        devices = [v.attach_data.device for v in volumes
-                   if v.attach_data.instance_id == instance_id]
+        devices = pset({v.attach_data.device for v in volumes
+                       if v.attach_data.instance_id == instance_id})
+        devices = devices.add(devices_in_use)
```
itamarst: This is wrong - you're adding the set as an *item*. So now you have `pset(["/dev/sdf", pset(["/dev/sdg"])])`.

You want `devices = devices | devices_in_use`.

Given the logic is broken I think this merits more thought on testing, I'll try to address that later in review.

flocker/node/agents/ebs.py
```diff
@@ -511,30 +515,50 @@ def attach_volume(self, blockdevice_id, attach_to):
...
-            # and interpret it as ours.
-            # Wait under lock scope to reduce false positives.
-            new_device = _wait_for_new_device(blockdevices, volume.size)
-
-            # end lock scope
+        attached = False
+        ignore_devices = pset([])
+        attach_attempts = 0
+        while (not attached and attach_attempts < MAX_RETRIES):
+            with self.lock:
```
itamarst: Move this block into a utility method (self._something).

flocker/node/agents/ebs.py
```diff
@@ -511,30 +515,50 @@ def attach_volume(self, blockdevice_id, attach_to):
...
+
+                try:
+                    self.connection.attach_volume(blockdevice_id,
+                                                  attach_to,
+                                                  device)
+                except EC2ResponseError as e:
+                    # If attach failed because selected device is currently
+                    # in use, retry MAX_RETRIES times to find another
+                    # unused device.
+                    used = u"Attachment point {0} is already in use".format(
```
itamarst: This is brittle. Instead I suggest just retrying no matter what the message - worst case the convergence loop will try again if there's a real problem.

flocker/node/agents/ebs.py
itamarst: I.e. the logic should be "do MAX_RETRIES attempts at calling this method".

itamarst:
We now have two strategies for finding devices: trying unused, and skipping failed ones. In an ideal world we would only have  one strategy, and the appealing one is the latter one: keep trying until you find one that succeeds and just forget about asking what's in-use. This would also make the test fail unconditionally lacking the fix. Unfortunately that may interact badly with AWS rate limits, at least in our test setups.

If we do stick to current approach, I'm starting to think we need simulated AWS where we can inject failures or eventual consistency. No time for that now. So.... (thinking).

itamarst:
Ah! Separate out code into two parts: first, split out the pure in-memory utility for dealing with data structure - I have a set and another set and a list of devices and I extract right one. Write a test for that - it will catch current logic bug you introduced. The second part hooks up that pure logic to boto calls.

Address, then merge (but note discussion on zulip about changed development process, may need to do another PR against reelease branch or something.)

flocker/node/agents/ebs.py
itamarst: Eh, probably not worth the bother now.

adamtheturtle:
> may need to do another PR against reelease branch or something.

I'll do this now

PR #1574, opened 2015-06-11
[FLOC-2209] Move the CentOS and Ubuntu instructions (above the Alpha and Testing sections)

davidoliverSP2:
CentOS and Ubuntu instructions moved so users don't miss essential instructions (these were previously below sections labelled as Alpha & Testing.

Note - while not part of this PR, we shouldn't be using Alpha in the docs. Issue raised: FLOC-2243

### DIFF
--- a/docs/using/installing/index.rst
+++ b/docs/using/installing/index.rst
@@ -619,8 +619,6 @@ Run the following commands to enable the agent service:
 .. task:: enable_flocker_agent ubuntu-14.04 ${CONTROL_NODE}
    :prompt: [root@agent-node]#
 
-.. _zfs-dataset-backend:
-
 What to do next
 ===============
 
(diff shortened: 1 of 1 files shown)


docs/using/installing/index.rst
```diff
@@ -557,6 +557,22 @@ The configuration item to use AWS should look like:
...
+   :prompt: [root@agent-node]#
+
+Ubuntu
+......
+
+Run the following commands to enable the agent service:
+
+.. task:: enable_flocker_agent ubuntu-14.04 ${CONTROL_NODE}
+   :prompt: [root@agent-node]#
+
```
exarkun: This is an improvement but this placement still doesn't quite make sense to me.  These things are still at the same level as the storage driver configuration documentation.  It seems like they belong in a different section or that the individual storage driver sections should be demoted to an inferior heading level.  I was also surprised to **return** to Flocker service configuration instructions after leaving the earlier Flocker installation section and spending time reading about `flocker-ca`.  I see why things happen in this order but I don't think that makes it less confusing to the user. :/

This whole document (and probably the activities it **documents**) probably needs to be revisited.  But for now maybe a change that makes sense is to:

  * Create a "Configuring the Flocker agent" section and put all of the agent.yml documentation beneath it
  * Create an "Enabling the Flocker agent service" section and put the `systemctl ...` etc stuff beneath that.

exarkun:
Thanks.  Please address that one point and then merge.

PR #1575, opened 2015-06-11
[FLOC-2242] Release maintenance/flocker 0.9.0pre1/fix content type

jongiddy:
Merge Vagrant metadata file fix from FLOC-2191 into 0.9.0pre1


adamtheturtle:
This should be a PR against the prerelease

PR #1576, opened 2015-06-11
[FLOC-2242] Release maintenance/flocker 0.9.0pre1/fix content type

jongiddy:
PR against 0.9.0pre1 of Vagrant fix from FLOC-2191


adamtheturtle:
Thanks. In future if you do something a bit odd (not the process described in the maintenance docs) it would be good to post exactly what was done. Then the reviewer can spot errors in the process instead of re-reviewing the code. Merging.

PR #1577, opened 2015-06-11
[FLOC-2221] Swap the security topics in the Using and Reference sections

davidoliverSP2:
The security topics were in the wrong sections - the concepts were in the the Using section, and the instructions were in the reference section. This PR moves them into the correct place, and changes some of the lead in text at the bottom of the Installing doc.


docs/concepts/security.rst
```diff
@@ -13,7 +13,7 @@ This ensures that the control service, convergence agents, and API end users are
 Mutual Authentication Overview
 ==============================
 
-When :doc:`installing the flocker-node package <installing/index>`, the ``flocker-ca`` tool (provided as part of ``flocker-cli``) generates a root certificate for your cluster, comprising a private key and public certificate, as well as certificates and private keys for the control service and convergence agents.
+When :ref:`installing the flocker-node package <installflocker>`, the ``flocker-ca`` tool (provided as part of ``flocker-cli``) generates a root certificate for your cluster, comprising a private key and public certificate, as well as certificates and private keys for the control service and convergence agents.
```
adamtheturtle: :question: Can we use `install-flocker` instead (like we have `generate-api` below). I find it much easier to read.

docs/using/authentication.rst
```diff
@@ -16,12 +16,14 @@ If you have not already followed these steps, see the :ref:`flocker-node install
...
-================================
+Generating an API user certificate
+==================================
 
 The CLI package includes the ``flocker-ca`` program which is used to generate certificate and key files.
 
-You can run ``flocker-ca --help`` for a full list of available commands. For API user certificates, run the ``flocker-ca create-api-certificate`` command from the same directory containing the certificate authority files generated when you first :ref:`installed the cluster <authentication>`.
+.. note:: You can run ``flocker-ca --help`` for a full list of available commands.
+
+For API user certificates, run the ``flocker-ca create-api-certificate`` command from the same directory containing the certificate authority files generated when you first :ref:`installed the cluster <authentication>`.
```
adamtheturtle: "from the same directory containing the certificate authority files" > "from the directory which contains" or something similar.

docs/using/installing/index.rst
```diff
@@ -677,6 +677,6 @@ So ensure that the firewall allows access to TCP port 22 on each node from the e
 Next Steps
 ----------
 
-Next, we will describe how to use cluster security and authentication.
+Next, we will describe how to set up an :ref:`authenticated user<authenticate>`.
```
adamtheturtle: :question: I'd prefer "The next section describes" or similar.

adamtheturtle:
Thanks @davidoliverSP2 . I have made a couple of points but from an engineering view I think that this does no harm. I got the impression from @robhaswell that he was happy to review these kinds of things for usability so I'll assign it to him.

@robhaswell unassign if that isn't what you'd like.

docs/concepts/security.rst
davidoliverSP2: thanks @adamtheturtle - I'm reluctant to make changes to ref tags at the moment due to the risk of breaking links in other docs. I've raised FLOC 2262 to revisit this across the whole doc set post v1.

docs/using/installing/index.rst
```diff
@@ -677,6 +677,6 @@ So ensure that the firewall allows access to TCP port 22 on each node from the e
 Next Steps
 ----------
 
-Next, we will describe how to use cluster security and authentication.
+The next section describes how to set up an :ref:`authenticated user<authenticate>`.
 However, you may want to perform the steps in :ref:`the MongoDB tutorial <movingapps>` to ensure that your nodes are correctly configured.
```
robhaswell: This needs to be removed and placed at the bottom of the next page. It is impossible to proceed with the MongoDB tutorial without creating an authenticated user.

Put another way, at this point creating a user is the ONLY thing a user can do, and they MUST do it.

robhaswell:
This is going to have a gigantic merge conflict with `addinst-genAPIcert-FLOC-2220`.

robhaswell:
Almost there, please correct and re-submit. I'd still like to check that it makes sense. Thanks!

robhaswell:
Great! Thanks.

docs/concepts/security.rst
adamtheturtle: There is no risk - Sphinx will tell you when you build the docs.

PR #1578, opened 2015-06-11
[FLOC-2243] Remove Alpha and replace with Experimental

davidoliverSP2:
Only one occurrence in the docs on master (installation doc). This is changed.

Definition of Stable and Experimental will be covered on FLOC 2045


exarkun:
Thanks.  Does what it says.  Please merge.

PR #1579, opened 2015-06-11
[FLOC-2244] Require zone when configuring AWS acceptance instances.

itamarst:
This has a corresponding config change to Buildbot (already in), so Buildbot will need to be restarted before the AWS acceptance tests will pass. Until this is merged *other* branches will fail to run AWS acceptance tests altogether, but its' not like they were passing before...

I do have a locally passing AWS acceptance test run, though, using this branch.


tomprince:
Looks good. Merge after restarting buildbot.

PR #1580, opened 2015-06-11
[FLOC-1884] Fix conch tests 

jml:
In lieu of actually fixing [the actual bug in conch](https://twistedmatrix.com/trac/ticket/7936), this changes the `ISession` implementation that we use so that it sets `HOME` to the avatar's configured home directory.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1580)

### DIFF
--- a/flocker/testtools/ssh.py
+++ b/flocker/testtools/ssh.py
@@ -13,14 +13,19 @@ from zope.interface import implementer
 
 from ipaddr import IPAddress
 
+from twisted.python.components import registerAdapter
 from twisted.internet import reactor
 from twisted.cred.portal import IRealm, Portal
 
 try:
     from twisted.conch.ssh.keys import Key
     from twisted.conch.checkers import SSHPublicKeyDatabase
+    from twisted.conch.interfaces import ISession
     from twisted.conch.openssh_compat.factory import OpenSSHFactory
-    from twisted.conch.unix import UnixConchUser
+    from twisted.conch.unix import (
+        SSHSessionForUnixConchUser,
+        UnixConchUser,
+    )
     _have_conch = True
 except ImportError:
     SSHPublicKeyDatabase = UnixConchUser = object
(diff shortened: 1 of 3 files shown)


tomprince:
This looks good if [buildbot](http://build.clusterhq.com/boxes-flocker?branch=refs/pull/1580/head) is happy (in particular the dev-box, which *always* seems to trigger this) then merge.

---

Review status: 0 of 3 files reviewed, 1 unresolved discussion, all commit checks successful.

---

<sup>**[flocker/testtools/ssh.py, line 92 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1580#-JrYbg7cYRGrGOA3C7jz)** ([raw file](https://github.com/clusterhq/flocker/blob/7f86a8de11af9748252fd8e540eab740a57ad353/flocker/testtools/ssh.py#L92)):</sup>
`e`

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1580)

tomprince:
Review status: all files reviewed, 1 unresolved discussion, all commit checks successful.
Reviewed files:
- admin/test/test_runner.py @ r1
- flocker/provision/test/test_ssh_conch.py @ r1
- flocker/testtools/ssh.py @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1580)

flocker/testtools/ssh.py
```diff
@@ -81,6 +86,23 @@ def getUserGroupId(self):
         return None, None
 
 
+@implementer(ISession)
+class _EnvironmentSSHSessionForUnixConchUser(SSHSessionForUnixConchUser):
+    """
+    SSH Session that correctly sets HOMe.
```
adamtheturtle: "HOMe" > "HOME"

adamtheturtle:
:question: With the one other patch we have for conch, there is a test which will fail when the patch can be removed - that is, it will fail when the behaviour is fixed upstream.

jml:
@adamtheturtle You're suggesting I add such a test here? If so, I think I'd rather not do it this time. Figuring out how to do that is roughly the same amount of work as writing the upstream patch, and I'd rather just do that.

@tomprince Buildbot is not happy, but is less sad. Tests now fail when run as root for entirely different reasons. The failures weren't visible before, perhaps because tox-root doesn't run if tox fails?

---

Review status: all files reviewed, 2 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/testtools/ssh.py, line 92 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1580#-JrY_REJPqrR3ItyLnJY-r1-92)** ([raw file](https://github.com/clusterhq/flocker/blob/7f86a8de11af9748252fd8e540eab740a57ad353/flocker/testtools/ssh.py#L92)):</sup>
Done.

---

<sup>**[flocker/testtools/ssh.py, line 92 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1580#-JrYbg7cYRGrGOA3C7jz)** ([raw file](https://github.com/clusterhq/flocker/blob/7f86a8de11af9748252fd8e540eab740a57ad353/flocker/testtools/ssh.py#L92)):</sup>
I've changed HOMe to HOME. Is this what you meant?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1580)

adamtheturtle:
I was suggesting that you make the judgement of whether it is worthwhile or
not.

PR #1581, opened 2015-06-11
[FLOC-2249] Release base

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2249

### DIFF
--- a/admin/release.py
+++ b/admin/release.py
@@ -113,13 +113,6 @@ class BranchExists(Exception):
     """
 
 
-class BaseBranchDoesNotExist(Exception):
-    """
-    Raised if trying to release a version for which the expected base branch
-    does not exist.
-    """
-
-
 class MissingPreRelease(Exception):
     """
     Raised if trying to release a pre-release for which the previous expected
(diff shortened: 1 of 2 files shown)


admin/release.py
```diff
@@ -926,10 +916,6 @@ def create_release_branch_main(args, base_path, top_level):
         sys.stderr.write("%s: No (previous) pre-release exists for this "
                          "release.\n" % (base_path.basename(),))
         raise SystemExit(1)
-    except BaseBranchDoesNotExist:
```
adamtheturtle: I think that this pattern for release tools causes problems. It might be suitable for user-facing tools.

admin/release.py
```diff
@@ -852,14 +845,11 @@ def calculate_base_branch(version, path):
     # We create a new branch from a branch, not a tag, because a maintenance
     # or documentation change may have been applied to the branch and not the
     # tag.
-    try:
-        base_branch = [
-            branch for branch in repo.branches if
-            branch.name == base_branch_name][0]
-    except IndexError:
-        raise BaseBranchDoesNotExist()
+    repo.git.checkout(base_branch_name)
```
jongiddy: Comment that this is required to make the branch available "locally" for the next step.

jongiddy:
OK to merge.

PR #1584, opened 2015-06-11
FLOC-2258: Create flocker-zfs meta package

ryao:
https://clusterhq.atlassian.net/browse/FLOC-2258

We need this so that we can pin ZFS versions to things we test.

Tom suggested the repository location, Conflicts statements and
BuildArch statement.

Signed-off-by: Richard Yao <richard.yao@clusterhq.com>


itamarst:
I believe we decided not to do this in the end. Re-open if that changes.

PR #1585, opened 2015-06-11
Output OpenStack OverLimit errors to Eliot.

myechuri:


flocker/node/agents/cinder.py
```diff
@@ -331,6 +347,7 @@ def compute_instance_id(self):
         """
         local_ips = get_all_ips()
         api_ip_map = {}
+        return u'cf3824fb-0342-4e3f-9dc6-b5d0bb4a97c3'
```
exarkun: :grin:

flocker/node/agents/cinder.py
```diff
@@ -356,17 +373,14 @@ def create_volume(self, dataset_id, size):
             CLUSTER_ID_LABEL: unicode(self.cluster_id),
             DATASET_ID_LABEL: unicode(dataset_id),
         }
-        action_type = u"blockdevice:cinder:create_volume"
-        with start_action(action_type=action_type):
-            requested_volume = self.cinder_volume_manager.create(
-                size=int(Byte(size).to_GiB().value),
-                metadata=metadata,
-            )
-            Message.new(blockdevice_id=requested_volume.id).write()
```
exarkun: Removing the `start_action` makes sense because there is now automatic logging applied by the logging decorator.  But the logging decorator doesn't log the resulting `blockdevice_id`, I think.  It would be nice to keep that in the logs to be able to easily see where various Cinder volumes came from.

Simply leaving the `Message...` line in may be good enough for this.  It will get logged in the context of some wider Eliot action, I think.

Unless you think that information is available somewhere else already?

flocker/node/agents/cinder.py
```diff
@@ -515,9 +529,18 @@ def _blockdevicevolume_from_cinder_volume(cinder_volume):
...
+
+@auto_openstack_logging(INovaVolumeManager, "_nova_volumes")
+class _LoggingNovaVolumeManager(PRecord):
+    _nova_volumes = field(mandatory=True)
+
+@auto_openstack_logging(INovaServerManager, "_nova_servers")
+class _LoggingNovaServerManager(PRecord):
+    _nova_servers = field(mandatory=True)
+
 def cinder_api(cinder_client, nova_client, cluster_id):
```
exarkun: Maybe it's just the late hour ... but ... I can't understand how this code could possibly have worked previously.  `auto_openstack_logging` is a class decorator but it was being used as a function decorator?  And a string with `.` in it was being used as an attribute name?

Was there any OpenStack logging enable at all prior to this change?

exarkun: Looks like `auto_openstack_logging` (or anything defined with `interface_decorator`) silently accepts functions and does nonsense with them - but in a way that doesn't immediately obviously break things. :/  Probably worth filing an issue for making that fail in an obvious way instead.

flocker/node/agents/cinder.py
```diff
@@ -67,30 +71,42 @@ def _run_with_logging(self, *args, **kwargs):
...
+            with OPENSTACK_ACTION(operation=[attempt, method_name, args, kwargs]):
+                try:
+                    return method(*args, **kwargs)
+                except KeystoneOverLimit as e:
+                    retry_after = int(e.retry_after)
+                    time.sleep(retry_after)
+                    retry = True
+                    attempt += 1
+                    if attempt == MAX_ATTEMPTS:
+                        raise
```
exarkun: I'm fairly tempted to suggest this retry logic belongs in a separate decorator, not in the logging decorator.  Maybe that refactoring isn't the highest priority task now - on the other hand, the retry logic should really have unit tests.  This code currently doesn't have direct tests for the retry behavior and it will be a lot easier to write them if you can exercise the retry code without all of this OpenStack baggage attached to it.

So .. yea.  I think it would be better to leave the logging decorator alone and create a new `interface_decorator` for this automatic sleep-and-retry behavior.  Parameterize enough of the pieces (apart from the method, I'm thinking about the exception type which prompts a retry) so that you can write some generic unit tests that don't interact with OpenStack at all.  Then you can apply that decorator in the same place `auto_openstack_logging` is being applied.

I'd also suggest adding `KeystoneOverLimit` to the logging decorator and applying the retry decorator "outside of" ("on top of"? eh) the logging decorator so that each over-limit exception that might prompt a retry also gets logged.

exarkun:
Thanks!  I think I probably should have explained/documented the `interface_decorator` stuff better and my failure to do that probably led you a bit down the wrong path (either here or on FLOC-2027).  Sorry about that.

Please do the refactoring and add the unit tests I suggested in my inline comment and then re-submit.

flocker/node/agents/cinder.py
myechuri: Sorry, should have added a note to ignore this change. Couldn't reproduce/test over limit issue (due to FLOC-2250) without hardcoding my instance.

flocker/node/agents/cinder.py
myechuri: You pointed out a very valid reason to keep ``Message``, will bring it back.
``start_action`` is no longer necessary since logging decorator has ``with OPENSTACK_ACTION..``.

flocker/node/agents/cinder.py
myechuri: > Maybe it's just the late hour ... but ... I can't understand how this code could possibly have worked previously.

It didn't work. i couldn't find OpenStack logs for ``retry_after`` in latest builder runs (that ran into rate over limit), debugged local run and realized the decorator method was not getting called at all.

> Probably worth filing an issue for making that fail in an obvious way instead.

Filed FLOC-2259 and assigned to self.

flocker/node/agents/cinder.py
myechuri: Thanks! Unit test is a good idea! This is how i reproduced / tested fix:

```
    def test_overlimit(self):
        """
        ``attach_volume`` can attach multiple block devices to a single host.
        """

        self.assertRaises(KeystoneOverLimit, self._generate_overlimit)

    def _generate_overlimit(self):
        i = 0
        while i < 50:
            self.api.create_volume(
                dataset_id=uuid4(),
                size=self.minimum_allocatable_size
            )
            i += 1
```

Did not include it in the review because it doesn't belong in functional test suite.

Unit test is a better idea, will move to that.

i also wanted review comments on using ``retry_after`` primarily versus max(retry_after, exponential back off generated value). The few reproductions i saw, ``retry_after`` has always been ``0``. Hence, i do not completely trust relying solely on that. What do you think, @exarkun ?

itamarst:
After some discussion we decided only logging is needed, thus this is the last PR standing.

itamarst:
I'm going to close this and recreate PR off of what was merged into release branch.

PR #1598, opened 2015-06-12
[FLOC-2240] Don't restart for dataset size changes

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2240

Note this is against master because a dependency hasn't been merged into 1.0.0pre1 yet.  I'll do 1.0.0pre1 after that dependency is merged but doing things in the right order now would prevent this from going up for review.

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -850,6 +850,7 @@ class ApplicationNodeDeployer(object):
         """
         def log(diverged, reason=None):
             Message.new(
+                message_type=_eliot_system(u"volume_is_diverged"),
                 volume_is_diverged=diverged,
                 state_is_none=state is None,
                 configuration_is_none=configuration is None,
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -849,15 +849,15 @@ class ApplicationNodeDeployer(object):
             one), ``True``.  If it does not differ or only differs in the
             allowed ways mentioned above, ``False``.
         """
-        def log(diverged, reason=None):
+        def log(restart, reason=None):
             Message.new(
                 message_type=_eliot_system(u"restart_for_volume_change"),
-                volume_is_diverged=diverged,
+                restart=restart,
                 state_is_none=state is None,
                 configuration_is_none=configuration is None,
                 reason=reason,
             ).write()
-            return diverged
+            return restart
 
         def restart_if_available(dataset_id):
             """
@@ -876,10 +876,10 @@ class ApplicationNodeDeployer(object):
             """
             if dataset_id in node_state.manifestations:
                 # We want it and we have it.
-                return log(True)
+                return log(True, "have configured dataset")
             else:
                 # We want it but we don't have it.
-                return log(False)
+                return log(False, "missing configured dataset")
 
         state_id = getattr(
             getattr(state, "manifestation", None), "dataset_id", None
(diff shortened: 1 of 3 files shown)


flocker/node/_deploy.py
```diff
@@ -816,45 +816,125 @@ def discover_state(self, local_state):
...
+
+        :return: If the state differs from the configuration in a way which
+            needs to be corrected by the convergence agent (for example, the
+            application is configured with a volume but is running without
+            one), ``True``.  If it does not differ or only differs in the
+            allowed ways mentioned above, ``False``.
+        """
+        def log(diverged, reason=None):
+            Message.new(
+                volume_is_diverged=diverged,
```
itamarst: `message_type` or eliot tree will be sad.

flocker/node/_deploy.py
```diff
@@ -816,45 +816,125 @@ def discover_state(self, local_state):
...
+            manifestation exists.
+
+            :param unicode dataset_id: The identifier of the dataset we want.
+
+            :return: If there is a manifestation of the given dataset on this
+                node, ``True``.  Otherwise, ``False``.
+            """
+            if dataset_id in node_state.manifestations:
+                # We want it and we have it.
+                return log(True)
```
itamarst: And this logging is going to be confusing to read.

itamarst: Ah, I see, never mind.

flocker/node/_deploy.py
```diff
@@ -816,45 +816,125 @@ def discover_state(self, local_state):
...
+            else:
+                # We want it but we don't have it.
```
itamarst: *This* logging is going to be super-confusing. "Why is it saying it didn't diverge when it very definitely is different?"

flocker/node/_deploy.py
```diff
@@ -816,45 +816,125 @@ def discover_state(self, local_state):
...
+        def log(diverged, reason=None):
+            Message.new(
+                volume_is_diverged=diverged,
+                state_is_none=state is None,
+                configuration_is_none=configuration is None,
+                reason=reason,
+            ).write()
+            return diverged
+
+        def want(dataset_id):
```
itamarst: This seems like the wrong place for this logic. Or at least you should rename the method? "_should_restart_due_to_mismatched_volume" or something.

itamarst: Also I wouldn't call this function `want`. `volume_available` or something?

flocker/node/_deploy.py
```diff
@@ -816,45 +816,125 @@ def discover_state(self, local_state):
...
 
         Certain differences are not considered divergences:
 
             - The running state of the application.  It may have exited
               normally and correctly after completing its task.
 
-            - The metadata for any volume of the application.  Volume metadata
-              only exists in the configuration.  It is always missing from the
-              state object.
+            - Certain volume differences.  See ``_volume_is_diverged``.
```
itamarst: Document new parameter.

flocker/node/test/test_deploy.py
```diff
@@ -1149,6 +1167,162 @@ def test_has_volume_no_changes(self):
...
+        local_config = to_node(local_state).set(
+            applications=[
+                changed_application,
+            ],
+        )
+        assert_application_calculated_changes(
+            self, local_state, local_config, set(),
+            restart(application, changed_application, local_state),
+        )
+
```
itamarst: None of these tests actually cover the original reported problem: configuration specifies no size, actual dataset in state has one.

itamarst:
OK, address above issues, then merge this and the companion PR.

PR #1602, opened 2015-06-12
[FLOC-2283] Use a realistic blocksize in test_application_volume_quotas.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2283

### DIFF
--- a/flocker/acceptance/test_deployment.py
+++ b/flocker/acceptance/test_deployment.py
@@ -20,7 +20,9 @@ from .testtools import (MONGO_APPLICATION, MONGO_IMAGE,
                         create_attached_volume, require_cluster,
                         require_moving_backend)
 
-SIZE_100_MB = u"104857600"
+from ..testtools import (
+    REALISTIC_BLOCKDEVICE_SIZE,
+)
 
 
 def api_configuration_to_flocker_deploy_configuration(api_configuration):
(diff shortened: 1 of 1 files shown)


itamarst:
I don't think this is the right solution. A realistic block device size in one backend may not in another (e.g. EMC XtremeIO requires stuff in multiples of 8 GiB). The whole presumption that dataset size will be exactly the same as requested size is wrong.

tomprince:
I think this is the expedient solution, pre-1.0. Doing the proper solution would require refactoring `wait_for_dataset` to be able to ignore some properties of the dataset. I think this is valuable, but disruptive enough that I don't want to do in pre-1.0.

exarkun:
> A realistic block device size in one backend may not in another (e.g. EMC XtremeIO requires stuff in multiples of 8 GiB).

I think this is a criticism of the entire idea of `REALISTIC_BLOCKDEVICE_SIZE` and I think it's a valid criticism.  We took a small step away from it in `flocker/node/agents/test/test_blockdevice.py` by introducing per-TestCase `minimum_allocatable_size` and `device_allocation_unit` parameters.  Unfortunately the acceptance tests aren't factored in a way that makes it easy to replicate that fix here (in many ways, from how the TestCase's themselves are defined to how a backend is chosen to run the tests against).  Also, it probably doesn't make sense to hold this code to a standard that prohibits `REALISTIC_BLOCKDEVICE_SIZE` while we let the rest of the acceptance tests continue to use it.

We'll come back and fix this but I think replacing the custom hard-coded size with the standard, test-suite-wide hard-coded size makes sense as an incremental step in the right direction.

exarkun:
So, assuming the test passes on buildbot, please merge.  Thanks.

itamarst:
GREEEEEEEEEN

tomprince:
Gah. This was a PR against the wrong branch (but the tests weren't).

PR #1603, opened 2015-06-12
[FLOC-2285] Postgres tests improvements

itamarst:
PR into master will be merged as soon as tests are slightly happier. Actual passing tests will be in a stacked PR.

PR #1604, opened 2015-06-12
[FLOC-2289] Don't use world-writeable permissions if dataset has been modified with files

itamarst:
Depends on #1603.

In theory this should fix postgres acceptance tests on AWS.


itamarst:
Wrong branch.

PR #1605, opened 2015-06-12
[FLOC-2285] Postgres tests improvements

itamarst:
Already merged into master. Don't actually expect these tests to pass - that will be in follow up PR #1606.


wallrj:
Looks fine and I verified that this helps the postgres acceptance tests pass on Rackspace and AWS in #1606.

Please merge.

PR #1626, opened 2015-06-15
[FLOC-1968] License

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1968

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2014-2015 ClusterHQ
+   Copyright 2014-2015 ClusterHQ Inc.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2014, 2015 ClusterHQ
+   Copyright 2014-2015 ClusterHQ
 
    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2014, 2015 ClusterHQ
+   Copyright 2014-2015 ClusterHQ
```
dwgebler: Seeing as this is a copyright notice, I suspect this should be "ClusterHQ Inc", seeing as that's the actual entity name that owns Flocker.

dwgebler:
There are some flake8 errors in `admin/test_release` - I'm not sure that matters, we have quite a few flake8 errors in our various admin code and we don't care about that in the same way we do for Flocker, but still, nice to fix them in any file you happen to be working on.

Other than the one thing about whether or not to use the full company name in the copyright line (my guess is we should), this looks good and tests fine so I will accept the JIRA issue.

adamtheturtle:
> There are some flake8 errors in admin/test_release - I'm not sure that matters, we have quite a few flake8 errors in our various admin code and we don't care about that in the same way we do for Flocker, but still, nice to fix them in any file you happen to be working on.

We do care and there is a follow up to fix it.

exarkun:
@adamtheturtle @dwgebler 

This change guarantees a failing test suite once a year.  It also uses local time so as to produce different results depending on where it is run. :/

dwgebler:
> This change guarantees a failing test suite once a year.

I thought that was the intention, that we therefore cannot miss when we need to update the file.

Local time I did notice was an issue, but seeing as we're only using the year and probably not ever doing a release in the 12 hour period on New Year's Eve where some parts of the world might consider it a different year to other parts, I didn't think that was worth doing anything about.

PR #1629, opened 2015-06-15
Revert "[FLOC-1968] License"

exarkun:
Reverts ClusterHQ/flocker#1626

Doesn't resolve the issue in a desirable way.


dwgebler:
So instead we'll.......check the copyright date and modify it as part of the release process script? And not have a test in admin that checks it? If that's yes and yes, please go ahead and merge the revert.

adamtheturtle:
Merging and then we can discuss options - doesn't matter what the conclusion is.

PR #1630, opened 2015-06-15
mkdir /etc/flocker on every node.

robhaswell:
https://clusterhq.atlassian.net/browse/FLOC-2319

### DIFF
--- a/docs/using/installing/index.rst
+++ b/docs/using/installing/index.rst
@@ -368,11 +368,11 @@ The Flocker CLI package includes the ``flocker-ca`` tool that is used to generat
 
         $ flocker-ca create-control-certificate example.org
 
-#. At this point you will need to create a :file:`/etc/flocker` directory:
+#. At this point you will need to create a :file:`/etc/flocker` directory on each node:
 
-   .. prompt:: bash
+   .. code-block:: console
 
-      mkdir /etc/flocker
+      root@centos-7:~/$ mkdir /etc/flocker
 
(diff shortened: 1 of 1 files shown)


robhaswell:
I don't know why these build built against the wrong branch, but this is the correct build: http://build.clusterhq.com/boxes-flocker?branch=release-maintenance/flocker-1.0.0/etc-flocker-everywhere-FLOC-2320

wallrj:
Ok. I guess this makes it a little clearer.

Personally I'd probably move all the certificate and key copying into a single code block and demonstrate commands that you might use to upload the certificates. eg

```
$ ssh root@node1 -- mkdir /etc/flocker
$ scp control-example.org.crt root@node1:/etc/flocker/control-service.crt
$ scp control-example.org.key root@node1:/etc/flocker/control-service.key
$ scp cluster.crt root@node1:/etc/flocker/
$ ssh root@node1 chmod 0700 /etc/flocker
$ ssh root@node1 chmod 0600 /etc/flocker/control-service.key
```

It assumes that the user has root SSH access to the box. 
But if they haven't at least this gives a clear idea of what how they are expected to transfer the certificates and keys.

Anyway, it can be improved (or better still automated) in a follow up issue.

Please merge unless you feel like discussing it / improving it further.

And open a follow up PR so that we can merge the change into master too.

(Sounds like we may not even do a 1.0.0 documentation release...but a 1.01 release instead...so the master PR will help with that too.)

robhaswell:
Thanks Richard! I agree with your suggestions but that's a different ticket.

PR #1641, opened 2015-06-16
[FLOC-2312] Add note to cURL instructions about providing password

davidoliverSP2:


adamtheturtle:
I think that this should instead be inline in the commands, as is shown in Terminal when the password prompts come up.

```
$ some-command --options
Enter your password:
$ More options
```

Related: I don't understand why / when we should use "note"s.

Please address and resubmit.

davidoliverSP2:
thanks.

My understanding is Notes are used to highlight important information that is outside of the context of the document, or document section. However, @alicesowerby and I can come up with a better rule, defined in the documentation style guide, then roll out across the site. See FLOC 2358

docs/using/authentication.rst
```diff
@@ -64,11 +64,13 @@ If you're not sure what the username is, you can find the common name like this:
     subject= /OU=164b81dd-7e5d-4570-99c7-8baf1ffb49d3/CN=user-allison
 
 In this example, ``user-allison`` is the common name.
-Import the client certificate into the ``Keychain`` and then refer to it by its common name:
+Import the client certificate into the ``Keychain`` and then refer to it by its common name (note that you will need to input a password of any length when creating ``user.p12``):
```
adamtheturtle: This note is not necessary now.

adamtheturtle:
Thanks, please remove the note (it doesn't add any information) and merge.

PR #1644, opened 2015-06-16
[FLOC-2210] Rebase master no agent yaml 

dwgebler:
This is FLOC-2210 rebased against master. There is a separate PR against the 1.0 release branch.


itamarst:
I am worried that merging rebased version into master will cause merge failures later on when we merge release branch into master. I suggest not merging this variant and instead following the recommended procedure in the docs (http://doc-dev.clusterhq.com/gettinginvolved/contributing.html#maintenance-branches). If that is a problem, as it probably will be because this is all terrible, see what @tomprince says.

itamarst:
(I may be worried about nothing; we'll see what @tomprince says.)

tomprince:
It would be better to not rebase this. If just merging to master has issues, I can have a look.

PR #1647, opened 2015-06-16
[FLOC-2256] Removed / changed broken links

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2256

Done by testing the Flocker links, changing those which have moved, removing one which has gone from the docs it seems.


davidoliverSP2:
Looks good to me - tested the links, they work fine. Thanks

PR #1650, opened 2015-06-16
Cannot make filesystem on device that has one

jml:
`mkfs` will blindly run and make a filesystem on a device that already
has one, and it won't even tell you about it.

Look before we leap by running `blkid` to see what's there.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1650)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -421,7 +421,11 @@ class CreateFilesystem(PRecord):
 
 def _ensure_no_filesystem(device):
     """
-    Raises an error if there's already a filesystem on 'device'.
+    Raises an error if there's already a filesystem on ``device``.
+
+    :raises: ``FilesystemExists`` if there is already a filesystem on
+        ``device``.
+    :return: ``None``
     """
     try:
         check_output(
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -424,7 +424,7 @@ def _ensure_no_filesystem(device):
     Raises an error if there's already a filesystem on 'device'.
     """
     try:
-        output = check_output(
+        check_output(
             [b"blkid", b"-p", b"-u", b"filesystem", device.path],
             stderr=STDOUT,
         )
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -8,7 +8,7 @@ devices.
 """
 
 from uuid import UUID, uuid4
-from subprocess import check_output
+from subprocess import CalledProcessError, check_output, STDOUT
 from stat import S_IRWXU, S_IRWXG, S_IRWXO
 from errno import EEXIST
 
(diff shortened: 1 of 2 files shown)


exarkun:
Conflicting with master apparently?  What branch was this branch based on (wish I knew git well enough to answer that question myself)?

FWIW it would probably be good to fix this in both master and the 1.0.1 release branch.

jml:
master

Will make a second PR for 1.0.1

flocker/node/agents/blockdevice.py
```diff
@@ -408,6 +419,30 @@ def run(self, deployer):
         return succeed(None)
 
 
+def _ensure_no_filesystem(device):
+    """
+    Raises an error if there's already a filesystem on 'device'.
```
exarkun: rST for API docs (though we don't use or verify this in any way).  So probably double backtick instead of single quotes.

exarkun: Would be good to have `:rtype:` here too

flocker/node/agents/blockdevice.py
```diff
@@ -408,6 +419,34 @@ def run(self, deployer):
...
+        #   the specified token was not found, or no (specified) devices
+        #   could be identified
+        #
+        # Experimentation shows that there is no output in the case of the
+        # former, and an error printed to stderr in the case of the
+        # latter.
+        if e.returncode == 2 and not e.output:
+            # There is no filesystem on this device.
+            return
+        raise
```
exarkun: It'd be nice if this case could be tested .... but Itamar and I couldn't see a reliable way to trigger the other error cases (at least not without screwing with the arguments).  Maybe just leave a comment noting it's untested, possibly linking to a follow-up issue for addressing this somehow (maybe by switching to a blkid library with a test double with support for easily exercising error paths).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2711,6 +2710,24 @@ def test_create_fails_on_mounted_filesystem(self):
...
+        """
+        mountroot = mountroot_for_test(self)
+        mountpoint = mountroot.child(b"mount-test")
+        scenario = self._run_success_test(mountpoint)
+        afile = mountpoint.child(b"file")
+        afile.setContent(b"data")
+        # Unmount the filesystem
+        run_process([b"umount", mountpoint.path])
+        # Try recreating filesystem; this should fail.
+        self.failureResultOf(scenario.create(), FilesystemExists)
```
exarkun: In addition to raising the exception, we probably want to prove the data is still there.  iow, do the rest of what `test_create_fails_on_mounted_filesystem` does.  And maybe refactor so those two tests have less duplication between them.

exarkun:
Thanks!  Looks pretty good.  Please address remainder of inline comments and then merge (and be sure to do the 1.0.1 follow-up).

flocker/node/agents/test/test_blockdevice.py
jml: Oh, I did that, cut it temporarily and forgot to yank. Added. 

Refactoring done. Probably could do more if we had a declaratively way of making files, ala [treeshape](https://github.com/jml/treeshape).

flocker/node/agents/blockdevice.py
jml: Agreed, which is why I didn't test it :wink:

Have filed [FLOC-2388](https://clusterhq.atlassian.net/browse/FLOC-2388). Comments welcome.

PR #1651, opened 2015-06-16
[FLOC-2368] Revert accidental readme updates

davidoliverSP2:


adamtheturtle:
Thanks @davidoliverSP2 . I can see that this only touches the README and I happen to know that that doesn't affect the build, so I'll merge.

PR #1655, opened 2015-06-16
[FLOC 2087] Example: Using the API with Python

davidoliverSP2:
These examples (this being the first of 3 currently under development) are currently tested manually in the following repo: https://github.com/ClusterHQ/flocker-api-examples

The README.md file details the process of testing, and the results of the tests can be viewed in the "_output" folder.

The process at the moment is that I run the test manually before copying the example files into the Flocker repo. We have an issue to bring these tests into the automated Buildbot tests post v1. See FLOC 2395.

When designing this example, @itamarst recommended that a good way to do this was using requests libraries. Having discussed this with @binocarlos, we decided the example "as-is" is affective, but we opened an issue to create an improved example using libraries post v1.0. See FLOC 2377.

### DIFF
--- a/docs/using/examples/api-example-python.rst
+++ b/docs/using/examples/api-example-python.rst
@@ -1,0 +1,27 @@
+=========================
+Using the API with Python
+=========================
+
+.. note:: You will need to have a running Flocker cluster set up, and at least `Python 2.7.9`_ to run this example.
+
+#. Download the :file:`example.py` file:
+
+   :download:`example.py`
+
+   .. literalinclude:: example.py
+
+#. Before you run the example, you will need to change the following values (either by editing the source file or by setting environment variables):
+
+   * **CONTROL_SERVICE** - the IP address of the Flocker control service
+   * **CONTROL_PORT** - the port the control service is listening on
+   * **KEY_FILE** - the API key file to use
+   * **CERT_FILE** - the API certificate file to use
+   * **CA_FILE** - the certificate authority file to use
+
+#. Now you can run the example:
+
+   .. prompt:: bash
+
+      python example.py
+
+.. _`Python 2.7.9`: https://www.python.org/downloads/
--- a/docs/using/examples/api-examples.rst
+++ b/docs/using/examples/api-examples.rst
@@ -1,0 +1,14 @@
+:tocdepth: 2
+
+====================
+Flocker API Examples
+====================
+
+You can find below an example of using the :ref:`Flocker REST API<api>` with Python.
+
+Examples of using the Flocker API with languages such as `cURL`, `Go`, and `Ruby` will be available soon.
+
+.. toctree::
+   :maxdepth: 2
+
+   api-example-python
--- a/docs/using/examples/example.py
+++ b/docs/using/examples/example.py
@@ -1,0 +1,60 @@
+import httplib
+import os
+import ssl
+import tempfile
+
+# NOTE - to run this example, you need at least python 2.7.9
+
+# Define the control IP, port, and the certificates for authentication.
+
+CONTROL_SERVICE = os.environ.get(
+    "CONTROL_SERVICE", "54.157.8.57")
+CONTROL_PORT = os.environ.get(
+    "CONTROL_PORT", 4523)
+KEY_FILE = os.environ.get(
+    "KEY_FILE", "/Users/kai/projects/flocker-api-examples/flockerdemo.key")
+CERT_FILE = os.environ.get(
+    "CERT_FILE", "/Users/kai/projects/flocker-api-examples/flockerdemo.crt")
+CA_FILE = os.environ.get(
+    "CA_FILE", "/Users/kai/projects/flocker-api-examples/cluster.crt")
+
+# Create a certificate chain and then pass that into the SSL system.
+
+certtemp = tempfile.NamedTemporaryFile()
+TEMP_CERT_CA_FILE = certtemp.name
+os.chmod(TEMP_CERT_CA_FILE, 0600)
+certtemp.write(open(CERT_FILE).read())
+certtemp.write("\n")
+certtemp.write(open(CA_FILE).read())
+certtemp.seek(0)
+ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+ctx.load_cert_chain(TEMP_CERT_CA_FILE, KEY_FILE)
+
+# Finally, create a HTTP connection.
+
+c = httplib.HTTPSConnection(CONTROL_SERVICE, CONTROL_PORT, context=ctx)
+
+def make_api_request(method, endpoint, data=None):
+    if method in ("GET", "DELETE"):
+        c.request(method, endpoint)
+    elif method == "POST":
+        c.request("POST", endpoint, data, 
+            headers={"Content-type": "application/json"})
+    else:
+        raise Exception("Unknown method %s" % (method,))
+
+    r = c.getresponse()
+    body = r.read()
+    status = r.status
+
+    print body
+
+# Make the first request to check the service is working.
+
+make_api_request("GET", "/v1/version")
+
+# Create a volume.
+
+make_api_request("POST", "/v1/configuration/datasets",
+    data= r'{"primary": "%s", "maximum_size": 107374182400, "metadata": {"name": "example_dataset"}}'
+        % ("5540d6e3-392b-4da0-828a-34b724c5bb80",))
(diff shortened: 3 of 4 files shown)


docs/using/examples/api-example-python.rst
```diff
@@ -0,0 +1,27 @@
...
+
+#. Download the :file:`example.py` file:
+
+   :download:`example.py`
+
+   .. literalinclude:: example.py
+
+#. Before you run the example, you will need to change the following values (either by editing the source file or by setting environment variables):
+
+   * **CONTROL_SERVICE** - the IP address of the Flocker control service
```
adamtheturtle: Sphinx has ``:envvar:`` http://sphinx-doc.org/markup/inline.html#role-envvar.

adamtheturtle:
This is marked as a task in progress in JIRA, is it meant to be an improvement in code review ready?

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
+import httplib
+import os
+import ssl
+import tempfile
+
+# NOTE - to run this example, you need at least python 2.7.9
```
adamtheturtle: :question: Maybe instead of noting this, let the code error. Or make it error with a "python version not sufficient" type error.

docs/using/examples/api-examples.rst
```diff
@@ -0,0 +1,14 @@
+:tocdepth: 2
+
+====================
+Flocker API Examples
+====================
+
+You can find below an example of using the :ref:`Flocker REST API<api>` with Python.
+
+Examples of using the Flocker API with languages such as `cURL`, `Go`, and `Ruby` will be available soon.
```
adamtheturtle: Single backticks are for italics, I think that double backticks are what you want.

Is it useful to say what will be available soon? It adds risk in case the roadmap changes (and this is software, it does :smile: ).

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Define the control IP, port, and the certificates for authentication.
+
+CONTROL_SERVICE = os.environ.get(
+    "CONTROL_SERVICE", "54.157.8.57")
+CONTROL_PORT = os.environ.get(
+    "CONTROL_PORT", 4523)
+KEY_FILE = os.environ.get(
+    "KEY_FILE", "/Users/kai/projects/flocker-api-examples/flockerdemo.key")
```
adamtheturtle: Hold on while I create /Users/kai :laughing:

docs/using/examples/api-example-python.rst
```diff
@@ -0,0 +1,27 @@
+=========================
+Using the API with Python
+=========================
+
+.. note:: You will need to have a running Flocker cluster set up, and at least `Python 2.7.9`_ to run this example.
+
+#. Download the :file:`example.py` file:
```
adamtheturtle: Name this file something more informative.

docs/using/examples/api-example-python.rst
```diff
@@ -0,0 +1,27 @@
+=========================
+Using the API with Python
+=========================
+
+.. note:: You will need to have a running Flocker cluster set up, and at least `Python 2.7.9`_ to run this example.
```
adamtheturtle: Python 3 is "at least Python 2.7.9", right?

adamtheturtle: How do I set up a Flocker cluster (link I guess).

docs/using/examples/api-example-python.rst
```diff
@@ -0,0 +1,27 @@
...
+
+#. Download the :file:`example.py` file:
+
+   :download:`example.py`
+
+   .. literalinclude:: example.py
+
+#. Before you run the example, you will need to change the following values (either by editing the source file or by setting environment variables):
```
adamtheturtle: The prerequisites are oddly split between here and the note above.

docs/using/examples/example.py
adamtheturtle: Ah, I see that a note says to change this. If we really want example users, let's stick with `alice` for consistency.

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Make the first request to check the service is working.
+
+make_api_request("GET", "/v1/version")
+
+# Create a volume.
+
+make_api_request("POST", "/v1/configuration/datasets",
+    data= r'{"primary": "%s", "maximum_size": 107374182400, "metadata": {"name": "example_dataset"}}'
+        % ("5540d6e3-392b-4da0-828a-34b724c5bb80",))
```
adamtheturtle: Newline at EOF

adamtheturtle:
> We have an issue to bring these tests into the automated Buildbot tests post v1

Now is post v1 :smile: .

Adding untested examples has problems including:
* If they stop working because of a Flocker change, we might miss that change, and then the examples will break. For example if we change an environment variable name, or add a new required one. We increase the risk of having non-working examples in our docs.
* If they stop working because of a non-Flocker change (in this case unlikely, but e.g. if we used a library and the library's API changed), we might not know. We increase the risk of having non-working examples in our docs.
* Every reviewer (read: every existing and new employee with access to merge on GitHub) has to consider these examples for every API change. That overhead slows down development.
* If we make a change to the API and we think "huh, maybe this stops the API example working" the testing of that is slow and manual.
* If we want to improve this example and test that it still works then:
    * That is slow and manual.
    * We might miss an edge case which the original author considered, but isn't documented because maybe they thought that it was "obvious".

docs/using/examples/api-examples.rst
```diff
@@ -0,0 +1,14 @@
+:tocdepth: 2
+
+====================
+Flocker API Examples
```
adamtheturtle: There is one example. This can be "Example**s**" when there are more.

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
...
+certtemp.write(open(CA_FILE).read())
+certtemp.seek(0)
+ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
+ctx.load_cert_chain(TEMP_CERT_CA_FILE, KEY_FILE)
+
+# Finally, create a HTTP connection.
+
+c = httplib.HTTPSConnection(CONTROL_SERVICE, CONTROL_PORT, context=ctx)
+
+def make_api_request(method, endpoint, data=None):
```
adamtheturtle: Let's meet our internal coding standards and best practices in our examples. Rather than enumerate them here, let's go through them together. A few guidelines are on http://doc-dev.clusterhq.com/gettinginvolved/contributing.html#merge-requirements but they are not complete.

docs/using/examples/api-example-python.rst
```diff
@@ -0,0 +1,27 @@
...
+   * **CERT_FILE** - the API certificate file to use
+   * **CA_FILE** - the certificate authority file to use
+
+#. Now you can run the example:
+
+   .. prompt:: bash
+
+      python example.py
+
+.. _`Python 2.7.9`: https://www.python.org/downloads/
```
adamtheturtle: Unfortunately instructing people on how to install Python is a game you can't win. Let's discuss how best to do this.

docs/using/examples/api-examples.rst
davidoliverSP2: I think your concerns about adding these examples without automated testing needs to be considered by Roger and Michael, as these docs were requested for Dockercon, and Kai has put a lot of effort in to bring these in with manual tests. I'll message them directly.

adamtheturtle:
Can we not use something to generate these examples? I understand that the ideal is to have human-crafted, idiomatic API requests in each language, but can we really do better than a tool? If so, is the difference worth the cost?

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
+import httplib
+import os
+import ssl
+import tempfile
+
+# NOTE - to run this example, you need at least python 2.7.9
+
+# Define the control IP, port, and the certificates for authentication.
+
+CONTROL_SERVICE = os.environ.get(
```
adamtheturtle: If we are going to go with manual testing (see other comments, and consider the results of our past attempts at doing manual testing for our docs / features) then it would be easier if there were a manual script to follow.

For example if we had a `flocker-provision` type tool which created a cluster which could be used with the default environment variables. Maybe hack `run-acceptance-tests` to help (I think it'll do most of the heavy lifting).

docs/using/examples/example.py
```diff
@@ -0,0 +1,60 @@
+import httplib
```
adamtheturtle: Buildbot should lint this file (file location can help).

davidoliverSP2:
closed, following review by @ferrantim - this content is not required at this time.

PR #1656, opened 2015-06-16
[FLOC-2374]  Avoid IBlockDeviceAPI.get_device_path where possible

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2374 for the 1.0.0 release maintenance branch.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -460,6 +472,10 @@ def _incorrect_device_path_test(self, bad_value):
             volume.blockdevice_id, self.api.compute_instance_id(),
         )
 
+        # XXX There are N different cases to test here.  At least: filesystem
+        # exists with uuid, filesystem exists without uuid, filesystem does not
+        # exist.  Only testing does-not-exist case now.
```
jml: :question: Do you normally have XXX comments without assigning them to a person or filing a follow-up task? 

Seems to me that since this is documenting raw work to do, rather than an infelicity in the code, it'd be better to ditch the comment and file a ticket.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -509,9 +525,30 @@ def test_unrelated_mounted(self):
         dataset returned by ``BlockDeviceDeployer.discover_state`` and not as a
         manifestation on the ``NodeState``.
         """
-        unrelated_device = FilePath(self.mktemp())
-        with unrelated_device.open("w") as unrelated_file:
-            unrelated_file.truncate(LOOPBACK_MINIMUM_ALLOCATABLE_SIZE)
+        # Create a different one so that the filesystem we get from it isn't
+        # accidentally treated specially by the API object we're going to let
+        # the deployer use.  In other words, this use of LoopbackBlockDeviceAPI
```
jml: :question: Suggested rephrasing:

```
# Create a different API so that the filesystem we get from it isn't
# accidentally treated specially by the API under test. 
```

i.e.
* s/one/API/
* fewer words at end of sentence.

flocker/node/agents/blockdevice.py
```diff
@@ -551,29 +551,18 @@ def run(self, deployer):
...
-        api = deployer.async_block_device_api
-        listing = api.list_volumes()
-        listing.addCallback(
-            _blockdevice_volume_from_datasetid, self.dataset_id
-        )
+        try:
+            device = get_device_for_dataset_id(self.dataset_id)
+        except KeyError:
+            # It was not actually found.
+            return fail(DatasetWithoutVolume(dataset_id=self.dataset_id))
```
jml: :question: If it makes sense, you could shave some lines by raising `DatasetWithoutVolume` from `get_device_for_dataset_id`, rather than `KeyError`.

jml:
Looks good. All comments are minor & optional. Please address and then merge.

flocker/node/agents/blockdevice.py
itamarst: Mmmm. I feel like it's not quite the same, the former is more like "pretty sure this is not a thing in the whole system" the latter is "sorry couldn't find it locally".

flocker/node/agents/test/test_blockdevice.py
itamarst: We usually require an issue, and then have a comment inline saying "We need to do X; see FLOC-1234".

In this case we expect to do that in next branch in the stack (Jean-Paul is working on it as we speak), so the XXX will be gone by end of day hopefully.

flocker/node/agents/test/test_blockdevice.py
itamarst: Hm hm. Probably won't bother since it's a test and sufficiently clear.

PR #1653, opened 2015-06-16
[FLOC-2369] Store dataset UUIDs on filesystems we create

itamarst:
By storing dataset IDs on the filesystem we create, we don't need to rely on problematic backends like EBS to return accurate information about device<->dataset mappings.

PR #1658, opened 2015-06-16
[FLOC-2372] Remember discovered device path in EBSBlockDeviceAPI instead of relying on EBS volume metadata

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2372 for the 1.0.0 release branch.

Stacked on https://github.com/ClusterHQ/flocker/pull/1653 so review that first.


exarkun:
WRONG TARGET BRANCH

PR #1659, opened 2015-06-16
[FLOC-2372] Remember discovered device path in EBSBlockDeviceAPI instead of relying on EBS volume metadata

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2372 for the 1.0.0 release branch.


flocker/node/agents/ebs.py
```diff
@@ -647,18 +645,17 @@ def get_device_path(self, blockdevice_id):
...
-            device = ebs_volume.tags[ATTACHED_DEVICE_LABEL]
+            return self._device_paths[blockdevice_id]
         except KeyError:
+            # Maybe we screwed up.  Do a sanity check instead of just claiming
+            # the caller screwed up?
+            #
+            ebs_volume = self._get_ebs_volume(blockdevice_id)
+            # volume = _blockdevicevolume_from_ebs_volume(ebs_volume)
+            # if volume.attached_to is None:
+            #     raise UnattachedVolume(blockdevice_id)
```
jml: Why the commented out code? Why assign to `ebs_volume` even though it's unused?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1722,6 +1723,28 @@ def test_get_device_path_unattached_volume(self):
         )
         self.assertEqual(new_volume.blockdevice_id, exception.blockdevice_id)
 
+    def test_get_device_path_detached_volume(self):
+        """
+        ``get_device_path`` raises ``UnattachedVolume`` if the supplied
+        ``blockdevice_id`` corresponds to a volume that was attached to the
+        node but has been detached from it.
+        """
```
jml: No change needed.

Out of curiosity, was this test failing before the changes to ebs?

itamarst: In old implementation EC2 APIs were queried each time. So it was same code path as never seen before unattached volume. In new implementation we can only give information about volumes that have been attached during the lifetime of the process, so this particular test covers a new code path in the new implementation.

itamarst:
We discovered a new requirement related to cde path you pointed out, so will resubmit once that's done.

itamarst:
OK ready for review again.

PR #1660, opened 2015-06-17
[FLOC-2400] Remove useless return of `Cluster` from `Cluster` methods

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2400

This is in preparation for getting proper eliot logs from the acceptance tests.

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -72,8 +72,8 @@ class ContainerAPITests(TestCase):
         """
         Create a container listening on port 8080.
 
-        :return: ``Deferred`` firing with a tuple of ``Cluster`` instance
-        and container dictionary once the container is up and running.
+        :return: ``Deferred`` firing with a container dictionary once the
+            container is up and running.
         """
         data = {
             u"name": random_name(self),
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -85,8 +85,7 @@ class ContainerAPITests(TestCase):
 
         d = cluster.create_container(data)
 
-        def check_result(result):
-            cluster, response = result
+        def check_result(response):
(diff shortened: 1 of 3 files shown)


flocker/acceptance/test_api.py
```diff
@@ -85,8 +85,7 @@ def _create_container(self, cluster):
 
         d = cluster.create_container(data)
 
-        def check_result(result):
-            cluster, response = result
+        def check_result(response):
```
jongiddy: Update `_create_container` docstring `:return:`. It only returns the container dictionary now.

jongiddy:
The decorator `log_method` should document that it does not work for certain types of objects.  Is there any way it can detect and remap such arguments?  Also, it seems Eliot should deal cleanly with problematic values.  Please file an upstream issue.

It is not good to be rewriting parameter lists in order to overcome constraints provided by the logging system. I would like another reviewer to OK this approach, so will put it back in review.

tomprince:
> It is not good to be rewriting parameter lists in order to overcome constraints provided by the logging system. I would like another reviewer to OK this approach, so will put it back in review.

I would agree with this, if the value was actually providing any value, but
- many places calling this function were actually ignoring the return.
- For those that weren't, they assign the returned 
`Cluster` to a fresh `cluster` that shadows another `cluster` variable holding the same `Cluster`.

exarkun:
> > It is not good to be rewriting parameter lists in order to overcome constraints provided by the logging system. I would like another reviewer to OK this approach, so will put it back in review.

Agree with this sentiment in general.

> I would agree with this, if the value was actually providing any value

And agree with this response specifically.

The interface was weird before and this change makes sense even ignoring the log-related motivation.

So - looks good to me, BuildBot is basically green - please merge.

PR #1663, opened 2015-06-17
Do not call mkfs on devices that already have filesystems

jml:
This applies the diff at #1650 to the release branch.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1663)


dwgebler:
There are some spurious build failures (resource limits again). I think those can be safely ignored, so good to merge :)

PR #1664, opened 2015-06-17
[FLOC-2399] Release maintenance/flocker 1.0.0/deployment config

dwgebler:
Fixes FLOC-2399


itamarst:
This means we will have configurations that work with 1.0.1 but not 1.0.0; expectation is that 1.0.x releases will be internally consistent in feature set. On the other hand docs say to do this so maybe someone will expect it to work.

And... it's not in critical path for new users.

So I suggest redoing this as branch against master, and since it's already written add a sentence to relevant docs to mention we support both formats.

For future reference the product teams want to get rid of `deployment.yml` so it's probably not worth doing much more work on it, so for similar issues maybe just update docs to be accurate instead of extending config parser.

PR #1666, opened 2015-06-17
[FLOC-2406] Skip tests if zpool not installed

jml:
Also use `run_process`, which is infinitely superior to `check_call`.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1666)


adamtheturtle:
Thanks @jml, this seems to work:

flocker.volume.functional.test_filesystems_zfs.StoragePoolTests failed on my Mac before, and now they pass. The tests are still running on BuildBot, I guess
that this is the success criteria. There is a mix of `docs.clusterhq.com` and `doc-dev` in our source code but I think that it is best to point developers to `doc-dev`, which is `master`'s docs.

Also, that page is big - I don't know what to look at. If I look for "zpool" there is nothing to do with "installing zpool", only "Creating a ZFS Pool".

If that is what is meant, link to that section (https://doc-dev.clusterhq.com/using/installing/index.html#creating-a-zfs-pool) and change the message slightly.

If only the minor changes suggested are done, please merge. Else if there is something more significant, please resubmit.

jml:
Linked to http://doc-dev.clusterhq.com/using/installing/index.html#optional-zfs-backend-configuration, since you don't need to create a zpool, you just need to be able to create one.

jml:
Yay! Only failing test is now expected failure!

PR #1667, opened 2015-06-17
[FLOC-2398] flocker deploy certs

dwgebler:
Fixes FLOC-2398. Also fixes FLOC-2349 and FLOC-2397.


itamarst:
I am not sure this should go into 1.0.x.. Regardless of the merit of the code, it seems like a very large change to fix a small problem as far as FLOC-2398 goes.

itamarst:
It is also not backwards compatible.

itamarst:
I don't think FLOC-2398 is particularly a problem for 1.0, for that matter. Re-do against master, and remove backwards incompatibilty.

(Ideally this would be two PRs, one for code improvements, one for script improvements.)

PR #1671, opened 2015-06-18
[FLOC-2376] Upgrade ext4 filesystems created by Flocker 1.0.0 to carry dataset identifying information

exarkun:
(With Itamar)
Fixes https://clusterhq.atlassian.net/browse/FLOC-2376 in the 1.0.0 release branch (for 1.0.1).
See also https://clusterhq.atlassian.net/browse/FLOC-314

This adds an upgrade step to the "discovery" phase of each iteration of the `BlockDeviceDeployer`-based convergence loop.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,10 @@
+Flocker 1.0.1 (2015-06-XX)
+==========================
+
+- Flocker-created filesystems now include information identifying the dataset they belong to.
+  Filesystems created by Flocker 1.0.0 will be upgraded to include this information automatically if and only if they are not in use.
+  See the upgrade documentation for details. (FLOC-2369, FLOC-2372, FLOC-2374, FLOC-2376)
+
 Flocker 1.0.0 (2015-06-15)
 ==========================
 
--- a/docs/using/administering/upgrading.rst
+++ b/docs/using/administering/upgrading.rst
@@ -1,0 +1,38 @@
+=========
+Upgrading
+=========
+
+A Flocker cluster can be upgraded to a newer version of Flocker while preserving all data and configuration of the Flocker cluster provided certain steps are followed.
+The correct steps to follow may vary depending on the versions of Flocker being upgraded from and to.
+A common requirement for all currently supported upgrade paths is that all nodes and the control service must be running the same version of Flocker.
+
+Flocker 1.0.1
+-------------
+
+Recommended Steps
+^^^^^^^^^^^^^^^^^
+
+  #. Stop the control service and the agent services on all nodes.
+  #. Install Flocker 1.0.1 on all hosts in the Flocker cluster.
+  #. Reboot each of the agent nodes.
+  #. If you have not configured the Flocker agents to start automatically on boot,
+     restart the agent services on all nodes.
+  #. Restart the control service.
+
+Details
+^^^^^^^
+
+The upgrade to Flocker 1.0.1 involves changing the metadata on the filesystems created on the volumes managed by the AWS EBS and OpenStack Cinder storage drivers.
+These metadata changes will be made automatically by ``flocker-dataset-agent`` as part of its normal operation.
+However, these metadata changes can only be performed when the filesystems themselves are unmounted (therefore not in use).
+The easiest way to ensure ``flocker-dataset-agent`` will be able to perform these upgrades is to reboot the node where the agent is running.
+When the node starts up again, all Flocker-managed filesystems will start unmounted.
+``flocker-dataset-agent`` will upgrade each filesystem before re-mounting the filesystem for use by applications.
+You can observe the upgrade process by finding ``agent:blockdevice:1.0.0:upgrade:fs`` and ``agent:blockdevice:1.0.0:mounted:fs`` events in the ``flocker-dataset-agent`` log.
+The former indicates a filesystem upgrade is being attempted.
+The latter indicates a filesystem was found mounted and no upgrade was attempted.
+
+Flocker 1.0.0
+-------------
+
+There are no supported upgrade paths from versions of Flocker older than 1.0.0.
--- a/docs/using/administering/upgrading.rst
+++ b/docs/using/administering/upgrading.rst
@@ -1,3 +1,5 @@
+.. _upgrading-flocker:
+
 =========
 Upgrading
 =========
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -128,6 +128,13 @@ class FilesystemExists(Exception):
         self.device = device
 
 
+class FilesystemMissing(Exception):
+    """
+    A block device on which a filesystem was anticipated did not have a
+    filesystem.
+    """
+
+
 DATASET = Field(
     u"dataset",
     lambda dataset: dataset.dataset_id,
@@ -1412,23 +1419,20 @@ def _manifestation_from_volume(volume):
 
 
 def _ext4_fs_uuid(device_path):
-    blkid = check_output([
-        b"blkid",
+    blkid = _blkid(
         # Display a slightly more parseable format.
         b"-o", b"udev",
         # Display information about this device
         device_path.path,
-    ])
+    )
     # Output is lines like:
     # ID_FS_UUID=a0d8c288-2dc9-4fa8-9039-f834d69730c6
     # ID_FS_UUID_ENC=a0d8c288-2dc9-4fa8-9039-f834d69730c6
     # ID_FS_TYPE=ext4
-
     for line in blkid.splitlines():
         key, value = line.split(b"=")
         if key == b"ID_FS_UUID":
             return value.decode("ascii")
-
     raise ValueError(
         "Failed to find UUID on {} in output: {}".format(
             device_path.path, blkid
@@ -1987,6 +1991,16 @@ class DuplicateFilesystemId(Exception):
     """
 
 
+def _blkid(*argv):
+    try:
+        return check_output([b"blkid"] + list(argv))
+    except CalledProcessError as e:
+        if e.returncode == 2:
+            raise FilesystemMissing()
+        # This code path is untested, unfortunately.
+        raise
+
+
 def get_device_for_dataset_id(dataset_id):
     """
     Lookup the path of the device for an attached dataset on this machine.
@@ -1999,17 +2013,14 @@ def get_device_for_dataset_id(dataset_id):
     :return: ``FilePath`` of device where the dataset is attached.
     """
     try:
-        result = check_output([b"blkid",
-                               # Only list the device:
-                               b"-o", b"device",
-                               # Search by UUID:
-                               b"-t", b'UUID="{}"'.format(dataset_id)])
-    except CalledProcessError as e:
-        if e.returncode == 2:
-            raise KeyError(dataset_id)
-        else:
-            # This code path is untested, unfortunately.
-            raise
+        result = _blkid(
+            # Only list the device:
+            b"-o", b"device",
+            # Search by UUID:
+            b"-t", b'UUID="{}"'.format(dataset_id),
+        )
+    except FilesystemMissing:
+        raise KeyError(dataset_id)
     paths = result.splitlines()
     if len(paths) > 1:
         raise DuplicateFilesystemId(paths)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -1560,7 +1560,7 @@ class BlockDeviceDeployer(PRecord):
             if device_path in mounted:
                 FLOCKER_1_0_0_FS_MOUNTED(
                     block_device_id=volume.blockdevice_id,
-                    device_path=device_path.path,
+                    block_device_path=device_path,
                 ).write()
                 continue
 
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -198,9 +198,10 @@ CREATE_BLOCK_DEVICE_DATASET = ActionType(
     u"A block-device-backed dataset is being created.",
 )
 
-FLOCKER_1_0_0_FS_UPGRADED = MessageType(
+FLOCKER_1_0_0_FS_UPGRADE = ActionType(
     u"agent:blockdevice:upgrade:1.0.0:fs",
-    fields(block_device_id=unicode, dataset_id=unicode, old_fs_uuid=unicode),
+    [BLOCK_DEVICE_ID, DATASET_ID],
+    fields(old_fs_uuid=unicode, tune2fs_output=unicode),
     u"A filesystem created by Flocker 1.0.0 was encountered and its "
     u"filesystem UUID was changed to match the dataset id it is meant to "
     u"represent.",
@@ -1403,6 +1404,35 @@ def _manifestation_from_volume(volume):
     return Manifestation(dataset=dataset, primary=True)
 
 
+def _ext4_fs_uuid(device_path):
+    blkid = check_output([
+        b"blkid",
+        # Display a slightly more parseable format.
+        b"-o", b"udev",
+        # Display information about this device
+        device_path.path,
+    ])
+    # Output is lines like:
+    # ID_FS_UUID=a0d8c288-2dc9-4fa8-9039-f834d69730c6
+    # ID_FS_UUID_ENC=a0d8c288-2dc9-4fa8-9039-f834d69730c6
+    # ID_FS_TYPE=ext4
+
+    for line in blkid.splitlines():
+        key, value = line.split(b"=")
+        if key == b"ID_FS_UUID":
+            return value.decode("ascii")
+
+    raise ValueError(
+        "Failed to find UUID on {} in output: {}".format(
+            device_path.path, blkid
+        )
+    )
+
+
+def tune2fs(*argv):
+    return check_output([b"tune2fs"] + list(argv), stderr=STDOUT)
+
+
 def _attached_volumes(compute_instance_id, volumes):
     """
     Yield volumes from the given iterable which are attached to the given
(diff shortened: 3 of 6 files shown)


exarkun:
It would be great to get input from @davidoliverSP2 or @alicesowerby or @ferrantim regarding the upgrade documentation changes that come with this PR.

exarkun:
We lack a framework for functional testing of upgrades.  It's late here so I'm not going to get to manual testing of this tonight.  I'll do it first thing tomorrow unless someone else wants to try (it would be interesting to see if someone else can figure out how to do the upgrade based on the documentation I wrote).

davidoliverSP2:
The line items under "recommended steps" on the Upgrading page have a visible indent bar - I think the line items do not need the 3 spaces before each #, as there is no preceding directive.

davidoliverSP2:
The Details paragraph is quite long, and could be broken up a little (I've also edited the wording a little):

The upgrade to Flocker 1.0.1 involves changing the metadata on the filesystems that were created on the volumes managed by the AWS EBS and OpenStack Cinder storage drivers. 

These metadata changes will be made automatically by flocker-dataset-agent as part of its normal operation. However, these metadata changes can only be performed when the filesystems themselves are unmounted, and therefore not in use. 

The easiest way to ensure flocker-dataset-agent will be able to perform these upgrades is to reboot the node where the agent is running. When the node starts up again, all Flocker-managed filesystems will start unmounted. flocker-dataset-agent will upgrade each filesystem before re-mounting the filesystem for use by applications. 

You can observe the upgrade process by finding agent:blockdevice:1.0.0:upgrade:fs and agent:blockdevice:1.0.0:mounted:fs events in the flocker-dataset-agent log. The former indicates a filesystem upgrade is being attempted. The latter indicates a filesystem was found mounted and no upgrade was attempted.

NEWS
```diff
@@ -1,3 +1,10 @@
+Flocker 1.0.1 (2015-06-XX)
```
adamtheturtle: Not completely related but:

1. Thank you for doing the NEWS in a PR, it makes the release engineer's job easier and I think that we should add it as a contribution requirement.
2. If everyone did this and release notes and we changed to "Next Release" as a title (like with the release notes), then weekly releases could remove the review stage and be much faster.

davidoliverSP2:
Thanks @exarkun - these docs are really helpful.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,40 @@
...
+
+Recommended Steps
+^^^^^^^^^^^^^^^^^
+
+  #. Stop the control service and the agent services on all nodes.
+  #. Install Flocker 1.0.1 on all hosts in the Flocker cluster.
+  #. Reboot each of the agent nodes.
+  #. If you have not configured the Flocker agents to start automatically on boot,
+     restart the agent services on all nodes.
+  #. Restart the control service.
```
wallrj: Too much indentation here; results in the list items being block quoted too.
 * http://build.clusterhq.com/results/docs/release-maintenance/flocker-1.0.0/blockdevicedeployer-upgrade-old-filesystems-FLOC-2376/build-8328/using/administering/upgrading.html#recommended-steps

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,40 @@
...
+The correct steps to follow may vary depending on the versions of Flocker being upgraded from and to.
+A common requirement for all currently supported upgrade paths is that all nodes and the control service must be running the same version of Flocker.
+
+Flocker 1.0.1
+-------------
+
+Recommended Steps
+^^^^^^^^^^^^^^^^^
+
+  #. Stop the control service and the agent services on all nodes.
```
wallrj: You have to stop the agent processes first, else the control service will be restarted by socket activation when the agents attempt to reconnect.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,40 @@
...
+  #. Install Flocker 1.0.1 on all hosts in the Flocker cluster.
```
wallrj: "hosts" here, "nodes" above. I vote for "nodes" (or "servers") throughout.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,40 @@
...
+  #. Reboot each of the agent nodes.
+  #. If you have not configured the Flocker agents to start automatically on boot,
+     restart the agent services on all nodes.
```
wallrj: Probably better to restart the control service first to avoid alarming AMP connection errors in the agent logs when they first start up.

flocker/node/agents/blockdevice.py
```diff
@@ -198,6 +205,22 @@ def __init__(self, device):
...
+    u"so it could not be upgraded.",
+)
+
+FLOCKER_1_0_0_FS_UPGRADE = ActionType(
+    u"agent:blockdevice:1.0.0:upgrade:fs",
+    [BLOCK_DEVICE_ID, DATASET_ID],
+    fields(old_fs_uuid=unicode, tune2fs_output=unicode),
+    u"A filesystem created by Flocker 1.0.0 was encountered and its "
+    u"filesystem UUID was changed to match the dataset id it is meant to "
+    u"represent.",
```
wallrj: :question: Maybe drop the "it is meant to represent."

wallrj: But this message doesn't actually show up in the logs, right? So maybe ignore that.

docs/using/administering/upgrading.rst
davidoliverSP2: we've been using nodes in place of hosts/servers, to be consistent with the code samples

docs/using/administering/upgrading.rst
davidoliverSP2: The line items under "recommended steps" on the Upgrading page have a visible indent bar - I think the line items do not need the 3 spaces before each #, as there is no preceding directive.

flocker/node/agents/blockdevice.py
```diff
@@ -1395,6 +1418,61 @@ def _manifestation_from_volume(volume):
...
+    Get the EXT4 filesystem UUID from the filesystem on the device at the given
+    path.
+
+    :param FilePath device_path: The path to the device on which the filesystem
+        resides.
+
+    :return: A ``UUID`` instance giving the UUID from the filesystem.
+    :rtype: ``UUID``
+    """
+    blkid = _blkid(
```
wallrj: :question: Apparently `blkid` uses a cache by default. Could it give us stale results?
```
man blkid
...
       -p     Switch to low-level superblock probing mode (bypassing the cache).

              Note that low-level probing also returns information about partition table type (PTTYPE tag) and partitions (PART_ENTRY_* tags).
```

flocker/node/agents/blockdevice.py
```diff
@@ -1395,6 +1418,61 @@ def _manifestation_from_volume(volume):
...
+    """
+    Run the ``tune2fs`` utility with the given arguments and return its stdout
+    and stderr.
+
+    :param argv: ``bytes`` giving the argument vector to pass to the process.
+
+    :return: ``bytes`` giving the stdout and stderr of tune2fs (mixed
+        together).
+    """
+    return check_output([b"tune2fs"] + list(argv), stderr=STDOUT)
```
wallrj: :question: And if the command fails, this'll mean that the STDERR is included in the ``CalledProcessError``, right?....which I think is a good thing for debugging if this fails for some reason.

flocker/node/agents/blockdevice.py
```diff
@@ -1395,6 +1418,61 @@ def _manifestation_from_volume(volume):
...
+
+
+def _attached_volumes(compute_instance_id, volumes):
+    """
+    Yield volumes from the given iterable which are attached to the given
+    instance.
```
wallrj: :question: Parameter and return type docs, ideally.

wallrj: :question: `compute_instance_id` may return `None` ...which is probably a mistake...do you think it ought to raise an exception instead? If so, comment on https://clusterhq.atlassian.net/browse/FLOC-2418

flocker/node/agents/blockdevice.py
```diff
@@ -1468,6 +1540,94 @@ def _get_system_mounts(self, volumes, compute_instance_id):
...
+        compute_instance_id = api.compute_instance_id()
+        for volume in _attached_volumes(compute_instance_id, volumes):
+            try:
+                get_device_for_dataset_id(volume.dataset_id)
+                # If we found the dataset_id like this the filesystem is
+                # sufficiently new to not need any changes.  Proceed to the
+                # next volume.
+                continue
+            except KeyError:
+                # There is no filesystem on the volume because volume setup
```
itamarst: I think the order here is wrong. By doing this first it means that if you have missing filesystem you will always detach and attach, which is sloooooooow. So missing filesystem goes from being annoyance to being pretty terrible.

Instead we should first try get_device_path maybe? THough even then you get slowdowns if you restart dataset agents.... So still a problem.

Big picture point: upgrades shouldn't cause slow downs ever when not upgrading,and right now they do. Can we somehow disable upgrades once it's clear they're not necessary?

flocker/node/agents/blockdevice.py
```diff
@@ -1453,13 +1532,6 @@ def _get_system_mounts(self, volumes, compute_instance_id):
             to ``volumes``.
```
wallrj: Also document the new `device_to_dataset_id` parameter.

flocker/node/agents/blockdevice.py
```diff
@@ -1468,6 +1540,94 @@ def _get_system_mounts(self, volumes, compute_instance_id):
...
+                # was created by 1.0.0.
+                #
+                # The filesystem isn't mounted because the 1.0.1 upgrade
+                # instructions tell the user to restart the node upon upgrade.
+                #
+                # Detach and then re-attach the volume so that get_device_path
+                # can tell us the device.  This is slow and we might be able to
+                # save time by calling get_device_path first.  It might give us
+                # an answer.  Maybe do that optimization.
+                api.detach_volume(volume.blockdevice_id)
```
wallrj: But this'll just hang if the filesystem *is* already mounted.
I don't think this is right.

wallrj: What I mean, is that I don't think you and detach a mounted block device....at least not in Rackspace / OpenStack.

flocker/node/agents/blockdevice.py
```diff
@@ -1468,6 +1540,94 @@ def _get_system_mounts(self, volumes, compute_instance_id):
...
+
+            # Find out of it is mounted.  Skip it (don't break the rest of the
+            # loop) if it is mounted because there's nothing we can do about
+            # mounted filesystems.  They're not supposed to be mounted because
+            # we told the user to restart the node.
+            if device_path in mounted:
+                FLOCKER_1_0_0_FS_MOUNTED(
+                    block_device_id=volume.blockdevice_id,
+                    block_device_path=device_path,
+                ).write()
```
wallrj: :question: Is there a way to mark eliot messages as errors?
There's a similar case in `CinderBlockDeviceAPI.compute_instance_id` where I log an Eliot message for the case where the instance_id isn't found, but it'd be difficult to find that error if I searched the logs for terms like exception, failure, traceback, `succeeded=false` etc.

flocker/node/agents/blockdevice.py
```diff
@@ -1468,6 +1540,94 @@ def _get_system_mounts(self, volumes, compute_instance_id):
...
+                # FLOC-1576
+                continue
+
+            # Put the dataset id into the ext4 filesystem UUID field for future
+            # reference.
+            with FLOCKER_1_0_0_FS_UPGRADE(
+                block_device_id=volume.blockdevice_id,
+                dataset_id=volume.dataset_id
+            ) as upgrade:
+                upgrade.addSuccessFields(old_fs_uuid=unicode(old_fs_uuid))
```
wallrj: :question: I wonder if the `clusterhq-flocker-node` package has an `e2fsprogs` dependency?

wallrj: No. https://github.com/ClusterHQ/flocker/blob/master/admin/packaging.py#L237

I might create an issue while I think about it.

flocker/node/agents/blockdevice.py
```diff
@@ -1489,6 +1649,17 @@ def is_existing_block_device(dataset_id, path):
...
+        # The state discovery method is basically the wrong place to do the
+        # filesystem upgrade.  Sadly, determining the necessity of each of
+        # these upgrades depends on more state than we can easily convey to
+        # `calculate_changes` (the technical term for this situation might be
+        # "design flaw").  Since we really do want to do these upgrades and
+        # we'd like to do them with a minimum of code churn, sticking this
+        # upgrade step into the discovery method makes the most sense.  This
+        # will be another datapoint to consider when revamping the way this
+        # IDeployer works.
+        self._upgrade_1_0_0_filesystem(api, volumes)
```
wallrj: :question: Seems like quite a lot of extra work, to be performing this check every time we loop for ever more, regardless of whether the volumes have already been upgraded.

 * Is it worth adding a `flocker-version` stamp or something to the volume metadata?
 * And maybe only performing this upgrade step when the agent starts up?
 * And what happens if the volume is mounted and can't be upgraded? Then the following code that calls `get_device_dataset_id` will fail every time....I think.

flocker/node/agents/blockdevice.py
```diff
@@ -1865,6 +2043,16 @@ class DuplicateFilesystemId(Exception):
     """
 
 
+def _blkid(*argv):
```
wallrj: Docstring.

flocker/node/agents/blockdevice.py
```diff
@@ -1865,6 +2043,16 @@ class DuplicateFilesystemId(Exception):
...
+    try:
+        return check_output([b"blkid"] + list(argv))
```
wallrj: :question: Capture stderr so that it's included in CalledProcessError and logged by Eliot rather than leaking and corrupting the logs.

flocker/node/agents/blockdevice.py
```diff
@@ -1865,6 +2043,16 @@ class DuplicateFilesystemId(Exception):
...
+    except CalledProcessError as e:
+        if e.returncode == 2:
```
wallrj: man page says
```
If the specified token was not found, or no (specified) devices could be identified, an exit code of 2 is returned.
```

So might it be worth logging and error message including the stderr here in case there's some other reason for this return code.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1110,6 +1119,108 @@ def test_no_unmount_if_in_use(self):
...
+    # An id that tests and logging decorators can use.
+    dataset_id = uuid4()
+
+    def _upgrade_ext4_uuid_test(self, mounted):
+        """
+        Test the handling of a Flocker 1.0.0-style filesystem by
+        ``BlockDeviceDeployer.discover_state``.
+
+        :param bool mounted: ``True`` to test the behavior when the filesystem
+            is mounted, ``False`` to tested it unmounted.
```
wallrj: typo "tested it unmounted"

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1110,6 +1119,108 @@ def test_no_unmount_if_in_use(self):
...
+            self, deployer,
+            expected_manifestations=expected_manifestations,
+            expected_devices={self.dataset_id: device_path},
+            expected_nonmanifest_datasets=expected_nonmanifest_datasets,
+        )
+        if not mounted:
+            # Only unmounted filesystems are upgraded.
+            self.assertEqual(
+                device_path, get_device_for_dataset_id(self.dataset_id),
+            )
```
wallrj: :question: Should there be an `assertRaises(KeyError,...` here to demonstrate that the filesystem hasn't been upgraded? Or is it simply not possible to run tune2fs on a mounted filesystem?

wallrj:
Thanks @exarkun and @itamarst 

This seems pretty nice.
I left some inline comments above, some of which I think need addressing.
I'll reject this for now and test the upgrade process after lunch.

flocker/node/agents/blockdevice.py
exarkun: > Big picture point: upgrades shouldn't cause slow downs ever when not upgrading

Not sure I agree with this.  We need to be pragmatic.  Some slowdowns don't matter.  This **particular** slowdown may be too egregious, though.

The other idea I considered (instead of asking get_device_path first) was only trying once per process-lifetime per dataset_id.  This would still be slow but only infrequently.

Asking get_device_path first might be the right thing to do.  I'm not sure.  I'm concerned it will produce the wrong results (detach/attach seems more likely to be reliable).  But we already depend on it being right after the detach/attach so that may not be a valid concern.  The real impediment to that approach is how to write tests for the extra cases.

itamarst: Once per process time per dataset id sounds reasonable.

flocker/node/agents/blockdevice.py
itamarst: Actions are the eliot abstraction that can succeed or fail. So make sure things that can fail are logged as actions, then raise an exception that finishes an action (e.g. in context manager or result of `eliot.twisted.DeferredContext`-wrapped Deferred).

exarkun:
Thanks for the review @wallrj.  Setting this work aside in favor of #1684 for the time being given some uncertainty about certain of the failure behaviors of this change, though.

docs/using/administering/upgrading.rst
myechuri: Fixed.

docs/using/administering/upgrading.rst
myechuri: Fixed.

docs/using/administering/upgrading.rst
myechuri: Changed to "nodes".

docs/using/administering/upgrading.rst
myechuri: Changed order to start control service before agents.

flocker/node/agents/blockdevice.py
tomprince: https://www.kernel.org/pub/linux/utils/util-linux/v2.21/libblkid-docs/libblkid-Cache.html suggest the results will always be up-to-date, if we have read access to the device (which we will, running as root).

PR #1673, opened 2015-06-18
[FLOC-2399] Allow either single app string or list in deployment config

dwgebler:
Fixes FLOC-2399

### DIFF
--- a/docs/using/config/configuration.rst
+++ b/docs/using/config/configuration.rst
@@ -302,8 +302,10 @@ Deployment Configuration
 The deployment configuration specifies which applications are run on what nodes.
 It consists of a version and a mapping from node names to application names.
 
-The required parameters are ``version`` and ``applications``.
+The required parameters are ``version`` and ``nodes``.
 For now the ``version`` must be ``1``.
+
+Each hostname in the deployment configuration should be either a single application name or a list of application names that match those defined in the application configuration.
 
 Here's an example of a simple but complete configuration defining a deployment of one application on one host:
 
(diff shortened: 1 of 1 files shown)


flocker/control/_config.py
```diff
@@ -1259,11 +1259,17 @@ def deployment_from_configuration(deployment_state, deployment_configuration,
     seen_applications = set()
     for hostname, application_names in (
             deployment_configuration['nodes'].items()):
+        # If application_names is a string type, convert to a list.
+        # This allows single a application name in a deployment configuration.
+        # The application will still be rejected if it does not exist in the
+        # application manifest.
+        if isinstance(application_names, (str, unicode)):
+            application_names = [application_names]
```
exarkun: Please remove this part.  Only a list of strings should be accepted.

dwgebler: So you just want the doc change? I don't have a personal preference, but I wonder why would we not want to accept a single app name as a string in the case of a deployment config only deploying one app? Forcing people to put a `-` or `[]` in the config for one application is possibly a more confusing experience for the user.

docs/using/config/configuration.rst
```diff
@@ -302,9 +302,11 @@ Deployment Configuration
 The deployment configuration specifies which applications are run on what nodes.
 It consists of a version and a mapping from node names to application names.
 
-The required parameters are ``version`` and ``applications``.
+The required parameters are ``version`` and ``nodes``.
 For now the ``version`` must be ``1``.
 
+Each hostname in the deployment configuration should be a list of application names that match those defined in the application configuration.
```
exarkun: I think another branch changes the example to stop using a hostname and use an IP address instead (which is the only thing that works).  So it would be nice to say something other than "hostname" here.  "Node address"?

Also the sentence reads oddly.  The hostname isn't a list of application names.  It's a key in a mapping and the associated value is a list of application names.

exarkun:
Thanks.  Address one minor comment, get a green build (rebuild failed builders if necessary), and then merge.

PR #1674, opened 2015-06-18
[FLOC-2398] Flocker deploy certs

dwgebler:
Fixes FLOC-2398. Also fixes FLOC-2349 and FLOC-2397.


docs/using/config/usage.rst
```diff
@@ -44,10 +44,11 @@ To authenticate with the control service, you will need a copy of the public clu
 
 For ``flocker-deploy``, your API user certificate and key should be in files named ``user.crt`` and ``user.key`` and the cluster certificate in file ``cluster.crt``.
 
-By default, ``flocker-deploy`` will look for these certificate files in the current working directory.
-If this is not where the files are located, you may specify the ``--certificates-directory`` option to ``flocker-deploy``:
+By default, ``flocker-deploy`` will look for these certificate files in the current working directory and expect them to be named ``cluster.crt`` (the public cluster certificate), ``user.crt`` (the API user certificate) and ``user.key`` (the API user's private key).
+
+You can override any of these defaults with the ``--cafile`` (cluster certificate), ``--cert`` (user certificate) and ``--key`` (user private key) options, specifying the full path to each file.
```
adamtheturtle: This seems to me like a weird interface, with most tools I can specify just the filename of a file in the current working directory.

dwgebler: As per normal Linux behaviour (don't know about OS X but I'd imagine it's the same), if the files are in your working directory, you can just specify the file name without an absolute path. Perhaps changing to "specifying the path to each file" would suffice.....I'm not sure. But generally in Linux, the "full path" rather than the "absolute path" is technically just the file name if it's in your CWD. Anyway, if you can think of a good wording suggestion for this sentence, please add it in a comment :)

adamtheturtle: I'd get rid of "full" but if it works it works, and so this is optional.

adamtheturtle: I was more worried about people trying it and it not working than people thinking that the tool was limited - but that is an incorrect concern.

docs/using/config/usage.rst
```diff
@@ -44,10 +44,11 @@ To authenticate with the control service, you will need a copy of the public clu
 
 For ``flocker-deploy``, your API user certificate and key should be in files named ``user.crt`` and ``user.key`` and the cluster certificate in file ``cluster.crt``.
 
-By default, ``flocker-deploy`` will look for these certificate files in the current working directory.
-If this is not where the files are located, you may specify the ``--certificates-directory`` option to ``flocker-deploy``:
+By default, ``flocker-deploy`` will look for these certificate files in the current working directory and expect them to be named ``cluster.crt`` (the public cluster certificate), ``user.crt`` (the API user certificate) and ``user.key`` (the API user's private key).
```
adamtheturtle: `:file:` for files. We aren't great at this but should get better.

docs/using/config/usage.rst
```diff
@@ -44,10 +44,11 @@ To authenticate with the control service, you will need a copy of the public clu
...
+
+You can override any of these defaults with the ``--cafile`` (cluster certificate), ``--cert`` (user certificate) and ``--key`` (user private key) options, specifying the full path to each file.
 
 .. code-block:: console
 
-   $ flocker-deploy --certificates-directory=/home/alice/flocker-credentials 172.16.255.250 clusterhq_deployment.yml clusterhq_app.yml
+   $ flocker-deploy --cafile=/home/alice/credentials/mycluster.crt --cert=/home/alice/credentials/alice.crt --key=/home/alice/credentials/alice.key 172.16.255.250 clusterhq_deployment.yml clusterhq_app.yml
```
adamtheturtle: :question: The only other `--cafile` option I know of is `openssl` which uses `--CAfile`.

dwgebler: Actually now I think about it, it should probably be `--cacert` which is what cURL uses and is also then consistent with cURL in terms of `--cert` and `--key`.....that may have been what I intended in the first place.

exarkun: The OpenSSL CLI actually uses `-CAfile`.  There are at least three things wrong with that convention so let's not emulate it.

adamtheturtle:
So my options are:

* Use no options, by default I get `$PWD/user.crt`, `$PWD/user.key`, `$PWD/cluster.crt`
* `--certificates-directory` which gets `$DIRECTORY/user.crt` etc
* Specifying the `.crt` files etc

Mixing the last two options should be disallowed, maybe. What do you think?

flocker/cli/script.py
```diff
@@ -65,6 +65,12 @@ class DeployOptions(Options):
     optParameters = [
         ["port", "p", REST_API_PORT,
          "The REST API port on the server.", int],
+        ["cafile", None, b"cluster.crt",
```
adamtheturtle: Full stops on descriptions (I don't mind either way as long as it is consistent, and the existing ones have full stops).

flocker/cli/script.py
```diff
@@ -108,11 +114,28 @@ def parseArgs(self, control_host, deployment_config, application_config):
...
+        self["cafile"] = FilePath(self["cafile"])
+        self["cert"] = FilePath(self["cert"])
+        self["key"] = FilePath(self["key"])
+
         if self["certificates-directory"] is None:
             self["certificates-directory"] = FilePath(os.getcwd())
         else:
             self["certificates-directory"] = FilePath(
                 self["certificates-directory"])
+            self["cafile"] = self["certificates-directory"].child(
```
adamtheturtle: :question: I'll test this but it seems like odd behaviour if I specify a certificates directory and a full path to a file in another directory that it'd strip the path apart from the basename.

dwgebler:
> Mixing the last two options should be disallowed, maybe. What do you think?

I originally took the `--certificates-directory` out when I added the other options, but @itamarst said I should retain the backwards compatibility. We could make the CLI throw an error if you try to mix those file path options with the directory option, but in this pass I made it so that both the original behaviour was retained and that if you specify the directory option and file name options, it looks for those file names in the specified directory path. So, e.g. `flocker-deploy --certificates-directory=/home/whatever/credentials/ --cafile=mycluster.crt` looks for `/home/whatever/credentials/mycluster.crt`

I don't have a strong opinion on which is more.....reasonable. The potential for confusion at the moment is that `flocker-deploy --certificates-directory=/home/whatever/credentials/ --cafile=/somethingelse/mycluster.crt` would still look for `/home/whatever/credentials/mycluster.crt` - but why would anyone try and use the options in that way? Certainly nothing in our docs suggests you can or should combine those options. That said, such confusion would be avoided by refusing to act if both `--certificates-directory` AND any of `--cacert`, `--cert` or `--key` are specified.

flocker/ca/_ca.py
```diff
@@ -214,44 +214,67 @@ def flocker_keypair():
...
+def load_certificate_from_path(path, key_filename, cert_filename):
+    """
+    Load a certificate and keypair from a specified path.
+
+    :param FilePath path: Directory where certificate and key files
+        are stored.
+    :param bytes key_filename: The file name of the private key.
+    :param bytes cert_filename: The file name of the certificate.
+
+    :return: A ``tuple`` containing the loaded key and certificate
```
adamtheturtle: :question: I'm not sure that we have a "policy" on this but the rest of the codebase but we seem to not use `tuple`s and instead use data structures like `CertAndKey`. Maybe that is a question for someone else.

flocker/ca/_ca.py
```diff
@@ -332,6 +355,36 @@ class UserCredential(PRecord):
     username = field(mandatory=True, type=unicode)
 
     @classmethod
+    def from_files(cls, certificate_path, key_path):
```
adamtheturtle: :question: I'll look into this more but first impression is why is this better than `from_path` - the docstring seems to say that it is because this loads "a user certificate and keypair" and `from_path` loads `a user certificate`. If that is the difference then either something should be moved or something should be renamed.

flocker/ca/_ca.py
dwgebler: We do return tuples in some places. Note that the return value of this method is not a behaviour that changes in this branch.

flocker/ca/_ca.py
dwgebler: `from_files` can take different file paths for the certificate and the key, `from_path` takes one directory path and expects to find `<username>.crt` and `<username.key>` inside that path - the latter being the original behaviour of the various `Credential` classes, which is still retained because it's still used, but not what we want for `flocker-deploy` anymore. There is a separate issue mentioned in a comment in the code changes in this PR to later and more generally improve the restricted `from_path` behaviour present throughout the credential classes.

docs/using/config/usage.rst
adamtheturtle: > don't know about OS X but I'd imagine it's the same

You imagine correctly.

flocker/ca/_ca.py
adamtheturtle: At least change the method names / docstrings to present this information more clearly.

docs/using/config/usage.rst
adamtheturtle: It may be better to replace this whole block with something like "X is the default, see `--help` for information on using custom paths".

adamtheturtle:
Thanks for the considered responses, summary so far is that this looks pretty good. I'm not sure on best practices for backwards compatibility, but it seems a bit weird to have both options (as you suggested, maybe throw an error when both are used). I'd like to find out whether we are dropping `flocker-deploy` before spending too much time optimising this but otherwise address and resubmit for a probably quick final review.

adamtheturtle:
> but why would anyone try and use the options in that way?

I think we should assume yes, and at worst we've done a little extra work

docs/using/config/usage.rst
```diff
@@ -44,10 +44,11 @@ To authenticate with the control service, you will need a copy of the public clu
 
 For ``flocker-deploy``, your API user certificate and key should be in files named ``user.crt`` and ``user.key`` and the cluster certificate in file ``cluster.crt``.
 
-By default, ``flocker-deploy`` will look for these certificate files in the current working directory.
-If this is not where the files are located, you may specify the ``--certificates-directory`` option to ``flocker-deploy``:
+By default, ``flocker-deploy`` will look for these certificate files in the current working directory and expect them to be named :file:`cluster.crt` (the public cluster certificate), :file:`user.crt` (the API user certificate) and :file:`user.key` (the API user's private key).
```
jongiddy: You don't need both this paragraph and the previous one. I suggest deleting the previous one.

docs/using/config/usage.rst
```diff
@@ -44,10 +44,11 @@ To authenticate with the control service, you will need a copy of the public clu
...
+
+You can override any of these defaults with the ``--cacert`` (cluster certificate), ``--cert`` (user certificate) and ``--key`` (user private key) options, specifying the full path to each file.
```
jongiddy: Delete "any of" - it doesn't add anything.

flocker/ca/_ca.py
jongiddy: I'm OK with this function returning a tuple, but in the docstring please indicate the types inside the returned tuple.

flocker/ca/_ca.py
jongiddy: :question: Rename `from_files` to `from_credential_files`

flocker/cli/script.py
```diff
@@ -65,6 +65,12 @@ class DeployOptions(Options):
...
          "The REST API port on the server.", int],
+        ["cacert", None, None,
+         "Path to cluster certificate file."],
+        ["cert", None, None,
+         "Path to user certificate file."],
+        ["key", None, None,
+         "Path to user private key file."],
         ["certificates-directory", "c",
          None, ("Path to directory where TLS certificate and keys can be "
                 "found. Defaults to current directory.")],
```
jongiddy: :question: Not currently part of this patch, but please change this to "Path to directory containing TLS certificates and keys. Defaults to current directory."

jongiddy:
:question: I would actually prefer that the `--certificates-directory` option could be used with the other options, as follows:

The default directory for credentials is the value of `--certificates-directory`, or else the current working directory.

The value of the other flags can be an absolute path, or a path relative to the default directory. Defaults as currently, hence relative to the setting of the default directory.

However, seeing as this has been discussed in previous reviews, I am happy to accept the current state.

flocker/cli/script.py
```diff
@@ -108,11 +114,45 @@ def parseArgs(self, control_host, deployment_config, application_config):
...
+                self["cacert"].basename())
+            self["cert"] = self["certificates-directory"].child(
+                self["cert"].basename())
+            self["key"] = self["certificates-directory"].child(
+                self["key"].basename())
+
+        for credential in ["cacert", "cert", "key"]:
+            if not self[credential].isfile():
+                raise UsageError(
+                    "File " + self[credential].path + " does not exist."
```
jongiddy: :question: Add an additional sentence: 
```
"Use the flocker-ca command to create the credential, or use the --" + credential + " flag to specify the credential location."
```

jongiddy:
Tested by moving my keys and certificates around, and setting them using the command line flags.  Please address the remaining comments, and as long as there are no major code changes from addressing these comment, this is OK to merge.

adamtheturtle:
I have addressed @jongiddy 's comments and the tests pass, so I'm merging.

Issue #1366, opened 2015-05-04
Can't use hostnames in deployment yaml file.


gcharot:
Hi,

I tried to switch IP to hostnames in your mongoDB example. The IP version works fine however when I use hostnames, flocker fails to deploy the stack. It seems to be related to the iptables redirection rule.
My hostname resolution is working fine (via hosts file). 

The deployment yaml looks like :
````
"version": 1
"nodes":
  "flocker01": []
  "flocker02": ["mongodb-example"] 
````
The interesting part of the stacktrace :

````
iptables v1.4.21: Bad IP address "flocker02"

Try `iptables -h' or 'iptables --help' for more information.
Unhandled Error

...
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['iptables', '--table', 'nat', '--append', 'PREROUTING', '--protocol', 'tcp', '--destination-port', '27017', '--match', 'addrtype', '--dst-type', 'LOCAL', '--match', 'comment', '--comment', 'flocker create_proxy_to', '--jump', 'DNAT', '--to-destination', 'flocker02']' returned non-zero exit status 2
Unhandled Error
...
````
As far as i known you can't specify hostname after the --to-destination option. Trying to manually run the iptables gives the same error :

````
# iptables --table nat --append PREROUTING --protocol tcp --destination-port 27017 --match addrtype --dst-type LOCAL --match comment  --comment "flocker create_proxy_to" --jump DNAT --to-destination flocker02
iptables v1.4.21: Bad IP address "flocker02"
````

Are hostname supported ? Using IPs is fine for testing but can become quite problematic in production. I agree that using hostnames with Iptables rules is not really advised for performances but you could do a gethostname before the rule insertion which is fine as long as the DNS reccord doesn't change over time... 

If you decide not to support hostname, it might be a good idea to check if the value is an IP (before SSHing to the nodes) and if not raise an "human readable" error to the user.

Cheers,
Greg

itamarst:
1. For more production-oriented use we have a new HTTP API for interacting with the cluster; in next release it will use stable UUIDs for identifying nodes rather than IPs (http://doc-dev.clusterhq.com/advanced/api.html#api-details).

2. `flocker-deploy` will continue to use IPs until we have a replacement client, so we'll probably wait to see what that UI looks like before adding this.

itamarst:
Or rather, `flocker-deploy` won't use UUIDs but might be extended to use hostnames (or the new command line client might).

[...]

PR #1676, opened 2015-06-18
[FLOC-2422] Flocker deploy ip addresses

dwgebler:
Fixes FLOC-2422


dwgebler:
I suppose another possibility is for the control service to attempt to resolve a hostname given in a deployment config to an IP (e.g. `socket.gethostbyname_ex`) and only return an error if it can't, but that's just a random thought, nothing to do with the scope of this issue.

exarkun:
This also fixes #1366 1366

docs/using/config/configuration.rst
```diff
@@ -300,9 +300,9 @@ Deployment Configuration
 ------------------------
 
 The deployment configuration specifies which applications are run on what nodes.
-It consists of a version and a mapping from node names to application names.
+It consists of a version and a mapping from node IP addresses to application names.
```
wallrj: "IPv4 address"

docs/using/config/configuration.rst
```diff
@@ -311,7 +311,7 @@ Here's an example of a simple but complete configuration defining a deployment o
 
   "version": 1
   "nodes":
-    "node017.example.com":
+    "172.16.255.251":
```
wallrj: :question: Maybe use an IPv4 address reserved for documentation:
 * https://tools.ietf.org/html/rfc5737#section-3

flocker/control/_config.py
```diff
@@ -1259,6 +1261,13 @@ def deployment_from_configuration(deployment_state, deployment_configuration,
     seen_applications = set()
     for hostname, application_names in (
             deployment_configuration['nodes'].items()):
+        try:
+            IPv4Address(hostname)
+        except AddressValueError:
+            raise ConfigurationError(
+                ("Node {node_name} specified, but deployment configuration "
+                 "expects an IPv4 address.").format(node_name=hostname)
+            )
```
wallrj: :question: Maybe quote the incorrect hostname or it ends up looking like this:
```
(2422)[~/flocker-tutorial]$ flocker-deploy --certificates-directory=/tmp/tmpqgFZ7D 119.9.108.69 volume-deployment.yml volume-application.yml 
Node node2.example.com specified, but deployment configuration expects an IPv4 address.
```

wallrj:
Thanks @dwgebler 

This seems to work.

And the code makes sense.

There's an acceptance test build failure, which I doubt is related to this branch, but there are no remote_logs so I can't tell:
 * http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fubuntu-14.04%2Faws/builds/896/steps/trial/logs/remote_logs.log

 I think we're not currently merging anything unless all builds are green.

So you might want to force another build.

Merge if you can get green build results.

wallrj:
Oh, and remember to close https://github.com/ClusterHQ/flocker/issues/1366 once you've merged this.

PR #1680, opened 2015-06-18
Flocker labsdoc floc 2032

robhaswell:


robhaswell:
I merge what I must, because I can.

PR #1682, opened 2015-06-18
Revert "[FLOC-2372] Remember discovered device path in EBSBlockDevice…

exarkun:
Reverts ClusterHQ/flocker#1659

For 1.0.1 we're going with a different implementation strategy to resolve this bug.  To minimize the size of the change, this change should be backed out because it will not be required for the alternate strategy.


exarkun:
exarkun: okay git diff `origin/release/flocker-1.0.0^^^..` is clean
itamar: merge then

PR #1684, opened 2015-06-18
Release maintenance/flocker 1.0.0/ebs device name mapping floc 2329

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2329

This is an entirely different approach to fixing the issue than was taken previously.  All of the filesystem UUID changes are gone.  This instead attempts to change the EBS storage driver so that it can reliably respond to `get_device_path` requests.

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -37,7 +37,6 @@ from ._logging import (
 DATASET_ID_LABEL = u'flocker-dataset-id'
 METADATA_VERSION_LABEL = u'flocker-metadata-version'
 CLUSTER_ID_LABEL = u'flocker-cluster-id'
-ATTACHED_DEVICE_LABEL = u'attached-device-name'
 BOTO_NUM_RETRIES = u'20'
 VOLUME_STATE_CHANGE_TIMEOUT = 300
 MAX_ATTACH_RETRIES = 3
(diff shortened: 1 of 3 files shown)


flocker/node/agents/ebs.py
```diff
@@ -71,6 +70,48 @@ def _enable_boto_logging():
...
+
+    __repr__ = __str__
+
+
+def _expected_device(requested_device):
+    """
+    Given a device we requested from AWS EBS, determine the OS device path that
+    will actually be created.
+
+    This maps EBS required ``/dev/sdX`` names to ``/dev/vbdX`` names that are
```
myechuri: Did you mean "This maps EBS required ``/dev/sdX`` names to ``/dev/xvdX`` names that are" ? Thanks!

flocker/node/agents/ebs.py
```diff
@@ -71,6 +70,48 @@ def _enable_boto_logging():
...
+    used by currently supported platforms (Ubuntu 14.04 and CentOS 7).
```
itamarst: Document the param.

flocker/node/agents/ebs.py
```diff
@@ -559,20 +603,33 @@ def attach_volume(self, blockdevice_id, attach_to):
                     # UserGuide/device_naming.html), wait for new block device
                     # to be available to the OS, and interpret it as ours.
                     # Wait under lock scope to reduce false positives.
-                    new_device = _wait_for_new_device(blockdevices,
-                                                      volume.size)
+                    device_path = _wait_for_new_device(
```
itamarst: I was kinda hoping to just delete the whole code path with the lock, since it could erroneously notice legit EBS devices attached by some other system. But I guess that failure mode would now mostly just result in a retry... so OK that seems fine for now.

flocker/node/agents/ebs.py
```diff
@@ -559,20 +603,33 @@ def attach_volume(self, blockdevice_id, attach_to):
...
+                    # a certain simple pattern.  Verify that now and signal an
+                    # error immediately if the assumption is violated.  If we
+                    # let it go by, a later call to ``get_device_path`` will
+                    # quietly produce the wrong results.
+                    #
+                    # To make this explicit, we *expect* that the device will
+                    # *always* be what we *expect* the device to be (sorry).
+                    # This check is only here in case we're wrong to make the
+                    # system fail in a less damaging way.
+                    if _expected_device(device) != device_path:
```
itamarst: These two variables are two confusingly similar.

flocker/node/agents/ebs.py
```diff
@@ -71,6 +70,48 @@ def _enable_boto_logging():
...
+
+    def __str__(self):
+        return self._template.format(
+            self.requested.path, self.discovered.path,
+        )
+
+    __repr__ = __str__
+
+
+def _expected_device(requested_device):
```
itamarst: `_expected_local_device` maybe.

itamarst:
The big thing this is missing is the upgrade docs and the what's new stuff. We still need those.

flocker/node/agents/ebs.py
itamarst: Ah, but that's existing issue. So some other time.

PR #1685, opened 2015-06-19
[FLOC-2456] Some news entries and documentation for 1.0.1

itamarst:
I have not actually previewed the result, please review carefully!

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,9 @@
+Flocker 1.0.1 (2015-06-XX)
+==========================
+
+- Fixed a bug where EBS volumes were mapped to the wrong device and therefore mounted in the wrong location. (FLOC-2329).
+- Added additional safety checks when formatting devices. (FLOC-2356)
+
 Flocker 1.0.0 (2015-06-15)
 ==========================
 
--- a/NEWS
+++ b/NEWS
@@ -1,4 +1,4 @@
-Flocker 1.0.1 (2015-06-XX)
+Flocker 1.0.1 (2015-06-19)
 ==========================
 
 - Fixed a bug where EBS volumes were mapped to the wrong device and therefore mounted in the wrong location. (FLOC-2329).
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -9,6 +9,14 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+v1.0.1
+======
+
+Upgrading is strongly recommended for all users of 1.0.0.
+
+* The EBS storage driver now more reliably selects the correct OS device file corresponding to an EBS volume being used.
+* Additional safety checks were added to ensure only empty volumes are formatted.
+
 v1.0
 ====
 
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -12,7 +12,7 @@ You can learn more about where we might be going with future releases by:
 v1.0.1
 ======
 
-Upgrading is strongly recommended for all users of 1.0.0.
+Upgrading is strongly recommended for all users of v1.0.0.
 
 * The EBS storage driver now more reliably selects the correct OS device file corresponding to an EBS volume being used.
 * Additional safety checks were added to ensure only empty volumes are formatted.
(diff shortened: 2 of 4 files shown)


NEWS
```diff
@@ -1,3 +1,9 @@
+Flocker 1.0.1 (2015-06-XX)
```
myechuri: Date needs to be changed from "2015-06-XX" to "2015-06-19" before merging.

docs/releasenotes/index.rst
```diff
@@ -9,6 +9,14 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+v1.0.1
+======
+
+Upgrading is strongly recommended for all users of 1.0.0.
```
davidoliverSP2: This is generally a question for the whole of the docs - do we refer to v1 as v1.0 or v1.0.0 - we use both on this page. 

Also, I would recommend always using the "v"

docs/releasenotes/index.rst
```diff
@@ -9,6 +9,14 @@ You can learn more about where we might be going with future releases by:
...
+
+* The EBS storage driver now more reliably selects the correct OS device file corresponding to an EBS volume being used.
```
davidoliverSP2: this sentence reads a bit awkwardly - maybe we can reword this to something like this:

The EBS storage driver is now more reliable when selecting the correct OS device file which corresponds with an EBS volume that is being used.

Also, is there a way to define "more reliably"? this is a bit vague.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,34 @@
+=========
+Upgrading
+=========
+
+A Flocker cluster can be upgraded to a newer version of Flocker while preserving all data and configuration of the Flocker cluster provided certain steps are followed.
+The correct steps to follow may vary depending on the versions of Flocker being upgraded from and to.
+A common requirement for all currently supported upgrade paths is that all nodes and the control service must be running the same version of Flocker.
```
davidoliverSP2: This sounds like an important piece of information related to upgrading. Please move this down a line so it starts a new paragraph.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,34 @@
...
+
+Flocker 1.0.1
```
davidoliverSP2: As previously mentioned, my recommendation would to use Flocker v1.0.1 and Flocker v1.0.0 when referring to the product version. @ferrantim - do you have a view on this?

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,34 @@
...
+-------------
+
+Recommended Steps
+^^^^^^^^^^^^^^^^^
+
+#. Stop the agent services on all nodes, and then stop control service.
+#. Install Flocker 1.0.1 on all nodes in the Flocker cluster.
+#. Restart the control service.
+#. If you are using the EBS storage backend reboot each of the agent nodes.
```
davidoliverSP2: missing comma:
If you are using the EBS storage backend, reboot each of the agent nodes.

docs/using/administering/upgrading.rst
```diff
@@ -0,0 +1,34 @@
...
+#. If you have not configured the Flocker agents to start automatically on boot,
+   restart the agent services on all nodes.
+
+Details
+^^^^^^^
+
+The upgrade to Flocker 1.0.1 involves changing the way the EBS storage backend maps volumes to devices: in version 1.0.0, there were occasional errors in this mapping.
+As a result, some devices may be mounted in the wrong location.
```
davidoliverSP2: Small change:
As a result, some devices may have been mounted in the wrong location.

davidoliverSP2:
Thanks @itamarst - only a few small recommendations from me, although we probably want a consistent way to refer to the version to use throughout the docs. Hopefully @ferrantim can provide some guidance for us.

docs/releasenotes/index.rst
wallrj: Done.

docs/releasenotes/index.rst
wallrj: I'll leave it I think.

docs/using/administering/upgrading.rst
wallrj: I've made it into a note.

docs/using/administering/upgrading.rst
wallrj: Changed to v1.0.1

docs/using/administering/upgrading.rst
wallrj: Done.

docs/using/administering/upgrading.rst
wallrj: Done.

wallrj:
Looks good. 

I addressed most of @davidoliverSP2 's comments; but not all.

I'll merge as soon as the docs builder completes.

PR #1696, opened 2015-06-20
Release maintenance/flocker 1.0.1pre1/disable restart policies floc 2467

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2467

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -9,6 +9,12 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+v1.0.2
+======
+
+* The "on-failure" and "always" restart policies for containers have been disabled.
+  See :ref:`restart configuration`.
+
 v1.0.1
 ======
 
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -13,7 +13,7 @@ v1.0.2
 ======
 
 * The "on-failure" and "always" restart policies for containers have been disabled.
-  See :ref:`restart configuration`.
+  See :ref:`restart policy<restart configuration>`.
 
 v1.0.1
 ======
--- a/docs/using/config/configuration.rst
+++ b/docs/using/config/configuration.rst
@@ -134,6 +134,9 @@ The following parameters are optional when defining an application:
 
   This specifies the restart policy for this Application.
   There must be a ``name`` sub-key whose value may be one of: ``never``, ``always``, or ``on-failure``.
+  **The supplied policy is recorded but not not used.
+  Containers will always behave as though created with the ``never`` restart policy.
+  It is recommended that all containers be created with this policy.**
   The ``on-failure`` restart policy accepts an optional ``maximum_retry_count`` sub-key, which specifies how many times Docker will attempt to restart the application container in the event of repeated failures.
   See the `Docker Restart Policy reference <https://docs.docker.com/reference/commandline/cli/#restart-policies>`_ for more information on restart policies.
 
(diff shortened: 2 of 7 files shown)


docs/releasenotes/index.rst
```diff
@@ -9,6 +9,12 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+v1.0.2
+======
+
+* The "on-failure" and "always" restart policies for containers have been disabled.
```
itamarst: I think this should have same text as the NEWS, explaining why and that it's temporary.

docs/using/config/configuration.rst
```diff
@@ -134,6 +134,9 @@ The following parameters are optional when defining an application:
 
   This specifies the restart policy for this Application.
   There must be a ``name`` sub-key whose value may be one of: ``never``, ``always``, or ``on-failure``.
+  **The supplied policy is recorded but not not used.
```
itamarst: Maybe add link to issue for re-enabling this? Public issue tracker, may as well use that.

docs/releasenotes/index.rst
itamarst: ANd maybe link to issue here too, so people who are interested can track progress instead of maybe filing new issues saying "I need this".

flocker/node/_deploy.py
```diff
@@ -934,6 +935,13 @@ def _restart_for_application_change(
         # we don't want to do anything:
         comparable_state = comparable_state.transform(["running"], True)
 
+        # Restart policies were briefly supported but they interact poorly with
```
itamarst: This comment is wrong, as is implementation I think. Given this patch Flocker will *not* change any existing running container; rather, it will continuously restart it.

1. Container X is configured with `always`.
2. Actual container may be `always`, may be new one with `never.
3. In either case, this function will indicate state is `never.
4. `always != never` and so container will be restarted (repeatedly).

What you want to change is the *configuration* to be `never` in this function. Then:

* Existing containers with policy `always` or `on-failure` in their state will be restarted as `never` containers, thus preventing problems on restart.
* Containers with state `never` will always compare equally to the configuration and therefore not getting restarted.
* When we re-introduce restarts the line can be deleted and the containers will be appropriately restarted.

In short: transform `comparable_configuration`, not `comparable_state`.

flocker/node/_deploy.py
```diff
@@ -171,7 +171,8 @@ def run(self, deployer):
             volumes=volumes,
             mem_limit=application.memory_limit,
             cpu_shares=application.cpu_shares,
-            restart_policy=application.restart_policy,
+            # The only supported policy is "never".  See FLOC-2449.
+            restart_policy=RestartNever(),
```
itamarst: This bit looks right.

itamarst:
Besides the comment above, merge is failing (?!). Merge forward, address, resubmit.

itamarst:
Oh, branch name is wrong. We really need a better policy...

flocker/node/_deploy.py
exarkun: Not sure what I was thinking.  Possibly I got confused between which of the three variations of the solution I was working on. :/  I think your analysis is correct with one minor correction: I also changed `StartApplication` to start applications with `RestartNever()` as the policy regardless of configuration... So this change accidentally succeeds in only restarting the applications once.

The suggested fix makes the code make more sense.

exarkun: And fixes the case where the application is running with "always" restart policy but the configuration is changed to "never" simultaneously with the Flocker upgrade.

exarkun: Aaaahhh restart policies aren't comparable in any meaningful way.

PR #1697, opened 2015-06-21
Make tox runs faster

achanda:
- Skip creating a source distribution. This takes time and is not
  necessary.
- Do not install distribute
- Require version 1.6 which added support for these flags

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,8 @@
 [tox]
 envlist = pypy, py27, lint, sphinx
+distribute = False
+minversion = 1.6
+skipsdist = True
 
 [testenv]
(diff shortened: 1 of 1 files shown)


lukemarsden:
Thank you for your contribution! We will take a look at this as soon as possible and get back to you.

itamarst:
While testing this I discovered that our `tox` support has been broken by a recent change. So I'll probably go fix that and maybe combine these improvements into that branch.

itamarst:
Some of this made it into #1821 - thanks for the PR! If you email thanks@clusterhq.com we'd be happy to send you a t-shirt.

PR #1701, opened 2015-06-22
[FLOC-2485] Font size should be set once, not on nested sections

msh100:
Fixes FLOC-2485


adamtheturtle:
Also fixes https://clusterhq.atlassian.net/browse/FLOC-2241.

adamtheturtle:
I don't understand the code but from looking at `/using/installing/index.html#what-to-do-next` the font sizes look fine, so I'm merging. Thanks @msh100 .

PR #1702, opened 2015-06-22
[FLOC-2484] Make padding top and bottom the same on headers

msh100:
Fixes FLOC-2484

### DIFF
--- a/docs/_themes/clusterhq/static/css/docs.css
+++ b/docs/_themes/clusterhq/static/css/docs.css
@@ -83,6 +83,7 @@ li.button.menu a:hover {
 h1, h2, h3, h4, h5 {
     font-family:'Soho Gothic W01 Medium';
     padding-bottom: 25px;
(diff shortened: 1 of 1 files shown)


adamtheturtle:
To confirm, this is because A is longer than B, right?

screenshot 2015-06-22 14 02 56

adamtheturtle:
If so I can't tell, but then the rest of the theme hasn't been through such scrutiny - the code is fine and the header looks fine to me and I'll leave it to your judgement. If the issue is something else then resubmit.

msh100:
Thanks @adamtheturtle, I think you're looking at the wrong part of the page. This is likely due to a poor description in JIRA.

http://i.imgur.com/pT5HyuT.png

On the left is the new version, the right is what we currently have.

The change is to increase the top padding for content headers.

adamtheturtle:
Ok, same comment applies, thanks. Merging.

PR #1703, opened 2015-06-22
[FLOC-2428] Release maintenance/flocker 1.0.1/remove logging

jongiddy:

PR #1704, opened 2015-06-22
[FLOC-2427] Remove logging

jongiddy:
This is the PR for `master` of the existing PR #1703 that has been merged into `release/flocker-1.0.1`


adamtheturtle:
Eyeballing shows that this is the same as the diff which was accepted in https://github.com/ClusterHQ/flocker/pull/1675/files. Merge if you're happy with the build results.

PR #1706, opened 2015-06-22
Increase the pyasn1 version

adamtheturtle:


adamtheturtle:
This is against master, does it need to go against a release branch? If so, which one?

setup.py
```diff
@@ -51,7 +51,7 @@
 
     # Some of the tests use Conch:
     "PyCrypto==2.6.1",
-    "pyasn1==0.1.7",
+    "pyasn1==0.1.8",
```
exarkun: I think a safer change would be to pin the version of pyasn1_modules to one that's compatible with pyasn1.

exarkun: Also easier to review since it's the change that means "keep using the software we've been using" instead of one that means "upgrade to some new software".

exarkun: Also avoids this problem happening again as soon as pyasn1 is releases again.

PR #1707, opened 2015-06-22
Build vagrant basebox floc 2494

Azulinho:
Add a new Vagrantfile for 'basebox' virtualbox VM
Update the Vagrantfile used by the buildbot job flocker-vagrant-dev to use the new basebox
disable tox tests using pypy

The first run of the modified flocker-vagrant-dev will fail until the PR https://github.com/ClusterHQ/build.clusterhq.com/pull/108 is merged and built once.

JIRA https://clusterhq.atlassian.net/browse/FLOC-2494

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = pypy, py27, lint, sphinx
+envlist = py27, lint, sphinx
 
 [testenv]
 # We want modern pip so we can install packages that only have wheels:
@@ -13,9 +13,6 @@ commands =
     trial --rterrors {posargs:flocker}
 setenv =
     PYTHONHASHSEED=random
-
-[testenv:pypy]
-basepython = pypy
 
 [testenv:py27]
 basepython = python2.7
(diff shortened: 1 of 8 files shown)


tomprince:
It isn't clear to me how this is different from the existing dev-box? It looks like the dev-box is now just a repackaging of the basebox.

Azulinho:
dev-box = basebox + tox tests.
But the build of dev-box doesn't build a vagrant image, simply consumes the most recent basebox image built.

tomprince:
Well, then why call it the base-box and not the dev-box?

tomprince:
That is, all we need to do is change when the dev-box gets built, not rename the dev-box and change when it is built.

Azulinho:
if we do that, then we would be discarding all the tox tests that get executed. if they dont provide any value (cant answer if they do or not) then it may be simpler not to run them at all.

tomprince:
> if we do that, then we would be discarding all the tox tests that get executed.

I don't follow. The name we give the box doesn't have anything to do with running the tests. Note that the tests aren't run by building the dev-box, but as a separate step afterwards.

Azulinho:
oh sorry, I get what you are saying now.
yeap, you're right, no reason it is called basebox, can look into it so that it uses the vgtfile from the existing dev-box

PR #1713, opened 2015-06-23
pin pyasn1-modules

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2506

See #1706 for comments on the approach.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,11 @@ dev_requirements = [
     # Some of the tests use Conch:
     "PyCrypto==2.6.1",
     "pyasn1==0.1.7",
+    # pyasn1-modules is tightly coupled to a certain version of pyasn1.
+    # Letting its version float while pyasn1 is pinned breaks things.  This is
+    # just the version of pyasn1-modules that works with the version of pyasn1
+    # pinned above.
+    "pyasn1-modules==0.0.5",
 
     # The acceptance tests interact with MongoDB
     "pymongo>=2.7.2",


adamtheturtle:
Thanks @exarkun - this seems like an improvement to me. However, I am wary of @tomprince 's concerns that pinning security modules is a risk. Maybe we should look at https://clusterhq.atlassian.net/browse/FLOC-1172 soon.

adamtheturtle:
Merge when you're happy (I guess now, this is a test fixer so it can be merged, right?).

exarkun:
> However, I am wary of @tomprince 's concerns that pinning security modules is a risk.

In general I agree with this concern.  So yea, we should do something better.  For this particular case, I'm not as concerned because our use of pyasn1 is in our build infrastructure only.

exarkun:
I am unhappy with the outcome of the builds:

  * installed-package/vagrant/centos-7 - hung, known issue, https://clusterhq.atlassian.net/browse/FLOC-2503
  *  acceptance/vagrant/centos-7/zfs
    * failed container tests, known issue, https://clusterhq.atlassian.net/browse/FLOC-2488
    * failed postgres tests, known issue (can't find the issue but I know it's filed)
  * Rackspace builders - Resource limits exceeded, known issue, https://clusterhq.atlassian.net/browse/FLOC-2441
  * acceptance/aws/ubuntu-14.04/aws - Resource limits exceeded, known issue, https://clusterhq.atlassian.net/browse/FLOC-2441

So I guess I'll merge.

PR #1714, opened 2015-06-23
[FLOC-2487] Pin pyasn1-modules

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2487

Forward-port of #1713


adamtheturtle:
Accepted on the same conditions as https://github.com/ClusterHQ/flocker/pull/1713. However this should be against a sub-task, not the main issue as merging this will mark the main issue as done even if the sub-task is outstanding.

PR #1715, opened 2015-06-23
test_exception_aborts_write: the writer sometimes can get IOError/EPIPE

avg-I:
The previous code had 2 problems and each of them ensured that the test
would pass (except for possible EPIPE) even if the tested code didn't behave as expected.

- previously the test would discard the first byte of the stream, so the
  receiving side could know that the stream is corrupt just from reading
  its header and thus it wouldn't need to read the rest of the stream
  and so the exception would change little
- even if a completely valid stream were sent then the receiving side would
  still reject it because 'snapshots' parameter was not passed to the
  sending side and, thus, the sending side would generate a full stream
  for a filesystem that already exists and that would be rejected

So now we make sure that an incremental stream is sent.
Also we send all of the stream except the last byte, so that the receiving
side sees a completely valid stream until the exception is raised and the
stream is aborted.  Note that if we sent a full stream then it would be
possible that the receiving side would apply it before the exception is
raised.

See https://clusterhq.atlassian.net/browse/FLOC-2507

### DIFF
--- a/flocker/volume/test/filesystemtests.py
+++ b/flocker/volume/test/filesystemtests.py
@@ -614,16 +614,23 @@ def make_istoragepool_tests(fixture, snapshot_factory):
                 path.child(b"anotherfile").setContent(b"hello")
 
                 to_filesystem = volume2.get_filesystem()
-                try:
-                    with from_filesystem.reader() as reader:
-                        with to_filesystem.writer() as writer:
-                            data = reader.read()
-                            writer.write(data[1:])
-                            raise ZeroDivisionError()
-                except ZeroDivisionError:
-                    pass
-                to_path = volume2.get_filesystem().get_path()
-                self.assertFalse(to_path.child(b"anotherfile").exists())
+                getting_snapshots = to_filesystem.snapshots()
+
+                def got_snapshots(snapshots):
+                    try:
+                        with from_filesystem.reader(snapshots) as reader:
+                            with to_filesystem.writer() as writer:
+                                data = reader.read()
+                                writer.write(data[:-1])
+                                raise ZeroDivisionError()
+                    except ZeroDivisionError:
+                        pass
+                    to_path = volume2.get_filesystem().get_path()
+                    self.assertFalse(to_path.child(b"anotherfile").exists())
+
+                getting_snapshots.addCallback(got_snapshots)
+                return getting_snapshots
+
             d.addCallback(got_volumes)
             return d
 


flocker/volume/test/filesystemtests.py
```diff
@@ -618,7 +620,11 @@ def got_volumes(copied):
                     with from_filesystem.reader() as reader:
                         with to_filesystem.writer() as writer:
                             data = reader.read()
-                            writer.write(data[1:])
+                            try:
+                                writer.write(data[1:])
+                            except IOError as e:
+                                if e.errno != errno.EPIPE:
```
jml: Please add a comment explaining why this is not a real error. 

Suggestion:

```python
# Sometimes the reader will quit before the writer is done writing. This is OK because ____.
```

jml:
There are test failures in several builders:
* vagrant/build/tutorial
* acceptance/rackspace/centos-7/rackspace
* acceptance/rackspace/ubuntu-14.04/rackspace
* osx-10.10
* functional/rackspace/centos-7/storage-driver

Merging master might fix some of them. 

Please make sure tests are passing on all builders before landing.

flocker/volume/test/filesystemtests.py
avg-I: Now that I've tried to formulate the suggested comment I am wondering why we write `data[1:]` here, that is, why we skip the first byte... That makes the stream invalid and, thus, a reader that works asynchronously may error out  after seeing the corrupted stream head and won't wait for the rest of the stream. This is what causes `EPIPE` if the writer is not fast enough.
From the test description it is not clear why we would want to write an invalid stream.
Perhaps the intention was to write `data[:-1]` to ensure that the reader does not get a complete stream before the exception happens? @itamarst ?

avg-I:
@jml I've rebased the branch on top of the latest master, but there are still some failures. However, it seems that all of them are because of failures to clone from github.

flocker/volume/test/filesystemtests.py
itamarst: Oh god can we just throw away this code.

itamarst: I have no idea what this code is doing. And I don't particularly care about it either, we're going to throw it away and replace the whole thing with something that isn't a horrible hack. So just do what you think best, it doesn't really matter.

avg-I: I understand the situation but as long as the test (and the API that it tests) remains in the current form, and I think that that could be for a while, I would like the test to test the right thing.

PR #1716, opened 2015-06-23
zfs test_maximum_size_enforced: flush the data to the kernel

avg-I:
If that is not explicitly done then some data may linger in userland
until the file is closed and so the error would be reported outside
the scope where we expect it.

See https://clusterhq.atlassian.net/browse/FLOC-2508


itamarst:
Merge when buildbot is happy (see discussion on how we're not merging until everything is green).

PR #1717, opened 2015-06-23
[FLOC-2492] Linkcheck

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2492

This changes redirected links, and then ignores links which are:
* Examples, to show a structure (e.g. a GitHub comparison)
* Password protected
* Failing because of https://clusterhq.atlassian.net/browse/FLOC-1156

FLOC-1156 is a follow up to this.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -130,8 +130,6 @@ Once your pull request is merged, as a small thank you for contributing to Flock
 Just send an email to thankyou@clusterhq.com with your t-shirt size, mailing address and a phone number to be used only for filling out the shipping form.
 We'll get something in the mail to you.
 
-.. _JIRA: https://clusterhq.atlassian.net
-
 
 Merge Requirements
 ^^^^^^^^^^^^^^^^^^
@@ -165,7 +163,7 @@ While we're happy to look at contributions in any state as GitHub PRs, the requi
 Project Development Process
 ===========================
 
-The core development team uses a `JIRA workflow`_ to track planned work.
+The core development team uses a `JIRA`_ workflow to track planned work.
 Issues are organized by sprints, and can reside in various states:
 
 Backlog
@@ -198,8 +196,6 @@ Done
     The issue has been closed.
     Some final work may remain to address review comments; once this is done and the branch is merged the GitHub PR will be closed.
 
-.. _JIRA workflow: https://clusterhq.atlassian.net/secure/Dashboard.jspa
-
 
 .. _reporting-security-issues:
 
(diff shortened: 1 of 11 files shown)


CONTRIBUTING.rst
```diff
@@ -41,7 +41,7 @@ Talk to Us
 
 Have questions or need help?
 
-* If you want to follow our development plans, our main issue tracker is https://clusterhq.atlassian.net.
+* If you want to follow our development plans, our main issue tracker is https://clusterhq.atlassian.net/secure/Dashboard.jspa.
```
exarkun: I'm not sure I like this change.  `https://clusterhq.atlassian.net/` seems like a better canonical URL to refer people to.  For all I know, Atlassian will change the redirect to point to `Dashboard.jspx` tomorrow and this will become a 404.

I take it there's no easy way to make the link checker allow URLs that are 302s?

CONTRIBUTING.rst
```diff
@@ -213,3 +209,5 @@ Flocker bugs should normally be :ref:`reported publicly<talk-to-us>`, but due to
 Instead, if you believe you have found something in Flocker (or any other ClusterHQ software) which has security implications, please send a description of the issue via email to security@clusterhq.com.
 Your message will be forwarded to the ClusterHQ security team (a small group of trusted developers) for triage and it will not be publicly readable.
 Once you have submitted an issue via email, you should receive an acknowledgment from a member of the security team within 48 hours, and depending on the action to be taken, you may receive further follow up emails.
+
+.. _JIRA: https://clusterhq.atlassian.net/secure/Dashboard.jspa
```
exarkun: Also, the duplication of this URL throughout the documentation is unfortunate.  Can we just spell this out once and then link to it from everywhere?

docs/labs/weave.rst
```diff
@@ -13,7 +13,7 @@ Installation
 ============
 
 Flocker and Weave both have Docker plugins.
-So you can simply install the :ref:`Flocker Docker plugin <labs-docker-plugin>` and install the `Weave Docker plugin <http://weave.works/docker-plugins>`_ and it should Just Work.
+So you can simply install the :ref:`Flocker Docker plugin <labs-docker-plugin>` and install the `Weave Docker plugin <http://weave.works/docker-plugins/index.html>`_ and it should Just Work.
```
exarkun: Fixes to labs docs, oh joy.

exarkun:
Thanks.  Please address inline comments to the extent reasonable.

This seems to make the doc builder green.  Is there a follow-up task to change the configuration so that the intermediate orange state is no longer allowed ( ... except for old release branches ... ) and, instead, any failed link checks produce a failed (red) build?  If not, can you file one?

Then go ahead and merge.  Thanks.

adamtheturtle:
> If not, can you file one?

Will do.

CONTRIBUTING.rst
adamtheturtle: Not that I can find.

CONTRIBUTING.rst
adamtheturtle: Not that I can see in an obvious way - what I have done is removed an instance of the duplication (it was unnecessarily twice in one document) so that the link is only in two places.

adamtheturtle:
> This seems to make the doc builder green. Is there a follow-up task to change the configuration so that the intermediate orange state is no longer allowed ( ... except for old release branches ... ) and, instead, any failed link checks produce a failed (red) build? If not, can you file one?

I created https://clusterhq.atlassian.net/browse/FLOC-2521 - however I'm not convinced that it is a good idea. It seems to have the same problem that the rejected LICENSE test - someone's branch will start failing even though they did nothing to cause it to fail (if an existing link breaks).

adamtheturtle:
I am merging because this makes the test suite green-er.

PR #1718, opened 2015-06-23
zfs test_maximum_size_enforced: test the right scenario and verify that

avg-I:
See https://clusterhq.atlassian.net/browse/FLOC-2509
Note that this pull request depends on #1716


avg-I:
Creating a proper pull request against master.

PR #1719, opened 2015-06-23
[FLOC-2512] Staging release

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2512

### DIFF
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -176,18 +176,9 @@ Preparing For a Release
       git config push.default current
       git push
 
-#. Go to the `BuildBot web status`_ and force a build on the just-created branch.
-
-#. Set up ``AWS Access Key ID`` and ``AWS Secret Access Key`` Amazon S3 credentials:
-
-   Creating the Vagrant machine attempts to copy the ``~/.aws`` configuration directory from the host machine.
-   This means that ``awscli`` may have correct defaults.
-
-   .. prompt:: bash [vagrant@localhost]$
-
-      aws configure
-
 #. Ensure all the required tests pass on BuildBot:
+
+   Pushing the branch in the previous step should have started a build on BuildBot.
 
    Unfortunately it is acceptable or expected for some tests to fail.
    Discuss with the team whether the release can continue given any failed tests.
@@ -273,6 +264,15 @@ Release
              Also, the package upload script currently expects the packages to be built from the tag, rather than the branch.
 
    Wait for the build to complete successfully.
+
+#. Set up ``AWS Access Key ID`` and ``AWS Secret Access Key`` Amazon S3 credentials:
+
+   Creating the Vagrant machine attempts to copy the ``~/.aws`` configuration directory from the host machine.
+   This means that ``awscli`` may have correct defaults.
+
+   .. prompt:: bash [vagrant@localhost]$
+
+      aws configure
 
 #. Publish artifacts and documentation:
 
--- a/docs/gettinginvolved/infrastructure/release-process.rst
+++ b/docs/gettinginvolved/infrastructure/release-process.rst
@@ -194,29 +194,6 @@ Preparing For a Release
    - Any ``docker-head`` builders.
    - Any builders in the "Expected failures" section.
 
-#. Set up ``AWS Access Key ID`` and ``AWS Secret Access Key`` Amazon S3 credentials:
-
-   Creating the Vagrant machine attempts to copy the ``~/.aws`` configuration directory from the host machine.
-   This means that ``awscli`` may have correct defaults.
-
-   .. prompt:: bash [vagrant@localhost]$
-
-      aws configure
-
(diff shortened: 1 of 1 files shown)


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -329,11 +313,4 @@ If there is no existing issue for the planned improvements then a new one should
 Look at `existing issues relating to the release process <https://clusterhq.atlassian.net/issues/?jql=labels%20%3D%20release_process%20AND%20status%20!%3D%20done>`_.
 The issue(s) for the planned improvements should be put into the next sprint.
 
-
-.. _wheel: https://pypi.python.org/pypi/wheel
-.. _Google cloud storage: https://console.developers.google.com/project/apps~hybridcluster-docker/storage/archive.clusterhq.com/
-.. _BuildBot web status: http://build.clusterhq.com/boxes-flocker
-.. _virtualenv: https://pypi.python.org/pypi/virtualenv
-.. _Homebrew: http://brew.sh
-.. _CloudFront: https://console.aws.amazon.com/cloudfront/home
```
adamtheturtle: I wanted to move the CloudFront link now it is only used once and then found that in fact many of the other links here are either no longer used or are only used once.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -290,6 +265,15 @@ Release
 
    Wait for the build to complete successfully.
 
+#. Set up ``AWS Access Key ID`` and ``AWS Secret Access Key`` Amazon S3 credentials:
```
adamtheturtle: I moved this because it only needs to be used at this point.

tomprince: I wonder if it makes sense to but per-requisite things like this in a section on its own at the very beginning. This step should only need to be done once per person.

adamtheturtle: This is done in the Vagrant box which is created at the start of each release process. Also, it is done once per person already sort-of: the next line explains that the `~/.aws` configuration is copied from the host if it exists, and the end of the release process copies it from the VM to the host.

tomprince: If it is copied automatically from the host system, then setting it up (on the host) is a step that can be done one time per host-machine, so it doesn't need to be a step in the release process.

tomprince: This could probably be done in a follow up though.

adamtheturtle: Doing it this way means that no software has to be installed on the host, but yes, this is a follow-up (https://clusterhq.atlassian.net/browse/FLOC-2514).

tomprince:
Other than that, this looks good.

jongiddy:
I don't think it is a good idea to merge this until some problems with the publish docs step have been resolved. See the JIRA issue https://clusterhq.atlassian.net/browse/FLOC-2512 for more info.

jongiddy:
Due to the usefulness of testing AWS credentials and S3 access before tagging, I have re-introduced the staging documentation section to the pre-tag release process, but made it optional.  This allows it to be removed from a future automated script, but keeps it as a sensible check for manual errors.

Putting the PR back up for review, as it is now more of a cleanup of the text rather than the original removal of this step.

In particular, it does 3 things:

1. removes the instruction to force Buildbot for the release branch (not needed - see FLOC-2580).
1. makes the staging documentation upload optional (useful but not required).
1. moves external links that occur only once out of footers and into text.

wallrj:
Not really sure what to say about this.

 * I agree that publishing the staging docs is a good test for correct AWS credentials etc.
 * But making it an optional step seems to defeat that goal.
 * And in any case, if the AWS credentials are wrong, then presumably I'll see an error when I run `publish-docs`, fix the error and then run the tool again.

On balance I'd say revert to plan-A and remove the "publish staging documentation" section.

And the reasoning behind the decision not to remove the "publish to staging" code seems a bit dodgy to me. Surely, for testing, what you want is the ability to publish to some other bucket than the production documentation bucket. In which case, just make the target bucket a command line option which defaults to `clusterhq-docs`.

PS Mixing in the fix for FLOC-2580 may seem like a time saver on the face of it, but it makes life hard for the reviewer.

jongiddy:
OK, I've removed the staging docs submission completely.  I think this should be OK now, according to the previous reviews, but would like a reviewer to confirm it is OK to merge.

Just to be clear on FLOC-2580, it was created after this PR. I then noticed that this PR solved FLOC-2580. There was no conscious decision to combine two issues in one PR.

itamarst:
This seems to match the claimed issue, so I will ignore confusing discussion in the middle and say go ahead and merge it.

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: Note that there is WIP for removing Vagrant as a release dependency which will mean that this is a one time per engineer task.

PR #1723, opened 2015-06-24
Pin everything.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2515

In master, we'll arrange for this to be generated from the dependencies we care about.


adamtheturtle:
I haven't looked properly at this PR or the requirements so sorry if this is off the mark, but we already have code which does similar using http://pythonhosted.org/tl.eggdeps/doc/tl.eggdeps.graph.html - the Homebrew recipe code. That might be useful.

Azulinho:
we're blocked by this,
it looks good,
merged

PR #1726, opened 2015-06-26
[FLOC-2504] OpenStack HTTPS

jongiddy:
Support private CA's and self-signed certificates for OpenStack.  Some OpenStack implementations do not use public well-known CA's. This PR adds fields to the agent backend configuration:

- `verify_peer` which determines whether to verify HTTPS connections, and
- `verify_ca_path` which selects a CA bundle other than the system CA's.


docs/using/installing/index.rst
```diff
@@ -516,6 +516,17 @@ The configuration item to use OpenStack should look like:
...
 OpenStack must be able to attach volumes created in that region to your Flocker agent nodes.
 
+If the OpenStack cloud uses certificates that are issued by a private Certificate Authority (CA), add the field ``verify_ca_path`` to the dataset stanza, with the path to the CA certificate.
+
+.. code-block:: yaml
+
+   dataset:
+       backend: "openstack"
+       region: "DFW"
+       verify_ca_path: "/etc/flocker/ca.crt"
```
itamarst: How about `/etc/flocker/openstack-ca.crt`? Just to be crystal clear about whose certificate we're talking about.

flocker/node/agents/cinder.py
```diff
@@ -600,6 +600,50 @@ def _openstack_auth_from_config(auth_plugin='password', **config):
...
+    verify, and a string for the path to the CA bundle) into a
+    requests-style single value.
+
+    If the ``verify_peer`` parameter is False, then no verification of
+    the certificate will occur.  This setting is insecure!  Although the
+    connections will be confidential, there is no authentication of the
+    peer.  We're having a private conversation, but we don't know to
+    whom we are speaking.
+
+    If the ``verify_peer`` parameter is True (the default), then the
```
itamarst: Why is `verify_peer` useful? And if it is useful, why is it not documented? As first pass seems like it's not useful.

itamarst:
There are two failure modes we can expect with this configuration option:

1. False positive: the client erroneously validates something it shouldn't. This is Very Bad since it means we can leak passwords to a MITM attack.
2. False negative: the client refuses to talk to a legitimate server. This is not good, but on the other hand we're no worse than we are now.

We should therefore at absolute minimum have functional test for the false positive case: in relevant storage-driver builders run a test that uses a known bogus CA cert (i.e. self-signed) to construct a Cinder `IBlockDeviceAPI` and tries to talk to whatever OpenStack API is configured by default. This should fail because the HTTPS API server will be using different CA cert.

As far as false negative, until we have a private OpenStack setup we can't really test this. I suggest a followup issue and for now we can just rely on user testing (i.e. once this is merged ask customer to test it; if it doesn't work for them then file a follow-up issue).

flocker/node/agents/cinder.py
jongiddy: If `verify_peer is False` you can distinguish between:

1. the system is almost working, and it is just the CA setup that is wrong
1. the system is not working for other reasons

Since it is an easy but insecure solution, it might be better to use it mainly for support. Hence, why I have not documented in the installation instructions.  However, I'm willing to document it as well.

jongiddy: Now documented.

itamarst:
Did we hear back from customer about this?

docs/using/installing/index.rst
```diff
@@ -516,6 +516,32 @@ The configuration item to use OpenStack should look like:
...
+       region: "DFW"
+       verify_ca_path: "/etc/flocker/openstack-ca.crt"
+       auth_plugin: "password"
+       ...
+
+For testing purposes, it is possible to turn off certificate verification, by setting the ``verify_peer`` field to ``false``.
+
+.. warning::
+
+   Only use this setting for troubleshooting, as it is vulnerable to Man-in-the-Middle (MITM) attacks.
```
itamarst: Maybe  "as it is insecure and could lead to your password being stolen."?

itamarst:
OK, address minor comment and merge when buildbot is happy. Thanks!

PR #1708, opened 2015-06-22
Release maintenance/flocker 1.0.1/disable restart policies floc 2467

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2467
Update of https://github.com/ClusterHQ/flocker/pull/1696 (changing branch name ... which matters ...)

PR #1730, opened 2015-06-29
[FLOC-2469] Disable restart policies

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2469
Forward-port of #1708


wallrj:
Thanks @exarkun 

I verified that the changes here are identical to those in #1708.

The two failed builds seem spurious so I forced a rebuild on both.

Please merge when you're happy with the build results.

PR #1731, opened 2015-06-30
Log subprocess output in the context of the subprocess action

jml:
`logged_run_process` has helped us debug issues with failing builders, but we're struggling to associate output from subprocesses with the correct subprocess. 

This (I think) fixes that by making sure that output from the subprocess is logged within the context of the action that corresponds to running the process.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1731)

### DIFF
--- a/flocker/testtools/__init__.py
+++ b/flocker/testtools/__init__.py
@@ -905,7 +905,7 @@ class _LoggingProcessProtocol(ProcessProtocol):
     Intended to be used by ``logged_run_process``.
     """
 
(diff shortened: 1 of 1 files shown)


itamarst:
Looks good, although for longer term might want utility (decorator?) for this particular idiom. Turns up any time one uses Eliot with Twisted protocols.

itamarst:
Merge when buildbot is happy and when you've verified log output actually is what you expect.

PR #1732, opened 2015-06-30
add python27-slim docker image to Vagrantfile

Azulinho:
https://clusterhq.atlassian.net/browse/FLOC-2550

* Add python27-slim image to the vagrant VM as part of the vagrant provisioning tasks.
* Fixes Vagrantfile according to the Ruby Style Guide


jml:
In future, please do style changes like this as separate commits so it's easier to review the actual change. 

Please land without changes when the tests pass.

PR #1733, opened 2015-06-30
Skip test_application_volume_quotas, because it's flaky

jml:
Raise `SkipTest` in test_application_volume_quotas, linking to the bug that tracks us fixing it.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1733)

### DIFF
--- a/flocker/acceptance/test_deployment.py
+++ b/flocker/acceptance/test_deployment.py
@@ -10,7 +10,7 @@ from uuid import uuid4
 
 from pyrsistent import pmap
 
-from twisted.trial.unittest import TestCase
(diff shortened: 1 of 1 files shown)


dwgebler:
Thanks @jml it might be preferable for readability purposes to set the `.skip` attribute on the test method rather than raising the skip in the test itself, e.g. `test_application_volume_quotas.skip = 'Sometimes times out on...'` but I'll leave that up to you to decide. Please merge when you can get a passing set of builds (have not checked out why you have some failures, but probably the usual suspects, request limits etc.)

PR #1737, opened 2015-07-01
[FLOC-2576] Skip conch tests in Flocker 1.0.1

jongiddy:
Back-porting fix to 1.0.1 to allow tests for other patches to pass.


jml:
Approved, pending passing tests.

PR #1739, opened 2015-07-01
[FLOC-2355] Fix test ports Mongo

dwgebler:
Fixes FLOC-2355.

### DIFF
--- a/flocker/acceptance/test_ports.py
+++ b/flocker/acceptance/test_ports.py
@@ -54,7 +54,8 @@ class PortsTests(TestCase):
             },
         }
 
-        cluster.flocker_deploy(self, port_deployment, port_application)
+        self.deployed = cluster.flocker_deploy(
+            self, port_deployment, port_application)
 
     def test_deployment_with_ports(self):
         """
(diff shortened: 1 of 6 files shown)


dwgebler:
Update: Cloud builders are also showing `test_moving_postgres_data` failing with a timeout, though this passes locally for me on CentOS VMs with ZFS. Perhaps whoever reviews this branch with a fresh pair of eyes might have an idea, or it may be a separate issue.

flocker/acceptance/test_ports.py
```diff
@@ -54,7 +54,8 @@ def setUp(self, cluster):
             },
         }
 
-        cluster.flocker_deploy(self, port_deployment, port_application)
+        self.deployed = cluster.flocker_deploy(
+            self, port_deployment, port_application)
```
itamarst: Instead of storing the Deferred, you can just return it from `setUp`. Trial also waits for Deferreds from `setUp`.

flocker/acceptance/test_ports.py
```diff
@@ -68,20 +69,23 @@ def test_deployment_with_ports(self):
 
         application = get_mongo_application().set("ports", ports)
 
-        d = self.cluster.assert_expected_deployment(self, {
```
itamarst: Not necessary if you do changes suggested above.

flocker/acceptance/test_ports.py
```diff
@@ -68,20 +69,23 @@ def test_deployment_with_ports(self):
...
+            })
+        )
+        return self.deployed
 
     @require_mongo
     def test_mongo_accessible(self):
         """
         The port is exposed on the host where Mongo was configured.
         """
-        getting_client = get_mongo_client(
```
itamarst: Ditto.

flocker/acceptance/test_ports.py
```diff
@@ -103,8 +107,10 @@ def test_traffic_routed(self):
         node, and data added to it, that data is visible when a client connects
         to a different node on the cluster.
         """
-        getting_client = get_mongo_client(
```
itamarst: Ditto.

flocker/acceptance/test_postgres.py
```diff
@@ -114,38 +114,47 @@ def setUp(self, cluster):
                 [u"applications", POSTGRES_APPLICATION_NAME, u"ports", 0,
```
itamarst: I think most of the changes here are unnecessary, since you can just return Deferred from `setUp`.

itamarst:
Thanks! Can't spot a bug at first reading, so maybe do the changes I suggested above, which will simplify the diff, and resubmit and we can see if it's easier to spot (or fixes it).

dwgebler:
> Can't spot a bug at first reading

Postgres tests should be passing now, waiting for buildbot....I forgot that `FlockerConfiguration` mutates the configuration dictionary it receives (unfortunately I think I'm the one who wrote it that way in the first place :cry:) so that was the source of the problem.

flocker/acceptance/test_postgres.py
itamarst: Or maybe I misread.

itamarst:
Great, merge when green.

PR #1740, opened 2015-07-01
[FLOC-2582] Clarify deploying Flocker on Nova-created nodes

lukemarsden:
Attempts to resolve https://clusterhq.atlassian.net/browse/FLOC-2582

Add an admonition (the yellow box below) underneath the text where the documentation first says this, in order to make it crystal clear:

---

screenshot 2015-07-01 16 52 00


wallrj:
I like the use of ``:note:`` box but I preferred the original wording in cfe5eb58e0126e4a4c6d4103ea44d0d952dcfa59

Consider using that content instead.

:question: Consider saying "virtual machines" instead of "VMs"

Otherwise looks fine. Please merge when the documentation builder passes.

wallrj:
I forced the failing builder -- rackspace resource limits.

But now I look closely, I see that this is a change to the installation instructions.

I think the two contributors who encountered this difficulty were actually attempting to run our functional tests. So I'd say we need an additional note in the Functional Testing section of our docs:
 * http://build.clusterhq.com/results/docs/cinder-nova-clarify-FLOC-2582/build-8828/gettinginvolved/acceptance-testing.html#functional-testing

wallrj:
After a further discussion with @lukemarsden we've decided that the extra functional testing note isn't worth the effort. That would only make sense if we also document the strategy of testing Flocker against an existing Cinder driver before proceeding to write a custom storage driver....but that's so specialised that it's not worth documenting.

I'll merge this as-is, since the builds have all now passed.

PR #1745, opened 2015-07-02
zfs test_maximum_size_enforced: test the right scenario and verify that

avg-I:
See https://clusterhq.atlassian.net/browse/FLOC-2509

This PR overrides #1718

### DIFF
--- a/flocker/volume/functional/test_filesystems_zfs.py
+++ b/flocker/volume/functional/test_filesystems_zfs.py
@@ -527,7 +527,7 @@ class FilesystemTests(TestCase):
         """
         pool = build_pool(self)
         service = service_for_pool(self, pool)
-        maximum_size = 64 * 1024 * 1024
+        maximum_size = 40 * 1024 * 1024
         volume = service.get(
             MY_VOLUME, size=VolumeSize(maximum_size=maximum_size))
         creating = pool.create(volume)
--- a/flocker/volume/functional/test_filesystems_zfs.py
+++ b/flocker/volume/functional/test_filesystems_zfs.py
@@ -529,8 +529,9 @@ class FilesystemTests(TestCase):
         service = service_for_pool(self, pool)
         # There is a lower-bound on the value of refquota in ZFS.  It seems to
         # be 64MB (but perhaps this isn't universal).
+        maximum_size = 64 * 1024 * 1024
         volume = service.get(
-            MY_VOLUME, size=VolumeSize(maximum_size=64 * 1024 * 1024))
+            MY_VOLUME, size=VolumeSize(maximum_size=maximum_size))
         creating = pool.create(volume)
 
         def write_flush(fObj, data):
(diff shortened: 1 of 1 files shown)


flocker/volume/functional/test_filesystems_zfs.py
```diff
@@ -527,10 +527,9 @@ def test_maximum_size_enforced(self):
         """
         pool = build_pool(self)
         service = service_for_pool(self, pool)
-        # There is a lower-bound on the value of refquota in ZFS.  It seems to
-        # be 64MB (but perhaps this isn't universal).
+        maximum_size = 40 * 1024 * 1024
```
jongiddy: Use the `bitmath` library to write this as `40 * MiB`. It's a minor change, but it makes things more obvious - easier to see if you meant GiB, for example.

flocker/volume/functional/test_filesystems_zfs.py
```diff
@@ -541,7 +540,9 @@ def created(filesystem):
             path = filesystem.get_path()
             # Try to write more than 64MB of data.
             with path.child(b"ok").open("w") as fObj:
-                self.assertRaises(
-                    IOError, write_flush, fObj, b"x" * 64 * 1024 * 1024)
+                exc = self.assertRaises(
+                    IOError, write_flush, fObj, b"x" * (maximum_size + 1))
```
jongiddy: This will be `maximum_size + Byte(1)` with `bitmath`

flocker/volume/functional/test_filesystems_zfs.py
```diff
@@ -527,10 +527,9 @@ def test_maximum_size_enforced(self):
         """
         pool = build_pool(self)
         service = service_for_pool(self, pool)
-        # There is a lower-bound on the value of refquota in ZFS.  It seems to
-        # be 64MB (but perhaps this isn't universal).
+        maximum_size = 40 * 1024 * 1024
         volume = service.get(
-            MY_VOLUME, size=VolumeSize(maximum_size=64 * 1024 * 1024))
+            MY_VOLUME, size=VolumeSize(maximum_size=maximum_size))
```
jongiddy: `maximum_size=int(maximum_size.to_Byte())`

flocker/volume/functional/test_filesystems_zfs.py
jongiddy: This creates a 40 MiB in-memory string. Is it possible to do this with a loop that writes a smaller byte string multiple times?

If so, this would be a place to use the `with` context manager:
```python
with self.assertRaises(IOError) as exc:
    bufsize = 8192
    buf = 'x' * bufsize
    for i in range(maximum_size / bufsize + 1):
        write_flush(fObj, buf)
```

flocker/volume/functional/test_filesystems_zfs.py
jongiddy: Add a comment explaining why 40 MiB is a good size to test

jongiddy:
Thanks @avg-I . I did think you could use the `bitmath` library here, but it probably makes it more complex than necessary.  Please address the comments, and resubmit.

wallrj:
Thanks @avg-I 

Looks good to me.

I verified that the tests pass on my local vagrant dev box and that it fails when I comment out the writing of the extra byte.

Please merge.

PR #1747, opened 2015-07-02
[FLOC-2538] An option to allow all flocker scripts to log to a file

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-2538

PR #1750, opened 2015-07-03
[FLOC-2595] An option to allow logging directly to a file

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2595

Based on design in #1747

### DIFF
--- a/flocker/common/functional/test_script.py
+++ b/flocker/common/functional/test_script.py
@@ -79,11 +79,15 @@ class FlockerScriptRunnerTests(TestCase):
         """
         code = b'''\
 from twisted.python.usage import Options
-from flocker.common.script import FlockerScriptRunner
+from flocker.common.script import FlockerScriptRunner, flocker_standard_options
 
 from flocker.common.functional.test_script import {}
 
-FlockerScriptRunner({}(), Options()).main()
+@flocker_standard_options
+class StandardOptions(Options):
+    pass
+
+FlockerScriptRunner({}(), StandardOptions()).main()
 '''.format(script.__name__, script.__name__)
         d = getProcessOutput(sys.executable, [b"-c", code], env=os.environ,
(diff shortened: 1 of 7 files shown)


wallrj:
Hmm. So ubuntu acceptance tests timeout. My guess is that the convergence agents aren't starting up because the logfile directory /var/log/flocker doesn't exist.....but there are no logs to tell me that :frowning:  

Perhaps ``run-acceptance-tests`` should check that the agent processes are running before commencing the tests.

flocker/common/functional/test_script.py
```diff
@@ -172,3 +184,35 @@ def test_sigint(self):
...
+            self, logfile_messages[1], {u"key": 123}
+        )
+
+    def test_file_logging(self):
+        """
+        Logged messages are written to ``logfile`` if ``--logfile`` is supplied
+        on the command line.
+        """
+        logfile = FilePath(self.mktemp()).child('foo.log')
+        logfile.parent().makedirs()
```
dwgebler: :question: doesn't this throw an `OSError` if the directory matching `logfile.parent()` already exists?

dwgebler:
Thanks @wallrj this looks good, the only thing I can really see is that there's no testing of the log rotation when 100MB is written. I would be tempted to say that Twisted's `LogFile' is tested and we're entitled to therefore assume in this branch that its behaviour works as intended. But I suppose on the flip-side, there's no testing here that Flocker's logs get rotated and therefore consume a limited quantity of disk space - and given that that's the specific issue the branch is intended to fix, I'm thinking that should be tested actually.

So yes I think please do add those tests, otherwise this looks good to merge when buildbot passes.

wallrj:
I guess I need to create a new branch and PR for merging this into `release/flocker-1.0.2`.

PR #1753, opened 2015-07-03
[FLOC-1194] Add forceable container removal if needed

dwgebler:
Probably fixes FLOC-1194 by allowing the affected test to force container removal during cleanup.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -546,7 +546,7 @@ class DockerClient(object):
         ).write()
         return result['State']['Running']
 
-    def remove(self, unit_name):
+    def remove(self, unit_name, force=False):
         container_name = self._to_container_name(unit_name)
 
         def _remove():
@@ -608,7 +608,7 @@ class DockerClient(object):
                     message_type="flocker:docker:container_remove",
                     container=container_name
                 ).write()
-                self._client.remove_container(container_name)
+                self._client.remove_container(container_name, force=force)
                 Message.new(
                     message_type="flocker:docker:container_removed",
                     container=container_name
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -140,6 +140,46 @@ class GenericDockerClientTests(TestCase):
 
         return d
 
+    def test_force_remove(self):
+        """
+        ``DockerClient.remove`` can forcefully remove a running container
+        when the ``force`` parameter is set to True.
+        """
+        client = DockerClient()
+        unit_name = u'test-force-remove'
+        expected_states = (u'active',)
+        d = client.add(
+            unit_name=unit_name,
+            image_name=u'clusterhq/flask'
+        )
+        d.addCallback(
+            lambda _: wait_for_unit_state(
+                client, unit_name, expected_states
+            )
+        )
+
+        def check_is_listed(_):
+            listed = client.list()
+
+            def check_unit(results):
+                self.assertTrue(unit_name in [unit.name for unit in results])
+            listed.addCallback(check_unit)
+            return listed
+
+        d.addCallback(check_is_listed)
+        d.addCallback(lambda _: client.remove(unit_name, force=True))
+
+        def check_is_removed(_):
+            listed = client.list()
+
+            def check_unit(results):
+                self.assertFalse(unit_name in [unit.name for unit in results])
+            listed.addCallback(check_unit)
+            return listed
+
+        d.addCallback(check_is_removed)
+        return d
+
     def test_default_base_url(self):
         """
         ``DockerClient`` instantiated with a default base URL for a socket
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -99,7 +99,7 @@ class GenericDockerClientTests(TestCase):
                         environment=None, volumes=(),
                         mem_limit=None, cpu_shares=None,
                         restart_policy=RestartNever(),
-                        command_line=None):
+                        command_line=None, force=False):
         """
         Start a unit and wait until it reaches the `active` state or the
         supplied `expected_state`.
@@ -114,6 +114,8 @@ class GenericDockerClientTests(TestCase):
         :param cpu_shares: See ``IDockerClient.add``.
         :param restart_policy: See ``IDockerClient.add``.
         :param command_line: See ``IDockerClient.add``.
+        :param bool force: Force container removal via SIGKILL
+            during cleanup.
 
         :return: ``Deferred`` that fires with the ``DockerClient`` when
             the unit reaches the expected state.
@@ -130,7 +132,7 @@ class GenericDockerClientTests(TestCase):
             restart_policy=restart_policy,
             command_line=command_line,
         )
-        self.addCleanup(client.remove, unit_name)
+        self.addCleanup(client.remove, unit_name, force)
 
         d.addCallback(lambda _: wait_for_unit_state(client, unit_name,
                                                     expected_states))
@@ -356,6 +358,7 @@ class GenericDockerClientTests(TestCase):
                 unit_name=unit_name,
                 image_name=image_name,
                 environment=Environment(variables=expected_variables),
+                force=True,
             )
         d.addCallback(image_built)
 


flocker/node/_docker.py
```diff
@@ -546,7 +546,7 @@ def _blocking_container_runs(self, container_name):
         ).write()
         return result['State']['Running']
 
-    def remove(self, unit_name):
+    def remove(self, unit_name, force=False):
```
exarkun: This method is an implementation of an interface method, `IDockerClient.remove`.

  * The interface needs to be updated similarly.
  * The other implementation needs to be updated similarly.
  * The interface-based test suite factory, `make_idockerclient_tests`, needs to have tests for the new functionality added.

flocker/node/functional/test_docker.py
```diff
@@ -356,6 +398,7 @@ def image_built(image_name):
                 unit_name=unit_name,
                 image_name=image_name,
                 environment=Environment(variables=expected_variables),
+                force=True,
```
exarkun: If the only real change in behavior is in the test suite then, hopefully, our test suite will stop failing spuriously with this problem.  What about the actual implementation, though?  If this failure can strike test code that's removing a container, it seems plausible it could strike `StopApplication` when in use on a real cluster.

Is there a reason to think this isn't the case?  Or does this problem eventually solve itself so that, as long as you're retrying (as the looping approach of the convergence agent will do) the removal does eventually succeed?

exarkun: Also, what about other tests that create and then remove containers (ie, all the other tests using `start_application`)?

Couldn't they all also fail with this error?  The error doesn't seem particularly container-with-environment-variables specific at first glance.  If any `start_container`-using test could encounter this problem, it would probably make more sense to just have force removal be the behavior `start_container` implements and not bother with the parameter and updating all the callers.

flocker/node/functional/test_docker.py
```diff
@@ -138,6 +140,46 @@ def start_container(self, unit_name,
...
+            lambda _: wait_for_unit_state(
+                client, unit_name, expected_states
+            )
+        )
+
+        def check_is_listed(_):
+            listed = client.list()
+
+            def check_unit(results):
+                self.assertTrue(unit_name in [unit.name for unit in results])
```
exarkun: `assertIn(unit_name, [unit.name for unit in results])`

This provides a better failure message because it has both the "needle" and the "haystack".  `assertTrue` just has `True` or `False`.

flocker/node/functional/test_docker.py
```diff
@@ -138,6 +140,46 @@ def start_container(self, unit_name,
...
+            return listed
+
+        d.addCallback(check_is_listed)
+        d.addCallback(lambda _: client.remove(unit_name, force=True))
+
+        def check_is_removed(_):
+            listed = client.list()
+
+            def check_unit(results):
+                self.assertFalse(unit_name in [unit.name for unit in results])
```
exarkun: Likewise.

flocker/node/functional/test_docker.py
```diff
@@ -138,6 +140,46 @@ def start_container(self, unit_name,
...
+        """
+        ``DockerClient.remove`` can forcefully remove a running container
+        when the ``force`` parameter is set to True.
+        """
+        client = DockerClient()
+        unit_name = u'test-force-remove'
+        expected_states = (u'active',)
+        d = client.add(
+            unit_name=unit_name,
+            image_name=u'clusterhq/flask'
```
exarkun: This image and flask are incidental complexity for this test.  The detail of which container to run doesn't really matter to what the test is concerned with testing.  All of the code from Flask that will run when this test runs is likewise irrelevant.  Depending on the execution environment of the test suite, it's also possible this will result in a Docker image being pulled from Dockerhub.

The incidental complexity increases the maintenance cost of the test method.  The Dockerhub dependency increases the chances that this will fail intermittently due to transient network or Dockerhub conditions.

A way to address the incidental complexity is to contain it and then isolate it from this particular test.  Add an attribute or method to the class responsible for giving back the name of an image that has the desired properties - in this case, an image that runs indefinitely, I think.

A way to address the external dependency is to build a custom image for this test as we do elsewhere.  I think a busybox-based container that runs forever should be pretty straightforward.

flocker/node/functional/test_docker.py
```diff
@@ -138,6 +140,46 @@ def start_container(self, unit_name,
 
         return d
 
+    def test_force_remove(self):
+        """
+        ``DockerClient.remove`` can forcefully remove a running container
+        when the ``force`` parameter is set to True.
```
exarkun: Something to note about this test is that, though it does exercise the new `force` parameter, it exercises a slightly different part of its behavior than we're most interested in.  "Remove a running container" isn't quite "avoid the internal server error that sometimes happens because 'Device is busy'".  It could be that `force=True` does fix this as well as its other (documented) behavior - but we haven't demonstrated it with this test.

I skimmed the Docker issue and didn't notice the section that suggests force=True as a work-around for this.  A reference to that suggestion, as well as to the issue as a whole somewhere in this code (perhaps near the callers that pass `force=True` to `remove`) would be nice.

exarkun:
Thanks.  Please address and re-submit.

flocker/node/functional/test_docker.py
dwgebler: > Also, what about other tests that create and then remove containers (ie, all the other tests using start_application)?
> 
> Couldn't they all also fail with this error?

Plausibly, yes - but the issue FLOC-1194 is only concerned with the environment variables test and is a "fix flaky test" category issue. So it possibly makes sense to not alter the behaviour of other tests using the Client change in this branch until such time as they are also identified as intermittently failing on the same device busy error.

dwgebler: > Is there a reason to think this isn't the case?

No, it is plausible this error could strike `StopApplication` on a real cluster, because the precise cause(s) of this Docker error are not known; some causes are known, specific to certain combinations of platforms, images and Docker versions, but I have thusfar not found a way of reliably reproducing the issue.

> Or does this problem eventually solve itself so that, as long as you're retrying (as the looping approach of the convergence agent will do) the removal does eventually succeed?

Maybe; I've seen at least one report (anecdote) that it was necessary for the host running Docker to be rebooted to resolve the problem, but again this seems to vary.

There's not even a guarantee that force removal via Docker's API prevents the device busy error, I've seen reports of it happening even with the `-f` switch supplied on Docker's CLI. It would be nice if we could write a test that reliably reproduced the problem and tested that forced-removal reliably circumvents it, but I don't have sufficient information at this time to do that. In the scope of this ticket, we're basically going for an approach of "based on discussions of this error found on historical Docker issues listings, add force removal capability to our Docker API client and tell the intermittently failing test to use it, then see if the test failure occurs again in future..." - not the most robust solution, I know, but it might help Buildbot stay green.

flocker/node/_docker.py
```diff
@@ -208,13 +208,14 @@ def exists(unit_name):
             otherwise ``False``.
         """
 
-    def remove(unit_name):
+    def remove(unit_name, force):
```
wallrj: I'd have thought `zope.verify` would have caught this, but the signature in the interface should match the signature in the implementation, so the interface ought to have `force=False`.

flocker/node/_docker.py
```diff
@@ -208,13 +208,14 @@ def exists(unit_name):
...
         """
         Stop and delete the given unit.
 
         This can be done multiple times in a row for the same unit.
 
         :param unicode unit_name: The name of the unit to stop.
+        :param bool force: Force removal of the unit.
```
wallrj: And add a note about the default behaviour here.

wallrj:
http://blog.hashbangbash.com/2014/11/docker-devicemapper-fix-for-device-or-resource-busy-ebusy/ seems to describe the problem and the fix.

Do we know if any of our supported Distros implement the suggested fix in their systemd / upstart units?

@binocarlos tells me that Docker on Ubuntu uses AUFs by default if the kernel module is detected; exactly because of this device mapper issue. Although judging by the @dwgebler 's other link (https://github.com/docker/docker/issues/9665) the problem also occurs with AUFs.
Does anyone know if AUFs is less flaky than devicemapper? If so, can we add AUFS to our list of package dependencies and document that we recommend using  docker with AUFS instead of devicemapper.

I'm not convinced about the `force=True` fix. 
Our DockerClientAPI.remove already attempts to stop the container before removal and waits for it to stop.
And I'm sure @jongiddy and others have considered adding the `force` option in the past for other reasons or perhaps for this reason. 

I also notice that docker API 1.19 has a `v` option which sounds related:
 * https://docs.docker.com/reference/api/docker_remote_api_v1.19/#remove-a-container

I don't really know what to say. I think I'll reject this so that we can discuss it further.

wallrj:
Has anyone tested this:
 * https://github.com/docker/docker/issues/9665#issuecomment-107361532

...work around?

Start another container for the same image then remove the old container.

dwgebler:
> Start another container for the same image then remove the old container.

I didn't try that one - `force` does definitely seem flaky; there was a suggestion on the Docker issue that it worked for some people in resolving the problem of removing the container, but I'm not sure why that would be the case, given that the only documented behaviour of `force` I can find is that it first terminates the process in the container via SIGKILL, which should be the same as stopping the container first - which we already do.

As per the discussion we were just having in the office, supporting a more up-to-date Docker, i.e. 1.7 and 1.19 API might also be a good thing to start thinking about at this time.

jml:

PR #1757, opened 2015-07-07
[FLOC-2540] Fix test_detach_volume failure.

myechuri:
Request for design review for FLOC-2540. The test failed because EBS did not update volume's state and attachment information atomically.


wallrj:
Thanks @myechuri 

I may not be the best person to review this design. We probably should get some feedback from @exarkun , but here are my thoughts:

 * The new state change table looks really neat.
 * AFAICS, the ``wait_for_volume_state_change`` function doesn't enforce the state change sequence. It returns when the state reaches `end_state` but it doesn't raise an exception if eg the volume starts in ``end_state`` or if eg the volume moves directly from ``start_state`` to ``end_state``....because we're polling and sleeping and we may have missed the intermediate states....but in that case, at best, I think the new function name is misleading and at worst, it makes me wonder why we are even checking for the start and transient states.
 * It seems to me that we should at least check that volume starts in the expected start state...which maybe implies that the helper function should be designed a little differently. Perhaps it should take the state change table plus the function responsible for triggering the state change. It can then check that the volume starts in `start_state`, call the supplied function, and wait for the end state.
 * ...but this starts to sound a lot like a job for  `machinist` which is where @exarkun  can weigh in.

Can you state the problem we're trying to fix with this code? The issue only shows the test failure.
My understanding is that boto sometimes reports: 
 * volume state `attached` but `attach_data.device` is not populated, or
 * volume state `detached` but `attach_data.device` is still populated.

Is that correct?

So perhaps we should discuss:
 * why we need to monitor the state transitions, 
 * how strict we should be about checking the start, transient and end states.
 * double check whether boto (or some other AWS library) provides an API where the volume state and attachment state are in sync.

myechuri:
Thanks for the detailed review, @wallrj !

> AFAICS, the wait_for_volume_state_change function doesn't enforce the state change sequence. It returns when the state reaches end_state but it doesn't raise an exception if eg the volume starts in end_state or if eg the volume moves directly from start_state to end_state....because we're polling and sleeping and we may have missed the intermediate states....but in that case, at best, I think the new function name is misleading and at worst, it makes me wonder why we are even checking for the start and transient states.

The reason i was checking for start and transient state (as opposed to just waiting for end state) was to return early if the state transition failed, instead of waiting for ``VOLUME_STATE_CHANGE_TIMEOUT`` (300 seconds).

myechuri:
> Can you state the problem we're trying to fix with this code? The issue only shows the test failure.
My understanding is that boto sometimes reports:

volume state attached but attach_data.device is not populated, or
volume state detached but attach_data.device is still populated.


The problem is that EBS does not update volume's state and attach_data.device atomically. 

The purpose of this change is to start cleaning up our driver. In ``ebs.py``, we sometimes look at volume's state, and at other times look at attach_data.device. i'd like to clean this up so that we have one source of truth, and the actions leading up to change in either state or attach_data declare success only after both are updated. Can you please let me know if that sounds reasonable?

myechuri:
> why we need to monitor the state transitions,

i want to monitor state transitions so that we exit early (< 300s) in failure cases.

myechuri:
> how strict we should be about checking the start, transient and end states.

Being strict is nice because our driver is clear about expected state transitions, and notices/logs unexpected states that might be useful in debugging problems.

myechuri:
> double check whether boto (or some other AWS library) provides an API where the volume state and attachment state are in sync.

Boto does not, unfortunately.

myechuri:
> And now that we have time to do things properly, the design should also probably sketch out how we can test this more thoroughly or at least explain why extra testing is not necessary here.

Testing is necessary, i just did not get to write out the test design before sending this out for review, apologies for that. i wanted to run the design review by you all before getting to testing. Once we converge on design, will update test plan. Sample test plan would include tests for create/attach/detach/destroy with expected and unexpected state transition tables.

myechuri:
> ...but this starts to sound a lot like a job for machinist which is where @exarkun can weigh in.

i looked into ``machinist`` usecases in ``node/_loop.py`` and decided to not use it because ``ebs.py`` is simply observing partial/full state transitions and is not actually executing the transitions. Please let me know if i am missing something?

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,93 @@
...
+    """
+    table = pmap_field(NamedConstant, VolumeStateFlow)
+
+
+def populate():
+    """
+    Initialize volume state table  with transitions for ``create_volume``,
+    ``attach_volume``, ``detach_volume``, ``delete_volume`` operations.
+    """
+    table = pmap()
```
itamarst: :question: This might be easier to read if you did e.g. `S = VolumeStates` and `O = VolumeOperations` at the top.

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,93 @@
...
+
+
+class VolumeStates(Values):
+    """
+    Expected EBS volume states during ``VolumeOperations``.
+    """
+    EMPTY = ValueConstant('')
+    CREATING = ValueConstant(u'creating')
+    AVAILABLE = ValueConstant(u'available')
+    ATTACHING = ValueConstant(u'attaching')
```
itamarst: The official docs don't even mention this one. Sigh.

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,93 @@
...
+    IN_USE = ValueConstant(u'in-use')
+    DETACHING = ValueConstant(u'detaching')
+    DELETING = ValueConstant(u'deleting')
```
itamarst: The docs do however mention `deleted`. Oh well.

flocker/node/agents/ebs.py
itamarst: :question: and maybe an inline utility function?

```
table = pmap()

def add_flow(operation, start, transient, end_state, has_attach_data):
    return table.set(operation, ...)

table = add_flow(O.CREATE, S.EMPTY, S.CREATING, S.AVAILABLE)
table = add_flow(...)
return table
```

flocker/node/agents/ebs.py
```diff
@@ -243,59 +335,88 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
+    :raises Exception: When input volume fails to reach expected backend
+        state for given operation within timeout seconds.
     """
     # It typically takes a few seconds for anything to happen, so start
     # out sleeping a little before doing initial check to reduce
     # unnecessary polling of the API:
     time.sleep(5.0)
 
-    # Wait ``VOLUME_STATE_CHANGE_TIMEOUT`` seconds for
+    # XXX Catch KeyError.
```
itamarst: Well... you'd get a KeyError if anyone ever passed in an unknown operation. But that's a coding bug that you can't actually handle. So best you can do is log a traceback, which raising KeyError will do anyway. So no need for this comment, or further action, I think.

flocker/node/agents/ebs.py
```diff
@@ -243,59 +335,88 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
+    :rtype: boto.ec2.volume
+
+    """
+    return volume.update()
+
+
+def _wait_for_volume_state_change(operation,
+                                  volume,
+                                  update=_get_ebs_volume_state,
+                                  timeout=VOLUME_STATE_CHANGE_TIMEOUT):
```
itamarst: This makes some progress towards what I was thinking of, but not the whole way.

Here's the logic I was going for. Instead of:

```
def wait(x, y, get_update):
    start = time.time()
    while time.time() - start < TIMEOUT:
        if x and y and get_update().foo:
            return
     raise TimeoutError()
```

We split this into two:
```
def should_finish(x, y, foo, start_time, current_time):
     if current_time - start_time < T:
         raise TimeoutError()
     return x and y and foo

def wait(x, y, get_update):
    start = time.time()
    while True:
        if should_finish(x, y, get_update().foo, start, time.time()):
            return
```

Notice that all the decision logic branches are now in `should_finish`. At the same time `should_finish` is completely side-effect free ("purely functional"): no calls to the outside universe, so you can test it all in isolation.

flocker/node/agents/ebs.py
```diff
@@ -650,17 +766,12 @@ def detach_volume(self, blockdevice_id):
             blockdevice_id is not currently 'in-use'.
         """
         ebs_volume = self._get_ebs_volume(blockdevice_id)
-        volume = _blockdevicevolume_from_ebs_volume(ebs_volume)
-        if (volume.attached_to is None or
-                ebs_volume.status != 'in-use'):
+        if ebs_volume.status != 'in-use':
```
itamarst: Is this the specific change that fixes the intermittent failing test?

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,93 @@
...
+
+class VolumeStateTable(PRecord):
+    """
+    Map of volume operation to expected volume state transitions
+    and expected update to volume's ``attach_data``.
+    """
+    table = pmap_field(NamedConstant, VolumeStateFlow)
+
+
+def populate():
```
itamarst: Make this private, you don't really need to call it more than once since you are setting a module-level object with its result.

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,93 @@
...
+    destroy_state_flow = VolumeStateFlow(
+        start_state=VolumeStates.AVAILABLE,
+        transient_state=VolumeStates.DELETING,
+        end_state=VolumeStates.EMPTY,
+        has_attach_data=False
+        )
+    table = table.set(VolumeOperations.DESTROY, destroy_state_flow)
+    return table
+
+volume_state_table = VolumeStateTable(table=populate())
```
itamarst: This is a constant, really, so probably should be upper-case.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -156,3 +160,273 @@ def test_next_device_in_use(self):
...
+        :rtype: ValueConstant
+        """
+        volume_state_table = VolumeStateTable(table=populate())
+        state_flow = volume_state_table.table[operation]
+        valid_states = set([state_flow.start_state,
+                            state_flow.transient_state,
+                            state_flow.end_state])
+        invalid_states = set(VolumeStates._enumerants.values()) - valid_states
+        invalid_state = invalid_states.pop()
+        if invalid_state is None:
```
itamarst: It doesn't seem like `None` should ever happen:

1. `pop()` will throw KeyError I suspect if set is empty.
2. Logically this oughtn't happen since number of states is > 3.

So you can just skip the `None` code path.

itamarst: More generally: "return successfully, or not at all." If an error occurred such that you can't handle it, you should throw an exception, not return `None`. http://c2.com/cgi/wiki?SamuraiPrinciple though I've never heard it called that before.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -156,3 +160,273 @@ def test_next_device_in_use(self):
...
+            return None
+        else:
+            return invalid_state.value
+
+    def test_error_volume_state_during_attach(self):
+        """
+        Assert that ``Exception`` is thrown when attach volume operation leads
```
itamarst: At this point it's probably time to use an actually specific Exception subclass (e.g. new TimeoutException one). The problem with merely asserting Exception is that e.g. maybe an `AttributeError` or `NameError` was raised due to typo in that logic branch's implementation... this indicates bug in your code, but this test will happily pass since those are both `Exception` subclasses.

flocker/node/agents/ebs.py
itamarst: (I edited the above since originally writing it)

itamarst: The downside to this approach is that you then need another test or two for `should_finish` integration with `wait`. Tradeoffs everywhere!

flocker/node/agents/functional/test_ebs.py
```diff
@@ -156,3 +160,273 @@ def test_next_device_in_use(self):
...
+
+            :param boto.ec2.volume.Volume volume: Volume to move to
+                invalid state.
+            """
+            volume.status = self._pick_invalid_state(operation)
+
+        self.assertRaises(Exception, _wait_for_volume_state_change,
+                          operation, volume, update, TIMEOUT)
+
+    def test_detach_stuck_detaching(self):
```
itamarst: Many of these tests are repetitive in their implementation. I suggest implementing a utility assert method, `assert_raises_on_unexpected_state` that does the repetitive part, and then call that from each test method.

itamarst: Alternatively, this seems like a good place to figure out how to use https://pypi.python.org/pypi/hypothesis

itamarst: Well. Probably not worth figuring out `hypothesis` as bottleneck for fixing an actual bug and intermittent test failure, so maybe not. But might be worth trying after this is merged to see if it makes things simpler.

itamarst:
A good improvement, just needs a little more work - address the comments above and resubmit. I didn't review the tests logic in depth on theory that next revision will be more succinct.

flocker/node/agents/ebs.py
myechuri: Thanks! Definitely makes for easier read. Fixed.

flocker/node/agents/ebs.py
myechuri: Thats an interesting state. Have never observed ``deleted``, but it could be a transient state right after ``deleting``. Might be worth it to add multiple transient state support per operation. Will take it up as a follow up. Thanks for pointing out!

flocker/node/agents/ebs.py
myechuri: Makes sense. Removed the comment. Thanks!

flocker/node/agents/ebs.py
myechuri: Thanks! Yes, this refactors the wait nicely. Working on test changes.

flocker/node/agents/ebs.py
myechuri: Yes, this change fixes one the two intermittent failing tests (FLOC-2540 and FLOC-2639).

flocker/node/agents/ebs.py
myechuri: Thanks for pointing out! Fixed.

flocker/node/agents/ebs.py

itamarst:
Is this ready for review again, or still work in progress?

myechuri:
> Is this ready for review again, or still work in progress?

Not ready for review again, needs test updates. Sorry for the delay!

itamarst:
No problem, just checking.

flocker/node/agents/functional/test_ebs.py
myechuri: Thanks for suggesting! Added TimeoutException, and see the value gained. Thanks!

flocker/node/agents/functional/test_ebs.py
myechuri: Thanks for pointing to hypothesis! Looks like a great fit for what i was intending to do. In the interest of moving this stalled PR along, went with refactoring asserts out as suggested initially. Hope to use ``hypothesis`` for next testing need that falls in similar category. Thanks!

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,141 @@
...
+        self.blockdevice_id = blockdevice_id
+        Exception.__init__(self, operation)
+        self.operation = operation
+        Exception.__init__(self, start_state)
+        self.start_state = start_state
+        Exception.__init__(self, transient_state)
+        self.transient_state = transient_state
+        Exception.__init__(self, end_state)
+        self.end_state = end_state
+        Exception.__init__(self, current_state)
```
jml: Unless I'm missing something, I don't think the repeated upcalls to `Exception.__init__` achieve anything. Normal practice is to upcall with the message that you'd like to see when the exception is raised. As it is, `str(TimeoutException)` will be `TimeoutException(current_state)`.

flocker/node/agents/ebs.py
```diff
@@ -42,6 +47,141 @@
...
+        self.current_state = current_state
```
jml: Likewise here.

flocker/node/agents/ebs.py
```diff
@@ -243,74 +383,125 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
+    if volume.status == end_state:
+        # If end state for the volume comes with attach data,
+        # declare success only upon discovering attach data.
+        if sets_attach:
+            if (volume.attach_data is not None and
+                    (volume.attach_data.device != '' and
+                     volume.attach_data.instance_id != '')):
+                return True
+            else:
+                return False
```
jml: I don't think the `if`/`else` here adds to the readability. Please change to just return the condition, e.g.

```
        return (volume.attach_data is not None and
                (volume.attach_data.device != '' and
                 volume.attach_data.instance_id != ''))
```

flocker/node/agents/ebs.py
```diff
@@ -243,74 +383,125 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
+        elif unsets_attach:
+            if (volume.attach_data is None or
+                    (volume.attach_data.device is None and
+                     volume.attach_data.instance_id is None)):
+                return True
+            else:
+                return False
```
jml: Same here, thanks.

flocker/node/agents/ebs.py
```diff
@@ -243,74 +383,125 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
+        else:
+            return True
+    elif volume.status not in [start_state, transient_state, end_state]:
+        raise UnexpectedStateException(unicode(volume.id), operation,
+                                       start_state, transient_state, end_state,
+                                       volume.status)
+    return False
```
jml: :question: I think whole chunk of conditional logic would be easier to follow if it were written in an early-escape style. The reason is that the `sets_attach`/`unsets_attach` bit is fairly intrinsically complex, and an early-escape on the irrelevant conditions can make it easier to focus / understand.

e.g.
```python
    if volume.status not in [start_state, transient_state, end_state]:
        raise UnexpectedStateException(unicode(volume.id), operation,
                                       start_state, transient_state, end_state,
                                       volume.status)

    if volume.status != end_state:
        return False

    # If end state for the volume comes with attach data,
    # declare success only upon discovering attach data.
    if sets_attach:
...
```

flocker/node/agents/ebs.py
jml: Please also make `current_time` a parameter, rather than accessing `time.time()` directly in this function.

flocker/node/agents/ebs.py
```diff
@@ -243,74 +383,125 @@ def _blockdevicevolume_from_ebs_volume(ebs_volume):
...
-            volume.update()
-        except EC2ResponseError as e:
-            # If AWS cannot find the volume, raise ``UnknownVolume``.
-            # (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/errors-overview.html
-            # for error details).
-            if e.code == u'InvalidVolume.NotFound':
-                raise UnknownVolume(volume.id)
-        if volume.status == end_status:
+    while True:
+        if _should_finish(operation, volume, update, start_time, timeout):
```
jml: Why not `while not _should_finish(...)`?

flocker/node/agents/ebs.py
jml: I guess this should become `VolumeStates.IN_USE`.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -156,3 +162,293 @@ def test_next_device_in_use(self):
...
+        if operation == self.V.CREATE:
+            self.assertEqual(volume.status, u'available')
+        elif operation == self.V.DESTROY:
+            self.assertEquals(volume.status, u'')
+        elif operation == self.V.ATTACH:
+            self.assertEqual([volume.status, volume.attach_data.device,
+                              volume.attach_data.instance_id],
+                             [u'in-use', u'/dev/sdf', u'i-xyz'])
+        elif operation == self.V.DETACH:
+            self.assertEqual(volume.status, u'available')
```
jml: It's generally a good idea to avoid `if` statements in tests (see http://xunitpatterns.com/Conditional%20Test%20Logic.html for excruciating amounts of detail on the subject). I think with `_assert_success` in particular, the tests would be easier to read if there were a helper that creates the volume, waits for state change and then returns the volume. The various tests would then assert things about the returned volume.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -156,3 +162,293 @@ def test_next_device_in_use(self):
...
+        update = self._custom_update(operation, testcase, attach_data_type)
+        start_time = time.time()
+
+        if testcase == self.S.ERROR:
+            self.assertRaises(UnexpectedStateException, _should_finish,
+                              operation, volume, update, start_time, TIMEOUT)
+        else:
+            finish_result = _should_finish(operation, volume, update,
+                                           start_time)
+            self.assertEqual(False, finish_result)
```
jml: I don't really understand the condition here. What's `testcase`? What is being tested when `testcase` is not `ERROR`?

jml:
Please merge master, resolve conflicts and ensure all tests pass.

jml:
Thanks for getting to this change. It looks like the code and tests are both becoming more robust. I've requested a few changes around code clarity. Please resubmit for review when you've responded.

flocker/node/agents/ebs.py
myechuri: Thanks for pointing out! Leaving ``blockdevice_id`` as the message, and rest of the parameters as state. Thanks!

flocker/node/agents/ebs.py
myechuri: Fixed here as well. Thanks!

flocker/node/agents/ebs.py
myechuri: Thanks for pointing out! Fixed.

flocker/node/agents/ebs.py
myechuri: Changed. Thanks!

flocker/node/agents/ebs.py
myechuri: Sounds reasonable. Thanks for suggesting! Changed.

flocker/node/agents/ebs.py
myechuri: Cleaner! Changed. Thanks!

flocker/node/agents/functional/test_ebs.py
myechuri: Thanks for raising concern. Reworked this part a little.

flocker/node/agents/functional/test_ebs.py
myechuri: Apologies for poor naming. ``testcase`` was a flag to indicate type of volume state transition test. Renamed ``testcase``, and separated out ``ERROR_STATE`` assert from assert module for other volume end states (``TRANSIT_STATE``, ``DESTINATION_STATE``).

flocker/node/agents/ebs.py
jml: @myechuri It doesn't look like you've addressed this comment. Is there a reason it shouldn't be `VolumeStates.IN_USE`?

jml:
Just one minor comment. Please address & merge. Thanks!

flocker/node/agents/ebs.py
itamarst: `VolumeStates.IN_USE.value` I think. I pushed a change with this, will see if tests pass.

PR #1756, opened 2015-07-06
[FLOC-2595] An option to allow logging directly to a file

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2595

Reviewed and accepted in #1750

Based on design in #1747

PR #1758, opened 2015-07-07
[FLOC-2596] An option to allow logging directly to a file

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2596

Merges #1756 to master.


dwgebler:
Confirmed this is the same set of changes as the previously reviewed branch, but for `master` this time, and that the tests all continue to pass. Please merge when Buildbot is green.

PR #1763, opened 2015-07-09
[FLOC-1040] test_added_is_listed intermittent failure

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1040

This changes the test to retry when it encounters this error.  In support of this, a new exception, `AddressInUse` is also introduced and the implementations updated to raise it when this particular failure is encountered.  In support for *this*, automatic port assignment (specify port 0, get a free port) is also added to the Docker client implementations.

Note the first revision in the branch is assorted refactoring I did while working on the rest.  Behavior changes are all in the second revision.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -230,7 +230,8 @@ class IDockerClient(Interface):
 
 @implementer(IDockerClient)
 class FakeDockerClient(object):
-    """In-memory fake that simulates talking to a docker daemon.
+    """
+    In-memory fake that simulates talking to a docker daemon.
 
     The state the the simulated units is stored in memory.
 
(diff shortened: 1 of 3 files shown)


flocker/node/_docker.py
```diff
@@ -434,6 +469,47 @@ def _serialize_restart_policy(self, restart_policy):
...
+        # Cannot start container <name>: Error starting userland proxy:
+        # listen tcp <ip>:<port>: bind: address already in use
+        #
+        # Or (without newline) like:
+        #
+        # Cannot start container <name>: Bind for <ip>:<port> failed:
+        # port is already allocated
+        #
+        # because Docker can't make up its mind about which format to use.
+        parts = apierror.explanation.split(b": ")
```
wallrj: :question: So doesn't docker-py give a unique error code for address in use?

wallrj:
Thanks @exarkun 

This all looks great.

 * Maybe add a line or two of documentation explaining the `port:0` random port assignment feature.
 * It looks like the API schema allows port 0, great.
 * And it looks like the fig / flocker config parser should also allow port 0
 * How will port:0 interact with the routing feature? An application using port:0 will / may get a different port when it's moved to another node. I suppose we really need to allocate a port that is known to be free on *all* nodes and have that port stick to the application as it is moved between nodes. 
 * But perhaps that can be considered in a followup and for now we consider the port:0 a feature for testing only.

The builds have all passed. Please merge after adding some documentation explaining port:0 and it's current limitations and after creating some follow up issues...if you agree with the notes above.

flocker/node/_docker.py
exarkun: Nope.  Instead, it sucks.

PR #1766, opened 2015-07-13
[FLOC-2640] Display debug information to help diagnose yum network problems

jongiddy:
This PR makes `yum` display details about low-level network operations while building the CentOS Dockerfile. This additional information might be useful next time FLOC-2452 shows up, or for other network problems.  The additional information is sent to stderr, and gets logged near the top of the `flocker-omnibus-centos-7` `build-package` step.


itamarst:
Yay more logs, I guess? Merge when buildbot is green (I'll restart a cuople of builds). I suggest moving parent issue out of sprint until it reoccurs.

PR #1767, opened 2015-07-14
[FLOC-2633] Retry HTTP requests in case it takes a while for the HTTP server to start up

itamarst:
Does seem to fix the issue - notice that if I add a print statement for count it sometimes needs to retry:

```
flocker.acceptance.test_api
  ContainerAPITests
    test_linking ... 
1
2
                                                      [OK]

```

### DIFF
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -314,14 +314,20 @@ class ContainerAPITests(TestCase):
         the HTTP server via Flocker in order to check if that particular
         setup succeeded.
 
+        We try three times since it may take a little time for the HTTP
+        server to start up.
+
         :param bytes host: Host to connect to.
         :param int port: Port to connect to.
         """
-        def query(host, port):
+        def query(host, port, count=0):
             req = get(
                 "http://{host}:{port}".format(host=host, port=port),
                 persistent=False
-            ).addCallback(content)
+            )
+            req.addCallback(content)
+            if count < 3:
+                req.addErrback(lambda _: query(host, port, count + 1))
             return req
 
         d = verify_socket(host, port)
--- a/flocker/acceptance/test_api.py
+++ b/flocker/acceptance/test_api.py
@@ -325,7 +325,12 @@ class ContainerAPITests(TestCase):
                 "http://{host}:{port}".format(host=host, port=port),
                 persistent=False
             )
-            req.addCallbacks(content, lambda _: False)
+
+            def failed(failure):
+                Message.new(message_type=u"acceptance:http_query_failed",
+                            reason=unicode(failure)).write()
+                return False
+            req.addCallbacks(content, failed)
             return req
 
         d = verify_socket(host, port)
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks @itamarst. If I'm reading this correctly this is relying on an HTTP server taking less time to start up than it takes to make three `get` requests. That doesn't seem particularly safe. Why three? I guess that in manual test there have been no failures with three `get`s. This seems more suited to `flocker.testtools.loop_until`.

In the past where we have had to wait up to 120 seconds to see a `TimeoutError`. Perhaps `loop_until` should have a `max_time` parameter or similar if you think that the fail-quick behaviour of a `count=3` loop here is valuable.

Please change this to use `loop_until`, or explain why not and use something else less race-y than three get requests vs a server starting up, and resubmit.

itamarst:
I *thought* there was a timeout within the HTTP client that was long enough that this should be enough time, but having hard time finding it...

jongiddy:
The interesting question is why is `verify_socket` succeeding, but `query(host, port)` failing?

The `verify_socket` call succeeds twice before the treq request is performed (once in `test_linking` and once in `assert_busybox_http`.

One possibility is that the `verify_socket` calls fill the listen queue at the remote end, and the treq request overtakes the connection close messages.

itamarst:
Yep, that'd be my guess. The http error if you *don't* retry is ResponseNeverReceived.

flocker/acceptance/test_api.py
```diff
@@ -314,6 +314,9 @@ def assert_busybox_http(self, host, port):
         the HTTP server via Flocker in order to check if that particular
         setup succeeded.
 
+        We try three times since it may take a little time for the HTTP
+        server to start up.
```
jongiddy: Remove this comment, now that we try until it succeeds.

flocker/acceptance/test_api.py
```diff
@@ -321,11 +324,12 @@ def query(host, port):
             req = get(
                 "http://{host}:{port}".format(host=host, port=port),
                 persistent=False
-            ).addCallback(content)
+            )
+            req.addCallbacks(content, lambda _: False)
```
jongiddy: Log the reason for failure, so we can diagnose persistent failures (that will otherwise just display a timeout message).

itamarst: You can tell that the request bit was started from the `loop_until` initial log message, so I'll just stick to logging the error.

wallrj:
> In the past where we have used loop_until we have had to wait up to 120 seconds to see a TimeoutError. Perhaps loop_until should have a max_time parameter or similar if you think that the fail-quick behaviour of a count=3 loop here is valuable.

Here's the issue for adding a time limit to `loop_until`.
 * https://clusterhq.atlassian.net/browse/FLOC-2621

wallrj:
Thanks @itamarst 

This all looks fine. 

I removed the obsolete comment that @jongiddy pointed out and once the builds pass I'll merge this to unblock the stacked branches.

itamarst:
OMG 5 hours end-to-end build.

PR #1768, opened 2015-07-14
[FLOC-2543] Continue on volume errors

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-2643

I wrote a test that attempts to attach two loopback block devices but raises an exception during the first attachment.
Unfortunately, this test passes. 
I need to look harder at the code and think about a better way to simulate the 'openstack volume in "error" state situation.'

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -3266,6 +3266,47 @@ class AttachVolumeTests(
             DatasetWithoutVolume(dataset_id=dataset_id), failure.value
         )
 
+    def test_run_error(self):
+        """
+        ``AttachVolume.run``
+        """
+        host = u"192.0.7.8"
+
+        deployer = create_blockdevicedeployer(self, hostname=host)
+        api = deployer.block_device_api
+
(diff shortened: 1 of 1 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3266,6 +3266,54 @@ def test_missing(self):
...
+
+        def fake_attach_volume(blockdevice_id, attach_to):
+            if blockdevice_id == volume1.blockdevice_id:
+                raise Exception(
+                    'Block device cannot be attached', blockdevice_id
+                )
+            return original_attach_volume(blockdevice_id, attach_to)
+
+        self.patch(api, 'attach_volume', fake_attach_volume)
+        changes = in_parallel([change1, change2])
```
itamarst: Note that "in_parallel" does a sort internally, so maybe disable and see if that causes breakage?

itamarst:
Unless that change manages to reproduce it, as discussed I suggest closing this and the issue and seeing if it ever happens again, since we can't figure it out.

jml:

PR #1770, opened 2015-07-15
Devbox release floc 2519

adamtheturtle:
This is a design for https://clusterhq.atlassian.net/browse/FLOC-2519.

I did a release for `1.2.0.dev1` using these instructions. I got stuck at the homebrew push phase, and it turned out that this was because of a git hook I had. I made the changes in `release.py` here and used the `--flocker-version` option (because the version was then dirty) and it all finished.

This includes attempts to:
* Provide instructions for installing the prerequisites which were installed in the dev box onto the host
* Do all the work on the host
* Remove permanent changes (e.g. use a temporary environment, don't change the git config, stop using an environment variable throughout because it might get overwritten and we don't want to write to .bashrc). These didn't matter in a temporary Vagrant box.


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -70,49 +86,26 @@ Preparing For a Release
...
-   Note that the ssh key you use must be linked to your GitHub account.
+#. Create an environment to do a release in:
 
-   .. prompt:: bash $
+   .. prompt:: bash $,(flocker-release)tmp-5234cf9a4c8fad37$,(flocker-release)flocker$ auto
 
-      [ -e "${SSH_AUTH_SOCK}" ] || eval $(ssh-agent)
-      ssh-add $HOME/.ssh/id_rsa
-
-   This copies your local git configuration from ``~/.gitconfig``.
```
jongiddy: :+1: for removing this ambiguous referent

adamtheturtle: I thought about keeping it just to solidify its status as ambiguous.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -36,17 +33,36 @@ Prerequisites
...
+   brew install createrepo dpkg
+
+Ubuntu
+******
 
-  .. prompt:: bash $
+.. prompt:: bash $
 
-     vagrant plugin install vagrant-scp
+   apt-get update
```
jongiddy: These commands and the `yum` commands require `sudo`.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -70,49 +86,26 @@ Preparing For a Release
...
-#. Create and log in to a new :doc:`Flocker development machine <vagrant>`:
-
-   This uses SSH agent forwarding so that you can push changes to GitHub using the keys from your workstation.
-
-   Add your SSH key to the ``sshd`` agent.
-   Note that the ssh key you use must be linked to your GitHub account.
+#. Create an environment to do a release in:
 
-   .. prompt:: bash $
+   .. prompt:: bash $,(flocker-release)tmp-5234cf9a4c8fad37$,(flocker-release)flocker$ auto
```
jongiddy: :question: For the purposes of making the docs less confusing, maybe halve the size of the temporary identifier

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: Ah thanks, I was doing this in a Docker container and didn't need that but I guess that isn't the normal case.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -70,49 +86,26 @@ Preparing For a Release
...
-
-      git clone git@github.com:ClusterHQ/flocker.git "flocker-${VERSION}"
-      cd flocker-${VERSION}
-      vagrant up
-      vagrant scp default:/home/vagrant/.bashrc vagrant_bashrc
-      echo export VERSION=${VERSION} >> vagrant_bashrc
-      vagrant scp vagrant_bashrc /home/vagrant/.bashrc
-      if [ -d ~/.aws ]; then vagrant scp "~/.aws" /home/vagrant; fi
-      vagrant ssh -- -A
+      $ mktmpenv --prompt="(flocker-release)"
```
jongiddy: Why the change to `mktmpenv`?

adamtheturtle: I put in the description:

> Remove permanent changes (e.g. use a temporary environment

I think that it means that the Python dependencies, and the files created in the temporary environment (the git clone) are all removed when the environment is deactivated. In the Vagrant box, permanent changes didn't matter, because the box could be destroyed. Here, I wanted there to be as few permanent changes as possible.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -70,49 +86,26 @@ Preparing For a Release
...
+      # The following command means that you will not be asked whether
+      # you want to continue connecting
+      (flocker-release)tmp-5234cf9a4c8fad37$ ssh-keyscan github.com >> ~/.ssh/known_hosts
+      (flocker-release)tmp-5234cf9a4c8fad37$ git clone git@github.com:ClusterHQ/flocker.git
+      (flocker-release)tmp-5234cf9a4c8fad37$ cd flocker
+      (flocker-release)flocker$ pip install --editable .[release]
```
jongiddy: `.[dev]` is the only supported target now.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -70,49 +86,26 @@ Preparing For a Release
...
 
-#. Create a release branch, and create and activate a virtual environment:
+#. Create a release branch, using the version number of the release being created, e.g.:
```
jongiddy: This step seems to duplicate parts of the previous step?  Is this section meant to be deleted, or are we doing things twice?

jongiddy:
Once the Vagrant box is out of the way, I'd prefer the simplicity of setting the VERSION environment variable, and setting other things (the branch name) off that.  Can you explain further why this would be a problem?

The subtraction of the prefix from the branch name seems somewhat fragile (if it fails due to some change to the branch name, accidental or policy-led, the very important tagging of the release could go very wrong).

jongiddy:
Thanks. I'd like to see the replies to these comments before pushing this forward to coding.

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: This is a mistake (bad conflict resolution?) - I'll fix it now.

adamtheturtle:
> Once the Vagrant box is out of the way, I'd prefer the simplicity of setting the VERSION environment variable, and setting other things (the branch name) off that. Can you explain further why this would be a problem?

With the Vagrant system, we stored the environment variable to `~/.bashrc`. I don't think that it is appropriate to do that here, on an engineer's host system. The reason that the environment variable was stored was that it became a pain to not store it - the release process took time, and has gaps (review, long builds etc.). Environment variables are prone to being lost across this long period of time, e.g. if the engineer closes their terminal window. Maybe we should store the version in a better way, but I don't like the environment variable only system. Maybe a file?

jongiddy:
OK to proceed with coding

adamtheturtle:
The next step is:
* Do a dev release following the instructions in this branch
* Change anything which comes up as an issue when doing that
* Submit for review

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -75,60 +91,22 @@ Preparing For a Release
...
-      mkvirtualenv flocker-release
-      pip install --editable .[dev]
-      admin/create-release-branch --flocker-version="${VERSION}"
-      admin/update-license
-      git commit -am "Updated copyright in LICENSE file"
+      $ mktmpenv --prompt="(flocker-release)"
+      (flocker-release)tmp-1234$ git clone git@github.com:ClusterHQ/flocker.git
+      (flocker-release)tmp-1234$ cd flocker
+      (flocker-release)flocker$ pip install --editable .[dev]
+      (flocker-release)flocker$ admin/create-release-branch --flocker-version="${VERSION}
```
jongiddy: How did the VERSION environment variable get set here? There is no longer a section saying to set this.

jongiddy: Also, missing closing double-quote.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -75,60 +91,22 @@ Preparing For a Release
...
-      # you want to continue connecting
-      ssh-keyscan github.com >> ~/.ssh/known_hosts
-      git clone git@github.com:ClusterHQ/flocker.git
-      cd flocker
-      mkvirtualenv flocker-release
-      pip install --editable .[dev]
-      admin/create-release-branch --flocker-version="${VERSION}"
-      admin/update-license
-      git commit -am "Updated copyright in LICENSE file"
+      $ mktmpenv --prompt="(flocker-release)"
```
jongiddy: The use of `mktempenv` makes it difficult to deal with the gap during tests.  I know you want to remove `$VERSION` from the second part of the release, to avoid the user having to keep it set.  However, in my case, I needed to reboot for updates, and then had to find the name of the temporary virtualenv to reconnect.  Luckily, I only had one tempenv to choose from. But having to remember the tempenv name is as bad as having to ensure `VERSION` is always set.  I suggest just going to a named virtualenv, probably `flocker-${VERSION}` since `VERSION` needs to be set for this stage of the release.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -75,60 +91,22 @@ Preparing For a Release
...
+      (flocker-release)tmp-1234$ git clone git@github.com:ClusterHQ/flocker.git
```
jongiddy: Also, suggest naming the cloned repo using the format `flocker-${VERSION}` as confirming which directory contains this clone is important too.

adamtheturtle: Yes. That was part of the reasoning for the temporary environment - it took you to a temporary location. I'll figure something out.

admin/release.py
```diff
@@ -409,7 +409,12 @@ def publish_homebrew_recipe(homebrew_repo_url, version, source_bucket,
     # Sometimes this raises an index error, and it seems to be a race
     # condition. There should probably be a loop until push succeeds or
     # whatever condition is necessary for it to succeed is met. FLOC-2043.
-    push_info = homebrew_repo.remotes.origin.push(homebrew_repo.head)[0]
+    push_info = homebrew_repo.remotes.origin.push(
+        homebrew_repo.head,
+        # Ignore any hooks which might prevent pushing (to master in this
+        # case). Without this, the release process can hang.
+        no_verify=True,
```
adamtheturtle: There are three options for this wrt testing:
* Convince `git` to use some particular fake hook in the test, test that the code in the hook is not run,
* Use a spy for push to check that the `no_verify` option is used
* Don't test this

I went for the latter as a judgement call of the tradeoff between effort and value

jml: :heavy_check_mark:

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -75,60 +91,29 @@ Preparing For a Release
...
 
    .. prompt:: bash $
 
-      export VERSION=0.1.2
+      VERSION=0.1.2
 
 #. Create an issue in JIRA:
 
-   This should be an "Improvement" in the current sprint, with "Release Flocker $VERSION" as the title, and it should be assigned to yourself.
+   This should be an "Improvement" in the current sprint, with "Release Flocker ${VERSION}" as the title, and it should be assigned to yourself.
```
jml: Please drop "in the current sprint".

jml:
Only one minor change suggested. Please merge once you've made it and tests pass. If fixing the tests require non-trivial changes, please bounce back for further review.

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: I had that lined up as a follow up but will do it here :smile:

adamtheturtle: Might as well change "Improvement" to "Feature" while I'm here.

adamtheturtle:
Thanks, I think that the failing builds were unrelated and hope that they will pass next time round.

PR #1778, opened 2015-07-17
[FLOC-2648] Restart rsyslogd after installing node package

jongiddy:
Restart rsyslog service after installing the configuration file that prevents logs going to `/var/log/messages`.

A simple way to test this is to start a CentOS 7 VM. Check the status of rsyslogd using `/bin/systemctl status rsyslog.service`. This will display the time that the service became active.

Create the file `/etc/yum.repos.d/clusterhq-build.repo`:
```
[clusterhq-build]
name=clusterhq-build
baseurl=http://build.clusterhq.com/results/omnibus/rsyslog-reconfig-on-install-FLOC-2648/centos-7
gpgcheck=0
enabled=1
```
Run `sudo yum -y install clusterhq-flocker-node`

Check that `/bin/systemctl status rsyslog.service` now shows a new started time.

(For a null example, replace `rsyslog-reconfig-on-install-FLOC-2648` with `master` in the `.repo` file.)


itamarst:
This is not the ultimate solution, insofar as apparently we can't rely on journalctl in CentOS 7. On the other hand it's definitely better than current situation.

itamarst:
Verified that this works correctly on CentOS 7.

itamarst:
Probably failures above were spurious, so I will merge forward. If tests pass then merge.

PR #1779, opened 2015-07-17
[FLOC-2671] Release/flocker 1.1.0dev1

zubron:
Waiting for tests to pass due to "no space" problem.
There is one error with the link check:
```
reference/api.rst:1: [redirected with Found] https://clusterhq.atlassian.net/browse/FLOC-2044 to https://clusterhq.atlassian.net/login?dest-url=%2Fbrowse%2FFLOC-2044&permission-violation=true
```
This is a JIRA link redirecting to the login page so will happen with all JIRA links. Does the filter need to be updated?

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -10,6 +10,14 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+Next Release
+============
+
+* The flocker-deploy command supports specification of the pathnames of certificate and key files.
+  See :ref:`flocker-deploy-authentication`.
+* The agent configuration file allows specification of a CA certificate for OpenStack HTTPS verification.
+  See :ref:`openstack-dataset-backend`.
+
 v1.0.3
 ======
 
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -13,7 +13,7 @@ You can learn more about where we might be going with future releases by:
 Next Release
 ============
 
-* The flocker-deploy command supports specification of the pathnames of certificate and key files.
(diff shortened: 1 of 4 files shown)


adamtheturtle:
@zubron Is that a local error? The docs build is successful on Buildbot. Also, you can write ``` :issue:`2044` ``` and it'll work and be neater.

docs/releasenotes/index.rst
```diff
@@ -10,6 +10,14 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+Next Release
+============
+
+* The flocker-deploy command supports specification of the pathnames of certificate and key files.
```
adamtheturtle: :question: Use ``` ``flocker-deploy`` ``` with backpacks like elsewhere in this document.

adamtheturtle:
The `NEWS` file and the release notes changes are suitable (with one optional comment).

adamtheturtle:
The review process says:

> The build should be passing to the team's satisfaction.

Usually with a dev release some tests don't really matter - we are testing the release process and not the software (most of the time), but in this case the packages are not being built, so the release scripts will fail. This release should therefore, unfortunately, be blocked on the build issues being fixed. Or if rebuilding works, do that and resubmit.

adamtheturtle:
I recon that this should be closed and perhaps we should do 1.1.0dev2. This highlights a release process problem - someone must remember in these failure cases to transfer the release notes / NEWS changes here to the next release. I'll leave what to do with this to you @zubron . Maybe try re-running the test suite but IIRC normal branches are merged with master before BuildBot runs the suite (removing some of the unrelated test failures) but release branches aren't.

zubron:
Sure, I can close this and make sure the release notes are added to the 1.1.0dev2 release. The docs issue above should no longer be there as the link in question has been made available without a log in. There have been some changes made to the AMIs for building so hopefully we shouldn't run into issues caused by no disk space.

PR #1780, opened 2015-07-17
[FLOC 2657] Move node installation steps into standalone doc

davidoliverSP2:
Fixes 2657.

Note, there is no new content here. The content in install-node.rst (and the install-client.rst created in 2656) was split from the old install/index.rst file.

This is a sub-task of 2644, which has 2 other sub-tasks which handles the rest of the breakdown of the install.rst.


jongiddy:
The rump `install.rst` file should now be entitled "Post-Installation Configuration" with "Configuring Cluster Authentication" as the first sub-section and "Enabling the Flocker control service" as the second sub-section.

The `install-node.rst` file is fine, but the Next Step should say something like, "In the next step, the node control and agent services will be configured and started."

docs/install/install-node.rst
```diff
@@ -0,0 +1,191 @@
...
+   :prompt: [root@ubuntu]#
+
+Finally, you will need to run the ``flocker-ca`` tool that is installed as part of the CLI package.
+This tool generates TLS certificates that are used to identify and authenticate the components of your cluster when they communicate, which you will need to copy over to your nodes.
+Please continue onto the next section, with the cluster authentication instructions.
+
+Next Step
+=========
+
+In the next step :ref:`the node control and agent services will be configured and started.<authentication>`.
```
itamarst: This link seems wrong? Ought to point to post-install?

docs/install/post-install.rst
```diff
@@ -577,7 +394,7 @@ To support moving data with the ZFS backend, every node must be able to establis
 So ensure that the firewall allows access to TCP port 22 on each node from the every node's IP addresses.
 
 Next Step
----------
+=========
 
 The next section describes your next step - setting up an :ref:`authenticated user<authenticate>`.
```
itamarst: Not actually next section in docs, but probably out of scope for this issue.

itamarst:
Looks good, address above if needed then merge.

docs/install/post-install.rst
davidoliverSP2: yep, the next subtask in this story (FLOC 2658) will handle this.

PR #1781, opened 2015-07-17
switch ZFS volume backend to pyzfs / libzfs_core

avg-I:
Please **DO NOT MERGE** yet.
At the moment this pull request and branch are for reviewing only.

Some special actions are needed to test this branch.
ZFS kernel module and userland components should be built and installed using source code in the following branch: https://github.com/ClusterHQ/zfs/tree/libzfs_core-wip
pyzfs / libzfs_core has to be installed from the following branch: https://github.com/ClusterHQ/pyzfs/tree/pyzfs-0.2

### DIFF
--- a/flocker/volume/filesystems/zfs.py
+++ b/flocker/volume/filesystems/zfs.py
@@ -9,6 +9,7 @@ from __future__ import absolute_import
 import os
 import libzfs_core
 
+from functools import wraps
 from contextlib import contextmanager
 from uuid import uuid4
 from subprocess import (
@@ -24,8 +25,12 @@ from eliot import Field, MessageType, Logger
 from twisted.python.failure import Failure
 from twisted.python.filepath import FilePath
 from twisted.internet.endpoints import ProcessEndpoint, connectProtocol
+from twisted.internet.interfaces import IReactorThreads
 from twisted.internet.protocol import Protocol
-from twisted.internet.defer import Deferred, succeed, gatherResults
+from twisted.internet.defer import (
+    Deferred, succeed, gatherResults, maybeDeferred
+)
+from twisted.internet.threads import deferToThreadPool
 from twisted.internet.error import ConnectionDone, ProcessTerminated
 from twisted.application.service import Service
 
@@ -67,6 +72,64 @@ class _AccumulatingProtocol(Protocol):
         else:
             self._result.errback(reason)
         del self._result
+
+
+class _AsyncLZC(object):
+    def __init__(self, reactor):
+        self._reactor = reactor
+        self._cache = {}
+
+    def callDeferred(self, func, *args, **kwargs):
+        return deferToThreadPool(self._reactor, self._reactor.getThreadPool(),
+                                 func, *args, **kwargs)
+
+    def __getattr__(self, name):
+        try:
+            return self._cache[name]
+        except KeyError:
+            func = getattr(libzfs_core, name)
+
+            @wraps(func)
+            def _async_wrapper(*args, **kwargs):
+                return self.callDeferred(func, *args, **kwargs)
+
+            self._cache[name] = _async_wrapper
+            return self._cache[name]
+
+
+class _FakeAsyncLZC(object):
+    def __init__(self):
+        self._cache = {}
+
+    def callDeferred(self, func, *args, **kwargs):
+        return maybeDeferred(func, *args, **kwargs)
+
+    def __getattr__(self, name):
+        try:
+            return self._cache[name]
+        except KeyError:
+            func = getattr(libzfs_core, name)
+
+            @wraps(func)
+            def _async_wrapper(*args, **kwargs):
+                return maybeDeferred(func, *args, **kwargs)
+
+            self._cache[name] = _async_wrapper
+            return self._cache[name]
+
+
+_reactor_to_alzc = {}
+
+
+def _async_lzc(reactor):
+    try:
+        return _reactor_to_alzc[reactor]
+    except KeyError:
+        if IReactorThreads.providedBy(reactor):
+            _reactor_to_alzc[reactor] = _AsyncLZC(reactor)
+        else:
+            _reactor_to_alzc[reactor] = _FakeAsyncLZC()
+        return _reactor_to_alzc[reactor]
 
 
 def zfs_command(reactor, arguments):
--- a/flocker/volume/filesystems/zfs.py
+++ b/flocker/volume/filesystems/zfs.py
@@ -585,16 +585,12 @@ class StoragePool(Service):
 
     def set_maximum_size(self, volume):
         filesystem = self.get(volume)
-        properties = []
         if volume.size.maximum_size is not None:
-            properties.extend([
-                u"refquota={0}".format(
-                    volume.size.maximum_size).encode("ascii")
-            ])
+            requota = volume.size.maximum_size
         else:
-            properties.extend([u"refquota=none"])
-        d = zfs_command(self._reactor,
-                        [b"set"] + properties + [filesystem.name])
+            # zero means no quota
+            requota = 0
+        d = self._async_lzc.lzc_set_prop(filesystem.name, b"refquota", requota)
         d.addErrback(self._check_for_out_of_space)
         d.addCallback(lambda _: filesystem)
         return d
--- a/flocker/volume/filesystems/zfs.py
+++ b/flocker/volume/filesystems/zfs.py
@@ -129,6 +129,26 @@ def _async_lzc(reactor):
         return _reactor_to_alzc[reactor]
 
 
+def ext_command(reactor, arguments):
+    """
+    Asynchronously run the given command-line tool with the given arguments.
+
+    :param reactor: A ``IReactorProcess`` provider.
+
+    :param arguments: A ``list`` of ``bytes``, the command and command-line
+    arguments.
+
+    :return: A :class:`Deferred` firing with the bytes of the result (on
+        exit code 0), or errbacking with :class:`CommandFailed` or
+        :class:`BadArguments` depending on the exit code (1 or 2).
+    """
+    endpoint = ProcessEndpoint(reactor, arguments[0], arguments,
+                               os.environ)
+    d = connectProtocol(endpoint, _AccumulatingProtocol())
+    d.addCallback(lambda protocol: protocol._result)
+    return d
+
+
 def zfs_command(reactor, arguments):
     """
     Asynchronously run the ``zfs`` command-line tool with the given arguments.
@@ -142,11 +162,7 @@ def zfs_command(reactor, arguments):
         exit code 0), or errbacking with :class:`CommandFailed` or
         :class:`BadArguments` depending on the exit code (1 or 2).
     """
-    endpoint = ProcessEndpoint(reactor, b"zfs", [b"zfs"] + arguments,
-                               os.environ)
-    d = connectProtocol(endpoint, _AccumulatingProtocol())
-    d.addCallback(lambda protocol: protocol._result)
-    return d
+    return ext_command(reactor, [b"zfs"] + arguments)
 
 
 _ZFS_COMMAND = Field.forTypes(
@@ -568,7 +584,7 @@ class StoragePool(Service):
             ], defer=False)
         d.addCallback(got_snapshots)
         d.addCallback(
-            lambda _: zfs_command(self._reactor, [b"umount", filesystem.name]))
+            lambda _: ext_command(self._reactor, [b"umount", filesystem.name]))
         d.addCallback(
             lambda _: self._async_lzc.lzc_destroy(filesystem.name))
         return d
@@ -601,7 +617,7 @@ class StoragePool(Service):
     def change_owner(self, volume, new_volume):
         old_filesystem = self.get(volume)
         new_filesystem = self.get(new_volume)
-        d = zfs_command(self._reactor,
+        d = ext_command(self._reactor,
                         [b"umount", old_filesystem.name])
         d.addCallback(lambda _: self._async_lzc.lzc_rename(
             old_filesystem.name, new_filesystem.name))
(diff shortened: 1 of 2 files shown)


flocker/volume/filesystems/zfs.py
```diff
@@ -7,11 +7,13 @@
 from __future__ import absolute_import
 
 import os
+import libzfs_core
```
wallrj: I guess we'll need to think about how to specify this dependency.
Simplest thing would be to add it to `requirements.txt` but first we'll have to make sure that `libzfs_core` is available from PyPI.
Perhaps it's better to have some separate `requirements.txt` files for each of the Flocker dataset backend plugins.

wallrj: We currently drag in all the `requirements.txt` in our "other" linux flocker-cli instructions which means that installing flocker-ca involves downloading all the AWS boto, OpenStack Nova / Cinder / Keystone libraries and now libzfs_core and its dependencies...which seems excessive. I wonder how other projects deal with this?

flocker/volume/filesystems/zfs.py
```diff
@@ -67,26 +71,100 @@ def connectionLost(self, reason):
         del self._result
 
 
-def zfs_command(reactor, arguments):
+class _AsyncLZC(object):
```
wallrj: I know it's early days, but a doc string here would help me understand the design of this wrapper.

Instead of wrapping every libzfs_core function with `_async_wrapper` consider creating an `ILibZFSCore` Zope Interface and using `auto_threaded` from https://github.com/ClusterHQ/flocker/blob/master/flocker/common/_thread.py

Or perhaps you'll have to create a slightly higher level abstraction in order to use that. eg `IZFSClient` which is wrapped to give `IZFSClientAsync`. That's how @exarkun organised the `IBlockDeviceAPI` .

flocker/volume/filesystems/zfs.py
```diff
@@ -67,26 +71,100 @@ def connectionLost(self, reason):
...
+    def __init__(self, reactor):
+        self._reactor = reactor
+        self._cache = {}
+
+    def callDeferred(self, func, *args, **kwargs):
```
wallrj: Why not just move this into the nested _async_wrapper function below?
And it seems like it should be private?

flocker/volume/filesystems/zfs.py
```diff
@@ -67,26 +71,100 @@ def connectionLost(self, reason):
...
+
+            @wraps(func)
+            def _async_wrapper(*args, **kwargs):
+                return self.callDeferred(func, *args, **kwargs)
+
+            self._cache[name] = _async_wrapper
+            return self._cache[name]
+
+
+class _FakeAsyncLZC(object):
```
wallrj: Maybe move the Fake to a `flocker.volume.zfs.testtools` module.

flocker/volume/filesystems/zfs.py
```diff
@@ -67,26 +71,100 @@ def connectionLost(self, reason):
...
+                return maybeDeferred(func, *args, **kwargs)
+
+            self._cache[name] = _async_wrapper
+            return self._cache[name]
+
+
+_reactor_to_alzc = {}
+
+
+def _async_lzc(reactor):
```
wallrj: And this looks like a function that'd only be used in tests...returning FakeAsyncLZC for synchronous unit tests and AsyncLZC for async functional tests.

flocker/volume/filesystems/zfs.py
```diff
@@ -241,39 +289,43 @@ def reader(self, remote_snapshots=None):
...
+        d = self._async_lzc.callDeferred(send_and_close)
+        d.addBoth(lambda _: None)
         try:
-            yield process.stdout
+            yield out
         finally:
-            process.stdout.close()
-            process.wait()
+            out.close()
+            queue.get()
```
wallrj: I may be misunderstanding but it seems this is mixing up Twisted style async with the alternative / standard Python threading. To make this truly async the method needs to return the deferred.
And in that case, there's no need to use the queue for signalling that the thread has completed. Twisted threadpool will do that for you and fire the deferred when complete.

flocker/volume/filesystems/zfs.py
```diff
@@ -289,33 +341,46 @@ def writer(self):
...
+
+        (rfd, wfd) = os.pipe()
+        wfile = os.fdopen(wfd, "w")
+        queue = Queue()
+
+        def recv_and_close():
+            try:
+                libzfs_core.lzc_receive(self.name, rfd, force)
+                success = True
+            except Exception:
```
wallrj: We'll lose the details of the Exception. And it'll be an informative libzfs exception, right? So it seems a shame hide it.

flocker/volume/filesystems/zfs.py
```diff
@@ -289,33 +341,46 @@ def writer(self):
...
+            try:
+                wfile.close()
+            except:
+                pass
+            succeeded = queue.get()
+            if succeeded and not force:
+                # a new filesystem
+                libzfs_core.lzc_set_prop(self.name, b"mountpoint",
+                                         self._mountpoint.path)
+                check_call([b"zfs", b"mount", self.name])
```
wallrj: Is there a libzfs API call for this?

flocker/volume/filesystems/zfs.py
```diff
@@ -289,33 +341,46 @@ def writer(self):
...
+                libzfs_core.lzc_receive(self.name, rfd, force)
+                success = True
+            except Exception:
+                success = False
+            finally:
+                os.close(rfd)
+                queue.put(success)
+
+        d = self._async_lzc.callDeferred(recv_and_close)
+        d.addBoth(lambda _: None)
```
wallrj: This seems to be the same as the reader / sender method above in that it uses the Twisted threadpool but then blocks until the thread has completed rather than returning the deferred.
Perhaps I'm missing some context. Perhaps we're working toward a fully async implementation of the ZFS backend in...but in steps. If so it'd be useful to read a few docstrings and comments explaining the plan.

wallrj:
@avg-I I only got halfway through, but I left some comments on what I read.

I know that libzfs_core is intended to be a low level wrapper. Perhaps this branch can contain a rough first version of higher level Sync and Async APIs with just enough Zope Interface methods to support Flocker. And tests and functional tests that exercise that API. Later that can be moved back into libzfs or into a separate external package.

flocker/volume/filesystems/zfs.py
itamarst: We could perhaps have a zfs-requirements.txt and corresponding separate OS packages you need to install, so e.g. `clusterhq-flocker-zfs` if you want ZFS backend.

avg-I: Publishing libzfs_core on PyPI is the plan.
I think that we should do that after its review is completed.

flocker/volume/filesystems/zfs.py
avg-I: This sounds a little bit too advanced for my skills.
I can try to do that or it can be deferred to a more skilled developer.
In either case I will work on this item and see what I can do. At the very least I'll add documentation.

flocker/volume/filesystems/zfs.py
avg-I: It's actually not private and is used in several places. The missing documentation would help, of course.
It turned out that in some cases it was insufficient to call a single `lzc_` function in the asynchronous fashion.
For example, `_list_snapshots` calls `_do_list_snapshots` via `callDeferred`.
Perhaps there was a better way to do that...

flocker/volume/filesystems/zfs.py
avg-I: Okay.

flocker/volume/filesystems/zfs.py
avg-I: The missing documentation would help here as well :-(
The main purpose of this function is to associate an `AsyncLZC` instance with a given reactor and cache it.

flocker/volume/filesystems/zfs.py
avg-I: You are absolutely correct, but the behavior is dictated by `IFilesystem` interface:
> Context manager that allows reading the contents of the filesystem.
> A blocking API, for now.

So, given the scope of this PR I had to jump through some hoops to preserve the existing behavior of this function and to conform to the interface.

Perhaps, there is a JIRA issue to change the interface? @itamarst @exarkun

flocker/volume/filesystems/zfs.py
avg-I: Maybe I could pass the exception through the queue and then raise it?
In my defense, the previous code hid any error details as well. But that's a poor defense :-)

flocker/volume/filesystems/zfs.py
avg-I: Same comment as for the `reader` method. The interface dictates this behavior.

flocker/volume/filesystems/zfs.py
avg-I: No. Please see https://clusterhq.atlassian.net/browse/FLOC-2632.

avg-I:
@wallrj your suggestion about having a higher level wrapper makes sense.
I think that I need some help designing it, so that it is sufficiently convenient both in general and for our ZFS backend specifically.

itamarst:
`IFilesystem` was basically "what we can do as quickly as possible", yeah, not ideal. There's no particular reason it needs to stay this way.

itamarst:
My general thought: if this consistently passes tests and basic comments are addressed, I don't think this needs to be perfect to be merged. The current code after all is not exactly a shining example of quality code. We do need to make sure we file issues in the appropriate epic for the all the necessary improvements (error handling, IFilesystem rewrite) etc. to make this workable.

That being said, we do use this for our demo machines, so we need passing tests (both unit, functional and acceptance) for this to be merged. The acceptance builder is currently in Expected Failures, so you'll have to check it manually on Buildbot.

itamarst:
Since we can't actually proceed with this (even given minimal quality requirements) until we've e.g. done the libzfs_core packaging, I'm going to move this back to Coding state.

flocker/volume/filesystems/zfs.py
avg-I: Looking at this again and it looks like `IFilesystem` does not specify any exception that could be raised in this case.
Also, the only other implementer of the interface, in-memory `DirectoryFilesystem`, does not raise any excepion either. More importantly, none of the functional tests expect an exception from `reader` or `writer` context managers.

So, I think that the refinement of the interface should be a separate issue.
@itamarst, do you agree?

What I can do right now is log the exception. Should that be a condition for merging this PR? Or could that be a separate issue for later?
If I do that now, can I re-use `ZFS_ERROR` message type here?

flocker/volume/filesystems/zfs.py
avg-I: At least, I've added the docstrings now.

flocker/volume/filesystems/zfs.py
avg-I: _FakeAsyncLZC has been completely removed as it has become disused now, even by the tests.

flocker/volume/filesystems/zfs.py
```diff
@@ -241,39 +299,47 @@ def reader(self, remote_snapshots=None):
...
+        out = os.fdopen(rfd)
+        queue = Queue()
+
+        def send_and_close():
+            try:
+                libzfs_core.lzc_send(snapshot, latest_common_name, wfd)
+            except Exception as e:
+                message = ZFS_ERROR(zfs_command="lzc_send " + snapshot,
+                                    output=str(e), status=e.errno)
+                message.write(self.logger)
```
exarkun: I think Eliot's `write_traceback` is probably produces better logging results in this case.  "except Exception" basically means "I dunno what went wrong" so gathering as much information in the logs as possible makes sense.  `write_traceback` should do a better job of that than this `ZFS_ERROR`.

avg-I: Actually only `ZFSError` is expected here (`Exception` is just in case, to not miss the clean-up) and that exception should have enough information.  Adding the traceback may be a good idea anyway.

Generally, it might be a good idea to add Eliot action logging in the same universal wrapper that turns synchronous pyzfs calls into asynchronous ones.  That way, all pyzfs calls and their results would get logged.

avg-I: Although, in this particular piece of code the synchronous API is used directly.

exarkun: > Actually only ZFSError is expected here (Exception is just in case, to not miss the clean-up) and that exception should have enough information. Adding the traceback may be a good idea anyway.

I guess I'm not entirely sure what information it will have.  I know that the exception that used to be handled like this included the output of the `zfs` child process and that was effectively all of the useful information (even though it excluded some Python stack frames having to do with the implementation of running child processes).  Now the exception comes from some cffi-based code and I'm less sure the traceback is irrelevant.  It still could be but I'd like to err on the side of seeing it until we're more confident about that.

Also, mostly separately, on this re-read of the code, I think that this function is being changed so that it no longer actually fails if the zfs send operation has failed.

Prior to the change, if the zfs child process exited with an error code, I think `process.wait()` would have raised an exception and that would have caused this `reader` function to raise an exception.

With the new version of the code, if `lzc_send` raises an exception then it gets logged and then `reader` returns normally - indicating to the caller that the necessary data has been generated.  I think the right thing to do might be to just delete all of this exception handling code and let the exception propagate to the caller of `reader` as it did before?

avg-I: My understanding is that `Popen.wait()` returns a status code and does *not* raise an exception if the status is not success.  In fact, I experimented with letting through the `lzc_send` exceptions and that caused many ZFS functional tests to fail.

exarkun: Yep, you're right.  My mistake.  So this code was terribly broken before.  It'll still be terribly broken in just the same way, now.  But I guess the point of this PR isn't to make the code work, just to make it use pyzfs. :smile: 

So I guess you can ignore the comments about this exception handling and someone will need to come fix this (and the other issues) later.

avg-I: And we probably want to change the IFilesystem reader and write interfaces as well...

PR #1782, opened 2015-07-17
[FLOC-2666] Delete tests that are duplicated by other, better tests.

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-2666


adamtheturtle:
Thanks @itamarst , please remove the comment:

> .. This section is tested in flocker.acceptance.test_moving_applications.MovingApplicationTests.
   Reflect any relevant changes here in those tests.

In `flocker/docs/using/tutorial/moving-applications.rst` and then merge when Buildbot is green.

adamtheturtle:
This one is fairly obvious - a test is being removed so time is saved. However, as there is sometimes some downside to removing a test, which should be weighed against the upside (e.g. taking less time to run the suite) it would be good if metrics (or a link to the relevant metrics pages on buildbot) could be supplied to justify any changes focused on performance.

itamarst:
In general I don't we should ever remove a test purely for performance reasons if it's testing something useful that is not covered by another test.

adamtheturtle:
Sure but consider the case where we have two tests, the first to check that deploying an application works at all and the second to test that with an option changed something is different about the deployed application. The second test covers the first - the second test will always fail if the first does (but not the other way around). Having the first has the advantage that if the second test fails, we can narrow down the cause of the second test's failure. Anyways this is all hypothetical and this change is good.

PR #1785, opened 2015-07-23
[FLOC-2664] Refactor postgress acceptance tests

itamarst:
Besides the fact this is much better in terms of goals, API usage, etc., on my EBS test cluster this takes Postgres tests from 200 seconds to 100 seconds.


flocker/acceptance/testtools.py
```diff
@@ -858,3 +859,34 @@ def clean(cluster):
             return calling_test_method
         return wrapper
     return decorator
+
+
+def create_dataset(test_case, cluster,
+                   maximum_size=REALISTIC_BLOCKDEVICE_SIZE):
```
myechuri: Should we have node number as a parameter? If so, this method could be reused to seed move tests in future? Thanks!

itamarst: Might be useful, yeah, but for now this is just refactoring existing tests. When someone needs that they can add it.

PR #1786, opened 2015-07-23
[FLOC-2709] Begin the switchover to purely Python-based acceptance tests

itamarst:
All internal tests (i.e. not tests for integration with external systems) should be as simple as possible. This is the beginning of the process to simplify everything so it only uses one container image.


myechuri:
Looks like a very useful change for leaner faster tests. Thanks!

itamarst:
test_linking is known failure, to be fixed by #1767.

PR #1787, opened 2015-07-23
[FLOC-2681] Explicit test for flocker-deploy

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2681

All other uses of flocker-deploy in acceptance tests will be going away as other issues in this epic are implemented.


wallrj:
Some tests failed, but it's not obvious why. 
Possibly another example of tests taking longer than the default timeout.

 * http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fcentos-7%2Faws/builds/1126

I'm re-running that builder.

flocker/acceptance/endtoend/test_cli.py
```diff
@@ -0,0 +1,55 @@
...
+                    u"image": u"openshift/busybox-http-app",
+                },
+            },
+        }
+
+        # flocker_deploy() does an assertion that the requested state has
+        # been reached. Once remaining uses of flocker_deploy elsewhere
+        # are removed that code could be moved off of Cluster and into
+        # this module since this is only module we expect will be calling
+        # flocker-deploy.
```
wallrj: :question: Add an issue link.

flocker/acceptance/endtoend/test_cli.py
```diff
@@ -0,0 +1,55 @@
...
+                    # Our config format doesn't support command-lines, so
+                    # we can't use standard image and need to pick
+                    # something small that will not immediately exit.
+                    u"image": u"openshift/busybox-http-app",
+                },
+            },
+        }
+
+        # flocker_deploy() does an assertion that the requested state has
+        # been reached. Once remaining uses of flocker_deploy elsewhere
```
wallrj: :question: Oh, I see, it calls `assert_expected_deployment`. I think the test would be much clearer if we called `assert_expected_deployment` explicitly when `flocker_deploy` has called back.

flocker/acceptance/endtoend/test_cli.py
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        minimal_application = {
+            u"version": 1,
+            u"applications": {
+                name: {
+                    # Our config format doesn't support command-lines, so
+                    # we can't use standard image and need to pick
+                    # something small that will not immediately exit.
+                    u"image": u"openshift/busybox-http-app",
```
wallrj: The "Systems tests" design document says:

> End-to-end tests. These verify the components of the system work together correctly. As such we should minimize reliance on external software, since that introduces irrelevant complication

* So shouldn't this at least use a ClusterHQ controlled image? (like #1281)
* And ideally an image that doesn't have to be pulled from DockerHub. (Maybe call `docker build`  or `docker load`)
* And if neither of those options appeal, perhaps we should at least fix the image version so that upstream changes to the Dockerfile are less likely to break our tests.

wallrj:
Thanks @itamarst 

I'm not quite sure what other test modules are planned for `flocker/acceptance/endtoend`. 
Will `flocker/acceptance/test_api.py` be moved there eventually?
But this looks like a good step forward.
So please merge after addressing or answering the comments above and when you get some green build results.

(Oh, the build results *are* green. Woohoo!)

flocker/acceptance/endtoend/test_cli.py
itamarst: Yeah, ideally we'd use our own image. `build` won't work since we don't support it.

On the other hand we're killing `flocker-deploy` so I'd like to minimize effort spent on this particular test. So I'll just harcode a tag, yeah.

flocker/acceptance/endtoend/test_cli.py
itamarst: Hm. I guess I will just wait until all other uses of flocker-deploy are merged before finishing this branch, and I'll move flocker_deploy and assert_expected_deployment into this module before merging it. Sounds reasonable?

flocker/acceptance/endtoend/test_cli.py
itamarst: We've discussed not filing issues for follow-up cleanups, just leaving comments in code. Or do you mean the existing issues for removing other uses of flocker-deploy?

flocker/acceptance/endtoend/test_cli.py
itamarst: Oh, except I want to delete this whole module. So I'll do that in the *next* branch.

PR #1788, opened 2015-07-23
[FLOC-2667] Swap out old Mongo tests with explicitly integration-oriented tests.

itamarst:


adamtheturtle:
We already have a little bit of confusion between unit/functional/acceptance. Now with acceptance/integration (and elsewhere acceptance/endtoend) I think that there deserves to be documentation about test organisation. This can be a follow-up but I'd prefer not to have these tests available for long without those docs.

I think that perhaps that's part of the reason we have got such a long test suite - acceptance tests being added without full consideration / discussion about when it is best to add such a test.

adamtheturtle:
Thanks @itamarst this looks great.

Potentially important:

* MongoDB is not a static piece of software. It gets updated and its assumptions (e.g. does a volume need to be empty on startup) may change. What is being tested here is one particular version of MongoDB. Is it a popular one? Is it one which was specifically chosen because it didn't interact badly with a ZFS bug (and therefore decreasing the value of the test)?

Not as important:

* Getting multiple client objects - this might be a non-issue but perhaps this test (or others later in a similar style) could be quicker if we somehow got one client object and passed it around.
* Perhaps the interface to `make_dataset_integration_testcase` would be better if it took as a parameter some data to be inserted and then checked for. Else there is repetition of the data.

do with these as you think appropriate. Merge when you're happy or resubmit if there are any major changes.

itamarst:
* I think we should be testing latest version of MongoDB official docker image. But I didn't want to make the switch in this branch since there are other tests that use MongoDB that haven't yet been removed, and I don't know if that'll work with ZFS, etc.. So one change at a time.
* I don't think time to open new client connection really impacts test time (although my internal Jean-Paul simulacra is saying something about "measuring all the things"). I would expect it to be 5ms, say, whereas e.g. AWS attach takes minimum of 10000ms in our current code.
* Data repetition can be fixed by having it as module-level constant. I don't care too much insofar as it's only the two functions. But I'll try to remember to do that for future integration tests.

PR #1791, opened 2015-07-24
[FLOC-2711] Remove busybox from acceptance tests

itamarst:
Also fixes https://clusterhq.atlassian.net/browse/FLOC-2401 as a side-effect.

This depends on #1767 being merged first.

PR #1792, opened 2015-07-24
[FLOC-2665] Simpler traffic routing test

itamarst:
Depends on #1791.


flocker/acceptance/test_api.py
```diff
@@ -382,6 +382,33 @@ def test_linking(self, cluster):
                 origin.public_address, origin_port))
         return running
 
+    @require_cluster(2)
+    def test_traffic_routed(self, cluster):
+        """
+        An application can be accessed even from a connection to a node
+        which it is not running on.
+        """
+        port = 8080
```
wallrj: :question: Is it safe to hard code this to such a common port? Might be better to use the `find_free_port` helper.

wallrj:
Thanks @itamarst 

The new test looks correct and is really neat. 

I forced a rebuild of the tutorial box. Please merge when that and the triggered acceptance tests have passed.

flocker/acceptance/test_api.py
wallrj: Oh ignore that. I already knew that `find_free_port` operates locally and won't help find a cluster-wide free port.

wallrj: https://clusterhq.atlassian.net/browse/FLOC-2401

PR #1794, opened 2015-07-27
Use more stable link to docker volumes plugin docs

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2722

### DIFF
--- a/docs/labs/docker-plugin.rst
+++ b/docs/labs/docker-plugin.rst
@@ -37,8 +37,8 @@ See also the `GitHub repo for this project <https://github.com/ClusterHQ/flocker
     The integration approach allows Docker to control Flocker via the Flocker Dataset API.
     This allows Flocker to be used in conjunction with other ecosystem tools like :ref:`Docker Swarm <labs-swarm>` and :ref:`Docker Compose <labs-compose>`, which is more experimental than ``flocker-deploy`` and the Flocker containers API.
 
-.. _`Docker volumes plugin`: https://github.com/docker/docker/blob/3ee15acaadc3f25c90ea2c4de2e5ab1a12d8e042/docs/extend/plugins_volume.md
-.. _`Docker documentation on volume plugins`: https://github.com/docker/docker/blob/3ee15acaadc3f25c90ea2c4de2e5ab1a12d8e042/docs/extend/plugins_volume.md
+.. _`Docker volumes plugin`: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md
+.. _`Docker documentation on volume plugins`: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md
 
 How it works
 ============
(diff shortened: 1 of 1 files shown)


docs/labs/docker-plugin.rst
```diff
@@ -37,6 +37,8 @@ See also the `GitHub repo for this project <https://github.com/ClusterHQ/flocker
     The integration approach allows Docker to control Flocker via the Flocker Dataset API.
     This allows Flocker to be used in conjunction with other ecosystem tools like :ref:`Docker Swarm <labs-swarm>` and :ref:`Docker Compose <labs-compose>`, which is more experimental than ``flocker-deploy`` and the Flocker containers API.
 
+.. _`Docker volumes plugin`: https://github.com/docker/docker/blob/3ee15acaadc3f25c90ea2c4de2e5ab1a12d8e042/docs/extend/plugins_volume.md
```
jongiddy: While the previous link can break if Docker choose to move files around, this link is forever stuck at this version, and will not change when significant information changes in the target page.

Since we can see the link break in the (hopefully) unusual case of Docker moving files around, why do we want to change to a more error-prone (success-prone?) link which will always succeed, even if we would have wanted it to change?

jongiddy:
There are two broken links pointing to the same location. Why has https://github.com/ClusterHQ/flocker/blame/b149602c30f7fa4f2f682032ac14a29a99e8f4ff/docs/labs/docker-plugin.rst#L25 been left unchanged?

jongiddy:
I would suggest simply fixing both links to point to the new location. This always shows the most recent page, and will trigger a warning in our tests if Docker break the link.

The current solution will always work, but requires us to remember to update our docs when the Docker page is updated.

adamtheturtle:
My thought was that this link was sent from @binocarlos and that engineering is not responsible for labs content, and I just want the tests to pass. But you are right and I have change both links to point to the newest location on master.

docs/labs/docker-plugin.rst
```diff
@@ -37,6 +37,9 @@ See also the `GitHub repo for this project <https://github.com/ClusterHQ/flocker
     The integration approach allows Docker to control Flocker via the Flocker Dataset API.
     This allows Flocker to be used in conjunction with other ecosystem tools like :ref:`Docker Swarm <labs-swarm>` and :ref:`Docker Compose <labs-compose>`, which is more experimental than ``flocker-deploy`` and the Flocker containers API.
 
+.. _`Docker volumes plugin`: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md
+.. _`Docker documentation on volume plugins`: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md
```
jongiddy: If these are meant to identical links, avoid the repetition by using: 
```rst
.. _`Docker documentation on volume plugins`: `Docker volumes plugin`_
```

jongiddy:
OK, one further comment to address, and I am happy for you to merge when green.

docs/labs/docker-plugin.rst
adamtheturtle: Ah tried that and my editor gave me an error - probably a bad editor plugin.

adamtheturtle: Thanks.

PR #1795, opened 2015-07-27
[FLOC-2685] Get rid of flocker-deploy based size movement test

itamarst:
Turns out the existing dataset movement test already covered this, so I didn't add a new test, but I did make it more obvious (and perhaps better tested).


flocker/acceptance/test_deployment.py
```diff
@@ -1,191 +0,0 @@
...
-        deploy_configuration['volume'] = volume
-    return deploy_configuration
-
-
-class DeploymentTests(TestCase):
-    """
-    Tests for deploying applications.
-
-    Similar to:
-    http://doc-dev.clusterhq.com/gettingstarted/tutorial/
```
adamtheturtle: Remove the comment:

> .. This section is tested in flocker.acceptance.test_deployment.DeploymentTests.
   Reflect any relevant changes here in those tests.

In the docs.

itamarst: Already gone via different branch, I think.

itamarst:
Urgh. Possibly this wasn't accepted? If you have any scruples I'll revert.

PR #1796, opened 2015-07-27
[FLOC-2712] Get rid of usage of clusterhq/flaskenv in acceptance tests.

itamarst:


flocker/acceptance/test_api.py
```diff
@@ -89,46 +90,25 @@ def test_create_container_with_environment(self, cluster):
         Create a container including environment variables on a single-node
```
adamtheturtle: This docstring has made the review of this issue harder - what I should be looking at mostly is whether the new test fits the description. Instead I am grokking the old test and looking at the new test to make sure that it is missing nothing I guess is important. I'm working on the assumption that it should be more like:

"If environment variables are specified when creating a container, those variables are available in the container's environment".

adamtheturtle:
I think that https://github.com/ClusterHQ/docker-images/tree/master/flaskenv is now not used. Add a note to the README that it is a legacy image.

flocker/acceptance/envhttp.py
```diff
@@ -0,0 +1,20 @@
+"""
+HTTP server that returns its environment variables as JSON.
+"""
+
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+from json import dumps
+from os import environ
+
+
+class Handler(BaseHTTPRequestHandler):
```
adamtheturtle: Docstrings

adamtheturtle:
Thanks @itamarst , please address the very minor comments and merge when Buildbot is green.

itamarst:
In theory this shoudl fix things, in practice have some local test failures. We'll see if buildbot is happy.

itamarst:
@exarkun -^

PR #1797, opened 2015-07-27
[FLOC-2710] Flocker api tests remove mongo

dwgebler:
Fixes FLOC-2710


flocker/acceptance/datahttp.py
```diff
@@ -0,0 +1,49 @@
...
+    from urlparse import parse_qs
+except ImportError:
+    from cgi import parse_qs
+
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+
+
+class Handler(BaseHTTPRequestHandler):
+    def do_POST(s):
+        length = int(s.headers['content-length'])
```
itamarst: :cry: HTTP, so terrible.

itamarst: Writing to arbitrary files is a bit of a security risk, especially since we're running these tests on publicly available nodes.

I suggest just writing to fixed location, rather than file specified by POST.

itamarst: If you want to make this flexible, so that you can write to different locations, I suggest making the path a command-line argument.

flocker/acceptance/datahttp.py
```diff
@@ -0,0 +1,49 @@
...
+            s.wfile.write(str(e.__class__) + ": " + str(e))
+        s.send_response(200)
+        s.end_headers()
+        s.wfile.write(b"ok")
+        s.wfile.close()
+
+    def do_GET(s):
+        s.send_response(200)
+        s.end_headers()
+        queryvars = parse_qs(urlparse(s.path).query)
```
itamarst: Could probably just always return the hardcoded file's contents. So test would fail if GET was called before POST since file would be missing.

itamarst:
Address the above, merge when buildbot is green.

itamarst:
Thanks!

PR #1800, opened 2015-07-28
[FLOC 2724] Move the ZFS backend config out of "What to do next" section

davidoliverSP2:
Fixes FLOC 2724 (and 2725, which is a duplicate)

Initially, it was planned to rename the What to do next section of http://doc-dev.clusterhq.com/install/post-install.html#what-to-do-next

This didn't make sense though, due to the reference to the experimental ZFS backend configuration content above: http://doc-dev.clusterhq.com/install/post-install.html#zfs-peer-to-peer-backend-configuration-experimental

I've moved the ZFS install instructions, and ZFS pool instructions into the backend config instructions, and removed the "What to do next" section entirely.


docs/install/post-install.rst
```diff
@@ -289,7 +289,60 @@ ZFS Peer-to-Peer Backend Configuration (Experimental)
...
+
+.. task:: create_flocker_pool_file
+   :prompt: [root@node]#
+
+.. note:: It is also possible to create the pool on a block device.
+
+.. XXX: Document how to create a pool on a block device: https://clusterhq.atlassian.net/browse/FLOC-994
+
+To support moving data with the ZFS backend, every node must be able to establish an SSH connection to all other nodes.
+So ensure that the firewall allows access to TCP port 22 on each node from the every node's IP addresses.
```
dwgebler: I think the line about "You must also set up SSH keys..." should also be here, or this should be moved to where that line is - either way, they should be together.

docs/install/post-install.rst
```diff
@@ -289,7 +289,60 @@ ZFS Peer-to-Peer Backend Configuration (Experimental)
...
+
+ZFS Backend Configuration
+.........................
+
+The ZFS backend has no infrastructure requirements: it can run no matter where the Flocker dataset agents run.
```
dwgebler: This line's meaning is unclear / confusing. Perhaps clarify that ZFS is local storage and that setting the values of "backend" and "pool" are the only requirements in the agent config file. Or just remove this sentence entirely, since that stuff has already been covered elsewhere in this document.

docs/install/post-install.rst
```diff
@@ -302,9 +355,8 @@ The configuration item to use ZFS should look like:
    Some of the suggested steps are not straightforward.
    FLOC-2092
 
-The pool name must match a ZFS storage pool that you have created on all of the Flocker agent nodes.
-This requires first installing :ref:`ZFS on Linux <installing-ZFS-CentOS-7>`.
-You must also set up SSH keys at ``/etc/flocker/id_rsa_flocker`` which will allow each Flocker dataset agent node to authenticate to all other Flocker dataset agent nodes as root.
+The pool name must match a ZFS storage pool that you have created on all of the Flocker agent nodes (see the installing :ref:`ZFS on Linux <installing-ZFS-CentOS-7>` steps above).
```
dwgebler: As a reader, my expectation was that this would link to the ZFS on Linux website http://zfsonlinux.org/
I'd suggest changing it to that - I'm not sure what benefit there is to linking back to a couple of paragraphs above on the same page, especially when it's within the same section (http://build.clusterhq.com/results/docs/install-renamewhatnext-FLOC-2724/build-9030/install/post-install.html#zfs-peer-to-peer-backend-configuration-experimental)

Another alternative is to link the ZoL website from the top of that section where we say "The ZFS backend uses node-local storage and ZFS filesystems as the storage for datasets."

dwgebler:
I left a couple of inline comments, please address / reply and then merge.

PR #1801, opened 2015-07-28
[FLOC 2716] Move authentication doc into Post-Install Config doc

davidoliverSP2:
Fixes 2716 (and duplicate 2343)


itamarst:
I'm not sure this is the best organizational scheme for this bit... in particular it's a bit tricky to find, since it's deep in. How about splitting what you have now into multiple pages or something? So install section ends up with following pages:

* Installing the client
* Installing the Node services
* Configuring authentication
* Configuring nodes and storage backends
* Enabling and running the services

davidoliverSP2:
Thanks @itamarst 

Looking at this again, I agree - I think your comment reflects the original plan for FLOC 2644, but I was trying to split the work into smaller step-changes in order to make the code review process more manageable.

davidoliverSP2:
@itamarst 

What I would like to do is merge this PR, which moves the Authentication doc out of the Using section. Then I have created FLOC 2745 (as another subtask of 2644) to move the content out of the Post-Installation Configuration topic into smaller, more specific docs, as you described above.

itamarst:
Sounds good, merge when buildbot is happy.

PR #1802, opened 2015-07-28
Sketch of minimal leases implementation

itamarst:


flocker/control/_model.py
```diff
@@ -504,6 +504,60 @@ def get_node(deployment, uuid, **defaults):
...
+    """
+    __key_type__ = UUID
+    __value_type__ = Lease
+    __invariant__ = _keys_match_dataset_id
+
+    def _check_lease(self, dataset_id, node_uuid):
+        if dataset_id in self and self[dataset_id].node_uuid != node_uuid:
+            raise ValueError("Lease already held by another node")
+
+    def acquire(self, now, dataset_id, node_uuid, expires=None):
```
jml: It occurs to me that we don't have any way to break the lock, which sounds like a thing an admin might have to do from time to time. Thoughts?

jml: Oh, I see, they can manually call the `release` API

flocker/control/_model.py
```diff
@@ -504,6 +504,60 @@ def get_node(deployment, uuid, **defaults):
...
+
+    Map dataset IDs to Leases.
+    """
+    __key_type__ = UUID
+    __value_type__ = Lease
+    __invariant__ = _keys_match_dataset_id
+
+    def _check_lease(self, dataset_id, node_uuid):
+        if dataset_id in self and self[dataset_id].node_uuid != node_uuid:
+            raise ValueError("Lease already held by another node")
```
jml: It's important that this error say which node, so the lock can be broken if necessary.

flocker/control/_model.py
```diff
@@ -504,6 +504,60 @@ def get_node(deployment, uuid, **defaults):
...
+
+class Leases(PMap):
+    """
+    Leases.
+
+    Map dataset IDs to Leases.
+    """
+    __key_type__ = UUID
+    __value_type__ = Lease
+    __invariant__ = _keys_match_dataset_id
```
jml: I didn't know you could do this w/ pyrsistent. *Awesome*

jml:
Going through the checklist we figured out in Cambridge:

### makes sure that edge cases are treated

* lock breaking discussed inline
* races around renewing / expiring / releasing
  * not addressed specifically in code, but will need explicit tests, and the design can handle it afaict

### has clean, maintainable APIs, with clear separation of concerns

Yes. Particularly like `Leases` not knowing about time.

### preserves backwards compatibility

New API. Irrelevant.

### doesn’t involve needless work

More of an issue for the epic. 

What could one do after this has landed in order to actually benefit from this change?

### avoids duplicating code

Not qualified to answer. Doesn't look like it.

### is quick to implement

Sure. Why not.

### solves a well-defined problem

I'm biased, but [Coordinating containers and data-sets](https://docs.google.com/document/d/1u6b6HAKZy4tLb_wwI6AulxD2uHsw11ftuHp2YHrsVoc/edit) might just be the best defined problem we've had.

flocker/control/_model.py
itamarst: I'm thinking maybe having a `list leases` API, which is more admin-oriented anyway. But yes need to make sure that's addressed. I'll add `list` API to epic if it's not there already.

jml:
You didn't answer the question:

What could one do after this has landed in order to actually benefit from this change?

itamarst:
I'm not sure I understand the question, or what you're asking me to do.

jml:
This adds an API, but not anything that uses it. Where do I look to see (the plans for) the bit that actually uses it?

itamarst:
1. 3rd party container frameworks might use this.
2. The Docker plugin will use this.

PR #1805, opened 2015-07-29
[FLOC-2035] [FLOC-2627]  Upgrade to docker-py-1.3.1 and requests-2.7.1

wallrj:
Fixes:
 * https://clusterhq.atlassian.net/browse/FLOC-2035
 * https://clusterhq.atlassian.net/browse/FLOC-2627

Docker-py-1.3.1 fixes https://github.com/docker/docker-py/pull/681 which is what prevented `flocker-container-agent` pulling images from private registries.

Docker-py depends on Requests >= 2.5.1
 * https://github.com/docker/docker-py/commit/9bed480c2745b06219c629dbbfd46b298c005447

So I updated that too.

I haven't updated `python-keystoneclient` - there doesn't seem any reason to. 

We can update all our dependencies properly when we start using `requires.io` or similar in 
 * https://clusterhq.atlassian.net/browse/FLOC-1172

FLOC-2035 says "restore test that depends on modern requests" but the commit ID in the issue description refers to something different. So I'll leave that unless @adamtheturtle can remember which test it was.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -312,7 +312,6 @@ class FakeDockerClient(object):
 
 # Basic namespace for Flocker containers:
 BASE_NAMESPACE = u"flocker--"
-BASE_DOCKER_API_URL = u'unix://var/run/docker.sock'
 
 
 class TimeoutClient(Client):
@@ -358,7 +357,7 @@ class DockerClient(object):
         long-running operations, particularly pulling an image.
     """
     def __init__(
-            self, namespace=BASE_NAMESPACE, base_url=BASE_DOCKER_API_URL,
+            self, namespace=BASE_NAMESPACE, base_url=None,
             long_timeout=600):
         self.namespace = namespace
         self._client = TimeoutClient(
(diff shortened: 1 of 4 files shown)


wallrj:
@tomprince already attempted the docker-py and requests upgrades in #1470 in order to support the latest version of python-keystoneclient.
In that PR he also updated the library versions used in our package building Docker images.
I don't see any reason to do that here. Those are dependencies required by the Flocker `admin/build-packages-entrypoint` which seems to still work fine with the existing package versions.

Perhaps the Flocker admin tools should be moved to a separate project with it's own setup.py , requirements.txt etc which can be monitored by requires.io.

itamarst:
The test was a test validation `requests` could use our certificates. It's not super important, but should be restored if it's not too much work. But maybe as an integration acceptance test, so let's just not bother for now.

You need to update admin deps too possibly insofar as I suspect some places install both, so you'll end up with conflict in versions of `requests`? But not seeing any relevant failure so maybe not.

Merge when buildbot is happy (or when you address relevant buildbot issues).

wallrj:
> Merge when buildbot is happy (or when you address relevant buildbot issues).

The docs build faliure was caused by https://github.com/pypa/pip/issues/1466 I think.
I created: https://clusterhq.atlassian.net/browse/FLOC-2749

> You need to update admin deps too possibly insofar as I suspect some places install both, so you'll end up with conflict in versions of requests? But not seeing any relevant failure so maybe not.

I don't think it works like that. Omnibus packages are built inside Docker containers which run `admin/build-package-entrypoint` and the requirements for that script are defined in https://github.com/ClusterHQ/flocker/blob/master/admin/build_targets/requirements.txt
That script in turn creates an omnibus virtualenv with --no-site-packages and installs Flocker there so there's no mixing of the dependencies.

PR #1808, opened 2015-07-29
[FLOC-2752] Sketch of fake client and minimal interface

itamarst:
The goal is just enough of a client interface to allow proceeding with implementing the Docker plugin business logic. Thus some of the client's API is not implemented, e.g. delete isn't really necessary at this point.

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -1,0 +1,90 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Client for the Flocker REST API.
+"""
+
+from uuid import UUID
+
+from zope.interface import Interface
+
+from pyrsistent import PClass, field, pmap_field
+
+
+class Dataset(PClass):
+    """
+    A dataset in the configuration.
+    """
+    dataset_id = field(type=UUID)
+    primary = field(type=UUID)
+    maximum_size = field(type=int)
+    deleted = field(type=bool)
+    metadata = pmap_field(unicode, unicode)
+
+
+class DatasetState(PClass):
+    """
+    The state of a dataset in the cluster.
+    """
+    dataset_id = field(type=UUID)
+    primary = field(type=UUID)
+    maximum_size = field(type=int)
+
+
+class DatasetAlreadyExists(Exception):
+    """
+    The suggested dataset ID already exists.
+    """
+
+
+class IFlockerAPIV1(Interface):
+    """
+    The Flocker REST API, v1.
+    """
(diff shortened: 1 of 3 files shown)


flocker/apiclient/_client.py
```diff
@@ -0,0 +1,90 @@
...
+
+        :return: ``Deferred`` firing with resulting ``Dataset``, or
+            errbacking with ``DatasetAlreadyExists``.
+        """
+
+    def move_dataset(primary, dataset_id):
+        """
+        Move the dataset to a new location.
+
+        :return: ``Deferred`` firing with resulting ``Dataset``.
```
robhaswell: And if the dataset does not exist?

itamarst: Presumably you'd get some sort of error. But specifying a particular error isn't necessary to proceed with basic docker plugin logic, so not going to bother with that yet.

flocker/apiclient/_client.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+
+from uuid import UUID
+
+from zope.interface import Interface
+
+from pyrsistent import PClass, field, pmap_field
+
+
+class Dataset(PClass):
```
wallrj: :question: ``PClass`` sounds like it's probably more appropriate than ``PRecord`` for ``Dataset`` and ``DatasetState``. Ought we to be using ``PClass`` instead of ``PRecord`` for ``flocker.control._model.Dataset`` etc?

wallrj: :question: And can't we re-use the existing types from `flocker.control._model` here? Or are you anticipating this package being installed without a `Flocker` dependency?

wallrj: Ah no. Because ``flocker.control.Dataset`` has no notion of `primary`. It just belongs to a `Node`.

wallrj:
Thanks @itamarst 

Looks good to me.

And this API can eventually replace (or be used by) the ad-hoc API in  ``flocker.acceptance.testtools``.
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/testtools.py#L338

That API has a ``wait_for_dataset`` method. Would that be useful here too? Or will you expect the library user to implement that?

Will you add Eliot logging to the library?

Please proceed.

flocker/apiclient/_client.py
itamarst: 1. We should probably switch to `PClass`, yeah. It was added after we wrote that code.
2. I am writing this was expectation that it may someday be standalone, yes.

itamarst:
I expect the Docker plugin will indeed have some equivalent of wait_for_dataset, but it may well look somewhat different.

PR #1814, opened 2015-07-30
[FLOC-2752] Flocker client interface and in-memory fake

itamarst:
The interface will be implemented by real client in follow-up issue; the fake will be used by the Docker plugin to test the plugin logic.


flocker/apiclient/_client.py
```diff
@@ -0,0 +1,139 @@
...
+from twisted.internet.defer import succeed, fail
+
+
+class Dataset(PClass):
+    """
+    A dataset in the configuration.
+
+    :param UUID primary: The node where the dataset should manifest.
+    :param int maximum_size: Size of new dataset, in bytes.
+    :param dataset_id: The UUID of the dataset.
```
jml: Should this also be `:param UUID dataset_id`?

flocker/apiclient/_client.py
```diff
@@ -0,0 +1,139 @@
...
+    metadata = pmap_field(unicode, unicode)
+
+
+class DatasetState(PClass):
+    """
+    The state of a dataset in the cluster.
+
+    :param UUID primary: The node where the dataset should manifest.
+    :param int maximum_size: Size of new dataset, in bytes.
+    :param dataset_id: The UUID of the dataset.
```
jml: And here?

flocker/apiclient/_client.py
```diff
@@ -0,0 +1,139 @@
...
+
+class DatasetAlreadyExists(Exception):
+    """
+    The suggested dataset ID already exists.
+    """
+
+
+class IFlockerAPIV1(Interface):
+    """
+    The Flocker REST API, v1.
```
jml: I feel ill at ease calling this the REST API when it is – AIUI – the interface to a couple of clients of the REST API. 

Perhaps "The Flocker REST API client, v1"?

flocker/apiclient/_client.py
```diff
@@ -0,0 +1,139 @@
...
+        """
+
+    def move_dataset(primary, dataset_id):
+        """
+        Move the dataset to a new location.
+
+        :param UUID primary: The node where the dataset should manifest.
+        :param dataset_id: Which dataset to move.
+
+        :return: ``Deferred`` firing with resulting ``Dataset``.
```
jml: I assume the `Deferred` fires after the move is complete? Please document when it fires.

flocker/apiclient/test/test_client.py
```diff
@@ -0,0 +1,181 @@
...
+            If no ``dataset_id`` is specified when calling ``create_dataset``,
+            a new one is generated.
+            """
+            return self.assert_creates(client_factory(), primary=self.node_1,
+                                       maximum_size=DATASET_SIZE)
+
+        def test_create_given_dataset(self):
+            """
+            If no ``dataset_id`` is specified when calling ``create_dataset``,
+            a new one is generated.
```
jml: This docstring duplicates the one above. I think you meant "If a ``dataset_id`` is specified when calling ``create_dataset``, that becomes the id of the created dataset.", or words to that effect.

flocker/apiclient/test/test_client.py
```diff
@@ -0,0 +1,181 @@
...
+                                       dataset_id=uuid4())
+
+        def test_create_with_metadata(self):
+            """
+            The metadata passed to ``create_dataset`` is stored with the
+            dataset.
+            """
+            return self.assert_creates(client_factory(), primary=self.node_1,
+                                       maximum_size=DATASET_SIZE,
+                                       metadata={u"hello": u"there"})
```
jml: :question: It's kind of difficult to look at this test and see that what it's testing aligns with the docstring. As a reader, you have to put a lot of trust in `assert_creates`. I think I would rather a bit more duplication.

flocker/apiclient/test/test_client.py
```diff
@@ -0,0 +1,181 @@
...
+                listed.addCallback(lambda l: (dataset, l))
+                return listed
+            d.addCallback(got_result)
+
+            d.addCallback(lambda result:
+                          self.assertEqual((result[0], result[0] in result[1]),
+                                           (Dataset(dataset_id=dataset_id,
+                                                    primary=self.node_2,
+                                                    maximum_size=DATASET_SIZE),
+                                            True)))
```
jml: I found this assertion difficult to follow. Partly it's because it's combining two assertions into one in order to maximize information in the assertion.

Suggestions to make it easier:
 * unpack `result` in the lambda to `(dataset, found_datasets)` or similar
 * use `@inlineCallbacks`

flocker/apiclient/test/test_client.py
```diff
@@ -0,0 +1,181 @@
...
+            d.addCallback(lambda _: synchronize_state(client))
+            d.addCallback(lambda _: client.list_datasets_state())
+            d.addCallback(lambda states:
+                          self.assertIn(
+                              DatasetState(dataset_id=dataset_id,
+                                           primary=self.node_1,
+                                           maximum_size=DATASET_SIZE * 2),
+                              states))
+            return d
+
```
jml: :question: Other tests to consider adding:

* what if `maximum_size` is zero, or negative
* initial state and configuration of supplied client

itamarst: Those are really tests of the server, it seems? I guess the first one could be something the client checks as well. But probably not essential for proceeding with leases so going to punt on it for now.

flocker/apiclient/test/test_client.py
itamarst: Jean-Paul hates inlineCallbacks, alas.

itamarst:
I've addressed all comments except the last one, which is answered.

jml:
Great, thanks!

Feel free to merge when the tests pass.

itamarst:
One hour later, three builds still pending.

PR #1816, opened 2015-07-31
[FLOC 2265] Specify that only one backend can be chosen

davidoliverSP2:
Fixes FLOC 2265


jml:
LGTM.

I think you should feel free to merge, since it passes the doc builder.

PR #1820, opened 2015-07-31
[FLOC-2769] Test instructions for client installed using pip

jongiddy:
This PR uses testable tasks for the pip installation of Flocker CLI provided for "other operating systems".

Although these instructions are a starting point for users on other OS's, we only test the instructions on CerntOS 7 (unsupported for clients) and Ubuntu 15.04 (a supported client system).  This is to check one YUM-based system, and one APT-based system, using the existing test resources.

The `admin/run-client-tests` command is modified to provide a `--pip` flag to install using pip instead of using packages.

For example, the command 
```./admin/run-client-tests --distribution=ubuntu-15.04 --flocker-version=1.0.3 --config-file=config.yml```
installs the client using 1.0.3 packages.

The command 
```./admin/run-client-tests --pip --distribution=ubuntu-15.04 --flocker-version=1.0.3 --config-file=config.yml```
installs the client using 1.0.3 wheel.

Note, for pip, using a non-release version number will use the previous production or dev release. This is necessary, since the wheel file is not published for each branch.

Note, there is another sub-issue to add this test to Buildbot (probably by combining it with the existing client tests).

### CODE REVIEW

File: docs/install/install-client.rst [hunk 1]
```diff
@@ -52,55 +52,31 @@ Other Linux Distributions
...
 
-.. version-code-block:: console
+.. task:: cli_pip_prereqs ubuntu-15.04
+   :prompt: alice@mercury:~$
 
-   alice@mercury:~$ cd flocker-tutorial
-   alice@mercury:~/flocker-tutorial$ bin/flocker-deploy --version
-   |latest-installable|
-   alice@mercury:~/flocker-tutorial$
+Then run the following commands to install ``flocker-cli`` in a Python virtualenv:
```
davidoliverSP2: This sentence should keep the same structure as the other instructions - 
Change
+Then run the following commands to install ``flocker-cli`` in a Python virtualenv:
to
+To install ``flocker-cli`` in a Python virtualenv, run:

docs/install/install-client.rst
```diff
@@ -52,55 +52,31 @@ Other Linux Distributions
...
 
-If you want to omit the prefix path you can add the appropriate directory to your ``$PATH``.
-You'll need to do this every time you start a new shell.
+.. task:: cli_pip_install flocker-client
+   :prompt: alice@mercury:~$
 
-.. version-code-block:: console
+Ensure you are in the virtualenv whenever you need to run Flocker CLI commands:
```
davidoliverSP2: For the same reason as above:
Change
+Ensure you are in the virtualenv whenever you need to run Flocker CLI commands:
to
+Whenever you need to run Flocker CLI commands, ensure you are in the virtualenv:

davidoliverSP2:
Thanks @jongiddy 
Couple of sentences to restructure, but no other changes required to the doc file.

admin/client.py
```diff
@@ -188,14 +191,13 @@ class RunOptions(Options):
         ['branch', None, None, 'Branch to grab packages from'],
         ['flocker-version', None, flocker.__version__,
          'Version of flocker to install'],
-        ['flocker-version', None, flocker.__version__,
-         'Version of flocker to install'],
```
jml: Was this just an incidental duplication that you removed?

admin/client.py
```diff
@@ -297,10 +297,19 @@ def main(reactor, args, base_path, top_level):
...
+                install_cli_pip(runner.package_source, nodes[0]))
+            result = yield run_client_tests(
+                reactor=reactor, node=nodes[0], commands=task_cli_pip_test())
+        else:
+            yield perform(
+                make_dispatcher(reactor),
+                install_cli(runner.package_source, nodes[0]))
+            result = yield run_client_tests(
+                reactor=reactor, node=nodes[0],
+                commands=task_client_installation_test())
```
jml: :question: 

Totally optional. I'd consider extracting the common bits between these two branches, like so:

```python
  if options['pip']:
    installer = install_cli_pip
    commands = task_cli_pip_test
  else:
    installer = install_cli
    commands=task_client_installation_test
  yield perform(make_dispatcher(reactor), installer(runner.package_source, nodes[0])
  result = yield run_client_tests(reactor=reactor, node=nodes[0], commands=commands)
```

admin/client.py
```diff
@@ -297,10 +297,19 @@ def main(reactor, args, base_path, top_level):
...
+        if options['pip']:
+            yield perform(
+                make_dispatcher(reactor),
+                install_cli_pip(runner.package_source, nodes[0]))
+            result = yield run_client_tests(
+                reactor=reactor, node=nodes[0], commands=task_cli_pip_test())
+        else:
+            yield perform(
+                make_dispatcher(reactor),
+                install_cli(runner.package_source, nodes[0]))
```
jml: Now that there's more than one way of installing the CLI, we should rename `install_cli` to `install_cli_system_packages`.

admin/client.py
jml: Likewise, `task_client_installation_test` is now an ambiguous name. Perhaps `task_cli_system_packages_test`?

flocker/provision/_ssh/_conch.py
```diff
@@ -131,8 +131,12 @@ def perform_run(dispatcher, intent):
             message_type="flocker.provision.ssh:run",
             command=intent.log_command_filter(intent.command),
         ).write()
+        if hasattr(intent, 'pre_command') and intent.pre_command:
```
jml: `hasattr` can sometimes explode in weird ways. I'd suggest:

```python
pre_command = getattr(intent, 'pre_command', None)
if pre_command:
   # ...
```

flocker/provision/_ssh/_model.py
```diff
@@ -52,15 +52,20 @@ class Run(PRecord):
     :ivar bytes command: The command to run.
     :ivar callable log_command_filter: A filter to apply to any logging
         of the executed command.
+    :ivar bytes pre_command: A command to run before the ``command``,
+        generally to setup the environment for the command.
```
jml: This seems unusual. Why not just run two commands?

jml: Oh right, because `virtualenv` weirdly doesn't provide a way of running arbitrary commands inside a virtualenv. Even so, its cleaner to change `command` to `commands`.

flocker/provision/_install.py
```diff
@@ -325,6 +325,103 @@ def install_cli(package_source, node):
...
+    """
+    if distribution in ('centos-7',):
+        return sudo_from_args([
+            'yum', '-y', 'install',
+            'gcc',
+            'libffi-devel',
+            'python',
+            'python-devel',
+            'python-virtualenv',
+            'openssl-devel',
```
jml: :question: Extract this list of packages to a top-level constant.

flocker/provision/_install.py
```diff
@@ -325,6 +325,103 @@ def install_cli(package_source, node):
...
+        return sequence([
+            sudo_from_args(['apt-get', 'update']),
+            sudo_from_args([
+                'apt-get', '-y', 'install',
+                'gcc',
+                'libffi-dev',
+                'libssl-dev',
+                'python2.7',
+                'python2.7-dev',
+                'python-virtualenv',
```
jml: :question: Extract this list of packages to a top-level constant.

flocker/provision/_ssh/_model.py
jml: An alternative would be to leave this as `command` and have a separate helper function somewhere called `venv_command` (or something), that takes an arbitrary command and sticks 'source bin/activate; ' in front of it.

jml:
Thanks for getting to this. It looks like a really useful thing to have.

I think a couple of things need to be renamed now that we've got more than one way of installing things. Also, I suspect the new `pre_command` will come back to byte us later. I've suggested a couple of alternative approaches.

Please go through them and resubmit for review.

flocker/provision/_ssh/_model.py
jongiddy: Each command runs in a separate channel, so the results of a `source` command are lost to subsequent commands.  This solution re-creates the environment for each subsequent command, but we don't want `source...` to appear multiple times in the docs.

The alternative solution is probably to change the RunRemotely implementation for conch to use a single CommandProtocol for multiple commands, so that the one ssh channel is used and the environment is kept between related commands.

I will consider the approach of setting a virtualenv instead of a generic pre_command. That sounds OK. The only problem is that the virtualenv locaiton is actually a directory rather than a global label. If we were using `mkvirtualenv` it would be more of a global label.

admin/client.py
jongiddy: Yes. Unrelated to issue, but uncontroversial (I hoped).

jml: Definitely OK, just wanted to make sure I understood.

flocker/provision/_ssh/_conch.py
jongiddy: @jml The only situation where `hasattr` can act weirdly is if the object has a weird (probably stateful) `__getattr__`/`__getattribute__` handler.  Nevertheless, I was dithering on these two versions myself, so happy to change it.

jongiddy:
I will get back to this PR once the fate of #1827 is known. That PR will allow this PR to use a cleaner solution than adding the `pre_command` parameter to the Run intent.

jongiddy:
Tests now run in a Docker container, and do not need the `pre_command` attribute.

Moved the really basic setup required for Docker images (e.g. install sudo) into a separate piece of code which does not appear in the docs. Instead, the docs now mention that sudo is a requirement.  This keeps the instructions a bit less verbose, and we expect that most people will already have sudo, so it's a no-op anyway.

Incidentally, the client docs still had an intro referring to the overall structure of CLI and nodes, so I moved that to the top-level installation page.

To handle some escaped commands within commands (su -c 'command') I rewrote the shell escaping to allow nested lists for internal commands.

Pip installation can now be tested on Debian, Fedora, CentOS, and Ubuntu.  Plan is to create a single test in Buildbot that can run through all these systems in one test.

admin/client.py
```diff
@@ -9,6 +9,7 @@
 import tempfile
 import yaml
 
+from characteristic import attributes
```
jml: No action required.

FWIW, the author of characteristic now recommends people use [attrs](http://attrs.rtfd.org/)

flocker/provision/_install.py
```diff
@@ -313,16 +348,85 @@ def task_install_cli(distribution, package_source=PackageSource()):
...
+        return sudo_from_args(
+            [package_manager, '-y', 'install'] + PIP_CLI_PREREQ_YUM
+        )
+    elif package_manager == 'apt':
+        return sequence([
+            sudo_from_args(['apt-get', 'update']),
+            sudo_from_args(['apt-get', '-y', 'install'] + PIP_CLI_PREREQ_APT),
+        ])
+    else:
+        raise UnsupportedDistribution()
```
jml: :question: There's now a couple of different places where the patch executes different code depending on the package manager. It might be worth extracting that out into a couple of classes (one for yum/dnf, one for apt), and making `ensure_minimal_setup` and `task_cli_pip_prereqs` methods on those. Removes the `if`s and associated duplication. 

I really don't think it's a necessary change though.

jml:
Thanks, LGTM. I've made an optional suggestion for a relatively large change. Bounce back to me if you decide to go for it. Likewise if you have to do anything major to fix tests. 

Please feel free to land when the tests are all green.

jongiddy:
Thanks for the review.

I have added the comment to FLOC-1670, since this has been discussed before for distribution-specific code, and I'd rather do it for the whole file as a consistent change. However, there doesn't seem to be much love for refactoring `flocker.provision` right now.

PR #1803, opened 2015-07-29
[FLOC-2746] Add network logging to release processes.

jongiddy:
This PR enables the requests library logging to display the URLs and status of HTTP connections.  This makes it easier to see which URLs have failed if an error occurs.

This is particularly useful now, as we start running dev releases with the PEP440 changes.

To test, run a command like:
```
./admin/publish-artifacts --target=clusterhq-archive-testing --flocker-version=0.9.0pre1
```

In other branches, the only information is a stack trace and:
```
flocker.provision._effect.SequenceFailed: <SequenceFailed(results=[], exc_info=(<class 'requests.exceptions.HTTPError'>, HTTPError('404 Client Error: Not Found',), <traceback object at 0x3a65488>))>
```

With this branch, before the stack trace, we also see:
```
INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): build.clusterhq.com
DEBUG:requests.packages.urllib3.connectionpool:"GET /results/omnibus/0.9.0pre1/centos-7/clusterhq-python-flocker-0.9.0-0.rc.1.x86_64.rpm HTTP/1.1" 404 145
```
which makes it quicker to see where an expected file is missing.

PR #1822, opened 2015-07-31
[FLOC-2775] Move logging setup from release code to publish-artifacts

jongiddy:
The change in #1803 made `publish-artifacts` display useful logging information. However, it also added logging to commands such as `test-redirects` which uses the same module for its core functions.

While the debugging info helps in `publish-artifacts`, it obscures in `test-redirects`. Hence, this PR moves the logging setup code into the top-level `publish-artifacts` script, so it only affects that code.

The command `./admin/publish-artifacts --flocker-version=0.9.0pre1 --target=clusterhq-archive-testing` still provides network logging.

The command `./admin/test-redirects` does not. (It will probably complain about the redirects, but will not display INFO and DEBUG messages).

### DIFF
--- a/admin/publish-artifacts
+++ b/admin/publish-artifacts
@@ -6,7 +6,14 @@ Publish release artifacts.
 
 from _preamble import TOPLEVEL, BASEPATH
 
+import logging
 import sys
+
+# Log information about network connections
+logging.basicConfig()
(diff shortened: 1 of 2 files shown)


jml:
Thanks for the clear description of the motivation behind this change! Much appreciated.

LGTM. Please merge when tests pass.

PR #1824, opened 2015-08-03
[FLOC-2734] Add leases to control service

dwgebler:
Fixes FLOC-2734. Adds the leases models used by the control service. Adds a simple looping lease service to the control service to check for and remove expired leases every second. Due to the limited scope of this branch (in particular that the Deployment model cannot be updated here), this looping service doesn't do anything here, but provides the skeletal service and test infrastructure to be completed in another branch (see the parent story at FLOC-2731).


flocker/control/_model.py
```diff
@@ -504,6 +504,126 @@ def get_node(deployment, uuid, **defaults):
...
+        :return: The updated ``Leases`` representation.
+        """
+        try:
+            self._check_lease(dataset_id, node_id)
+        except ValueError as e:
+            message = (
+                u"Cannot acquire lease " + unicode(dataset_id) + " for node "
+                + unicode(node_id) + ": " + e.message
+            )
+            raise LeaseAcquisitionError(message)
```
jml: I think this can be simplified a bit:

1. There doesn't seem to be a need for two special exception types, one `LeaseError` will do.
2. There's no need for it to subclass `ValueError`.
3. :question: The exception class itself can construct the message.

This means that `_check_lease` could instead have `raise LeaseError(dataset_id, node_id)`, and the `try`/`except` blocks can be removed from `acquire` and `release`.

flocker/control/_persistence.py
```diff
@@ -87,7 +93,58 @@ def decode_object(dictionary):
...
+    :param persistence_service: The persistence service to which the
+        updated configuration will be saved.
+
+    :return Deferred: Fires when the persistence service has saved.
+    """
+    # XXX we cannot manipulate leases in this branch since the configuration
+    # doesn't know anything about them yet. See FLOC-2731 subtasks.
+    # configuration = persistence_service.get()
+    # return persistence_service.save(
+    #     configuration.set("leases", transform(configuration.leases)))
```
jml: Please delete this commented out code for now. If you need it for reference, then put it in the appropriate FLOC-2731 subtask.

flocker/control/test/test_model.py
```diff
@@ -1460,3 +1462,190 @@ def test_case_insensitive(self):
...
+        """
+        # Acquire a lease on a node
+        leases = self.leases.acquire(self.now, self.dataset_id, self.node_id)
+        # Check the lease for a different node UUID
+        exception = self.assertRaises(
+            ValueError, leases._check_lease, self.dataset_id, uuid4()
+        )
+        self.assertEqual(
+            exception.message, "Lease already held by another node"
+        )
```
jml: :question: Reasonable people disagree on this point, but I generally don't see the point of writing tests for internal methods, especially ones that are this simple. Consider deleting this test.

flocker/control/test/test_model.py
```diff
@@ -1460,3 +1462,190 @@ def test_case_insensitive(self):
...
+    def test_lease_expires(self):
+        """
+        An acquired lease expires after the specified number of seconds and
+        is removed from the ``Leases`` map.
+        """
+        leases = self.leases.acquire(
+            self.now, self.dataset_id, self.node_id, self.lease_duration
+        )
+        # Assert the lease has been acquired successfully.
+        self.assertTrue(self.dataset_id in leases)
```
jml: `assertIn(self.dataset_id, leases)` will get you a better error message when this fails.

flocker/control/test/test_model.py
```diff
@@ -1460,3 +1462,190 @@ def test_case_insensitive(self):
...
+        # Fake a time the first lease has expired.
+        now = self.now + datetime.timedelta(seconds=self.lease_duration + 1)
+        leases = leases.expire(now)
+        # Assert the lease has been removed successfully.
+        self.assertFalse(self.dataset_id in leases)
```
jml: Likewise `assertNotIn`

flocker/control/test/test_model.py
```diff
@@ -1460,3 +1462,190 @@ def test_case_insensitive(self):
...
+    def test_indefinite_lease_never_expires(self):
+        """
+        An acquired lease set to never expire is not removed from ``Leases``
+        map.
+        """
+        leases = self.leases.acquire(self.now, self.dataset_id, self.node_id)
+        self.assertTrue(self.dataset_id in leases)
+        now = self.now + datetime.timedelta(seconds=self.lease_duration)
+        leases = leases.expire(now)
+        self.assertTrue(self.dataset_id in leases)
```
jml: Please use `assertIn` here & elsewhere.

flocker/control/_persistence.py
```diff
@@ -87,7 +93,58 @@ def decode_object(dictionary):
...
+        self._lc.clock = self._reactor
+        self._lc.start(1)
+
+    def stopService(self):
+        self._lc.stop()
+
+    def _expire(self):
+        now = datetime.now(tz=UTC)
+        return update_leases(lambda leases: leases.expire(now),
+                             self._persistence_service)
```
jml: AFAICT, this never gets called in this branch, and it's certainly not tested. Which ticket is the one that gets this to be used?

jml:
Thanks! A few small comments and a couple of big questions. Happy to talk further in a forum of your choice.

flocker/control/_persistence.py
dwgebler: So this is the only comment that remains (partially) unaddressed at this stage. Basically, this branch includes a minimal Lease service as part of the control service (now a `MultiService`) and this runs a loop every second to check for expired leases. However, in this branch, we aren't updating the `Deployment` model to be lease-aware (because that would break things if it were done now), which means our service here doesn't do anything yet. `_expire` here calls `update_leases`, which will save the configuration through the persistence service after running the `Leases` model's `expire` method. But in this branch it just returns a `Deferred` with a success result of `None`. The problem with doing anything further than I've done in this branch with this code and the skeletal test infrastructure in the persistence unit tests is that it's not really possible until `Deployment` has a concept of leases. Suggestions are welcome.

jml: Thanks for the added test. Could you please update the `XXX` comment in the issue below with the exact ticket for when this will be actually used (rather than the higher-level ticket)? If you can do that, then please feel free to merge. If it's not possible for some reason, then please bounce back to me.

PR #1825, opened 2015-08-03
[FLOC-2783] Simple API calls for Docker plugin API

itamarst:
1. Move a bunch of utility code out of `flocker.control.test.test_httpapi` and into `flocker.restapi.testtools`. It is not modified in any way.
2. Add basic (output) schema and API call implementations for three trivial Docker plugin API endpoints.

The relevant documentation:
https://github.com/docker/docker/blob/master/docs/extend/plugin_api.md#pluginactivate
https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md#volumedriverremove
https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md#volumedriverunmount

PR #1826, opened 2015-08-03
[FLOC-2784] Docker plugin: create volume

itamarst:
1. Builds on #1825, which is mergeable when buildbot is happy. Hopefully by the time you're reviewing this you can click Merge yourself.
2. Relevant docs: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md#volumedrivercreate

### DIFF
--- a/flocker/dockerplugin/test/test_api.py
+++ b/flocker/dockerplugin/test/test_api.py
@@ -61,7 +61,8 @@ class APITestsMixin(APIAssertionsMixin):
         name = u"myvol"
         d = self.assertResult(b"POST", b"/VolumeDriver.Create",
                               {u"Name": name}, OK, {u"Err": None})
-        d.addCallback(lambda _: self.flocker_client.list_datasets_configuration())
+        d.addCallback(
+            lambda _: self.flocker_client.list_datasets_configuration())
         d.addCallback(self.assertItemsEqual, [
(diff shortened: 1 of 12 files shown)


flocker/dockerplugin/test/test_api.py
```diff
@@ -0,0 +1,124 @@
...
+        return self.assertResult(b"POST", b"/VolumeDriver.Unmount",
+                                 {u"Name": u"vol"}, OK, {u"Err": None})
+
+    def test_create_creates(self):
+        """
+        ``/VolumeDriver.Create`` creates a new dataset in the configuration.
+        """
+        name = u"myvol"
+        d = self.assertResult(b"POST", b"/VolumeDriver.Create",
+                              {u"Name": name}, OK, {u"Err": None})
```
jml: :question: I'd normally have a separate test that asserts that the initial `list_datasets_configuration` was empty. Not strictly necessary here, but add it if you feel that it helps.

jml:
One minor & ignorable suggestion. No need for follow-up review either way.

Thanks!

flocker/dockerplugin/test/test_api.py
itamarst: That seems like more of a test for the Flocker client, so I'll probably skip it.

PR #1827, opened 2015-08-04
[FLOC-2771] Run client installation tests using Docker

jongiddy:
The client tests currently start cloud nodes for testing. Since the tests are very short, this is wasteful in time and money, especially since the tests do not need the same capabilities as the acceptance tests.

This PR changes the tests to use a local Docker container.

Apart from the resource usage argument above, this PR is motivated by
- FLOC-2750, regarding client test failures due to cloud request limits, and 
- the code review of FLOC-2769.  The problem with the PR #1820 is that the cloud tests do not maintain the environment between commands. Each command runs in a separate SSH channel. This prevents `virtualenv` being used sensibly. The Docker Runner in this PR supports keeping the environment between commands in a sequence, so will allow a cleaner solution to that PR.

Although the commits are a set of gradual changes to the files, for review it is probably easier to consider the file `admin/client.py` as completely re-written code rather than looking at the diff.

To test, run a command like:
```
./admin/run-client-tests --distribution ubuntu-15.04 --flocker-version=1.1.0.dev2
```

### DIFF
--- a/admin/client.py
+++ b/admin/client.py
@@ -1,4 +1,4 @@
-# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
 """
 Run the client installation tests.
 """
--- a/admin/client.py
+++ b/admin/client.py
@@ -84,7 +84,7 @@ def make_script_file(dir, effects):
     Create a shell script file from a sequence of effects.
 
     :param bytes dir: The directory in which to create the script.
-    :param Effect effects: An effect which contain the commands,
+    :param Effect effects: An effect which contains the commands,
         typically a Sequence containing multiple commands.
     :return: The base filename of the script.
     """
@@ -98,7 +98,7 @@ def make_script_file(dir, effects):
 
 class DockerRunner:
     """
-    Run commands on Docker.
+    Run commands in a Docker container.
     """
 
     def __init__(self, image):
--- a/admin/client.py
+++ b/admin/client.py
@@ -1,6 +1,6 @@
 # Copyright Hybrid Logic Ltd.  See LICENSE file for details.
 """
-Run the acceptance tests.
+Run the client installation tests.
 """
 
 import os
@@ -10,21 +10,20 @@ import tempfile
 import yaml
 
 import docker
+from effect import TypeDispatcher, sync_performer, perform
 from twisted.python.usage import Options, UsageError
 from twisted.python.filepath import FilePath
+
+import flocker
 from flocker.common.version import make_rpm_version
 from flocker.provision import PackageSource
-import flocker
-from flocker.provision._install import (
-    task_client_installation_test,
-    task_install_cli,
-)
-from effect.twisted import perform
-
-from effect import TypeDispatcher, sync_performer
 from flocker.provision._effect import Sequence, perform_sequence
-from flocker.provision._ssh._model import Run, Sudo, Put, Comment
+from flocker.provision._install import (
+    task_install_cli,
+    task_client_installation_test,
+)
 from flocker.provision._ssh._conch import perform_sudo, perform_put
+from flocker.provision._ssh._model import Run, Sudo, Put, Comment
 
 DOCKER_IMAGES = {
     'centos-7': 'centos:7',
@@ -44,7 +43,10 @@ class ScriptBuilder(TypeDispatcher):
     """
 
     def __init__(self, effects):
-        self.lines = ['#!/bin/bash', 'set -ex']
+        self.lines = [
+            '#!/bin/bash',
+            'set -ex'
+        ]
         TypeDispatcher.__init__(self, {
             Run: self.perform_run,
             Sudo: perform_sudo,
--- a/admin/client.py
+++ b/admin/client.py
@@ -109,19 +109,25 @@ class DockerRunner:
 
     def execute(self, commands):
         """
-        :param commands: An Effect containing the commands to run, probably a
-            Sequence of Effects, one for each command to run.
+        Execute a set of commands in the Docker container.
+
+        The set of commands provided to one call of ``execute`` will be
+        executed in a single session. This means commands will see the
+        environment created by previous commands.
+
+        :param commands: An Effect containing the commands to run,
+            probably a Sequence of Effects, one for each command to run.
+        :return int: The exit status of the commands.  If all commands
+            succeed, this will be zero. If any command fails, this will
+            be non-zero.
         """
         script_file = make_script_file(self.tmpdir, commands)
         script = '/mnt/script/{}'.format(script_file)
         session = self.docker.exec_create(self.container_id, script)
         session_id = session[u'Id']
-        output = self.docker.exec_start(session)
-        status = self.docker.exec_inspect(session_id)[u'ExitCode']
-        if status == 0:
+        for output in self.docker.exec_start(session, stream=True):
             sys.stdout.write(output)
-        else:
-            sys.exit(output)
+        return self.docker.exec_inspect(session_id)[u'ExitCode']
 
 
 class RunOptions(Options):
@@ -210,6 +216,8 @@ def main(args, base_path, top_level):
     runner.start()
     try:
         for commands in steps:
-            runner.execute(commands)
+            status = runner.execute(commands)
+            if status != 0:
+                sys.exit(status)
     finally:
         runner.stop()
--- a/admin/client.py
+++ b/admin/client.py
@@ -174,7 +174,7 @@ class RunOptions(Options):
          'Version of flocker to install'],
         ['build-server', None, 'http://build.clusterhq.com/',
          'Base URL of build server for package downloads'],
-        # XXX - remove the remaining flags once Buildbot is updated
+        # XXX - remove the remaining flags once Buildbot is updated (FLOC-2813)
         ['provider', None, None, 'No longer used.'],
         ['config-file', None, None, 'No longer used.'],
     ]
--- a/admin/client.py
+++ b/admin/client.py
@@ -96,9 +96,6 @@ class DockerRunner:
     def install_client(self, distribution, package_source):
         tmpdir = tempfile.mkdtemp()
         try:
-            install = make_script_file(
-                tmpdir, task_install_cli(distribution, package_source))
-            dotest = make_script_file(tmpdir, task_client_installation_test())
             image = DOCKER_IMAGES[distribution]
             self.docker.pull(image)
             container = self.docker.create_container(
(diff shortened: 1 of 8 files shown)


admin/client.py
```diff
@@ -1,176 +1,164 @@
 # Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
adamtheturtle: :question: Not new. I don't know why this is in `admin`.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
-    """
-    Remove all keys belonging to hostname from a known_hosts file.
 
-    :param reactor: Reactor to use.
-    :param bytes hostname: Remove all keys belonging to this hostname from
-        known_hosts.
+class ScriptBuilder(TypeDispatcher):
     """
-    return run(reactor, ['ssh-keygen', '-R', hostname])
+    Convert an Effect sequence to a shell script.
```
adamtheturtle: I don't like that we're building upon our internal sequence when it is something we want to get rid of.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
+    """
+    Create a shell script file from a set of effects.
 
-@implementer(INodeRunner)
-@attributes(RUNNER_ATTRIBUTES + [
-    'provisioner',
-], apply_immutable=True)
-class LibcloudRunner(object):
+    :param dir: The directory in which to create the script.
+    :param effects: The effects which contain the commands.
```
adamtheturtle: A list, a `sequence`?

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
-    def __init__(self):
-        self.nodes = []
 
-        self.metadata = self.config.get('metadata', {})
-        try:
-            creator = self.metadata['creator']
-        except KeyError:
-            raise UsageError("Must specify creator metadata.")
+    def __init__(self, image):
+        self.docker = docker.Client(version='1.18')
```
adamtheturtle: Is it bad that we're using different versions in different places (e.g. `version="1.15"` in `DockerClient`)?

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
+        Execute a set of commands in the Docker container.
+
+        The set of commands provided to one call of ``execute`` will be
+        executed in a single session. This means commands will see the
+        environment created by previous commands.
+
+        The output of the commands is sent to the ``out`` file object,
+        which must have a ``write`` method.
+
+        :param commands: An Effect containing the commands to run,
```
adamtheturtle: Document the types of params.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
+from flocker.provision import PackageSource
+from flocker.provision._effect import Sequence, perform_sequence
 from flocker.provision._install import (
+    task_install_cli,
     task_client_installation_test,
-    install_cli,
 )
-from effect.twisted import perform
-from flocker.provision._ssh._conch import make_dispatcher
+from flocker.provision._ssh._conch import perform_sudo, perform_put
```
adamtheturtle: If these things need to be imported perhaps they should be moved to a non-private module.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
+DISTRIBUTIONS = DOCKER_IMAGES.keys()
 
-def remove_known_host(reactor, hostname):
-    """
-    Remove all keys belonging to hostname from a known_hosts file.
 
-    :param reactor: Reactor to use.
-    :param bytes hostname: Remove all keys belonging to this hostname from
-        known_hosts.
+class ScriptBuilder(TypeDispatcher):
```
adamtheturtle: :question: The commands to put in the script are the same as those to render in sphinx. Therefore this overlaps with `run_for_docs`. I wonder if it would be possible to combine the two somehow, or share some code.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
-from twisted.internet.defer import inlineCallbacks, returnValue
 
-from flocker.common.version import make_rpm_version
-from flocker.provision import PackageSource, CLOUD_PROVIDERS
 import flocker
-from flocker.provision._ssh import (
-    run_remotely)
+from flocker.common.version import make_rpm_version
+from flocker.provision import PackageSource
+from flocker.provision._effect import Sequence, perform_sequence
```
adamtheturtle: Maybe it'd be good to have a look at how much time https://clusterhq.atlassian.net/browse/FLOC-2810 would take, and consider whether that time is worth it to avoid building so much on `Sequence` which we want to get rid of - merging this PR will make that take longer.

admin/client.py
```diff
@@ -180,26 +168,19 @@ class RunOptions(Options):
...
          'Version of flocker to install'],
-        ['flocker-version', None, flocker.__version__,
-         'Version of flocker to install'],
         ['build-server', None, 'http://build.clusterhq.com/',
          'Base URL of build server for package downloads'],
-    ]
-
-    optFlags = [
-        ["keep", "k", "Keep VMs around, if the tests fail."],
+        # XXX - remove the remaining flags once Buildbot is updated
```
adamtheturtle: Is there an issue to link to? The one for updating Buildbot.

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -226,15 +226,15 @@ The nodes should be configured to allow key based SSH connections as user ``root
 
 .. _client-acceptance-tests:
```
adamtheturtle: :question: Perhaps this page should be merged somehow with http://doc-dev.clusterhq.com/gettinginvolved/client-testing.html?highlight=docker

flocker/provision/_install.py
```diff
@@ -241,9 +241,9 @@ def install_cli_commands_ubuntu(distribution, package_source):
         # Minimal images often have cleared apt caches and are missing
         # packages that are common in a typical release.  These commands
         # ensure that we start from a good base system with the required
-        # capabilities, particularly that the add-apt-repository command
-        # and HTTPS URLs are supported.
-        # FLOC-1880 will ensure these are necessary and sufficient.
+        # capabilities, particularly that sudo and add-apt-repository
+        # commands are available, and HTTPS URLs are supported.
+        run("which sudo || su root -c 'apt-get -y install sudo'"),
```
adamtheturtle: There is a comment on https://clusterhq.atlassian.net/browse/FLOC-1880 which suggests that that issue will be covered, for the client, by this issue. However this adds to the list of necessary commands which are not in the docs for the people using those minimal images. Either improve how this deals with the requirements of FLOC-1880 (commands should be in the docs, docker images tested should be minimal) or update that issue accordingly to mention this series of commands.

admin/client.py
jongiddy: There was a decision propagated at stand-up to leave minor, non-breaking issues in the code rather than adding JIRA issues.

admin/client.py
```diff
@@ -1,176 +1,164 @@
...
     task_client_installation_test,
-    install_cli,
 )
-from effect.twisted import perform
-from flocker.provision._ssh._conch import make_dispatcher
+from flocker.provision._ssh._conch import perform_sudo, perform_put
+from flocker.provision._ssh._model import Run, Sudo, Put, Comment
 
-from .runner import run
+DOCKER_IMAGES = {
```
adamtheturtle: :question: I think that this could be better with `Distribution` objects mapped to Docker images (using `DISTRIBUTION_NAME_MAP`?). We would then have less passing around of strings. Maybe a `Distribution` could hold whether we support it on the client and/or node?

flocker/provision/_install.py
jongiddy: Can you please clarify? The code tested is the same as that shown in the docs, including the addition of the new command.

admin/client.py
jongiddy: OK, although there is no JIRA issue for removing, maybe I should identify the "once Buildbot is updated" to point to FLOC-2813.

adamtheturtle: That or update FLOC-2813 to mention this

flocker/provision/_install.py
adamtheturtle: I misunderstood about the commands being displayed.

However I do think that the commands used here, as they are meant to support minimal images, should be tested on minimal images. Comment near the image map why those images were chosen (e.g. this image does not include `sudo`).

admin/client.py
jongiddy: If Sequence is replaced for docs and cloud tests, it won't take much further effort to change it here.  It's not clear how long it will take for FLOC-2810 to be started and accepted, so I'd rather proceed using the current methods here.

admin/client.py
jongiddy: I consider this to be a separate issue, which will need to be solved here and everywhere else that uses these strings.

admin/client.py
jongiddy: I've gone halfway on this, moving the non-conch dependent performers in the `_conch` module into the `_model` module, and then making that importable through `__init.py__` (`from flocker.provision._ssh import...`)

admin/client.py
jongiddy: It's not bad, in that the client tests are a different application to Flocker node, so each should be able to choose a suitable protocol to talk to Docker, as long as they both choose working versions.

However, it probably makes sense to use the most recent version of the API that works with the lowest supported version of Docker.  And does Docker have a mapping of Docker server versions to supported Docker APIs? No. https://github.com/docker/docker/issues/14008

admin/client.py
jongiddy: Also, added FLOC-2821 to re-factor this properly.

docs/gettinginvolved/acceptance-testing.rst
jongiddy: Moved the client tests off the acceptance testing page, and into the separate client testing page.

flocker/provision/_install.py
jongiddy: The images were chosen primarily because they are the official Docker images for each distribution.  The instructions were then modified to ensure they run on these images. They were not chosen because they are minimal, but they do happen to be about the most minimal images we would expect our users to try running stuff on.

admin/client.py
```diff
@@ -1,176 +1,165 @@
...
+    """
+    Create a shell script file from a sequence of effects.
 
-@implementer(INodeRunner)
-@attributes(RUNNER_ATTRIBUTES + [
-    'provisioner',
-], apply_immutable=True)
-class LibcloudRunner(object):
+    :param bytes dir: The directory in which to create the script.
+    :param Effect effects: An effect which contain the commands,
```
adamtheturtle: contain > contains

admin/client.py
```diff
@@ -1,176 +1,165 @@
...
+    return os.path.basename(filename)
 
-    :ivar LibcloudProvioner provisioner: The provisioner to use to create the
-        nodes.
-    :ivar VolumeBackend volume_backend: The volume backend the nodes are
-        configured with.
+
+class DockerRunner:
+    """
+    Run commands on Docker.
```
adamtheturtle: :question: "on Docker" > "in a Docker container"

adamtheturtle:
Great, thanks @jongiddy . I'll accept so you can continue by making the Buildbot side. If there are significant changes to make that you find once the Buildbot side is done then resubmit, else merge when you're happy.

admin/client.py
jongiddy: Our Ubuntu build slaves use Docker 1.3, which only supports 1.15. Since these tests run on a CentOS buildslave, this was not detected. Found while converting to Jenkins.

PR #1829, opened 2015-08-04
[FLOC-2785] Docker plugin mount API

itamarst:
Relevant documentation: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md#volumedrivermount


flocker/dockerplugin/_api.py
```diff
@@ -157,3 +159,36 @@ def got_configured(configured):
...
+                if datasets and datasets[0].primary == self._node_id:
+                    return datasets[0].path
+                return deferLater(self._reactor, 0.05, get_state)
+            d.addCallback(got_state)
+            return d
+        d.addCallback(lambda _: get_state())
+        d.addCallback(lambda path: {u"Err": None,
+                                    u"Mountpoint": path.path})
+        return d
+
```
jml: There's lint here: flocker/dockerplugin/_api.py:194:1: W391 blank line at end of file (via [buildbot](http://build.clusterhq.com/builders/flocker-lint/builds/8909/steps/lint/logs/stdio)).

flocker/dockerplugin/_api.py
```diff
@@ -157,3 +159,36 @@ def got_configured(configured):
...
+        d = self._flocker_client.move_dataset(self._node_id, dataset_id)
+
+        def get_state():
+            d = self._flocker_client.list_datasets_state()
+
+            def got_state(datasets):
+                datasets = [dataset for dataset in datasets
+                            if dataset.dataset_id == dataset_id]
+                if datasets and datasets[0].primary == self._node_id:
+                    return datasets[0].path
```
jml: What does it mean if there's more than one dataset with that id? I *think* if that happens then we actually want to fail hard because if we have multiple datasets with the same id we're not going to be able to do a good job with data integrity.

What do you think?

flocker/dockerplugin/test/test_api.py
```diff
@@ -112,13 +122,39 @@ def create_after_list():
...
+        """
+        name = u"myvol"
+        dataset_id = UUID(dataset_id_from_name(name))
+        # Create dataset on a different node:
+        d = self.flocker_client.create_dataset(
+            self.NODE_B, DEFAULT_SIZE, metadata={u"name": name},
+            dataset_id=dataset_id)
+
+        # In 0.1 seconds the dataset arrives as state:
+        reactor.callLater(0.1, self.flocker_client.synchronize_state)
```
jml: Is this `0.1` deliberately chosen to be clearly greater than `0.05`? If so, then please make the `0.05` a named constant and make this `2 * self.flocker_client._POLLING_INTERVAL`.

flocker/dockerplugin/_api.py
jml: I kind of would like to read & understand your answer before approving, but in order to prevent stalling roundtrips:

If you decide not to make a change here (and you address the two other points), then feel free to merge. If you do decide to make a change, please bounce back to me for another (quick!) round of review.

jml:
Marking as "Accepted" for now, but bounce back for review if major changes. Must fix tests before landing.

flocker/dockerplugin/_api.py
itamarst: In our current model dataset_id is always unique; lots of code relies on that internally too, and it's explicitly enforced as part of the API interface: https://github.com/ClusterHQ/flocker/blob/master/flocker/control/httpapi.py#L222

flocker/dockerplugin/test/test_api.py
itamarst: Good point, will do.

flocker/dockerplugin/_api.py
jml: I see that as all the more reason to fail hard if it's not the case here (or to encode that assertion in a `get_dataset(dataset_id) -> dataset|None` function). However, reasonable people reasonably disagree about defensive programming, and it's not necessary for this patch.

itamarst: As such I'm going to go with "that's not possible" and leave it at that.

PR #1830, opened 2015-08-05
[FLOC-2733] Versioned configuration

dwgebler:
Design for FLOC-2733, this PR should be closed without merging.

As per discussion with @itamarst the test infrastructure here includes `configuration_vX.json` files containing the actual JSON blob for a particular versioned config, as well as a corresponding `.py` file that is capable of generating such a versioned config. The supporting Python file references the Git commit hash for a version of Flocker itself that generates JSON configs in that version, with the generating code also provided for reference.

Migrations are performed sequentially up to the latest config version, so for example for a version 4 config where a version 1 is found when the persistence service starts, the service will sequentially upgrade 1-2, 2-3, then 3-4.


flocker/control/_persistence.py
```diff
@@ -21,12 +23,98 @@
...
+    """
+    def up(configuration):
+        """
+        Migrate a Vx source configuration dict to a Vy target
+        configuration format and return the updated dictionary.
+
+        :param dict configuration: The Vx configuration.
+        """
+
+    def down(configuration):
```
itamarst: This seems unnecessary. Downgrading is a significantly harder problem than upgrading, so I'd start with not implementing it.

flocker/control/_persistence.py
```diff
@@ -163,15 +252,66 @@ def __init__(self, reactor, path):
...
+        config_files = [
+            (
+                version,
+                self._path.child(_VERSIONED_CONFIG_FILE % version)
+            )
+            for version in range(_CURRENT_VERSION, -1, -1)
+        ]
+        # Check for each possible versioned config file, from newest
+        # to oldest and return the first match.
+        for version, config_file in config_files:
```
itamarst: This method could be improved. Right now it does two distinct things:
1. Find latest existing file.
2. Give file name in case config doesn't exist.

These are not really the same thing; notice how you're duplicating a bunch of code. So probably you want two separate functions for each use case.

flocker/control/_persistence.py
```diff
@@ -163,15 +252,66 @@ def __init__(self, reactor, path):
...
+            )
+        )
+
+    def load_configuration(self):
+        """
+        Load the persisted configuration, upgrading the configuration format
+        if an older version is detected.
+        """
+        self._config_version, self._config_path = self._versioned_config()
+        if self._config_version < _CURRENT_VERSION:
```
itamarst: I think you can just do `<=`; if it's current version it'll just skip the upgrade code. This means you don't need special code path for current version.

flocker/control/_persistence.py
```diff
@@ -163,15 +252,66 @@ def __init__(self, reactor, path):
...
+                b"current_configuration.v%d.json" % _CURRENT_VERSION
+            )
+        )
+
+    def load_configuration(self):
+        """
+        Load the persisted configuration, upgrading the configuration format
+        if an older version is detected.
+        """
+        self._config_version, self._config_path = self._versioned_config()
```
itamarst: There's no reason to set these as instance attributes; they're temporary state. In general avoid mutating or modifying objects when it can be avoid; this will result in code that is simpler to reason about.

flocker/control/test/configurations/configuration_v1.py
```diff
@@ -0,0 +1,71 @@
...
+                        name=u'myapp',
+                        image=DockerImage.from_string(u'postgresql:7.6'),
+                        volume=AttachedVolume(
+                            manifestation=MANIFESTATION,
+                            mountpoint=FilePath(b"/xxx/yyy"))
+                    )],
+                manifestations={DATASET.dataset_id: MANIFESTATION})])
+
+
+class _ConfigurationEncoder(JSONEncoder):
```
itamarst: You shouldn't need to copy/paste this (or any) code. Insofar as you gave a commit hash, someone who wants to reproduce the file can checkout that version of the code and run the script with that version of the code.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +252,32 @@ def test_complex_keys(self):
...
+    def test_v0_v1_configuration(self):
+        """
+        A V0 JSON configuration blob is transformed to a V1 configuration
+        blob, with the result validating when loaded.
+        """
+        v0_json = self.configurations_dir.child(
+            'configuration_v0.json').getContent()
+        v1_json = migrate_configuration(0, 1, v0_json)
+        v1_config = wire_decode(v1_json)
+        self.assertEqual(v1_config, Deployment(nodes=frozenset()))
```
itamarst: This is a reasonable semantic test...

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +252,32 @@ def test_complex_keys(self):
...
+
+    def test_v1_v0_configuration(self):
+        """
+        A V1 JSON configuration blob is transformed to a V0 configuration
+        blob.
+        """
+        config_dict = {_CLASS_MARKER: u"Deployment"}
```
itamarst: But this doesn't actually test that the upgrade did anything useful?

itamarst: Oh, it's downgrade test. You should just not bother with downgrades.

flocker/control/test/test_persistence.py
itamarst: ... although "and it comes up empty" probably isn't the *most* useful assertion to make.

@exarkun can you give some examples of what this sort of test would contain?

dwgebler: Yeah, the problem I had here is that "version 0" configs don't really exist and are made up for the purposes of testing, plus the scope of this branch doesn't define a version 2 config (which will become a thing only when we do the branch that adds Leases to a Deployment) - so I couldn't think of any other example.

flocker/control/test/test_persistence.py
dwgebler: :question: is it not plausible that someone might run a v2 config from a later version of Flocker and then at some point need to revert Flocker to an older version that doesn't support v2? I'm happy to skip downgrades (it means I do less work on this branch, after all), but I've always been of the view that model migrations in a system should be bidirectional.

itamarst: Hm.

itamarst: The thing is that this is just one bit of state among many others: metadata we store on cloud storage, on-disk format in volumes we manage, etc.. Unless we make concerted effort to make *all* of those downgradeable, no point in doing it in just one place.

itamarst: So I'll leave it up to Jean-Paul to decide the big picture question.

itamarst:
I'm mostly happy with this, given comments above, with the caveat that the important bit is testing. So I'll assign this to @exarkun to flesh out that part of the review since he has relevant experience.

flocker/control/_persistence.py
```diff
@@ -21,16 +23,151 @@
...
+        Deserialize the supplied JSON to a ``Configuration`` model.
+
+        :param bytes config: The JSON configuration to deserialize.
+        :return Configuration: The configuration model.
+        """
+
 
-class _ConfigurationEncoder(JSONEncoder):
+@implementer(IConfiguration)
+class Configuration_V1(object):
```
itamarst: In real code you wouldn't typically keep around the old IConfiguration classes since they may not work at all, no? So Configuration_V1 ought not exist.

flocker/control/_persistence.py
```diff
@@ -21,16 +23,151 @@
...
+        Migrate a v1 JSON configuration to v2.
+
+        :param bytes config: The v1 JSON data.
+        :return bytes: The v2 JSON data.
+        """
+        # In reality, this will transform JSON in some way to ensure
+        # the models line up, e.g. by creating a Leases object,
+        # wire_encode'ing it, then adding that JSON in to the Deployment
+        # object's JSON before deserializing and reserialzing.
+        v1_config = Configuration_V1.deserialize(config)
```
itamarst: This is not a JSON structure transformation, and shouldn't rely on Configuration_V1 AFAICT.

I think given this PR actually introduces a new version of the config you don't need to refer to leases at all. Just try to make this code realistic within the context of this particular upgrade.

flocker/control/_persistence.py
```diff
@@ -21,16 +23,151 @@
...
+class ConfigurationMigration(object):
+    """
+    Migrate a JSON configuration from one version to another.
+    """
+    @classmethod
+    def configuration_v1_v2(cls, config):
+        """
+        Migrate a v1 JSON configuration to v2.
+
+        :param bytes config: The v1 JSON data.
```
itamarst: If it's actually specifically upgrading JSON, why is it taking bytes? Either it shouldn't be talking about JSON since it might eventually support other on-disk formats, or maybe it should be taking output of `json.loads`?

itamarst: In later case algorithm would be something like:

```
json_v1 = json.loads(f.read())
json_v2 = migrate_1_to_2(data)
return Configuration_V2.deserialize(json.dumps(json_v2))
```

flocker/control/_persistence.py
```diff
@@ -158,20 +295,50 @@ def __init__(self, reactor, path):
...
+        MultiService.startService(self)
+        _LOG_STARTUP(configuration=self.get()).write(self.logger)
+
+    def load_configuration(self):
+        """
+        Load the persisted configuration, upgrading the configuration format
+        if an older version is detected.
+        """
+        self._config_path = self._path.child(b"current_configuration.json")
+        v1_config_path = self._path.child(
```
itamarst: This seems overcomplicated.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +250,25 @@ def test_complex_keys(self):
...
+    configurations_dir = FilePath(__file__).sibling('configurations')
+
+    def test_v1_v2_configuration(self):
+        """
+        A V0 JSON configuration blob is transformed to a V1 configuration
+        blob, with the result validating when loaded.
+        """
+        v1_json = self.configurations_dir.child(
+            b"configuration_v1.json").getContent()
+        v1_deployment = wire_decode(v1_json)
```
itamarst: You can't use wire_decode here, wire_decode is always the *latest* version.

flocker/control/_persistence.py
dwgebler: :question: we don't want to keep the old `.v1.json` naming convention though? but we do need to see if we're dealing with an old v1 config that needs to be upgraded and such a config will have the old filename.

flocker/control/test/test_persistence.py
itamarst: Remember this test may have to run on v3 and v4 and v5.

flocker/control/_persistence.py
itamarst: Ah, I see.

itamarst:
More generally, can you highlight how this is different from the original proposal? Versioning in file itself, what else?

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +250,24 @@ def test_complex_keys(self):
...
+
+    def test_v1_v2_configuration(self):
+        """
+        A V1 JSON configuration blob is transformed to a V2 configuration
+        blob, with the result validating when loaded.
+        """
+        v1_json = self.configurations_dir.child(
+            b"configuration_v1.json").getContent()
+        v2_json = migrate_configuration(1, 2, v1_json)
+        v2_config = wire_decode(v2_json)
```
itamarst: Still can't use wire_decode because eventually it's going to be the v3 wire_decode. Possibly you need more of a "test_v1_to_latest_configuration" whose assertions may occasionally need to be updated, but ideally not the rest of it. and test a more complex object... or maybe i'm thinking of this wrong. @exarkun how did the axiom tests deal with this?

dwgebler:
> More generally, can you highlight how this is different from the original proposal? Versioning in file itself, what else?

So yeah, this introduces a ``Configuration`` model object that contains the version as part of the persisted state. We drop the `.v1.json` file naming and replace that with all future versions having a set filename of `current_configuration.json`. Downward migrations have been dropped, with migrations taking place in a single ``ConfigurationMigration`` class, which defines individual upgrade methods for Vx to Vy. These are called by a ``migrate_configuration`` method that sequentially performs upgrades from x to y (so that logic's been moved from the original proposal) and will throw a specific exception if a migration path is not defined (e.g. no `v2_v3` migration method exists). The migration methods themselves will take JSON (in this design, as you pointed out, that's a JSON string but can be a dictionary of parsed JSON in the implementation) and transform it before passing their results to the existing generic serialization and deserialization methods (the `wire_` methods). The differences in structure should make it easier to keep consistent configuration tests and see these tests break if we change a model (because we will no longer be able to parse previously valid configurations in to models with changed fields).

Not radically different to my first proposal, but different enough that I feel this will better accommodate further improvement for a long-term solution. I played around with a lot of different ideas in previous iterations of this branch, including versioned model classes and describing models purely as JSON schema (with the resulting `PRecord` derivative classes being generated on the fly). I got encouraging results playing around with the latter, but also determined a full implementation of that strategy would be quite complicated and probably out of scope for this issue.

flocker/control/test/test_persistence.py
dwgebler: Indeed, this is mostly a hangover from a previous design iteration where I had different Configuration parsing classes that determined the encoder and decoder methods used as the hooks in the wire methods. Then I decided that was unnecessarily complicated and reverted to keeping the wire methods as generic serialization functions, partly because they are also used in the AMP messaging service to encode/decode data chunks. But bear in mind these functions can operate on subsets of an overall config, so writing effective v1 to v2 tests will probably not be hard, we just won't be able to `wire_decode` an entire (non-latest) config in the real test implementations.

dwgebler:
I also looked on Friday last week at the idea of producing some kind of hashed representation of a model class's fields in the persisted state to ensure that matched up against the current model definitions in our code, but a) it was problematic to do that in any consistent, reliable and not grossly hacky way and b) it didn't really add much value to the developer experience.

Mainly we want to see a test break if some part of the persisted state changes in our code, e.g. if we add a field to a model. But we can accomplish that in this branch's implementation by exercising a greater range of model tests than we currently have. The only case I can think of where adding a field might then not automatically break a test would be a field that was not mandatory, or had an initial default value and in either of those cases, by definition those would be changes that would not break code utilising the models, so there'd be no reason for them to make existing tests fail - you'd just need new tests to exercise the new field.

dwgebler:
I have simplified the design a bit now, using a `v1_to_latest` config test in the test suite. It is possible to write individual upgrade path tests for the unit testing of the `ConfigurationMigration` methods, which we would do in the implementation in addition to an overall "upgrade version 1 to latest" test.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +251,24 @@ def test_complex_keys(self):
...
+        """
+        A V1 JSON configuration blob is transformed to the latest
+        configuration, with the result validating when loaded.
+        """
+        v1_json = self.configurations_dir.child(
+            b"configuration_v1.json").getContent()
+        upgraded_json = migrate_configuration(1, 2, v1_json)
+        upgraded_config = wire_decode(upgraded_json)
+        expected_configuration = Configuration(
+            version=_CONFIG_VERSION, deployment=TEST_DEPLOYMENT
```
itamarst: The assertion below will break if TEST_DEPLOYMENT changes, so maybe want specific object for these tests? But yeah, basically should work.

dwgebler: Possibly. although an assertion on TEST_DEPLOYMENT or an equivalent is something we *want* to break if we change the Deployment model, or any of the other models it encapsulates.

dwgebler:
Oh and this design version doesn't include logging the upgrades, but logging will be added in the implementation.

itamarst:
OK, so with all my comments above addressed this part of the design should work. However, Jean-Paul's point that we need way to automatically catch model changes still stands, and model hashes do seem the easiest way to do this. I'll open a separate issue for that, though.

Please pick someone to work with you on this and proceed.

PR #1831, opened 2015-08-05
[FLOC-2751] Check if node exists every 15 seconds instead of every 3 seconds

itamarst:
… to reduce chances of hitting rate limits on cloud provider.

Node startup takes a while anyway, so it's unlikely to slow us down much. Longer term we will pre-provision nodes so we won't have this speed hit at all.

### DIFF
--- a/flocker/provision/_libcloud.py
+++ b/flocker/provision/_libcloud.py
@@ -100,7 +100,8 @@ class LibcloudNode(object):
         def do_reboot(_):
             self._node.reboot()
             self._node, self.addresses = (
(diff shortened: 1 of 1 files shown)


jml:
Looks good. Merge when it passes tests & you're ready.

wallrj:
There were a couple of build errors:
 * Known flaky-test error: https://clusterhq.atlassian.net/browse/FLOC-2668
 * New flaky-test error: https://clusterhq.atlassian.net/browse/FLOC-2826

I restarted the failed builds.

PR #1832, opened 2015-08-06
[FLOC-2753] Real Flocker client

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-2753

I filed a follow-up issue to replace as much of possible of the existing acceptance code's client with the new client, now that it is implemented.


flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        :param int port: Port to connect to:
+        :param FilePath ca_cluster_path: Path to cluster's CA certificate.
+        :param FilePath cert_path: Path to user certificate.
+        :param FilePath key_path: Path to user private key.
+        """
+        self._treq = treq_with_authentication(reactor, ca_cluster_path,
+                                              cert_path, key_path)
+        self._base_url = b"https://{}:{}/v1".format(host, port)
+
+    def _request(self, method, path, body, success_code, error_codes={}):
```
wallrj:
```
error_codes=None
if error_codes is None:
    error_codes = {}
``` etc.

File: flocker/apiclient/_client.py [hunk 2]
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        :return: ``Deferred`` firing with decoded JSON.
+        """
+        def error(body, code):
+            if code in error_codes:
+                raise error_codes[code](body)
+            raise ResponseError(code, body)
+
+        def got_result(result):
+            if result.code == success_code:
+                return json_content(result)
```
wallrj: :question: I guess `json_content` deals with empty strings for empty response body cases.

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+            else:
+                d = content(result)
```
wallrj: :question: Does the content need to be `utf-8` decoded before being passed to the exception initializer?

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        :param reactor: Reactor to use for connections.
+        :param bytes host: Host to connect to.
+        :param int port: Port to connect to:
+        :param FilePath ca_cluster_path: Path to cluster's CA certificate.
+        :param FilePath cert_path: Path to user certificate.
+        :param FilePath key_path: Path to user private key.
+        """
+        self._treq = treq_with_authentication(reactor, ca_cluster_path,
+                                              cert_path, key_path)
+        self._base_url = b"https://{}:{}/v1".format(host, port)
```
wallrj: :question: I've been told in previous code reviews not to use format on byte strings in anticipation of Python3 (I think) where the method isn't available.

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+                return json_content(result)
+            else:
+                d = content(result)
+                d.addCallback(error, result.code)
+                return d
+
+        headers = None
+        data = None
+        if body is not None:
+            headers = {b"content-type": b"application/json"}
```
wallrj: :question: Should this be a `Headers` instance? http://treq.readthedocs.org/en/latest/api.html#treq.request

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+            persistent=False
+        )
+        request.addCallback(got_result)
+        return request
+
+    def _parse_configuration_dataset(self, dataset_dict):
+        """
+        Convert a dictionary decoded from JSON with a dataset's configuration.
+
+        :param dataset_dict: Dictionary describing a dataset.
```
wallrj: :question: `:param dict dataset_dict:` if you feel like it.

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        :return: ``Dataset`` instance.
+        """
+        return Dataset(primary=UUID(dataset_dict[u"primary"]),
+                       maximum_size=dataset_dict[u"maximum_size"],
+                       dataset_id=UUID(dataset_dict[u"dataset_id"]),
+                       metadata=dataset_dict[u"metadata"])
+
+    def create_dataset(self, primary, maximum_size, dataset_id=None,
+                       metadata=pmap()):
```
wallrj: Aha, immutable default arguments!

wallrj: :question: Can / should `maximum_size` be an optional argument? `flocker-dataset-agent` will assign minimum_allocatable size (or something), right?

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        dataset = {u"primary": unicode(primary),
+                   u"maximum_size": maximum_size,
+                   u"metadata": dict(metadata)}
+        if dataset_id is not None:
+            dataset[u"dataset_id"] = unicode(dataset_id)
+
+        request = self._request(b"POST", b"/configuration/datasets",
```
wallrj: :question: Doesn't `twisted.web` have constants for HTTP methods?

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+                                dataset, CREATED,
+                                {CONFLICT: DatasetAlreadyExists})
+        request.addCallback(self._parse_configuration_dataset)
+        return request
+
+    def move_dataset(self, primary, dataset_id):
+        request = self._request(
+            b"POST", b"/configuration/datasets/{}".format(dataset_id),
```
wallrj: :question: Maybe don't use `bytes.format` or maybe use unicode here and elsewhere in this branch when specifying paths for requests.

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        request = self._treq.request(
+            method, self._base_url + path,
+            data=data, headers=headers,
+            # Keep tests from having dirty reactor problems:
+            persistent=False
+        )
+        request.addCallback(got_result)
+        return request
+
+    def _parse_configuration_dataset(self, dataset_dict):
```
wallrj: :question: This might as well be a free function I think.

flocker/apiclient/_client.py
```diff
@@ -144,4 +150,127 @@ def synchronize_state(self):
...
+        request = self._request(b"GET", b"/configuration/datasets", None, OK)
+        request.addCallback(
+            lambda results:
+            [self._parse_configuration_dataset(d) for d in results])
+        return request
+
+    def list_datasets_state(self):
+        request = self._request(b"GET", b"/state/datasets", None, OK)
+
+        def parse_dataset_state(dataset_dict):
```
wallrj: :question: This could be a module level function too.

wallrj: Documented too.

flocker/apiclient/test/test_client.py
```diff
@@ -31,23 +42,22 @@ def make_clientv1_tests(client_factory, synchronize_state):
     control of this process. So when testing a real client it will be
     talking to a in-process server.
 
-    :param client_factory: 0-argument callable that returns a
-        ``IFlockerAPIV1Client`` provider.
+    The ``TestCase`` should have two 0-argument methods:
```
wallrj: :question: You prefer this style? It forces us to create a subclass, but we do that anyway as a place to hang the TestCase specific documentation. So fine.

flocker/apiclient/test/test_client.py
```diff
@@ -31,23 +42,22 @@ def make_clientv1_tests(client_factory, synchronize_state):
...
 
-    :param synchronize_state: 1-argument callable that takes a client
-        instances and makes or waits for state to match configuration.
+    create_client: Returns a ``IFlockerAPIV1Client`` provider.
+    synchronize_state: Make state match the configuration.
     """
     class InterfaceTests(TestCase):
```
wallrj: :question:  Might be nice to move the tests to a mixin in the style of the tests for other Flocker verified fakes.

wallrj: Probably not here though -- it'll make the patch unnecessarily large.

wallrj:
Thanks @itamarst 

This all looks very clean and elegant.
The build two failures seem unrelated.

Please:
 * answer or address the comments above,
 * create followup issues where applicable,
 * create issues (if necessary) for the two build failures, then
 * merge.

wallrj:
PS coverage looks complete. I wonder why the interface methods aren't measured?
 * http://build.clusterhq.com/results/complete-coverage/flocker-client-FLOC-2753/build-8335/flocker_apiclient__client.html

I'd have thought `z.i.verify.verifyObject` would take care of that.

flocker/apiclient/_client.py
itamarst: Apparently `%` got added to Python 3.5?!?! Sigh.

flocker/apiclient/_client.py
itamarst: Apparently, yeah.

itamarst: Or actually we don't have any cases of that sort yet.

flocker/apiclient/_client.py
itamarst: You can pass any object you want, really.

flocker/apiclient/_client.py
itamarst: Documentation uses dict elsewhere, and it works, so I'll stick to this I guess.

flocker/apiclient/_client.py
itamarst: It will, but that's kinda bogus picked-out-of-a-hat number. Better to force the caller to pick a number out of a hat, as the Docker plugin does :wink:

flocker/apiclient/_client.py
itamarst: Just response codes AFAICT.

flocker/apiclient/_client.py
itamarst: Switched to `%`.

flocker/apiclient/_client.py
itamarst: If it ever becomes generally useful, yeah.

flocker/apiclient/test/test_client.py
itamarst: Don't prefer this style, but was forced to due to Reasons of code structure.

flocker/apiclient/test/test_client.py
itamarst: Yeah...

itamarst:
`zope.interface.Interface` is... special somehow.

PR #1833, opened 2015-08-06
[FLOC-2786] Docker plugin get path API

itamarst:
Relevant Docker docs: https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md#volumedriverpath

### DIFF
--- a/flocker/dockerplugin/_api.py
+++ b/flocker/dockerplugin/_api.py
@@ -66,7 +66,7 @@ class VolumePlugin(object):
     can't be sure they won't change things in minor ways. We do validate
     outputs to ensure we output the documented requirements.
     """
-    _POLL_INTERNVAL = 0.05
+    _POLL_INTERVAL = 0.05
 
     app = Klein()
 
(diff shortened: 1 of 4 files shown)


flocker/dockerplugin/_api.py
```diff
@@ -162,23 +162,18 @@ def got_configured(configured):
...
 
-        :return: Result indicating success.
+        :return: Result that includes the mountpoint.
         """
-        dataset_id = UUID(dataset_id_from_name(Name))
-        d = self._flocker_client.move_dataset(self._node_id, dataset_id)
-
+        # If we ever get rid of dataset_id hashing hack we'll need to
+        # lookup the dataset by its metadata, not its id.
         def get_state():
```
jml: :question: Is there any particular need for this inner function? I think it would be simpler without it.

flocker/dockerplugin/test/test_api.py
```diff
@@ -153,6 +153,29 @@ def test_mount(self):
...
+    def test_path(self):
+        """
+        ``/VolumeDriver.Path`` returns the mount path of the given volume.
+        """
+        name = u"myvol"
+        dataset_id = UUID(dataset_id_from_name(name))
+
+        d = self.create(name)
+        # After two polling intervals the dataset arrives as state:
+        reactor.callLater(VolumePlugin._POLL_INTERVAL,
```
jml: This looks like one polling interval, rather than two.

jml:
LGTM. I don't think this needs another review. But please bounce back if there are changes required to make the tests pass.

flocker/dockerplugin/_api.py
adamtheturtle: I'm not sure how to do this more neatly, given `get_state` is currently the callable in the `deferLater` after this.

adamtheturtle:
Note that the one failing test was `coverage` which is no longer a required builder. I will merge when the tests have finished.

flocker/dockerplugin/_api.py
jongiddy: Maybe add a comment explaining the implementation here: that `get_state` and `got_state` loop, polling for the state, and terminate when the primary manifestation of the dataset is on the local node.

PR #1834, opened 2015-08-11
[FLOC-2850] Reduce API version to support Docker 1.3

jongiddy:
Ubuntu buildslaves currently have Docker 1.3.3 on them, which supports up to API 1.15.  1.15 is also the Docker API version used in `flocker.node._docker`.

However, the client tests use API v1.18. This has not been a problem, since the client tests, although they run on Ubuntu, are run in a Docker container on a CentOS slave.  CentOS has Docker 1.5 installed, which supports API v 1.18.

On Jenkins, the DEB files are created as part of the test, and lintian complains about the permissions set on files due to the umask setting.  Hence it is simpler to use the Ubuntu slave to run the tests, since it already has this problem solved for the acceptance tests.

Since there is no particular reason to use API v1.18, it makes sense to keep it consistent with `_docker.py`, at 1.15, especially since it does solve this problem.  Once all our systems are confirmed to be running later versions of Docker, we can upgrade the API consistently throughout.


jongiddy:
Closing this PR as the test does not work with a server whose max API is 1.15 due to use of `exec_inspect` call. See https://github.com/docker/docker-py/pull/711

It passed the test on my machine and on Buildbot because both have a later version of Docker. The Docker server is happy to respond to a recognised post-1.15 API request from a client claiming to be 1.15.

PR #1836, opened 2015-08-11
[FLOC-2851] Configuration model change detection

itamarst:
I have deliberately introduced a change to the model which should cause the model detection to trigger, see Buildbot results below which ought to have a failing test.

### DIFF
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -379,9 +379,9 @@ def persist_model():
     git_hash = check_output(
         [b"git", b"show", b'--format=%H', b"-s"]).strip()
     model = generate_model()
-    PERSISTED_MODEL.setContent(dumps({
-        u"git_hash": git_hash, u"model": model,
-    }))
+    PERSISTED_MODEL.setContent(dumps(
+        {u"git_hash": git_hash, u"model": model}, sort_keys=True, indent=4,
+        separators=(',', ': ')))
     print("Persisted model for git hash {}.".format(git_hash))
 
 
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -10,7 +10,9 @@ code is always implemented when necessary.
 from json import loads, dumps
 from subprocess import check_output
 
-from pyrsistent import PRecord
+from pyrsistent import (
+    PRecord, PClass, CheckedPSet, CheckedPVector, CheckedPMap, field,
+)
 
 from twisted.trial.unittest import SynchronousTestCase
 from twisted.python.filepath import FilePath
@@ -25,18 +27,21 @@ PERSISTED_MODEL = FilePath(__file__).sibling(b"persisted_model.json")
 ROOT_CLASS = Deployment
 
 
-def generate_model():
+def generate_model(root_class=ROOT_CLASS):
     """
     Generate a data-structure that represents the current configuration
     model.
 
     Changes to output may require regenerating the persisted version
     on-disk.
+
+    This should switch to Pyrsistent's introspection API once it exists:
+    https://github.com/tobgu/pyrsistent/issues/47
     """
     classes_result = {}
-    result = {u"root": fqpn(ROOT_CLASS),
+    result = {u"root": fqpn(root_class),
               u"classes": classes_result}
-    classes = {ROOT_CLASS}
+    classes = {root_class}
     while classes:
         klass = classes.pop()
         klass_name = fqpn(klass)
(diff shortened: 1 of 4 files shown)


flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,104 @@
...
+        klass_name = fqpn(klass)
+        if klass_name in classes_result:
+            continue
+        record = None
+        if issubclass(klass, PRecord):
+            record = {u"category": u"record",
+                      u"fields": {}}
+            # XXX file issue with pyrsistent to add introspection API if I
+            # can't find anything in docs:
+            for name, field in klass._precord_fields.items():
```
dwgebler: This was the only way I could find to map the fields in Pyrsistent record types.

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,104 @@
...
+                record[u"fields"][name] = list(fqpn(cls) for cls in field.type)
+                for cls in field.type:
+                    classes.add(cls)
+        # XXX CheckedPMap and friends
```
dwgebler: I can see this turning in to quite a long function with a lot of repetitive similar logic for the varying introspection that has to occur for each field type. The next one would be a handler for the generated `flocker.control._model.NodePSet` which contains the mapping to `_model.Node`, where `NodePSet` is some kind of (undocumented) meta type. Then there are all the others.

Also, what happens if we add something to a model of a field type that isn't covered here? The test won't break, but the model will have changed. I'm not sure how likely that is to happen, but it seems plausible.

dwgebler:
I'd be interested in seeing a design that used hashing of (for example) a tuple of the field names in each of our defined models and then checking that against a hash generated on the fly from the current models in the test. I think that might be easier and clearer than introspection - not sure it would be any harder to implement, either.

This does sort of work, but the reason I shyed away from doing something like this in the versioned config branch was because it seems very forced....to me, if you change a model, a test somewhere should fail because some aspect of the software no longer works with the changed model, rather than because a test is introspecting your code and detecting it's not the same as it used to be.

But maybe this is sufficient for our purposes of developer experience.

wallrj:
Thanks @itamarst 

Dave's made some points above. Please respond to those.

I note that the tests failed not because of the schema change but because the new JSON file isn't included in the `MANIFEST.in` file (I think)

```
[ERROR]
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/control/test/test_model_change.py", line 98, in test_model_changed
    previous_model = loads(PERSISTED_MODEL.getContent())[u"model"]
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/filepath.py", line 290, in getContent
    fp = self.open()
  File "/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/twisted/python/filepath.py", line 845, in open
    return open(self.path, mode + 'b')
exceptions.IOError: [Errno 2] No such file or directory: '/srv/buildslave/flocker-centos-7/venv/lib/python2.7/site-packages/flocker/control/test/persisted_model.json'

flocker.control.test.test_model_change.ConfigurationModelChanged.test_model_changed
```


The dumped JSON model looks reasonable to me:

```
{
  "git_hash": "dea57a9396e9b86b3afaf69f50f03d73b284f3f4",
  "model": {
    "root": "flocker.control._model.Deployment",
    "classes": {
      "flocker.control._model.Deployment": {
        "fields": {
          "nodes": [
            "flocker.control._model.NodePSet"
          ]
        },
        "category": "record"
      },
      "flocker.control._model.NodePSet": null
    }
  }
}

```

This will highlight changes in the allowed types in the model, but what about other constraints and invariants?

I think @dwgebler made the point above, but what if, for example, we change the model to so that a previously optional field becomes mandatory. That would require a schema upgrade -- configurations which had a null field value would need to be assigned a sensible default -- but this test wouldn't catch that, right?

Or what if, for example, we introduce a new invariant? Should those be (somehow) included in the json dump?

If those are problems perhaps they can be addressed in follow up issues.

Please proceed unless you want more discussion / feedback.

flocker/control/test/test_model_change.py
itamarst: I filed an issue with `pyrsistent` to get something better, yeah.

flocker/control/test/test_model_change.py
itamarst: We currently need to support PRecord/PClass, CheckedPSet and CheckedPMap and that'll cover pretty much everything. If we change to a different class ... then that will be caught by the fact that the output changed. So we'd need to write more code, but we'd know that since test would fail.

Note that we don't cover "field types", there really isn't any such thing. A field points to types, it doesn't have a type.

itamarst:
Re Dave's long comment: the purpose of failing test when model changes is to tell the developer they need to write upgrade code. That's it. And existing test coverage is useless for that.

itamarst:
"Optional" field is ambiguous; currently it means "can be type None" which would be covered by listing possible types. If you mean "does not need to be set at all" then that's probably a thing we shouldn't be doing at all (and currently don't AFAIK).

Bigger picture: right now if you change model nothing happens to tell you that you need to write an upgrader. This code is likely to not be perfect, but catch let's say 90% of changes. Telling developers 90% of time that they need to write an upgrader is a lot better than never telling them that. The *absence* of this warning doesn't mean you shouldn't write an upgrader; anytime you touch the model you should be considering that.

PR #1837, opened 2015-08-11
[FLOC-1059] [FLOC-2081] [FLOC-2279] [FLOC-2402] [FLOC-2447] [FLOC-2524]

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1059
Fixes https://clusterhq.atlassian.net/browse/FLOC-2081
Fixes https://clusterhq.atlassian.net/browse/FLOC-2279
Fixes https://clusterhq.atlassian.net/browse/FLOC-2402
Fixes https://clusterhq.atlassian.net/browse/FLOC-2447
Fixes https://clusterhq.atlassian.net/browse/FLOC-2524

This removes the dependency declared on Docker from the native packages.  Since this means that installing the Flocker packages no longer drags in the Docker package, it also updates the provisioning code to explicitly install Docker.  It uses the Docker-supplied script for doing so (and therefore lets the version of Docker float based on the whim of that script).  Some tests which were broken by that change and were highly duplicative of the implementation they tested have also been deleted.

It also updates the installation documentation to instruct users to install Docker.


docs/install/install-node.rst
```diff
@@ -159,11 +159,8 @@ Paste them into a root console on the target node:
 .. task:: install_flocker centos-7
    :prompt: [root@centos]#
 
-Installing ``flocker-node`` will automatically install Docker, but the ``docker`` service may not have been enabled or started.
-To enable and start Docker, run the following commands in a root console:
-
-.. task:: enable_docker centos-7
-   :prompt: [root@centos]#
+Flocker's container management features depend on Docker.
+Make sure `a recent version of Docker is installed <https://docs.docker.com/docker/installation/>` and running.
```
jongiddy: I think we should include a minimum version here.

docs/install/install-node.rst
```diff
@@ -181,6 +178,9 @@ Setup the pre-requisite repositories and install the ``clusterhq-flocker-node``
 .. task:: install_flocker ubuntu-14.04
    :prompt: [root@ubuntu]#
 
+Flocker's container management features depend on Docker.
+Make sure `a recent version of Docker is installed <https://docs.docker.com/docker/installation/>` and running.
+
```
jongiddy: Ditto, minimum version here.

jml:
The four failing `client/(aws|rackspace)/ubuntu-1[45].04` tests can be ignored, since those builders have gone away and been replaced by simpler ones.

jongiddy:
Since we're not pinning the Docker version, why is it preferable to use the Docker downloads rather than using the CentOS `docker` package from the `extras` repo (currently 1.7.1) and the Ubuntu 15.04 `docker.io` package (currently 1.5) for testing?  If we used the `wget` approach for Ubuntu 14.04, then we would have a good range of Docker versions being tested. This would seem to give us a better chance of finding problems before users.

jongiddy:
CentOS 7 acceptance test on AWS is failing. running the steps manually gives:
```
# systemctl start docker.service
Job for docker.service failed. See 'systemctl status docker.service' and 'journalctl -xn' for details.
# systemctl status docker.service -l
docker.service - Docker Application Container Engine
   Loaded: loaded (/usr/lib/systemd/system/docker.service; enabled)
   Active: failed (Result: exit-code) since Wed 2015-08-12 11:24:11 UTC; 17s ago
     Docs: https://docs.docker.com
  Process: 8013 ExecStart=/usr/bin/docker daemon -H fd:// (code=exited, status=127)
 Main PID: 8013 (code=exited, status=127)

Aug 12 11:24:11 ip-172-31-13-88 docker[8013]: /usr/bin/docker: relocation error: /usr/bin/docker: symbol dm_task_get_info_with_deferred_remove, version Base not defined in file libdevmapper.so.1.02 with link time reference
Aug 12 11:24:11 ip-172-31-13-88 systemd[1]: docker.service: main process exited, code=exited, status=127/n/a
Aug 12 11:24:11 ip-172-31-13-88 systemd[1]: Failed to start Docker Application Container Engine.
Aug 12 11:24:11 ip-172-31-13-88 systemd[1]: Unit docker.service entered failed state.
```

jongiddy:
I think the problem with CentOS on AWS is related to the device-mapper-libs package problem listed in https://rhn.redhat.com/errata/RHBA-2015-0993.html 

This could be another reason to use the OS package for CentOS, at least.

docs/install/install-node.rst
jongiddy: The minimum version that supports API v1.15 is 1.3.0, but FLOC-1059 says our minimum version should be 1.3.3, due to security problems.

jongiddy:
It seems like we should have some check in the agent service for the existence of Docker, and for the minimum required version, to ensure that any error specifically identifies the problem of a missing or too-old Docker.  I will try to see what error is currently given if Docker is missing, but in the meantime, returning this to Code Backlog for addressing the above comments.

docs/install/install-node.rst
exarkun: Yep, 1.8 is the minimum required version if we want all the issues I claimed this PR fixes to actually be fixed.  1.8 is also what you'll get (as of today) if you follow the linked installation instructions.

I could put some words about installing 1.8 here.  I have mixed feelings about it.  Further input welcome.

exarkun:
> why is it preferable to use the Docker downloads

Probably so that users get the newest version of Docker which seems to be what we want users to get.  For example, 1.7.1 still has the firewalld mis-interaction but 1.8 fixes this.

jongiddy:
OK, I am happy for the docs to suggest that users install the latest Docker using the Docker downloads. But I expect some users will use OS packages to install Docker, probably before they come anywhere near Flocker (e.g. the people who use Ansible, Puppet, etc. to keep their systems uniform). Hence, I still think it is worth testing using the different methods described above for each supported OS. (3 tests in total for 3 platforms, not multiple combinations of platform and install method).

jongiddy:
> I will try to see what error is currently given if Docker is missing, but in the meantime, returning this to Code Backlog for addressing the above comments.

If Docker is not installed, there is no error until a client attempts to start a container. For such a singular error, it would be better for the Flocker agent to check, on startup, its connectivity to Docker, verify the version, and exit with a specific error if it is not happy.  This would work for the container agent, not sure how it translates to the dataset agent.

exarkun:
Points not yet addressed:

  * I haven't changed the behavior of the container agent in the case where Docker isn't installed.  I agree the current behavior isn't ideal.  Considering our plans for the container agent and the lack of a general way to do centralized error reporting, I'm not sure there's much that can be done about this that's actually worthwhile.
  * I haven't done anything to expand testing to cover the different Docker packages.  I'm reluctant to undertake any non-trivial BuildBot reconfiguration at the moment given our plans for BuildBot.  Without knowing more about which packages users actually want to use (not just which packages they happen to have arbitrarily selected) I'm also not sure how much value there is in testing additional configurations.  It would be nice to have a systematic way to determine which of these packages users might actually benefit from using so we can focus our efforts on those.

flocker/provision/_install.py
```diff
@@ -376,7 +388,7 @@ def task_upgrade_kernel(distribution):
     """
     Upgrade kernel.
     """
-    if distribution == 'centos-7':
+    if is_centos(distribution):
```
adamtheturtle: :question: Use `flocker.admin.DISTRIBUTION_NAME_MAP` and `Distribution.name`.

docs/install/install-node.rst
```diff
@@ -159,11 +159,8 @@ Paste them into a root console on the target node:
 .. task:: install_flocker centos-7
    :prompt: [root@centos]#
 
-Installing ``flocker-node`` will automatically install Docker, but the ``docker`` service may not have been enabled or started.
-To enable and start Docker, run the following commands in a root console:
-
-.. task:: enable_docker centos-7
-   :prompt: [root@centos]#
+Flocker's container management features depend on Docker.
+Make sure `a recent version of Docker (at least 1.8) is installed <https://docs.docker.com/docker/installation/>` and running.
```
adamtheturtle: :question: Remove "recent" - it doesn't add anything if you've already got "at least 1.8" and increases the likelihood of our docs looking outdated if we neglect to update this and leave the word "recent" there for a long time.

flocker/provision/_install.py
exarkun: It's `admin.DISTRIBUTION_NAME_MAP` and `flocker.provision` can't depend on `admin`.  Also probably not really worthwhile considering we should be replacing `flocker.provision` with something not hand-rolled as soon as we can.

itamarst:
Re container agent, we could add a "Requires=docker.service" for systemd in the `[Unit]` section, and equivalent upstart thing. And then it would fail to start in a meaningful way. E.g. if I configure `Requires=xxx.service` then I get failure on `systemctl start flocker-container-agent`:

```
Failed to issue method call: Unit xxx.service failed to load: No such file or directory.
```

itamarst:
Re testing, I'd say something like "Test with current and previous major release" and that's about it, given pace of development. But insofar as we're switching from hard-coded version we can do that as follow-up.

flocker/provision/_install.py
```diff
@@ -822,9 +895,6 @@ def task_install_flocker(
             run_from_args([
                 "apt-get", "-y", "install", "apt-transport-https",
                 "software-properties-common"]),
-            # Add Docker repo for recent Docker versions
```
itamarst: woah, yeah, sketchy.

vagrant/tutorial/post-reboot-bootstrap.py
```diff
@@ -66,12 +66,14 @@
 # Install ZFS.
 check_call(['yum', 'install', '-y', 'zfs'])
 
-# Enable docker.
-# We don't need to start it, since when the box is packaged,
+# Install Docker.
+check_call(["curl", "-o", "/tmp/install-docker.sh", "https://get.docker.com/"])
```
itamarst: How did this end up being installed before?

itamarst:
I did only a minimal review insofar as it's been through some rounds.

1. Address minor comments.
2. Add the `Requires=` thing for flocker-container-agent systemd unit file, and equivalent for upstart if there is one. Mysteriously non-working clusters are bad.
3. Then merge when green.

itamarst:
Oh, and:
4. Add appropriate news entry to documentation.

itamarst:
And I guess:
5. Does anything need to be done to upgrade an existing install? If so document it.

vagrant/tutorial/post-reboot-bootstrap.py
exarkun: The Flocker OS package declared a dependency on some Docker package and the OS package manager arranged for that Docker package to get installed when we installed the Flocker OS package.

exarkun:
> Add appropriate news entry to documentation.

I added to the release notes even though you said "news entry".  There's a NEWS file but I think you meant the release notes.  If not, well, oops.

> Does anything need to be done to upgrade an existing install? If so document it.

Probably not.  Although maybe we should suggest that people do upgrade to a newer version of Docker?  Dunno.

Really, we should have some kind of testing in place to verify it's possible to upgrade our software...

PR #1840, opened 2015-08-12
[FLOC-1968] Automate copyright changes in license

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-1968

This duplicates the LICENSE file into a template and makes two changes to it:
* A set of braces "`{}`" are replaced with "`{{}}`", but this does not change the rendered file
* Instead of `2014, 2015` the LICENSE file now has `2014-{current_year}`.

### DIFF
--- a/admin/LICENSE.template
+++ b/admin/LICENSE.template
@@ -1,0 +1,201 @@
+                                Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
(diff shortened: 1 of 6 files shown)


admin/LICENSE.template
```diff
@@ -0,0 +1,201 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
jongiddy: Remove the APPENDIX, down to the Copyright notice. No-one can use this section once the braces have been replaced, and it is clearly not a part of the actual licence.

admin/release.py
```diff
@@ -1076,3 +1077,16 @@ def publish_dev_box_main(args, base_path, top_level):
             ),
         ]),
     )
+
+
+def update_license_file(args, top_level, year=datetime.now().year):
```
jongiddy: By default `now()` returns the date in the local timezone. This has the perverse effect that the copyright could go backwards. OK, we're unlikely to do a release during our New Year's party, but it's also pretty simple to add `pytz.UTC` as a parameter. And then we're good for the NYEReleaseathon.

adamtheturtle: > OK, we're unlikely to do a release during our New Year's party

Hopefully Jenkins will be doing it for us.

docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -122,6 +122,8 @@ Preparing For a Release
       mkvirtualenv flocker-release
       pip install --editable .[dev]
       admin/create-release-branch --flocker-version="${VERSION}"
+      admin/update-license
+      git commit -am "Updated copyright in LICENSE file"
```
jongiddy: :question: Comment that this last line exits with 1 if the license does not need to be updated. May be useful to remember if this step is automated in the future.

jongiddy:
Thanks. Address comments and merge when green.

docs/gettinginvolved/infrastructure/release-process.rst
adamtheturtle: How about `git commit --allow-empty` ?

jongiddy: No, that would actually create an empty commit, and suggets that the copyright was updated when it wasn't. The behaviour of `git` is fine as is. It is just that a future automated script will need to deal with this returning non-zero.

Maybe have `update-licence` return a specific exit code meaning, "no error, but also no update". This can be ignored by the release engineer now, but would help develop a solid automated script later.

adamtheturtle: I'll leave this - I think that whoever puts this into a script can work it out.

PR #1842, opened 2015-08-13
[FLOC-2851] Configuration model change detection

itamarst:
If someone changes `flocker.control._model` this should detect that and cause a failing test, warning the developer that they need to write upgrade code for our on-disk format.

Support for upgrade code is being implemented in a different branch.

### DIFF
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -43,7 +43,7 @@ def _precord_model(klass):
     record = {u"category": u"record",
               u"fields": {}}
     for name, field_info in getattr(klass, attr_name).items():
-        record[u"fields"][name] = list(
+        record[u"fields"][name] = sorted(
             fqpn(cls) for cls in field_info.type)
         for cls in field_info.type:
             further_classes.add(cls)
@@ -60,9 +60,9 @@ def _pmap_model(klass):
     record = {
         u"category": u"map",
         u"fields": {
-            u"key": list(
+            u"key": sorted(
                 fqpn(cls) for cls in klass._checked_key_types),
-            u"value": list(
+            u"value": sorted(
                 fqpn(cls) for cls in klass._checked_value_types)}}
     further_classes = set()
     for cls in klass._checked_key_types + klass._checked_value_types:
@@ -81,7 +81,7 @@ def _psequence_model(klass):
     category = u"set" if issubclass(klass, CheckedPSet) else u"list"
     record = {
         u"category": category,
-        u"type": list(fqpn(cls) for cls in klass._checked_types),
+        u"type": sorted(fqpn(cls) for cls in klass._checked_types),
     }
     further_classes = set(klass._checked_types)
     return record, further_classes
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -379,9 +379,9 @@ def persist_model():
     git_hash = check_output(
         [b"git", b"show", b'--format=%H', b"-s"]).strip()
     model = generate_model()
-    PERSISTED_MODEL.setContent(dumps({
-        u"git_hash": git_hash, u"model": model,
-    }))
+    PERSISTED_MODEL.setContent(dumps(
+        {u"git_hash": git_hash, u"model": model}, sort_keys=True, indent=4,
+        separators=(',', ': ')))
     print("Persisted model for git hash {}.".format(git_hash))
 
 
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -17,14 +17,10 @@ from twisted.trial.unittest import SynchronousTestCase
 from twisted.python.filepath import FilePath
 from twisted.python.reflect import qual as fqpn
 
-from .._model import Deployment
+from .._persistence import ROOT_CLASS
 from ... import __version__
 
 PERSISTED_MODEL = FilePath(__file__).sibling(b"persisted_model.json")
-
-# The class at the root of the configuration tree. This may need to be
-# changed if the configuration's root class changes.
-ROOT_CLASS = Deployment
 
 
 def _precord_model(klass):
--- a/flocker/control/test/test_model_change.py
+++ b/flocker/control/test/test_model_change.py
@@ -27,6 +27,66 @@ PERSISTED_MODEL = FilePath(__file__).sibling(b"persisted_model.json")
 ROOT_CLASS = Deployment
 
 
+def _precord_model(klass):
+    """
+    Serialize a ``PRecord`` or ``PClass`` model to something
+    JSON-encodable.
+
+    :param klass: A ``PRecord`` or ``PClass`` subclass.
+    :return: Tuple of (model dictionary, further classes to process).
+    """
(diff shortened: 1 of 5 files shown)


flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+    result = {u"root": fqpn(root_class),
+              u"classes": classes_result}
+    classes = {root_class}
+    while classes:
+        klass = classes.pop()
+        klass_name = fqpn(klass)
+        if klass_name in classes_result:
+            continue
+        record = None
+        if issubclass(klass, (PRecord, PClass)):
```
exarkun: It would be nice to just break this function into tinier pieces -- probably along the lines of this type check -- now, rather than waiting for it to be 200 lines long and moaning about how awful it is for six months.

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+            if issubclass(klass, PRecord):
+                attr_name = "_precord_fields"
+            else:
+                attr_name = "_pclass_fields"
+            record = {u"category": u"record",
+                      u"fields": {}}
+            for name, field_info in getattr(klass, attr_name).items():
+                record[u"fields"][name] = list(
+                    fqpn(cls) for cls in field_info.type)
```
exarkun: It seems that we can't rely on the order of `field_info.type` being stable.  The test suite is failing now, at least in part because:

```
              'flocker.control._model.Application': {u'category': u'record',
                                                     u'fields': {'command_line': ['flocker.control._model.UnicodePVector',
                                                                                  '__builtin__.NoneType'],
...
```
does not equal
```
              u'flocker.control._model.Application': {u'category': u'record',
                                                      u'fields': {u'command_line': [u'__builtin__.NoneType',
                                                                                    u'flocker.control._model.UnicodePVector'],
...
```

Maybe there are other places where order is non-deterministic.  I stopped looking for differences after I found one.

(Awful, awful, awful presentation of the failure, btw :/ Reminds me that we haven't yet upgraded to the version of Twisted that includes the better formatting functionality, though I don't know how much it would help in this specific case.)

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+    A sub-type used in ``GenerateModelTests``.
+    """
+
+
+class ChangedSubtype(PRecord):
+    """
+    A changed variant of ``Subtype``.
+    """
+    x = field()
+ChangedSubtype.__name__ = "Subtype"
```
exarkun: :question: I wonder if this idiom would be better:

```py
class Subtype(PRecord):
    ...
OriginalSubtype = Subtype

class Subtype(OriginalSubtype):
    ...
```

in case `__name__` isn't the on place the name gets recorded.

exarkun: (previous comment edited)

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+        when you need to write upgrade code.
+        """
+        current_model = generate_model()
+        previous_model = loads(PERSISTED_MODEL.getContent())[u"model"]
+        self.assertEqual(current_model, previous_model,
+                         self.test_model_changed.__doc__)
+
+
+if __name__ == '__main__':
+    persist_model()
```
exarkun: The convention for running a test module is to run the tests in it.  Not sure this is a problem here.  The alternative, making a real script for this functionality, is certainly more work (and since we probably don't want to distribute/install the script alongside the other scripts ... I don't even really know how we would do it ...).  Just thought it was worth a comment.

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+
+
+class ConfigurationModelChanged(SynchronousTestCase):
+    """
+    Detect when the configuration model has changed.
+    """
+    def test_model_changed(self):
+        """
+        Most of the time if the configuration model changes this test will
+        fail.
```
exarkun: If you can change the model and this test doesn't fail, that's probably a bug we want to fix.  Maybe say something about that?

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+
+def persist_model():
+    """
+    Store the current model to disk.
+
+    We also store the git hash of current checkout, so it's clear what
+    version of code was used to generate the model.
+    """
+    git_hash = check_output(
+        [b"git", b"show", b'--format=%H', b"-s"]).strip()
```
exarkun: If there are changes to the working copy, they'll change the output of this function but they won't be reflected in this git hash.

Perhaps `flocker.__version__`?  I think it's strictly more informational.

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+from .._model import Deployment
+
+PERSISTED_MODEL = FilePath(__file__).sibling(b"persisted_model.json")
+
+# The class at the root of the configuration tree. This may need to be
+# changed if the configuration's root class changes.
+ROOT_CLASS = Deployment
+
+
+def generate_model(root_class=ROOT_CLASS):
```
exarkun: Maybe this belongs in `flocker.control._model`?  `flocker.control._persistence`?

flocker/control/test/test_model_change.py
exarkun: Some runs of the test suite with PYTHONHASHSEED=random suggests that if there are other cases where the ordering is non-deterministic, they're at least not covered by the current test suite.

flocker/control/test/test_model_change.py
```diff
@@ -0,0 +1,418 @@
...
+    We also store the git hash of current checkout, so it's clear what
+    version of code was used to generate the model.
+    """
+    git_hash = check_output(
+        [b"git", b"show", b'--format=%H', b"-s"]).strip()
+    model = generate_model()
+    PERSISTED_MODEL.setContent(dumps(
+        {u"git_hash": git_hash, u"model": model}, sort_keys=True, indent=4,
+        separators=(',', ': ')))
+    print("Persisted model for git hash {}.".format(git_hash))
```
exarkun: :question: Meh `print`, mucking up stdout.  Maybe this could at least be in the `__main__` block so one can call this function without dumping out this user feedback, if one wanted.

exarkun:
Thanks.  This looks good.  I think it will give us the protection I was looking for.  Please address inline comments and merge when tests are passing.

flocker/control/test/test_model_change.py
itamarst: I'm going to just add a bunch of sorts, and no tests, since writing non-determenistically failing tests for test code just seems... overly paranoid. If there *is* a false positive we'll notice and fix it, after all.

exarkun: This is why you like tox, isn't it?  Because it sets PYTHONHASHSEED=random every time?

A small number of simple tests should quickly show whether there's a problem or not.  That seems worth doing now rather than possibly making this someone else's problem in a few days.

itamarst: OK, will do.

PR #1844, opened 2015-08-13
[FLOC-2863] REST API endpoints can choose to ignore the request body

itamarst:


flocker/restapi/test/test_infrastructure.py
```diff
@@ -412,17 +418,20 @@ def test_decode(self, logger):
         render(app.app.resource(), request)
         self.assertEqual(objects, app.kwargs)
 
-    def assertNoDecodeLogged(self, logger, method):
+    def assertNoDecodeLogged(self, logger, method, path=b"/foo/bar"):
         """
         The I{JSON}-encoded request body is ignored when the given method is
```
exarkun: Hm.  How about a request body that's not JSON-encoded?  Is that important?  I suppose `null` and `12345` can both be loaded by the `json` module (but neither is actually technically legal).

flocker/restapi/test/test_infrastructure.py
```diff
@@ -445,6 +454,14 @@ def test_noDecodeDELETE(self, logger):
         """
         self.assertNoDecodeLogged(logger, b"DELETE")
 
+    @validateLogging(_assertRequestLogged(b"/foo/ignore_body", b"POST"))
+    def test_noBodyPOST(self, logger):
+        """
+        The I{JSON}-encoded request is body is ignored for methods with
+        {{ignore_body}} set to {{True}}.
```
exarkun: Your brain overloaded I guess.  Sorry.  `C{ignore_body}` and `C{True}` I think.

flocker/restapi/_infrastructure.py
```diff
@@ -165,6 +166,9 @@ def structured(inputSchema, outputSchema, schema_store=None):
     :param schema_store: A mapping between schema paths
         (e.g. ``b/v1/types.json``) and the JSON schema structure, allowing
         input/output schemas to just be references.
+    :param ignore_body: If true, ignore the contents of the body for all
```
exarkun: Maybe expand on what "ignore" means here.  The body is still read.  It just isn't decoded.  Maybe `decode_body=True` is closer to the intent?

exarkun:
Thanks.  Apart from minor inline comments, some tests have failed on AWS.  Address comments and get a green build and then merge.

flocker/restapi/test/test_infrastructure.py
itamarst: Next issue in queue is "don't require application/json", so can add a test there for another ignore_body variant.

PR #1845, opened 2015-08-14
[FLOC-2733] Versioned configuration

dwgebler:
Fixes FLOC-2733.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1845)


flocker/control/_model.py
```diff
@@ -628,6 +628,13 @@ class Deployment(PRecord):
 
     get_node = _get_node(Node)
 
+    @classmethod
+    def create_empty(cls):
+        """
+        Return a ``Deployment`` object representing an empty configuration.
+        """
+        return cls(nodes=frozenset())
```
exarkun: `Deployment.create_empty()` produces the same result as `Deployment()` so I don't think this method provides much value.

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+    :return bytes: The updated JSON configuration after migration.
+    """
+    upgraded_config = config
+    current_version = source_version
+    for upgrade_version in range(source_version + 1, target_version + 1):
+        with _LOG_UPGRADE(configuration=upgraded_config,
+                          source_version=current_version,
+                          target_version=upgrade_version):
+            migration_method = (
+                u"configuration_v%d_v%d"
```
exarkun: Since this driver only ever upgrades from N to N+1 the `v%d_v%d` mechanism - ie, upgrades between two arbitrary versions - seems like more than we need and is somewhat misleading (since `v1_v5` won't actually be used if it is defined).  Supporting that mechanism also introduces more complexity into the system (primarily once you have both `v1_v3` and `v1_v2` + `v2_v3` - even if you don't ever have them in the code base at the same time - since you can't be as sure how some state was upgraded anymore).  I'd suggest something like `upgrade_from_N` which is always used to convert state from `N` to `N + 1`.

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+                % (current_version, upgrade_version)
+            )
+            try:
+                migration = getattr(migration_class, migration_method)
+            except AttributeError:
+                message = (
```
exarkun: Might be easier/clearer to construct this string with `u"...".format(...)`

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+                    u"Unable to find a migration path for a version " +
+                    unicode(source_version) + u" to version " +
+                    unicode(target_version) + u" configuration. " +
+                    u"No migration method exists for v" +
+                    unicode(current_version) + u" to v" +
+                    unicode(upgrade_version) + u"."
+                )
+                raise ConfigurationMigrationError(message)
```
exarkun: I'd suggest that this should actually be an error much earlier in the process.  For example, it would be fairly useless to try to run code that could upgrade, say, 1 to 2 and 3 to 4, but not from 2 to 3.

Please file a follow-up for detecting situations like this earlier (for example, in the test suite).

exarkun: (And mark that follow-up as relating to the versioning work and please pull it onto the board for implementation soon)

flocker/control/_persistence.py
exarkun: Also, string formatting like this is better done in `ConfigurationMigrationError.__str__`.  Then you can construct the exception with useful objects (eg `ConfigurationMigrationError(source_version, target_version, ...)` (or perhaps better, something like `MissingMigrationPath(...)`)) and offer them as attributes while still presenting a human-readable error message

flocker/control/_persistence.py
```diff
@@ -38,7 +116,9 @@ def default(self, obj):
             result[_CLASS_MARKER] = obj.__class__.__name__
             return result
         elif isinstance(obj, PMap):
-            return {_CLASS_MARKER: u"PMap", u"values": dict(obj).items()}
+            return {
+                _CLASS_MARKER: u"PMap", u"values": dict(obj).items()
+            }
```
exarkun: Seems like this is only a formatting change?

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+
+        :param bytes config: The v1 JSON data.
+        :return bytes: The v2 JSON data.
+        """
+        v2_config = Configuration(
+            version=2, deployment=wire_decode(config))
+        return wire_encode(v2_config)
+
+
+class _Configuration_Encoder(JSONEncoder):
```
exarkun: Why the rename?

flocker/control/_persistence.py
```diff
@@ -91,6 +169,10 @@ def decode_object(dictionary):
                            [_DEPLOYMENT_FIELD])
 _LOG_SAVE = ActionType(u"flocker-control:persistence:save",
                        [_DEPLOYMENT_FIELD], [])
+_LOG_UPGRADE = ActionType(u"flocker-control:persistence:migrate_configuration",
+                          [_DEPLOYMENT_FIELD,
+                           Field(u"source_version", repr),
+                           Field(u"target_version", repr)], [])
```
exarkun: I think these should instead be something like `Field.for_types(u"source_version", [int], ...)`.  Also fields should always have documentation strings (3rd argument to `for_types`).

flocker/control/_persistence.py
```diff
@@ -163,15 +245,51 @@ def __init__(self, reactor, path):
...
+        """
+        Load the persisted configuration, upgrading the configuration format
+        if an older version is detected.
+        """
+        self._config_path = self._path.child(b"current_configuration.json")
+        v1_config_path = self._path.child(
+            b"current_configuration.v1.json")
+        # Check for a v1 config and upgrade to latest if found.
+        if v1_config_path.exists():
+            if not self._config_path.exists():
```
exarkun: It would probably be reasonable to either:

  * Move the v1 configuration out of the way as part of the upgrade process, ensuring it doesn't get picked up in the future
  * Reverse the logic here so that newer configuration files are always preferred over older ones

A third option, delete the old configuration, might also be reasonable but I'm happy to err on the side of keeping stuff around for longer than necessary if it helps us avoid even one accidental loss if information.

flocker/control/_persistence.py
```diff
@@ -163,15 +245,51 @@ def __init__(self, reactor, path):
...
+                    self._config_path.setContent(updated_json)
         if self._config_path.exists():
-            self._deployment = wire_decode(
-                self._config_path.getContent())
+            config_json = self._config_path.getContent()
+            config_dict = loads(config_json)
+            if 'version' in config_dict:
+                config_version = config_dict['version']
+            else:
+                config_version = 1
```
exarkun: This is an impossible case, right?  The v1 configuration is handled above.

flocker/control/_persistence.py
```diff
@@ -163,15 +245,51 @@ def __init__(self, reactor, path):
...
+        self.load_configuration()
+        MultiService.startService(self)
+        _LOG_STARTUP(configuration=self.get()).write(self.logger)
+
+    def load_configuration(self):
+        """
+        Load the persisted configuration, upgrading the configuration format
+        if an older version is detected.
+        """
+        self._config_path = self._path.child(b"current_configuration.json")
```
exarkun: This seems to be a constant.  It might be a good thing to make a property or return from a helper function or to compute at `__init__` time when `_path` is learned.

flocker/control/test/test_persistence.py
```diff
@@ -131,7 +139,53 @@ def test_file_is_created(self):
         """
         path = FilePath(self.mktemp())
         self.service(path)
-        self.assertTrue(path.child(b"current_configuration.v1.json").exists())
+        self.assertTrue(path.child(b"current_configuration.json").exists())
+
+    @validate_logging(assertHasAction, _LOG_UPGRADE, succeeded=True,
+                      startFields=dict(configuration=V1_TEST_DEPLOYMENT_JSON,
+                                       source_version=1, target_version=2))
```
exarkun: The `target_version` will change (and the test will break) when `_CONFIG_VERSION` changes.  Probably makes sense to put `_CONFIG_VERSION` into this assertion.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +303,88 @@ def test_complex_keys(self):
                                manifestations={}, paths={},
                                devices={uuid4(): FilePath(b"/tmp")})
         self.assertEqual(node_state, wire_decode(wire_encode(node_state)))
+
+
+class FakeMigration(object):
```
exarkun: This is more like a stub (https://jml.io/pages/test-doubles.html) - specifically, since the only configuration it can deal with is one with a fixed item, `{"configuration: "fake"}` - regardless of the input given.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +303,88 @@ def test_complex_keys(self):
...
+        if config['version'] != 2:
+            raise ConfigurationMigrationError(
+                "Supplied configuration was not a valid v2 config."
+            )
+        return json.dumps({"version": 3, "configuration": "fake"})
+
+
+class ConfigurationMigrationTests(SynchronousTestCase):
+    """
+    Tests for ``ConfigurationMigration`` and ``migrate_configuration``.
```
exarkun: This test suite only demonstrates that the version number gets bigger.  It doesn't demonstrate that the content or structure of the configuration can be changed.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +303,88 @@ def test_complex_keys(self):
...
+        upgraded_json = migrate_configuration(
+            1, _CONFIG_VERSION, V1_TEST_DEPLOYMENT_JSON,
+            ConfigurationMigration)
+        upgraded_config = wire_decode(upgraded_json)
+        expected_configuration = Configuration(
+            version=_CONFIG_VERSION, deployment=TEST_DEPLOYMENT
+        )
+        self.assertEqual(upgraded_config, expected_configuration)
+
+    def test_v1_v2_configuration(self):
```
exarkun: I think this test belongs somewhere else.  One sensible collection of tests is for `migrate_configuration`, independent of the specifics our actual configuration schema/types/values/objects.  Another sensible collection is for the upgrades defined by `ConfigurationMigration`.

exarkun: There should also be some additional tests here.  `V1_TEST_DEPLOYMENT_JSON` gets us through one particular case of the configuration data.  There are other cases that it won't exercise.  There are lots of possible variations.  I think this is probably somewhere Hypothesis would be really valuable.  For starters, it's probably worth exercising the 0, 1, >1 datasets and 0, 1, >1 applications cases.  (Hypothesis should let us exercise a ton more variations while probably writing less code, which would be pretty nice.)

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+
+def migrate_configuration(source_version, target_version,
+                          config, migration_class):
+    """
+    Migrate a persisted configuration from one version to another
+    in sequential upgrades, e.g. a source version of 1 and target
+    version of 3 will perform two upgrades, from version 1 to 2,
+    followed by 2 to 3.
+
+    Calls the correct ``ConfigurationMigration`` class methods for
```
exarkun: Or whatever `migration_class` is.

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+    the suppled source and target versions.
+
+    :param int source_version: The version to migrate from.
+    :param int target_version: The version to migrate to.
+    :param bytes config: The JSON-encoded source configuration.
```
exarkun: I think all of the JSON logic is inside the upgrade methods.  So really this is just whatever bytes were found in the file.

exarkun: Alternatively, we could suppose that we're going to stick with JSON (at least for a long while) and do the json encoding/decoding outside of the upgrade methods.  The upgrade methods then get to deal with decoded objects and don't have to all replicate the same json encode/decode steps.

exarkun: This is probably a sensible refactoring but it can wait for a follow-up branch.

flocker/control/_persistence.py
```diff
@@ -21,16 +21,94 @@
...
+    def configuration_v1_v2(cls, config):
+        """
+        Migrate a v1 JSON configuration to v2.
+
+        :param bytes config: The v1 JSON data.
+        :return bytes: The v2 JSON data.
+        """
+        v2_config = Configuration(
+            version=2, deployment=wire_decode(config))
+        return wire_encode(v2_config)
```
exarkun: Using `wire_decode` here will eventually lead to problems.  This use only works as long as the model objects in the current version of the code can be deserialized from the older version of the configuration.  This works for this particular upgrade for the time being but it's more or less guaranteed to break at some point.  Ideally, old upgrade code would be as little of a maintenance burden as possible.  So having old upgraders break when someone is developing a new feature isn't ideal.

We can be fairly confident that the `json` module isn't going to undergo any drastic changes that render it incapable of deserializing the configuration to some mid-level Python objects, though.  Those mid-level Python objects are probably the right level at which upgrades can operate.  That means this upgrade method should look something like:

```py
v1_config = json.loads(config)
v2_config = {"$__class__$": ..., "deployment": v1_config}
return json.dumps(v2_config)
```

The details of the most robust/reliable way to construct the `v2_config` wrapper/skeleton might vary a bit - but in this case it's so simple there's not really much choice, I think.

exarkun:
Thanks.

Please address and re-submit.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +332,193 @@ def test_complex_keys(self):
...
+                maximum_size=random.randint(1000000, 100000000)
+            )
+        )
+        manifestations[dataset_id] = manifestation
+    return manifestations
+
+
+ST_MANIFESTATIONS = st.basic(generate=manifestation_generator)
+UUIDS = st.basic(generate=lambda r, _: UUID(int=r.getrandbits(128)))
+ALPHABET = u"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijlmnopqrstuvwxyz"
```
jml: Python already defines this as `string.letters`

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +332,193 @@ def test_complex_keys(self):
...
+    alias=SIMPLE_TEXT
+)
+APPLICATIONS = st.builds(
+    Application, name=SIMPLE_TEXT, image=IMAGES,
+    # XXX at the moment there is a problem including AttachedVolume
+    # instances in each Application, because the Manifestation in
+    # a volume must also be present on the manifestations in a Node.
+    # So far I cannot find a way to do this with Hypothesis.
+    volume=st.just(None),
+    ports=st.sets(PORTS, min_size=0, max_size=3),
```
jml: `min_size=0` is a bit redundant (here & elsewhere).

jml: Where does `max_size=3` come from? Why 3? I'd suggest removing it, or at the least adding a comment explaining why 3.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +332,193 @@ def test_complex_keys(self):
...
+)
+PORTS = st.builds(
+    Port,
+    internal_port=st.integers(min_value=1, max_value=32000),
+    external_port=st.integers(min_value=1, max_value=32000)
+)
+LINKS = st.builds(
+    Link,
+    local_port=st.integers(min_value=1, max_value=32000),
+    remote_port=st.integers(min_value=1, max_value=32000),
```
jml: 1. Would be cleaner to define a strategy for port numbers and re-use it here
2. 32000 is an unusual number to choose. Why did you pick it? [Registered ports](https://en.wikipedia.org/wiki/Registered_port)

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +332,193 @@ def test_complex_keys(self):
...
+    return manifestations
+
+
+ST_MANIFESTATIONS = st.basic(generate=manifestation_generator)
+UUIDS = st.basic(generate=lambda r, _: UUID(int=r.getrandbits(128)))
+ALPHABET = u"ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijlmnopqrstuvwxyz"
+SIMPLE_TEXT = st.text(
+    alphabet=ALPHABET, min_size=4, max_size=10, average_size=6
+)
+IMAGES = st.builds(DockerImage, tag=SIMPLE_TEXT, repository=SIMPLE_TEXT)
```
jml: Why restrict these values to simple text?

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +332,193 @@ def test_complex_keys(self):
...
+NONE_OR_INT = st.one_of(
+    st.just(None),
+    st.integers(min_value=1000000, max_value=100000000)
```
jml: Where do these integer values come from?

flocker/control/test/test_persistence.py
dwgebler: If I remove it, how many combinations of Ports is Hypothesis going to try by default? We want a reasonable range, but in practice an Application is never going to realistically expose 1000 ports, around 0-3 seems more typical.

flocker/control/test/test_persistence.py
dwgebler: They're just a range of possible values for things like memory limit, representing in this case 1MB to 100MB (ish). They're fairly arbitrary choices, intended only to ensure we test a reasonable range of values.

flocker/control/test/test_persistence.py
dwgebler: I didn't think there was a need to test anything beyond ordinary alphabet characters. They don't technically have to be `string.letters`, the objects can be constructed with any unicode, I've just never encountered a Docker image name that uses a snowman symbol in its name.

flocker/control/test/test_persistence.py
dwgebler: We're primarily testing manipulation of JSON-serialized and deserialized objects here, so the choice of 32000 is arbitrary, just in a range that are valid port numbers. Seeing as we're not testing any actual containers or use of these port numbers, any particular number would seem arbitrary.

flocker/control/test/test_persistence.py
jml: We discussed this a little on Zulip. Part of the point of using hypothesis is that it tests cases we would never have thought of. We should only limit the inputs if there is an explicit reason (e.g. it makes it faster; we deliberately don't handle that case; etc.). When we do add such limits, we should have a comment explaining why.

flocker/control/test/test_persistence.py
jml: Port numbers can go from 1 to 2**16-1, inclusive. Let's use that range in our tests.

flocker/control/test/test_persistence.py
jml: There's no particular need to test beyond there, but there's equally no reason to restrict our tests to just those. Why turn down cheap extra test coverage?

As with other strategies, please also remove the limits.

flocker/control/test/test_persistence.py
jml: Again, if they are arbitrary, don't provide them. Trust hypothesis to get it right until we've been given a reason to believe otherwise.

dwgebler:
Seems Buildbot has trouble handling the Hypothesis pressure, http://build.clusterhq.com/builders/flocker%2Funit-test%2Fcentos-7%2Fzfs-head/builds/698/steps/trial/logs/stdio

We may need to further restrict the range of examples min and max values.

jml:
Or we may need to try a different approach. The [latest release of hypothesis](http://hypothesis.readthedocs.org/en/latest/changes.html#id2) has a bugfix that would make a filtering-based approach for node work much better.

dev-requirements.txt
```diff
@@ -10,6 +10,7 @@ Fabric==1.10.2
 flake8==2.4.1
 gitdb==0.6.4
 GitPython==1.0.1
+hypothesis==1.10.1
```
jml: Please change to 1.10.3

flocker/control/_persistence.py
```diff
@@ -160,21 +266,65 @@ def __init__(self, reactor, path):
...
+            with _LOG_UPGRADE(self.logger,
+                              configuration=v1_json,
+                              source_version=1,
+                              target_version=_CONFIG_VERSION):
+                updated_json = migrate_configuration(
+                    1, _CONFIG_VERSION, v1_json,
+                    ConfigurationMigration
+                )
+                self._config_path.setContent(updated_json)
+                v1_config_path.moveTo(v1_archived_path)
```
jml: There doesn't seem to be much that's actually specific to v1 about this code, apart from the paths and the variable names. 

Please make these parameters and pass them in from `load_configuration`

flocker/control/_persistence.py
```diff
@@ -160,21 +266,65 @@ def __init__(self, reactor, path):
...
+            config_version = config_dict['version']
+            if config_version < _CONFIG_VERSION:
+                with _LOG_UPGRADE(self.logger,
+                                  configuration=config_json,
+                                  source_version=config_version,
+                                  target_version=_CONFIG_VERSION):
+                    config_json = migrate_configuration(
+                        config_version, _CONFIG_VERSION,
+                        config_json, ConfigurationMigration)
+            config = wire_decode(config_json)
```
jml: It took me a long while to figure out how and why this code was different from `_process_v1_config`. I *guess* it's because in v1 we stored the version number in the filename rather than in the config itself, hence the need for a special case. 

Please add a comment explaining why we special-case v1. It'll probably save future readers some time.

flocker/control/_persistence.py
```diff
@@ -160,21 +266,65 @@ def __init__(self, reactor, path):
...
+            self._deployment = config.deployment
+            self._sync_save(config.deployment)
```
jml: :question: It might be possible to refactor this so it shares more code with _process_v1_config. I don't think it's strictly necessary (see [three strikes and you refactor](http://c2.com/cgi/wiki?ThreeStrikesAndYouRefactor)), but might be worth looking at.

flocker/control/test/test_persistence.py
```diff
@@ -3,31 +3,40 @@
...
-DATASET = Dataset(dataset_id=unicode(uuid4()),
+DATASET = Dataset(dataset_id=u'4e7e3241-0ec3-4df6-9e7c-3f7e75e08855',
                   metadata={u"name": u"myapp"})
 MANIFESTATION = Manifestation(dataset=DATASET, primary=True)
 TEST_DEPLOYMENT = Deployment(
-    nodes=[Node(uuid=uuid4(),
+    nodes=[Node(uuid=UUID(u'ab294ce4-a6c3-40cb-a0a2-484a1f09521c'),
                 applications=[
                     Application(
                         name=u'myapp',
```
jml: I'm guessing that the reason behind this change is so that these variables match the sample data files added in the PR?

If so, please add a comment explaining that these values were not chosen at random, but that they must match those files.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+UUIDS = st.basic(generate=lambda r, _: UUID(int=r.getrandbits(128)))
+
+DATASETS = st.builds(Dataset, dataset_id=UUIDS, maximum_size=st.integers())
+
+# Constrain primary to be True so that we don't get invariant errors from Node
+# due to having two differing manifestations of the same dataset id.
+MANIFESTATIONS = st.builds(
+    Manifestation, primary=st.just(True), dataset=DATASETS)
+SIMPLE_TEXT = st.text(
+    alphabet=string.letters, min_size=4, max_size=20, average_size=12
```
jml: Please delete the `min_size` and `max_size` or add a comment explaining why they are necessary.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+)
+IMAGES = st.builds(DockerImage, tag=SIMPLE_TEXT, repository=SIMPLE_TEXT)
```
jml: Please change this to use `st.text()`. We should only constrain the test inputs if there's a concrete & demonstrated need. When we do so, we should document that need.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+NONE_OR_INT = st.one_of(
+    st.just(None),
+    st.integers()
+)
```
jml: Please use `st.none()` instead of `st.just(None)` (this was missing from the docs due to a bug that's since been fixed).

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+LINKS = st.builds(
+    Link,
+    local_port=ST_PORTS,
+    remote_port=ST_PORTS,
+    alias=SIMPLE_TEXT
+)
+APPLICATIONS = st.builds(
+    Application, name=SIMPLE_TEXT, image=IMAGES,
+    ports=st.sets(PORTS, max_size=10),
+    links=st.sets(LINKS, max_size=10),
```
jml: Please remove the `max_size` constraint.

flocker/control/test/test_persistence.py
jml: :question: FWIW, you can also write this as `st.none() | st.integers()`. I think either works.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+    environment=st.dictionaries(keys=SIMPLE_TEXT, values=SIMPLE_TEXT),
+    memory_limit=NONE_OR_INT,
+    cpu_shares=NONE_OR_INT,
+    running=st.booleans()
+)
+FILEPATHS = st.text(alphabet=string.printable).map(FilePath)
+VOLUMES = st.builds(
+    AttachedVolume, manifestation=MANIFESTATIONS, mountpoint=FILEPATHS)
+APPLICATIONS_WITH_VOLUMES = st.tuples(
+    APPLICATIONS, VOLUMES).map(lambda (a, v): a.set(volume=v))
```
jml: There's no need for a separate `APPLICATIONS_WITH_VOLUMES`. Instead, the `APPLICATIONS` strategy should set `volume` to `st.none() | VOLUMES`. That way, we'll test all cases.

This will require some changes further down, as `_build_node` assumes that `app.volume` is not `None`.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+    return st.builds(
+        Node, uuid=UUIDS,
+        applications=st.just(applications),
+        manifestations=manifestations)
+
+
+NODES = st.sets(APPLICATIONS_WITH_VOLUMES).map(
+    lambda apps: pset(dict((
+        app.volume.manifestation.dataset_id, app) for app in apps).values())
+).flatmap(_build_node)
```
jml: ISTR you saying this was failing due to timeouts. I think hypothesis 1.10.3 fixes that problem. If it doesn't, we're going to have to figure out a different strategy.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+    """
+    Tests for ``ConfigurationMigration`` class that performs individual
+    configuration upgrades.
+
+    There should be
+    """
+    @given(DEPLOYMENTS)
+    def test_upgrade_configuration_v1_latest(self, deployment):
+        """
+        Test a range of generated configurations (deployments) can be
```
jml: No need for `Test`. https://jml.io/test-docstrings/

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+        upgraded_config = wire_decode(upgraded_json)
+        expected_configuration = Configuration(
+            version=_CONFIG_VERSION, deployment=deployment
+        )
+        self.assertEqual(upgraded_config, expected_configuration)
+
+    @given(st.tuples(
+        st.integers(min_value=1, max_value=2),
+        st.integers(min_value=2, max_value=2)
+    ))
```
jml: Since `versions` is never used as a tuple, this is better specified as `@given(integers(...), integers(...))`, and then writing the test as `test_upgrade_configuration_versions(self, from_version, to_version)`.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+    def test_upgrade_configuration_versions(self, versions):
+        """
+        Test a range of version upgrades by ensuring the configuration
```
jml: No need for `Test`. https://jml.io/test-docstrings/

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +333,202 @@ def test_complex_keys(self):
...
+
+        See flocker/control/test/configurations for individual
+        version JSON files and generation code.
+
+        Increment the given integer values in the decorator above to define
+        the test range when new configuration upgraders are added. The first
+        range is the "upgrade from" versions range, where min_value should
+        always be 1. The second range is the "upgrade to" versions range,
+        where min_value should be 2 and max_value should be the latest
+        supported configuration version number.
```
jml: I very much appreciate that you've explained where the numbers come from. Thank you.

If I understand correctly, we already have a variable in code that says what the latest supported config version is: `_CONFIG_VERSION`. If that's right, we should use that, rather than a literal 2.

Second, I don't think we support downgrades. This test only passes because we have two versions. The real constraint is **not** that the "upgrade to" value should be more than 2, it's that it should be more than "upgrade from".

As such, a better approach would be to do something like this:

```python
SUPPORTED_VERSIONS = integers(1, _CONFIG_VERSION)

@given(tuples(SUPPORTED_VERSIONS, SUPPORTED_VERSIONS).map(lambda x: tuple(sorted(x)))
def test_u_c_v(self, (upgrade_from, upgrade_to)):
    ....
```

flocker/control/test/test_persistence.py
jml: I've added a comment below that invalidates this one.

flocker/control/test/test_persistence.py
jml: Also, it's not clear why we are testing migration from 2 to 2 but not 1 to 1. Please update the docstring to explain why.

flocker/control/_persistence.py
```diff
@@ -30,10 +30,110 @@
...
+    migrations_sequence = []
+    for upgrade_version in range(source_version + 1, target_version + 1):
+        with _LOG_UPGRADE(configuration=upgraded_config,
+                          source_version=current_version,
+                          target_version=upgrade_version):
+            migration_method = u"upgrade_from_v%d" % current_version
+            try:
+                migration = getattr(migration_class, migration_method)
+                migrations_sequence.append(migration)
+                current_version = current_version + 1
```
jml: Please change to `current_version += 1`

flocker/control/_persistence.py
```diff
@@ -30,10 +30,110 @@
...
+            except AttributeError:
```
jml: As a general rule, you want to make `try` clauses as small as possible in order to reduce the likelihood of accidentally catching something that you didn't mean to, because that can lead to masking serious errors and making code much harder to debug.

This is **especially** true for `AttributeError`, which is often raised due to programming mistakes (e.g. `self.dataRecieved(some_data)`).

The right thing to do here is either:
  1. Restrict the `try` block to just the line with the `getattr`
  2. Don't use `try` at all. Instead `getattr(migration_class, migration_method, None)` and then raise the exception if `migration` is `None`.

jml:
Thanks David. It's exciting to see hypothesis make its way into the flocker code base. Please address these comments & resubmit.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +338,199 @@ def test_complex_keys(self):
...
+    def test_upgrade_configuration_versions(self, versions):
+        """
+        A range of versions can be upgraded and the configuration
+        blob after upgrade will matche that which is expected for the
+        particular version.
+
+        See flocker/control/test/configurations for individual
+        version JSON files and generation code.
+        """
+        configs_dir = FilePath(__file__).sibling('configurations')
```
jml: If you add a line:

```python
old_version, new_version = versions
```

And use those variables instead of `versions[0]` and `versions[1]`, the code will become magically more self-documenting.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +338,199 @@ def test_complex_keys(self):
...
+
+NODES = st.sets(APPLICATIONS).map(
+    lambda apps: pset(dict((
+        app.volume.manifestation.dataset_id, app)
+        for app in apps if app.volume).values())
+).flatmap(_build_node)
+
+
+DEPLOYMENTS = st.builds(
+    Deployment, nodes=st.sets(NODES, min_size=0, max_size=3)
```
jml: 1. You never need `min_size=0` on a set. No reasonable reader is going to assume the default is anything but. Please drop it.
2. Please either delete the maximum size constraint, or give a concrete reason in the comments for constraining `max_size` to 3, explaining why 3 was chosen.

(I'm not 100% sure, but think I made these comment in a previous pass)

The more general point is that code should communicate intent. No one is going to be able to look at `max_size=3` and guess why we chose 3, or what we were trying achieve by doing so.

flocker/control/test/test_persistence.py
```diff
@@ -249,3 +338,199 @@ def test_complex_keys(self):
...
+    local_port=ST_PORTS,
+    remote_port=ST_PORTS,
+    alias=st.text(alphabet=string.letters, min_size=1)
+)
+FILEPATHS = st.text(alphabet=string.printable).map(FilePath)
+VOLUMES = st.builds(
+    AttachedVolume, manifestation=MANIFESTATIONS, mountpoint=FILEPATHS)
+APPLICATIONS = st.builds(
+    Application, name=st.text(), image=IMAGES,
+    # A MemoryError will likely occur without the max_size limits on
```
jml: Just checking, did you observe a `MemoryError` without these set?

dwgebler: I did.

jml:
Thanks for these changes David. I've only got a couple of minor tweaks that I've suggested, but I've also got one fairly open ended question, so I'd appreciate it if you could resubmit when you've answered / made the changes.  Feel free to ping me on Zulip to ensure prompt review.

flocker/control/test/test_persistence.py
jml: Ah, interesting. I don't have any insights off the top of my head, and it's been a while since I've debugged Python memory usage errors. (https://pypi.python.org/pypi/memory_profiler seems a good place to start though). 

It might be interesting to try to make a minimal example that reproduces the memory crash, and submit it as a bug against Hypothesis. But, not necessary, and probably not a good use of ClusterHQ time.

jml:
Looks good, thanks... apart from the failing test. I've asked on `#hypothesis` and tried a couple of things to make it faster but no joy yet. Will do some profiling. Feel free to investigate in parallel, or we can pair on it on Monday.

jml:
I've done a bit of exploring and it seems the bottleneck seems to be the unboundedness of the size of `nodes`.

I chased this up on `#hypothesis`:

> [18:38:16]  <jml>	so one piece of code review advice I've been giving is to not constrain our arbitrary data generators without an explicit reason
> [18:39:07]  <jml>	because a large part of the value of something like hypothesis is that it checks cases that haven't thought about
> [18:39:31]  <jml>	but it's looking like size on containers is probably always going to need some kind of constraint?
> [18:41:13]  <jml>	DRMacIver: does that seem right?
> [18:50:27]  <DRMacIver>	Because you can assign the defaults
> [18:50:33]  <DRMacIver>	Historically that didn't.
> [18:50:54]  <DRMacIver>	jml: So ultimately your data shape is constrained by what's feasible to generate. If you don't want to hard constraint it you can use size hints.
> [18:51:10]  <DRMacIver>	jml: It *should* be the case that Hypothesis can figure out what's reasonable automatically, but right now it just flat out isn't.
> [18:51:33]  <DRMacIver>	So I would recommend using average_size but not necessarily using max_size
> [18:51:53]  <DRMacIver>	(average_size is a bit of a misnomer due to possible other constraints. It's more of a size hint)

I suggest making the following changes to `NODES` and `DEPLOYMENTS` to fix the test.

```python
NODES = st.lists(
    APPLICATIONS,
    # If we add this hint on the number of applications, Hypothesis is able to
    # run many more tests.
    average_size=10,
    unique_by=lambda app:
    app if not app.volume else app.volume.manifestation.dataset_id).map(
        pset).flatmap(_build_node)


DEPLOYMENTS = st.builds(
    # If we leave the number of nodes unbounded, Hypothesis will take too long
    # to build examples, causing intermittent timeouts. Making it roughly 3
    # should give us adequate test coverage.
    Deployment, nodes=st.sets(NODES, average_size=3)
)

```

PR #1847, opened 2015-08-14
[FLOC-2417] Operate with SELinux enabled

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2417

This removes the acceptance testing provisioning step which disables SELinux on CentOS 7.  It changes the way we use Docker to enable a volume bound to a container to be usable by that container.


flocker/node/_docker.py
```diff
@@ -516,13 +516,12 @@ def add(self, unit_name, image_name, ports=None, environment=None,
         restart_policy_dict = self._serialize_restart_policy(restart_policy)
 
         def _create():
-            binds = {
-                volume.node_path.path: {
-                    'bind': volume.container_path.path,
-                    'ro': False,
-                }
+            binds = list(
+                u"{}:{}:Z".format(
```
itamarst: Apparently this is in "the other docs"? Add a link please.

docs/releasenotes/index.rst
```diff
@@ -16,6 +16,8 @@ v1.2
 * Flocker ``.deb`` and ``.rpm`` packages no longer declare any dependency on a Docker package.
   Docker is required for the container management functionality but a Docker package must be selected and installed manually.
   This provides more control over the version of Docker used with Flocker.
+* Flocker's container management functionality now integrates with SELinux.
```
itamarst: `NEWS` entry too I guess? Or maybe we shoudl just give up on `NEWS`.

exarkun: I'm not clear on why there are two copies of all of this stuff.

itamarst:
Beyond a couple of minor comments, the one worry I have is that this breaks Vagrant somehow, since we're not testing it. All the tests failed on that builder... I reran it.

So: address minor comments, merge if buildbot is green and there is some minimal evidence we haven't completely broken Vagrant (e.g. a single passing container+dataset acceptance test on the builder).

flocker/node/_docker.py
```diff
@@ -516,13 +516,17 @@ def add(self, unit_name, image_name, ports=None, environment=None,
...
-                    'bind': volume.container_path.path,
-                    'ro': False,
-                }
+            binds = list(
+                # The "Z" mode tells Docker to "relabel file objects" on the
+                # volume.  This makes things work when SELinux is enabled, at
+                # least in the default configuration on CentOS 7.  See
+                # <https://docs.docker.com/reference/commandline/run/>, in the
+                # `--volumes-from` section (or just search for SELinux).
+                u"{}:{}:Z".format(
```
itamarst: Maybe this change is breaking Ubuntu? The test that times out is using a volume.

itamarst: In particular the test checks that a file was written to a volume *while the container is still running*. So if that flag has some impact on visibility in Ubuntu that might make it wait forever.

itamarst: Also this is suggestive:

flocker.node.functional.test_docker.GenericDockerClientTests.test_add_with_volumes ... [FAILURE]

PR #1849, opened 2015-08-14
[FLOC-2865] Support some edge cases that previously broke the REST API client

itamarst:


flocker/apiclient/_client.py
```diff
@@ -25,14 +25,14 @@ class Dataset(PClass):
     A dataset in the configuration.
 
     :attr UUID primary: The node where the dataset should manifest.
-    :attr int maximum_size: Size of new dataset, in bytes.
+    :attr maximum_size: Size of new dataset in bytes (as ``int``) or ``None``.
```
jongiddy: Remove the word "new". There doesn't seem to be any constraint that this structure can only be used for a new dataset. It is used in the `move` and `list` methods later on.

flocker/apiclient/_client.py
```diff
@@ -25,14 +25,14 @@ class Dataset(PClass):
...
     :attr UUID dataset_id: The UUID of the dataset.
     :attr metadata: A mapping between unicode keys and values.
     :attr bool deleted: If true indicates this dataset should be deleted.
     """
     dataset_id = field(type=UUID, mandatory=True)
     primary = field(type=UUID, mandatory=True)
-    maximum_size = field(type=int, mandatory=True)
+    maximum_size = field(type=(int, None.__class__), mandatory=True)
```
jongiddy: :question: Use `type(None)` instead of `None.__class__` here and in other places below.

flocker/apiclient/_client.py
```diff
@@ -41,16 +41,18 @@ class DatasetState(PClass):
     """
     The state of a dataset in the cluster.
 
-    :attr UUID primary: The node where the dataset should manifest.
-    :attr int maximum_size: Size of new dataset, in bytes.
+    :attr primary: The ``UUID`` of the node where the dataset is manifest,
+        or ``None`` if has no manifestation.
```
jongiddy: "...if the dataset has no primary manifestation."

flocker/apiclient/_client.py
```diff
@@ -41,16 +41,18 @@ class DatasetState(PClass):
...
+    :attr maximum_size: Size of new dataset in bytes, or ``None`` if it
+        has none.
```
jongiddy: Again, remove the "new".

This definition implies that `maximum_size` really means actual size, in which case, what does it mean for a dataset state to have no actual size (`maximum_size is None`)?  Alternatively, if `maximum_size` is a limit and it is guaranteed that the actual size <= `maximum_size`, then `None` makes sense, but `maximum_size` is not the "size of the dataset".  Please clarify the definition.

flocker/apiclient/_client.py
jongiddy: In `DatasetState`, the code has `:attr FilePath|None path:`. This "or-typing" could be applied to these changed definitions, or that definition could be changed to match this "prose-typed" definition, but we should be consistent within a module, at least.

jongiddy:
Looks good to me. My comments are all to do with the docstrings, so PR can be merged once these are addressed.

flocker/apiclient/_client.py
itamarst: I felt like `type=(UUID, type(None))` was hard to read. Maybe I'll just use a module constant.

PR #1850, opened 2015-08-17
[FLOC-2864] Flocker API client delete datasets

dwgebler:
Fixes FLOC-2864

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -94,13 +94,14 @@ class IFlockerAPIV1Client(Interface):
 
         :param dataset_id: The UUID of the dataset to be deleted.
 
-        :return: ``Deferred`` that fires after the configuration has been
-            updated with the ``Dataset`` that has just been deleted.
+        :return: ``Deferred`` that fires with the ``Dataset`` that has just
+        been deleted, after the configuration has been updated.
         """
 
     def list_datasets_configuration():
         """
-        Return the configured datasets.
+        Return the configured datasets, excluding any datasets that
+        have been deleted.
 
         :return: ``Deferred`` firing with iterable of ``Dataset``.
         """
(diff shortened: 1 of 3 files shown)


flocker/apiclient/_client.py
```diff
@@ -90,6 +88,16 @@ def move_dataset(primary, dataset_id):
             updated with the resulting ``Dataset``.
         """
 
+    def delete_dataset(dataset_id):
+        """
+        Delete a dataset.
+
+        :param dataset_id: The UUID of the dataset to be deleted.
+
+        :return: ``Deferred`` that fires after the configuration has been
```
itamarst: This should note that it fires with the deleted ``Dataset`` object.

flocker/apiclient/test/test_client.py
```diff
@@ -144,6 +144,39 @@ def got_result(dataset):
...
+                                    dataset_id=dataset_id)
+            d.addCallback(
+                lambda _: self.client.delete_dataset(dataset_id))
+
+            def got_result(dataset):
+                expected_removed = Dataset(
+                    dataset_id=dataset_id, primary=self.node_1,
+                    maximum_size=DATASET_SIZE
+                )
+                self.assertEqual(expected_removed, dataset)
```
itamarst: :question: Unless you're using a single high-level assertion method that internally does multiple assertions, it's better to have only one assertion in each unit test.

itamarst: I.e. you'd have one test for return result, one test for impact on listing.

flocker/apiclient/_client.py
```diff
@@ -90,6 +88,16 @@ def move_dataset(primary, dataset_id):
...
+            updated with the ``Dataset`` that has just been deleted.
+        """
+
     def list_datasets_configuration():
         """
         Return the configured datasets.
```
itamarst: Document that deleted datasets are not listed.

itamarst:
Looks good; address and merge when buildbot is happy.

itamarst:
Needs to be merged forward apparently.

itamarst:
OK merged forward.

PR #1859, opened 2015-08-18
[FLOC-2880] Switch to using the real `requests.Response`

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2880

Replaces the test double with tested use of the real type.  Still not what I'd call ideal but there's not much we can easily do about the fact that the way `Response` is made to be used is error prone.


itamarst:
Looks good, thanks.

PR #1860, opened 2015-08-19
[FLOC 2481] Update CONTRIBUTING.rst for consistency

davidoliverSP2:
Fixes 2481

In addition, this PR removes ref tags from the CONTRIBUTING.rst document, which don't work in GitHub, and replaces them with links to the latest pages on the ClusterHQ website.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -131,10 +131,12 @@ You can view the result by opening ``docs/_build/html/index.html`` in your brows
 Contributing to Flocker
 =======================
 
+If you have any feature requests or suggestions, we would love to hear about them.
+
 At a minimum you can simply submit a GitHub Pull Request with your changes.
 In order to maximize your chances of getting your code accepted, and to keep you from wasting time:
 
-* Discuss your ideas with us in advance in a `JIRA`_ or GitHub issue.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -143,6 +147,8 @@ Once your pull request is merged, as a small thank you for contributing to Flock
 Just send an email to thankyou@clusterhq.com with your t-shirt size, mailing address and a phone number to be used only for filling out the shipping form.
 We'll get something in the mail to you.
 
+.. _UserVoice: https://feedback.clusterhq.com/
+.. _GitHub issue: https://github.com/clusterhq/flocker/issues
```
adamtheturtle: This link is used twice so it can be at the bottom of the document only once.

adamtheturtle:
Thanks @davidoliverSP2. I prefer the change proposed in the commit mentioned in the JIRA issue - https://github.com/ClusterHQ/flocker/commit/5aa1361bf81f4ff820b86f082327cdd05c3a1f5e. This allows us to have one place only where we can update the ways of talking to us.

Are the new links checked automatically by BuildBot? If not, please make an issue to do this. If not, is it better to risk broken links than to have links not work in GitHub's view?

I'll leave those judgements to you. If you make any significant changes, resubmit, else merge when you are happy and the docs builder is happy (no need to wait for the other builders).

PR #1861, opened 2015-08-19
fix typo in build.yaml \ -> /

Azulinho:
fixes typo in build.yaml

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -188,7 +188,7 @@ common_cli:
 
   build_sdist: &build_sdist  |
     # package the goodies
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks @Azulinho . I'll accept this on the basis that we'll figure out if this command is incorrect by the time we start using Jenkins, and I assume it is tested. As a reviewer I'd like to know not just that this is a typo being fixed but instead what functionality is not working, and is desired to work after this in the issue. 

I guess in this case that would be something like "Creating a source distribution fails because it uses the wrong path to python".

Also, the convention is to use hyphens instead of underscores in branch names.

Issue #1851, opened 2015-08-17
Install flocker encounter "clusterhq-flocker-node : Depends: docker.io (>= 1.3.0) but it is not going to be installed"


coolljt0725:
Hi all, when I install the flocker on ubuntu 14.04 follow the steps from https://docs.clusterhq.com/en/1.1.0/install/install-node.html 
I encountered the error
<pre><code>Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 clusterhq-flocker-node : Depends: docker.io (>= 1.3.0) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.</code></pre>
and I have installed docker version 1.7.1. 
Anyone can help?

lukemarsden:
Hi there, thanks for trying Flocker!

We are aware of this issue and it has already been fixed in master. I believe that our 1.1.1 release will contain the fix, and I believe that release will be done (hopefully) later this week, so please watch out for that and then try again.

Meanwhile, there are some workarounds documented in the comments on [this issue](https://clusterhq.atlassian.net/browse/FLOC-2447).

Out of interest, what are you thinking of using Flocker for? What kind of applications/databases, how big are they, and on what infrastructure/cloud provider?

Thanks!
Luke

lukemarsden:
By the way @coolljt0725, you can sign up for email updates on Flocker releases by signing up for email updates at in the footer of our website at https://clusterhq.com/

coolljt0725:
@lukemarsden  Thanks for your help
>what are you thinking of using Flocker for? What kind of applications/databases, how big are they, and on what infrastructure/cloud provider?
I'm interested in Flocker. I just trying the the flocker and see what it can do and how it works.

lukemarsden:
You can now install Flocker alongside any version of Docker, @coolljt0725.

PR #1863, opened 2015-08-20
Fix Running MRO error:

moypray:
  File "/usr/local/bin/flocker-dataset-agent", line 7, in <module>
    from flocker.node.script import flocker_dataset_agent_main
  File "/usr/local/lib/python2.7/dist-packages/flocker/node/__init__.py", line 17, in <module>
    from .script import BackendDescription, DeployerType
  File "/usr/local/lib/python2.7/dist-packages/flocker/node/script.py", line 37, in <module>
    from ._loop import AgentLoopService
  File "/usr/local/lib/python2.7/dist-packages/flocker/node/_loop.py", line 395, in <module>
    class AgentLoopService(object, MultiService):
TypeError: Error when calling the metaclass bases
    Cannot create a consistent method resolution
order (MRO) for bases object, MultiService

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -8,3 +8,4 @@ The following people and organizations contributed to its development; please ad
 
 * ClusterHQ (formerly Hybrid Logic Ltd.)
 * Scotch Media (Base for documentation theme)
+* Wentao Zhang (zhangwentao234@huaewi.com, Huawei Ltd.)
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -392,7 +392,7 @@ def build_convergence_loop_fsm(reactor, deployer):
 
 @implementer(IConvergenceAgent)
 @attributes(["reactor", "deployer", "host", "port"])
-class AgentLoopService(object, MultiService):
+class AgentLoopService(MultiService, object):
     """
     Service in charge of running the convergence loop.
 


exarkun:
Thanks.

This is caused by an incompatible change made in a recent release of Twisted.  We test Flocker against a specific version of Twisted which doesn't contain this problem.  If you use any of the OS packages we distribute, you won't run into this problem because we include versions of our dependencies in those packages which we test against.  If you install Flocker using anything that respects Python's requirements metadata (either in setup.py or in our requirements.txt file) then you should also get the same version of Twisted that we test Flocker against.  So this should also avoid this (and similar) problems.

You can find detailed instructions about the recommended way to install Flocker from source (ie, not using the native OS packages) at <https://docs.clusterhq.com/en/1.1.0/install/install-client.html#other-linux-distributions>.  The instructions more or less come down to "install a couple native OS dependencies, make a virtualenv, use pip to install Flocker into it".

Since it looks like Twisted is going to stick with this incompatible change, Flocker will eventually need to change to be compatible with it.  The change in this PR looks like it's the right one for that.  We might hold off on merging this until we have testing in place on our CI system, though (letting us avoid introducing regressions into Flocker compatibility with the newer version of Twisted).

Thanks again.

moypray:
Hi   @exarkun .  Thanks for your reply and explanation.

I am using Ubuntu 14.04, but it failed to install flocker following the way of in  https://docs.clusterhq.com/en/1.1.0/install/install-client.html

So I install it according to "other linux distribution", but I got lots of package dependency errors.
Finally I dropped using virtualenv, installing all the packages (pip install) which required by flocker, and this issue come up, but after this fix, it works, and I success to bring flocker up and could create container using flocker-deploy.

my python is: 2.7.6.
root@ubuntu:~# python --version
Python 2.7.6
root@ubuntu:~#

I am not sure why do not reproduce in your environment.

exarkun:
Thanks for the follow-up moypray.  I'm curious about the errors you encountered when following <https://docs.clusterhq.com/en/1.1.0/install/install-client.html>.  I'd really like those instructions to work. :smile:

moypray:
thanks,  I met the error is  the same with https://github.com/ClusterHQ/flocker/issues/1851

jongiddy:
Can you please provide some more context about what you were doing when the problems occurred?

I'd like to clarify this, because the error in #1851 may occur when installing the package `clusterhq-flocker-node`, but the instructions at https://docs.clusterhq.com/en/1.1.0/install/install-client.html don't mention that package.

Did this problem occur while you were installing the Flocker client or installing a Flocker node?  Were you modifying the "Other Linux" instructions on the client installation page to install a Flocker node manually?

Thanks.

moypray:
hi , @jongiddy, 

I just use "pip install https://clusterhq-archive.s3.amazonaws.com/python/Flocker-1.1.0-py2-none-any.whl" to install flocker.

source flocker-client/bin/activate
flocker-deploy  version    could work.

but when I try to start flocker-control (not using virtualenv) it does not work. so i use pip install some required package.
run flocker-control met this problem. after the fix, it works, and could run containers. and then submit this PR.

itamarst:
`flocker-control` is not really intended to be installed this way, only the client, so it's not surprising it didn't work. It is surprising that you ended up with wrong version of Twisted, though, I tried following instructions locally and ended up with correct version...

Anyway I'll try to do a CI run on this and see if we can get it merged - thanks for the patch!

adamtheturtle:
@moypray As per https://docs.clusterhq.com/en/1.1.0/gettinginvolved/contributing.html#id2 feel free to add your name to `AUTHORS.rst` and once this is merged get in touch for some swag.

moypray:
@adamtheturtle  Update the PR, Am I right?  
I hope I could submit some PR on Flocker too. I am very interesting in it.

thanks

moypray:
@adamtheturtle 
And more, Is there something about debuging flocker docs or wiki ?

Now I setup a control service on a Host,  and a node in another Host(Node B),  and Node B could run container success with flocker-plugin volumes.  but the local control service Node could work.

adamtheturtle:
@moypray I will run this build when GitHub is no longer suffering from the DDOS attack (see https://status.github.com).

The latest released docs on debugging Flocker are at https://docs.clusterhq.com/en/1.1.0/using/administering/debugging.html. However those have been improved in `master` since the last release. We will be doing another release shortly and it will include a diagnostics tool.

moypray:
@adamtheturtle  ok, thanks

adamtheturtle:
@moypray First GitHub had problems, then PyPI, and now it turns out that your name and company set off our spellchecker! I have a Pull Request accepted (modulo tests passing) for the spellchecker to ignore the AUTHORS file at https://github.com/ClusterHQ/flocker/pull/1883. When that is merged (and GitHub is still having problems) then I will run the tests on this and continue tomorrow.

Thanks for your patience

adamtheturtle:
The GitHub report is wrong, there is no Build pending for the vagrant/build/tutorial builder and a build already passed - http://build.clusterhq.com/boxes-flocker?branch=refs%2Fpull%2F1863%2Fhead. Thanks @moypray , I'll merge this. Remember to get in touch if you'd like some swag, and thanks again for your contribution.

PR #1866, opened 2015-08-21
[FLOC-2903] Remove duplicate packge installation code

itamarst:
Mostly just removing duplication (3 copies of one bit of logic, 2 copies of other logic) in preparation of adding code to install yet another package. Some minor changes were however made along the way:

1. Client package installs no longer use sudo.
2. No longer run `yum clean all` before installing flocker node package.


docs/install/install-client.rst
```diff
@@ -16,24 +16,20 @@ The following sections describe how to install the Flocker client on your platfo
...
 ============
 
-.. note:: These instructions require that you have ``sudo`` access.
-
-On Ubuntu 15.04, the Flocker CLI can be installed from the ClusterHQ repository:
+On Ubuntu 15.04, the Flocker CLI can be installed from the ClusterHQ repository as ``root``, e.g. by running ``sudo bash``:
 
 .. task:: cli_pkg_install ubuntu-15.04
-   :prompt: alice@mercury:~$
+   :prompt: root@mercury:~$
```
adamtheturtle: :question: `root` usually uses `#` instead of `$`, right? At least in our docs.

adamtheturtle:
Thanks @itamarst.

IIRC the tests for these instructions were removed earlier, but in refactoring like this it would be useful to see them pass.

I'm wondering why `yum clean all` was deemed necessary at some point, but is not necessary now.

I tested the Ubuntu CLI instructions manually with the instructions at http://doc-dev.clusterhq.com/gettinginvolved/client-testing.html and unfortunately the output was:

```
root@vagrant-ubuntu-trusty-64:~# apt-get -y --force-yes install clusterhq-flocker-cli
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following packages were automatically installed and are no longer required:
  libdrm-intel1 libdrm-nouveau2 libdrm-radeon1 libfontenc1 libgl1-mesa-dri
  libgl1-mesa-glx libglapi-mesa libice6 libllvm3.4 libpciaccess0 libpixman-1-0
  libsm6 libtxc-dxtn-s2tc0 libx11-xcb1 libxaw7 libxcb-dri2-0 libxcb-dri3-0
  libxcb-glx0 libxcb-present0 libxcb-sync1 libxcomposite1 libxdamage1
  libxfixes3 libxfont1 libxkbfile1 libxmu6 libxpm4 libxrandr2 libxrender1
  libxshmfence1 libxt6 libxxf86vm1 x11-common x11-xkb-utils xfonts-base
  xfonts-encodings xfonts-utils xserver-common xserver-xorg-core
Use 'apt-get autoremove' to remove them.
The following extra packages will be installed:
  clusterhq-python-flocker
The following NEW packages will be installed:
  clusterhq-flocker-cli clusterhq-python-flocker
0 upgraded, 2 newly installed, 0 to remove and 0 not upgraded.
Need to get 27.2 MB of archives.
After this operation, 97.4 MB of additional disk space will be used.
WARNING: The following packages cannot be authenticated!
  clusterhq-python-flocker clusterhq-flocker-cli
Get:1 https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/  clusterhq-python-flocker 1.2.0-0.dev.1 [27.2 MB]
Err https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/  clusterhq-python-flocker 1.2.0-0.dev.1
  HttpError403
Get:2 https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/  clusterhq-flocker-cli 1.2.0-0.dev.1 [1,164 B]
Err https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/  clusterhq-flocker-cli 1.2.0-0.dev.1
  HttpError403
E: Failed to fetch https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/.//clusterhq-python-flocker_1.2.0-0.dev.1_amd64.deb  HttpError403

E: Failed to fetch https://clusterhq-archive.s3.amazonaws.com/ubuntu-testing/14.04/amd64/.//clusterhq-flocker-cli_1.2.0-0.dev.1_all.deb  HttpError403

E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

So I then tried the same with the instructions from doc-dev (http://doc-dev.clusterhq.com/install/install-client.html#ubuntu-14-04). I got the same error.

With `ubuntu-testing` replaced with `ubuntu` these instructions work. It seems therefore that something has broken the Ubuntu packages in the last dev release, perhaps the new `dpkg` tool used.

I then tried the CentOS node instructions and all seemed to work.

Please merge when you're happy, and I'll look into / make issues for the broken dev package.

adamtheturtle:
I've got a patch for the above mentioned issue in another PR.

jongiddy:
The reason for using `sudo` in the client instructions is that it is expected that this will be installed on the user's local laptop, and therefore they will generally use `sudo`.

The node installation specifically creates a root login, so it makes sense for those instructions to use root directly.

This seems an unfortunate regression in usability.

itamarst:
If necessary we can fix the usability issues in a way that doesn't involve copy/pasting code.

itamarst:
More specifically, if you feel that it's a sufficiently big deficit in this branch then I can fix that before merging.

itamarst:
OK, got a fix I think, will push soon.

itamarst:
Since I made some more changes (docs now use sudo) might merit another review round.

flocker/provision/_install.py
```diff
@@ -316,48 +315,85 @@ def install_cli_commands_ubuntu(distribution, package_source):
...
     return sequence(commands)
 
 
 _task_install_commands = {
-    'centos-7': install_cli_commands_yum,
-    'ubuntu-14.04': install_cli_commands_ubuntu,
-    'ubuntu-15.04': install_cli_commands_ubuntu,
+    'centos-7': install_commands_yum,
+    'ubuntu-14.04': install_commands_ubuntu,
+    'ubuntu-15.04': install_commands_ubuntu,
```
jongiddy: Keeping this mapping and adding the `is_centos/ubuntu` functions means that there are now multiple ways to convert between a specific distribution and a base system (analogous to supported packaging system).  It would be better to have a single source of truth for this.  See https://clusterhq.atlassian.net/browse/FLOC-1670 for some ideas.

flocker/provision/_install.py
```diff
@@ -316,48 +315,85 @@ def install_cli_commands_ubuntu(distribution, package_source):
...
         package.
 
     :return: a sequence of commands to run on the distribution
     """
-    return _task_install_commands[distribution](distribution, package_source)
+    commands = task_package_install("clusterhq-flocker-cli", distribution,
+                                    package_source)
+    if is_ubuntu(distribution):
+        # We want to use sudo for better documentation output, even though
+        # it's not actually necessary since this will run as root.
```
jongiddy: Change this to "client testing is currently done using root". The only reason we test the client installation as root is that the recent change to use Docker for client testing makes it easier to use the root account.

flocker/provision/_install.py
```diff
@@ -316,48 +315,85 @@ def install_cli_commands_ubuntu(distribution, package_source):
...
     :param bytes distribution: The distribution the node is running.
     :param PackageSource package_source: The source from which to install the
         package.
 
     :return: a sequence of commands to run on the distribution
     """
-    return _task_install_commands[distribution](distribution, package_source)
+    commands = task_package_install("clusterhq-flocker-cli", distribution,
+                                    package_source)
+    if is_ubuntu(distribution):
```
jongiddy: This is not required. We only run CLI installations for Ubuntu, both for docs and testing. If we add other distributions in the future, they will probably use `sudo` as well.

flocker/provision/_install.py
```diff
@@ -316,48 +315,85 @@ def install_cli_commands_ubuntu(distribution, package_source):
...
+        # We want to use sudo for better documentation output, even though
+        # it's not actually necessary since this will run as root.
+        return sequence([
+            (Effect(Sudo(command=e.intent.command,
+                         log_command_filter=e.intent.log_command_filter))
+             if isinstance(e.intent, Run) else e)
+            for e in commands.intent.effects])
```
jongiddy: This is pretty ugly and slightly obscure. Maybe pass Run/Sudo as a parameter to `task_package_install` :question:

flocker/provision/_install.py
```diff
@@ -273,39 +279,32 @@ def install_cli_commands_ubuntu(distribution, package_source):
...
-
     commands = [
         # Minimal images often have cleared apt caches and are missing
         # packages that are common in a typical release.  These commands
         # ensure that we start from a good base system with the required
         # capabilities, particularly that the add-apt-repository command
         # is available, and HTTPS URLs are supported.
-        sudo_from_args(["apt-get", "update"]),
-        sudo_from_args([
+        run_from_args(["apt-get", "update"]),
```
jongiddy: It's nice that this is now consistent between client and node installations.  Previously, the node installation did not have this. It's usually only needed on Docker images.  But it may prevent problems if we Dockerise Flocker.

jongiddy:
A few issues to address for `task_cli_pkg_install` but looking good otherwise.

flocker/provision/_install.py
itamarst: Fedora doesn't do `sudo` out of the box I think? Or has that changed in modern Fedora?

flocker/provision/_install.py
itamarst: I just got rid of the dict.

flocker/provision/_install.py
itamarst: Yeah, that would be better. But this sufficiently better than before that it's no longer a blocker for my work, so I'll probably just not bother.

flocker/provision/_install.py
```diff
@@ -58,6 +58,18 @@ def is_centos(distribution):
...
+    Determine whether the named distribution is a version of Ubuntu.
+
+    :param bytes distribution: The name of the distribution to inspect.
+
+    :return: ``True`` if the distribution named is a version of Ubuntu,
+        ``False`` otherwise.
+    """
+    return distribution.startswith("ubuntu-")
+
+
```
jml: No action required. 

I mentioned on a previous review that having an `is_centos` function and a bunch of methods with `centos` in the name was a hint that maybe there should be an object. This change makes that more true. However, out of scope for this refactoring.

jml:
:+1: LGTM. Merge when it passes the tests, unless there are significant changes required to make it do so.

PR #1868, opened 2015-08-24
[FLOC 2905] Fix broken link in install node doc

davidoliverSP2:
Fixes 2905


jongiddy:
It's a bit awkwardly phrased, and maybe it should go ahead of the Flocker installation instructions. But it resolves the issue as is.

OK to merge. (It was agreed that patches that only affect `.rst` files can be merged when `docs` test passes).

PR #1869, opened 2015-08-24
Ubuntu packages floc 2904

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2904

I ran `admin/publish-artifacts --flocker-version=1.2.0.dev1` from the tag of the last dev release with these changes. This has been tested by running the Ubuntu CLI instructions from http://doc-dev.clusterhq.com/install/install-client.html#ubuntu-14-04.

The initial problem was that various versions of `dpkg-scanpackages` had different outputs - we didn't notice until we started doing releases outside of Vagrant. The approach that this uses is to introduce a Python dependency which has consistent output across the `dpkg-scanpackages` versions I've tested and has the side-effect of making the code nicer, in my opinion.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -3,13 +3,19 @@ apache-libcloud==0.17.0
 awscli==1.7.35
 backports.ssl-match-hostname==3.4.0.2
 botocore==1.0.0rc1
+cached-property==1.2.0
+chardet==2.3.0
 colorama==0.3.3
+coloredlogs==1.0.1
+deb-pkg-tools==1.34
 docutils==0.12
 ecdsa==0.13
+executor==4.4
 Fabric==1.10.2
 flake8==2.4.1
 gitdb==0.6.4
 GitPython==1.0.1
+humanfriendly==1.33
 hypothesis==1.10.3
 Jinja2==2.7.3
 jmespath==0.7.1
@@ -28,6 +34,7 @@ Pygments==2.0.2
 pymongo==3.0.2
 PyMySQL==0.6.6
 python-dateutil==2.4.2
+python-debian==0.1.23
 requests-file==1.3.1
 rsa==3.1.4
 smmap==0.9.0
--- a/admin/yum.py
+++ b/admin/yum.py
@@ -8,8 +8,9 @@ import requests
 from requests_file import FileAdapter
 from characteristic import attributes
 from effect import sync_performer, TypeDispatcher
-from subprocess import check_call, check_output
+from subprocess import check_call
 from gzip import GzipFile
+from deb_pkg_tools.repo import scan_packages
 
 from flocker.common.version import make_rpm_version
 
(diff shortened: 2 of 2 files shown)


exarkun:
I can't understand the problem description or how the change fixes it.

adamtheturtle:
The problem was that installing the latest development release of the Flocker CLI on Ubuntu failed (with an error as mentioned in the JIRA issue). The `Packages` file in the `ubuntu-testing` repository had filenames like `.//clusterhq-flocker-cli_1.2.0-0.dev.1_all.deb` (note `.//` at the beginning) while the working `Packages` file in the `ubuntu` repository had filenames like `./clusterhq-flocker-cli_1.2.0-0.dev.1_all.deb` (one fewer `/` at the beginning).

The change made which caused the difference is that we used to use just one version of `dpkg-scanpackages` (the one in the Vagrant dev box), but now we are using various versions (on users' machines). The version I used in the last dev release (`1.18.1`) created a `Packages` file which was unusable. I plan to write an upstream issue.

The options then included:
* Fixing the version of `dpkg-scanpackages`
* Writing something which normalises the output so it doesn't matter which version you use

I found this Python package (deb-pkg-tools) which does the latter so chose to use it. I ran the release script as a test and it seems to work.

dev-requirements.txt
```diff
@@ -3,13 +3,19 @@ apache-libcloud==0.17.0
 awscli==1.7.35
 backports.ssl-match-hostname==3.4.0.2
 botocore==1.0.0rc1
+cached-property==1.2.0
+chardet==2.3.0
 colorama==0.3.3
+coloredlogs==1.0.1
+deb-pkg-tools==1.34
```
dwgebler: I'm guessing all the other additions to this file are dependencies of `deb-pkg-tools`?

dwgebler:
The use of `deb-pkg-tools` seems reasonable here to solve the problem. It requires a few external programs (outside of Python packages) to be on the user's machine, but mostly things that will already be on any Debian-type machine.

Note: you say "which has consistent output across the `dpkg-scanpackages` versions I've tested" - my understanding of this package is that it doesn't use the `dpkg-scanpackages` tool but rather reimplements its behaviour in Python. That's probably a good thing for our use case here.

I'm OK with this, given that it avoids us all requiring a fixed scanpackages version to be able to do a release. I'm not enough of an expert in packaging to know if it's the best solution possible, but it's at least better than what it's replacing.

Assuming the answer to the question on the additions to `dev-requirements.txt` is a yes, please merge.

dev-requirements.txt
adamtheturtle: That is correct.

adamtheturtle:
> Note: you say "which has consistent output across the dpkg-scanpackages versions I've tested" - my understanding of this package is that it doesn't use the dpkg-scanpackages tool but rather reimplements its behaviour in Python. That's probably a good thing for our use case here.

It uses the `dpkg` tools under the hood. Those are installed on each platform in the release process at http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/release-process.html#prerequisites. Though this reimplements `dpkg-scanpackages -m`, it uses the `dpkg` developer external program (see https://deb-pkg-tools.readthedocs.org/en/latest/index.html?highlight=error#dependencies). This can be tested by running the `scan_packages` function after uninstalling the dpkg dev tools - you'll see an error.

I tested the output with dpkg-scanpackages version 1.18.1 on OS X and version 1.17.5 on Ubuntu, two versions which had different outputs with the old code. This was with one package downloaded from our S3 bucket. The re-release of 1.2.0.dev1 was done with 1.18.1, which caused the initial problem but worked with this code.

PR #1870, opened 2015-08-24
[FLOC-2831] Acceptance tests use the new API client

itamarst:


flocker/acceptance/testtools.py
```diff
@@ -304,11 +304,15 @@ class Cluster(PRecord):
     :ivar Node control_node: The node running the ``flocker-control``
         service.
     :ivar list nodes: The ``Node`` s in this cluster.
-    :ivar treq: A ``treq`` client.
+
+    :ivar treq: A ``treq`` client, eventually to be completely replaced by
+        ``FlockerClient`` usage.
```
jml: Why eventually? Was there any particular reason for not going the whole way in this PR? 

I try to put judgement calls like this in the description of a PR.

flocker/acceptance/testtools.py
```diff
@@ -322,133 +326,44 @@ def base_url(self):
...
-            self.base_url + b"/state/datasets", persistent=False)
-        request.addCallback(check_and_decode_json, OK)
-        return request
+        :param Dataset expected_dataset: The configured dataset that
+            we're waiting for in state.
 
-    @log_method
-    def wait_for_dataset(self, dataset_properties):
+        :returns: A ``Deferred`` which fires with ``expected_datasets``
+            when cluster state matches the configuration for the given
```
jml: Nit: s/when cluster/when the cluster/

flocker/acceptance/testtools.py
```diff
@@ -322,133 +326,44 @@ def base_url(self):
...
-            request.addCallback(got_body)
+            request = self.client.list_datasets_state()
+
+            def got_results(results):
+                # State has unpredictable path, so we don't bother
+                # checking for its contents:
+                for dataset in results:
+                    if dataset.set(path=None) == expected_dataset_state:
+                        return True
+                return False
```
jml: It took me a while to puzzle the logic of this one. I don't have any good ideas on what can make it more clear. Perhaps

```python
return expected_dataset_state in [dataset.set(path=None) for dataset in results]
```

:question:

flocker/acceptance/testtools.py
```diff
@@ -322,133 +326,44 @@ def base_url(self):
...
-                dataset_id.encode('ascii'),
-            ),
-            headers={b"content-type": b"application/json"},
-            persistent=False
-        )
-
-        request.addCallback(check_and_decode_json, OK)
-        return request
-
-    @log_method
```
jml: Glorious. Thanks.

jml:
:+1: Thanks, it really is nice to see code being deleted and simplified.

One minor change, one optional suggestion, and one question. Although the question is kind of fundamental, the patch as it stands looks like a solid improvement to me so I'm happy for it to go in. 

The tests are failing. Please make sure they pass before landing. Please resubmit for review if major changes are required.

flocker/acceptance/testtools.py
itamarst: Because the client doesn't support the container API yet, so you can't replace all the calls.

PR #1871, opened 2015-08-24
[FLOC-2909] Use openshift/busybox-http-app:latest

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2909

This is a work-around for a Docker 1.8.1 / Docker hub bug.  Hopefully Docker upstream will fix whatever the underlying problem is.  This avoids us having hanging/failing test suite meanwhile.

Note that BuildBot is still using an older version of Docker - probably around 1.3.3.  Ideally we would have coverage on the CI system for the old and new versions of Docker until we finish the transition.  Failing that, build results against Docker 1.8.1 are available at http://54.200.233.212/boxes-flocker?branch=http-app-latest-FLOC-2909 (for as long as I happen to be running that alternate master).

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -31,7 +31,7 @@ from ...testtools import (
     loop_until, find_free_port, DockerImageBuilder, assertContainsAll,
     random_name)
 
-from ..test.test_docker import make_idockerclient_tests
+from ..test.test_docker import ANY_IMAGE, make_idockerclient_tests
 from .._docker import (
     DockerClient, PortMap, Environment, NamespacedDockerClient,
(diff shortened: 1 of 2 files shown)


itamarst:
I'm seeing red on the linked buildbot...

exarkun:
Yea.  Sorry, my explanation was incomplete.  This is step 1 of 5 or so to get the suite passing against Docker 1.8.1.  See sub-tasks of FLOC-2902 for the other steps.  Maybe http://54.200.233.212/boxes-flocker?branch=master is a good point of comparison for this work (master against Docker 1.8.1).  Unfortunately those results vs the results I linked to also don't show an obvious improvement either...

itamarst:
Ah, OK. Merge when green.

PR #1874, opened 2015-08-24
[FLOC-2910] Factor out code for running a registry and pushing an image

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2910

This will be used to help resolve FLOC-2911 and FLOC-2912, at least.

Note diff is wrong, it includes already-merged changes from #1871

PR #1875, opened 2015-08-24
[FLOC-2911] [FLOC-2882] Don't use Docker hub in test_pull_timeout

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2911
Fixes https://clusterhq.atlassian.net/browse/FLOC-2882

This removes the possibility of intermittent failures of this test caused by problems with Docker hub itself or with the test client's network connection to Docker hub.

FLOC-2882 is fixed as a side-effect due to the deterministic control afforded (and exercised) over the local registry the test now talks to.

See FLOC-2902 and FLOC-2910 for further context.

Depends on #1874 (note diff is currently wrong, including those changes)

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -564,31 +564,97 @@ class GenericDockerClientTests(TestCase):
         registry_listening.addCallback(lambda ignored: registry)
         return registry_listening
 
+    def _pull_timeout(self):
+        """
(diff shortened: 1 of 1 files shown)


itamarst:
Please address comment above; if you have demonstrated this fails well then merge, otherwise if it's a bigger change resubmit.

PR #1876, opened 2015-08-24
[FLOC-2914] Rename test to make it reflect what it does

itamarst:


.gitignore
```diff
@@ -44,6 +44,7 @@ htmlcov/
 nosetests.xml
 coverage.xml
 _trial_temp*
+.hypothesis
```
jml: Thanks, I had meant to mention that in a previous review, but multi-tasking got the better of my brain.

flocker/control/test/test_persistence.py
```diff
@@ -407,7 +359,7 @@ def test_sequential_migrations(self):
         self.assertEqual(result, StubMigration.upgrade_from_v2(v2_config))
 
 
-UUIDS = st.basic(generate=lambda r, _: UUID(int=r.getrandbits(128)))
+UUIDS = st.sampled_from([uuid4() for i in range(1000)])
```
jml: Any evidence that this helps?

flocker/control/test/test_persistence.py
```diff
@@ -479,7 +431,7 @@ def _build_node(applications):
     APPLICATIONS,
     # If we add this hint on the number of applications, Hypothesis is able to
     # run many more tests.
-    average_size=10,
+    average_size=3,
```
jml: Related, from `#hypothesis` earlier today:

```
[15:30:13]  <DRMacIver>	jml: BTW it looks like you can speed your tests up quite a lot if you just replace your usage of text() with text(average_size=5)
```

I would suggest giving this idiom a name: e.g. `short_text`, or `label`.

(UK time)

flocker/control/test/test_persistence.py
```diff
@@ -495,32 +447,62 @@ def _build_node(applications):
...
+        self.assertFalse(isinstance(wire_decode(data), Temp))
+
+    def test_complex_keys(self):
+        """
+        Objects with attributes that are ``PMap``\s with complex keys
+        (i.e. not strings) can be roundtripped.
+        """
+        node_state = NodeState(hostname=u'127.0.0.1', uuid=uuid4(),
+                               manifestations={}, paths={},
+                               devices={uuid4(): FilePath(b"/tmp")})
```
jml: No action suggested.

One of the nice things with hypothesis is that you can easily use it to fill in data in a specific case like this, e.g.

```python
node_state = NodeState(hostname=IP_ADDRESSES.example(), ...)
```

(Which is the [Anonymous Creation Method](http://xunitpatterns.com/Creation%20Method.html) pattern, for those playing at home)

But I don't think we have generators for host names or filepaths, and the uuid generator is exactly the same.

jml:
Thanks for the cleanup! One optional change suggested and one question that I'd love to know the answer to. No need for follow-up review.

flocker/control/test/test_persistence.py
itamarst: I'll try that as alternative to the other two. They definitely combined cut time from 60 seconds to 36 seconds on my machine.

itamarst: The short text thing sped up tests from 35 seconds to 26 seconds. Unfortunately applying it would involve another Buildbot CI round which means I'm not going to be able to commit this PR today, so I'm not going to bother.

PR #1877, opened 2015-08-24
[FLOC-2912] Don't use Docker hub in test_pull_timeout_pull

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2912

See FLOC-2902, FLOC-2910, and FLOC-2911 for further context.

Depends on #1875 (note diff is currently wrong, including those changes)

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1877)


itamarst:
`test_pull_timeout_pull` is failing... in two different ways. So you might need to fix that first. This specifically seems fine, if you can get green build.

exarkun:
> `test_pull_timeout_pull` is failing... in two different ways

First way it fails - port auto-assignment is unreliable.  Already ran into this elsewhere in the test suite ... so I'll factor that solution out so I can re-use it in `run_registry`.

exarkun:
Second way it fails - pulling images from `registry:2` is unreliable so ...

exarkun:
Here's the 2nd failure Itamar mentioned, http://build.clusterhq.com/builders/flocker-ubuntu-14.04/builds/5973/steps/trial/logs/problems

I failed to reproduce it locally.  Some additional logging, and perhaps less insanely verbose container and image names, would probably make debugging it easier ... supposing it's possible to reproduce.

PR #1879, opened 2015-08-25
[FLOC 2339] Update to authentication docs from MY feedback

davidoliverSP2:
Fixes 2339

This issue resolves the feedback from @myechuri, following an installation run-through.

Note that this only fixes point 4 in the issue. I think points 1 - 3 have already been addressed in previous PRs. Please advise if that is not the case.


jml:
:+1: LGTM. Merge when ready.

PR #1880, opened 2015-08-25
Change instance types floc 2891

Azulinho:
Disabling SElinux setenforce 0 calls, as selinux is disabled on the new baked images.
Had to update the instance types as I had a few out-mem conditions with the T2micros.
With the new images we aren't requited to use the M3 old generation instances and can use T2 all around


itamarst:
Looks fine (but unrelated to FLOC-2891 so a little confusing).

PR #1881, opened 2015-08-25
Remove setenforce from aws centos jenkins jobs floc 2891

Azulinho:
need to disable selinux setenforce 0 calls, as selinux is disabled on the baked image.
This causes the job to fail


itamarst:
And also we're in theory supporting SELinux any hour now.

itamarst:
Looks good, feel free to merge.

adamtheturtle:
Ideally there would be a link to a failing job (old code) and a passing job (new code).

PR #1884, opened 2015-08-25
[FLOC-2885] Docker plugin acceptance test

itamarst:
This is just enough to setup the package, enable the remote Docker API, and run a single simple test. Later issues will add more tests.

Using the remote Docker API may come in handy for other kinds of tests as well.


flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -0,0 +1,90 @@
...
+        tls = TLSConfig(
+            client_cert=(get_path(b"user.crt"), get_path(b"user.key")),
+            # Blows up if not set
+            # (https://github.com/shazow/urllib3/issues/695):
+            ssl_version=ssl.PROTOCOL_TLSv1,
+            # Don't validate hostname, we don't generate it correctly, but
+            # do verify certificate authority signed the server certificate:
+            assert_hostname=False,
+            verify=get_path(b"cluster.crt"))
+        client = Client(base_url="https://{}:2376".format(address), tls=tls)
```
jml: Why 2376? Can you please give this number a name?

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+        data = "hello world"
+        node = cluster.nodes[0]
+
+        volume_name = random_name(self)
+        self.run_python_container(
+            cluster, node.public_address,
+            {"host_config": create_host_config(
+                binds=["{}:/data".format(volume_name)],
+                port_bindings={8080: 8080}),
```
jml: Is this source port and destination port? Are the 8080s below references to the first or the last?

Defining variables for each and using them (even if it's simply `port = 8080` just after `node` is defined) will make the test a bit more self-documenting.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -0,0 +1,90 @@
...
+
+from docker import Client
+from docker.tls import TLSConfig
+from docker.utils import create_host_config
+
+from ...testtools import random_name
+from ..testtools import (
+    require_cluster, post_http_server, assert_http_server,
+)
+from ..obsolete.test_containers import CURRENT_DIRECTORY
```
jml: What's `obsolete`?  It doesn't seem like a great idea to be importing from there in new code.

flocker/provision/_install.py
```diff
@@ -711,6 +772,27 @@ def task_enable_flocker_control(distribution):
...
+
+    :param bytes distribution: The distribution name.
+    """
+    if is_centos(distribution):
+        return sequence([
+            run_from_args(['systemctl', 'enable', 'flocker-docker-plugin']),
+            run_from_args(['systemctl', START, 'flocker-docker-plugin']),
+            run_from_args(['systemctl', START, 'docker']),
+        ])
+    elif distribution == 'ubuntu-14.04':
```
jml: Maybe it hasn't landed yet, but another recent PR adds an `is_ubuntu` checker. Should you be using that?

jml:
:+1: Merge when tests pass.

flocker/provision/_install.py
itamarst: No, because newer versions of ubuntu will use `systemd`.

itamarst: Should be doing something *better*, certainly.

flocker/acceptance/endtoend/test_dockerplugin.py
itamarst: Oh, hm, did I forget to move stuff? Bleh.

itamarst: This is what happens when I have to continuiously context switch.

`obsolete` is for stuff that we are supporting only because we have to, in particular container API. I will move stuff out of there, yeah.

PR #1892, opened 2015-08-27
Use selinux baked images on jenkins floc 2924

Azulinho:
Jenkins tests passing,
see attached pdf on FLOC-2924.

This commit updates the build.yaml file so that the jenkins jobs use a SELinux enabled instance on both rackspace and AWS.


build.yaml
```diff
@@ -282,8 +282,8 @@ common_cli:
 job_type:
   run_trial:
   # http://build.clusterhq.com/builders/flocker-centos-7
-    run_trial_on_AWS_CentOS_7:
-      on_nodes_with_labels: 'aws-centos-7-T2Small'
+    run_trial_on_AWS_CentOS_7_with_SELinux:
```
tomprince: :question: Given that you are just changing the images used for the build, is there any reason to also rename the builds? It seems just like a new configuration of the environment, rather than a different builder.

build.yaml
```diff
@@ -427,9 +427,9 @@ common_cli:
 # flocker.node.functional is hanging, so we don't run it
 job_type:
   run_trial:
-  # http://build.clusterhq.com/builders/flocker-centos-7
+    # http://build.clusterhq.com/builders/flocker-centos-7
     run_trial_on_AWS_CentOS_7:
-      on_nodes_with_labels: 'aws-centos-7-T2Small'
+      on_nodes_with_labels: 'aws-centos-7-SELinux-T2Small'
```
adamtheturtle: Why is this particular configuration detail in the node label?
Surely there are many configuration details which are not in the labels.
Is it because the labels must be differentiated from the last set of AMI labels? If so, maybe we should come up with a system (dates? version numbers?) for labelling.

This can be left for a later date but please reply letting me know your thoughts.

build.yaml
```diff
@@ -554,7 +553,7 @@ job_type:
       clean_repo: true
 
     run_trial_for_cinder_storage_driver_on_Ubuntu_trusty:
-      on_nodes_with_labels: 'rackspace-ubuntu14-jenkins-slave-DFW-standard-4'
+      on_nodes_with_labels: 'rackspace-jenkins-slave-ubuntu14-standard-4-dfw'
```
adamtheturtle: Not new so don't address now ftr I think that `ubuntu-14-` is a bad part of the label as there are multiple 2014 Ubuntus. Perhaps we should standardise on what the Flocker code uses which is `ubuntu-14.04`, `centos-7` etc.

All of these labels seem to have a different standard. Some include `jenkins-slave`, some say the region. I guess that there are good reasons for these but we should document them so that anyone can add a new image.

build.yaml
```diff
@@ -1,4 +1,4 @@
-# vim ft=yaml:fdm=indent
+# vim: ai ts=4 sts=4 et sw=4 ft=yaml fdm=indent et foldlevel=0
```
adamtheturtle: I'll just assume that this is right.

build.yaml
```diff
@@ -427,9 +427,9 @@ common_cli:
 # flocker.node.functional is hanging, so we don't run it
 job_type:
   run_trial:
-  # http://build.clusterhq.com/builders/flocker-centos-7
+    # http://build.clusterhq.com/builders/flocker-centos-7
```
adamtheturtle: Is there some kind of formatter / standard used here which people without vi could use.

I ask because before we had `flake8` tests on our Python we had horrible diffs when e.g. someone's editor formatted things and a reviewer had to manually discern what was a relevant change and what wasn't. Again, something to consider but not address now.

adamtheturtle:
As the PDF shows something working this is preliminarily accepted. Re-run the tests ( :cry: ) until they pass.

As there is a related PR, don't merge this until that is accepted.

build.yaml
Azulinho: "because the labels must be differentiated from the last set of AMI labels"
that's pretty much it, the config in jenkins for nodes with 'aws-centos-7-T2small' use an AMI which doesn't contain SELinux enabled.

"should come up with a system (dates? version numbers?) for labelling"
Probably worth thinking of, question is if a new image for a slave should require changes to the build.yaml, in my view it shouldn't if one is simply updating the slave with OS patches.

On the other hand there are some benefits in doing so, for instance it would become visible in the build.yaml that the base slave image has changed  and  that any issues that occurred could be related back to that change

build.yaml
Azulinho: they do lack a standard

as we move to multi-region across different cloud providers we can tackle that
<provider><os><version><region><vm-type>  

the EC2-plugin requires different AMIs for capping instance numbers across different VM-types.
If we were to use the jclouds plugin for EC2 as we do for Rackspace we could possibly get rid of the 'vm-type'

build.yaml
Azulinho: ha! 
http://vim.wikia.com/wiki/Modeline_magic

build.yaml
Azulinho: nope, that's just me indenting the comment to match the line it refers to below that comment.
Not really vi related, the top line 'the modeline' simply says 2 spaces no tabs.
Keep in mind this is YAML not python

Azulinho:
"As the PDF shows something working this is preliminarily accepted. Re-run the tests ( :cry: ) until they pass."

The tests don't pass, and that is unrelated to this PR

PR #1894, opened 2015-08-27
[FLOC-2913] Disable pyrsistent C extension earlier in the startup process

itamarst:
Eliot imports pyrsistent, so we need to disable C extension first.

Also fix docstring and copyright, since I was touching the module anyway.


jml:
:+1: LGTM. It gets much further on my debug build, so it's likely that it fixes the problem.

PR #1878, opened 2015-08-24
[FLOC-2915] Stop using Docker hub in test_pull_image_if_necessary

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-2915

Depends on #1877 (note the diff is wrong, including those changes)

PR #1896, opened 2015-08-28
[FLOC-2949] Mark failing Docker tests as expected failing

jml:
As discussed on [FLOC-2949](https://clusterhq.atlassian.net/browse/FLOC-2949), I prematurely upgraded the version of docker on the slaves, causing the tests to fail. 

This patch disables those tests in the sure and certain hope that we can land [FLOC-2912](https://clusterhq.atlassian.net/browse/FLOC-2912) (#1877) and [FLOC-2915](https://clusterhq.atlassian.net/browse/FLOC-2915) (#1878) quickly.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1896)

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -104,6 +104,10 @@ class GenericDockerClientTests(TestCase):
     """
     clientException = APIError
 
+    UPGRADING_DOCKER = (
+        "FLOC-2902: Some of these tests will fail while we are upgrading "
+        "docker.")
+
     @if_docker_configured
     def setUp(self):
         self.namespacing_prefix = namespace_for_test(self)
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks @jml  for taking this on.

So, we have a few TODO tests. I think that we need a policy / system for dealing with them and making sure that we don't just forget them. In this case I'll trust that you're on top of this one.

As well as Expected Failures it would be really nice to have a section of "Failures we care about now but are known so you can ignore them in your branch". Maybe that's a ratchet system.

Relatedly I think we need a policy for how and when to upgrade Docker on slaves, and how to test first that the tests won't be broken.

Anyways, merge when the tests pass.

jml:
All valid points. Re:

> Relatedly I think we need a policy for how and when to upgrade Docker on slaves, and how to test first that the tests won't be broken.

I would rather automate the problem away, but point taken.

Thanks for the review!

PR #1897, opened 2015-08-28
[FLOC-2529] Add notice for out of date docs

msh100:
Added javascript to AJAX GET https://docs.clusterhq.com/en/lates/version.html and compare to DOCUMENTATION_OPTIONS

Added new theme with no markup for version output. Added logic to use plain layout for version.rst.

Fixes FLOC-2529

### DIFF
--- a/docs/version.rst
+++ b/docs/version.rst
@@ -1,0 +1,1 @@
+|version|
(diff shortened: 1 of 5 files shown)


docs/version.rst
```diff
@@ -0,0 +1 @@
+|version|
```
jml: We're getting warnings about how this file isn't in the table of contents:
```
/srv/buildslave/flocker-docs/build/docs/version.rst:: WARNING: document isn't included in any toctree
```
e.g. http://build.clusterhq.com/builders/flocker-docs/builds/9718

Can you please either add it to the toctree or suppress the warning? (I'm afraid I don't know how to do the latter).

msh100: @adamtheturtle has pushed a fix for this

jml:
Lovely, thanks. :+1:

PR #1898, opened 2015-08-28
[FLOC-2872] Support incoming X-Eliot header in the REST API for log tracing.

itamarst:
This will allow us to trace log messages from the Docker plugin as it calls into the Flocker REST API.


flocker/restapi/_infrastructure.py
```diff
@@ -110,6 +110,35 @@ def failure(reason):
...
+    The remote task's context will be extracted from a C{X-Eliot} HTTP header.
+
+    :param original: Function to wrap.
+
+    :return: Wrapped function.
+    """
+    @wraps(original)
+    def logger(self, request, **routeArguments):
+        serialized_remote_task = request.requestHeaders.getRawHeaders(
+            "X-Eliot", [None])[0]
```
tomprince: Any reason not to use the header used in the example from the documentation (`X-Eliot-Task-Id`)?

flocker/restapi/test/test_infrastructure.py
```diff
@@ -653,3 +653,51 @@ def test_notFound(self):
...
+
+        logged_parent = LoggedAction.ofType(logger.messages, parent)[0]
+        child = _assertRequestLogged(b"/foo/bar")(self, logger)
+        self.assertIn(child, logged_parent.descendants())
+
+    @capture_logging(_assertRequestLogged(b"/foo/bar", b"GET"))
+    def test_malformed_task(self, logger):
+        """
+        If the contents of the ``X-Eliot`` header are malformed, processing
+        continues as normal and logging just starts a new task.
```
tomprince: :question: This doesn't appear to verify the second part of the description (that a new task is started). I don't see any reason that needs to be verified, but if we don't the description shouldn't include it, probably.

tomprince:
This looks good. Merge after optionally addressing the above two comments.

PR #1899, opened 2015-08-28
[FLOC-2740] Respect leases when calculating changes

itamarst:
If a lease is held on a dataset on the current node we shouldn't delete it or resize it or unmount it, since that means it's in use. (Eventually leases will be the only way to make things in-use, but for now we also consider application state.)


jml:
Lovely, thanks.

PR #1901, opened 2015-08-30
[FLOC-2736] REST API endpoint: list leases

itamarst:
1. Expand schema to know about leases. The tests were structured so the good/bad examples can be re-used in later schema tests.
2. Add a new endpoint for listing leases.

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -280,3 +280,36 @@ definitions:
       primary:
         '$ref': '#/definitions/primary'
     additionalProperties: false
+
+  lease_expiration:
+    title: "Lease Expiration"
+    description: |
+      The number of seconds until a lease expires, or null to indicate no expiration.
+    type:
+      - "number"
+      - "null"
+
+  lease_dataset_id:
+    title: "Dataset"
+    description: |
+      An opaque identifier, unique across the cluster, identifying a
+      particular dataset.
+    type: string
+    # The length of a stringified uuid
+    minLength: 36
+    maxLength: 36
+
+  lease:
+    type: object
+    properties:
+      dataset_id:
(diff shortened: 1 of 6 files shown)


itamarst:
Missing docs, will add those tomorrow I guess.

flocker/control/test/test_httpapi.py
```diff
@@ -42,6 +45,7 @@
 from .._clusterstate import ClusterStateService
 from .._config import (
     FlockerConfiguration, FigConfiguration, model_from_configuration)
+from .._model import Leases
```
dwgebler: Minor quibble, but the rest of our models used in these tests are imported from module level defined in `__init__.py`

flocker/control/schema/types.yml
```diff
@@ -280,3 +280,36 @@ definitions:
...
+
+  lease_expiration:
+    title: "Lease Expiration"
+    description: |
+      The number of seconds until a lease expires, or null to indicate no expiration.
+    type:
+      - "number"
+      - "null"
+
+  lease_dataset_id:
```
dwgebler: Any reason to not reuse the existing `dataset_id` type definition?

dwgebler:
Couple of minor inline comments to address or answer, otherwise please resubmit with the docs when they've been added. Thanks.

flocker/control/schema/types.yml
itamarst: The prose doesn't quite make sense in this API endpoint's context, alas.

itamarst: Which suggests a bigger problem but not one I'm going to think too hard about for now.

dwgebler:
Thanks, LGTM when the build passes.

PR #1904, opened 2015-08-31
[FLOC-2739] Delete leases endpoint

itamarst:
Builds on and refactors #1901 so that should be reviewed first.

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -894,7 +894,7 @@ class ConfigurationAPIUserV1(object):
         now = datetime.fromtimestamp(self.clock.seconds(), UTC)
         result = []
         for lease in self.persistence_service.get().leases.values():
-            result.append(lease_to_json(lease, now))
+            result.append(lease_response(lease, now))
         return result
 
     @app.route("/configuration/leases/<dataset_id>", methods=['DELETE'])
@@ -938,7 +938,7 @@ class ConfigurationAPIUserV1(object):
                 # safety that adds... so just accept all releases.
                 lambda leases: leases.release(dataset_id, lease.node_id),
                 self.persistence_service)
-            d.addCallback(lambda _: lease_to_json(lease, now))
+            d.addCallback(lambda _: lease_response(lease, now))
             return d
         else:
             # Didn't find the lease:
@@ -1172,7 +1172,7 @@ def create_api_service(persistence_service, cluster_state_service, endpoint,
     )
 
 
-def lease_to_json(lease, now):
+def lease_response(lease, now):
     """
     Convert a ``Lease`` to the corresponding objects for JSON
     serialization.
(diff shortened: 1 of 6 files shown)


flocker/control/httpapi.py
```diff
@@ -853,6 +865,74 @@ def replace_configuration(self, applications, deployment):
...
+
+        :param unicode dataset_id: The dataset whose lease is being
+            removed.
+
+        :return: A ``Deferred`` firing with an ``EndpointResponse`` or
+            serializable JSON.
+        """
+        now = datetime.fromtimestamp(self.clock.seconds(), UTC)
+        leases = self.persistence_service.get().leases
+        dataset_id = UUID(dataset_id)
```
dwgebler: This could throw a `ValueError` in the event of `dataset_id` in the request path not being a valid UUID.

flocker/control/httpapi.py
```diff
@@ -1079,3 +1159,21 @@ def create_api_service(persistence_service, cluster_state_service, endpoint,
             Site(api_root)
         )
     )
+
+
+def lease_to_json(lease, now):
```
dwgebler: Kind of a misleading function name, since it doesn't actually return JSON. The containers equivalent is called `container_configuration_response`.

flocker/control/schema/types.yml
```diff
@@ -280,3 +280,37 @@ definitions:
...
+      An opaque identifier, unique across the cluster, identifying a
+      particular dataset.
+    type: string
+    # The length of a stringified uuid
+    minLength: 36
+    maxLength: 36
+
+  lease:
+    type: object
+    description: "A specific lease."
```
dwgebler: "A lease on a dataset" or something like that might be a better description, "a specific lease" doesn't really say anything more than you'd get from just the word "lease".

flocker/control/httpapi.py
```diff
@@ -853,6 +865,74 @@ def replace_configuration(self, applications, deployment):
...
+        return result
+
+    @app.route("/configuration/leases/<dataset_id>", methods=['DELETE'])
+    # This can stop being private as part of FLOC-2741:
+    @private_api
+    @user_documentation(
+        u"""
+        This will release a lease on a dataset, allowing it to be moved
+        elsewhere, deleted and otherwise modified.
+        """,
```
dwgebler: :question: Do you think we should clarify in the reference docs that this doesn't do anything to the dataset the lease concerns?

dwgebler:
Thanks, a few inline comments to address or answer, then please merge after #1901 has been merged.

flocker/control/httpapi.py
itamarst: Filed FLOC-2695; this impacts a bunch of other endpoints as well and so should be solved for all of them. It therefore isn't going to be an isolated fix, thus a new issue.

flocker/control/httpapi.py
itamarst: Fixed.

flocker/control/schema/types.yml
itamarst: Fixed.

flocker/control/httpapi.py
itamarst: Fixed.

PR #1905, opened 2015-09-01
[FLOC-2965] Validate rest API UUIDs

dwgebler:
Fixes FLOC-2965

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -11,7 +11,7 @@ definitions:
     minLength: 36
     maxLength: 36
     # UUIDv4, as defined by RFC 4122. See http://tools.ietf.org/html/rfc4122.html#section-4.1
-    pattern: "^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-4[A-Fa-f0-9]{3}-[89ab][A-Fa-f0-9]{3}-[A-Fa-f0-9]{12}$"
+    pattern: "^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-4[A-Fa-f0-9]{3}-[89abAB][A-Fa-f0-9]{3}-[A-Fa-f0-9]{12}$"
 
   dataset_id:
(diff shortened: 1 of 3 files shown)


flocker/control/schema/types.yml
```diff
@@ -1,16 +1,19 @@
 $schema: http://json-schema.org/draft-04/schema#
 id: http://api.clusterhq.com/v1/types.json
 definitions:
+  uuid:
+    type: string
+    minLength: 36
+    maxLength: 36
+    pattern: "^[A-Fa-f0-9]{8}-[A-Fa-f0-9]{4}-4[A-Fa-f0-9]{3}-[89ab][A-Fa-f0-9]{3}-[A-Fa-f0-9]{12}$"
```
jml: Maybe add a comment linking to the relevant bit of RFC 4122?

flocker/control/schema/types.yml
```diff
@@ -1,16 +1,19 @@
 $schema: http://json-schema.org/draft-04/schema#
 id: http://api.clusterhq.com/v1/types.json
 definitions:
+  uuid:
```
jml: Tests are failing because this is missing a title.

jml:
Thanks! Couple of documentation suggestions. Please merge once the tests pass, or resubmit for review if more changes are required to pass tests.

flocker/control/schema/types.yml
dwgebler: Dang and blast it, the `$ref` to this definition in the types that use it mean that their titles and descriptions are ignored. Some sort of solution using `allOf` is the only thing I can think of, but I don't think our docs builder for the API reference can handle that. I'll look at the options in more detail - obviously the simplest solution is to use a generic title and description in the type definition here, but that would probably make some sections of the API reference confusing, because we use UUIDs for both nodes and datasets.

itamarst: I suspect supporting the specific case of allOf with only a single item wouldn't be too hard.

PR #1906, opened 2015-09-01
[FLOC-2737] API endpoint for creating leases

itamarst:

PR #1907, opened 2015-09-01
[FLOC-2738] API endpoint for renewing leases

itamarst:
Builds on #1906 so review that first.


jml:
Nice. Thanks. No comments. Please merge it passes the tests.

Also, you might want to cancel some of the pending builds from past pushes in order to get the latest build further up in the queue.

PR #1908, opened 2015-09-01
[FLOC-2954] Leases support in API client

itamarst:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2954

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -79,12 +79,12 @@ class Lease(PClass):
     A lease on a dataset.
 
     :attr UUID dataset_id: The dataset for which the lease applies.
-    :attr UUID node_id: The node for which the lease applies.
+    :attr UUID node_uuid: The node for which the lease applies.
     :attr None|float|int expires: Time in seconds until lease expires, or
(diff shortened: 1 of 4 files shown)


flocker/apiclient/_client.py
```diff
@@ -186,6 +247,34 @@ def synchronize_state(self):
...
+    def release_lease(self, dataset_id):
+        # We don't handle the case where lease doesn't exist yet, since
+        # it's not clear that's necessary yet. If it is we'll need to
+        # expand this logic.
+        lease = self._leases[dataset_id]
+        self._leases = self._leases.release(dataset_id, lease.node_id)
+        return succeed(
+            Lease(dataset_id=dataset_id, node_uuid=lease.node_id,
+                  # Just lie about expiration time, it doesn't really
+                  # matter when releasing:
```
tomprince: If it doesn't matter, that suggest suggests the client (and the REST API?) shouldn't bother including it either. But, since it is easy enough to get here, we might as well populate it correctly.

tomprince:
Address the one comment and then merge.

PR #1910, opened 2015-09-02
Improve cinder get device path floc 2931

zubron:
These are the changes to fix issue FLOC-2931. The current implementation of ```cinder.get_device_path``` returns the device mapping provided by libvirt which does not recognise externally attached disks. This results in incorrect device paths being returned when non-Cinder volumes have been attached.

If a non-Cinder device is attached to the guest at device path ```/dev/vdc```, the device will actually attach to ```/dev/vdb```. If a Cinder volume is then attached, the Cinder API will report that it attached to ```/dev/vdb``` but it actually attached to ```/dev/vdc```.

To fix this issue, we instead return the udev symlink created for attached Cinder volumes, which is constructed using the first 20 characters of the UUID of the Cinder volume. In the case where this device path is not available (for example, OpenStack on Rackspace or if the udev rules have been modified by the user) we rely on the correct device path being returned from Cinder.

Three tests have been introduced with this change covering different scenarios.

```test_get_device_path_no_attached_disks``` is a test for the current behaviour. The devices under ```/dev``` are compared before and after a volume attachment. Only one new device should be added and the path for this should be the same path returned from ```cinder.get_device_path```.

The next two tests rely on using virsh on the host machine to attach external disks to the guests which causes the previously described behaviour to be exhibited. Unfortunately these tests can not be run using our existing CI system and are currently being run manually on a devstack instance running on an on-metal Rackspace server. To run these tests, ssh into the Rackspace server then ssh into the guest node on this server. The code has already been checked out on this guest and the tests can be run there. Please ask me for connection details.

An additional key ```devstack-openstack``` has been added to the ```MINIMUM_ALLOCATABLE_SIZES``` dictionary as devstack/OpenStack allows smaller volumes to be attached. This also removes the need to increase the size of devstack Cinder LVM pool.

In order for the guest to interact with libvirt on the devstack/OpenStack host machine, TLS access is provided for libvirtd. The connections are unverified as this removes the need to transfer files from the host to the guest during test set-up.

```test_get_device_path_correct_with_attached_disk``` covers the issue described above where the incorrect device path is being returned. This is the test that now passes after this change.

```test_disk_attachment_fails_with_conflicting_disk``` is for another issue that was discovered while working on this issue. If a non-Cinder volume is attached at the next available device slot (in the test case, ```/dev/vdb```), Cinder will fail when attempting to attach a volume which would result in this device path being used. The API will initially report success and the device will go into 'attaching' state, however it will then fail to attach and will not go into the 'in-use' state. Currently, an exception is raised on time-out. This will be addressed in FLOC-2970.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1910)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -76,3 +76,4 @@ vagrant/dev/*.box
 .Trashes
 ehthumbs.db
 Thumbs.db
+jobs.groovy
--- a/admin/build_targets/centos-7/Dockerfile
+++ b/admin/build_targets/centos-7/Dockerfile
@@ -3,15 +3,8 @@
 # A Docker image for building RPMs in a clean CentOS 7 build environment.
 #
 
-FROM centos:centos7
+FROM clusterhqci/fpm-centos-7
 MAINTAINER ClusterHQ <contact@clusterhq.com>
-# URLGRABBER_DEBUG=1 to log low-level network info - see FLOC-2640
-RUN ["env", "URLGRABBER_DEBUG=1", "yum", "groupinstall", "--assumeyes", "Development Tools"]
-RUN ["env", "URLGRABBER_DEBUG=1", "yum", "install", "--assumeyes", "epel-release"]
-RUN ["env", "URLGRABBER_DEBUG=1", "yum", "install", "--assumeyes", "git", "ruby-devel", "python-devel", "libffi-devel", "openssl-devel", "python-pip", "rpmlint"]
-# Despite being a packaging tool, fpm isn't yet packaged for CentOS.
-# See https://github.com/jordansissel/fpm/issues/611
-RUN ["gem", "install", "fpm"]
 COPY requirements.txt /tmp/
 RUN ["pip", "install", "-r", "/tmp/requirements.txt"]
 VOLUME /flocker
--- a/admin/build_targets/ubuntu-14.04/Dockerfile
+++ b/admin/build_targets/ubuntu-14.04/Dockerfile
@@ -4,15 +4,10 @@
 # environment.
 #
 
-FROM ubuntu:14.04
+FROM clusterhqci/fpm-ubuntu-trusty
 MAINTAINER ClusterHQ <contact@clusterhq.com>
-RUN ["apt-get", "update"]
-RUN ["apt-get", "install", "--no-install-recommends", "-y", "git", "ruby-dev", "libffi-dev", "libssl-dev", "build-essential", "python-pip", "python2.7-dev", "lintian"]
-# Despite being a packaging tool, fpm isn't yet packaged for Ubuntu
-# See https://github.com/jordansissel/fpm/issues/611 and
-# https://github.com/jordansissel/fpm/issues/657
-RUN ["gem", "install", "fpm"]
 COPY requirements.txt /tmp/
+RUN echo UMASK 022 >> /etc/logins.defs
 RUN ["pip", "install", "-r", "/tmp/requirements.txt"]
 VOLUME /flocker
 ENTRYPOINT ["/flocker/admin/build-package-entrypoint", "--destination-path=/output"]
--- a/admin/build_targets/ubuntu-15.04/Dockerfile
+++ b/admin/build_targets/ubuntu-15.04/Dockerfile
@@ -4,14 +4,8 @@
 # environment.
 #
 
-FROM ubuntu:15.04
+FROM clusterhqci/fpm-ubuntu-vivid
 MAINTAINER ClusterHQ <contact@clusterhq.com>
-RUN ["apt-get", "update"]
-RUN ["apt-get", "install", "--no-install-recommends", "-y", "git", "ruby-dev", "libffi-dev", "libssl-dev", "build-essential", "python-pip", "python2.7-dev", "lintian"]
-# Despite being a packaging tool, fpm isn't yet packaged for Ubuntu
-# See https://github.com/jordansissel/fpm/issues/611 and
-# https://github.com/jordansissel/fpm/issues/657
-RUN ["gem", "install", "fpm"]
 COPY requirements.txt /tmp/
 RUN ["pip", "install", "-r", "/tmp/requirements.txt"]
 VOLUME /flocker
--- a/build.yaml
+++ b/build.yaml
@@ -1,0 +1,602 @@
+# vim ft=yaml:fdm=indent
+# To modify an existing job, update the dictionary 'jobs_type'
+# commit, push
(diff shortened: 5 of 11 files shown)


jml:
> The next two tests rely on using virsh on the host machine to attach external disks to the guests which causes the previously described behaviour to be exhibited. Unfortunately these tests can not be run using our existing CI system and are currently being run manually on a devstack instance running on an on-metal Rackspace server. 

Can you please file a bug / subtask for this, and maybe add a comment near `require_virsh`?


Patch looks good! Thanks so much for tracking this done. A few minor suggestions and one question. If the question has an easy answer, then feel free to merge when the tests pass.

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/agents/cinder.py, line 479 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyCYJwiM4D11zKdYPe3)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/cinder.py#L479)):</sup>
This seems like a good place to link to the upstream bug in openstack, if that's actually relevant.

---

<sup>**[flocker/node/agents/cinder.py, line 481 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyCYTzuTBGHche7f_fw)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/cinder.py#L481)):</sup>
I don't know enough about UUIDs. 

What happens if there's another device with the same first twenty digits?

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 251 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_FxpCGSJUc54bqQr)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L251)):</sup>
Please use `flocker.testtools.run_process` here. It has better logging, captures stderr, and raises nicer exceptions.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 261 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_NJqmiLhMnDnvWlI)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L261)):</sup>
Likewise here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1910)

zubron:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/agents/cinder.py, line 481 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyCYTzuTBGHche7f_fw)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/cinder.py#L481)):</sup>
I'm not sure what would happen. The device paths are created by udev rules, so if another device happened to match the first twenty characters of the UUID, I don't know if OpenStack would handle this situation or whether it would result in some type of collision within udev when creating the new device path. It would obviously be better if the full UUID was used in the device path but unfortunately this is the most distinct way of identifying the device. I think it's unlikely that a collision will occur, but if one did occur, it would be a much more serious issue for OpenStack in general.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1910)

zubron:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/agents/cinder.py, line 479 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyCYJwiM4D11zKdYPe3)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/cinder.py#L479)):</sup>
I've added the relevant Cinder bug and libvirt documentation.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 251 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_FxpCGSJUc54bqQr)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L251)):</sup>
Done.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 261 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_NJqmiLhMnDnvWlI)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L261)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1910)

zubron:
Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 253 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_FxpCGSJUc54bqQr)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L253)):</sup>
Done.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 263 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1910#-JyC_NJqmiLhMnDnvWlI)** ([raw file](https://github.com/clusterhq/flocker/blob/b317c3cd5534b83bbedff30284902850b441fd3c/flocker/node/agents/functional/test_cinder.py#L263)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1910)

jml:
Thanks. I'm happy for this to be merged when the tests pass.

---

Review status: 0 of 3 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1910)

PR #1911, opened 2015-09-02
[FLOC-2741] End to end tests for leases

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2741

This isn't quite what we designed, notably:
 * We can't just move / delete a dataset from underneath a running container; the Docker bind mount prevents the Flocker dataset being unmounted.
 * The 10s delay is really ugly, but how else can we catch the potential important error that would result in flocker-dataset-agent unmounting a leased dataset?
 * We'd originally planned to write tests for expiring leases too, but that's difficult because we can't really know what expiry time to use when acquiring the lease. It depends on how quickly docker responds to API requests. We could set a really long expiry time, but that would slow down the tests.

### DIFF
--- a/flocker/acceptance/endtoend/test_leases.py
+++ b/flocker/acceptance/endtoend/test_leases.py
@@ -34,64 +34,51 @@ class LeaseAPITests(TestCase):
 
         def acquire_lease(dataset):
             # Call the API to acquire a lease with the dataset ID.
+            # Give the lease an expiry of 60 seconds.
             pass
 
-        def attempt_move_dataset(dataset):
-            pass
+        def start_http_container(dataset):
+            # In the 60 seconds before the lease expires, we can launch
+            # the acceptance tests' data HTTP container and make POST requests
+            # to it in a looping call every second.
+            # return looping call deferred
 
-        XXX what do we do here? Some options:
-        * Modify the dataset configuration API:
-          to return an error if the dataset has a lease?
+        def request_move_dataset(self):
+            # We can then request to move the dataset attached to the container.
+            return deferred
 
-        * Wait and timeout?
-          The dataset-agent won't move the dataset because it has a lease so
-          the state will not be updated.
-          But how do we distinguish this from a dataset move which has not
-          yet been performed (delayed for some reason)?
+        def wait_some_amount_of_time(self):
+            # Because the dataset is leased, we should be able to continue
+            # writing data via HTTP requests to the running container.
+            # We should be able to do this for some number of seconds.
+            # The looping call running in parallel to this is continuing to
+            # write data.
 
-        * Check the dataset-agent logs?
-          except that `calculate_changes` doesn't currently log anything if a
-          dataset is leased to nodeA but configured to be on nodeB.
+            # wait some time and return a deferred
 
-        * Don't attempt an acceptance test for this situation.
-          It's already tested in the blockdevice API tests...
-        """
-        self.fail("not implemented yet")
+        def stop_container(self):
+            # We stop the container, so that there is no constraint outside
+            # of leases to prevent the volume from being unmounted.
 
-    @require_moving_backend
-    @require_cluster(2)
-    def test_move_dataset_after_lease_release(self, cluster):
-        """
-        A dataset can be moved once a lease held on it by a
-        particular node is released.
+            def confirmed_stopped(self):
+                # When the container is confirmed as stopped,
+                # restart the container again, to prove the dataset
+                # hasn't moved.
+                # If the dataset had moved, the host path on the volume would
+                # have been unmounted. # XXX is this actually true, does
+                # Flocker remove mountpoint directories?
 
-        ...instead we could write only this test which issues the same API
-        calls that the Docker Plugin is likely to make eg
+        def stop_container_again(self):
+            # Stop the container again.
 
-        Kubernetes schedules Postgres on NodeA with -v PostgresData
-            Docker(NodeA) calls ``VolumePlugin.mount`` PostgresData
-                Flongle responds by:
-                    Create dataset for FooBar
-                    Acquire lease for FooBar on NodeA
-            Docker(NodeA) starts Postgres
+            def stopped_again(self):
+                # Ask for the lease to be released, causing the dataset to move.
+                # Wait a couple of seconds.
 
-        Kubernetes schedules Postgres on NodeB with -v PostgresData
-            Docker(NodaA) stops container
-            Docker(NodeA) calls ``VolumePlugin.umount`` PostgresData
-                Flongle releases lease for FooBar on NodeB
-
-            Docker(NodeB) calls ``VolumePlugin.mount`` PostgresData
-                Flongle
-                    Checks for PostgresData -- exists
-                    Checks for PostgresData leases -- none
-                    Move PostgresData to NodeB
-                    Acquire lease for FooBar on NodeB
-            Docker(NodeB) starts Postgres
-
-        @itamarst, is this what you mean by:
-        > so perhaps try thinking about leases from point of view of users,
-          e.g. docker plugin - how and when will they use it? what would they
-          care about?
+        def try_to_start_again(self):
+            # After a couple of seconds, we can try to recreate the container
+            # and that should fail, because the host mount path for the volume
+            # should no longer exist.
         """
         self.fail("not implemented yet")
 
--- a/flocker/acceptance/endtoend/test_leases.py
+++ b/flocker/acceptance/endtoend/test_leases.py
@@ -33,7 +33,6 @@ class LeaseAPITests(TestCase):
         """
         http_port = 8080
         dataset_id = uuid4()
-        dataset_path = []
         datasets = []
         client = get_docker_client(cluster, cluster.nodes[0].public_address)
         d = create_dataset(
@@ -51,7 +50,6 @@ class LeaseAPITests(TestCase):
                 getting_datasets = cluster.client.list_datasets_state()
 
                 def extract_dataset_path(datasets):
-                    dataset_path.insert(0, datasets[0].path)
                     return datasets[0].path
 
                 getting_datasets.addCallback(extract_dataset_path)
@@ -66,7 +64,6 @@ class LeaseAPITests(TestCase):
             # Launch data HTTP container and make POST requests
             # to it in a looping call every second.
             # return looping call deferred
-            # import pdb;pdb.set_trace()
             script = SCRIPTS.child("datahttp.py")
             script_arguments = [u"/data"]
             docker_arguments = {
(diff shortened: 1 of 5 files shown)


flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,132 @@
...
+    @require_cluster(2)
+    def test_move_dataset_after_lease_release(self, cluster):
+        """
+        A dataset can be moved once a lease held on it by a
+        particular node is released.
+
+        ...instead we could write only this test which issues the same API
+        calls that the Docker Plugin is likely to make eg
+
+        Kubernetes schedules Postgres on NodeA with -v PostgresData
```
itamarst: No postgres please.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,119 @@
...
+            return deferred
+
+        def wait_some_amount_of_time(self):
+            # Because the dataset is leased, we should be able to continue
+            # writing data via HTTP requests to the running container.
+            # We should be able to do this for some number of seconds.
+            # The looping call running in parallel to this is continuing to
+            # write data.
+
+            # wait some time and return a deferred
```
wallrj: We're not quite sure about this bit. 
In discussions with @itamarst we guessed that Flocker would be able to unmount a leaseless dataset as long as the container had no *open* files in the volume. The datahttpserver script doesn't keep open files, so fine, but...
Another possible reason that Flocker wouldn't be able to unmount the leaseless dataset is that Docker has bind mounted the Flocker mount point.
So the next step may be necessary.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,119 @@
...
+
+        def stop_container(self):
+            # We stop the container, so that there is no constraint outside
+            # of leases to prevent the volume from being unmounted.
```
wallrj: i.e. Docker has bind mounted the `/flocker/dataset-id` directory which would prevent Flocker from unmounting the dataset, regardless of  a lease.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,119 @@
...
+
+        def acquire_lease(dataset):
+            # Call the API to acquire a lease with the dataset ID.
+            # Give the lease an expiry of 60 seconds.
+            pass
+
+        def start_http_container(dataset):
+            # In the 60 seconds before the lease expires, we can launch
+            # the acceptance tests' data HTTP container and make POST requests
+            # to it in a looping call every second.
```
jml: What happens if we can't do that in 60s?

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,119 @@
...
+        self.fail("not implemented yet")
+
+    @require_moving_backend
+    @require_cluster(2)
+    def test_move_dataset_after_lease_expiry(self, cluster):
+        """
+        A dataset can be moved once a lease held on it by a
+        particular node has expired.
+        """
+        self.fail("not implemented yet")
```
jml: Isn't this case handled by the big test earlier? (Line 75 of this diff.)

jml:
Looks fine to me, thanks. I've got a couple of questions, but nothing that should prevent coding.

flocker/acceptance/endtoend/test_leases.py
wallrj: Yeah, I think we should probably acquire a lease without an expiry time.

flocker/acceptance/endtoend/test_leases.py
wallrj: The big test explicitly releases the lease rather than waiting for the lease to expire.
But since the lease expiration is tested elsewhere, we might not bother with this acceptance test.
Perhaps the one acceptance test is sufficient.

flocker/acceptance/testtools.py
```diff
@@ -544,21 +600,34 @@ def delete_items(items):
...
                 lambda item: self.client.delete_dataset(item.dataset_id),
             )
 
-        return cleanup_containers().addCallback(lambda _: cleanup_datasets())
+        def cleanup_leases():
+            get_items = self.client.list_leases()
+
+            def release_all(leases):
+                for lease in leases:
+                    self.client.release_lease(lease.dataset_id)
```
jml: Shouldn't this gather the results of release_lease into a deferred and return that? i.e. is it a problem that this doesn't wait for the leases to be released?

If it isn't a problem, then please add a comment explaining why.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,229 @@
...
+        * Wait for the previously requested operation to complete.
+
+        :param Cluster cluster: The cluster on which to operate.
+        :param operation: The ``FlockerClient`` method to call before releasing
+            the lease.
+        :param dict additional_kwargs: Any additional arguments to pass to
+            ``operation``.
+        :param state_method: A callable which returns a ``Deferred`` that fires
+            when the requested operation has been performed.
+        :returns: A ``Deferred`` that fires when the all the steps have
```
jml: Nit: "the all the" -> "all the"

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,229 @@
...
+            completed.
+        """
+        http_port = 8080
```
jml: What happens if something is already listening on this port? 

JP recently added `add_with_port_collision_retry` to `flocker.node.testtools`. Could we use that here to make the test more reliable?

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,229 @@
...
+            d.addCallback(lambda lease: leases.insert(0, lease))
+            return d
+        acquiring_lease = getting_dataset_state.addCallback(acquire_lease)
+
+        def start_http_container(ignored):
+            """
+            Launch data HTTP container and make POST requests
+            to it in a looping call every second.
+            return looping call deferred
+            """
```
jml: This docstring doesn't seem to match the code. 

It doesn't return a looping call deferred, and doesn't do any POSTing that I can see.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,229 @@
...
+
+        def wait_for_unexpected_umount(ignored):
+            """
+            If the dataset agent is broken and not respecting leases then we
+            expect that 10 seconds is long enough for it to begin performing
+            the requested operation.
+            And the first step will always be to unmount the filesystem which
+            should happen quickly since there are no open files and no Docker
+            bind mounts to the filesystem.
+            """
```
jml: I'm sorry, I'm probably missing something obvious, but I don't quite understand this. 

Does this mean that if the dataset agent is broken then it will unbreak after 10 seconds? Or does it mean that if it's not working after 10 seconds then it's definitely broken? 

Also, I don't quite see how the second sentence relates to the first sentence. It says "And the first step" but I can't tell what it's the first step of.

jml:
Thanks!

Caveats 1 & 3 are fine. I don't really understand 2 (re: 10s delay), and my lack of understanding shines through in my comments on the PR. 

I'm probably going to spend this afternoon dealing with a customer emergency, but I'd be happy to have a call to resolve my confusion if you think that would help.

I'd like to take another look after you've replied to the review, if that's OK.

jml:
Thanks. I made a couple of comments on the [commit itself](https://github.com/ClusterHQ/flocker/commit/d7290deab89003e7e7111e80569c53483eaadaaf). I'd appreciate it if you could reply to them and get back to me.

dwgebler:
> Using add_with_port_collision_retry wasn't an option?

No, in this test we need more granular control over our Docker client. This does find a free host port though.

dwgebler:
> would it be possible to have a retry loop here that stops retrying at 10s?

I'm not sure, I can't immediately envision how that would work. In the 10s period, we don't expect anything to happen. If we were to try restarting the container immediately in a retry-loop, we'd expect it to also succeed immediately - but that could happen even if leases weren't being properly respected, because there may not even have been time to unmount, so it wouldn't tell us anything useful to this test case.

jml:
> > Using add_with_port_collision_retry wasn't an option?
> No, in this test we need more granular control over our Docker client. This does find a free host port though.

It does, but the port that was free when you called `find_free_port` might get taken before you start the container. That's why there needs to be a retry in the case of port collision.

flocker/acceptance/endtoend/test_leases.py
```diff
@@ -0,0 +1,238 @@
...
+            dataset) operation.
+            The first step of such an operation will always be to unmount
+            the filesystem, which should happen quickly since there are no
+            open files and no Docker bind mounts to the filesystem.
+            Therefore if after 10 seconds, we can restart the container and
+            successfully write some data to it, we can conclude that the
+            dataset agent has respected the lease and not attempted to
+            unmount.
+            """
+            return deferLater(reactor, 10, lambda: None)
```
jml: > > would it be possible to have a retry loop here that stops retrying at 10s?

> I'm not sure, I can't immediately envision how that would work. In the 10s period, we don't expect anything to happen. If we were to try restarting the container immediately in a retry-loop, we'd expect it to also succeed immediately - but that could happen even if leases weren't being properly respected, because there may not even have been time to unmount, so it wouldn't tell us anything useful to this test case.

OK, I think I understand now. You're trying to prove a negative, which is intrinsically hard. This is a good compromise.

dwgebler:
> It does, but the port that was free when you called find_free_port might get taken before you start the container. That's why there needs to be a retry in the case of port collision.

In the Docker tests we fire up lots of containers and it takes time for them to shut down, while we asynchronously continue operations, so a port collision is more likely. In these acceptance tests we're using the Docker API directly in a blocking manner, so is it really a risk here?

jml:
Thanks for the changes and the explanations.

Regarding the port collision retry, it's unclear to me whether it's worth getting it right now, so I'll leave the decision up to you. If you do fix it to retry, please submit for review. Otherwise, merge when the tests pass.

jml:
> > It does, but the port that was free when you called find_free_port might get taken before you start the container. That's why there needs to be a retry in the case of port collision.

> In the Docker tests we fire up lots of containers and it takes time for them to shut down, while we asynchronously continue operations, so a port collision is more likely. In these acceptance tests we're using the Docker API directly in a blocking manner, so is it really a risk here?

It's always a risk when you assume that everything else is going to well behaved. It sounds like it's an acceptable risk here though. (http://www.kitchensoap.com/2014/03/19/paradigm-check-point-prefacing-debriefings/ is pretty good on this)

PR #1912, opened 2015-09-03
[FLOC-2970] Improved wait error

jongiddy:
This PR adds a more explicit check of the states that Cinder volumes traverse when performing create/delete and attach/detach operations.  This allows unexpected states to raise a more useful exception. See, in particular, the test `test_disk_attachment_fails_with_conflicting_disk` in `test_cinder.py`, which now sees an UnexpectedStateError rather than a TimeoutError.  The error is both more informative and occurs sooner.

To quote https://github.com/ClusterHQ/flocker/pull/1910:
"Unfortunately these tests can not be run using our existing CI system and are currently being run manually on a devstack instance running on an on-metal Rackspace server. To run these tests, ssh into the Rackspace server then ssh into the guest node on this server. The code has already been checked out on this guest and the tests can be run there. Please ask [@zubron] for connection details."

There are still some potential problems, such as another instance attaching a volume before the polling sees a final 'available' state, but these can be worked out for FLOC-1832.

### DIFF
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -516,6 +516,7 @@ class CinderBlockDeviceAPI(object):
             volume_manager=self.cinder_volume_manager,
             expected_volume=cinder_volume,
             desired_state=u'available',
+            transient_states=(u'detaching',)
         )
 
     def destroy_volume(self, blockdevice_id):
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -240,7 +240,7 @@ class TimeoutException(Exception):
         return (
             'Timed out while waiting for volume. '
             'Expected Volume: {!r}, '
-            'Expected Status: {!r}, '
+            'Expected State: {!r}, '
             'Elapsed Time: {!r}, '.format(
                 self.expected_volume, self.desired_state, self.elapsed_time)
             )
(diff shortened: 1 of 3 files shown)


flocker/node/agents/cinder.py
```diff
@@ -227,46 +227,141 @@ def list():
...
-    # Log stuff happening in this loop.  FLOC-1833.
-    while True:
+    def __init__(self, volume_manager, expected_volume,
+                 desired_state, transient_states=(), time_limit=60):
+        self.volume_manager = volume_manager
+        self.expected_volume = expected_volume
+        self.desired_state = desired_state
+        self.transient_states = transient_states
+        self.time_limit = time_limit
+        self.start_time = time.time()
```
jml: :question: Not strictly necessary, but I'd make `start_time` a parameter here, and make `now` a parameter of `reached_desired_state` so this class doesn't need to know about IO at all.

jml:
Looks good. One optional suggestion. I noticed a test failure so kicked off another run. Please merge when the tests pass.

Thanks!

flocker/node/agents/cinder.py
jongiddy: I looked at this, but there is no reason that the `wait_for_volume_state` would pass the current time in general. It only makes sense when the code knows a timeout is possible.  I think it would make more sense to pass a Clock object to the VolumeStateMonitor, which knows about timeouts.  Although this might help in future testing, it doesn't seem necessary right now, especially as there are several issues (FLOC-1831,-1832,-1833) which may change this code significantly later. (I'd like the EC2 and Cinder code to share more at some point).

PR #1915, opened 2015-09-04
[FLOC-2975] Cleanup PKI keys used for testing

jongiddy:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2975

This is a cleanup for the tests added in #1910 

Those tests copied PKI keys and certs into the `/etc/pki` directories, and left them there.

This PR creates the keys and certs in a temporary directory, which is removed at the end of test.

This PR also renames the factory function for the `VirtIOClient` to give a better idea of it's main characteristics; `from_instance_id` was a terrible but simple name. `using_insecure_tls` is more informative, and emphasises that this should not be used outside it's current context without detailed consideration.

To quote PR #1910:
> Unfortunately these tests can not be run using our existing CI system and are currently being run manually on a devstack instance running on an on-metal Rackspace server. To run these tests, ssh into the Rackspace server then ssh into the guest node on this server. The code has already been checked out on this guest and the tests can be run there. Please ask [@zubron] for connection details.

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -203,6 +203,19 @@ class VirtIOClient:
         communication to the routing gateway (in particular from VM
         guest to VM host), where a MITM attack is unlikely.
 
+        The tests require that disks are attached using libvirt, but not
+        using Cinder, as the problem they test is libvirt disks that are
+        not known by Cinder.  Note, this rules out solutions using
+        ``mknod`` directly on the guest.
+
+        Creating a TLS connection is one of the simplest ways to set-up
+        libvirtd to listen on a network socket.  Disabling the actual
+        certificate verification on both ends of the connection allows
+        connection of the TLS endpoint without sharing any files (e.g.
+        CA cert and key, or a CSR).  This means the tests are contained
+        on one guest, with only a network connection required to attach
+        and delete nodes from the host.
+
         :param instance_id: The UUID of the guest instance.
         :param FilePath tempdir: A temporary directory that will exist
             until the VirtIOClient is done.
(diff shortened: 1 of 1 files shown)


flocker/node/agents/functional/test_cinder.py
```diff
@@ -335,7 +329,9 @@ def test_get_device_path_correct_with_attached_disk(self):
         instance_id = self.blockdevice_api.compute_instance_id()
 
         host_device = "/dev/null"
-        virtio = VirtIOClient.from_instance_id(instance_id)
+        tmpdir = FilePath(self.mktemp())
+        tmpdir.makedirs()
+        virtio = VirtIOClient.using_insecure_tls(instance_id, tmpdir)
```
wallrj: :question: It's a matter of opinion, but I think if I was doing this, I might: 
 * make `create_credentials` a free function (rather than a static method) which returns the `FilePath` 
 * then use it as follows: 
```
virtio = VirtIOClient.using_insecure_tls(
    instance_id=instance_id, 
    credentials_directory=create_credentials(
        base_directory=FilePath(self.mktemp())
    )
) 
```

wallrj:
Thanks @jongiddy 

* I ran the tests and noted that there is no longer any polution of the /etc/pki directory
* We just discussed it and you explained the reason, but it does seem slightly pointless to be using TLS and client certificates if the libvirt daemon has been configured to accept any certificate signed by any random authority.
* I think you should write some notes / comments explaining why it's done this way and also create a follow up issue containing ideas for  more secure mechanisms of forcing disks to be attached to the guest.
* Or even consider whether there's a simpler way to occupy `/dev/vd?` devices ....perhaps by calling `mknod` in the guest with the correct major and minor device numbers?

But that's all future work, so please merge after answering or addressing my comments.

flocker/node/agents/functional/test_cinder.py
jongiddy: It's a good idea, and passing a tmpdir is kind of awkward. But the credential creation code is dependent on how virsh works, particularly the requirement for specific filenames. For this reason, I'll leave it as is.

PR #1916, opened 2015-09-04
[FLOC-2976] Respect leases when updating node

dwgebler:
Fixes FLOC-2976


jongiddy:
Looks good. Test fails without patch, passes with it.  Please merge when green.

PR #1917, opened 2015-09-08
[FLOC-2985] Remove duplication in block device test setup

jongiddy:
The block device test code duplicates code for setting up `IBlockDeviceAPI` instances. In `master`, a good example is https://github.com/ClusterHQ/flocker/blob/9ebddf05cf6fc68dbd0f1af5a1d983efa232d716/flocker/node/agents/test/blockdevicefactory.py#L181-L189 which duplicates https://github.com/ClusterHQ/flocker/blob/9ebddf05cf6fc68dbd0f1af5a1d983efa232d716/flocker/node/agents/cinder.py#L773-L782.

This PR rewrites this code to reduce this repetition. The main change is that `get_blockdeviceapi_args`, which returned a factory function and a dictionary to apply to that function, is replaced by a simpler `get_blockdevice_config`, which just returns the configuration stanza relevant to the tested platform.  This means the caller needs to do a bit more work (see the change to the `get_blockdeviceapi` function, for example), but the work that is getting done is more explicit.

For the Openstack clients in particular, the tests now use utility functions to obtain the (Cinder and Nova) clients that they need. Again, while more verbose, this is more clear than the previous code, which created a dictionary containing all clients, and then extracted the required clients from the dictionary.

Where previously, configuration values that needed to be overridden for a test were passed as parameters to `get_blockdeviceapi_args`, they can now be set explicitly in the test, making the changes more explicit. Compare https://github.com/ClusterHQ/flocker/blob/9ebddf05cf6fc68dbd0f1af5a1d983efa232d716/flocker/node/agents/functional/test_cinder.py#L170-L176 in `master` with https://github.com/ClusterHQ/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/functional/test_cinder.py#L167-L179 in this branch.

There are some changes to the AWS code, as well, to keep the changes consistent, but the changes for AWS were kept as minimal as possible.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1917)

### DIFF
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -764,12 +764,12 @@ def get_nova_client(session, region):
 
 def cinder_from_configuration(region, cluster_id, **config):
     """
-    Build a ``CinderBlockDeviceAPI`` using configuration and credentials in
-    ``config``.
+    Build a ``CinderBlockDeviceAPI`` using configuration and credentials
+    in ``config``.
 
-    :param str region: The region "slug" for which to configure the object.
-    :param cluster_id: The unique cluster identifier for which to configure the
-        object.
+    :param str region: The Openstack region to access.
+    :param cluster_id: The unique identifier for the cluster to access.
+    :param config: A dictionary of configuration options for Openstack.
     """
     session = get_keystone_session(**config)
     cinder_client = get_cinder_client(session, region)
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -734,15 +734,12 @@ def _openstack_verify_from_config(
 
 def get_keystone_session(**config):
     """
-    Create Cinder and Nova volume managers suitable for use in the creation of
-    a ``CinderBlockDeviceAPI``.  They will be configured to use the region
-    where the server that is running this code is running.
+    Create a Keystone session from a configuration stanza.
 
-    :param config: Any additional configuration (possibly provider-specific)
-        necessary to authenticate a session for use with the CinderClient and
-        NovaClient.
+    :param dict config: Configuration necessary to authenticate a
+        session for use with the CinderClient and NovaClient.
 
-    :return: A ``dict`` of keyword arguments for ``cinder_api``.
+    :return: keystoneclient.Session
     """
     return Session(
         auth=_openstack_auth_from_config(**config),
@@ -750,13 +747,27 @@ def get_keystone_session(**config):
         )
 
 
-def get_cinder_client(session, region):
+def get_cinder_v1_client(session, region):
+    """
+    Create a Cinder (volume) client from a Keystone session.
+
+    :param keystoneclient.Session session: Authenticated Keystone session.
+    :param str region: Openstack region.
+    :return: A cinderclient.Client
+    """
     return CinderClient(
         session=session, region_name=region, version=1
     )
 
 
-def get_nova_client(session, region):
+def get_nova_v2_client(session, region):
+    """
+    Create a Nova (compute) client from a Keystone session.
+
+    :param keystoneclient.Session session: Authenticated Keystone session.
+    :param str region: Openstack region.
+    :return: A novaclient.Client
+    """
     return NovaClient(
         session=session, region_name=region, version=2
     )
@@ -772,8 +783,8 @@ def cinder_from_configuration(region, cluster_id, **config):
     :param config: A dictionary of configuration options for Openstack.
     """
     session = get_keystone_session(**config)
-    cinder_client = get_cinder_client(session, region)
-    nova_client = get_nova_client(session, region)
+    cinder_client = get_cinder_v1_client(session, region)
+    nova_client = get_nova_v2_client(session, region)
 
     logging_cinder = _LoggingCinderVolumeManager(
         _cinder_volumes=cinder_client.volumes
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -661,35 +661,6 @@ class _LoggingNovaServerManager(PRecord):
     _nova_servers = field(mandatory=True)
 
 
-def cinder_api(cinder_client, nova_client, cluster_id):
-    """
-    :param cinderclient.v1.client.Client cinder_client: The Cinder API client
-        whose ``volumes`` attribute will be supplied as the
-        ``cinder_volume_manager`` parameter of ``CinderBlockDeviceAPI``.
-    :param novaclient.v2.client.Client nova_client: The Nova API client whose
-        ``volumes`` attribute will be supplied as the ``nova_volume_manager``
-        parameter of ``CinderBlockDeviceAPI``.
-    :param UUID cluster_id: A Flocker cluster ID.
-
-    :returns: A ``CinderBlockDeviceAPI``.
-    """
-    logging_cinder = _LoggingCinderVolumeManager(
-        _cinder_volumes=cinder_client.volumes
-    )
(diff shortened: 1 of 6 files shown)


tomprince:
<sup>**[flocker/node/agents/test/blockdevicefactory.py, line 149 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj4WONuq0txxVTYQnD)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/test/blockdevicefactory.py#L149)):</sup>
It isn't clear to me why this is a separate function, when the corresponding code for AWS is just part of 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1917)

tomprince:
The changes here are a definite improvement over the existing code, so can be merged as is. However, there are a couple of things that jump out at me.

- Although this is intended to get rid of dupication of code, it does look like it really just shuffles around which code is duplicated and where.
- The AWS code also has the same duplication.
- The OpenStack and AWS code is strangely asymmetric.
   - At least part of this is because the interface to AWS and OpenStack are different (AWS has a single object to interact with, OpenStack appears to have three).
   - But some of it isn't: both AWS and OpenStack deal with getting the region to use from an environment variable.
     - AWS does this in a test-specific function to get the API client.
     - OpenStack has a function that does this that needs to be called from multiple places.
    
     I'm not sure how to resolve that asymmetry, though. One thought is to move the code dealing with the environment variable into `get_blockdevice_config`; this would probably involve making `get_<openstack>_client` take a config instead of a region, and extracting the bit it cares about out.

I don't have any specific suggestions on how to address any of that, so do with it what you will.

---

Reviewed 5 of 5 files at r1.

---

<sup>**[flocker/node/agents/cinder.py, line 791 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyig0P4aIUbyuB6oDuM)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/cinder.py#L791)):</sup>
:question: Looking at the code, it appears that wrapping with `_Logging*Manager` here means that usage in tests won't be logged. It seems like it would be useful to always log all the API interactions, which suggests moving the wrapping into the above functions.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 290 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj3J5exEloSZ4a_R9F)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/functional/test_cinder.py#L290)):</sup>
This block looks very much like the duplicated block that prompted this PR (and there are a couple of other occurrences above.

---

<sup>**[flocker/node/agents/test/blockdevicefactory.py, line 59 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj419F5RpRrrIV0rJP)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/test/blockdevicefactory.py#L59)):</sup>
This has nothing to do with the current PR, but it seems like `provider` and `provider_type` are backwards or something.

---

<sup>**[flocker/node/agents/test/blockdevicefactory.py, line 142 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj4I3S7TIdsE7VMkal)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/test/blockdevicefactory.py#L142)):</sup>
This comment isn't entirely clear.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1917)

jongiddy:
The number of lines didn't change much, but the shuffling around provides better hooks for testing. In production, we just want to go straight from config file to working BlockDeviceAPI. For testing, it is useful to have a hook after reading the config stanza but before using it (to allow modification of the config stanza), and to have a hook after creating the clients, but before creating the Flocker BlockDeviceAPI, to allow tests to perform supporting steps without using Flocker code.

I think you're right that the asymmetry between AWS and OpenStack is a problem, in terms of readability, and potential code reuse. I'm certainly open to another round to bring them closer together, but this was a low priority fix to solve a particular problem that hit me the last two times I touched this code. (I originally added the `override` parameter that was a messy way to change configuration options for testing purposes). 

Thanks for your comments.

---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/agents/cinder.py, line 791 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyig0P4aIUbyuB6oDuM)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/cinder.py#L791)):</sup>
I believe that the amount of logging is consistent before and after the patch.

In master, some tests in `functional/test_cinder.py` call `get_blockdeviceapi_args`and then extract the direct clients from the returned `kwargs` without passing it through `cinder_api`, and hence without logging.

In this branch, the same code takes a more direct route to get the non-logged clients. But any calls that eventually call `_openstack` get the logging added as before.

But, yes, it may be good to log all calls, whether through the BlockDeviceAPI, or directly to the nova/cinder modules.

---

<sup>**[flocker/node/agents/functional/test_cinder.py, line 290 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj3J5exEloSZ4a_R9F)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/functional/test_cinder.py#L290)):</sup>
The difference between this PR's repetition and the repetition in master is that this PR's repetition provides the right hooks for testing. See my response to your overall comment.

I thought about having a function to squash these multiple statements into a single function call, but you'll notice these repeated sections usually do something different at these hook points.

Also, I'm just not as bothered by some repetition in the individual tests as by repetition of code in two different files, one of which is production code and one of which is a support module for the tests.  I'd rather have production and tests use the same code path.

---

<sup>**[flocker/node/agents/test/blockdevicefactory.py, line 142 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj4I3S7TIdsE7VMkal)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/test/blockdevicefactory.py#L142)):</sup>
Rewrote to be more detailed.

---

<sup>**[flocker/node/agents/test/blockdevicefactory.py, line 149 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1917#-Jyj4WONuq0txxVTYQnD)** ([raw file](https://github.com/clusterhq/flocker/blob/d8b67278ca90f48ae9cb7f412518eac03f392c89/flocker/node/agents/test/blockdevicefactory.py#L149)):</sup>
It is separate for OpenStack because the region needs to be known both in `_openstack` which creates a BlockDeviceAPI, and in some tests which require the direct client module.  It isn't currently required for AWS because I only changed the AWS module to keep it working, not to refactor in any way.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1917)

PR #1919, opened 2015-09-08
[FLOC-2947] Docker plugin persist restarts

dwgebler:
Fixes FLOC-2947

### DIFF
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -25,12 +25,13 @@ class DockerPluginTests(TestCase):
     """
     Tests for the Docker plugin.
     """
-    def restart_docker(self, address):
+    def docker_service(self, address, action):
         """
         Restart the Docker daemon on the specified address.
 
         :param bytes address: The public IP of the node on which Docker will
             be restarted.
+        :param bytes action: The action to perform on the service.
         """
         distro = []
         get_distro = run_ssh(
@@ -38,22 +39,23 @@ class DockerPluginTests(TestCase):
             handle_stdout=distro.append)
         get_distro.addCallback(lambda _: distro[0].lower())
 
-        def restart_docker(distribution):
+        def action_docker(distribution):
             if 'ubuntu' in distribution:
-                command = ["service", "docker", "restart"]
+                command = ["service", "docker", action]
             else:
-                command = ["systemctl", "restart", "docker"]
+                command = ["systemctl", action, "docker"]
             d = run_ssh(reactor, b"root", address, command)
 
             def handle_error(_):
                 self.fail(
-                    "Restarting Docker failed. See logs for process output.")
+                    "Docker {} failed. See logs for process output.".format(
+                        action))
 
             d.addErrback(handle_error)
             return d
 
-        restarting = get_distro.addCallback(restart_docker)
-        return restarting
+        acting = get_distro.addCallback(action_docker)
+        return acting
 
     def run_python_container(self, cluster, address, docker_arguments, script,
                              script_arguments, cleanup=True):
@@ -123,8 +125,20 @@ class DockerPluginTests(TestCase):
         # assert the data has been written
         d.addCallback(lambda _: assert_http_server(
             self, node.public_address, host_port, expected_response=data))
-        # restart the Docker daemon
-        d.addCallback(lambda _: self.restart_docker(node.public_address))
+        # stop the Docker daemon
+        d.addCallback(lambda _: self.docker_service(
+            node.public_address, b"stop"))
+
+        # ensure the container HTTP service has stopped
+
+        def poll_http_server_stopped(_):
+            ds = verify_socket(node.public_address, host_port, closed=True)
+            return ds
+
+        d.addCallback(poll_http_server_stopped)
+        # start Docker daemon
+        d.addCallback(lambda _: self.docker_service(
+            node.public_address, b"start"))
         # attempt to read the data back again; the container should've
         # restarted automatically, though it may take a few seconds
         # after the Docker daemon has restarted.
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -63,9 +63,7 @@ class DockerPluginTests(TestCase):
                 command = ["service", "docker", "restart"]
             else:
                 command = ["systemctl", "restart", "docker"]
-            #import pdb;pdb.set_trace()
-            zx=run_ssh_command(b"root", address, command)
-            #import pdb;pdb.set_trace()
+            run_ssh_command(b"root", address, command)
         except SSHError as e:
             self.fail("Restart docker failed: " + e)
 
@@ -124,20 +122,19 @@ class DockerPluginTests(TestCase):
             cluster, node.public_address,
             {"host_config": create_host_config(
                 binds=["{}:/data".format(volume_name)],
-                port_bindings={"8080/tcp": [{"HostPort": "8080"}]}),
-                #port_bindings={http_port: http_port}),
+                port_bindings={http_port: http_port}),
              "ports": [http_port]},
             SCRIPTS.child(b"datahttp.py"),
             # This tells the script where it should store its data,
             # and we want it to specifically use the volume:
-            [u"/data"], "always")
+            [u"/data"], restart_policy="always")
 
         # write some data to it via POST
         d = post_http_server(self, node.public_address, http_port,
                              {"data": data})
         # assert the data has been written
-        #d.addCallback(lambda _: assert_http_server(
-        #    self, node.public_address, http_port, expected_response=data))
+        d.addCallback(lambda _: assert_http_server(
+            self, node.public_address, http_port, expected_response=data))
         # restart the Docker daemon
         d.addCallback(lambda _: self.restart_docker(node.public_address))
         # attempt to read the data back again; the container should've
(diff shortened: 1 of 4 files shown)


flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -48,6 +48,25 @@ def get_path(name):
...
+
+        :param bytes address: The public IP of the node on which Docker will
+            be restarted.
+        """
+        try:
+            distro = run_ssh_command(
+                b"root", address, ["python", "-m", "platform"])
+            distro = distro.output.lower()
+            if 'ubuntu' in distro:
+                command = ["service", "docker", "restart"]
```
dwgebler: `service docker restart` might work in all cases, simply going via `systemctl` on CentOS and other non-Ubuntu platforms, but I'm not 100% on that...

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -48,6 +50,25 @@ def get_path(name):
...
+            else:
+                command = ["systemctl", "restart", "docker"]
+            run_ssh_command(b"root", address, command)
```
dwgebler: As far as I know, we don't have root SSH for our cloud instances, but the only ways I can think of to restart the Docker daemon on a node is either SSH or an interface to restart the machine. I'd be glad to know someone else's thoughts on what best to do here - we may want to do something like set the SSH username (and perhaps a key path) in the environment..

flocker/acceptance/endtoend/test_dockerplugin.py
jml: Another approach would be to try one and then if it fails with "command not found" try the other. The most important thing is if _neither_ command works, we get a good error message and a failed test.

flocker/acceptance/endtoend/test_dockerplugin.py
jml: I don't have any good ideas here. I would have thought that however we connect to the machines, we would be able to use `sudo`.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -57,6 +78,46 @@ def run_python_container(self, cluster, address, docker_arguments, script,
...
+        d = post_http_server(self, node.public_address, host_port,
+                             {"data": data})
+        # assert the data has been written
+        d.addCallback(lambda _: assert_http_server(
+            self, node.public_address, host_port, expected_response=data))
+        # restart the Docker daemon
+        d.addCallback(lambda _: self.restart_docker(node.public_address))
+        # attempt to read the data back again; the container should've
+        # restarted automatically.
+        d.addCallback(lambda _: assert_http_server(
```
jml: There's a race condition here. The docker service might say that it is restarted before the container has fired up. The right thing to do is poll a few times with a timeout.

flocker/testtools/__init__.py
```diff
@@ -1074,6 +1074,30 @@ def run_process(command, *args, **kwargs):
...
+    :param list command: A list containing the command to execute and
+        any arguments.
+    :return A ``_ProcessResult`` instance describing the result of the child
+         command.
+    """
+    try:
+        return run_process(
+            [b'ssh', b'{}@{}'.format(user, host)] + command)
+    except _CalledProcessError as e:
+        raise SSHError(e.returncode, e.cmd, e.output)
```
jml: Probably should be using `flocker.common.runner.run_ssh`

jml:
Overall design is fine. I've made some comments that might help with the implementation.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -20,6 +25,36 @@ class DockerPluginTests(TestCase):
...
+    def restart_docker(self, address):
+        """
+        Restart the Docker daemon on the specified address.
+
+        :param bytes address: The public IP of the node on which Docker will
+            be restarted.
+        """
+        distro = []
+        get_distro = run_ssh(
+            reactor, b"root", address, ["python", "-m", "platform"],
```
wallrj: Huh. `python -m platform`. That's cool.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -20,6 +25,36 @@ class DockerPluginTests(TestCase):
...
+            handle_stdout=distro.append)
+        get_distro.addCallback(lambda _: distro[0].lower())
+
+        def restart_docker(distribution):
+            if 'ubuntu' in distribution:
+                command = ["service", "docker", "restart"]
+            else:
+                command = ["systemctl", "restart", "docker"]
```
wallrj: Although the `service` command seems to be available in Centos-7 where it redirects to systemctl.
```
[root@centos-flocker-node1 ~]# service docker restart
Restarting docker (via systemctl):                         [  OK  ]
```

I'd be inclined to use that instead of detecting the distro.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -20,6 +25,36 @@ class DockerPluginTests(TestCase):
...
+            d = run_ssh(reactor, b"root", address, command)
+
+            def handle_error(_):
+                self.fail(
+                    "Restarting Docker failed. See logs for process output.")
```
wallrj: :question: Isn't there anything useful in the supplied `Failure` (`_`) which would make debugging easier?

dwgebler: Unfortunately not, it's a `ProcessTerminatedError` that can give you the exit code, but not any output.

flocker/acceptance/endtoend/test_dockerplugin.py
dwgebler: Yeah, so it turns out on an out-of-the-box CentOS 7, `service` is aliased to `systemctl`, but it seems that the preferred convention is nonetheless to use `systemctl` directly - we know this command will always be available on CentOS 7, whereas strictly, the availability of `service` is probable but not guaranteed.

wallrj:
I just wanted to see the restart behaviour for myself. The container was restarted.

```
$ docker run --detach --restart=always --publish 8000:8000 python:2.7-slim python -m SimpleHTTPServer
f1e7434239f3f60224acc859f887f62b88839b73f61b6aa1c2c30d26982337bf
$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS                    NAMES
f1e7434239f3        python:2.7-slim     "python -m SimpleHTT   3 seconds ago       Up 2 seconds        0.0.0.0:8000->8000/tcp   cranky_poitras      
$ sudo service docker restart
Redirecting to /bin/systemctl restart  docker.service
(devstack-acceptance-test-runner-FLOC-3017 ? =)[~/.../HybridLogic/flocker]$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS                    NAMES
f1e7434239f3        python:2.7-slim     "python -m SimpleHTT   29 seconds ago      Up 2 seconds        0.0.0.0:8000->8000/tcp   cranky_poitras      
```

wallrj:
Thanks @dwgebler 

Looks good.

What happens if:
 * the `service docker restart` command returns before docker has actually restarted (not sure if that's ever going to happen), or
 * if it doesn't actually restart the docker daemon (broken init script or it's doing some sort of graceful restart), or
 * if docker leaves behind running container processes when the daemon exits (bugs such as https://github.com/docker/docker/issues/4766 for example) ?

There's a `--full-restart` option which forces a stop and a start

```
$ man service
...
As a special case, if COMMAND is --full-restart, the script is run twice, first with the stop command, then with the start
       command.
```

But I think it'd be worth verifying that the container does really stop by:
 * stopping the docker daemon, 
 * looping until the HTTP server stops responding.
 * starting the docker daemon
 * looping until the HTTP server re-appears.

What do you think?

Do that if you agree it's worthwhile. Otherwise merge when green.

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -57,6 +100,70 @@ def run_python_container(self, cluster, address, docker_arguments, script,
...
+                ds = check_http_server(node.public_address, host_port)
+                ds.addCallback(lambda succeeded: not succeeded)
+                return ds
+            looping = loop_until(http_closed)
+            return looping
+
+        # d.addCallback(poll_http_server_stopped)
+        # start Docker daemon
+        # d.addCallback(lambda _: self.docker_service(
+        #     node.public_address, b"start"))
```
dwgebler: > But I think it'd be worth verifying that the container does really stop by:
> 
> stopping the docker daemon,
> looping until the HTTP server stops responding.
> starting the docker daemon
> looping until the HTTP server re-appears.

I thought this was worth doing and made some applicable changes and this may have uncovered a bug either in Docker, Flocker or our plugin - I am yet to determine where exactly the problem lies.

If you change the "restart" command a few lines up to "stop", then uncomment this block, you'll have a version of the test that works in the suggested way. This test will pass in that manner, however on Buildbot, this will consistently cause numerous test failures in the acceptance tests that are run *after* this one passes. There are no build errors if I revert to just running a "restart" command as the branch does here and as it did originally when reviewed.

I've tried various changes but it seems to specifically be the use of "stop" and "start" against the Docker daemon that's causing the build failures.

dwgebler: The container agent is stopping when Docker stops, but `systemd` isn't restarting it. One solution that strikes me as reasonable is to use `Wants=docker.service` in the unit file instead of `Requires=`.

PR #1920, opened 2015-09-08
[FLOC-1657] Traige intermittent test_pull_image_if_necessary failure

myechuri:
Ran failing test 200 times on Ubuntu 14.04 and 100 times on CentOS 7, could not reproduce failure.

Have a theory that Docker server could be returning cached content to clients:
1) Client1 establishes connection
2) Client2 starts registry, adds image
3) Client1 attempts to ``inspect_image`` -> test failure sees state before event (2).
Have a small change where (3) is replaced with:
Client3 establishes new connection, then attempts to ``inspect_image``.

If we see a reproduction on buildbot with this change, this theory can be invalidated. Since this change does not add much overhead to tests, hope it is worth the experiment.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1920)


jml:
Sorry this languished so long. For stupid reasons it wasn't appearing on the kanban board. I've triggered a re-run of the tests. 

Please feel free to merge when the tests pass.

exarkun:
> Please feel free to merge when the tests pass.

Is this actually worth merging?  It's not clear the change actually fixes the problem, even if the problem is correctly diagnosed.  What are Docker's consistency guarantees across different client connections?  I strongly suspect there are none.

The experiment seems fine to run but just merging this into master won't accomplish that.  Someone actually needs to pay attention to the results - and the results can be gathered about as easily without merging this change.

exarkun:
On Sep 1, test_pull_image_if_necessary-refactor-FLOC-2915 was merged.  It changed this test so that it no longer uses the busybox image directly and no longer tries to remove the busybox image.  This was done as a way to avoid having to pull any images from Dockerhub (which has reliability issues of its own).  A side-effect may have been to fix FLOC-1657, though.  On 448 runs of this test (with some concurrency and while creating, starting, stopping, and removing some busybox containers) on master@HEAD (2832748f93e7bf8c5383139dacb867603ecabc79), I didn't see this test fail.

On top of this, if I run the test on 27005333683036407894333df6a2dd909d253383 then it fails as soon as I concurrently have a running a busybox container.

So I think we can say FLOC-1657 most likely fixed this problem.

PR #1921, opened 2015-09-09
 [FLOC-2991] Only use virtual disk serial number if the virtio_blk driver is detected

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2991

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -178,6 +178,19 @@ require_moving_backend = skip_backend(
     reason="doesn't support moving")
 
 
+def get_default_volume_size():
+    """
+    :returns int: the default volume size (in bytes) supported by the
+        backend the acceptance tests are using.
+    """
+    default_volume_size = environ.get("FLOCKER_ACCEPTANCE_DEFAULT_VOLUME_SIZE")
+    if default_volume_size is None:
+        raise SkipTest(
+            "Set acceptance testing default volume size using the " +
+            "FLOCKER_ACCEPTANCE_DEFAULT_VOLUME_SIZE environment variable.")
+    return int(default_volume_size)
+
+
 def get_mongo_client(host, port=27017):
     """
     Returns a ``Deferred`` which fires with a ``MongoClient`` when one has been
@@ -727,7 +740,7 @@ def create_python_container(test_case, cluster, parameters, script,
 
 
 def create_dataset(test_case, cluster,
-                   maximum_size=REALISTIC_BLOCKDEVICE_SIZE):
+                   maximum_size=None):
     """
     Create a dataset on a cluster (on its first node, specifically).
 
@@ -738,6 +751,9 @@ def create_dataset(test_case, cluster,
     :return: ``Deferred`` firing with a ``flocker.apiclient.Dataset``
         dataset is present in actual cluster state.
     """
+    if maximum_size is None:
+        maximum_size = get_default_volume_size()
+
     configuring_dataset = cluster.client.create_dataset(
         cluster.nodes[0].uuid, maximum_size=maximum_size, dataset_id=uuid4(),
         metadata={u"name": u"my_volume"})
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -34,7 +34,7 @@ from ..common import gather_deferreds
 
 from ..control.httpapi import REST_API_PORT
 from ..ca import treq_with_authentication
-from ..testtools import loop_until, random_name, REALISTIC_BLOCKDEVICE_SIZE
+from ..testtools import loop_until, random_name
 from ..apiclient import FlockerClient, DatasetState
 
 try:
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -583,7 +583,7 @@ class CinderBlockDeviceAPI(object):
             "/dev/disk/by-id/virtio-{}".format(volume.id[:20])
         )
         if expected_path.exists():
-            return expected_path
+            return expected_path.realpath()
         else:
             raise UnattachedVolume(volume.id)
 
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -571,7 +571,7 @@ class CinderBlockDeviceAPI(object):
                 break
             time.sleep(1.0)
 
-    def get_device_path(self, blockdevice_id):
+    def _get_device_path_virtioblk(self, volume):
         # libvirt does not return the correct device path when additional
         # disks have been attached using a client other than cinder. This is
         # expected behaviour within Cinder and libvirt
@@ -579,29 +579,42 @@ class CinderBlockDeviceAPI(object):
         # http://libvirt.org/formatdomain.html#elementsDisks (target section)
         # However, the correct device is named as a udev symlink which includes
         # the first 20 characters of the blockedevice_id.
-        device_path = FilePath(
-            "/dev/disk/by-id/virtio-{}".format(blockdevice_id[:20]))
-        if not device_path.exists():
-            # If the device path does not exist, either virtio driver is
-            # not being used (e.g. Rackspace), or the user has modified
-            # their udev rules.  The following code relies on Cinder
-            # returning the correct device path, which appears to work
-            # for Rackspace and will work with virtio if no disks have
-            # been attached outside Cinder.
-            try:
-                cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
-            except CinderNotFound:
-                raise UnknownVolume(blockdevice_id)
+        expected_path = FilePath(
+            "/dev/disk/by-id/virtio-{}".format(volume.id[:20])
+        )
+        if expected_path.exists():
+            return expected_path
+        else:
+            raise UnattachedVolume(volume.id)
 
-            # As far as we know you can not have more than one attachment,
-            # but, perhaps we're wrong and there should be a test for the
-            # multiple attachment case.  FLOC-1854.
-            try:
-                [attachment] = cinder_volume.attachments
-            except ValueError:
-                raise UnattachedVolume(blockdevice_id)
+    def _get_device_path_other(self, volume):
+        # If the device path does not exist, either virtio driver is
+        # not being used (e.g. Rackspace), or the user has modified
+        # their udev rules.  The following code relies on Cinder
+        # returning the correct device path, which appears to work
+        # for Rackspace and will work with virtio if no disks have
+        # been attached outside Cinder.
 
-            device_path = FilePath(attachment['device'])
+        # As far as we know you can not have more than one attachment,
+        # but, perhaps we're wrong and there should be a test for the
+        # multiple attachment case.  FLOC-1854.
+        try:
+            [attachment] = volume.attachments
+        except ValueError:
+            raise UnattachedVolume(volume.id)
+
+        return FilePath(attachment['device'])
+
+    def get_device_path(self, blockdevice_id):
+        try:
+            cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
+        except CinderNotFound:
+            raise UnknownVolume(blockdevice_id)
+
+        if FilePath('/sys/bus/virtio/drivers/virtio_blk').isdir():
+            device_path = self._get_device_path_virtioblk(cinder_volume)
+        else:
+            device_path = self._get_device_path_other(cinder_volume)
 
         # It could be attached somewhere else...
         # https://clusterhq.atlassian.net/browse/FLOC-1830
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -573,16 +573,17 @@ class CinderBlockDeviceAPI(object):
 
     def _get_device_path_virtio_blk(self, volume):
         """
-        The virtio_blk driver allows a serial number to be assigned to virtual blockdevices.
+        The virtio_blk driver allows a serial number to be assigned to virtual
+        blockdevices.
         OpenStack will set a serial number containing the first 20
         characters of the Cinder block device ID.
         The udev daemon will read the serial number and create a
         symlink to the canonical virtio_blk device path.
 
-        We do this because libvirt does not return the correct device path when additional
-        disks have been attached using a client other than cinder. This is
-        expected behaviour within Cinder and libvirt
-        See https://bugs.launchpad.net/cinder/+bug/1387945 and
+        We do this because libvirt does not return the correct device path when
+        additional disks have been attached using a client other than
+        cinder. This is expected behaviour within Cinder and libvirt See
+        https://bugs.launchpad.net/cinder/+bug/1387945 and
         http://libvirt.org/formatdomain.html#elementsDisks (target section)
 
         :param volume: The Cinder ``Volume`` which is attached.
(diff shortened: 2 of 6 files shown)


flocker/node/agents/cinder.py
```diff
@@ -579,29 +579,42 @@ def get_device_path(self, blockdevice_id):
...
+        else:
+            raise UnattachedVolume(volume.id)
+
+    def _get_device_path_other(self, volume):
+        # If the device path does not exist, either virtio driver is
+        # not being used (e.g. Rackspace), or the user has modified
+        # their udev rules.  The following code relies on Cinder
+        # returning the correct device path, which appears to work
+        # for Rackspace and will work with virtio if no disks have
+        # been attached outside Cinder.
```
jongiddy: This function should now never be run when the virtio driver is being used, so the comment does not need to refer to cases where virtio would work using this code.

flocker/node/agents/cinder.py
```diff
@@ -579,29 +579,42 @@ def get_device_path(self, blockdevice_id):
...
+        return FilePath(attachment['device'])
 
-            device_path = FilePath(attachment['device'])
+    def get_device_path(self, blockdevice_id):
+        try:
+            cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
+        except CinderNotFound:
+            raise UnknownVolume(blockdevice_id)
+
+        if FilePath('/sys/bus/virtio/drivers/virtio_blk').isdir():
```
jongiddy: :question: wrap this in a simple `is_virtio` function to clarify the decision being made.  The docstring for this simple function should explain how the existence of the this directory shows which driver is in operation.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -415,3 +419,117 @@ def test_disk_attachment_fails_with_conflicting_disk(self):
...
+        self.addCleanup(self._cleanup, instance_id, cinder_volume)
+        volume = wait_for_volume_state(
+            volume_manager=self.cinder.volumes, expected_volume=cinder_volume,
+            desired_state=u'available', transient_states=(u'creating',))
+
+        # Suspend udevd before attaching the disk
+        [udev_process] = [
+            p for p in psutil.process_iter()
+            if p.name().endswith('-udevd')
+        ]
```
jongiddy: :question: If you use a real for loop here, you could `break` when the process is found.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -415,3 +419,117 @@ def test_disk_attachment_fails_with_conflicting_disk(self):
...
+        volume.get()
+        if volume.attachments:
+            self._detach(instance_id, volume)
+        self.cinder.volumes.delete(volume.id)
+
+    @require_virsh
+    def test_getdevicepath_error_without_udev(self):
+        """
+        ``getdevicepath`` on systems using the virtioblk driver raises
+        ``TimeoutException`` if ``/dev/disks/by-id/xxx`` does not appear within
```
jongiddy: Exception is `UnattachedVolume` and there is no 5 second timeout.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -415,3 +419,117 @@ def test_disk_attachment_fails_with_conflicting_disk(self):
...
+            device=None,
+        )
+        volume = wait_for_volume_state(
+            volume_manager=self.cinder.volumes,
+            expected_volume=attached_volume,
+            desired_state=u'in-use',
+            transient_states=(u'attaching',),
+        )
+        udev_process.resume()
+        sleep(5)
```
jongiddy: Not clear why the udevd suspend and 5 second wait are required for this test. If it is required, please add more info to the docstring why we're testing this under these circumstances.

jongiddy:
To answer your points above:
 * I think you should solve this for the OpenStack platforms to which we have access (Rackspace, devstack, and cust0), but add code to check expectations and raise exceptions if those expectations aren't met.  There is a separate question of how Flocker could better propagate errors, without entering a never-converging loop, but that is separate to this issue.
 * Sounds like you should raise a different exception to `UnattachedVolume`, to distinguish between complete disaster, and possibly just a matter of trying again later. A new exception class is fairly cheap.
 * I prefer the immediate and specific exception to a timeout. Lots of questions raised with a timeout - how long to wait, what does it mean (was it a permanent failure or just a too-short timeout)?

jongiddy:
It sounds like you're planning to make further changes to this branch.  As well as the above, there are some lint errors. Also, be aware that FLOC-2985 / #1917 may merge into `master` before this branch, which will require that you replace the test `setUp` with slightly different code (should be similar to the previous test class).

flocker/node/agents/functional/test_cinder.py
adamtheturtle: Can one use a set and `.next()`?

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/functional/test_cinder.py
wallrj: I won't do this. There's a chance that there are multiple processes ending in 'udev' so and we want the test to blow up in that case.

flocker/node/agents/functional/test_cinder.py

flocker/node/agents/functional/test_cinder.py
wallrj: Yeah, the udev manipulation isn't necessary here. We removed it.

flocker/node/agents/cinder.py
```diff
@@ -571,43 +571,89 @@ def destroy_volume(self, blockdevice_id):
...
+                Message.new(
+                    message_type=(
+                        u'flocker:node:agents:blockdevice:openstack:get_device_path:'
+                        u'unexpected_multiple_attachments'
+                    ),
+                    volume_id=unicode(volume.id),
+                    attachment_devices=u','.join(
+                        unicode(a['device']) for a in volume.attachments
+                    ),
+                ).write()
```
wallrj: This isn't really related to this branch, but I noticed that in catching ``ValueError`` we weren't distinguishing between the no attachements and the multiple attachments situations. I'm now logging the multiple attachments situation in case it ever occurs.

wallrj:
Ready for another review.

Things to discuss:
 * Just because we're on an OpenStack that uses virtio_blk, doesn't mean that the devices will have the expected serial numbers. Find out when OpenStack / libvirt / (whatever) started doing this.
 * There may be other virtual block device drivers that populate the serial number. Do Xen Virtual Disks ever have  a serial number? Not on Rackspace, but perhaps on a recent devstack using the xen hypervisor.

I haven't done looked into these yet but @jongiddy replied: 
> I think you should solve this for the OpenStack platforms to which we have access (Rackspace, devstack, and cust0), but add code to check expectations and raise exceptions if those expectations aren't met. There is a separate question of how Flocker could better propagate errors, without entering a never-converging loop, but that is separate to this issue.

 * In this solution, get_device_path raises `UnattachedVolume` if the `/dev/disk/by-id/xxx` mapping is missing. How will the dataset agent loop react to that? Until now, `UnattachedVolume` has been an unexpected error...indicating that our state change logic is wrong. Now it will be a routine error on some OpenStack platforms and may occur every time a volume is attached until udevd reacts to the disk attachment event.

@jongiddy replied:
> Sounds like you should raise a different exception to UnattachedVolume, to distinguish between complete disaster, and possibly just a matter of trying again later. A new exception class is fairly cheap.

True, but a new exception would need to be added to the `IBlockDeviceAPI` and it seems reasonable that delays in udev creating the device symlinks should be treated as a device not being completely attached.

 * Alternatively, we discussed modifying `get_device_path` to block until the `/dev/disk/by-id/xxx` appeared and raising `TimeoutError` if it doesn't appear within `X` seconds (5 seconds). But that makes it difficult to write tests without adding unreliable sleeps to the tests.

@jongiddy replied:
> I prefer the immediate and specific exception to a timeout. Lots of questions raised with a timeout - how long to wait, what does it mean (was it a permanent failure or just a too-short timeout)?

Finally, it's important to note that my / our original diagnosis was wrong. We originally thought that  ``flocker-dataset-agent`` was mounting volumes before udev had populated the `by-id` symlinks.
That is certainly a possibility and our new tests demonstrate that.

But the main problem was that although we supply `/dev/disk/by-id/xxx` to `mount`; the mounted filesystems are listed (in eg `/proc/mounts`) with the target of the ``/dev/disk/by-id/xxx`` symlink.
Which means that ``flocker-dataset-agent`` was mis-calculating that the attached volume had not been mounted and attempted to mount the device repeatedly.

Running the `flocker.acceptance.endtoend.test_dataset` tests on our devstack environment demonstrated that clearly.
Re-running those tests after against the code in this branch demonstrates that that problem is fixed.

There's a `tmux` session running on our devstack `10.0.0.3` node where you can run the acceptance tests against both ubuntu and centos nodes.

jml:
Thanks. Code, tests, and comments all look good. I'll do a second pass where I try to grok the higher-level concerns addressed in your last comment (or maybe we should have a call, or maybe @jongiddy should do it).

jml:
> Just because we're on an OpenStack that uses virtio_blk, doesn't mean that the devices will have the expected serial numbers. Find out when OpenStack / libvirt / (whatever) started doing this.

What would we do with that information once we found out?

> There may be other virtual block device drivers that populate the serial number. Do Xen Virtual Disks ever have a serial number? Not on Rackspace, but perhaps on a recent devstack using the xen hypervisor.

Does that mean that this patch might cause flocker to misbehave on other systems because it assumes things about udev, or does it mean that this patch addresses a problem on one system that might also need addressing on others? Or something else entirely?

> Raise `UnattachedVolume` or something else

I don't have a good opinion about this. On the one hand, I don't see a problem with adding a new documented exception to `IBlockDeviceAPI`. On the other, until it has behavioural consequences, trying to distinguish between "keep looping" and "something is wrong and requires intervention" exceptions is going to be an uphill battle.

flocker/acceptance/testtools.py
```diff
@@ -727,7 +740,7 @@ def created(response):
 
 
 def create_dataset(test_case, cluster,
-                   maximum_size=REALISTIC_BLOCKDEVICE_SIZE):
+                   maximum_size=None):
```
jml: Therefore, need to remove `REALISTIC_BLOCKDEVICE_SIZE` from imports

flocker/node/agents/cinder.py
```diff
@@ -571,43 +571,89 @@ def destroy_volume(self, blockdevice_id):
...
-            # returning the correct device path, which appears to work
-            # for Rackspace and will work with virtio if no disks have
-            # been attached outside Cinder.
-            try:
-                cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
-            except CinderNotFound:
-                raise UnknownVolume(blockdevice_id)
+    def _get_device_path_virtio_blk(self, volume):
+        """
+        The virtio_blk driver allows a serial number to be assigned to virtual blockdevices.
```
jml: Please wrap this line to <80 cols.

flocker/node/agents/cinder.py
```diff
@@ -571,43 +571,89 @@ def destroy_volume(self, blockdevice_id):
...
+        OpenStack will set a serial number containing the first 20
+        characters of the Cinder block device ID.
+        The udev daemon will read the serial number and create a
+        symlink to the canonical virtio_blk device path.
+
+        We do this because libvirt does not return the correct device path when additional
```
jml: Please wrap this line to <80 cols.

flocker/node/agents/cinder.py
```diff
@@ -571,43 +571,89 @@ def destroy_volume(self, blockdevice_id):
...
+        if volume.attachments:
+            attachment = volume.attachments[0]
+            if len(volume.attachments) > 1:
+                # As far as we know you can not have more than one attachment,
+                # but, perhaps we're wrong and there should be a test for the
+                # multiple attachment case.  FLOC-1854.
+                # Log a message if this ever happens.
+                Message.new(
+                    message_type=(
+                        u'flocker:node:agents:blockdevice:openstack:get_device_path:'
```
jml: Please wrap this line to <80 cols.

jml:
LGTM, barring lint changes. I have some questions that I'd like answered, but don't feel it's very useful for me to be blocking this.

flocker/acceptance/testtools.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

wallrj:
>> Just because we're on an OpenStack that uses virtio_blk, doesn't mean that the devices will have the expected serial numbers. Find out when OpenStack / libvirt / (whatever) started doing this.
> What would we do with that information once we found out?

If we knew which version of OpenStack introduced the the setting the disk serial number to match the Cinder volume UUID, we could document our minimum OpenStack version requirement.

@zubron  and I did some digging and found  the revision where OpenStack nova started populating the disk SERIAL number with the Cinder volume ID:
 * https://github.com/openstack/nova/commit/3a47c02c58cefed0e230190b4bcef14527c82709
 * https://bugs.launchpad.net/nova/+bug/1004328

That was committed in Aug 2012 so we probably don't need to worry too much about documenting minimum requirements.

wallrj:
>> There may be other virtual block device drivers that populate the serial number. Do Xen Virtual Disks ever have a serial number? Not on Rackspace, but perhaps on a recent devstack using the xen hypervisor.

> Does that mean that this patch might cause flocker to misbehave on other systems because it assumes things about udev, or does it mean that this patch addresses a problem on one system that might also need addressing on others? Or something else entirely?

The second. I was wondering if we could use the disk serial number on other OpenStack Cinder hypervisors / virtual disk drivers.

@zubron just double checked on a Rackspace VM (which uses Xen Hypervisoer and XVD driver); there's no sign of a disk serial number on that platform at least.

wallrj:
>> Raise `UnattachedVolume` or something else
> I don't have a good opinion about this. On the one hand, I don't see a problem with adding a new documented exception to `IBlockDeviceAPI`. On the other, until it has behavioural consequences, trying to distinguish between "keep looping" and "something is wrong and requires intervention" exceptions is going to be an uphill battle. 

``UnattachedVolume`` does fall under the "keep looping" category, I think.

``flocker-dataset-agent`` may calculate that a volume needs detaching and when it comes to detach, finds that the volume has already been detached by forces out side of Flocker.

On the other hand, our ``attach_volume`` and ``detach_volume`` implementations try hard to block until the volume has reached the desired state, so up until now we shouldn't see ``UnattachedVolume`` under normal circumstances.

Once this  branch lands we may see ``UnattachedVolume`` if udev is not updating the `/dev/disk/by-id/xxx` directory....but again it's not going to be common.

wallrj:
We've moved this work to a release maintenance branch in #1922.
And that will then be merged into master, so closing this.

PR #1922, opened 2015-09-10
[FLOC-3009] Only use virtual disk serial number if the virtio_blk driver is detected

wallrj:
A release maintenance branch for #1921 which has already been reviewed and approved.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -14,6 +14,7 @@ from zope.interface import Interface, implementer
 from characteristic import attributes
 from eliot import add_destination, write_failure
 from pyrsistent import pvector
+from bitmath import GiB
 
 from twisted.internet.error import ProcessTerminated
 from twisted.python.usage import Options, UsageError
@@ -95,6 +96,9 @@ def get_trial_environment(cluster):
             for node in cluster.agent_nodes
             if node.private_address is not None
         }),
+        'FLOCKER_ACCEPTANCE_DEFAULT_VOLUME_SIZE': bytes(
+            cluster.default_volume_size
+        ),
     }
 
 
@@ -335,11 +339,16 @@ def configured_cluster_for_nodes(
     :returns: A ``Deferred`` which fires with ``Cluster`` when it is
         configured.
     """
+    default_volume_size = GiB(1)
+    if dataset_backend_configuration.get('auth_plugin') == 'rackspace':
+        default_volume_size = GiB(100)
+
     cluster = Cluster(
         all_nodes=pvector(nodes),
         control_node=nodes[0],
         agent_nodes=nodes,
         dataset_backend=dataset_backend,
+        default_volume_size=int(default_volume_size.to_Byte().value),
         certificates=certificates
     )
 
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -472,41 +472,89 @@ class CinderBlockDeviceAPI(object):
                 break
             time.sleep(1.0)
 
+    def _get_device_path_virtio_blk(self, volume):
+        """
+        The virtio_blk driver allows a serial number to be assigned to virtual
+        blockdevices.
+        OpenStack will set a serial number containing the first 20
+        characters of the Cinder block device ID.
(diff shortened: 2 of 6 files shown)


jml:
:+1: YOLO

wallrj:
The rackspace/ubuntu acceptance test failures revealed a problem with our `_is_virtio_blk` function.

```
root@acceptance-test-richardw-0:~# ls -l /sys/bus/virtio/drivers/virtio_blk
total 0
--w------- 1 root root 4096 Sep 14 10:05 bind
--w------- 1 root root 4096 Sep 14 10:05 uevent
--w------- 1 root root 4096 Sep 14 10:05 unbind
```

We need to find another way to detect that the guest uses virtio_blk.

wallrj:
I merged forward to pick up @jongiddy 's zfs fix (#1929) and to trigger a new build.
Let's see if it goes green this time.

admin/acceptance.py
```diff
@@ -335,11 +339,16 @@ def configured_cluster_for_nodes(
     :returns: A ``Deferred`` which fires with ``Cluster`` when it is
         configured.
     """
+    default_volume_size = GiB(1)
+    if dataset_backend_configuration.get('auth_plugin') == 'rackspace':
+        default_volume_size = GiB(100)
```
jongiddy: This looks like it should come from `node.agents.test.blockdevicefactory.MINIMUM_ALLOCATABLE_SIZES`.  Or maybe they should both come from a non-test file.

flocker/node/agents/cinder.py
```diff
@@ -472,43 +472,103 @@ def destroy_volume(self, blockdevice_id):
...
+
+        :param volume: The Cinder ``Volume`` which is attached.
+        :returns: ``FilePath`` of the device created by the virtio_blk
+            driver.
+        """
+        expected_path = FilePath(
+            "/dev/disk/by-id/virtio-{}".format(volume.id[:20])
+        )
+        if expected_path.exists():
+            return expected_path.realpath()
```
jongiddy: Is there a reason you chose to return the real device path, rather than the symbolic link?  It would be good to document any reason, because the symbolic link has the advantage of being less ambiguous in the face of concurrent detaching/attaching activity.

flocker/node/agents/cinder.py
```diff
@@ -472,43 +472,103 @@ def destroy_volume(self, blockdevice_id):
...
-        # It could be attached somewhere else...
-        # https://clusterhq.atlassian.net/browse/FLOC-1830
         return device_path
 
 
+def _is_virtio_blk(device_path):
+    """
+    Check whether the supplied device path is a virtio_blk device.
+
+    XXX: We assume that virtio_blk device name always begin with `vd` where as
```
jongiddy: I don't see any action to take on this assumption, so remove the `XXX`.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -394,3 +395,86 @@ def test_disk_attachment_fails_with_conflicting_disk(self):
                 expected_volume=attached_volume,
                 expected_status=u'in-use',
             )
+
+    @require_virsh
```
jongiddy: This test doesn't require virsh. Either the decorator should be renamed, or maybe a different decorator for tests that require virtio, but not particularly virsh.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -394,3 +395,86 @@ def test_disk_attachment_fails_with_conflicting_disk(self):
...
+            expected_status=u'in-use',
+        )
+
+        self.assertRaises(
+            UnattachedVolume,
+            self.blockdevice_api.get_device_path,
+            volume.id,
+        )
+
+    @require_virsh
```
jongiddy: Ditto, decorator name.

jongiddy:
I will go with your testing on Devstack. And it works for Rackspace acceptance tests, so go ahead and merge into 1.3.0 branch.

None of my comments are critical for making the system work. They are cleanups (DRY, clear names and comments) so I recommend merging this branch, and addressing the comments in the branch to be merged into `master`.

PR #1927, opened 2015-09-14
[FLOC-3016] Add support for ZFS 0.6.5

jongiddy:
ZFS 0.6.5 has stopped loading modules when the `zpool` command is run.  This prevents our Vagrant Tutorial box from building, and may cause problems for people using our instructions to build a ZFS backend.

It is possible to add the environment variable `ZFS_MODULE_LOADING=yes` to continue to load modules, as for older versions of ZFS.  This is a temporary provision in the ZFS code, which will be removed at a later date. This PR applies this to our ZFS commands, to ensure our tests continue to pass when ZFS 0.6.5 is installed.

An additional issue https://clusterhq.atlassian.net/browse/FLOC-3018 exists to provide a more permanent solution.

As this issue prevents a test from passing, I would like to merge it, even if some other tests fail: particularly those that do not deal with ZFS, and in particular the acceptance tests that are currently running up against AWS request limits.

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -877,7 +877,9 @@ def task_create_flocker_pool_file():
     return sequence([
         run('mkdir -p /var/opt/flocker'),
         run('truncate --size 10G /var/opt/flocker/pool-vdev'),
-        run('zpool create flocker /var/opt/flocker/pool-vdev'),
+        # XXX - See FLOC-3018
+        run('ZFS_MODULE_LOADING=yes '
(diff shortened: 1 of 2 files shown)


jongiddy:
Note, see http://build.clusterhq.com/builders/flocker%2Fvagrant%2Fbuild%2Ftutorial to see that this PR fixes a sequence of failing tests since the weekend release of ZFS 0.6.5.

dwgebler:
Thanks @jongiddy this looks fine to me, please merge when the builds are happy. Hopefully we'll do the permanent solution sooner rather than later.

PR #1928, opened 2015-09-14
[FLOC-3017] Allow run-acceptance-tests to boot devstack guests

wallrj:
Design / Partial implementation for: https://clusterhq.atlassian.net/browse/FLOC-3017

TODO
 * Figure out how to get libcloud to wait for public IP of instance on Rackspace and private IP on devstack, OR
 * Figure out how to assign public IPs on devstack (needn't be actual public IPs, maybe just so called floating IPs)
 * devstack guests need to be able to access the internet but can't without some nat IPtables rules on the devstack host. We'd been adding the rules individually for each guest but perhaps somethign like `iptables -t nat  -I POSTROUTING --source 10.0.0.0/8 -j MASQUERADE` will do the trick.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -2,6 +2,7 @@ Flocker 1.4.0.dev1 (2015-09-08)
 ===============================
 
 - Improved the error handling when Cinder volumes enter an unexpected state. (FLOC-2970)
+- ``run-acceptance-tests`` now supports all OpenStack variants, not just RackSpace. (FLOC-3017)
 
 Flocker 1.3.0 (2015-09-04)
 =====================================
--- a/flocker/provision/_openstack.py
+++ b/flocker/provision/_openstack.py
@@ -13,16 +13,22 @@ from ._ssh import run_remotely
 
 from ._effect import sequence
 
+# XXX: Copied from _aws. Needs refactoring
+_usernames = {
+    'centos-7': 'centos',
+    'ubuntu-14.04': 'ubuntu',
+    'ubuntu-15.04': 'ubuntu',
+}
+
 
 def get_default_username(distribution):
     """
     Return the username available by default on a system.
 
     :param str distribution: Name of the operating system distribution
-    :return str: The username made available by Rackspace for this
-        distribution.
+    :return str: The username made available by AWS for this distribution.
     """
-    return 'root'
+    return _usernames[distribution]
 
 
 def provision_openstack(node, package_source, distribution, variants):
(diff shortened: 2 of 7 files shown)


wallrj:
It kind of works:
```
./admin/run-acceptance-tests --no-pull --distribution=ubuntu-14.04 --provider=openstack --dataset-backend=devstack-openstack --config-file=$PWD/config.yml --branch=master --flocker-version='' flocker.acceptance.endtoend.test_dataset
...

[ubuntu@10.0.0.9]: If you would like to use Docker as a non-root user, you should now consider
[ubuntu@10.0.0.9]: adding your user to the "docker" group with something like:
[ubuntu@10.0.0.9]: 
[ubuntu@10.0.0.9]:   sudo usermod -aG docker ubuntu
[ubuntu@10.0.0.9]: 
[ubuntu@10.0.0.9]: Remember that you will have to log out and back in for this to take effect!
[ubuntu@10.0.0.9]: 
[ubuntu@10.0.0.9]: Running apt-get update
[ubuntu@10.0.0.9]: E: Could not open lock file /var/lib/apt/lists/lock - open (13: Permission denied)
[ubuntu@10.0.0.9]: E: Unable to lock directory /var/lib/apt/lists/
[ubuntu@10.0.0.9]: E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
[ubuntu@10.0.0.9]: E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
Destroying acceptance-test-richardw-0
Destroying acceptance-test-richardw-1


...
```


And `iptables -t nat  -I POSTROUTING --source 10.0.0.0/8 -j MASQUERADE` allows new guests to access the internet.

NEWS
```diff
@@ -2,6 +2,7 @@ Flocker 1.4.0.dev1 (2015-09-08)
 ===============================
 
 - Improved the error handling when Cinder volumes enter an unexpected state. (FLOC-2970)
+- ``run-acceptance-tests`` now supports all OpenStack variants, not just RackSpace. (FLOC-3017)
```
tomprince: I'd avoid say all, just some other than rackspace.

flocker/provision/_openstack.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Return the username available by default on a system.
+
+    :param str distribution: Name of the operating system distribution
+    :return str: The username made available by AWS for this distribution.
+    """
+    return _usernames[distribution]
+
+
+def provision_openstack(node, package_source, distribution, variants):
```
tomprince: Is this any different than rackspace? If not, we should combine them.

tomprince:
I think the right solution is to enable floating IPs. Or, better but I don't know if it is possible, mark the fixed IPs that are assigned as public, rather than private.

flocker/provision/_openstack.py
tomprince: Looking at the commit messages it may be different. In that case, probably every bit of it needs comments to explain why the bits are there (as contrast to all the other slightly different copies of this code.

PR #1929, opened 2015-09-15
[FLOC-3021] Backport ZFS 0.6.5. fix to release 1.3.0

jongiddy:
This patch needs to be applied to the 1.3 branch in order for tests to pass, so that we can release 1.3.1 for other significant bugs.


zubron:
The change looks good. I'm happy for this to be merged when the tests pass.

PR #1931, opened 2015-09-15
Flocker create jenkins osx job floc 2854

Azulinho:
Adds OSX job for creating a homebrew package and testing it on the OSX slave.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1931)


build.yaml
```diff
@@ -101,13 +108,12 @@ common_cli:
     # choose to configured.
     # The list of ASCII codes it uses can be found here.
     # https://wiki.archlinux.org/index.php/Color_Bash_Prompt
-    parse_logs() {
+    function parse_logs {
       # ubuntu defaults to mawk, which causes some things to break
```
Azulinho: OSX doesn't eat:
 name() {
but it is happy with:
function name {

build.yaml
```diff
@@ -101,13 +108,12 @@ common_cli:
...
       # it seems to be related to buffering, -W interactive seems to fix it.
-      case "\$(os)"
-      in
-        "ubuntu") alias awk='mawk -W interactive'
-      esac
```
Azulinho: no point in having a case statement for a single option

build.yaml
```diff
@@ -155,8 +161,9 @@ common_cli:
     # permissions/owners.
     # This is to be tackled as part of:
     # https://clusterhq.atlassian.net/browse/FLOC-2689
-    sudo chmod 777 /var/run/docker.sock
-
+      if [ 'OSX' != "\$(os)" ]; then
+        sudo chmod 777 /var/run/docker.sock
+      fi
```
Azulinho: we don't have docker on our OSX cloudbees boxes

build.yaml
```diff
@@ -168,10 +175,20 @@ common_cli:
...
+      rm -rf \$venv
+      rm -f "Flocker*.rb"
+      rm -f "dist/Flocker-*.tar.gz"
+    fi
+    if [ 'ubuntu' == "\$(os)" ] || [ 'centos' == "\$(os)" ]; then
+      sudo rm -rf \$venv
+      sudo rm -f results.xml
+      sudo rm -f trial.log
+      sudo rm -rf _trial_temp/
+    fi
```
Azulinho: OSX cloudbees doesn't allows us to sudo (no tty)

build.yaml
```diff
@@ -196,13 +213,21 @@ common_cli:
...
+        parse_logs pip install -i \$PIP_INDEX_URL "Flocker[dev]" --trusted-host \$TRUSTED_HOST
+        # using the caching-layer, install junix for our coverage report
+        parse_logs pip install -i \$PIP_INDEX_URL python-subunit junitxml  --trusted-host \$TRUSTED_HOST
+    fi
+    if [ "\$(os)" == "OSX" ]; then
+      pip install --upgrade pip
+      pip install .
+      pip install  "Flocker[dev]"
+      pip install python-subunit junitxml
+    fi
```
Azulinho: we're not injecting PIP caching env variables to our OSX boxes, that is done on centos and ubuntu through the slave_setup jenkins plugin, which we can't execute on cloudbees

build.yaml
```diff
@@ -710,6 +754,19 @@ job_type:
...
+        - { type: 'shell',
+            cli: [ *hashbang,
+                   *cleanup, *setup_venv, *setup_flocker_modules,
+                   'export DISTRIBUTION_NAME=OSX',
+                   *check_version,
+                   *build_sdist, *build_homebrew_package,
+                   *uninstall_old_homebrew_flocker_packages,
+                   *install_homebrew_package ]
+          }
+      clean_repo: false
```
Azulinho: new client job for OSX client test

build.yaml
adamtheturtle: Works for me but doesn't matter:

```
~$ uname
Darwin
~$ name () {
> echo "hi"
> }
~$
~$ name
hi
```

build.yaml
```diff
@@ -155,8 +161,10 @@ common_cli:
     # permissions/owners.
     # This is to be tackled as part of:
     # https://clusterhq.atlassian.net/browse/FLOC-2689
-    sudo chmod 777 /var/run/docker.sock
-
+    # we only run this on centos, ubuntu as OSX slaves don't have docker
```
adamtheturtle: Let's instead check if Docker is available and then run this if it is

build.yaml
```diff
@@ -168,10 +176,21 @@ common_cli:
     # https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/20
     # So we will place the virtualenv in /tmp/v instead
     #
-    sudo rm -rf \$venv
-    sudo rm -f results.xml
-    sudo rm -f trial.log
-    sudo rm -rf _trial_temp/
+    if [ 'OSX' == "\$(os)" ]; then
```
adamtheturtle: :question: Might be neater if above we set variables like `is_osx`.

build.yaml
adamtheturtle: Put this as a comment in the code

build.yaml
```diff
@@ -317,6 +344,25 @@ common_cli:
...
     parse_logs \$venv/bin/python admin/build-package --destination-path repo --distribution \${DISTRIBUTION_NAME} /flocker/dist/Flocker-\${FLOCKER_VERSION}.tar.gz
 
+  build_homebrew_package: &build_homebrew_package  |
+    # and build a homebrew
+    SDIST_URL=file://\${PWD}/dist/Flocker-\${FLOCKER_VERSION}.tar.gz
+    RECIPE_FILE=\${PWD}/Flocker\${GIT_COMMIT}.rb
+    \${venv}/bin/python -m admin.homebrew --flocker-version \${GIT_COMMIT} --sdist \${SDIST_URL} --output-file \${RECIPE_FILE}
+    \$venv/bin/python admin/build-package --destination-path repo --distribution \${DISTRIBUTION_NAME} /flocker/dist/Flocker-\${FLOCKER_VERSION}.tar.gz
+
+  install_homebrew_package: &install_homebrew_package
```
adamtheturtle: We want to not be in our virtualenv here.

build.yaml
```diff
@@ -317,6 +344,25 @@ common_cli:
...
+    # install the new freshly baked package
+    brew update
```
adamtheturtle: Link to an issue (which should be a follow up to this) to use the existing homebrew testing task, instead of hardcoding these instructions here.

build.yaml
```diff
@@ -317,6 +344,25 @@ common_cli:
...
+    brew tap ClusterHQ/tap
+    brew install \${RECIPE_FILE}
+    brew test \${RECIPE_FILE}
+
+  uninstall_old_homebrew_flocker_packages: &uninstall_old_homebrew_flocker_packages
```
adamtheturtle: We should be testing on fresh boxes so this should not be necessary

build.yaml
```diff
@@ -168,10 +176,21 @@ common_cli:
...
-    sudo rm -rf \$venv
-    sudo rm -f results.xml
-    sudo rm -f trial.log
-    sudo rm -rf _trial_temp/
+    if [ 'OSX' == "\$(os)" ]; then
+      # OSX doesn't allow us to sudo (no tty)
+      rm -f results.xml
+      rm -f trial.log
+      rm -rf _trial_temp/
+      rm -rf \$venv
```
adamtheturtle: Same as another comment, let's make it so we're on a clean box and these lines aren't necessary

build.yaml
```diff
@@ -168,10 +176,21 @@ common_cli:
...
+      rm -f "Flocker*.rb"
+      rm -f "dist/Flocker-*.tar.gz"
+    fi
+    if [ 'ubuntu' == "\$(os)" ] || [ 'centos' == "\$(os)" ]; then
```
adamtheturtle: Comment why you need `sudo` here - and not on OS X.

build.yaml
Azulinho: actually, this might alread be done as part of FLOC-2689

build.yaml
```diff
@@ -196,13 +215,25 @@ common_cli:
...
+        parse_logs pip install -i \$PIP_INDEX_URL "Flocker[dev]" --trusted-host \$TRUSTED_HOST
+        # using the caching-layer, install junix for our coverage report
+        parse_logs pip install -i \$PIP_INDEX_URL python-subunit junitxml  --trusted-host \$TRUSTED_HOST
+    fi
+    if [ "\$(os)" == "OSX" ]; then
+      # we can't use our pip caching layer on our OSX cloudbees boxes
+      # as we're unable run the slave_setup plugin that injects the environment
+      # variables to the slave.
+      # So we just go upstream for the python modules.
+      pip install --upgrade pip
```
adamtheturtle: Why are these not `parse_logs` as above?

Ideally we'd have variables set for the `PIP_INDEX_URL` and `TRUSTED_HOST` for OS X (say, `https://pypi.python.org/simple` and another default). Then we could avoid repeating the commands here (can be an XXX and follow up later).

build.yaml
```diff
@@ -317,6 +348,25 @@ common_cli:
     # and build a rpm/deb package using docker
     parse_logs \$venv/bin/python admin/build-package --destination-path repo --distribution \${DISTRIBUTION_NAME} /flocker/dist/Flocker-\${FLOCKER_VERSION}.tar.gz
 
+  build_homebrew_package: &build_homebrew_package  |
+    # and build a homebrew
```
adamtheturtle: Expand on this, we are building a source distribution and then a Homebrew recipe which installs that source distribution and its dependencies.

build.yaml
```diff
@@ -90,24 +90,52 @@ common_cli:
...
+      fi
+    }
+    function is_osx {
+      if [ 'OSX' == "\$(os)" ]; then
+        echo "true"
+      else
+        echo "false"
+      fi
+    }
+
```
Azulinho: added is_<os?> functions

build.yaml
```diff
@@ -151,15 +179,9 @@ common_cli:
...
-    # fix the docker permission issue, the base image doesn't have the correct
-    # permissions/owners.
-    # This is to be tackled as part of:
-    # https://clusterhq.atlassian.net/browse/FLOC-2689
-    sudo chmod 777 /var/run/docker.sock
-
 
   cleanup: &cleanup |
-    export PATH=/usr/local/bin:\$PATH
+    export PATH=/usr/local/bin:\${PATH}
```
Azulinho: wrap every $VAR, make it consistent and avoid spending another hour troubleshooting a broken groovy file due to a misslaced $\{

build.yaml
```diff
@@ -168,14 +190,28 @@ common_cli:
...
+      sudo rm -f results.xml
+      sudo rm -f trial.log
+      sudo rm -rf _trial_temp/
+    fi
 
   setup_venv: &setup_venv |
     # setup the new venv
-    virtualenv -p python2.7 --clear \$venv
+    virtualenv -p python2.7 --clear \${venv}
     . \${venv}/bin/activate
```
Azulinho: same ${ var thing

build.yaml
```diff
@@ -196,13 +232,26 @@ common_cli:
     # SSL on our caching box.
     # The --trusted-host option is only available with pip 7.
     #
-    parse_logs pip install --upgrade pip
+    if [ "\$(is_ubuntu)" == "true" ] || [ "\$(is_centos)"  == "true" ]; then
+        PIP_ADDITIONAL_OPTIONS=" -i \${PIP_INDEX_URL} --trusted-host \${TRUSTED_HOST}"
+        PARSE_LOGS="parse_logs "
+    fi
+    if [ "\$(is_osx)" == "true" ]; then
+      # we can't use our pip caching layer on our OSX cloudbees boxes
```
Azulinho: use new is_OSs functions

build.yaml
```diff
@@ -196,13 +232,26 @@ common_cli:
     # SSL on our caching box.
     # The --trusted-host option is only available with pip 7.
     #
-    parse_logs pip install --upgrade pip
+    if [ "\$(is_ubuntu)" == "true" ] || [ "\$(is_centos)"  == "true" ]; then
+        PIP_ADDITIONAL_OPTIONS=" -i \${PIP_INDEX_URL} --trusted-host \${TRUSTED_HOST}"
+        PARSE_LOGS="parse_logs "
```
Azulinho: set a cpl of helpers for the PIP arguments when used with caching layer

build.yaml
```diff
@@ -196,13 +232,26 @@ common_cli:
...
+    fi
+    if [ "\$(is_osx)" == "true" ]; then
+      # we can't use our pip caching layer on our OSX cloudbees boxes
+      # as we're unable run the slave_setup plugin that injects the environment
+      # variables to the slave.
+      # So we just go upstream for the python modules.
+      PIP_ADDITIONAL_OPTIONS=""
+      PARSE_LOGS=""
+    fi
+
```
Azulinho: no caching layer for OSX

Azulinho: and no parsing of logs yet

build.yaml
```diff
@@ -196,13 +232,26 @@ common_cli:
...
-    # using the caching-layer, install flocker
-    parse_logs pip install -i \$PIP_INDEX_URL "Flocker[dev]" --trusted-host \$TRUSTED_HOST
-    # using the caching-layer, install junix for our coverage report
-    parse_logs pip install -i \$PIP_INDEX_URL python-subunit junitxml  --trusted-host \$TRUSTED_HOST
+    \${PARSE_LOGS} pip install --upgrade pip \${PIP_ADDITIONAL_OPTIONS}
+    \${PARSE_LOGS} pip install . \${PIP_ADDITIONAL_OPTIONS}
+    # install Flocker
+    \${PARSE_LOGS} pip install  "Flocker[dev]" \${PIP_ADDITIONAL_OPTIONS}
+    # install junix for our coverage report
+    \${PARSE_LOGS} pip install python-subunit junitxml \${PIP_ADDITIONAL_OPTIONS}
```
Azulinho: use the same cli for both Linux and OSX

Azulinho:
@adamtheturtle think I addressed all your comments

build.yaml
```diff
@@ -90,24 +90,52 @@ common_cli:
     #
```
tomprince: Looking at the comment above this, it seems this has long stopped being the '"first-line".

build.yaml
```diff
@@ -90,24 +90,52 @@ common_cli:
...
+      _nix_flavour=\$(uname)
+      if [ 'Darwin' == "\${_nix_flavour}" ]; then
+        echo "OSX"
+      else
+        # Determine OS platform
+        source /etc/os-release
+        echo \${ID}
+      fi
     }
+    function is_ubuntu {
```
tomprince: I'm curious why these functions don't use the exit code. Then they could be used like
```sh
if is_ubuntu; then
    ...
fi
```

build.yaml
tomprince: From dash(1):

>    If the program is not a normal executable file (i.e., if
    it does not begin with the "magic number" whose ASCII representation is
    "#!", so execve(2) returns ENOEXEC then) the shell will interpret the
    program in a subshell. The child shell will reinitialize itself in
    this case, so that the effect will be as if a new shell had been
    invoked to handle the ad-hoc shell script, except that the location of
    hashed commands located in the parent shell will be remembered by the
    child.

I bet the script isn't set as executable, so the shell that spawns the script spawns a subshell. This is probably explicitly /bin/sh (via /bin/sh -c <our-script> probably?). So, since /bin/sh is dash on ubuntu, the script is run with that on ubuntu, despite the shebang line. Making the script executable should fix that.

tomprince: :exclamation: I think rather than trying to work around various differences in `/bin/sh` we should actually make sure we are running in bash.

Or better, consider using a language other than shell that is available everywhere (python for example).

build.yaml
tomprince: :exclamation: Except you appear to have changed the meaning.

build.yaml
```diff
@@ -168,14 +196,28 @@ common_cli:
...
-    sudo rm -rf \$venv
-    sudo rm -f results.xml
-    sudo rm -f trial.log
-    sudo rm -rf _trial_temp/
+    if [ "\$(is_osx)" == "true" ]; then
+      # OSX doesn't allow us to sudo (no tty)
+      rm -f results.xml
+      rm -f trial.log
+      rm -rf _trial_temp/
+      rm -rf \${venv}
```
tomprince: :exclamation: It would be nice to get rid of this duplication here (have a list of files to delete?).

I'm also curious how this relates to https://github.com/ClusterHQ/flocker/blob/flocker-create-jenkins-osx-job-FLOC-2854/jobs.groovy.j2#L124-L137 which seems to serve the same purpose.

build.yaml
```diff
@@ -196,15 +238,26 @@ common_cli:
     # SSL on our caching box.
     # The --trusted-host option is only available with pip 7.
     #
-    parse_logs pip install --upgrade pip
+    if [ "\$(is_ubuntu)" == "true" ] || [ "\$(is_centos)"  == "true" ]; then
+        PIP_ADDITIONAL_OPTIONS=" -i \${PIP_INDEX_URL} --trusted-host \${TRUSTED_HOST}"
+        PARSE_LOGS="parse_logs "
+    fi
+    if [ "\$(is_osx)" == "true" ]; then
```
tomprince: If the problem is that the environment variables aren't set on OSX, then it seems like checking for those environment variables would be a better choice here, since that is what we care about.

build.yaml
tomprince: OSX does appear to have awk, so I'm not sure why there isn't parsing of logs.

build.yaml
```diff
@@ -394,15 +460,12 @@ common_cli:
...
-      "centos")
-        sudo createrepo .
-        ;;
-    esac
+    if [ "\$(is_ubuntu)" == "true" ]; then
+      sudo sh -c 'dpkg-scanpackages --multiversion . | gzip > Packages.gz'
+    fi
+    if [ "\$(is_centos)" == "true" ]; then
+      sudo createrepo .
+    fi
```
tomprince: :exclamation: I find the new code harder to understand. The original code makes it clear that there are multiple options being chosen between. With the new code, I need to realize that `is_ubuntu` and `is_centos` are mutually exclusive and (currently!) exhaustive, and that if a new distribution is added, I should add another `if`.

(It would be even better if the original case had a default case that errored out, so one would know that another case needs to be added by a noisy failure).

build.yaml
```diff
@@ -762,12 +819,7 @@ job_type:
       # flocker.provision is responsible for creating the test nodes on
       # Rackspace, so we can actually run run-acceptance-tests from AWS
       on_nodes_with_labels: 'aws-ubuntu-trusty-T2Medium'
-      with_modules:
-        - flocker.acceptance.endtoend.test_dataset
-        - flocker.acceptance.integration.test_mongodb
-        - flocker.acceptance.integration.test_postgres
-        - flocker.acceptance.obsolete.test_cli
-        - flocker.acceptance.obsolete.test_containers
+      with_modules: *run_acceptance_modules
```
tomprince: These changes are unrelated.

tomprince:
There are also a number of (now hidden) comments by @adamtheturtle that were addressed or responded to (I found them by looking at reviewable).

Please address the above issue an re-submit.

build.yaml
Azulinho: no clue about what your comment meant

build.yaml
Azulinho: agree, however these were suggestions from the first round of code review.

build.yaml
Azulinho: these variables are not set on OSX (yet), however I would rather have a method to explicitly define what variables I set for different OS

build.yaml
Azulinho: no particular reason, different approach

build.yaml
Azulinho: "consider using a language other than shell that is available everywhere (python for example)"
I would rather have you creating a JIRA issue for that as it's not related at all to this flocker issue

build.yaml
Azulinho: its the shebang, which is still the first like of the script.
not sure what you meant here

build.yaml
Azulinho: because the escape codes didn't provide colour on the output.
it can be tackled on a different ticket.
my preference would be not to use awk but something easier on the eyes

build.yaml
Azulinho: the pre-scm plugin is not executed on OSX boxes as they're slaves from cloudbees

build.yaml
Azulinho: they are, however I don't think they justify an expensive cycle of new JIRA ticket, new branch, commit, PR, code review request

tomprince: Each unrelated change makes reviewing a PR harder.

build.yaml
tomprince: Can you expand on that somewhat (and add a comment explaining it).

It doesn't explain why there is also the cleanup on centos/ubuntu, which presumably does have the pre-scm plugin run.

In fact, is this even needed on OSX? My understanding is that this was necessary to cleanup root-owned files, and that there is already a step that does `git clean` making this unnecessary on OSX.

tomprince:
The comments on line 439, 445, 447 are what I was referring to.

@Azulinho I know you seem to dislike reviewable.io, but there is enough stuff and multiple reviews here that using anything else is unwieldly.

---

Reviewed 1 of 1 files at r6.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

<sup>**[build.yaml, line 90 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-90)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L90)):</sup>
There is *much* more than just the shebang here now (a bunch of helper functions, and variable defintions etc.

---

<sup>**[build.yaml, line 105 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-105)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L105)):</sup>
I'm just always leery of anything that passes values around in strings and then parses them. It is hard to avoid in a shell, but this seems like somewhere where avoiding it is the natural approach.

---

<sup>**[build.yaml, line 112 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r1-112)** ([raw file](https://github.com/clusterhq/flocker/blob/8a99cdb30e458cfe32e373dd350b2e846f10db42/build.yaml#L112)):</sup>
Certainly a reasonable response to the second part of my comment is to file an issue, or start a discussion elsewhere. I don't think it is unreasonable to leave a comment with the suggestion at the point in the code that prompted the thought though.

I do think making sure we are in fact running on bash is appropriate to address either here, or at least before this branch lands. Looking at 
http://52.88.192.98:8080/job/ClusterHQ-flocker/job/flocker-create-jenkins-osx-job-FLOC-2854/job/run_client_installation_on_OSX/22/consoleFull
I see 
```
17:41:58 [run_client_installation_on_OSX] $ /bin/sh -xe /var/folders/t3/qty_skss0ts40zrv1q0ypyjm000fmr/T/hudson6342074241276611136.sh
```
which suggests that my hypothesis is correct.

---

<sup>**[build.yaml, line 204 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-204)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L204)):</sup>
Actually, I guess that `${venv}` may need to be deleted on OSX, but on both OSX and Linux, that shouldn't need sudo.

---

<sup>**[build.yaml, line 245 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-245)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L245)):</sup>
I'm not sure that I understand. You are checking on whether to set additional pip arguments, which needs to be done because some places don't set the environment variables we need.  It seems to make the most sense to check whether the environment we need are set, and pass the additional arguments in that case; rather than testing for OSX which happens, for non-essential reasons, to be the only place where this currently occurs.

Based on your comment below, I guess parse_logs is broken for some reason on OSX. So for that, it makes sense to test against OSX specifically. But there should be a comment explaining why we don't try to do that on OSX.

But ... I wonder if that test would make more sense in `parse_logs` rather than everywhere where we might happen to want to parse logs?  We already have a conditional in that function to do something different on ubuntu, it seems simpler (and clearer for later readers) to make it a no-op on OSX, than to conditionally call it everywhere.

---

<sup>**[build.yaml, line 468 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-468)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L468)):</sup>
Looking at all the comments on the PR, I don't see any such comments. Given that we both think the explicit rather than implicit case is better, I suggest you drop this hunk.

I guess you may be referring to
> Might be neater if above we set variables like is_osx.

I think that that is because it is clearer to read in the context where that comment was made, but the original code is clearer in this context.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

build.yaml
tomprince: The case statement talks about ubuntu, the if statement about OSX.

Azulinho:
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks failed.

---

<sup>**[build.yaml, line 90 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-90)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L90)):</sup>
ah! gotcha, renamed it to add_shell_functions.
It wasn't clear what you meant initially. makes sense now


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 14 unresolved discussions, all commit checks successful.

---

<sup>**[build.yaml, line 105 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-105)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L105)):</sup>
Done.


---

<sup>**[build.yaml, line 136 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r1-112)** ([raw file](https://github.com/clusterhq/flocker/blob/8a99cdb30e458cfe32e373dd350b2e846f10db42/build.yaml#L136)):</sup>
there's already a ticket for addressing that issue,
https://clusterhq.atlassian.net/browse/FLOC-2986 as mentioned in the comments


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, all commit checks successful.

---

<sup>**[build.yaml, line 136 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r1-112)** ([raw file](https://github.com/clusterhq/flocker/blob/8a99cdb30e458cfe32e373dd350b2e846f10db42/build.yaml#L136)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 201 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r2-179)** ([raw file](https://github.com/clusterhq/flocker/blob/0d86f1c72b0977aabd076d6ea7e8ccc197272f85/build.yaml#L201)):</sup>
Done.


---

<sup>**[build.yaml, line 206 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r3-184)** ([raw file](https://github.com/clusterhq/flocker/blob/57ad6772a0aa8aed37f8314b46dc2123359b3852/build.yaml#L206)):</sup>
added comment about cloudbees behaviour, where we can't expect to receive a new slave on every build


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 206 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r3-184)** ([raw file](https://github.com/clusterhq/flocker/blob/57ad6772a0aa8aed37f8314b46dc2123359b3852/build.yaml#L206)):</sup>
Done.


---

<sup>**[build.yaml, line 206 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-204)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L206)):</sup>
it could be that is no longer necessary on ubuntu and centos, since we now use the pre-scm plugin.
However I would rather tackle that as a separate JIRA.
https://clusterhq.atlassian.net/browse/FLOC-3139


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 206 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-204)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L206)):</sup>
Done.


---

<sup>**[build.yaml, line 210 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r3-188)** ([raw file](https://github.com/clusterhq/flocker/blob/57ad6772a0aa8aed37f8314b46dc2123359b3852/build.yaml#L210)):</sup>
Done.


---

<sup>**[build.yaml, line 252 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r3-232)** ([raw file](https://github.com/clusterhq/flocker/blob/57ad6772a0aa8aed37f8314b46dc2123359b3852/build.yaml#L252)):</sup>
Done.


---

<sup>**[build.yaml, line 255 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r1-231)** ([raw file](https://github.com/clusterhq/flocker/blob/8a99cdb30e458cfe32e373dd350b2e846f10db42/build.yaml#L255)):</sup>
Done.


---

<sup>**[build.yaml, line 441 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r3-352)** ([raw file](https://github.com/clusterhq/flocker/blob/57ad6772a0aa8aed37f8314b46dc2123359b3852/build.yaml#L441)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 456 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r2-361)** ([raw file](https://github.com/clusterhq/flocker/blob/0d86f1c72b0977aabd076d6ea7e8ccc197272f85/build.yaml#L456)):</sup>
I am setting an expectaction that it may not be the case on cloudbees, and that I may be using an used slave


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 452 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r2-356)** ([raw file](https://github.com/clusterhq/flocker/blob/0d86f1c72b0977aabd076d6ea7e8ccc197272f85/build.yaml#L452)):</sup>
can you raise one ?
I am not entirelly sure of the details how the current testing task is cooked on bb


---

<sup>**[build.yaml, line 456 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r2-361)** ([raw file](https://github.com/clusterhq/flocker/blob/0d86f1c72b0977aabd076d6ea7e8ccc197272f85/build.yaml#L456)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 833 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-822)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L833)):</sup>
and each rejection increases the total cost of delivering the feature.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

Azulinho:
Review status: 0 of 1 files reviewed at latest revision, 15 unresolved discussions, some commit checks pending.

---

<sup>**[build.yaml, line 450 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r2-354)** ([raw file](https://github.com/clusterhq/flocker/blob/0d86f1c72b0977aabd076d6ea7e8ccc197272f85/build.yaml#L450)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

tomprince:
<sup>**[build.yaml, line 112 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r1-112)** ([raw file](https://github.com/clusterhq/flocker/blob/8a99cdb30e458cfe32e373dd350b2e846f10db42/build.yaml#L112)):</sup>
I mildly object to changing coding-style to workaround broken shells when we
don't intend to use them, especially when the issues is (I think) debugged.

On the other hand, I don't have strong opinions about style.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)

tomprince:
Review status: 0 of 1 files reviewed at latest revision, 13 unresolved discussions, some commit checks failed.

---

<sup>**[build.yaml, line 105 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-105)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L105)):</sup>
:question: This should be equivalent to
```shell
function is_blah () {
    [ 'ubuntu' == "\$(os)"]
}
<!-- Reviewable comment -K-Ji04wgZmSc0DnILvn:-89671205 -->

---

<sup>**[build.yaml, line 822 \[r6\]](https://reviewable.io:443/reviews/clusterhq/flocker/1931#-JzlTRRH7--vWCagHU_e-r6-822)** ([raw file](https://github.com/clusterhq/flocker/blob/d51fe76bea14cf29760d51a128078a014ab3b3e8/build.yaml#L822)):</sup>
Which is why I didn't reject on the basis of the changes, just griped about them.
<!-- Reviewable comment -K-Jif6Amb-s4_1kSF5d:-957715891 -->

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1931)
<!-- Sent from Reviewable.io -->

Azulinho:
if you weren't happy with it, why did you accept it ?
https://clusterhq.atlassian.net/browse/FLOC-2854?page=com.atlassian.streams.streams-jira-plugin:activity-stream-issue-tab

tomprince:
Because 
1. the last comment I made suggested addressing the issue before landing this PR.
2. I didn't care enough to block the PR.

PR #1933, opened 2015-09-15
[FLOC-2770] Check release packages

jongiddy:
Currently, we do not check that the artifacts published during release are correct. This has lead to problems only observed when not using Buildbot repos, hence usually undetected by any of our automated tests. (e.g. https://clusterhq.atlassian.net/browse/FLOC-2904 and https://clusterhq.atlassian.net/browse/FLOC-3008).

This PR adds an additional verification step to the release process, to check that the client packages can be installed in a Docker container OK. Many of the issues that could affect our packaging (S3 permissions, errors in packaging format) should be obvious from these tests.  Although this test only detects problems after a release, at least we're ahead of our users in awareness of problems with the release, as opposed to waiting for them to complain.

The test script is created as a Python unittest, allowing it to perform all tests and summarise the results.  Error outputs are saved to a file for later perusal.

Tests for the client now not only print the version number, but also check it against the expected version number.  Since this is inappropriate for our documentation, the client tests have been removed as task directives, and replaced in the docs.

To test the new code, run `admin/test-artifacts`.

At this moment, the CentOS RPM tests fails, due to FLOC-3008.  This is useful to demonstrate the error output. Since FLOC-3008 is resolved, merging this branch with `master` eliminates this problem.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1933)


tomprince:
> Tests for the client now not only print the version number, but also check it against the expected version number. Since this is inappropriate for our documentation, the client tests have been removed as task directives, and replaced in the docs.

We could just add new functions, rather than changing this around. But making this change also seems reasonable.

---

Merge after addressing the comments and buildbot is green.

---

Reviewed 7 of 7 files at r1.

---

<sup>**[admin/test-artifacts, line 67 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1933#-Jzt3u1Eti_yIpP_PO1X)** ([raw file](https://github.com/clusterhq/flocker/blob/05821439c9c1887bc7cbc801705e834a05ce32a2/admin/test-artifacts#L67)):</sup>
`unittest.main` isn't a very nice test runner. If it isn't too difficult, it might be nice to replace this with trial.

---

<sup>**[flocker/provision/_install.py, line 215 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1933#-JzsyIIbXG_IKcpLS3UN)** ([raw file](https://github.com/clusterhq/flocker/blob/05821439c9c1887bc7cbc801705e834a05ce32a2/flocker/provision/_install.py#L215)):</sup>
You should use `pipes.quote` to quote the expected package version.

---

<sup>**[flocker/provision/_install.py, line 246 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1933#-Jzt0SSm1Yvm3dvswbsL)** ([raw file](https://github.com/clusterhq/flocker/blob/05821439c9c1887bc7cbc801705e834a05ce32a2/flocker/provision/_install.py#L246)):</sup>
`pipes.quote`

---

<sup>**[flocker/provision/_install.py, line 487 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1933#-Jzt0KFZs-1py7OzeSdM)** ([raw file](https://github.com/clusterhq/flocker/blob/05821439c9c1887bc7cbc801705e834a05ce32a2/flocker/provision/_install.py#L487)):</sup>
`pipes.quote`

---

<sup>**[vagrant/tutorial/post-reboot-bootstrap.py, line 47 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1933#-JzswD4VmvfixcaZ2QN3)** ([raw file](https://github.com/clusterhq/flocker/blob/05821439c9c1887bc7cbc801705e834a05ce32a2/vagrant/tutorial/post-reboot-bootstrap.py#L47)):</sup>
There should be a comment on why this is here. Presumably this is to record debugging information.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1933)

PR #1934, opened 2015-09-15
[FLOC-3030] Let me specify the number of libcloud nodes to create and provision.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3030


dwgebler:
LGTM, one builder failed which I am re-running as it appears to be a known intermittent error.

PR #1936, opened 2015-09-15
[FLOC-3032] Re-install flocker-docker-plugin on acceptance test upgrades

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3032

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1936)

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -1072,8 +1072,13 @@ def task_install_flocker(
 
     :raises: ``UnsupportedDistribution`` if the distribution is unsupported.
     """
-    return task_package_install("clusterhq-flocker-node",
-                                distribution, package_source)
+    return sequence(list(
+        task_package_install(package, distribution, package_source)
+        for package in [
+            "clusterhq-flocker-node",
+            "clusterhq-flocker-docker-plugin"
+        ]
+    ))
 
 
 ACCEPTANCE_IMAGES = [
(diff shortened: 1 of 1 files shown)


dwgebler:
LGTM subject to builds passing. Several builds failed but they all look coincidental, a mix of known intermittent failures, I am re-running them.

wallrj:
Forced another rebuild on two builds that failed.

wallrj:
Ah, I think @itamarst may have done separate docker plugin installation to avoid adding it to the docs (until the flocker-docker-plugin is officially released)
 * See http://build.clusterhq.com/results/docs/reinstall-flocker-docker-plugin-FLOC-3032/build-9999/install/install-node.html#installing-on-centos-7

itamarst:
Well, it's about to be official once docs branch is merged. But I think Docker plugin installation is maybe something that should be optional, yeah, in which case putting it here is problematic, yeah. In particular it'll be installed but broken because it's missing a certificate file.

dwgebler:
:question: We could make a reasonably small change to introduce the concept of a "private" provisioning step in a task that does not get processed through the Sphinx extension.

adamtheturtle:
> :question: We could make a reasonably small change to introduce the concept of a "private" provisioning step in a task that does not get processed through the Sphinx extension.

We already have used this before. See for example `locker.provision._install.ensure_minimal_setup` and its use.

PR #1937, opened 2015-09-15
[FLOC-3033] Stop things before removing them

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3033

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -980,14 +980,31 @@ def _uninstall_flocker_centos7():
     Return an ``Effect`` for uninstalling the Flocker package from a CentOS 7
     machine.
     """
-    return sequence([
-        run_from_args([
-            b"yum", b"erase", b"-y", b"clusterhq-python-flocker",
-        ]),
-        run_from_args([
-            b"yum", b"erase", b"-y", b"clusterhq-release",
-        ]),
-    ])
+    def maybe_disable(unit):
+        return run(
+            u"{{ "
+            u"systemctl is-enabled {unit} && "
+            u"systemctl stop {unit} && "
+            u"systemctl disable {unit} "
+            u"; }} || /bin/true".format(unit=unit).encode("ascii")
+        )
+
+    return sequence(
+        list(
+            # XXX There should be uninstall hooks for stopping services.
+            maybe_disable(unit) for unit in [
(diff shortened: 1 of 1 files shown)


flocker/provision/_install.py
```diff
@@ -980,14 +980,31 @@ def _uninstall_flocker_centos7():
...
+            u"{{ "
+            u"systemctl is-enabled {unit} && "
+            u"systemctl stop {unit} && "
+            u"systemctl disable {unit} "
+            u"; }} || /bin/true".format(unit=unit).encode("ascii")
+        )
+
+    return sequence(
+        list(
+            # XXX There should be uninstall hooks for stopping services.
```
adamtheturtle: I'll have a look at this tomorrow but it looks like this (or the XXX follow up) could borrow from http://doc-dev.clusterhq.com/gettinginvolved/contributing.html?highlight=test#testing-code-on-nodes

wallrj: Here's some stuff about post install RPM hooks: http://www.ibm.com/developerworks/library/l-rpm2/

wallrj:
Forced rebuild of failed builds.

flocker/provision/_install.py
Azulinho: we use fpm don't we ?
it supports pre/post hooks

```
--after-install FILE          A script to be run after package installation
    --before-install FILE         A script to be run before package installation
    --after-remove FILE           A script to be run after package removal
    --before-remove FILE          A script to be run before package removal
    --after-upgrade FILE          A script to be run after package upgrade. If not specified,
                                  --before-install, --after-install, --before-remove, and 
                                  --after-remove wil behave in a backwards-compatible manner
                                  (they will not be upgrade-case aware).
                                  Currently only supports deb and rpm packages.
    --before-upgrade FILE         A script to be run before package upgrade. If not specified,
                                  --before-install, --after-install, --before-remove, and 
                                  --after-remove wil behave in a backwards-compatible manner
                                  (they will not be upgrade-case aware).
                                  Currently only supports deb and rpm packages.
```

exarkun:
Builds are passing now.

adamtheturtle:
Thanks, merge if you have run this and it is removing the dangling units.

PR #1938, opened 2015-09-16
[FLOC-3036] Send node agent state only if it changed or new client.

myechuri:
Send node agent state to control agent only if:

- it differs from previously reported state, or,
- previous state send failed, or,
- last known state was sent to a different client.

@tomprince and i paired on this issue.

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -120,8 +120,6 @@ class ClusterStatus(object):
 
     def output_STORE_CLIENT(self, context):
         self.client = context.client
-        # Clear convergence loop's last sent local status cache
-        self.convergence_loop_fsm.last_sent_local_status = None
 
     def output_UPDATE_STATUS(self, context):
         self.convergence_loop_fsm.receive(
@@ -306,13 +304,19 @@ class ConvergenceLoop(object):
         self.reactor = reactor
         self.deployer = deployer
         self.cluster_state = None
+        self.client = None
 
         # Save last known local state
         self.last_sent_local_state = None
 
     def output_STORE_INFO(self, context):
+        old_client = self.client
         self.client, self.configuration, self.cluster_state = (
             context.client, context.configuration, context.state)
+        if old_client is not self.client:
+            # State updates are now being sent somewhere else.  At least send
+            # one update using the new client.
+            self.last_sent_local_state = None
 
     def _maybe_send_state_to_control_service(self, state_changes):
         """
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -307,9 +307,7 @@ class ConvergenceLoop(object):
         self.deployer = deployer
         self.cluster_state = None
 
-        # Save last known local state, and AMP client we sent it to.
-        # Since only one control service can exist at the moment, there can
-        # no more than one AMP client at any given point in time.
+        # Save last known local state
         self.last_sent_local_state = None
 
     def output_STORE_INFO(self, context):
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -290,9 +290,10 @@ class ConvergenceLoop(object):
 
     :ivar fsm: The finite state machine this is part of.
 
-    :ivar last_sent_local_state: Last reported local state to
-        control agent.
-    :type last_sent_local_state: tuple of IClusterStateChange
+    :ivar last_acknowledged_state: The last state that was sent to and
+        acknowledged by the control service over the most recent connection
+        to the control service.
+    :type last_acknowledged_state: tuple of IClusterStateChange
     """
     def __init__(self, reactor, deployer):
         """
@@ -305,9 +306,7 @@ class ConvergenceLoop(object):
         self.deployer = deployer
         self.cluster_state = None
         self.client = None
-
-        # Save last known local state
-        self.last_sent_local_state = None
+        self.last_acknowledged_state = None
 
     def output_STORE_INFO(self, context):
         old_client = self.client
@@ -316,7 +315,7 @@ class ConvergenceLoop(object):
         if old_client is not self.client:
             # State updates are now being sent somewhere else.  At least send
             # one update using the new client.
-            self.last_sent_local_state = None
+            self.last_acknowledged_state = None
 
     def _maybe_send_state_to_control_service(self, state_changes):
         """
@@ -326,7 +325,7 @@ class ConvergenceLoop(object):
         :param state_changes: State to send to the control service.
         :type state_changes: tuple of IClusterStateChange
         """
-        if self.last_sent_local_state != state_changes:
+        if self.last_acknowledged_state != state_changes:
             context = LOG_SEND_TO_CONTROL_SERVICE(
                 self.fsm.logger, connection=self.client,
                 local_changes=list(state_changes),
@@ -338,13 +337,10 @@ class ConvergenceLoop(object):
                     eliot_context=context)
                 )
 
-                def set_sent_state(_):
-                    self.last_sent_local_state = state_changes
+                def record_acknowledged_state(ignored):
+                    self.last_acknowledged_state = state_changes
 
-                def clear_sent_state(f):
-                    self.last_sent_local_state = None
-                    return f
-                d.addCallbacks(set_sent_state, clear_sent_state)
+                d.addCallback(record_acknowledged_state)
                 d.addErrback(
                     writeFailure, self.fsm.logger,
                     u"Failed to send local state to control node.")
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -317,34 +317,37 @@ class ConvergenceLoop(object):
             # one update using the new client.
             self.last_acknowledged_state = None
 
+    def _send_state_to_control_service(self, state_changes):
+        context = LOG_SEND_TO_CONTROL_SERVICE(
+            self.fsm.logger, connection=self.client,
+            local_changes=list(state_changes),
+        )
+        with context.context():
+            d = DeferredContext(self.client.callRemote(
+                NodeStateCommand,
+                state_changes=state_changes,
+                eliot_context=context)
+            )
+
+            def record_acknowledged_state(ignored):
+                self.last_acknowledged_state = state_changes
+
+            d.addCallback(record_acknowledged_state)
+            d.addErrback(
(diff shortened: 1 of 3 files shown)


flocker/node/_loop.py
```diff
@@ -288,6 +291,10 @@ class ConvergenceLoop(object):
         ``None``.
 
     :ivar fsm: The finite state machine this is part of.
+
+    :ivar last_sent_local_state: Last reported local state to
+        control agent.
+    :type last_sent_local_state: tuple of IClusterStateChange
```
wallrj: It doesn't seem to be a tuple any more. Perhaps it was originally going to be a `2-tuple(state, client)` or something.
Document the `last_sent_client` attribute too.

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
         self.deployer = deployer
         self.cluster_state = None
 
+        # Save last known local state, and AMP client we sent it to.
```
wallrj: typo "state, and"

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+        # Since only one control service can exist at the moment, there can
+        # no more than one AMP client at any given point in time.
```
wallrj: typo "there can no more"

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+        self.lock = threading.Lock()
```
wallrj: Document the `lock` attribute.

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+    def _maybe_send_state_to_control_service(self, state_changes):
+        """
+        If current state_changes differ from last sent local state,
+        transmit the new state to control agent.
+
+        :param state_changes: State to send to the control service.
+        :type state_changes: tuple of IClusterStateChange
+        """
+        if (self.last_sent_local_state != state_changes or
+           self.last_sent_client != self.client):
```
wallrj: Can `AgentAMP` protocol instances be meaningfully compared or are we just checking for identity... that these are different instances?
Are we comparing `client` in order to handle the situation where the dataset agent is disconnected and reconnects to an existing control service, or
Are we doing it to handle the situation where the control service is restarted and loses it's state?
Or maybe both?
We should add comments / docstring explaining why we're comparing clients (like you've done in the PR) and explain the situations where we consider the client may change.

wallrj: And on that note, we're using a `ReconnectingClientFactory` and I started worrying whether we'd get a *new* client on disconnect or re-use the existing client. I think it must be the latter right? Protocols don't get re-used. I'll read the code.

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
 
+    def _maybe_send_state_to_control_service(self, state_changes):
+        """
+        If current state_changes differ from last sent local state,
+        transmit the new state to control agent.
+
+        :param state_changes: State to send to the control service.
+        :type state_changes: tuple of IClusterStateChange
+        """
+        if (self.last_sent_local_state != state_changes or
```
wallrj: The indentation below looks wrong.

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+            )
+            with context.context():
+                d = DeferredContext(self.client.callRemote(
+                    NodeStateCommand,
+                    state_changes=state_changes,
+                    eliot_context=context)
+                )
+
+                def set_sent_state(_):
+                    with self.lock:
```
wallrj: Yeah, I don't think this is necessary. 
`flocker-dataset-agent` is single threaded so e.g.  `self.client` can't have been re-assigned from another thread. 
Is that the sort of problem you were anticipating?
I've seen people do e.g. `self.last_sent_local_state, self.last_sent_client = state_changes, self.client` and I've always understood that to be an atomic operation, but I think that might just be folk lore...at least according to:
 * http://stackoverflow.com/questions/13649684/is-sequence-unpacking-atomic

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+                        self.last_sent_local_state = state_changes
+                        self.last_sent_client = self.client
+
+                def clear_sent_state(f):
+                    with self.lock:
```
wallrj: Remove this lock too if you agree.

flocker/node/_loop.py
```diff
@@ -300,10 +307,54 @@ def __init__(self, reactor, deployer):
...
+                        self.last_sent_local_state = None
+                        self.last_sent_client = None
+                    return f
+                d.addCallbacks(set_sent_state, clear_sent_state)
+                d.addErrback(
+                    writeFailure, self.fsm.logger,
+                    u"Failed to send local state to control node.")
```
wallrj: Does eliot still require the `logger` to be supplied? i thought not.

flocker/node/test/test_loop.py
```diff
@@ -9,6 +9,8 @@
 from eliot.testing import validate_logging, assertHasAction, assertHasMessage
 from machinist import LOG_FSM_TRANSITION
 
+from pyrsistent import s
```
wallrj: `s` ? Is there a more readable version of this?

flocker/node/test/test_loop.py
```diff
@@ -303,6 +307,167 @@ def test_convergence_done_notify(self, logger):
         self.assertEqual(client.calls, [(NodeStateCommand,
                                          dict(state_changes=(local_state,)))])
 
+    def test_convergence_done_unchanged_notify(self):
+        """
+        A FSM doing convergence that gets a discovery result that is unchanged
+        from the last time it sent data does not send the discoverd state to
```
wallrj: typo "discoverd"

flocker/node/test/test_loop.py
```diff
@@ -303,6 +307,167 @@ def test_convergence_done_notify(self, logger):
...
+        the control service.
+        """
```
wallrj: I got this far. I've got to a meeting now.

wallrj:
Thanks @tomprince and @myechuri 

I didn't finish reading all the code, but I think the locking needs to be removed and I spotted one or two typos missing docs etc.

@exarkun measured the performance of this loop and got encouraging results. Perhaps he'll post those and do a final review.

flocker/node/_loop.py
tomprince: I wonder if it would make more sense to just clear the last sent state when we get a new client, rather than trying to track the client?

exarkun:
I measured this using the procedure I outlined elsewhere.  On a ten node cluster, this branch brings flocker-control CPU usage to around 15%.

exarkun:
(Compared to current master@HEAD where it's close to 100%)

tomprince:
While we don't need locks to update the last sent state, the current code updates the last sent state asynchronously. It occurred to me that this might cause `last_sent_state` to get updated out of order. Perhaps it should be set synchronously, and only cleared asynchronously on failure. Another possibility would be to wait for the server to acknowledge the state before proceeding to the next iteration of convergence (but still in parallel with the current iterations convergence.

flocker/node/test/test_loop.py
myechuri: Fixed.

flocker/node/test/test_loop.py
myechuri: Fixed. Used ``pset`` instead. Thanks!

flocker/node/_loop.py
myechuri: Convergence loop seems to have specific logger (https://github.com/ClusterHQ/flocker/blob/master/flocker/node/_loop.py#L418), and associated tests (https://github.com/ClusterHQ/flocker/blob/master/flocker/node/test/test_loop.py#L384). Removing logger parameter from above line regressed one of the 37 tests:

```
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 1368, in _runCleanups
    f(*args, **kwargs)
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/eliot/testing.py", line 274, in <lambda>
  File "/home/ubuntu/flocker/flocker/node/test/test_loop.py", line 559, in assert_full_logging
    u"desired_configuration": self.configuration})
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/eliot/testing.py", line 385, in assertHasAction
    testCase.assertTrue(actions, "No actions of type %s" % (actionType,))
  File "/home/ubuntu/flocker/.tox/py27/local/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 403, in assertTrue
twisted.trial.unittest.FailTest: No actions of type <eliot._validation.ActionType object at 0x7f49e9387c90>

flocker.node.test.test_loop.ConvergenceLoopFSMTests.test_convergence_done_delays_new_iteration

```

In the interest of not refactoring this code, am leaving the ``logger`` parameter as is. Please let me know if you think this change needs to be checked in with current PR. Thanks!

flocker/node/_loop.py
myechuri: Removed.

flocker/node/_loop.py
myechuri: Thanks for the Zulip conversation. Removed the lock.

flocker/node/_loop.py
myechuri: Flake8 wants it this way, so, leaving it as is. Thanks!

flocker/node/_loop.py
myechuri: > I wonder if it would make more sense to just clear the last sent state when we get a new client, rather than trying to track the client?

Yes, that seems cleaner. Testing out the change. Thanks!

flocker/node/_loop.py
myechuri: Fixed.

flocker/node/_loop.py
myechuri: Not needed anymore since we got rid of last sent client. Thanks!

PR #1943, opened 2015-09-16
[FLOC 2760] Documentation for the Flocker plugin for Docker 

davidoliverSP2:
Fixes 2760

@ferrantim and I are already aware that we require a new diagram for the content in the Introduction, and we already have a rough draft that will be sent to fffunction.

Do not merge this branch until the updated image is added (or the current one is removed)

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1943)


docs/labs/index.rst
```diff
@@ -33,14 +22,14 @@ Currently Flocker supports both volume-centric operation via the dataset API and
...
 * See what's happening in your Flocker cluster with a :ref:`CLI <labs-volumes-cli>` and a :ref:`GUI <labs-volumes-gui>`.
 * Make it easier to spin up a Flocker cluster in the first place with an :ref:`installer <labs-installer>`.
 * Integrate Flocker with other popular tools, like :ref:`Weave <labs-weave>`, :ref:`Mesosphere <labs-mesosphere>` and eventually :ref:`Kubernetes <labs-kubernetes>`.
 
 **We believe that Flocker will be more successful if, as well as focusing on making it useful for managing data volumes, we work on integrating it with other components in the emerging Docker and container ecosystem.**
 
-Our biggest step towards this goal so far is the :ref:`Flocker Docker plugin <labs-docker-plugin>`, which makes Flocker pluggable directly into the Docker Engine and directly usable from the ``docker run`` CLI.
+Our biggest step towards this goal so far is the Flocker Docker plugin, which enables you to integrate Flocker with tools like :ref:`Swarm <labs-swarm>` and :ref:`Compose <labs-compose>`, and pluggable directly into the Docker Engine and directly usable from the ``docker run`` CLI.
```
dwgebler: I don't know what we're looking to do in respect of referring to the plugin as Flongle throughout the docs - need to clarify that with marketing and product though.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
...
+
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "echo hello > /data/file.txt"
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "cat /data/file.txt"
+
+The second command should print ``hello`` on the screen if everything has worked; the same data written to the Flocker volume in the first command.
+
+Known limitations
+=================
+
+* If the volume exists on a different node and is currently being used by a container, the Flocker plugin does not stop it being migrated out from underneath the running container.
```
dwgebler: There is a question mark over this. The sentence here is technically correct, but Docker itself may prevent a volume from being ripped out from an underlying container. Further, other ecosystem orchestration tools should prevent this happening anyway. It is a limitation, but I wouldn't want to accidentally present it to people as something that's likely to happen in the course of ordinary usage.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
...
+=========================
+
+Once installed, two simple Docker tests can be run to verify that the plugin is working correctly with the Flocker agents:
+
+.. prompt:: bash $
+
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "echo hello > /data/file.txt"
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "cat /data/file.txt"
+
+The second command should print ``hello`` on the screen if everything has worked; the same data written to the Flocker volume in the first command.
```
dwgebler: Probably worth going in to a bit more detail about what these commands are doing and why it verifies the plugin is working. Clarify that we are creating a named volume (apples) using Flongle as the volume driver and that therefore in the second container, we are reattaching the same volume from the first container and verifying that its data has been preserved.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
...
+   
+   DOCKER_OPTS="-s aufs"
+
+On each of your container agent servers, install the Flocker plugin:
+
+.. prompt:: bash $
+
+   apt-get install clusterhq-flocker-docker-plugin
+
+The service can be started, stopped, or restarted using ``{[service flocker-docker-plugin restart}}``
```
dwgebler: I think these steps are all that's required (other than generating and copying your `plugin` certificate and key files), but let's double-check with @itamarst and @lukemarsden that nothing has been overlooked or missed out here - but AFAIK, there is no other config and installing the package should set up `systemd` to launch the plugin service on system startup, or launch it via upstart on Ubuntu.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
...
+.. prompt:: bash $
+
+   wget -qO- https://get.docker.com/ | sudo sh
+
+Now you can use the following instructions to install the Flocker plugin for Docker on each of the nodes in your cluster.
+
+On CentOS 7
+===========
+
+On each of your container agent servers, install the Flocker plugin:
```
dwgebler: "container agent servers" is confusing wording - the Flocker container agent has no part in Flongle, it's the dataset agents that the plugin uses. Elsewhere we refer to the package containing these agents that users have installed as "Flocker Node", Remember the difference in architecture here is that the Flocker volume manager is being controlled by Docker.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
...
+
+On the same machine where you ran ``flocker-ca`` while installing Flocker, you will need to generate a new API user certificate and key for a user named ``plugin``, as described in the :ref:`generate-api` instructions.
+
+Upload these files to :file:`/etc/flocker/plugin.key` and :file:`/etc/flocker/plugin.crt` on the nodes where you want to run the Flocker plugin for Docker.
+
+You will also need to have Docker (at least version 1.8 or later) installed. The following command will install the latest version available:
+
+.. prompt:: bash $
+
+   wget -qO- https://get.docker.com/ | sudo sh
```
dwgebler: Better to link to Docker's installation docs than to reproduce any specific instructions here that may be subject to change, or vary between platforms.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
+
+========================================
+Installing the Flocker Plugin for Docker
+========================================
+
+Before installing the Flocker plugin for Docker, you will need to have installed Flocker on some nodes, using the instructions in the previous topics.
+
+On the same machine where you ran ``flocker-ca`` while installing Flocker, you will need to generate a new API user certificate and key for a user named ``plugin``, as described in the :ref:`generate-api` instructions.
```
dwgebler: :question: would it be easier here to give an example in a prompt section of the `flocker-ca` command to run? e.g.

```
flocker-ca create-api-certificate plugin
scp ./plugin.crt root@172.16.255.251:/etc/flocker/plugin.crt
scp ./plugin.key root@172.16.255.251:/etc/flocker/plugin.key
```

perhaps worth saying more clearly that Flongle requires access to the Flocker REST API, therefore you will need to create an API certificate and install it on each node you are also using the plugin.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,74 @@
+
+========================================
+Installing the Flocker Plugin for Docker
+========================================
+
+Before installing the Flocker plugin for Docker, you will need to have installed Flocker on some nodes, using the instructions in the previous topics.
```
dwgebler: This prose can probably be expanded and clarified a bit more and it might also be a good idea to link again to the node installation instructions where we say "installed Flocker on some nodes".

dwgebler:
Thanks @davidoliverSP2 I've left some initial comments for some changes and would also like someone else to review the next pass of this branch, as I've been too closely involved in getting the plugin ready for release and would like someone with less internal exposure to how it all works to have a say on how the docs read for them.

docs/labs/index.rst
davidoliverSP2: confirmed with @ferrantim - the plugin will be called "the Flocker plugin for Docker"

Michael or @robhaswell, please let me know if this is not correct.

docs/install/docker-plugin.rst
davidoliverSP2: I could resolve this with either:
1. Removing this limitation
2. Amending the text to something like this:

> If the volume exists on a different node and is currently being used by a container, the Flocker plugin would not stop it being migrated out from underneath the running container. It is possible that Docker, or your orchestration tool would prevent this from happening, but Flocker itself does not.

@lukemarsden or @robhaswell - could you offer an opinion here?

docs/labs/index.rst
robhaswell: Sorry guys. I tried. It's the Flocker plugin for Docker, or what ever Michael says.

itamarst:
@robhaswell wants first release to *not* be targetted at Docker CLI (since that will be broken) so maybe that shouldn't be mentioned, or at least the warning about safety should be right there.

docs/install/docker-plugin.rst
robhaswell: Sure, sounds good to me.

itamarst:
If we're going to ship a potentially data-corrupting plugin we really need to include some documentation on how to integrate safely (presuming it's possible). I'll write an outline tomorrow based on outcome of conversation with Rob.

itamarst:
This should also document that upgrading the plugin requires restarting Docker, as per https://clusterhq.atlassian.net/browse/FLOC-2846

dwgebler:
One thing to note is that branch https://github.com/ClusterHQ/flocker/pull/1936 adds installation of the plugin package to the `install_flocker` task, which adds the corresponding commands to the documentation for installing `flocker-node` - which is something we don't want. But we do want the provisioning change effected when upgrading nodes for running the acceptance tests in CI.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. prompt:: bash $
+
+   systemctl restart flocker-docker-plugin
+
+
+On Ubuntu 14.04
+===============
+
+On Ubuntu, it's best to ensure that Docker is using the ``AUFS`` storage driver.
```
lukemarsden: unrelated to flongle

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,97 @@
...
+
+.. prompt:: bash $
+
+   flocker-ca create-api-certificate plugin
+   scp ./plugin.crt root@172.16.255.251:/etc/flocker/plugin.crt
+   scp ./plugin.key root@172.16.255.251:/etc/flocker/plugin.key
+
+Upload these files to :file:`/etc/flocker/plugin.key` and :file:`/etc/flocker/plugin.crt` on the nodes where you want to run the Flocker plugin for Docker.
+
+You will also need to have Docker (at least version 1.8 or later) installed. For more information, see `the Docker installation instructions <https://docs.docker.com/>`_.
```
lukemarsden: this should say where Docker 1.8+ should be installed. on the nodes? on my local workstation? also, isn't that a prerequisite for installing Flocker that Docker is installed? so maybe the point of this section is to make it clear about the docker version dependency (1.8+)

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,97 @@
...
+* You should not move a volume from one node to another unless you are sure no containers are using the volume.
+
+  The Flocker plugin will not stop volumes from being migrated out from underneath a running container.
+  It is possible that Docker or your orchestration tool will prevent this from happening, but Flocker itself does not.
+* ``--volumes-from`` and equivalent Docker API calls will only work if both containers are on the same machine.
+
+  Some orchestration frameworks may not schedule containers in a way that respects this restriction, so check before using ``--volumes-from``.
+* We recommend only using named volumes when using the Flocker plugin.
+
+  If you use volumes in your Docker run commands without specified names, anonymous volumes can be created.
```
lukemarsden: ... "or images which specify a VOLUME"...

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,97 @@
...
+  This occurs as Docker defines volume drivers for the entire run command, not per-volume.
```
lukemarsden: "The result of this is that each time such a container is started on a different host, a new random name for the anonymous volumes is created. This can waste resources when the underlying volumes are provisioned from, for example, EBS. Therefore, we recommend not using Docker images that specify VOLUMEs, or anonymous (`-v /path`) style volumes when running Docker containers with the Flocker Docker plugin."

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,97 @@
...
+=================
+
+* You should not move a volume from one node to another unless you are sure no containers are using the volume.
+
+  The Flocker plugin will not stop volumes from being migrated out from underneath a running container.
+  It is possible that Docker or your orchestration tool will prevent this from happening, but Flocker itself does not.
+* ``--volumes-from`` and equivalent Docker API calls will only work if both containers are on the same machine.
+
+  Some orchestration frameworks may not schedule containers in a way that respects this restriction, so check before using ``--volumes-from``.
+* We recommend only using named volumes when using the Flocker plugin.
```
lukemarsden: ... "that is, volumes which are specified using the `-v name:/path` syntax in `docker run`."

docs/labs/index.rst
```diff
@@ -33,14 +22,14 @@ Currently Flocker supports both volume-centric operation via the dataset API and
...
+The Flocker Docker plugin started out as an unofficial labs project, but is now supported in the :ref:`official Flocker Docker plugin <docker-plugin>` documentation.
```
lukemarsden: yay!

docs/labs/index.rst
```diff
@@ -15,17 +15,6 @@ If we get lots of positive feedback about any one of these projects then we will
 
 .. _labs-demo:
 
-Flocker Docker plugin with CLI and GUI
```
lukemarsden: Please revert deleting this section. We can link to the official plugin docs, but this is a really important demo and it remains valid.

lukemarsden:
@dwgebler @itamarst @davidoliverSP2 

I did a short review of this PR so far, I would appreciate it if the comments I left could be addressed before merging, especially insofar as it touches the labs docs.

I'd also like to ensure that the unofficial installer gets updated to support installing the new plugin *and that the docs get updated on flocker* before we release this change. That doesn't have to block merging this change, but I'll put change the installer and then put a PR in ASAP and we can make sure they get released together.

itamarst:
I guess I'm going to go with "I don't care if the installer is broken, it's unofficial and you get to support it when it breaks" so sure :grin:

docs/labs/index.rst
robhaswell: This page ranks, and references the labs plugin as if it is the only plugin (which is fair, it is).

Instead of being removed, it should be rewritten in such a way that this it is clear that this is not a supported confguration, that this demo is legacy, and that they can get the real plugin somewhere else.

Probably this entire section, and all the plugin references in the labs section, should carry a warning. @davidoliverSP2 what do you reckon?

docs/install/docker-plugin.rst
itamarst: Isn't it fine if you specify an explicit path for the ``VOLUME`` on the command line?

itamarst:
OK, addressed Luke's comments. The installation instructions in the unofficial installer docs have been modified to match what a hypothetical updated unofficial installer would do given the new plugin system.

docs/labs/index.rst
lukemarsden: The only change that needs to happen to this section is that `:ref:\`Flocker Docker plugin <labs-docker-plugin>\`` needs to be replaced with `:ref:\`Flocker Docker plugin <docker-plugin>\``.

lukemarsden:
Thanks @itamarst. [This comment](https://github.com/ClusterHQ/flocker/pull/1943/files#r39988209) still needs addressing, otherwise this looks good.

itamarst:
Which comment?

lukemarsden:
My comment contained a hyperlink to the comment: https://github.com/ClusterHQ/flocker/pull/1943/files#r39988209

The comment is the one requesting the undeletion of the "Flocker Docker plugin with CLI and GUI" section with the demo. I still want that demo to be on the main page of the labs docs, even though the flocker docker plugin is official now (it also shows off two other labs projects, the CLI and GUI, which work really well together with the flocker docker plugin because they allow users to see what's going on in their cluster).

davidoliverSP2:
Thanks @itamarst 

I've just changed a few small things, and fixed a merge conflict on the Labs/Installer page, which occurred due to a branch merged by @lukemarsden 

I'll hand this back for a technical review.

itamarst:
Final review should probably actually work through the instructions and make sure they work.

itamarst:
@davidoliverSP2 Note docs are failing to build at the moment.

robhaswell:
This comment was discussed in the office:

[] There should be some information under "Using Flocker" that talk about how the plugin works.
[] Users should be directed to install the FDP by default, so there should be additional guidance. @davidoliverSP2 knows more about this.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Testing Your Installation
+=========================
+
+Once installed, two simple Docker tests can be run to verify that the plugin is working correctly with the Flocker agents:
+
+.. prompt:: bash $
+
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "echo hello > /data/file.txt"
+   docker run -v apples:/data --volume-driver flocker busybox sh -c "cat /data/file.txt"
```
robhaswell: This does not really test the installation, it just shows that data can be written to the filesystem.

Perhaps the user should be instructed to run this test over different nodes.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,94 @@
...
+In this example, the first command uses the Flocker plugin for Docker as the volume driver to create a named volume called ``apples``.
+
+In the second command we are reattaching the same volume from the first container, and verifying that the data (``hello``) has been preserved.
+
+Upgrading the Plugin
+====================
+
+If you are upgrading from an earlier version of the plugin, make sure to stop the Docker daemon before doing so and then start it back up once the plugin has been upgraded.
+
+Known Limitations
```
robhaswell: This section should be (possibly duplicated in) "Using the Flocker Docker Plugin".

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,94 @@
...
+=================
+
+* You should not move a volume from one node to another unless you are sure no containers are using the volume.
+
+  The Flocker plugin will not stop volumes from being migrated out from underneath a running container.
+  It is possible that Docker or your orchestration tool will prevent this from happening, but Flocker itself does not.
```
robhaswell: AWS also prevents a volume from being unmounted if it is in use. I have just tested this.

itamarst: It's the OS, and that only works in some cases... I don't think docs need to explain *every* case it'll be prevented, if reader doesn't know enough to figure it out for their own application then they shouldn't rely on it. Not to mention there's still issues with race conditions even in the best case.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,94 @@
+.. _install-docker-plugin:
+
+========================================
+Installing the Flocker Plugin for Docker
+========================================
+
+Before installing the Flocker plugin for Docker, you will need to have installed Flocker on some nodes, using the :ref:`node installation instructions <installing-flocker-node>`.
+
+The Flocker plugin for Docker requires access to the Flocker REST API, and therefore you will need to create an API user certificate and key for a user named ``plugin`` on each node, as described in the :ref:`generate-api` instructions.
```
exarkun: Short sentences are better than run-ons and comma splices.

docs/install/docker-plugin.rst
```diff
@@ -0,0 +1,94 @@
...
+
+
+On Ubuntu 14.04
+===============
+
+On each of your Flocker nodes, install the Flocker plugin:
+
+.. prompt:: bash $
+
+   apt-get install clusterhq-flocker-docker-plugin
```
exarkun: `-y`?  If it's good for yum it's probably good for apt.

docs/install/docker-plugin.rst
exarkun: You don't need two nodes, you just need to look inside whatever's mounted under `/flocker` to see if the data ended up on the Flocker-managed filesystem.  The downside is this involves some UUIDs.

docs/introduction/index.rst
```diff
@@ -40,7 +41,11 @@ With the Flocker API or CLI you can:
 * Attach and detach data volumes from containers as they change hosts
 * Migrate local data volumes between servers (currently Experimental)
 
+You can also use Flocker to manage only your volumes, while continuing to manage your containers however you choose.
+To use Flocker to manage your volumes while tools like Docker, Docker Swarm or Mesos manage your containers, :ref:`use the Flocker plugin for Docker <docker-plugin>`.
```
exarkun: Why this repetition?

docs/introduction/index.rst
```diff
@@ -51,7 +56,74 @@ Flocker supports block-based shared storage such as Amazon EBS, Rackspace Cloud
...
+----
+
+This demo shows both the Flocker Docker plugin in conjunction with the :ref:`Volumes CLI <labs-volumes-cli>` and :ref:`Volumes GUI <labs-volumes-gui>`.
+
+.. raw:: html
+
+   <iframe width="100%" height="450" src="https://www.youtube.com/embed/OhWxJ_hOPx8?rel=0&amp;showinfo=0" frameborder="0" allowfullscreen style="margin-top:1em;"></iframe>
+
+Also check out the `DockerCon Plugin Demos <https://plugins-demo-2015.github.io/>`_ site to see a joint project between ClusterHQ and Weaveworks.
+This is the "ultimate integration demo" — a pre-built demo environment that includes Flocker, Weave, Swarm, Compose & Docker, all working together in harmony.
```
exarkun: `Compose, and Docker —` (if you really need to use em dashes but you should really use commas here).

davidoliverSP2:
@robhaswell comment:

> * There should be some information under "Using Flocker" that talk about how the plugin works.
* Users should be directed to install the FDP by default, so there should be additional guidance. @davidoliverSP2 knows more about this.

FLOC 3129 will create documentation for using the plugin.
(+ The FPD install instructions are now part of the installation steps as part of this branch)

robhaswell:
There are changes in https://github.com/ClusterHQ/flocker/pull/2000 which should be merged into here.

docs/labs/index.rst
```diff
@@ -18,9 +18,9 @@ If we get lots of positive feedback about any one of these projects then we will
 Flocker Docker plugin with CLI and GUI
 ======================================
 
-By way of example, here is a 55 second demo of the :ref:`Flocker Docker plugin <labs-docker-plugin>` provisioning portable Flocker volumes and moving them between hosts directly from the Docker CLI.
+By way of example, here is a 55 second demo of the :ref:`Flocker Docker plugin <docker-plugin>` provisioning portable Flocker volumes and moving them between hosts directly from the Docker CLI.
```
dwgebler: For consistency, I think references on this page should be to the "Flocker Plugin for Docker" rather than "Flocker Docker plugin"

docs/labs/swarm.rst
```diff
@@ -4,9 +4,9 @@
 Flocker with Swarm
 ==================
 
-With the :ref:`Flocker Docker plugin <labs-docker-plugin>`, and with Swarm support for the ``--volume-driver`` option, you can use Flocker together with Docker Swarm.
+With the :ref:`Flocker Docker plugin <docker-plugin>`, and with Swarm support for the ``--volume-driver`` option, you can use Flocker together with Docker Swarm.
 
-First, you need to :ref:`install Flocker <labs-installer>` and the :ref:`Flocker Docker plugin <labs-docker-plugin>`.
+First, you need to :ref:`install Flocker <labs-installer>` and the :ref:`Flocker Docker plugin <docker-plugin>`.
```
dwgebler: Likewise, Flocker Plugin for Docker. Also I think we should link this to the plugin installation instructions rather than the introduction.

docs/labs/index.rst
robhaswell: I considered updating all these references, and didn't. Here's why:

* The labs plugin is actually a different project, different code, used differently. It follows that it has a different name.
-- or --
* Now that FDP is release, all these docs should be blown away / ported to main docs.

On balance I prefer the latter.

docs/introduction/index.rst
```diff
@@ -41,6 +42,7 @@ With the Flocker API or CLI you can:
 * Migrate local data volumes between servers (currently Experimental)
 
 Flocker supports block-based shared storage such as Amazon EBS, Rackspace Cloud Block Storage, and EMC ScaleIO, as well as local storage (currently Experimental using our ZFS storage backend) so you can choose the storage backend that is best for your application.
+:ref:`View a complete list of supported storage backends <storage-backends>`.
```
dwgebler: This link seems out-of-place here, in the manner of a "click here to see..." type link. I'd suggest linking this reference in the narrative text above it somewhere, e.g. "so you can choose any of the `supported storage backends <...>` that is best for your application..."

dwgebler:
Thanks @davidoliverSP2 I've left a few more minor inline comments to address, then this is good to merge.

docs/labs/index.rst
dwgebler: I too prefer the latter, but given that these docs are still here, I'd recommend updating all references to be consistent. Labs projects can get away with different coding structures and standards than we use in engineering, but the way I look at it, their docs are still part of our docs and should be held to the same standard of consistency. The labs plugin had a different name "Flocker Docker Plugin" for labs, but it's not a labs project anymore, it's an official product being referenced by labs docs, so I think it should be referenced with the correct name.

PR #1947, opened 2015-09-17
[FLOC-3024] Slower agent loop

dwgebler:
Slows the dataset agent loop to 3 second iterations (up from 1).
We may or may not want to keep this change pending the performance change results from FLOC-3036.

In testing against AWS, this change did not break any acceptance tests and resulted in a small decrease in CPU usage of the control service, as well as a more considerable decrease in the CPU usage of the dataset agents.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1947)


wallrj:
I merged forward to resolve conflicts and force a new build.

dwgebler:
Some of the changes introduced in `master` have added agent loop tests that affect this branch, so I've updated those tests to reflect this branch's 3 second delay.

wallrj:
We did slow down the container agent loop and fixed a bug in the dataset agent that was causing it to wake up too quickly.

PR #1949, opened 2015-09-17
[FLOC-3015] Fix intermittent lease test failure

dwgebler:
Sometimes, it takes long enough to communicate with a real API that a shorthold lease has expired before we attempt to reacquire it, hence the intermittent failure. This change gives the lease a 60 second expiry and ensures we have successfully acquired it before attempting to acquire a lease on the same dataset on a different node.

### DIFF
--- a/flocker/apiclient/test/test_client.py
+++ b/flocker/apiclient/test/test_client.py
@@ -30,7 +30,7 @@ from .._client import (
 )
 from ...ca import rest_api_context_factory
 from ...ca.testtools import get_credential_sets
-from ...testtools import find_free_port
+from ...testtools import find_free_port, loop_until
 from ...control._persistence import ConfigurationPersistenceService
 from ...control._clusterstate import ClusterStateService
 from ...control.httpapi import create_api_service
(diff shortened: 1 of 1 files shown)


jongiddy:
This appears to prevent the problem, but I'm not sure why this much effort is required to claim a lease. Looking at the code, it seems your first commit should work. In fact, it does not look necessary to even call `list_leases`. The first call to `acquire_lease` should not succeed until the API call has returned a lease. Hence, just setting the lease timeout to 60 should be enough (as long as the test completes within 60 seconds).

This needs further explanation as to why it is happening. The non-test clients should be able to rely on the behaviour in the original test (albeit maybe 10 seconds is not long enough for the test).

dwgebler:
> In fact, it does not look necessary to even call list_leases

It probably isn't, I was just being thorough in listing the leases and checking we had the expected result as an additional step. I have a build running now that just changes the 10 second expiry to 60 seconds and leaves everything else about the test unchanged. Let's see if that works.

jongiddy:
This doesn’t appear to be triggering the error. I re-ran the ubuntu-14.04 and centos-7 tests that were suffering the problem.  As an intermittent error anyway, that doesn't prove that it is fixed. However, it occurs frequently enough that we should be able to merge this, and see whether it reoccurs.

Merge when green.

flocker/apiclient/test/test_client.py
```diff
@@ -325,7 +325,7 @@ def test_acquire_lease_conflict(self):
             acquire a lease that is held by another node.
             """
             dataset_id = uuid4()
-            d = self.client.acquire_lease(dataset_id, self.node_1, 10)
+            d = self.client.acquire_lease(dataset_id, self.node_1, 60)
```
jongiddy: :question: You could still check the result of `acquire_lease` to check the returned lease matches the spec, but really that is a separate test.

PR #1951, opened 2015-09-18
[FLOC-3039] Enable profiling

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3039

### DIFF
--- a/docs/using/administering/debugging.rst
+++ b/docs/using/administering/debugging.rst
@@ -146,6 +146,8 @@ To dump the current profile run the following command as root on the control nod
    pkill -SIGUSR2 flocker-control
 
 This will output the profiling data to a file named :file:`/etc/flocker/profile-<TIMESTAMP>`.
+This will also disable profiling.
+When profiling is re-enabled, new profiling data will be added to the existing profiling data until the profile is dumped again.
 
 See :py:mod:`pstats` for details on how to extract information from this file.
 For example:
--- a/docs/using/administering/debugging.rst
+++ b/docs/using/administering/debugging.rst
@@ -145,8 +145,7 @@ To dump the current profile run the following command as root on the control nod
 
    pkill -SIGUSR2 flocker-control
 
-.. TODO determine where the profile file should be written to
-This will output the profiling data to a file named :file:`profile-<TIMESTAMP>`.
+This will output the profiling data to a file named :file:`/etc/flocker/profile-<TIMESTAMP>`.
 
 See :py:mod:`pstats` for details on how to extract information from this file.
 For example:
--- a/docs/using/administering/debugging.rst
+++ b/docs/using/administering/debugging.rst
@@ -119,6 +119,46 @@ Alternatively, the information can be gathered manually using the following comm
 
 * Flocker log files (see :ref:`Flocker logging <flocker-logging>` above)
 
+Profiling
+---------
+
+.. warning::
+
+   It is not recommended to use profiling while relying on Flocker within a production environment.
+
+Control Service
+^^^^^^^^^^^^^^^
+
+It is possible to obtain :py:mod:`cProfile` profiling data of the :ref:`Control Service <control-service>` between two user defined intervals.
+
+Profiling is disabled by default.
+To enable profiling of the Control Service run the following command as root on the control node:
+
+.. prompt:: bash #
+
+   pkill -SIGUSR1 flocker-control
+
+Profiling data will then be collected until a signal to dump the profiling data is received.
+To dump the current profile run the following command as root on the control node:
+
+.. prompt:: bash #
+
+   pkill -SIGUSR2 flocker-control
+
+.. TODO determine where the profile file should be written to
+This will output the profiling data to a file named :file:`profile-<TIMESTAMP>`.
+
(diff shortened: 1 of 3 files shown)


flocker/control/script.py
```diff
@@ -77,6 +81,35 @@ def main(self, reactor, options):
...
+    def output_profile(signal, frame):
+        """
+        Dump profiling statistics to a file.
+        This disables profiling.
+
+        :param int signal: See ``signal.signal``.
+        :param frame: None or frame object. See ``signal.signal``.
+        """
+        current_time = time.strftime("%Y%m%d%H%M%S")
+        path = FilePath('/etc/flocker/profile-{}'.format(current_time))
```
jongiddy: `/etc` is the wrong place to put this file.  Either `/var/...` or `/tmp` would be better.

adamtheturtle: This is something we weren't sure about. Do you have a preference for an exact path?

jongiddy: Well, `/var/flocker` would be good, but requires that you ensure that directory is created and has correct permissions. So, `/tmp` may be easier, or `/var/tmp` if you need it to survive a reboot.

docs/using/administering/debugging.rst
```diff
@@ -119,6 +119,47 @@ Alternatively, the information can be gathered manually using the following comm
...
+
+See :py:mod:`pstats` for details on how to extract information from this file.
+For example:
+
+.. code-block:: python
+
+   import pstats
+
+   profile = pstats.Stats('profile-20150917161214')
+   profile.print_stats()
```
jongiddy: Use a more useful line, like (copied from pstats page):
```python
profile.sort_stats('cumulative').print_stats(10)
```

docs/using/administering/debugging.rst
```diff
@@ -119,6 +119,47 @@ Alternatively, the information can be gathered manually using the following comm
...
+Profiling data will then be collected until a signal to dump the profiling data is received.
+To dump the current profile run the following command as root on the control node:
+
+.. prompt:: bash #
+
+   pkill -SIGUSR2 flocker-control
+
+This will output the profiling data to a file named :file:`/etc/flocker/profile-<TIMESTAMP>`.
+This will also disable profiling.
+When profiling is re-enabled, new profiling data will be added to the existing profiling data until the profile is dumped again.
```
jongiddy: This line is a bit confusing. Do you mean something like: "If profiling is re-enabled in the same run of the control server, new profiling data will be added to all previously collected profiling data when the profile is dumped again"

docs/using/administering/debugging.rst
```diff
@@ -119,6 +119,47 @@ Alternatively, the information can be gathered manually using the following comm
...
+
+Profiling data will then be collected until a signal to dump the profiling data is received.
+To dump the current profile run the following command as root on the control node:
+
+.. prompt:: bash #
+
+   pkill -SIGUSR2 flocker-control
+
+This will output the profiling data to a file named :file:`/etc/flocker/profile-<TIMESTAMP>`.
+This will also disable profiling.
```
jongiddy: :question: Change this to emphasise first that SIGUSR1 starts profiling and SIGUSR2 stops profiling. Then mention that when profiling is stopped, all profiling data collected up to that point, including from previous intervals of profiling, will be dumped. This clearly matches the two signals as delimiting phases of profiling, while also explaining how to get the profiling data.

flocker/control/script.py
```diff
@@ -77,6 +81,35 @@ def main(self, reactor, options):
...
+    def enable_profiling(signal, frame):
+        """
+        Enable profiling of the control service.
+
+        :param int signal: See ``signal.signal``.
+        :param frame: None or frame object. See ``signal.signal``.
+        """
+        pr.enable()
+
+    def output_profile(signal, frame):
```
jongiddy: :question: As for the docs, consider renaming to `disable_profiling` to emphasise on/off semantics, and document profile output as a side effect.

jongiddy:
Tested with a control service on CentOS 7. Signal handling works to start and stop profiling.

Please address the comments, and resubmit.

flocker/control/script.py
adamtheturtle: I'll go with `/tmp` unless there are any objections.

docs/using/administering/debugging.rst
adamtheturtle: Yes, I will change this.

flocker/control/script.py
adamtheturtle: :+1:

flocker/control/script.py
```diff
@@ -77,6 +81,35 @@ def main(self, reactor, options):
 
 
 def flocker_control_main():
+    pr = cProfile.Profile()
```
jongiddy: Bonus comment: Use `Profile(time.clock)` to get cpu-time instead of wallclock time.

adamtheturtle:
All points addressed but the script couldn't seem to write to `/tmp` or `/var/tmp` (I didn't investigate why) so I used `var/lib/flocker`.

jongiddy:
Tested on Ubuntu 14.04 this time round. Works fine.

Please merge when all-green.

PR #1952, opened 2015-09-18
[FLOC-3014] Link check for UserVoice is forbidden

jongiddy:
The UserVoice website hosted at https://feedback.clusterhq.com is sending 403 Forbidden to Buildbot requests, probably because it frequently performs robot requests.

This PR stops checking the single link to the forum, to prevent linkcheck tests being permanently orange, and to be nice to someone who apparently doesn't want us knocking.

Note, I haven't actually confirmed that this is why Buildbot is getting 403.  The check works when running `make linkcheck` from a laptop.


adamtheturtle:
Thanks, I can't think of anything better to do than this. Merge when green.

davidoliverSP2:
Thanks @jongiddy

PR #1953, opened 2015-09-18
[FLOC-3050] Make leases API public

itamarst:
1. Make leases API endpoints public.
2. Fix various minor content and formatting issues in the examples and API endpoints.
3. Make the JSON API property documentation format correctly in the face of multi-type properties, in particular the "expires" property which supports both numbers and `null`.

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -10,6 +10,12 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+Next release
+============
+
+* The :ref:`dataset API <api>` added support for leases.
+  Leases prevent a dataset from being deleted or moved off a node.
+
 v1.3.1
 ======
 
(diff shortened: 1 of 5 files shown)


docs/releasenotes/index.rst
```diff
@@ -10,6 +10,12 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
+Next release
+============
+
+* The :ref:`dataset API <api>` added support for leases.
+  Leases prevent a dataset from being deleted or moved off a node.
```
dwgebler: :question: Worth expanding the sentence? Maybe not, seeing as it's a release note, but OTOH we don't have prose anywhere explaining the concept of leases.

Even just a couple of extra words might make it a bit clearer though:

`Leases prevent a dataset from being deleted or moved off a node for as long as the lease is held.`

dwgebler:
One entirely optional comment, otherwise please merge when green.

docs/releasenotes/index.rst
itamarst: We definitely need a tutorial for the API, which I think is vaguely on @davidoliverSP2 todo list, but current text is probably fine, it's implicit in what "lease" means.

PR #1955, opened 2015-09-21
[FLOC-3070] Clean the clusterhq-build repo metadata before installing packages from there

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3070

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1955)


itamarst:
You're only doing this in one branch, possibly you should do it unconditionally? You might be switching from branch to whatever the default is or back.

itamarst:
You can skip design review phase for something this small in future.

flocker/provision/_install.py
```diff
@@ -255,6 +255,20 @@ def install_commands_yum(package_name, distribution, package_source,
...
         repo_options = ['--enablerepo=clusterhq-build']
+
+        # There might be a clusterhq-build repo already and its metadata will
+        # be cached.  Wipe out the cache so that metadata gets loaded from the
+        # latest build URL.
+        # XXX: Would it be better to do this during uninstallation?
+        commands.append(
+            run_from_args([
+                b"yum",
+                b"--disablerepo='*'",
```
itamarst: These two lines are so we only clear metadata from clusterhq-build? Might want comment explaining why you are doing specific options.

dwgebler: Was about to ask why we weren't just doing `yum clean all` so I'd +1 that.

tomprince: Since the clusterhq-build repo isn't enabled by default, we need to enable it explicitly in order to clean the data from it.

flocker/provision/_install.py
```diff
@@ -212,6 +212,24 @@ def task_cli_pkg_test():
...
+    Return a command which will clear all of yum's cached data for a particular
+    repository.
+
+    :param bytes repository: The name of the repository to clear.
+    """
+    return run_from_args([
+        b"yum",
+        b"--disablerepo='*'",
+        # XXX If there are shell special characters in the repository name this
+        # will break.
```
tomprince: This comment is false. `run_from_args` is designed to properly quote all the arguments.

One can observe this by seeing that the above argument is rendered as
`'--disablerepo='"'"'*'"'"''`.

tomprince:
I wonder if a simpler solution would be to just put `metadata_expire=0` in the clusterhq-build repository definition.

I guess maybe we install and/or update released versions infrequently enough that we won't hit stale metadata for the release repositories. The default metadata expiry time is 6h, so installing the latest version within 6h of a release will trigger the same issue.

This suggests that perhaps we expire both clusterhq-build (when we have a branch), and clusterhq/clusterhq-testing unconditionally.

tomprince:
Also, it would be better to use `expire-cache` rather than `all`, since that will force yum to check for new metadata, without deleting anything (thus no need to redownload data if it is up to date.

itamarst:
Assuming this works as promised... the only problem I can see is that this impacts the installation docs. Are these actually steps we want all users running?

itamarst:
I guess it doesn't hurt users, so I guess we can live with some extra noise in instructions, they're already not great as it is. Merge when buildbot (other than irrelevant issues like lint failure) is green.

tomprince:
> the only problem I can see is that this impacts the installation docs. Are these actually steps we want all users running?

Probably not. I think this is probably easy enough to do to be worth fixing.

tomprince:
> Assuming this works as promised...

I haven't tested that it works for installs from release packages (one can only test if one installs within 6h before a release). I have observed
- metadata is expired when upgrading a `managed` cluster
- it is possible to succesfully (at least as far as package installation goes) switch branches with a `managed` cluster.

tomprince:
Since the metadata only needs to be cleaned when upgrading, I moved the steps to the uninstall task.

jml:
LGTM. I've triggered a re-run for the failing acceptance test, since it looks unrelated.

PR #1956, opened 2015-09-21
[FLOC-3071] Backport to Flocker 1.3 #1938 Send node state only on change

jongiddy:
This is the backport of PR #1938 to the 1.3 branch. The change here is identical to the change in that PR.


adamtheturtle:
So we're planning to do 1.3.2 from 1.3.1? Or is this just in case that happens? Merge when the build is green and you are happy.

PR #1957, opened 2015-09-21
[FLOC-3072] don't use local docker cached layers on fpm nightly builds

Azulinho:
Docker fails to execute apt-get update on a docker image if the layer is already cached locally, failling the following apt-get install commands

https://github.com/docker/docker/issues/3313


tomprince:
A better solution would be to combine the `apt-get update` and `apt-get install`, I think. That way we get the benefit of caching if possible.

At least if the point is fix the described problem. If we actually always want to get the latest packages, then using `--no-cache` makes sense.

Azulinho:
These Docker images are built every 24h,  and the intention is to make sure
they have the latest updates installed.
I thought apt-get update && apt upgrade would do the business but quickly
come to terms,  that the command would be cached once and prevent updates
in the next runs.
Its one of those nice 'features' of Docker that can do more harm than good.
--no-cache is defo what we need here

jongiddy:
Given the point of these commands is to create a periodically updated image, I agree that `--no-cache` is a sensible option.

Please merge when all-green. You will have to merge `origin/master` to clear the lint error.

adamtheturtle:
:question: Add the explanation from the issue / these comments near the new line

PR #1958, opened 2015-09-21
[FLOC-2944] Retry if EBS returns bogus error when asked to list volumes.

itamarst:


flocker/node/agents/ebs.py
```diff
@@ -706,8 +707,18 @@ def list_volumes(self):
         """
         Return all volumes that belong to this Flocker cluster.
         """
+        try:
+            ebs_volumes = self.connection.get_all_volumes()
+        except EC2ResponseError as e:
+            # Work around some internal race-condition in EBS by retrying,
+            # since this error makes no sense:
+            if e.code == NOT_FOUND:
+                return self.list_volumes()
```
myechuri: Should we limit number of retries? Thanks!

itamarst: Hm. Well, if we have infinite failures then presumably the agent won't succeed at doing anything anyway, so it doesn't matter?

myechuri: Sounds reasonable, although node agent's one second sleep would help avoiding hitting api limits.

PR #1960, opened 2015-09-22
[FLOC-3081] Only persist the configuration when leases have changed

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3081

### DIFF
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -228,7 +228,7 @@ _LOG_UPGRADE = ActionType(u"flocker-control:persistence:migrate_configuration",
 _LOG_EXPIRE = MessageType(
     u"flocker-control:persistence:lease-expired",
     [Field(u"dataset_id", unicode), Field(u"node_id", unicode)],
-    "A lease for a dataset has expired.")
+    u"A lease for a dataset has expired.")
 
 
 class LeaseService(Service):
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -230,6 +230,13 @@ _LOG_EXPIRE = MessageType(
     [Field(u"dataset_id", unicode), Field(u"node_id", unicode)],
     u"A lease for a dataset has expired.")
 
+_LOG_UNCHANGED_DEPLOYMENT_NOT_SAVED = MessageType(
+    u"flocker-control:persistence:unchanged-deployment-not-saved",
+    [],
+    u"The persistence service was told to save a deployment which is the same "
+    u"as the already-saved deployment.  It has optimized this away."
+)
+
 
 class LeaseService(Service):
     """
@@ -403,6 +410,10 @@ class ConfigurationPersistenceService(MultiService):
 
         :return Deferred: Fires when write is finished.
         """
+        if deployment == self._deployment:
+            _LOG_UNCHANGED_DEPLOYMENT_NOT_SAVED().write(self.logger)
+            return
+
         with _LOG_SAVE(self.logger, configuration=deployment):
             self._sync_save(deployment)
             self._deployment = deployment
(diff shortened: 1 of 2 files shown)


flocker/control/_persistence.py
```diff
@@ -279,10 +279,13 @@ def update_leases(transform, persistence_service):
         persistence service has saved.
     """
     config = persistence_service.get()
-    new_config = config.set("leases", transform(config.leases))
-    d = persistence_service.save(new_config)
-    d.addCallback(lambda _: new_config.leases)
-    return d
+    new_leases = transform(config.leases)
+    if config.leases != new_leases:
```
jongiddy: This appears to compare the whole dictionary, including all key and value pairs.  However, the `Leases.expire` code returns the `config.leases` object if no changes are made.

Hence, it should be faster to use `is not` here.

itamarst:
Personally I'd rather do this in configuration persistence layer (i.e. in `save()`), rather than in leases layer: "if configuration hasn't changed from old version, don't bother saving" in general, rather than implementing the logic in each particular thing that saves configuration.

flocker/control/_persistence.py
itamarst: The CPU cost of doing a comparison once a second is very low. Using `is not` introduces brittleness requiring a very specific implementation, though to be fair the test should catch something that breaks this logic.

jongiddy:
At the point highlighted, we know we can use `is not` safely to do a very fast check of a potentially large dictionary (actually a PMap).  The code currently returns the same object when no leases expire, so this should be much faster.

If a transform function is ever used that returns a different object that compares equal, then the code still works correctly, it just attempts to persist more often than needed. And that case could be stopped by Itamar's suggested change.

flocker/control/_persistence.py
exarkun: The safe way to get this optimization would probably be to have pyrsistent implement equality by starting with an identity check.  This puts the complexity where it is minimized and provides a wider benefit at the same time.

wallrj:
We implemented the general change in  `_persistence.py` as @itamarst suggested.

We ran some rough benchmarks of ``==`` vs `is` comparison of Pyrsistent structures and the speed up wasn't compelling

```
(flocker)jean-paul@dilaton:~/Projects/ClusterHQ/flocker/develop$ python -m timeit -s '    
> from flocker.control._persistence import wire_decode                                    
> with open("current_configuration.jsonbig-broken-whatever") as conf:                     
>     config = wire_decode(conf.read())                                                   
> new_config = config.set(version=4)                                                      
> ' 'new_config == config'                                                                
100000 loops, best of 3: 4.91 usec per loop                                               
(flocker)jean-paul@dilaton:~/Projects/ClusterHQ/flocker/develop$ python -m timeit -s '    
> from flocker.control._persistence import wire_decode                                    
> with open("current_configuration.jsonbig-broken-whatever") as conf:                     
>     config = wire_decode(conf.read())                                                   
> new_config = config.set(version=4)                                                      
> ' 'new_config is config'                                                                
100000000 loops, best of 3: 0.018 usec per loop                                           

```

flocker/control/test/test_persistence.py
```diff
@@ -363,6 +367,52 @@ def saved(_):
...
+        with a deployment that is the same as the already-saved deployment.
+        """
+        service = self.service(FilePath(self.mktemp()), None)
+
+        saving = service.save(TEST_DEPLOYMENT)
+
+        def saved_old(ignored):
+            saving = service.save(TEST_DEPLOYMENT)
+            saving.addCallback(self.assertEqual, None)
+
```
itamarst: Need to return `saving` here. And perhaps give it different name so this is less confusing.

itamarst:
Looks good , just one minor fix to tests and then merge when buildbot is happy.

PR #1961, opened 2015-09-22
[FLOC-3083] Log the causes of flocker:controlservice:send_cluster_state

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3083

Note this introduces only one new Eliot action.  It turns out the "node changed" code path is already describe **if** you look at both the agent and control service logs together.


itamarst:
This can just be code review next time.

flocker/control/_model.py
```diff
@@ -804,6 +804,8 @@ class IClusterStateSource(Interface):
     This is presently used for activity/inactivity tracking to inform change
     wiping.
     """
+    # XXX Should this interface be expanded to include the node UUID and the
+    # type of agent (container or dataset) that reported the state change?
```
dwgebler: :question: so yeah, looking through the code that uses the implementation of `IClusterStateSource` I can see *where* we use it, but I'm not clear on what it's doing that's actually useful. It seems we should either not bother logging a `STATE_SOURCE` field, or an implementation of `IClusterStateSource` that reports specific information about the source (UUID, agent, etc.) should be used. Or possibly this information is already represented somewhere in the log tree by the time we get to logging these causes of `send_cluster_state` and when the logs are correctly parsed, there's no benefit to additional source logging here.

dwgebler:
Other than my question over the use of `ChangeSource` in the logging, this looks fine to proceed.

flocker/control/_protocol.py
```diff
@@ -342,8 +365,14 @@ def node_changed(self, source, state_changes):
         :param list state_changes: One or more ``IClusterStateChange``
             providers representing the state change which has taken place.
         """
-        self.cluster_state.apply_changes_from_source(source, state_changes)
-        self._send_state_to_connections(self.connections)
+        # XXX The IClusterStateSource source doesn't seem terribly useful.  I
+        # hoped it might have an IP address or node UUID which we could log but
+        # it only has `last_activity`. On the other hand there will be a parent
+        # Action that includes information about which connection called this
+        # method remotely.
```
exarkun: It seems there isn't such a parent action.

exarkun: Or ... yes, there is.  It's just from another node.  If you examine the merged logs then the log structure does tell you who triggered this.

itamarst:
Looks good, except for the failing tests... fix those and merge.

PR #1962, opened 2015-09-22
[FLOC-3079] Intermittently failing lease test

dwgebler:
Fixes FLOC-3079. The test was failing because the persistence service and API service in this test suite are instantiated using different `IReactorTime` implementations, one using `Clock` and the other using the global `reactor`. This meant that leases created via the API were being created with an expiry calculated as seconds from Unix time, whereas the leases expiry loop (which compares datetimes) was checking the expiry against the real current time. Hence leases that had seemingly expired over 40 years ago :scream:


itamarst:
Good catch, thanks.

PR #1963, opened 2015-09-22
[FLOC-3093] flocker 1.4.0.dev2

ryao:
All buildbot tests passed.

`docs/releasenotes/index.rst` does not appear to need any changes.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,16 @@
+Flocker 1.4.0.dev2 (2015-09-17)
+===============================
+
+- Flocker container agent systemd unit no longer requires docker
+- Changed api.crt/key to plugin.crt/key in docker plugin (FLOC-2973)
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Many of the tests are failing, some because of the force push. However, because the description says that the tests pass, and the only change is to NEWS, I'll let this pass. Please don't do any more force pushes to Flocker - the commit history looking clean is of very little value. A history of which tests passed when is more valuable.

As this is a dev release, these notes will be looked at at the time of the next marketing release. I will comment but in the interest of spending < 2 days on this release there is no need to address these comments, just note them for next time:

The NEWS has an audience of people who want to know details about Flocker and what has changed for them as users. "installer fixes" for example isn't useful. This is a judgment call but only put what you would like to read as a user.

Be consistent with the style of the document. This is almost there but there are some minor details:
* Periods at the ends of sentences,
* "docker" is "Docker"

I've been wanting to get rid of NEWS for a while, and since forever we've had an issue to document how it should be done.

Please continue with the release.

PR #1966, opened 2015-09-22
Run acceptance tests floc 3089

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3089

I have tested this by successfully running acceptance tests against Vagrant.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1966)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -30,9 +30,6 @@ from admin.vagrant import vagrant_version
 from flocker.common.version import make_rpm_version
 from flocker.provision import PackageSource, Variants, CLOUD_PROVIDERS
 import flocker
-from flocker.node.agents.test.blockdevicefactory import (
-    MINIMUM_ALLOCATABLE_SIZES,
-)
 from flocker.provision._ssh import (
     run_remotely,
     ensure_agent_has_ssh_key,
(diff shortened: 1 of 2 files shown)


tomprince:
A better solution would be to allow not specifying a size, if the backend doesn't require it (like ZFS). But that is much more invasive change that isn't warranted at this point.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[admin/acceptance.py, line 430 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1966#-JzstIdqJdlyWpip7hmn)** ([raw file](https://github.com/clusterhq/flocker/blob/86f808cf631fcb026cb475befa6759b090823618/admin/acceptance.py#L430)):</sup>
Since vagrant doesn't support the block-device backends, using `blockdevice.MINIMUM_ALLOCATABLE_SIZES` seems silly here. I think it is better to just hard-code 1G here with a note that the backend will always be ZFS, or something along those lines. While it would be nice to automatically pick a size from elsewhere, I don't think making vagrant work with a block-device backend is likely, this specific flexibility here is illusory.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1966)

PR #1968, opened 2015-09-22
[FLOC-1796] Fix problem where log messages were being wrapped

itamarst:
By logging directly using the journald APIs we can avoid a bug where log messages are line wrapped on 2048 byte boundaries. This means log messages will now be much easier to process with tools like `eliot-tree`.

The `sleep` in the tests is problematic; not sure what else to do about fact that logging takes a little bit of time to show up in `journalctl` output. Maybe retry a few times?

### DIFF
--- a/flocker/common/_journald.py
+++ b/flocker/common/_journald.py
@@ -1,0 +1,26 @@
+"""
+Minimal journald logging support.
+
+Eventually should be moved into eliot with more extensive functionality:
+https://github.com/ClusterHQ/eliot/issues/207
+"""
+
+from cffi import FFI
+
+_ffi = FFI()
+_ffi.cdef("""
+int sd_journal_send(const char *format, ...);
+""")
+_journald = _ffi.dlopen("libsystemd-journal.so")
+
+
+def sd_journal_send(message):
+    """
+    Send a message to the journald log.
+
+    :param bytes message: Some bytes to log.
+    """
+    # The function uses printf formatting, so we need to quote
+    # percentages.
+    _journald.sd_journal_send(b"MESSAGE=" + message.replace(b"%", b"%%"),
+                              _ffi.NULL)
--- a/flocker/common/_journald.py
+++ b/flocker/common/_journald.py
@@ -11,7 +11,7 @@ _ffi = FFI()
 _ffi.cdef("""
 int sd_journal_send(const char *format, ...);
 """)
-_journald = _ffi.dlopen("libsystemd-journal.so")
+_journald = _ffi.dlopen("libsystemd-journal.so.0")
 
 
 def sd_journal_send(message):
--- a/flocker/common/functional/test_script.py
+++ b/flocker/common/functional/test_script.py
@@ -5,12 +5,11 @@ Functional tests for ``flocker.common.script``.
 from __future__ import print_function
 
 import os
-import time
 import sys
 from json import loads
 from signal import SIGINT
 from unittest import skipUnless
-from subprocess import check_output, CalledProcessError, STDOUT
+from subprocess import check_output, CalledProcessError, STDOUT, Popen
 
 from bitmath import MiB
 
@@ -27,7 +26,7 @@ from twisted.python.filepath import FilePath
 from twisted.python.procutils import which
 
 from ..script import ICommandLineScript
-from ...testtools import random_name, if_root
+from ...testtools import random_name, if_root, loop_until
 
 
 def _journald_available():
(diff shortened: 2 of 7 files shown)


itamarst:
I should add that longer term plan is to move this into Eliot and make it both less hacky and more useful (so e.g. you have ability to find traceback and failed action messages based on priority, and filter by task UUID). This is just enough work to make messages non-corrupt.

flocker/common/_journald.py
```diff
@@ -0,0 +1,26 @@
...
+def sd_journal_send(message):
+    """
+    Send a message to the journald log.
+
+    :param bytes message: Some bytes to log.
+    """
+    # The function uses printf formatting, so we need to quote
+    # percentages.
+    _journald.sd_journal_send(b"MESSAGE=" + message.replace(b"%", b"%%"),
+                              _ffi.NULL)
```
exarkun: There's still a limit, right?  Maybe worth documenting or emitting a warning if the message exceeds it or something like that.

flocker/common/functional/test_script.py
```diff
@@ -272,3 +286,62 @@ def verify_logfiles(stdout_messages, logfile):
...
+    def test_journald_large_message(self):
+        """
+        When ``--journald`` option is used large messages end up in the
+        journal and are not broken up, demonstrating we're not hitting
+        https://bugs.freedesktop.org/show_bug.cgi?id=86465
+        """
+        d = self.run_journald_script(EliotLargeScript)
+        d.addCallback(lambda messages: assertContainsFields(
+            self, messages[1], {u"key": EliotLargeScript.key}))
+        return d
```
exarkun: Do these two tests need to run a separate script?  Couldn't the test process call the journal send function?

flocker/common/script.py
```diff
@@ -22,6 +22,12 @@
 
 from .. import __version__
 
+try:
+    from ._journald import sd_journal_send
+except OSError:
+    # This platform doens't have journald.
+    pass
```
exarkun: This will make `--journald` behavior on non-journald-having platforms pretty unfriendly.  Rejecting `--journald` in this case with a message would be nice.

exarkun:
Thanks.  This will be really nice to have.  Just minor comments inline to address - but buildbot also seems really angry.  Please address those failures and re-submit.

flocker/common/_journald.py
itamarst: Possibly, but if so I haven't reached it (I believe I went up to 800KB and it still worked.) I imagine no one will write out messages *that* big. I'll see if I can figure it out, though.

flocker/common/functional/test_script.py
itamarst: Ideally we'd have separate tests, instead of just end-to-end tests, yeah... but we do need to test `--journald` option. Given final location of code is in Eliot, which will test the low-level bits on its own, my thought was we should just do the minimal amount of needed testing and make it end-to-end so it's realistic.

itamarst:
Hopefully that'll fix buildbot. Will resubmit once initial unit test runs are done.

itamarst:
Hm, `centos` is failing with "ERROR: Journald unavailable on this machine."... Guess it's not fixed yet.

flocker/common/functional/test_script.py
```diff
@@ -84,23 +128,13 @@ def run_script(self, script, options=None):
...
-
-FlockerScriptRunner({}(), StandardOptions()).main()
-'''.format(script.__name__, script.__name__)
+        code = _SCRIPT_CODE.format(script.__name__, script.__name__)
         d = getProcessOutput(
             sys.executable, [b"-c", code] + options,
             env=os.environ,
             errortoo=True
         )
+        d.addCallback(lambda data: (msg(b"script output: " + data), data)[1])
```
tomprince: He He.https://twistedmatrix.com/trac/ticket/6292

flocker/common/functional/test_script.py
```diff
@@ -272,3 +306,94 @@ def verify_logfiles(stdout_messages, logfile):
...
+            b"systemd-run", [b"--unit=" + name, b"--description=testing",
+                             sys.executable, b"-u", b"-c", code,
+                             b"--journald"],
+            env=os.environ, errortoo=True,
+        )
+        d.addCallback(lambda result: msg("systemd-run output: " + result))
+        # systemd-run doesn't wait for process to exit, so we need to ask
+        # systemd when it's done:
+        d.addCallback(lambda _: loop_until(lambda: Popen(
+            [b"systemctl", b"-q", b"is-active", name]).wait() != 0))
```
tomprince: The mix of `Popen` and `getProcessOutput` is saddening.

flocker/common/functional/test_script.py
```diff
@@ -272,3 +306,94 @@ def verify_logfiles(stdout_messages, logfile):
...
+
+
+class JournaldOptionsTests(TestCase):
+    """
+    Tests for the ``--journald`` option.
+    """
+    # _journald_available() is too strong of an assertion; a non-root user
+    # on system with journald installed will get False but this test
+    # shouldn't be run in that case.
+    @skipIf(which("journalctl"), "Journald is available on this machine.")
```
tomprince: How come this uses a different test than above?

tomprince:
Merge once buildbot is happy.

flocker/common/functional/test_script.py
tomprince: But probably not worth changing.

flocker/common/functional/test_script.py
itamarst: See comment above it.

PR #1969, opened 2015-09-22
Remove news, do release notes per branch

adamtheturtle:
Design for https://clusterhq.atlassian.net/browse/FLOC-3094

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1969)


adamtheturtle:
I kept the first Buildbot build in because we've been having trouble with release branches recently, but in theory we're branching from known-good branches and then only potentially changing the license. I think that step can go.

adamtheturtle:
> I'm thinking of something like twisted's topfiles.

We've had an issue for using something like Newsbuilder for ~1 year (https://clusterhq.atlassian.net/browse/FLOC-523). My thinking is that we can wait until that is prioritised (never?) or do this as a part-way point. The release notes file is so short (~150 lines) that I think we can put a tiny bit of effort in when we modify it, and choose to use something like topfiles if it becomes worthwhile.

> That works for the documentation built for the release, but it will become incorrect as soon as any other commits are made.

What do you think of something which just replaces a marker in the text (like the LICENSE script) with the version being released if it is a marketing release and has a comment for the next author to add that marker?

^ Is the design I'd like reviewed

adamtheturtle:
In summary, I think that the benefits of not doing this in the release process, and then later when we have sorted out the branching removing the build step, and then past that allowing our CI system to do a release, outweigh the downsides of a little bit of hassle when changing the release notes a little bit. And if that hassle gets bigger we can try something like topfiles.

tomprince:
Having just done the part of the release that involves creating the NEWS and release notes,
- I don't think that it is a significant chunk of time (~30m or so)
- I ended up doing editing of the NEWS and release notes to give them a (somewhat) unified feel. I'm not sure that this can be done in advance.
- We aren't currently consistent about adding appropriate release notes.

Given the above points, I don't think it is appropriate *at this time* to remove this step from the release process. If we go several major releases where we don't need to update the release notes, we can remove the step then.
---
I do suspect that there is likely nobody that cares about reading `NEWS`, so I wouldn't be opposed to removing that. (One minor counter point is that I found constructing `NEWS` helpful in constructing the release notes).

Regarding topfiles, I suggested that not because of the annoyance of ordering things, but because using them allows our CI to enforce that there are release notes for every change, or specific acknowledgement that no release notes are needed.

One thing that occurs to me, we don't actually need tool-support for creating release notes out of top-files, to create them and use them in the release process and CI.

adamtheturtle:
Thanks for the feedback @tomprince .

> I do suspect that there is likely nobody that cares about reading NEWS, so I wouldn't be opposed to removing that.

Would you be open to removing NEWS now?

> Having just done the part of the release that involves creating the NEWS and release notes, I don't think that it is a significant chunk of time (~30m or so)

I also can do this quickly. CI cannot do it quickly, and I'd like to move towards having a release done mostly by a CI system.
Also, people doing the release for the first time take longer, particularly with the review process back and forth if changes are required.

I think that 30 mins per release is a not insignificant amount of time.

I guess the options now are:
* Ditch this for now
* Remove NEWS and ditch the rest of this for now
* Build a system with a Newsbuilder-like tool
* Use topfiles as a start, and later we can consider a Newsbuilder-like tool

I'd rather not do the first but will let you choose.

tomprince:
> Would you be open to removing NEWS now?

Yes.

Regarding the other, I'm not opposed to doing it now, but it should be done separately from NEWS.

PR #1970, opened 2015-09-23
[FLOC-3090] Implement back-pressure on agents sending updates.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3090

This is the expedient solution. One issue with it, is that it delays convergence iterations after this one until after the control node acknowledges the state update.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1970)

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -356,13 +356,6 @@ class ConvergenceLoop(object):
         :type state_changes: tuple of IClusterStateChange
         """
         if self._last_acknowledged_state != state_changes:
-            # XXX If a prior attempt to send the state is still in progress,
-            # it's not a great idea to initiate a new attempt.  The control
-            # service should respond to requests in order, though, so it may
-            # not be catastrophic.  At the very least, the comparison above
-            # doesn't account for this case.  The last acknowledged state may
-            # be stale while state in the process of being sent could be fully
-            # up-to-date.
             return self._send_state_to_control_service(state_changes)
         else:
             return succeed(None)
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -28,11 +28,13 @@ from machinist import (
 
 from twisted.application.service import MultiService
 from twisted.python.constants import Names, NamedConstant
+from twisted.internet.defer import succeed
 from twisted.internet.protocol import ReconnectingClientFactory
 from twisted.protocols.tls import TLSMemoryBIOFactory
 
 from . import run_state_change
 
+from ..common import gather_deferreds
 from ..control import (
     NodeStateCommand, IConvergenceAgent, AgentAMP,
 )
@@ -343,7 +345,7 @@ class ConvergenceLoop(object):
             d.addErrback(
                 writeFailure, self.fsm.logger,
                 u"Failed to send local state to control node.")
-            d.addActionFinish()
+            return d.addActionFinish()
 
     def _maybe_send_state_to_control_service(self, state_changes):
         """
@@ -361,7 +363,9 @@ class ConvergenceLoop(object):
             # doesn't account for this case.  The last acknowledged state may
             # be stale while state in the process of being sent could be fully
             # up-to-date.
-            self._send_state_to_control_service(state_changes)
+            return self._send_state_to_control_service(state_changes)
+        else:
+            return succeed(None)
 
     def output_CONVERGE(self, context):
         known_local_state = self.cluster_state.get_node(
@@ -386,18 +390,19 @@ class ConvergenceLoop(object):
 
             # XXX And for this update to be the side-effect of an output
             # resulting.
-            self._maybe_send_state_to_control_service(state_changes)
+            send_ack = self._maybe_send_state_to_control_service(state_changes)
 
             action = self.deployer.calculate_changes(
                 self.configuration, self.cluster_state
             )
             LOG_CALCULATED_ACTIONS(calculated_actions=action).write(
                 self.fsm.logger)
-            return run_state_change(action, self.deployer)
+            state_ran = run_state_change(action, self.deployer)
+            # If an error occurred we just want to log it and then try
+            # converging again; hopefully next time we'll have more success.
+            state_ran.addErrback(writeFailure, self.fsm.logger, u"")
+            return gather_deferreds([send_ack, state_ran])
         d.addCallback(got_local_state)
-        # If an error occurred we just want to log it and then try
-        # converging again; hopefully next time we'll have more success.
-        d.addErrback(writeFailure, self.fsm.logger, u"")
 
         # It would be better to have a "quiet time" state in the FSM and
         # transition to that next, then have a timeout input kick the machine
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -390,18 +390,21 @@ class ConvergenceLoop(object):
 
             # XXX And for this update to be the side-effect of an output
             # resulting.
-            send_ack = self._maybe_send_state_to_control_service(state_changes)
+            sent_state = self._maybe_send_state_to_control_service(
+                state_changes)
 
             action = self.deployer.calculate_changes(
                 self.configuration, self.cluster_state
             )
             LOG_CALCULATED_ACTIONS(calculated_actions=action).write(
                 self.fsm.logger)
-            state_ran = run_state_change(action, self.deployer)
-            # If an error occurred we just want to log it and then try
-            # converging again; hopefully next time we'll have more success.
-            state_ran.addErrback(writeFailure, self.fsm.logger, u"")
-            return gather_deferreds([send_ack, state_ran])
+            ran_state_change = run_state_change(action, self.deployer)
+            DeferredContext(ran_state_change).addErrback(
+                writeFailure, self.fsm.logger)
+
+            # Wait for the control node to acknowledge the new
+            # state, and for the convergence actions to run.
+            return gather_deferreds([sent_state, ran_state_change])
         d.addCallback(got_local_state)
 
         # If an error occurred we just want to log it and then try
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -332,7 +332,14 @@ class ConvergenceLoop(object):
             def record_acknowledged_state(ignored):
                 self._last_acknowledged_state = state_changes
 
-            d.addCallback(record_acknowledged_state)
+            def clear_acknowledged_state(failure):
+                # We don't know if the control service has processed the update
+                # or not. So we clear the last acknowledged state so that we
+                # always send the state on the next iteration.
+                self._last_acknowledged_state = None
+                return failure
+
+            d.addCallbacks(record_acknowledged_state, clear_acknowledged_state)
             d.addErrback(
                 writeFailure, self.fsm.logger,
                 u"Failed to send local state to control node.")
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -404,6 +404,10 @@ class ConvergenceLoop(object):
             return gather_deferreds([send_ack, state_ran])
         d.addCallback(got_local_state)
 
+        # If an error occurred we just want to log it and then try
+        # converging again; hopefully next time we'll have more success.
+        d.addErrback(writeFailure, self.fsm.logger)
+
         # It would be better to have a "quiet time" state in the FSM and
         # transition to that next, then have a timeout input kick the machine
         # back around to the beginning of the loop in the FSM.  However, we're
--- a/flocker/node/test/test_loop.py
+++ b/flocker/node/test/test_loop.py
@@ -491,7 +491,7 @@ class ConvergenceLoopFSMTests(SynchronousTestCase):
         self.assertTupleEqual(
             (deployer.calculate_inputs, client.calls),
             (
-                # Check that the loop has run twice
+                # Check that the loop has run thrice
                 [(local_state, configuration, state),
                  (changed_local_state, configuration, changed_state),
                  (local_state, configuration, state)],
--- a/flocker/node/test/test_loop.py
+++ b/flocker/node/test/test_loop.py
@@ -20,7 +20,7 @@ from twisted.internet.task import Clock
 from twisted.internet.ssl import ClientContextFactory
 from twisted.protocols.tls import TLSMemoryBIOFactory, TLSMemoryBIOProtocol
 
-from ...testtools import FakeAMPClient
+from ...testtools import FakeAMPClient, DelayedAMPClient
 from .._loop import (
     build_cluster_status_fsm, ClusterStatusInputs, _ClientStatusUpdate,
     _StatusUpdate, _ConnectedToControlService, ConvergenceLoopInputs,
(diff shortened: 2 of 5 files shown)


flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,55 @@ def test_convergence_sent_state_fail_resends(self):
...
+        configuration = Deployment(nodes=[to_node(local_state)])
+        state = DeploymentState(nodes=[local_state])
+        changed_state = DeploymentState(nodes=[changed_local_state])
+        deployer = ControllableDeployer(
+            local_state.hostname,
+            [succeed(local_state), succeed(changed_local_state),
+             succeed(local_state)],
+            [no_action(), no_action(), no_action()])
+        client = self.make_amp_client(
+            [local_state, changed_local_state.copy()],
```
wallrj: :question: It's immutable, so maybe `copy` isn't necessary here. @exarkun and I discussed this in our branch yesterday.

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,55 @@ def test_convergence_sent_state_fail_resends(self):
...
+            successes=[True, False],
+        )
+        reactor = Clock()
+        loop = build_convergence_loop_fsm(reactor, deployer)
+        loop.receive(_ClientStatusUpdate(
+            client=client, configuration=configuration, state=state))
+        reactor.advance(1.0)
+        reactor.advance(1.0)
```
wallrj: :question: @dwgebler here's another test that might fail if this gets merged before the slower loop branch.

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,55 @@ def test_convergence_sent_state_fail_resends(self):
...
+        state = DeploymentState(nodes=[local_state])
+        changed_state = DeploymentState(nodes=[changed_local_state])
+        deployer = ControllableDeployer(
+            local_state.hostname,
+            [succeed(local_state), succeed(changed_local_state),
+             succeed(local_state)],
+            [no_action(), no_action(), no_action()])
+        client = self.make_amp_client(
+            [local_state, changed_local_state.copy()],
+            successes=[True, False],
```
wallrj: :question: This is another key part of this test, right? The second AMP request fails. Maybe worth a comment.

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,55 @@ def test_convergence_sent_state_fail_resends(self):
...
+        changed_local_state = local_state.set(
+            used_ports=pset([80]),
+        )
+        configuration = Deployment(nodes=[to_node(local_state)])
+        state = DeploymentState(nodes=[local_state])
+        changed_state = DeploymentState(nodes=[changed_local_state])
+        deployer = ControllableDeployer(
+            local_state.hostname,
+            [succeed(local_state), succeed(changed_local_state),
+             succeed(local_state)],
```
wallrj: :question: Why do we have `changed_state`. Shouldn't we be priming the `ControllableDeployer` with identical states for this test? Does it matter that the un-acknowledged state was different?

flocker/node/test/test_loop.py
```diff
@@ -557,6 +611,38 @@ def test_convergence_done_delays_new_iteration(self, logger):
              [(NodeStateCommand, dict(state_changes=(local_state,)))])
         )
 
+    def test_convergence_done_delays_new_iteration_ack(self):
+        """
+        An FSM completing the changes from one convergence iteration doesn't
+        start another iteration, if the control node hasn't acknowledged the
+        last state update.
```
wallrj: :question: I need some commentary comments in this test.

tomprince:
<sup>**[flocker/node/_loop.py, line 405 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-Jzthkhi-9afY7rwQAlI)** ([raw file](https://github.com/clusterhq/flocker/blob/42b292dbb097d7d653e9d3151356dbe4bccb7c66/flocker/node/_loop.py#L405)):</sup>
There appears to be something wrong with the logging changes, but I'm unsure what. It appears like an error from `run_state_change` is now being written to twisted (instead of eliot) logging.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1970)

flocker/node/test/test_loop.py
```diff
@@ -557,6 +611,38 @@ def test_convergence_done_delays_new_iteration(self, logger):
...
+        self.cluster_state = received_state = DeploymentState(nodes=[])
+        self.action = action = ControllableAction(result=succeed(None))
+        deployer = ControllableDeployer(
+            local_state.hostname, [succeed(local_state)], [action]
+        )
+        client = self.make_amp_client([local_state])
+        reactor = Clock()
+        loop = build_convergence_loop_fsm(reactor, deployer)
+        loop.receive(_ClientStatusUpdate(
+            client=DelayedAMPClient(client), configuration=configuration,
```
wallrj: :question: I don't understand the use of `DelayedAMPClient` here. We never seem to call its `respond` method. Perhaps that's the point, but the fact that it has a `respond` method that we don't use in this branch is a bit confusing. Add a few explanatory comments.

wallrj:
Thanks @tomprince 

I understand the code changes, but I struggled to understand the new tests.

Please add a few explanatory comments and resubmit.

Meanwhile, I'm trying to benchmark the CPU usage of flocker-control with the packages from this branch.

tomprince:
Review status: 0 of 5 files reviewed at latest revision, 6 unresolved discussions, some commit checks pending.

---

<sup>**[flocker/node/test/test_loop.py, line 456 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8c-r1-456)** ([raw file](https://github.com/clusterhq/flocker/blob/42b292dbb097d7d653e9d3151356dbe4bccb7c66/flocker/node/test/test_loop.py#L456)):</sup>
I guess it perhaps isn't necessary. But it better matches the actual situation where clearing the last acknowledged state is necessary.

1. Agent sends original state to control node, which records and acknowledges it.
2. Agent sends changed state to control node, which records it, but errors out before acknowledging it.
3. State returns to original state. If we don't clear the acknowledged state, the agent wont send a state update, but the control node will think the state is still the changed state.

---

<sup>**[flocker/node/test/test_loop.py, line 631 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8c-r1-631)** ([raw file](https://github.com/clusterhq/flocker/blob/42b292dbb097d7d653e9d3151356dbe4bccb7c66/flocker/node/test/test_loop.py#L631)):</sup>
I had originally had a test that called `respond`, but realized that it was redundant. I didn't bother to remove `respond`.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1970)

flocker/node/_loop.py
```diff
@@ -332,11 +334,18 @@ def _send_state_to_control_service(self, state_changes):
             def record_acknowledged_state(ignored):
                 self._last_acknowledged_state = state_changes
 
-            d.addCallback(record_acknowledged_state)
+            def clear_acknowledged_state(failure):
+                # We don't know if the control service has processed the update
+                # or not. So we clear the last acknowledged state so that we
+                # always send the state on the next iteration.
+                self._last_acknowledged_state = None
+                return failure
```
exarkun: I think this is unrelated to the optimization.  It would have been nice to see this (and the related test changes) in a different PR so I could think about it independently.

flocker/node/_loop.py
```diff
@@ -354,7 +363,9 @@ def _maybe_send_state_to_control_service(self, state_changes):
             # doesn't account for this case.  The last acknowledged state may
             # be stale while state in the process of being sent could be fully
             # up-to-date.
-            self._send_state_to_control_service(state_changes)
+            return self._send_state_to_control_service(state_changes)
```
exarkun: Note the long comment immediately above this line.

flocker/node/_loop.py
```diff
@@ -379,18 +390,26 @@ def got_local_state(state_changes):
...
 
             action = self.deployer.calculate_changes(
                 self.configuration, self.cluster_state
             )
             LOG_CALCULATED_ACTIONS(calculated_actions=action).write(
                 self.fsm.logger)
-            return run_state_change(action, self.deployer)
+            ran_state_change = run_state_change(action, self.deployer)
+            DeferredContext(ran_state_change).addErrback(
+                writeFailure, self.fsm.logger)
```
exarkun: I think it would be beneficial to supply a system here.  `DeferredContext` makes sure `writeFailure` is run in the context of the ... `LOG_CONVERGE` action (I think?  at the very least, whatever action is on active here) but it won't associate the failure with the `run_state_change` operation.  I think making this association is the purpose of passing a system to `writeFailure`.

flocker/node/_loop.py
```diff
@@ -379,18 +390,26 @@ def got_local_state(state_changes):
...
+
+            # Wait for the control node to acknowledge the new
+            # state, and for the convergence actions to run.
+            return gather_deferreds([sent_state, ran_state_change])
         d.addCallback(got_local_state)
+
         # If an error occurred we just want to log it and then try
         # converging again; hopefully next time we'll have more success.
-        d.addErrback(writeFailure, self.fsm.logger, u"")
+        d.addErrback(writeFailure, self.fsm.logger)
```
exarkun: Likewise here (who put `u""` there :frowning: )

flocker/node/test/test_loop.py
```diff
@@ -267,23 +268,27 @@ def test_new_status_update_starts_discovery(self):
...
 
         :param local_states: The node states we expect to be able to send.
-        :param bool succeed: ``True`` to make a client which responds to
-            requests with results, ``False`` to make a client which response
-            with failures.  succeed or fail.
+        :param successes: List indicating whether the response to the
+            corresponding states should fail.  ``True`` to make a client which
+            responds to requests with results, ``False`` to make a client which
+            response with failures. Defaults to always succeeding.
+        :type successes: ``list`` of ``bool`` or ``None``
```
exarkun: "None or list of bool" is less ambiguous

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
             )
         )
 
+    def test_convergence_sent_state_fail_resends_alternating(self):
+        """
+        If sending state to the control node fails the next iteration will send
+        state even if the state is the same as the last acknowledge state.
+
+        The situation this is intnded to model is the following sequence:
```
exarkun: typo

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
...
+        1. Agent sends original state to control node, which records and
+           acknowledges it.
+        2. Agent sends changed state to control node, which records it, but
+           errors out before acknowledging it.
+        3. State returns to original state. If we don't clear the acknowledged
+           state, the agent wont send a state update, but the control node
```
exarkun: `won't`

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
...
+           will think the state is still the changed state.
+        """
+        local_state = NodeState(
+            hostname=u'192.0.2.123',
```
exarkun: `hostname` without `uuid` is deprecated

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
...
+        # Wait for all three iterations to occure.
+        reactor.advance(1.0)
+        reactor.advance(1.0)
+
+        # Calculating actions happened, result was run... and then we did
+        # whole thing again:
+        self.assertTupleEqual(
+            (deployer.calculate_inputs, client.calls),
+            (
+                # Check that the loop has run twice
```
exarkun: Looks like three times

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
...
+            # local_state will be acknowledge
+            # changed_local_state will result in an error.
+            successes=[True, False],
+        )
+        reactor = Clock()
+        loop = build_convergence_loop_fsm(reactor, deployer)
+        loop.receive(_ClientStatusUpdate(
+            client=client, configuration=configuration, state=state))
+
+        # Wait for all three iterations to occure.
```
exarkun: `occure`

flocker/node/test/test_loop.py
```diff
@@ -550,7 +623,44 @@ def test_convergence_done_delays_new_iteration(self, logger):
...
+        self.assertTupleEqual(
+            (deployer.calculate_inputs, client.calls),
+            ([(local_state, configuration, expected_cluster_state)],
+             [(NodeStateCommand, dict(state_changes=(local_state,)))])
+        )
+
+    def test_convergence_done_delays_new_iteration_ack(self):
+        """
+        An FSM completing the changes from one convergence iteration doesn't
+        start another iteration, if the control node hasn't acknowledged the
```
exarkun: spurious `,`

flocker/node/test/test_loop.py
exarkun: I think the really important intent of this test is that a state update isn't sent if the control node hasn't acked the last state update.  The fact that we do this by delaying the entire subsequent iteration is an implementation detail and probably one we want to change (since introducing a round-trip to the control service for every step of convergence is going to slow down the already very slow convergence process - particularly for the multistep convergence operations like dataset creation).  So the docstring should be clear about this intent and perhaps note the expedient implementation decision that was made (bonus would be if the test also actually made an assertion directly tied to that intent rather than being based on the implementation details as well).

flocker/testtools/test/test_amp.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    arguments = [
+        ('argument', Integer()),
+    ]
+    response = [
+        ('response', Integer()),
+    ]
+
+
+class DelayedAMPClientTests(SynchronousTestCase):
```
exarkun: More unverified fakes.  Bummer.  Maybe someday we can stop writing such things and even come back and fix up these things we're accumulating.

flocker/node/test/test_loop.py
```diff
@@ -430,6 +435,74 @@ def test_convergence_sent_state_fail_resends(self):
...
+            # changed_local_state will result in an error.
+            successes=[True, False],
+        )
+        reactor = Clock()
+        loop = build_convergence_loop_fsm(reactor, deployer)
+        loop.receive(_ClientStatusUpdate(
+            client=client, configuration=configuration, state=state))
+
+        # Wait for all three iterations to occure.
+        reactor.advance(1.0)
```
exarkun: This could be self-documenting if it were in a method like `run_next_convergence_iteration(reactor)`.

exarkun:
Thanks.  Please address inline comments and merge when builds are green.

tomprince:
Review status: 0 of 5 files reviewed at latest revision, 8 unresolved discussions, some commit checks pending.

---

<sup>**[flocker/node/_loop.py, line 342 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8b-r2-342)** ([raw file](https://github.com/clusterhq/flocker/blob/bda5fce3c5f2965a4bcc6f52ba06da7cecb74bc7/flocker/node/_loop.py#L342)):</sup>
The seemed linked in my head, when I started on this. On reflection, they probably could have been split. (Or, it could still be split.)

---

<sup>**[flocker/node/_loop.py, line 403 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8b-r3-403)** ([raw file](https://github.com/clusterhq/flocker/blob/730aa322a0c2d764d9e606dfbd1afb188437b00a/flocker/node/_loop.py#L403)):</sup>
`system` is deprecated and a noop.

---

<sup>**[flocker/node/_loop.py, line 405 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-Jzthkhi-9afY7rwQAlI)** ([raw file](https://github.com/clusterhq/flocker/blob/42b292dbb097d7d653e9d3151356dbe4bccb7c66/flocker/node/_loop.py#L405)):</sup>
The error was coming from failed discovery, which was crashing the loop.

---

<sup>**[flocker/node/test/test_loop.py, line 453 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8c-r3-453)** ([raw file](https://github.com/clusterhq/flocker/blob/730aa322a0c2d764d9e606dfbd1afb188437b00a/flocker/node/test/test_loop.py#L453)):</sup>
But `ControllableDeployer` hard-codes the derived UUID. 

---

<sup>**[flocker/node/test/test_loop.py, line 456 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8c-r1-456)** ([raw file](https://github.com/clusterhq/flocker/blob/42b292dbb097d7d653e9d3151356dbe4bccb7c66/flocker/node/test/test_loop.py#L456)):</sup>
Actually, it is necessary. We need to try to send a second state, which will only occur if the state has changed.

---

<sup>**[flocker/node/test/test_loop.py, line 485 \[r3\]](https://reviewable.io:443/reviews/clusterhq/flocker/1970#-JztQ7CTtSHBzm68aN8c-r3-485)** ([raw file](https://github.com/clusterhq/flocker/blob/730aa322a0c2d764d9e606dfbd1afb188437b00a/flocker/node/test/test_loop.py#L485)):</sup>
I'll leave that for the branch that adjusts the iteration length.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1970)

PR #1971, opened 2015-09-23
[FLOC-3104] Configure acceptance testing nodes in parallel

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3104


tomprince:
Merge when buildbot is green.

PR #1972, opened 2015-09-23
[FLOC-3100] Memoize wire encode when sending updates to agents

itamarst:
Instead of serializing config and state to JSON once for each destination, we only do it once.

The actual end-to-end result of this (lower CPU usage, i.e. the fact that we have caching enabled in `_send_state_to_connections`) is not tested, it's not clear to me how one would given current abstractions. Ideas welcome.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1972)

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -55,6 +55,38 @@ from ._model import (
 PING_INTERVAL = timedelta(seconds=30)
 
 
+class CachingEncoder(object):
+    """
+    Cache results of ``wire_encode`` and re-use them, relying on the fact
+    we're encoding immutable objects.
+
+    Not thread-safe, so should only be used by Twisted code (or this
+    module).
+    """
+    def __init__(self):
+        self.clear()
+
+    def encode(self, obj):
+        """
+        Encode an object to bytes using ``wire_encode``, or return cached
+        result if available.
+
+        :param obj: Object to encode.
+        :return: Resulting ``bytes``.
+        """
+        if obj not in self.results:
+            self.results[obj] = wire_encode(obj)
+        return self.results[obj]
+
+    def clear(self):
+        """
+        Clear the cache.
+        """
+        self.results = {}
+
+_caching_encoder = CachingEncoder()
+
+
 class SerializableArgument(Argument):
     """
     AMP argument that takes an object that can be serialized by the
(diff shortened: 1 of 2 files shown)


tomprince:
> The object is pretty much always immutable.

That "pretty much" is slightly worrying. One thought, is that SerializableArgument should check what kind of objects it can take, and make sure they are immutable? Probably over thinking things, though.

---

This looks good, just a couple of cosmetic changes below.. If the benchmarks look good and buildbot is green, go ahead and merge.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[flocker/control/_protocol.py, line 95 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1972#-JzvsSrEr83YytcmbsZ9)** ([raw file](https://github.com/clusterhq/flocker/blob/bb72f188dfe556a55b4a6b8a770f547965e31454/flocker/control/_protocol.py#L95)):</sup>
`:var:`

---

<sup>**[flocker/control/_protocol.py, line 347 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1972#-JzvrhMNqPyy398Pob9V)** ([raw file](https://github.com/clusterhq/flocker/blob/bb72f188dfe556a55b4a6b8a770f547965e31454/flocker/control/_protocol.py#L347)):</sup>
I think maybe we want a comment here about how everything sent via SerializableArgument needs to be immutable here.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1972)

itamarst:
Addressed, can be merged as soon as buildbot is green.

tomprince:
Review status: 1 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/control/_protocol.py, line 95 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1972#-JzvsSrEr83YytcmbsZ9)** ([raw file](https://github.com/clusterhq/flocker/blob/bb72f188dfe556a55b4a6b8a770f547965e31454/flocker/control/_protocol.py#L95)):</sup>
I was actually suggesting that `_caching_encoder` be documented.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1972)

PR #1976, opened 2015-09-24
[FLOC-3108] A big composeable AMP argument type

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3108

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -59,10 +59,29 @@ PING_INTERVAL = timedelta(seconds=30)
 
 
 class Big(Argument):
+    """
+    An ``Argument`` type which can handle objects which are larger than AMP's
+    MAX_VALUE_LENGTH when serialized.
+
+    Thanks to Glyph Lefkowitz for the idea:
+    * http://bazaar.launchpad.net/~glyph/+junk/amphacks/view/head:/python/amphacks/mediumbox.py  # noqa
+    """
     def __init__(self, another_argument):
+        """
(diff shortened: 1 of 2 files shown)


flocker/control/_protocol.py
```diff
@@ -55,6 +58,62 @@
...
+        method of the wrapped ``Argument`` is then called supplied with the
+        updated ``strings`` dictionary, deserializes the large value and
+        populates the ``objects`` dictionary with the result.
+
+        See ``IArgumentType`` for argument and return type documentation.
+        """
+
+        value = BytesIO()
+        for counter in count(0):
+            chunk = strings.get("%s.%d" % (name, counter))
```
wallrj: We could `pop` here instead to avoid having two copies of the bytes in memory.

itamarst: Mutation is a bad idea; we should avoid if possible.

itamarst:
I merged forward to deal with some conflicts.

flocker/control/_protocol.py
itamarst: Moreover this doesn't copy, it just adds another reference to the same string.

itamarst: Python never copies stuff unless you explicitly say so with `.copy()` or `[:]` or `copy.copy` or whatever.

itamarst: Oh, you mean a copy in `BytesIO`. And `strings` is being severely mutated anyway. Probably fine as is.

itamarst:
OK, merge when green.

PR #1977, opened 2015-09-24
[FLOC-3107] Include CPU shares and memory limit in Applications

dwgebler:
Fixes FLOC-3107. Previously we overlooked the memory limit and CPU shares coming back from a Unit when listing container state and didn't include these in the Applications we construct when discovering node state, hence erroneously constructing a model that represented a different state than that desired from the configuration and calculating changes accordingly - so containers were continuously created, destroyed and recreated when using these options.

### DIFF
--- a/flocker/node/_deploy.py
+++ b/flocker/node/_deploy.py
@@ -757,6 +757,8 @@ class ApplicationNodeDeployer(object):
                 volume=volume,
                 environment=environment if environment else None,
                 links=frozenset(links),
+                memory_limit=container.mem_limit,
+                cpu_shares=container.cpu_shares,
                 restart_policy=container.restart_policy,
                 running=(container.activation_state == u"active"),
                 command_line=container.command_line,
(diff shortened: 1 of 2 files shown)


wallrj:
Thanks @dwgebler 

This looks like it'll work.

* I read the docker-py docs and notice that mem_limit might be a string with a unit suffix....but not sure if docker ever reports it that way....and if so I guess it might cause a Pyrsistent type error. http://docker-py.readthedocs.org/en/latest/api/#create_container

* I forced a rebuild of two intermittent build failures.

* The tests seem a bit repetitive, but now's probably not the time to tidy them up.

Please merge when those are green.

dwgebler:
> I read the docker-py docs and notice that mem_limit might be a string with a unit suffix....but not sure if docker ever reports it that way....and if so I guess it might cause a Pyrsistent type error. http://docker-py.readthedocs.org/en/latest/api/#create_container

FWIW docker-py's `inspect_container`, although poorly documented, will always return `Memory` as an integer representing bytes.

PR #1978, opened 2015-09-24
[FLOC-3110] Support big cluster configurations and numbers of nodes

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3110

This allows arbitrarily large (more or less) `Deployment` objects to be passed around as cluster configuration (using the recently introduced `Big` argument type).  It does likewise for `DeploymentState` objects.

Note that arbitrarily large `NodeState` objects passed from agents to the control service in `NodeStateCommand` are not supported.  They have a small serialization quirk and will need a slightly different strategy applied.

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -59,10 +59,29 @@ PING_INTERVAL = timedelta(seconds=30)
 
 
 class Big(Argument):
+    """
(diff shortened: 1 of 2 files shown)


itamarst:
Looks good, merge when buildbot is green.

exarkun:
BuildBot will never be green.  Merging anyway because the red is unrelated.

PR #1982, opened 2015-09-24
[FLOC-3117] Use a wait instead of polling for container stop.

itamarst:


flocker/node/_docker.py
```diff
@@ -710,8 +693,7 @@ def _remove():
                 # attempting removal or use -f")'
                 # This code should probably be removed once the above
                 # issue has been resolved. See [FLOC-1850]
-                while self._blocking_container_runs(container_name):
-                    sleep(0.01)
+                self._client.wait(container_name)
```
adamtheturtle: :question: The description of this mentions a 404 error, but the issue comments describe that it is a 500 error now.

adamtheturtle:
Thanks @itamarst , ideally this would come with some benchmarking numbers, or an explanation of why this would use less CPU. I understand that we are in a rush now so merge when you are confident that this is better for CPU usage and Buildbot is happy.

exarkun:
Still seems to work: I set up an add/remove loop and didn't encounter any cases where Docker was complaining about this new remove logic.  Code makes sense, too, and seems like the most straightforward implementation anyway (if you know about `wait`, at least).  Will merge when buildbot shows me I haven't overlooked something else.

exarkun:
Lots of problems on buildbot.  None clearly attributable to this change.  Merging.

PR #1983, opened 2015-09-24
[FLOC-3118] Instead of polling with an expensive inspect, poll with the action we care about care about

itamarst:


exarkun:
This looks like a good change.  I also increased the sleep in the loop since I don't think 1000 checks per second is a reasonable goal to try to achieve.  If Docker is infinitely fast then we'll now only run this loop 100 times a second (still probably excessive, maybe when we have some real profile/benchmark data we can make an informed decision about this cost/responsiveness trade-off).

If buildbot doesn't throw up any surprises, I think this should be merged (I'll do that later tonight if no one else does it first).

exarkun:
Some failures, none surprising, merging.

PR #1984, opened 2015-09-25
[FLOC-3122] Remove useless test test_in_use_device_log.

myechuri:
This test assumes that '/dev/sda1' is the only device in use at the beginning of the test, which is not true in all test run environments. For example, my EC2 instance had the following devices before any functional tests ran:
```
brw-rw----  1 root disk    202,   0 May 13 22:37 xvda
brw-rw----  1 root disk    202,  80 Jun 11 04:09 xvdf
```

Attempted to cleanup volumes (``detach_destroy_volumes``) at the beginning of the testcase, but realized this is not sufficient for predictable test outcome either.

### DIFF
--- a/flocker/node/agents/functional/test_ebs.py
+++ b/flocker/node/agents/functional/test_ebs.py
@@ -16,7 +16,7 @@ from boto.exception import EC2ResponseError
 
 from twisted.python.constants import Names, NamedConstant
 from twisted.trial.unittest import SkipTest, TestCase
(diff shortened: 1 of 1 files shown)


itamarst:
You can do logging assertions in the body of the test, if it's easier... so this is how I would fix the test:

1. Attach device.
2. Get its path.
3. Assert that the device path is in the logged message.

The only question is whether it's worth the time of doing so

itamarst:
Sadly, probably not, until we're focusing on EBS again.

PR #1986, opened 2015-09-25
[FLOC-3114] Intermittent deployer test failure

dwgebler:
Fixes FLOC-3114 - this should make the two affected tests less flaky by a) ensuring the unit we assert on is the correct unit, rather than assuming it's the only unit and b) checking that an assertion on a unit definitely happened.

### DIFF
--- a/flocker/node/functional/test_deploy.py
+++ b/flocker/node/functional/test_deploy.py
@@ -360,10 +360,11 @@ class DeployerTests(TestCase):
             deferred_list = docker_client.list()
 
             def app_memory(units):
-                [
+                checked = [
                     self.assertEqual(unit.mem_limit, EXPECTED_MEMORY_LIMIT)
                     for unit in units if unit.name == application_name
                 ]
+                self.assertNotEqual(checked, [])
             deferred_list.addCallback(app_memory)
             return deferred_list
         d.addCallback(inspect_application)
@@ -407,10 +408,11 @@ class DeployerTests(TestCase):
             deferred_list = docker_client.list()
 
             def app_cpu_shares(units):
-                [
+                checked = [
                     self.assertEqual(unit.cpu_shares, EXPECTED_CPU_SHARES)
                     for unit in units if unit.name == application_name
                 ]
+                self.assertNotEqual(checked, [])
             deferred_list.addCallback(app_cpu_shares)
             return deferred_list
         d.addCallback(inspect_application)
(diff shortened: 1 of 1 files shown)


flocker/node/functional/test_deploy.py
```diff
@@ -360,11 +360,13 @@ def inspect_application(_):
...
 
             def app_memory(units):
-                unit = units.pop()
-                self.assertEqual(unit.mem_limit, EXPECTED_MEMORY_LIMIT)
-                return deferred_list
-
+                checked = [
+                    self.assertEqual(unit.mem_limit, EXPECTED_MEMORY_LIMIT)
+                    for unit in units if unit.name == application_name
+                ]
```
exarkun: :question: `list(generator expression)` instead of `[list comprehension]` is a good habit to get in to.  Generator expressions don't leak the loop variable into the enclosing scope - and that's preferable almost all the time.  Make list comprehensions the exception rather than the default.

exarkun: It's sort of weird to put the `assertEqual` into this expression.  Instead, I'd collect units with a matching name and make the assertion afterwards.

flocker/node/functional/test_deploy.py
```diff
@@ -405,11 +407,13 @@ def test_cpu_shares(self):
...
-                self.assertEqual(unit.cpu_shares, EXPECTED_CPU_SHARES)
-                return deferred_list
-
-            deferred_list.addCallback(app_memory)
+            def app_cpu_shares(units):
+                checked = [
+                    self.assertEqual(unit.cpu_shares, EXPECTED_CPU_SHARES)
+                    for unit in units if unit.name == application_name
+                ]
+                self.assertNotEqual(checked, [])
```
exarkun: How about a shared callback for filtering `units` down to one matching a particular name?  `deferred_list.addCallback(pick_unit_named, application_name); deferred_list.addCallback(app_cpu_shares)`

Avoids some duplication between these two tests.

flocker/node/functional/test_deploy.py
exarkun: To know if this has had a beneficial effect, it would be great to log something about what containers this new code is filtering out.  A log message with, say, the names of all containers listed would give us something to point at and say "Aha, there were 2 other containers and the `units.pop()` previously picked the wrong one, the assumption is validated and the bug is really fixed."

exarkun:
Thanks.  Minor comments inline.  Please address those and merge.

PR #1988, opened 2015-09-25
[FLOC-3119] Cache inspected image data in memory

dwgebler:
Fixes FLOC-3119

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -242,32 +242,6 @@ class IDockerClient(Interface):
         :return: ``Deferred`` firing with ``set`` of :class:`Unit`.
         """
 
-    def _image_data(image, container, state):
-        """
-        Supply data about an image, by either inspecting it or returning
-        cached data if available.
-
-        :param unicode image: The ID of the image.
-
-        :param unicode container: The ID of the container for which this
-            image is being used.
-
-        :param unicode state: The state of the container.
-
-        :return: ``dict`` representing data about the image properties.
-        """
-
-    def _image_cached_data(image):
-        """
-        Supply data about an image from the IDockerClient's image data
-        cache if available.
-
-        :param unicode image: The ID of the image.
-
-        :return: Either a ``dict`` representing data about the image, or
-            None if the image does not exist in the cache.
-        """
-
 
 def make_response(code, message):
     """
@@ -307,7 +281,6 @@ class FakeDockerClient(object):
             units = {}
         self._units = units
         self._used_ports = pset()
-        self._image_cache = {}
 
     def add(self, unit_name, image_name, ports=frozenset(), environment=None,
             volumes=frozenset(), mem_limit=None, cpu_shares=None,
@@ -360,19 +333,6 @@ class FakeDockerClient(object):
         units = set(self._units.values())
         return succeed(units)
 
-    def _image_data(self, image, container, state):
-        cached_data = self._image_cached_data(image)
-        if cached_data is not None:
-            return cached_data
-        # Fake client does not care about accurate data.
-        image_data = {u"Id": image, u"Config": {u"Env": [], u"Cmd": []}}
-        self._image_cache[image] = image_data
-        return image_data
-
-    def _image_cached_data(self, image):
-        if image in self._image_cache:
-            return self._image_cache[image]
-        return None
 
 # Basic namespace for Flocker containers:
 BASE_NAMESPACE = u"flocker--"
@@ -569,32 +529,48 @@ class DockerClient(object):
         return AddressInUse(address=(ip, int(port)), apierror=apierror)
 
     def _image_data(self, image, container, state):
-        cached_data = self._image_cached_data(image)
-        if cached_data is not None:
-            return cached_data
+        """
+        Supply data about an image, by either inspecting it or returning
+        cached data if available.
+
+        :param unicode image: The ID of the image.
+
+        :param unicode container: The ID of the container for which this
+            image is being used.
+
+        :param unicode state: The state of the container.
+
+        :return: ``dict`` representing data about the image properties.
+        """
+        # This caching technique isn't thread-safe, but allowable
+        # in this case because a) the data is for our purposes
+        # immutable, so two writes of same key isn't a problem and
+        # b) we never remove entries.
+        if image in self._image_cache:
+            return self._image_cache[image]
         try:
             image_data = self._client.inspect_image(image)
-            self._image_cache[image] = image_data
+            self._image_cache[image] = {
+                u"Config": {
+                    u"Cmd": image_data[u"Config"][u"Cmd"],
+                    u"Env": image_data[u"Config"][u"Env"],
+                }
+            }
         except APIError as e:
             if e.response.status_code == NOT_FOUND:
                 # Image has been deleted, so just fill in some
                 # stub data so we can return *something*. This
                 # should happen only for stopped containers so
                 # some inaccuracy is acceptable.
+                # We won't cache stub data though.
                 Message.new(
                     message_type="flocker:docker:image_not_found",
                     container=container, running=state
                 ).write()
                 image_data = {u"Config": {u"Env": [], u"Cmd": []}}
-                self._image_cache[image] = image_data
             else:
                 raise
         return image_data
-
-    def _image_cached_data(self, image):
-        if image in self._image_cache:
-            return self._image_cache[image]
-        return None
 
     def add(self, unit_name, image_name, ports=None, environment=None,
             volumes=(), mem_limit=None, cpu_shares=None,
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -560,10 +560,6 @@ class DockerClient(object):
 
         :return: ``dict`` representing data about the image properties.
         """
-        # This caching technique isn't thread-safe, but allowable
-        # in this case because a) the data is for our purposes
-        # immutable, so two writes of same key isn't a problem and
-        # b) we never remove entries.
         cached_image = self._image_cache.get(image)
         if cached_image is not None:
             LOG_CACHED_IMAGE(image=image).write()
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -242,6 +242,32 @@ class IDockerClient(Interface):
         :return: ``Deferred`` firing with ``set`` of :class:`Unit`.
         """
 
+    def _image_data(image, container, state):
+        """
+        Supply data about an image, by either inspecting it or returning
+        cached data if available.
+
+        :param unicode image: The ID of the image.
+
+        :param unicode container: The ID of the container for which this
+            image is being used.
+
+        :param unicode state: The state of the container.
+
+        :return: ``dict`` representing data about the image properties.
+        """
+
+    def _image_cached_data(image):
+        """
+        Supply data about an image from the IDockerClient's image data
+        cache if available.
+
+        :param unicode image: The ID of the image.
+
+        :return: Either a ``dict`` representing data about the image, or
+            None if the image does not exist in the cache.
+        """
+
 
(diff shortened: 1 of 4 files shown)


flocker/node/_docker.py
```diff
@@ -242,6 +242,32 @@ def list():
         :return: ``Deferred`` firing with ``set`` of :class:`Unit`.
         """
 
+    def _image_data(image, container, state):
```
itamarst: The interface is for things that are common to all implementations. Private methods are by definition not part of a public interface, and moreover these are needed only due to issues with one implementation; the in-memory implementation doesn't need them.

So, take them off the interface.

flocker/node/_docker.py
```diff
@@ -333,6 +360,19 @@ def list(self):
         units = set(self._units.values())
         return succeed(units)
 
+    def _image_data(self, image, container, state):
```
itamarst: The in-memory implementation doesn't need these.

flocker/node/functional/test_docker.py
```diff
@@ -208,6 +208,23 @@ def started(_):
         d.addCallback(started)
         return d
 
+    def test_list_image_data_cached(self):
+        """
+        ``DockerClient.list`` will only an inspect an image ID once, caching
```
itamarst: The test doesn't actually prove that, though. It just proves the second part, that the information is cached.

flocker/node/_docker.py
```diff
@@ -527,6 +568,34 @@ def _address_in_use(self, apierror):
             return None
         return AddressInUse(address=(ip, int(port)), apierror=apierror)
 
+    def _image_data(self, image, container, state):
+        cached_data = self._image_cached_data(image)
+        if cached_data is not None:
+            return cached_data
+        try:
+            image_data = self._client.inspect_image(image)
+            self._image_cache[image] = image_data
```
itamarst: The data will end up using up a bunch of memory if there's lots of images used. I suggest only caching the information we actually care about.

flocker/node/_docker.py
dwgebler: The reason they ended up here is because `NamespacedDockerClient` proxies this interface through its `_client` attribute - have a look at the difference between this and my first commit to the branch and you can get an idea of how I was originally approaching the problem, but that doesn't work for the namespaced client tests. Any suggestions on how best to proceed?

flocker/node/_docker.py
```diff
@@ -527,6 +568,34 @@ def _address_in_use(self, apierror):
...
+                    message_type="flocker:docker:image_not_found",
+                    container=container, running=state
+                ).write()
+                image_data = {u"Config": {u"Env": [], u"Cmd": []}}
+                self._image_cache[image] = image_data
+            else:
+                raise
+        return image_data
+
+    def _image_cached_data(self, image):
```
itamarst: I don't think this method adds much. If the test needs to poke at private details it can poke at the data structure.

flocker/node/_docker.py
```diff
@@ -527,6 +568,34 @@ def _address_in_use(self, apierror):
...
+                # Image has been deleted, so just fill in some
+                # stub data so we can return *something*. This
+                # should happen only for stopped containers so
+                # some inaccuracy is acceptable.
+                Message.new(
+                    message_type="flocker:docker:image_not_found",
+                    container=container, running=state
+                ).write()
+                image_data = {u"Config": {u"Env": [], u"Cmd": []}}
+                self._image_cache[image] = image_data
```
itamarst: We shouldn't cache ignorance permanently.

flocker/node/_docker.py
```diff
@@ -527,6 +568,34 @@ def _address_in_use(self, apierror):
             return None
         return AddressInUse(address=(ip, int(port)), apierror=apierror)
 
+    def _image_data(self, image, container, state):
```
itamarst: A general note on thread-safety: we have potentially multiple threads reading and writing to the dictionary. This is OK, but only because:
1. The data is for our purposes immutable, so two writes of same key isn't a problem.
2. We never remove entries.

You may wish to document these constraints.

flocker/node/functional/test_docker.py
itamarst: A different way to demonstrate this; gross but workable:
1. `DockerClient.list()` to get container's info.
2. Patch underlying Docker client so that `inspect_image` returns (for that particular image) different data or maybe fails.
3. Run `DockerClient.list()` again, note it still has correct info.

itamarst:
Thanks, back to you for some improvements.

flocker/node/_docker.py
```diff
@@ -527,6 +528,50 @@ def _address_in_use(self, apierror):
...
+            return self._image_cache[image]
+        try:
+            image_data = self._client.inspect_image(image)
+            self._image_cache[image] = {
+                u"Config": {
+                    u"Cmd": image_data[u"Config"][u"Cmd"],
+                    u"Env": image_data[u"Config"][u"Env"],
+                }
+            }
+        except APIError as e:
```
exarkun: Ideally only the code that might actually raise the exception being handled should be inside the try block.  The second statement, the statement populating the cache, should be outside the try block, near the `return image_data`.

flocker/node/_docker.py
```diff
@@ -527,6 +528,50 @@ def _address_in_use(self, apierror):
...
+        """
+        Supply data about an image, by either inspecting it or returning
+        cached data if available.
+
+        :param unicode image: The ID of the image.
+
+        :param unicode container: The ID of the container for which this
+            image is being used.
+
+        :param unicode state: The state of the container.
```
exarkun: Neither `container` nor `state` is necessary for the cache.  These parameters are only used for logging.  An alternative that makes this method simpler is to start a "looking up image data" action at the caller and log the container and state there.  Then the log message this method writes about the image not being found is written in a context that provides that information but without this method actually needing to take the extra arguments.

flocker/node/_docker.py
```diff
@@ -527,6 +528,50 @@ def _address_in_use(self, apierror):
...
+
+        :return: ``dict`` representing data about the image properties.
+        """
+        # This caching technique isn't thread-safe, but allowable
+        # in this case because a) the data is for our purposes
+        # immutable, so two writes of same key isn't a problem and
+        # b) we never remove entries.
+        if image in self._image_cache:
+            return self._image_cache[image]
```
exarkun: Cache hit and miss logging would make it possible to evaluate how effective this cache is on a real system.  Cache hit information could also help with debugging - for example, to reveal that the image data is stale because it's being served from a cache.

flocker/node/functional/test_docker.py
```diff
@@ -208,6 +208,51 @@ def started(_):
...
+                docker = Client()
+                data = docker.inspect_container(self.namespacing_prefix + name)
+                # This is kind of nasty, but NamespacedDockerClient represents
+                # its client via a proxying attribute, so we can't assume
+                # client here directly has an _image_cache.
+                if isinstance(client, NamespacedDockerClient):
+                    image_cache = client._client._image_cache
+                else:
+                    image_cache = client._image_cache
+                self.assertIn(data['Image'], image_cache)
```
exarkun: We don't really care about the implementation details of the cache.  Ideally, the only difference the cache makes is that some things get faster.  So, ideally, the best way to justify this change is to demonstrate that the desired outputs are unchanged (ie, it still "works" - hopefully just by passing all of the existing tests) and to demonstrate that it provides those outputs faster than it did before.

Demonstrating things about fastness requires benchmarks.  We've skipped over that step for the past two weeks because of Reasons.  Since we're still lacking any general tool chain or infrastructure for having benchmarks, it's probably okay to skip over it for this change as well.

But what we don't want is new unit "tests" that just reveal optimization-motivated implementation details.  The thing we want to prove is that this is *fast*.  We don't care if there's an `_image_cache`.

Itamar's suggestion for a test, involving patching the Docker client's `inspect_image` method is a sort of compromise between these two.  It drops the assertions this test currently makes about the irrelevant `_image_cache` - that's good.  In place of measuring how fast the old and new code is, though, it supposes that fewer calls to `inspect_image` must mean things are getting faster - so it proves that `inspect_image` is not called in a certain case (where we know it was called before).  It's still not ideal but it is a little better.

flocker/node/_docker.py
```diff
@@ -386,6 +386,7 @@ def __init__(
         self.namespace = namespace
         self._client = TimeoutClient(
             version="1.15", base_url=base_url, long_timeout=long_timeout)
+        self._image_cache = {}
```
exarkun: This should be documented.

flocker/node/_docker.py
```diff
@@ -527,6 +528,50 @@ def _address_in_use(self, apierror):
...
+        # b) we never remove entries.
+        if image in self._image_cache:
+            return self._image_cache[image]
+        try:
+            image_data = self._client.inspect_image(image)
+            self._image_cache[image] = {
+                u"Config": {
+                    u"Cmd": image_data[u"Config"][u"Cmd"],
+                    u"Env": image_data[u"Config"][u"Env"],
+                }
```
exarkun: The calling code wants the command and the environment.  It doesn't need this Docker-invented structure to represent it.  I suggest caching a PClass with `command` and `environment` attributes.  This will make the job of documenting `_image_cache` easier, too.

flocker/node/_docker.py
```diff
@@ -527,6 +528,50 @@ def _address_in_use(self, apierror):
...
+            image is being used.
+
+        :param unicode state: The state of the container.
+
+        :return: ``dict`` representing data about the image properties.
+        """
+        # This caching technique isn't thread-safe, but allowable
+        # in this case because a) the data is for our purposes
+        # immutable, so two writes of same key isn't a problem and
+        # b) we never remove entries.
```
exarkun: "never remove entries" strikes me as a memory usage problem waiting to happen.  These values probably do actually need to be invalidated, eventually, somehow.

Lacking any specific information about what the realistic usage patterns for this image data is, the simpler the solution the better here.  Something like an LRU is probably a good start (and there's no shortage of implementations of LRU caches for Python - eg http://docs.repoze.org/lru/ seems to be somewhat popular, and it has a decorator form that may even simplify the rest of this `_image_data` method).

exarkun: (A maximum size for the cache is something we want in order to limit memory usage of the process; that's another thing we should write a benchmark for but aren't able to do easily yet - so a unit test may also need to serve to demonstrate this is working as desired for the time being.)

exarkun:
Thanks.  Back to you.

dwgebler:
> This is the wrong requirements file, I think.

Oops, good catch! Absent-minded Monday morning (AMMM) syndrome.

exarkun:
> `image` is a unicode ID in the callable here though, as things stand atm.

Ah.  Okay.  Probably better to use the `Field.for_types` constructor, then, and just tell Eliot that the value is unicode rather than supplying `repr` as a serializer.

flocker/node/functional/test_docker.py
```diff
@@ -208,6 +211,46 @@ def started(_):
         d.addCallback(started)
         return d
 
+    @capture_logging(assertHasMessage, LOG_CACHED_IMAGE)
+    def test_list_image_data_cached(self, logger):
+        """
+        ``DockerClient.list`` will only an inspect an image ID once, caching
+        the resulting data.
+        """
+        image_name = u"openshift/busybox-http-app:latest"
```
exarkun: `ANY_IMAGE` would do here

flocker/node/_docker.py
exarkun: > b) we never remove entries.

`repoze.lru` will remove entries for us, now.  Fortunately, it is thread-safe.

exarkun:
Thanks!  Just a few more points, marked with inline comments.  Please address those and merge when buildbot is green.

PR #1990, opened 2015-09-25
[FLOC 3095] Add latest storage backends to config docs

davidoliverSP2:
Fixes 3095

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1990)


docs/config/configuring-nodes-storage.rst
```diff
@@ -49,5 +49,10 @@ The following pages describe how to configure the backends currently supported b
    openstack-configuration
    aws-configuration
    emc-configuration
+   vmware-configuration
```
adamtheturtle: Not a new issue, but please address at some point. "supported by Flocker" indicates (or, I infer) that we at ClusterHQ, makers of Flocker, support using these backends.

adamtheturtle: In particular as ZFS says "(EXPERIMENTAL)" that indicates we support the third party drivers more than ZFS. In reality we don't test against these third party drivers. cc @ferrantim .

adamtheturtle:
The issue description has paragraphs:

```
visit the following GitHub repository: https://github.com/XXX
```

but this change puts them in (weird to me) bulleted lists with one item. Please change if you think that change is an improvement, and merge when you are happy and the docs builder is green.

PR #1992, opened 2015-09-25
Runremotely floc 3124

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3124

Without this, some acceptance tests don't run with `pypy`, and we were passing unicode to a function which has a docstring explaining that it expects bytes.

See http://52.25.42.219/builders/flocker%2Facceptance%2Frackspace%2Fcentos-7%2Frackspace/builds/6 for a failure on master and http://52.25.42.219/builders/flocker%2Facceptance%2Frackspace%2Fcentos-7%2Frackspace/builds/10 for a success on this branch.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1992)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -946,7 +946,7 @@ def main(reactor, args, base_path, top_level):
                 parallel([
                     run_remotely(
                         username='root',
-                        address=node.address,
+                        address=node.address.encode('ascii'),
                         commands=task_pull_docker_images()
                     ) for node in cluster.agent_nodes
                 ]),
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -1242,7 +1242,7 @@ def configure_cluster(cluster, dataset_backend_configuration):
             sequence([
                 run_remotely(
                     username='root',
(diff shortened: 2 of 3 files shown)


admin/acceptance.py
```diff
@@ -946,7 +946,7 @@ def main(reactor, args, base_path, top_level):
                 parallel([
                     run_remotely(
                         username='root',
-                        address=node.address,
+                        address=node.address.encode('ascii'),
```
exarkun: `run_remotely` creates a `RunRemotely`.  `RunRemotely.address` is already constrained to be of type `bytes`.  So `node.address` must be of type `bytes` (on CPython) or this code wouldn't have worked without this change.

So it seems more likely that the problem is that `node.address` varies between `bytes` and `unicode` as a result of a bug somewhere else.

`bytes.encode('ascii')` is allowed by Python but it's a wrong thing to do and I think that's what this change is causing to happen on CPython.

dwgebler:
Closing, see #2040

PR #1993, opened 2015-09-25
[FLOC-3135] Handle exceptions from IDeployer.discover_state in the convergence loop

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3135

This ended up not being OpenStack specific.  As long as we actually retry the OpenStack operation, we'll eventually get results again.


exarkun:
Test changes and fix look good to me.  Will merge when buildbot is done.

exarkun:
Failures on CentOS 7 builder are weird.  I don't think they're related (they're from Docker, this code is pretty far from all of our Docker interaction).  So I'm going to suppose they're not introduced by this change and merge.

PR #1995, opened 2015-09-27
[FLOC-3137] Containers should always be restarted if they are stopped

exarkun:
Luke and Richard's design for https://clusterhq.atlassian.net/browse/FLOC-3137

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1995)


flocker/acceptance/reboot_httpserver.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    Respond to HTTP requests with bootup time of first time process was
+    run, as well as current run's bootup time.
+    """
+
+    def do_GET(s):
+        s.send_response(200)
+        s.send_header("Content-type", "text/plain")
+        s.end_headers()
+        s.wfile.write(file(FIRST_BOOT_PATH).read() + b"\n")
```
exarkun: Potentially leaked file here

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,109 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Tests for restarting and reboots and their interactions.
```
exarkun: Restarting and rebooting what things?

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,109 @@
...
+            # We now have the initial response of the server. This should
+            # be the initial boot time of the node, repeated twice.  Next,
+            # we restart the machine.
+            print "INITIAL RESPONSE", initial_response
+            initial_reboot_time = initial_response.splitlines()[0]
+            # XXX Checking exit code is problematic insofar as reboot
+            # kills the ssh process...
+            print "Rebooting!"
+            call([b"ssh", b"root@{}".format(node.public_address),
+                  b"shutdown", b"-r", b"now"])
```
exarkun: The (un)reliability of this approach to node control may turn this into a perpetually flaky test.  We should figure out something that's actually going to work reliably if we're going to have acceptance tests rebooting instances.

We should also consider whether an acceptance test is the best way to verify this functionality.

exarkun: A different strategy for rebooting the node could be to use an interface provided by the system providing the node.  We might make that accessible via an API like `cluster.nodes[0].reboot()` and have the node talk to libcloud or vagrant or whatever is appropriate.  The cloud APIs for rebooting an instance should be more reliable (or at least have more well-defined failure modes) than running shell commands via SSH.

admin/acceptance.py
```diff
@@ -122,6 +122,9 @@ def check_result(f):
         else:
             return f
 
+    print "To re-run tests against running cluster, run:"
+    print " ".join("%s=%s" % (k ,v) for (k, v) in get_trial_environment(cluster).iteritems()),
+    print " ".join(['trial'] + list(trial_args))
```
exarkun: Perhaps this is useful but it seems like part of an unrelated change.

flocker/acceptance/reboot_httpserver.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    :return: ``bytes`` with time this machine was booted.
+    """
+    return check_output([b"uptime", b"--since"])
+
+
+class RebootHandler(BaseHTTPServer.BaseHTTPRequestHandler):
+    """
+    Respond to HTTP requests with bootup time of first time process was
+    run, as well as current run's bootup time.
```
exarkun: The boot time doesn't seem relevant.  Any random piece of data would do, so long as it persists.  The data could even be supplied by the test instead of generated by the container - in which case perhaps there is a much simpler implementation of the container.

exarkun:
Thanks.

This goes somewhat beyond what a design should include.  Many of the implementation details will have to change if the design changes.  The design stage exists in large part to let you avoid having to write such details only to have to change them.  This PR has gone into the coding stage and could nearly be submitted for code review.  I've tried to skip commenting on such implementation details for now.

At the same time, this is missing some things a design should include - such as answers to the questions raised in the code near the XXX and TODO comments.

There are also documentation considerations.  This is a change in behavior.  It could be considered an incompatible change.  Since the previous behavior was really two behaviors, selected semi-randomly (based on boot up timings, I guess) and the new behavior is to deterministically select one of those behaviors, I wouldn't consider it incompatible in a bad way.  However, the documentation claimed the behavior was deterministic and corresponded to the behavior you've selected against here.  So there's at least an incompatibility in user expectations that needs to be communicated clearly.

Also, this fix is basically a hack.  It adds complexity to the system (tighter coupling between the two agents) with only a very expensive-to-run test.  We need a plan for how to fix this for real.  I wouldn't say we necessarily need it before merging this fix... But, apart from blocking this fix until we have such a plan, I don't know how to ensure we actually get one.

lukemarsden:
Thanks @exarkun.

The design is quite implementation-y on the testing side because I borrowed a test that @itamarst had already written.

The longer term plan for fixing this for real is switching the container API to use Flongle.

I'll include a documentation update in this branch and then try to work with someone on an implementation, if that's OK with you?

I had another idea for simulating the behaviour without a reboot - delete a container and unmount the filesystem at around the same time triggers the new exception in the code, we could probably use that to more reliably (?) and quickly test the code path.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,109 @@
...
+            ).addCallbacks(content)
+            return req
+
+        def created_dataset(dataset):
+            print "DATASET", dataset
+            http_server = {
+                u"name": random_name(self),
+                u"node_uuid": str(cluster.nodes[0].uuid),
+                u"image": u"python:2.7-slim",
+                u"ports": [{u"internal": 12345, u"external": 12345}],
```
wallrj: :question: Use `find_free_port` for external port.

flocker/acceptance/reboot_httpserver.py
lukemarsden: I think the boot time is relevant, because we specifically want to assert that the node rebooted by comparing different boot times before and after the reboot.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,109 @@
...
+                # We expect restarts to occur after reboot with this policy:
+                u'restart_policy': {u'name': u'always'},
```
wallrj: :question: The restart policy will be discarded by the container API, right? It confused me for a moment. Let's remove it.

wallrj:
Move the new test module  to one of the acceptance subdirectories....probably `endtoend` or alongside other container acceptance tests in the `obsolete` directory.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,111 @@
...
+        volume becomes available.
+        """
+        if not os.environ.get("RUN_REBOOT_TESTS"):
+            raise SkipTest(
+                    "Don't want to run this on buildbot, for now at least.")
+        node = cluster.nodes[0]
+        # Implicitly uses first node:
+        creating_dataset = create_dataset(self, cluster)
+
+        def query_server():
```
wallrj: We have a function for this already...`verify_socket`
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/testtools.py#L845

flocker/acceptance/test_restarts.py
wallrj: We have another acceptance test that connects to the node and restarts the docker daemon.
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/endtoend/test_dockerplugin.py#L53

Still flaky...but at least it gives a useful error if it fails.

wallrj: Although, my experience is that calling `shutdown -r now` will kill the remote SSH process and cause the local SSH process to exit with an error status....which will cause `run_ssh` to raise an exception which we'll need to handle.

wallrj:
We should probably write separate tests for:
 * Container restarting after reboot.
 * Container restarting after docker daemon restart.
 * Container not restarting until dataset is mounted. 
(And try and figure out how to verify that without relying on the race between container agent converging before the dataset agent)

lukemarsden:
A-ha! @wallrj and I realised that there's no stale state in the test because the first node also has the control service on it!

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+REBOOT_SERVER = FilePath(__file__).sibling(b"reboot_httpserver.py")
+
+def _service(address, name, action):
+    """
+    :param bytes address: The public IP of the node on which the service will
+        be changed.
+    :param bytes action: The action to perform on the service.
+    """
+    from twisted.internet import reactor
+    command = ["systemctl", action, name]
```
wallrj: Add an Ubuntu compatible version of this. 
Figure out how to disable / enable services on Ubuntu.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+            # be the initial boot time of the node, repeated twice.  Next,
+            # we restart the machine.
+            print "INITIAL RESPONSE", initial_response
+            initial_reboot_time = initial_response.splitlines()[0]
+            initial_container_id = initial_response.splitlines()[2].encode("ascii")
+            # XXX Checking exit code is problematic insofar as reboot
+
+            # kills the ssh process...
+            print "Rebooting!"
+            disabling = _service(node.public_address, b'flocker-dataset-agent', b'disable')
```
wallrj: `addCleanup` to re-enable `flocker-dataset-agent` after test completes.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+                      b"shutdown", b"-r", b"now"])
+            rebooting = disabling.addCallback(reboot)
+
+            # Now, keep trying to get a response until it's different than
+            # the one we originally got, thus indicating a reboot:
+            def query_until_different():
+                print "query!"
+                return query_server().addCallbacks(
+                    lambda response: response != initial_response,
+                    lambda _: False)
```
wallrj: :question: doesn't this prove that the server rebooted? i.e. No need to add an extra ensure reboot step.

wallrj: Oh wait, I'm confused, this doesn't get called yet. It might be clearer if we defined this function immediately before it gets assigned to the deferred.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+            # If the bug that this test is designed to catch occurs, then the container
+            # will have been stopped and a new one started before the dataset is in place.
+            # The dataset cannot be in place because the dataset agent isn't running.
+            # So by the time (just after) the container agent kills the first container
+            # ID, the bad thing will already have happened, and we can start the dataset
+            # agent in order to allow a correct implementation (where the bug is avoided)
+            # to eventually get the dataset in place and start the container correctly.
+            def old_container_gone(ignored):
+                if call([b"ssh", b"root@{}".format(node.public_address),
+                         b"true"]) == 0:
```
wallrj: :question: Is this necessary? If so, add an explanatory comment.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+                    if call([b"ssh", b"root@{}".format(node.public_address),
+                             b"docker", b"inspect", initial_container_id]) == 0:
```
wallrj: The docker HTTP API is directly accessible, so we can use docker-py and the Docker API here.

flocker/acceptance/test_restarts.py
```diff
@@ -0,0 +1,169 @@
...
+            # the one we originally got, thus indicating a reboot:
+            def query_until_different():
+                print "query!"
+                return query_server().addCallbacks(
+                    lambda response: response != initial_response,
+                    lambda _: False)
+
+            # Now, poll the host until the now-stopped Docker container that existed
+            # on the first run is destroyed. This is an external sign that the container
+            # agent tried to "restart" (by destroying and recreating) the container.
```
wallrj: I've got a bad feeling about this. It seems like we're relying on an implementation detail....but I need to re-read the code to be sure.
e.g. I could imagine a `flocker-container-agent` that cleaned up all its stopped containers when it first starts up.

flocker/acceptance/reboot_httpserver.py
tomprince: `/proc/sys/kernel/random/boot_id` is probably more explicit.

admin/acceptance.py
wallrj: Done.

flocker/node/_deploy.py
```diff
@@ -817,6 +830,32 @@ def discover_state(self, local_state):
...
+        # control service before the dataset agent and receives the (now stale)
+        # dataset state reported prior to the reboot.
+        # In that situation a stateful container may be started before the
+        # dataset agent has attached and mounted its dataset.
+        all_manifest = True
+        for manifestation in local_state.manifestations.values():
+            if not _is_mounted(
+                    # XXX This isn't right. Need a clean way to get the dataset
+                    # mountpoint from here. How?
+                    u'/flocker/{}'.format(manifestation.dataset.dataset_id)
```
lukemarsden: That's definitely wrong (e.g. ZFS backend won't work with this since it uses different format). We need to operate on the path which is definitely going to be passed to docker -- that was possible with the previous implementation approach...

tomprince:
I've left a bunch of comments at various levels of detail. The one major comment I have is that I'm unsure of your method of detecting that a reboot is completed.

The container agent will report ignorance about everything until it gets state from the data-set agent. Since the control service ignores that ignorance, the control service will continue to report stale container state until the timeout (2m). Perhaps we should be differentiating authoritatively ignorant (like the container agent about containers) vs. true ignorance (like the container agent about datasets). And then have the control service actually update its state in the face of authoritative ignorance.

I'm not sure how that plays out against the desire to avoid having flapping state, though.

---

I wonder if a better solution to this would be just to aggressively expire state at least when agents disconnect. If we want to avoid flapping on user-visible state, we could preserve just the data that is sent to the user, but report correct ignorance to the agents. It seems like the solution in here is basically working around the fact that we don't expire state aggressively in the control service.

Granted, there would still be a (much smaller) window where the race could occur, if the agents shutdown uncleanly. That suggests to me that there should be a way of forcing state expiry on a reboot. Perhaps a service that pushes authoritative ignorance that starts before either agent. Or perhaps the agents should report `/proc/sys/kernel/random/boot_id` and if it changes, expire all information about the node.

---

I'm not entirely sure what of the above is within the scope of this issue. But resolving that definitely is required to approve the design.


---

Reviewed 2 of 7 files at r9, 1 of 1 files at r10.
Review status: 3 of 5 files reviewed at latest revision, 25 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 26 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-M7VzlrsBZDBIUL46N)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L26)):</sup>
This definitely needs to support Ubuntu (as mentioned in a now hidden comment). We can probably use `['service', name, action]` which (I think) works on both CentOS and Ubuntu.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 47 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MRJTIC87ZQwqkH97K)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L47)):</sup>
:question: s/on reboots/after a reboot/


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 70 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-M8EFiUxWEZ5NRmTmm)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L70)):</sup>
This comment shouldn't live in the final version (or at least it should be fixed to be accurate once it has been merged.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 85 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MBhw-rQNqevxpd5ur)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L85)):</sup>
I think it it would be better to use some deterministic data here, rather than random times. i think using `datahttp.py` and writing some data to it would work.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 95 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MRdgHNqSwWuPZaM6z)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L95)):</sup>
This won't work with the current state of the code. The container agent will report ignorance, since the correct datasets aren't mounted, so the control service will continue to report stale container state until it expires. 


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 101 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MCxu5Bc3cCXEMFZsy)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L101)):</sup>
Why do we care about either of these things? The user visible behavior we care about is
- the container is started
- the container has mounted the correct dataset.
Any other details are implementation details.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 106 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MLf0GC9AOPxz0R8by)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L106)):</sup>
This should probably be a helper function.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 107 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-M9eI0f1bNuCESCFqZ)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L107)):</sup>
No `print`s. Instead use some eliot logging.


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 123 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-JnPhKBMJzAOU2auo2-r2-55)** ([raw file](https://github.com/clusterhq/flocker/blob/25422e2e0be65f68ca069e6cc4a570d012624f69/flocker/acceptance/endtoend/test_restarts.py#L123)):</sup>
The current`find_free_port` is non-nonsensical in acceptance tests, since it finds a port on client machine, not the machine the container is going to run on. 


---

<sup>**[flocker/acceptance/endtoend/test_restarts.py, line 163 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-McCNAYmiBqTrUC8Jt)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/endtoend/test_restarts.py#L163)):</sup>
As mentioned elsewhere, we'll never get empty application state, only stale state and then (after 2m) ignorance.


---

<sup>**[flocker/acceptance/testtools.py, line 817 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MHmd5qTFeI3xdSoid)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/acceptance/testtools.py#L817)):</sup>
This is now out of date. Also, document the new parameter.


---

<sup>**[flocker/node/_deploy.py, line 614 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MXhXRkwtmKFpG7xRC)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/node/_deploy.py#L614)):</sup>
Do ZFS volumes show up here?


---

<sup>**[flocker/node/_deploy.py, line 834 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-M_r7rPMcLtm82JD4-)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/node/_deploy.py#L834)):</sup>
There should be a big warning that the following logic is a big hack. (Also it should be associated with the following, not the preceding code)


---

<sup>**[flocker/node/_deploy.py, line 855 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-Mcca1K5XBVez3qtza)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/node/_deploy.py#L855)):</sup>
This logic needs some tests.


---

<sup>**[flocker/node/_deploy.py, line 950 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MfvQs5w3x_v-o71wi)** ([raw file](https://github.com/clusterhq/flocker/blob/5722dac72cab5972ee728d8094c90bde8a221c27/flocker/node/_deploy.py#L950)):</sup>
This probably needs some test changes.


---

<sup>**[flocker/node/_deploy.py, line 983 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MdpumuTLAthqWaNMK)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/node/_deploy.py#L983)):</sup>
It is good that there is a link to the relevant ticket, but I think it would be clearer for later readers if that was a link at the end `(see FLOC-XXXX)` and the rest of this comment was worded to describe the desired behavior on its own merits.

Also, having a comment here describing code that doesn't exist and doesn't obviously have a reason for existing seems a bit weird. (Since the presence of original comment suggested the code existing was exceptional, not having it seems unexceptional and not deserving of a comment).

----

I'm torn here. I guess somebody needs to look at this bit of code with fresh eyes, and see if the comment makes sense in/is useful in context, for somebody who didn't know there used to be code here. If not, the comment should go.


---

<sup>**[flocker/node/_deploy.py, line 1004 \[r10\]](https://reviewable.io:443/reviews/clusterhq/flocker/1995#-K-MdGjwwI-ASKUjNZtW)** ([raw file](https://github.com/clusterhq/flocker/blob/65c1bede1bac6e5867790181425e4a8d3d871cac/flocker/node/_deploy.py#L1004)):</sup>
This branch is (in addition to dealing with stuff about state and restarts, apparently changing the default (only) restart policy. Except that we are always going to pass `never` to docker still? In any case, there is work to be done to clear up how restart policy in flocker and docker interact.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1995)

lukemarsden:
I had a worry about the design which I think I convinced myself isn't really a problem, but for completeness I wanted to record it here. If *any* dataset is missing from the host, *no* containers will be started. This means that a single missing dataset will stop the host converging on any configuration, even for containers with healthy datasets which could be happily be started.

The reason I think this isn't a problem is that the error checking here is to protect about *incorrect (stale) knowledge*. Once the dataset agent starts up on a host again, it should correctly determine the absence of any missing/unmountable (for whatever reason, e.g. filesystem is corrupt, I/O error on the block device) datasets, and correctly report that to the control service which reports it to the container agent. At that point, the container agent will correctly start up whichever containers it can start, maintaining the desirable "start what you can" semantics.

WDYT?

PR #1996, opened 2015-09-28
[FLOC-2894] Specify branching policy for release vs. patch vs. dev releases.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-2894

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1996)

PR #1998, opened 2015-09-28
[FLOC-3138] Release Flocker 1.4.1

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3138

I used the branch naming from https://github.com/ClusterHQ/flocker/pull/1996 for this.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1998)


tomprince:
Note that this is based of master, since the relevant fix wasn't based of the last release. However, the only non-test change ([diff](https://github.com/ClusterHQ/flocker/compare/1.4.0...release/flocker-1.4)) is [FLOC-1292](https://clusterhq.atlassian.net/browse/FLOC-1292). That also seems reasonable to include in a release.

adamtheturtle:
Thanks @tomprince . The changes here look good.

I will have a look through the release publishing code to see if there is any problem I can spot with the new branching. I think that in these cases where the old system is known to work then it is probably best to stick with them until the alternative branching system is accepted, but for now this can be continued.

NEWS
```diff
@@ -1,3 +1,11 @@
+Flocker 1.4.1 (2015-09-27)
+==========================
+
+- Unexpected errors in agent state discovery no longer break the agent convergence loop. (FLOC-3135)
+- Fixed formating of links in admonitions. (FLOC-1292)
```
exarkun: `formatting`

tomprince:
> I will have a look through the release publishing code to see if there is any problem I can spot with the new branching.

The only thing that cares about branching is the broken script to create branches (since it demands pre-releases which we aren't doing.

adamtheturtle:
@tomprince sorry I stopped - as per the JIRA issue we aren't doing a release today.

PR #1999, opened 2015-09-29
[FLOC-3137] Always restart containers

lukemarsden:
Resolves: https://clusterhq.atlassian.net/browse/FLOC-3137

Flocker is currently documented as implementing the equivalent of the `restart=never` Docker policy. However, this is not the current behaviour, nor is it desirable behaviour. Currently, Flocker *does* restart containers when they are stopped. This PR removes the line of code which attempted -- but failed -- to stop these restarts previously, and updates the documentation to accurately reflect the actual behaviour. It also updates the unit test which tested the previous code in what must have been an unrealistic context (because it never worked in practice) and adds an acceptance test which verifies that the desired behaviour is actually correct.

Behaviour with respect to node reboots is out of scope for this PR.

@wallrj and I paired on this branch.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/1999)


tomprince:
This looks good. Please merge once buildbot is green.

I think we should, if possible, address the UI issue of reporting an unused restart policy before releasing this. Please file an issue to track that.


---

Review status: 0 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[docs/reference/api_examples.yml, line 383 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/1999#-K-Pj5lD460Gq9AAkMmY)** ([raw file](https://github.com/clusterhq/flocker/blob/3bfab16986252970237735be8d1a47ef77f818cf/docs/reference/api_examples.yml#L383)):</sup>
It seems like we should have containers always report that they have a restart policy of always then?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1999)

tomprince:
As an aside, I think i would have been inclined to start a new branch for this change, rather than reusing the branch that was done for restarts, since they share almost no code. (Not worth worrying about at this point, though.


---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1999)

tomprince:
Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/1999)

lukemarsden:
Thanks for the review, @tomprince!

I've filed the follow-up issue you asked for here: https://clusterhq.atlassian.net/browse/FLOC-3153

PR #2002, opened 2015-09-30
[FLOC-3140] Coalesce redundant cluster parameter updates to agents

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3140

The general idea here is that because many different things trigger a full update of cluster parameters (state and configuration) to agents, it is often the case that several full updates will be sent to an agent before it has even had a chance to process the first update.  In these cases, the first update is unavoidable and the last update is useful but the intermediate updates are useless and we can avoid sending them.  So, do.

Since we only have manual, ad hoc benchmarking in place at the moment, here are some results from my manual, ad hoc benchmarks:

Without this change, it takes longer than 10 minutes (I didn't wait to see much much longer) for flocker-control to finish processing all of the events generated on the startup of a 40 node cluster.

With this change, it takes about 3 minutes.

With this change, on an empty 40 node cluster it takes slightly less than 6 minutes to configure 593 containers (out of an attempted 800, the difference is accounted for by errors from the API server).  Measuring from when the first container is configured, it takes a little less than 9 minutes to converge on this configuration.  While convergence is underway, it takes a little under 2 minutes for the control service to respond to a read-only API request.  Once the cluster has converged, it takes 14 seconds to add one more container to the configuration and an additional 16 seconds to converge on that configuration.

As a follow-up to this, I've made an additional change which cuts down the cost of some Eliot logging near the code changed by this PR.  That involved some refactoring which somewhat improves the readability of this code.  I didn't have a chance this evening to push that refactoring into this PR so what's here is a bit of a mess.  Please have a look at https://github.com/ClusterHQ/flocker/compare/coalesce-calls-probably-even-better-FLOC-3140...send_cluster_state_logging-FLOC-3151?expand=1 to see the improved factoring and don't count the mess in this PR as a reason not to accept this change (since it is required before send_cluster_state_logging-FLOC-3151 can be merged).

As far as I know, send_cluster_state_logging-FLOC-3151 is also ready for a PR but since it's stacked on this branch github will make a mess of the diff - so I'm aiming for this to be merged before creating that PR.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2002)

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -460,23 +460,25 @@ class ControlAMPService(Service):
                         # shouldn't be allowed to).  No test coverage for
                         # either of these cases.
                         try:
-                            current_command = self._current_command_for_connection[
+                            (current_command, already_scheduled) = self._current_command_for_connection[
                                 connection
                             ]
                         except KeyError:
                             current_command = self._update_connection(
                                 connection, configuration, state, action
                             )
-                            self._current_command_for_connection[connection] = current_command
+                            self._current_command_for_connection[connection] = (current_command, False)
 
                             def finished_update(ignored, connection):
                                 del self._current_command_for_connection[connection]
                             current_command.addCallback(finished_update, connection)
                         else:
-                            current_command.addCallback(
-                                lambda ignored, connection: self._send_state_to_connections([connection]),
-                                connection,
-                            )
+                            if not already_scheduled:
+                                current_command.addCallback(
+                                    lambda ignored, connection: self._send_state_to_connections([connection]),
+                                    connection,
+                                )
+                                self._current_command_for_connection[connection] = (current_command, True)
 
 
     def _update_connection(self, connection, configuration, state, action):
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -36,7 +36,7 @@ from io import BytesIO
 from itertools import count
 from contextlib import contextmanager
 
-from eliot import Logger, ActionType, Action, Field
+from eliot import Logger, ActionType, Action, Field, MessageType
 from eliot.twisted import DeferredContext
 
 from characteristic import with_cmp
@@ -395,6 +395,17 @@ AGENT_CONNECTED = ActionType(
     "An agent connected to the control service."
 )
 
+AGENT_UPDATE_ELIDED = MessageType(
+    "flocker:controlservice:agent_update_elided",
+    [AGENT],
+    u"An update to an agent was elided because a subsequent update supercedes it.",
+)
+
+AGENT_UPDATE_DELAYED = MessageType(
+    "flocker:controlservice:agent_update_delayed",
+    [AGENT],
+    u"An update to an agent was delayed because an earlier update is still in progress.",
+)
 
 class ControlAMPService(Service):
     """
@@ -471,7 +482,10 @@ class ControlAMPService(Service):
                             del self._current_command_for_connection[connection]
                         current_command.addCallback(finished_update, connection)
                     else:
-                        if not already_scheduled:
+                        if already_scheduled:
+                            AGENT_UPDATE_ELIDED(agent=connection).write()
+                        else:
+                            AGENT_UPDATE_DELAYED(agent=connection).write()
                             current_command.addCallback(
                                 lambda ignored, connection: self._send_state_to_connections([connection]),
                                 connection,
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -370,20 +370,11 @@ def _serialize_agent(controlamp):
     """
     Serialize a connected ``ControlAMP`` to the address of its peer.
 
-    :raise TypeError: If the given protocol is not an instance of
-        ``ControlAMP``.
-
     :return: A string representation of the Twisted address object describing
         the remote address of the connection of the given protocol.
 
     :rtype str:
     """
-    if not isinstance(controlamp, ControlAMP):
-        raise TypeError(
-            "Logged agent field must be ControlAMP, not {}".format(
-                fullyQualifiedName(controlamp.__class__),
-            )
-        )
     return str(controlamp.transport.getPeer())
 
 
@@ -455,12 +446,11 @@ class ControlAMPService(Service):
         """
         configuration = self.configuration_service.get()
         state = self.cluster_state.as_deployment()
-        with LOG_SEND_CLUSTER_STATE(self.logger,
-                                    configuration=configuration,
+        with LOG_SEND_CLUSTER_STATE(configuration=configuration,
                                     state=state):
             with _caching_encoder.cache():
                 for connection in connections:
-                    action = LOG_SEND_TO_AGENT(self.logger, agent=connection)
+                    action = LOG_SEND_TO_AGENT(agent=connection)
                     with action.context():
                         # XXX If callRemote raises an exception, the loop won't
                         # finish and the rest of the connections won't receive
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -451,46 +451,45 @@ class ControlAMPService(Service):
                                     state=state):
             with _caching_encoder.cache():
                 for connection in connections:
-                    action = LOG_SEND_TO_AGENT(agent=connection)
-                    with action.context():
-                        # XXX If callRemote raises an exception, the loop won't
-                        # finish and the rest of the connections won't receive
-                        # the updated state.  Asynchronous exceptions aren't a
-                        # problem since they won't interrupt the loop (and they
-                        # shouldn't be allowed to).  No test coverage for
-                        # either of these cases.
-                        try:
-                            (current_command, already_scheduled) = self._current_command_for_connection[
-                                connection
-                            ]
-                        except KeyError:
-                            current_command = self._update_connection(
-                                connection, configuration, state, action
+                    # XXX If callRemote raises an exception, the loop won't
+                    # finish and the rest of the connections won't receive
+                    # the updated state.  Asynchronous exceptions aren't a
+                    # problem since they won't interrupt the loop (and they
+                    # shouldn't be allowed to).  No test coverage for
+                    # either of these cases.
+                    try:
+                        (current_command, already_scheduled) = self._current_command_for_connection[
+                            connection
+                        ]
+                    except KeyError:
+                        current_command = self._update_connection(
+                            connection, configuration, state, action
+                        )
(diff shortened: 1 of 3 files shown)


flocker/control/test/test_protocol.py
```diff
@@ -974,20 +1078,16 @@ def test_logging(self, logger):
...
-            "callRemote",
-            lambda *args, **kwargs: succeed({})
-        )
+        agent = FakeAgent()
+        client = AgentAMP(Clock(), agent)
+        server = LoopbackAMPClient(client.locator)
+
+        control_amp_service.connected(server)
+
         self.patch(control_amp_service, 'logger', logger)
```
dwgebler: I think this line is not necessary anymore.

flocker/control/test/test_protocol.py
```diff
@@ -974,20 +1078,16 @@ def test_logging(self, logger):
...
 
-        control_amp_service._send_state_to_connections(
-            connections=[connection_protocol])
+        control_amp_service._send_state_to_connections(connections=[server])
 
         assertHasAction(
```
dwgebler: This test has failed on a couple of builds on this assertion with no action logged (e.g. http://build.clusterhq.com/builders/flocker%2Funit-test%2Fcentos-7%2Fzfs-head/builds/1467/steps/trial/logs/stdio) but I can't reproduce locally or see any immediately obvious cause.

flocker/control/test/test_protocol.py
```diff
@@ -50,6 +54,26 @@
...
+def random_ip():
+    return inet_ntop(AF_INET6, urandom(16))
+
+
+def arbitrary_transformation(deployment):
+    """
+    Make some change to a deployment configuration.  Any change.
+
+    The exact change made is unspecified but the resulting ``Deployment`` will
+    be different from the given ``Deployment``.
```
dwgebler: The change is specified though, a new empty node is added to the deployment, so the function name and docstring should be adjusted. Also I don't think you need `random_ip` because you don't need to specify a hostname when you specify a UUID.

dwgebler:
There are lint errors in the branch coming from `test_protocol.py`

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
-                        # either of these cases.
-                        d = DeferredContext(connection.callRemote(
-                            ClusterStatusCommand,
-                            configuration=configuration,
-                            state=state,
-                            eliot_context=action
-                        ))
-                        d.addActionFinish()
-                        d.result.addErrback(lambda _: None)
+                    # XXX If callRemote raises an exception, the loop won't
```
dwgebler: This comment looks like it belongs in `_update_connection` now.

flocker/control/test/test_protocol.py
```diff
@@ -50,6 +54,26 @@
 from .clusterstatetools import advance_some, advance_rest
 
 
+def random_ip():
+    return inet_ntop(AF_INET6, urandom(16))
```
dwgebler: See other comment about specifying UUID on a `Node`; I don't think you need this function at all.

flocker/control/test/test_protocol.py
```diff
@@ -586,25 +611,30 @@ def test_nodestate_notifies_all_connected(self):
...
-        self.protocol.makeConnection(StringTransport())
-        another_protocol = ControlAMP(self.reactor, self.control_amp_service)
-        another_protocol.makeConnection(StringTransport())
-        sent1 = []
-        sent2 = []
 
-        self.patch_call_remote(sent1, self.protocol)
-        self.patch_call_remote(sent2, protocol=another_protocol)
+        agents = [FakeAgent(), FakeAgent()]
+        clients = list(AgentAMP(Clock(), agent) for agent in agents)
```
dwgebler: :question: shouldn't these all be the same `Clock` instance? I guess it doesn't really matter here...

flocker/control/test/test_protocol.py
exarkun: It's not specified by the interface.  The interface only specifies that something will change, not what.  Code shouldn't depend on the exact change, only that `deployment != arbitrary_transformation(deployment)`.

flocker/control/test/test_protocol.py
jml: It doesn't really matter either way, but I think since each of these represents a notionally different computer it makes sense for them to have different clocks.

exarkun: They could be.  Or, since each agent actually runs on its own computer, it might makes sense for them to all be different.  But it's correct that it doesn't matter since this test doesn't exercise any code that uses the clock.

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
+                            AGENT_UPDATE_ELIDED(agent=connection).write()
+                        else:
+                            AGENT_UPDATE_DELAYED(agent=connection).write()
+                            current_command.addCallback(
+                                lambda ignored, connection: self._send_state_to_connections([connection]),
+                                connection,
+                            )
+                            self._current_command_for_connection[connection] = (current_command, True)
+
+    def _update_connection(self, connection, configuration, state):
```
wallrj: Docstring needed.

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
+                    # finish and the rest of the connections won't receive
+                    # the updated state.  Asynchronous exceptions aren't a
+                    # problem since they won't interrupt the loop (and they
+                    # shouldn't be allowed to).  No test coverage for
+                    # either of these cases.
+                    try:
+                        (current_command, already_scheduled) = self._current_command_for_connection[
+                            connection
+                        ]
+                    except KeyError:
```
wallrj:
```
# There are no un-acknowledged state change messages for this connection.
# Start a new message immediately.
# Save the un-acknowledged message (Deferred) for this connection.
```

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
+                        current_command = self._update_connection(
+                            connection, configuration, state,
+                        )
+                        self._current_command_for_connection[connection] = (current_command, False)
+
+                        def finished_update(ignored, connection):
+                            del self._current_command_for_connection[connection]
+                        current_command.addCallback(finished_update, connection)
+                    else:
```
wallrj:
```
# A state change message has been sent to this agent but agent that has not been acknowleged.
# Don't send another message.
```

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
+                        if already_scheduled:
+                            AGENT_UPDATE_ELIDED(agent=connection).write()
+                        else:
```
wallrj: # Schedule another state change message.
# When the outstanding message is eventually acknowledged, the latest state and configuration will be immediately sent to this agent.

flocker/control/_protocol.py
```diff
@@ -455,27 +458,52 @@ def _send_state_to_connections(self, connections):
...
+                        current_command = self._update_connection(
+                            connection, configuration, state,
+                        )
+                        self._current_command_for_connection[connection] = (current_command, False)
+
+                        def finished_update(ignored, connection):
+                            del self._current_command_for_connection[connection]
+                        current_command.addCallback(finished_update, connection)
+                    else:
+                        if already_scheduled:
```
wallrj:
```
# Yet another state change comes in, but a state update message has already been scheduled for this agent. 
# No point doing any thing.
```

flocker/control/test/test_protocol.py
```diff
@@ -693,25 +717,99 @@ def test_configuration_change(self):
...
+        service = build_control_amp_service(self)
+        service.startService()
+
+        configuration = service.configuration_service.get()
+        modified_configuration = arbitrary_transformation(configuration)
+
+        server = LoopbackAMPClient(client.locator)
+        delayed_server = DelayedAMPClient(server)
+        # Send first update
+        service.connected(delayed_server)
```
wallrj: :question: This and the following test simulate the sending of state change when agent first connects to control service...which is special in that the first state is only sent to that agent, not all agents. Does that matter? I guess not...

flocker/control/test/test_protocol.py
```diff
@@ -693,25 +717,99 @@ def test_configuration_change(self):
...
+        client = AgentAMP(Clock(), agent)
+        service = build_control_amp_service(self)
+        service.startService()
+
+        configuration = service.configuration_service.get()
+        modified_configuration = arbitrary_transformation(configuration)
+        more_modified_configuration = arbitrary_transformation(modified_configuration)
+
+        server = LoopbackAMPClient(client.locator)
+        delayed_server = DelayedAMPClient(server)
```
wallrj: :question: It could just be me, but the variable names in these tests confused me for while....here are some of my difficulties:
 * `DelayedAMPClient` wraps `LoopbackAMPClient` but assigning to `server` makes it look like server is significant. Perhaps `DelayedAMPClient(wraps=LoopbackAMPClient(client.locator)` would be clearer.
 * `first_agent_desired`, `second_agent_desired` ... got me thinking that we were simulating multiple connected agents....instead of the state of a single agent at three points in time.....perhaps `agent_desired_1`, `agent_desired_2` etc might be clearer.
 * `service` could be `control_service` for the avoidance of doubt.
 * `delayed_server` could be `delayed_agent_endpoint` (or something)

wallrj:
Thanks @exarkun and @jml 

This looks great.

 * It took me a while to work out the situations that lead to each of `send now`, `schedule send`, `already scheduled` cases. I thought they could do with some extra comments....I suggested some, but feel free to discard those.

 * You should probably explain the new behaviour of `_send_cluster_state` in its docstring.

 * I added a few other comments which you can address if you agree.

Please merge when you've fixed the lint errors and got a green build.

wallrj:
Ah, I see that you've cleaned up / refactored the code in the follow up branch  (send_cluster_state_logging-FLOC-3151) and  added the explanatory comments there.

exarkun:
> Ah, I see that you've cleaned up / refactored the code in the follow up branch (send_cluster_state_logging-FLOC-3151) and added the explanatory comments there.

Indeed.  Sorry about that.  I realize it's not ideal to have the cleaner factoring / improved documentation in the follow-up branch instead of here.  If I hadn't been rushing to deliver something for today then I definitely wouldn't have done this.

PR #2003, opened 2015-09-30
[FLOC-3155] Add 'available' as a possible transient state when attaching Cinder volumes

jongiddy:
This PR adds 'available' as an expected state when attaching volumes using the Cinder client. Although the volume goes through the sequence `('available', 'attaching', 'in-use')`, the state 'available' was missing as an expected state because we have never seen 'available' occur in the attachment sequence. 

For detaching, we had seen the state 'in-use' occur at the start of the sequence `('in-use', 'detaching', 'available')`.  Hence, the equivalent detach code includes the initial state. 

Perhaps due to speedups introduced through recent patches, we have now seen the unexpected 'available' state occur in attachment.  Hence, we add it to the code as an expected state.

This has the nice effect of making the `attach_volume` and `detach_volume` functions more symmetric.

Note, there is another sequence that can occur `('available', 'attaching', (some inability of Cinder to attach), 'available')`.  If this occurs, one of two things will happen:
1. If the 'attaching' state is seen, then 'available' is no longer a valid state, and the unexpected state exception will be raised, correctly this time.
2. If the 'attaching' state is not seen, it is ambiguous whether we are seeing the first or second 'available' state. In this case, the client will hang until a timeout exception is raised. Since we regard failure to attach within the timeout as an error, it is appropriate to raise an exception, and try again on the next convergence loop.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2003)

### DIFF
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -527,7 +527,7 @@ class CinderBlockDeviceAPI(object):
             volume_manager=self.cinder_volume_manager,
             expected_volume=nova_volume,
             desired_state=u'in-use',
-            transient_states=(u'attaching',),
+            transient_states=(u'available', u'attaching',),
         )
 
         attached_volume = unattached_volume.set('attached_to', attach_to)
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -331,7 +331,7 @@ class CinderAttachmentTests(SynchronousTestCase):
             volume_manager=self.cinder.volumes,
             expected_volume=attached_volume,
             desired_state=u'in-use',
-            transient_states=(u'attaching',),
+            transient_states=(u'available', u'attaching',),
         )
 
         devices_after = set(FilePath('/dev').children())
@@ -376,7 +376,7 @@ class CinderAttachmentTests(SynchronousTestCase):
             volume_manager=self.cinder.volumes,
             expected_volume=attached_volume,
             desired_state=u'in-use',
-            transient_states=(u'attaching',),
+            transient_states=(u'available', u'attaching',),
         )
 
         devices_after = set(FilePath('/dev').children())
@@ -421,7 +421,7 @@ class CinderAttachmentTests(SynchronousTestCase):
                 volume_manager=self.cinder.volumes,
                 expected_volume=attached_volume,
                 desired_state=u'in-use',
-                transient_states=(u'attaching',),
+                transient_states=(u'available', u'attaching',),
             )
         self.assertEqual(e.exception.unexpected_state, u'available')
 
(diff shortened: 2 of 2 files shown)


flocker/node/agents/cinder.py
```diff
@@ -527,7 +527,7 @@ def attach_volume(self, blockdevice_id, attach_to):
             volume_manager=self.cinder_volume_manager,
             expected_volume=nova_volume,
             desired_state=u'in-use',
-            transient_states=(u'attaching',),
+            transient_states=(u'available', u'attaching',),
```
zubron: We should update the `CinderAttachmentTests` to include this state. Currently they only include `attaching` as a possible transient state.

zubron:
This change makes sense to me, although I would prefer if we were able to test it in some way. I'm not sure how easy this would be given that there's no guarantee that the volume will be in the `available` state for long enough. I've added a suggestion to update the `CinderAttachmentTests` to include this state.

zubron:
Following discussions, I'm happy for this to be merged when the tests pass.

wallrj:
Yep, I agree. This change looks good. 

If we have time, we might be able to test this on devstack by running the functional tests on devstack with the cinder scheduler process paused (or something like that).

On master that might cause wait_for_volume_state to see the unexpected `available` transient state and blow up while with this branch it'd just keep waiting until we then unpause the scheduler.

In https://github.com/ClusterHQ/flocker/pull/1912 (where the transient states were introduced) and in previous such PRs we've taken this approach to testing, in the absence  of automated devstack test environment.

exarkun:
Please record any testing that has been done to demonstrate this change fixes the problem it's intended to fix.  Please include sufficient details for that testing to be reproduced in the future.  Please also file an issue for adding automated test coverage for this case.

jongiddy:
The testing that has been done at this stage is that the 'available' state was seen to occur on a large (25 nodes and ~200 containers) customer cluster during attachment.  It was not clear whether this was the initial 'available' state being seen before the attachment transition is started, or a final 'available' state due to an error occurring after the Cinder operation has returned (Cinder operations are asynchronous to the calls that start them).

However, if an unexpected state occurs during attachment, the creation and mounting of the volume is interrupted, and does not get repeated cleanly, because `CreateBlockDeviceDataset` contains multiple changes in one step. [FLOC-1771, FLOC-2396].

Until these issues are resolved, it is necessary to avoid having `CreateBlockDeviceDataset` fail. Allowing state 'available' and increasing the timeout for Cinder to wait for states from 60 seconds to 300 seconds, was seen to work on the customer cluster.

We are currently investigating whether the valid transient states can occur in any order, or will always occur in a specific order, and also whether we can sensibly refine the value for the timeout between 60 and 300 s.

wallrj:
@zubron and I have been trying to recreate the problem by running `test_attach_unattached_volume` with a pause between creating and attaching.

After creating we stop the cinder scheduling service.

Then continue the test in the hope that the API will continue to report state `available`.

We're still working on this and may have broken our devstack environment.

Ok to merge as-is without demoing the fix on devstack?

wallrj:
Followup issue for testing this change is https://clusterhq.atlassian.net/browse/FLOC-3161

I'll merge this now.

PR #2004, opened 2015-09-30
[FLOC-3151] Reduce the amount of logging done about the "send cluster state" operation

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3151

Against 1.4.0+260.g8cfd264, 40 nodes, starting from zero containers and datasets configured or running:

  * 5m14s to configure 621 containers (out of attempted 800; difference comes from API errors)
  * 35s to respond to a read-only API request while convergence is in progress
  * 6m7s to converge on 606 (?) containers running
  * 15s to configure 1 additional container from that state
  * 18s to converge on the new configuration

Note this is stacked on https://github.com/ClusterHQ/flocker/pull/2002.  The diff to review here is https://github.com/ClusterHQ/flocker/compare/coalesce-calls-probably-even-better-FLOC-3140...send_cluster_state_logging-FLOC-3151

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2004)


exarkun:
On investigation, I see that Docker on one of the cluster has gotten itself wedged and so that node is incapable of starting containers.  This entirely explains the failure to reach 621 running containers (iow, not a bug in these changes).

flocker/control/_protocol.py
```diff
@@ -354,15 +355,19 @@ def connectionLost(self, reason):
...
 
-DEPLOYMENT_CONFIG = Field(u"configuration", repr,
+# These two logging fields use wire_encode as the serializer so that they can
+# share the encoding cache with the network code related to this logging.  This
+# reduces the overhead of logging these (potentially quite large) data
+# structures.
+DEPLOYMENT_CONFIG = Field(u"configuration", wire_encode,
                           u"The cluster configuration")
-CLUSTER_STATE = Field(u"state", repr,
+CLUSTER_STATE = Field(u"state", wire_encode,
```
dwgebler: :question: Am I misunderstanding how this works - shouldn't these be using `_caching_encoder.encode` as the serializer?

dwgebler:
Thanks @exarkun this mostly looks fine, other than my question - without an answer to that, I can see the encoder caching in the serializable config and state sent to the agent, but I'm not understanding at the moment how this reduces the logging's serialization work. Please answer / address that and then merge when there's a passing build.

flocker/control/_protocol.py
exarkun: I think you're right.  Using `wire_encode` directly here is bypassing the caching I was trying to bring in.

exarkun:
Just a follow-up with results from the code that was actually merged...

Against 1.4.0+266.g9ba8930, 39 nodes (one died of unrelated causes since my last round of testing), starting from zero containers and datasets configured or running:

    5m2s to configure 610 containers (out of attempted 800; difference comes from API errors)
    41s to respond to a read-only API request while convergence is in progress
    5m54s to converge on 548 containers running
    15s to configure 1 additional container from that state
    18s to converge on the new configuration

Some other measurements I wish I'd been taking previously:

   5s to respond to a read-only API request when quiescent
   1m30s for flocker-control to quiesce after being restarted

PR #2005, opened 2015-09-30
[FLOC-3158] Re-raise `RequestLimitExceeded` instead of `CloudKeyNotFound`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3158

My hypothesis is that https://clusterhq.atlassian.net/browse/FLOC-3152 is caused by `RequestLimitExceeded`. By propagating the error, we can see if that is in fact the case.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2005)


itamarst:
Gotta love libcloud error handling. Merge when ... one of the acceptance builds passes demonstrating it's not completely broken?

PR #2006, opened 2015-09-30
[FLOC-3159] Bump timeout to 600.

lukemarsden:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3159

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2006)

### DIFF
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -49,6 +49,8 @@ CLUSTER_ID_LABEL = u'flocker-cluster-id'
 # a volume.
 DATASET_ID_LABEL = u'flocker-dataset-id'
 
+# The longest time we're willing to wait for a Cinder API call to complete.
+CINDER_TIMEOUT = 600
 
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks @lukemarsden 

@zubron and I have:

* Fixed a lint error.
* checked for other users of `f.n.a.cinder.wait_for_volume_state`; they are create, attach, detach_volume and a number of uses in the cinder specific functional tests. @zubron notes that tests may take longer to fail, but not a show stopper. If we ever write tests to reproduce timeouts (perhaps on devstack) we might need to increase the test timeout too, to match.
* The original 60s was an arbitrary value which isn't explicitly tested and isn't part of the IBlockDeviceAPI. Perhaps it should be? But not now.
* @zubron found that there are EBS timeout tests which use what looks like a fake EBS volume API. 
* We should probably create a followup issue to unit test the `VolumeStateMonitor` and `wait_for_volume`.

We'll merge this when we get green builds.

PR #2000, opened 2015-09-29
[FLOC-3129] "Using" section for FPD docs

robhaswell:
https://clusterhq.atlassian.net/browse/FLOC-3129

This PR adds instructions on using the Flocker plugin for Docker, and duplicates the known limitations into this section by way of an include.

One current problem is that the remote API is undocumented at Docker. This will be fixed soon, and is stable. It's debatable whether we should even include this documentation. My position is that it does no harm.

Another important change here is to add an `_include` directory into `docs` and ignore it (similar to `_build`).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2000)

PR #2007, opened 2015-09-30
[FLOC 3129] Document using the Flocker Plugin

davidoliverSP2:
Includes changes proposed in https://github.com/ClusterHQ/flocker/pull/2000, and makes additional changes to the Using Flocker section.

We have provided another method of controlling Flocker, so the Using Flocker section has been renamed Controlling Flocker. The information about using the plugin has been placed within here.

The tutorials and examples have been moved out of this section into a new top level section called Tutorials and Examples

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2007)


adamtheturtle:
@davidoliverSP2 What does "This is still work in progress, so should not be merged." and "CODE REVIEW READY" mean? Would you like a check that in general this is OK, so you can continue to refine it? If so use "DESIGN REVIEW READY".

davidoliverSP2:
It means I am still pushing changes to this branch - I'm not sure what the protocol is for a branch in that state?

exarkun:
The JIRA issue is in the "Coding" state.  That seems exactly right.

docs/install/plugin-restrictions.rst
```diff
@@ -0,0 +1,16 @@
+Known Limitations
+=================
+
+* You should not move a volume from one node to another unless you are sure no containers are using the volume.
+
+  The plugin will not stop volumes from being migrated out from underneath a running container.
+  It is possible that Docker or your orchestration tool will prevent this from happening, but Flocker itself does not.
+* ``--volumes-from`` and equivalent Docker API calls will only work if both containers are on the same machine.
```
jml: I don't know whether this is true, but I'll take your word for it.

Do these known limitations have JIRA bugs? They need not be referenced here, but it would be good if the JIRA bugs had notes saying "Update plugin-restrictions.rst when you fix". Otherwise, this risks getting out of date. (Or maybe you have another system for keeping on top of that sort of thing?)

jml:
:+1: Thanks.

PR #2014, opened 2015-10-01
[FLOC-3172] Test syslog configuration

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3172

Not for merging yet.  Note the test fails because the desired functionality regressed (in #1968 I think).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2014)


flocker/acceptance/integration/test_platform.py
```diff
@@ -0,0 +1,47 @@
...
+
+        getting.addCallback(got_messages)
+        return getting
+
+    @require_cluster(1)
+    def test_flocker_control_not_logged(self, cluster):
+        """
+        Log messages from ``flocker-control`` do not appear in syslog.
+        """
+        return self._assert_not_logged(cluster, b"flocker:controlservice:")
```
jml: :question: Why don't you have to exercise the system? I guess it's because `require_cluster` does stuff that we know for a fact logs things? 

I think the test would more clearly communicate intent if they did a log-triggering action, or at least had a comment explaining why we know something has been logged.

jml:
I don't understand why the failing builders don't actually have failures for the tests you added. None of them have a 'problems' field.

jml:
:+1: I'm OK with this code being merged as-is, along with code that actually fixes the test failures.

exarkun:
> I don't understand why the failing builders don't actually have failures for the tests you added. None of them have a 'problems' field.

Because all the test runs crashed before they finished and buildbot doesn't try to parse out the problems for steps that don't complete.

jml:
> > I don't understand why the failing builders don't actually have failures for the tests you added. None of them have a 'problems' field.
> Because all the test runs crashed before they finished and buildbot doesn't try to parse out the problems for steps that don't complete.

Ah, I misread your comment "the test fails because the desired functionality regressed". They do, but there's no evidence of this here, because an unrelated crash happened. Is that right?

exarkun:
Or at least the evidence is hard to dig up.  If you read through the stdout logs I think you do see the desired test failures.

jml:
Ah, right. Not an option on my current net connection.

exarkun:
Unfortunately, an error on one of the builds:

```
[ERROR]
Failure: twisted.internet.error.ProcessTerminated: A process has ended with a probable error condition: process ended with exit code 1.

flocker.acceptance.integration.test_platform.SyslogTests.test_flocker_agent_not_logged
flocker.acceptance.integration.test_platform.SyslogTests.test_flocker_control_not_logged
```

http://build.clusterhq.com/builders/flocker%2Facceptance%2Frackspace%2Fubuntu-14.04%2Frackspace/builds/2728

exarkun:
From the logs,

```
{"timestamp": 1443813282.914059, "task_uuid": "0e1c3e79-9b48-4823-a0da-58f8c6cfd75d", "message_type": "flocker.common.runner:run:stderr", "line": "scp: /var/log/messages: No such file or directory", "task_level": [2]}
```

Hm.

exarkun:
Maybe our Ubuntu systems are configured to have `/var/log/messages` at all, just some Upstart stuff?

PR #2015, opened 2015-10-02
update fpm cron jobs to use flocker AMI FLOC-3180

Azulinho:
Update fpm-cron nightly jobs to use the same AMIs as flocker.
This means one less AMI to maintain.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2015)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -900,7 +900,7 @@ job_type:
   cronly_jobs:
     run_docker_build_centos7_fpm:
       at: '0 0 * * *'
-      on_nodes_with_labels: 'aws-centos-7-T2Medium'
+      on_nodes_with_labels: 'aws-centos-7-SELinux-T2Medium'
       with_steps:
         - { type: 'shell',
             cli: [ *add_shell_functions,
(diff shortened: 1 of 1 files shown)


jml:
Code change is fine. Some builds failed but they look spurious. I've triggered rebuilds. Please wait until tests pass before merging. If tests are genuine failures that require fixes, please resubmit for review.

Azulinho:
" If tests are genuine failures that require fixes, please resubmit for review."
Since the tests don't exercise the build.yaml, what should I fix if they fail ?

(just a side question)

jml:
If the tests fail then:
* it's a spurious failure caused by some failed dependency -> file a bug; retry
* it's an intermittent failure caused by a race condition in a buggy test -> file a bug; retry
* master is broken -> file a bug; fix it
* the tests **do** actually exercise `build.yaml` 

Although neither you nor I think the tests fail because of `build.yaml`, we've both been known to be wrong.

Azulinho:
merging with unrelated failed tests

PR #2016, opened 2015-10-02
Flocker venv random FLOC-3179

Azulinho:
Allow multiple builds on the same slave.
One point of contention was the placement of virtualenv under /tmp/v/, this change uses a random short number under /tmp

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2016)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -931,3 +931,21 @@ job_type:
                    *build_docker_image,
                    *push_image_to_dockerhub ]
           }
+    build_vagrant_basebox_for_flocker_tutorial:
+      at: '0 2 * * *'
+      on_nodes_with_labels: 'mesos-vbox'
+      with_steps:
+        - { type: 'shell',
+            cli: [ *add_shell_functions, *setup_pip_cache,
+                   *cleanup, *setup_venv, *setup_flocker_modules,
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -86,11 +86,9 @@ common_cli:
     # https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/20
     # http://stackoverflow.com/questions/10813538/shebang-line-limit-in-bash-and-linux-kernel
     # https://github.com/spotify/dh-virtualenv/issues/10
-    # so we set our virtualenv dir to live in /tmp/v. We avoid any race
-    # conditions by only executing one build per jenkins_slave at any moment
-    # in time, and cleaning up old virtualenvs at the begining of the build.
+    # so we set our virtualenv dir to live in /tmp/<random number>
     #
-    export venv=/tmp/v
+    export venv=/tmp/\${RANDOM}
```
jml: We do a lot of builds, and  `RANDOM` is only 0-32k, so we need to handle the case where `venv` already exists. Ideally, we'd retry, but I think it's acceptable for now to fail early with an informative error message.

PR #2017, opened 2015-10-02
Flocker venv random FLOC-3179

Azulinho:
Use a different virtualenv path so that we can start having multiple builders per jenkins slave.
This removes a contention point where virtualenv was placed under /tmp/v

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2017)


jml:
LGTM. Same thing as before re tests failures.

Azulinho:
merging with unlreaded failed tests

PR #2018, opened 2015-10-02
Better error messages for admin/run-acceptance-tests

jml:
Raises a slightly more informative error message when running the acceptance tests on a non-existent distro.

Makes some assumptions about the layout of `admin`, but they seem consistent with the rest of the assumptions in the code base.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2018)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -393,12 +393,15 @@ class VagrantRunner(object):
         :raise UsageError: if no such distribution found.
         :return: ``FilePath`` of the vagrant directory.
         """
-        vagrant_path = top_level.descendant([
-            'admin', 'vagrant-acceptance-targets', distribution,
+        vagrant_dir = top_level.descendant([
+            'admin', 'vagrant-acceptance-targets'
         ])
+        vagrant_path = vagrant_dir.child(distribution)
         if not vagrant_path.exists():
-            raise UsageError("Distribution not found: %s."
-                             % (self.distribution,))
+            distributions = vagrant_dir.listdir()
+            raise UsageError(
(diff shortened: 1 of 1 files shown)


admin/acceptance.py
```diff
@@ -375,18 +375,35 @@ class VagrantRunner(object):
...
         if self.variants:
             raise UsageError("Variants unsupported on vagrant.")
 
+    def _get_vagrant_path(self, top_level, distribution):
+        """
+        Get the path to the Vagrant directory for ``distribution``.
+
+        :param FilePath top_level: the directory containing the ``admin``
+            package.
+        :param str distribution: the name of a distribution
```
exarkun: `bytes`?

exarkun:
Somewhat sad we don't have tests for this code ...  Oh well.

Seems like a nice improvement.  Please merge.

admin/acceptance.py
```diff
@@ -375,18 +375,35 @@ class VagrantRunner(object):
...
+        :raise UsageError: if no such distribution found.
+        :return: ``FilePath`` of the vagrant directory.
+        """
+        vagrant_dir = top_level.descendant([
+            'admin', 'vagrant-acceptance-targets'
+        ])
+        vagrant_path = vagrant_dir.child(distribution)
+        if not vagrant_path.exists():
+            distributions = vagrant_dir.listdir()
```
adamtheturtle: I think that this should use the `DISTRIBUTIONS` variable which is created in this file and used in the `optParameters`.

adamtheturtle: Ignore that, different thing.

PR #2020, opened 2015-10-02
[FLOC-3188] Make remote_logs.log JSON

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3188

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2020)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -917,7 +917,13 @@ def journald_json_formatter(output_file):
             accumulated[key] = value
         else:
             if accumulated:
-                message = json.loads(accumulated.get(b"MESSAGE", b"{}"))
+                raw_message = accumulated.get(b"MESSAGE", b"{}")
+                try:
+                    message = json.loads(raw_message)
+                except ValueError:
+                    # Docker log messages are not JSON
+                    message = dict(message=raw_message)
+
                 message[u"_HOSTNAME"] = accumulated.get(
                     b"_HOSTNAME", b"<no hostname>"
                 )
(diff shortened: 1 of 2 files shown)


admin/acceptance.py
```diff
@@ -896,9 +898,41 @@ def capture_journal(reactor, host, output_file):
...
+                    message = dict(message=raw_message)
+
+                message[u"_HOSTNAME"] = accumulated.get(
+                    b"_HOSTNAME", b"<no hostname>"
+                )
+                message[u"_SYSTEMD_UNIT"] = accumulated.get(
+                    b"_SYSTEMD_UNIT", b"<no unit>"
+                )
+                output_file.write(json.dumps(message) + b"\n")
+                accumulated.clear()
```
jml: :grey_question: Not necessary, but I'd probably write this with a yield, for a clearer separation of IO & parsing.

jml:
I think this also fixes https://clusterhq.atlassian.net/browse/FLOC-2097

I guess ideally this should be a thing that's in eliot, since extracting mangled logs from systemd is a general problem. However, happy to start with this.

jml:
Thanks for doing this. I've wanted to do this pretty much since I started at ClusterHQ (but fell over at the first hurdle). 

I've made one optional suggestion. Feel free to merge without further review whether or not you do it.

admin/acceptance.py
exarkun: Agree that'd be better factoring.  I'll leave a note but I don't think putting this off 'til later will hurt anything.

PR #2022, opened 2015-10-02
[FLOC-3192] Switch to newer pyrsistent

itamarst:
In theory this version doesn't segfault when the C extension is used... and using the C extension speeds things up a lot. Also this new version has way less bugs, some of them serious.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2022)

### DIFF
--- a/flocker/control/test/test_model.py
+++ b/flocker/control/test/test_model.py
@@ -524,9 +524,9 @@ class DeploymentTests(SynchronousTestCase):
                 image=DockerImage.from_string(u"image"))}),
         )
         deployment = Deployment(nodes=frozenset([node, another_node]))
-        self.assertItemEqual(set(deployment.applications()),
-                             set(node.applications) +
-                             set(another_node.applications))
+        self.assertItemsEqual(set(deployment.applications()),
+                              set(node.applications) |
+                              set(another_node.applications))
 
     def test_update_node_retains_leases(self):
         """
--- a/flocker/control/test/test_model.py
+++ b/flocker/control/test/test_model.py
@@ -524,9 +524,9 @@ class DeploymentTests(SynchronousTestCase):
                 image=DockerImage.from_string(u"image"))}),
         )
         deployment = Deployment(nodes=frozenset([node, another_node]))
-        self.assertEqual(sorted(list(deployment.applications())),
-                         sorted(list(node.applications) +
-                                list(another_node.applications)))
+        self.assertItemEqual(set(deployment.applications()),
+                             set(node.applications) +
+                             set(another_node.applications))
 
     def test_update_node_retains_leases(self):
         """
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -100,7 +100,7 @@ class _InParallel(PRecord):
         # Sort the changes for the benefit of comparison.  Stick with a vector
         # (rather than, say, a set) in case someone wants to run the same
         # change multiple times in parallel.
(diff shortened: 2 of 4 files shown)


flocker/node/_change.py
```diff
@@ -100,7 +100,7 @@ class _InParallel(PRecord):
         # Sort the changes for the benefit of comparison.  Stick with a vector
         # (rather than, say, a set) in case someone wants to run the same
         # change multiple times in parallel.
-        factory=lambda changes: pvector(sorted(changes)),
+        factory=lambda changes: pvector(sorted(changes, key=lambda o: id(o))),
```
jml: I'm pretty suWhy not just `key=id`?

jml:
LGTM, assuming test failures are spurious.

exarkun:
neowm

`flocker.control.test.test_persistence.WireEncodeDecodeTests.test_roundtrip ... process killed by signal 11`

itamarst:
Even with C extension disabled there's a whole bunch of bugs in older versions of pyrsistent, many of which we encountered, so I think it's worth upgrading regardless.

itamarst:
0.11.7 should fix the segfault, but we'll see.

flocker/control/test/test_model.py
```diff
@@ -524,9 +524,9 @@ def test_applications(self):
                 image=DockerImage.from_string(u"image"))}),
         )
         deployment = Deployment(nodes=frozenset([node, another_node]))
-        self.assertEqual(sorted(list(deployment.applications())),
-                         sorted(list(node.applications) +
-                                list(another_node.applications)))
+        self.assertItemEqual(set(deployment.applications()),
```
tomprince: typo.

dwgebler:
I can only say the segfault has not occurred on the latest build and I haven't been able to reproduce a segfault locally from this branch either....not really a guarantee it's been fixed, but as you say, even if we have to re-disable the C extension, the newer version of Pyrsistent is still worth using, so I'll accept this as LGTM.

exarkun:
Why re-enable the C extension as part of this change?  Can't we upgrade for the bug fixes without doing that?

Wouldn't it be nice to have a repeatable benchmark showing the C extension is at least worth some risk of crashing our processes?

dwgebler:
> Why re-enable the C extension as part of this change?

We have at least some reason to believe the segfault problem is fixed in this version - http://pyrsistent.readthedocs.org/en/latest/changes.html - and Itamar's benchmarking on the JIRA issue indicates a >40% performance increase in `wire_encode`

> Can't we upgrade for the bug fixes without doing that?

Can't we just as easily re-disable the C extension if we find any future problems?

exarkun:
> > Can't we upgrade for the bug fixes without doing that?
> Can't we just as easily re-disable the C extension if we find any future problems?

Not as easily as we can *not* re-disable it if we *don't* re-enable it now.

Change for the sake of change is a bad thing.  Performance changes should be justified by measurements demonstrating their value - just like functional changes are justified by tests demonstrating their correctness.

dwgebler:
> Performance changes should be justified by measurements demonstrating their value

I haven't done any benchmarking myself, but Itamar indicates on the issue that "Re-enabling the C extension changes time to wire_encode a 30 node, 60 application deployment from about 4.2ms per encode to 2.4ms."

exarkun:
The key word is "repeatable".

Please require repeatable benchmarks in code reviews on future optimizations.

PR #2023, opened 2015-10-03
Create jenkins job vagrant basebox FLOC-2852

Azulinho:
This adds the BB vagrant build-tutorial job to Jenkins,
it runs on an ubuntu mesos cluster and produces the basebox for the build-tutorial.
(it doesn't get pushed anywhere, it is simply stored locally on the local box)

It is not running any of the triggered jobs that currently run on BB, those can be added later on a separate ticket (if they're still wanted) as per the BB logs they don't seem to have passed a build for a long time.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2023)


build.yaml
```diff
@@ -235,7 +241,9 @@ common_cli:
     # during the bootstrapping of the slave.
     # This file is located in /etc/jenkins_slave on the Master box.
     # These variables contain the PIP URL and IP address of our caching server.
-    . /tmp/pip.sh
+    if [ -e /tmp/pip.sh ]; then
+      . /tmp/pip.sh
+    fi
```
Azulinho: only set the PIP caching if the file exists (currently not the case on Mesos and OSX)

build.yaml
```diff
@@ -243,25 +251,26 @@ common_cli:
...
+        PIP_ADDITIONAL_OPTIONS="\${PIP_ADDITIONAL_OPTIONS} -i \${PIP_INDEX_URL} "
+    fi
+    if [ \${TRUSTED_HOST} ]; then
+        PIP_ADDITIONAL_OPTIONS="\${PIP_ADDITIONAL_OPTIONS} \
+                                --trusted-host \${TRUSTED_HOST} "
+    fi
+
+    # parse_logs() currently only works on Ubuntu or Centos builds
+    # due to a different awk implementation on OSX cloudbees servers
     if is_ubuntu || is_centos; then
```
Azulinho: suggestion from Tom on a different ticket, only set the pip options if they're available in the environment

build.yaml
```diff
@@ -243,25 +251,26 @@ common_cli:
...
-      # variables to the slave.
-      # So we just go upstream for the python modules.
-      PIP_ADDITIONAL_OPTIONS=""
-      PARSE_LOGS=""
-    fi
 
     # using the caching-layer, install all the dependencies
-    \${PARSE_LOGS} pip install --upgrade pip \${PIP_ADDITIONAL_OPTIONS}
+    \${PARSE_LOGS} pip install --upgrade pip
     \${PARSE_LOGS} pip install . \${PIP_ADDITIONAL_OPTIONS}
```
Azulinho: if our pip is <7, using the pip cache and 'trusted-host' would fail as pip<7 doesn't have --trusted-host option

build.yaml
```diff
@@ -474,10 +483,10 @@ common_cli:
...
     # create a repo on either centos or ubuntu
-    if is_ubuntu; then
+    if [ "\${DISTRIBUTION_NAME}" == "ubuntu-14.04" ]; then
       sudo sh -c 'dpkg-scanpackages --multiversion . | gzip > Packages.gz'
     fi
-    if is_centos; then
+    if [ "\${DISTRIBUTION_NAME}" == "centos-7" ]; then
       sudo createrepo .
     fi
     cd -
```
Azulinho: we actually build a package for a distribution and not a package for the OS we are running on

Azulinho: this should work on Archlinux too : - )

build.yaml
```diff
@@ -567,6 +576,12 @@ common_cli:
...
     echo "RUN pip install -r /tmp/requirements.txt" >> Dockerfile
 
+  build_vagrant_tutorial_basebox: &build_vagrant_tutorial_basebox |
+    vagrant destroy -f
+    parse_logs \$venv/bin/python admin/build-vagrant-box --box tutorial \
+      --branch \${TRIGGERED_BRANCH} --build-server  http://\${eth0_ip}
+    parse_logs vagrant box add --force "clusterhq/flocker-tutorial" \
+      vagrant/tutorial/flocker-tutorial-\${FLOCKER_VERSION}.box
 
 #-----------------------------------------------------------------------------#
```
Azulinho: our vagrant build-tutorial job

build.yaml
```diff
@@ -934,3 +949,18 @@ job_type:
...
+            cli: [ *add_shell_functions, *setup_pip_cache,
+                   *cleanup, *setup_venv, *setup_flocker_modules,
+                   *check_version,
+                   'export DISTRIBUTION_NAME=centos-7',
+                   *build_sdist, *build_package,
+                   *build_repo_metadata,
+                   'export HOME=/root',
+                   *build_vagrant_tutorial_basebox,
+                   *clean_packages]
+          }
```
Azulinho: we only build the master branch once every 24hours, all other branches can be built on demand as needed

jobs.groovy.j2
```diff
@@ -149,8 +149,11 @@ folder("{{folder}}") { displayName("{{display_name}}") }
...
 #}
     triggers {
-{#  the job_type 'cron' is used by the docker_build jobs running every 24h   #}
-{%      if _type == 'cron'                                                   %}
+{#  the job_type 'cron' is used by the docker_build jobs running every 24h
+    but we only configure the scheduler if the jobs is for the master branch.
+    If we were to schedule the job on every branch we would have multiple jobs
+    running at the same time                                                 #}
+{%      if _type == 'cron' and _branch == 'master'                           %}
 {#  the cron  string below is a common crontab style string                  #}
```
Azulinho: only add the schedulle if this is the master branch

jobs.groovy.j2
```diff
@@ -435,7 +438,7 @@ branches.each {
 
 
 {# the multijob is responsible for running all configured jobs in parallel   #}
-  multiJob("${dashProject}/${dashBranchName}/_main_multijob") {
+  multiJob("${dashProject}/${dashBranchName}/__main_multijob") {
     wrappers {
```
Azulinho: make sure our main multijob is the first one to show up when sorted by name

Azulinho: the cron ones are listed just below
(maybe the could actually be called cron?)

jobs.groovy.j2
```diff
@@ -594,12 +597,13 @@ branches.each {
...
 {%    if job_type == 'cronly_jobs'                                          -%}
 {%        set _job_name = job_name                                          -%}
-{# lets place these jobs inside the 'master' branch folder, as its likely
-   people will be looking into that view when monitoring the master branch   #}
-  job("${dashProject}/master/_{{ _job_name }}") {
+  job("${dashProject}/${branchName}/_{{ _job_name }}") {
+    parameters {
+        textParam("TRIGGERED_BRANCH", "${branchName}", "Branch that triggered this job" )
+    }
     label("{{ job_values.on_nodes_with_labels }}")
```
Azulinho: we need to pass the branch name to our vagrant build tutorial box so that it can build the package for the right OS

build.yaml
```diff
@@ -243,25 +251,26 @@ common_cli:
     # We have a devpi cache in AWS which we will consume instead of going
     # upstream to the PyPi servers.
     # We specify that devpi caching server using -i \$PIP_INDEX_URL
-    # which requires as to include --trusted_host as we are not (yet) using
+    # which requires as to include --trusted-host as we are not (yet) using
     # SSL on our caching box.
     # The --trusted-host option is only available with pip 7.
     #
+    if [ \${PIP_INDEX_URL} ]; then
+        PIP_ADDITIONAL_OPTIONS="\${PIP_ADDITIONAL_OPTIONS} -i \${PIP_INDEX_URL} "
```
adamtheturtle: Is it possible to just upgrade pip here and then use the `--trusted-host` option knowing that it will be available?

Azulinho: it may not be available, the contents of those vars are loaded to the slave. But not all slaves have them (OSX builds on cloudbees don't have them).
So it's better to check if they're defined (this was actually Tom's suggestion a cpl of weeks ago)

zubron:
The current box is not uploaded anywhere as it will conflict with the same box being uploaded by buildbot. We should file an issue to make sure that this is enabled before we transition from buildbot.

Currently the box is added directly (not using a JSON metadata file) and as a result the output will show that the box is version "v0", not the relevant flocker version. I don't know how important this version is. We should also determine whether we will need to create the metadata file. If we do need it, we should also file an issue that is resolved when moving to Jenkins.

Otherwise, I am happy with the changes. Please merge once the relevant issues have been created.

zubron:
See [FLOC 3208](https://clusterhq.atlassian.net/browse/FLOC-3208) for uploading the box to S3.

Azulinho:
thanks for raising the backlog ticket and for the review,
effort and time spent on it very appreciated.

ice-cream for you

zubron:
See [FLOC 3209](https://clusterhq.atlassian.net/browse/FLOC-3209) for figuring out whether we need the metadata or the version.

PR #2024, opened 2015-10-05
[FLOC-3203] Raise error if we can't compute instance ID

jml:
In some circumstances, `compute_instance_id` can return `None`. This is not a valid response according to the documentation, and most of our code base assumes that `compute_instance_id` returns a valid instance ID.

Thus, rather than returning `None`, we here raise an error.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2024)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -120,7 +120,9 @@ class UnknownInstanceID(Exception):
     Could not compute instance ID for block device.
     """
     def __init__(self, blockdevice):
-        Exception.__init__(self, blockdevice)
+        Exception.__init__(
+            self,
+            'Could not find valid instance ID for {}'.format(blockdevice))
         self.blockdevice = blockdevice
 
 
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -115,6 +115,15 @@ class FilesystemExists(Exception):
         self.device = device
 
 
+class UnknownInstanceID(Exception):
+    """
+    Could not compute instance ID for block device.
+    """
+    def __init__(self, blockdevice):
+        Exception.__init__(self, blockdevice)
+        self.blockdevice = blockdevice
+
+
 DATASET = Field(
     u"dataset",
     lambda dataset: dataset.dataset_id,
@@ -806,7 +815,8 @@ class IBlockDeviceAsyncAPI(Interface):
 
         :returns: A ``Deferred`` that fires with ``unicode`` of a
             provider-specific node identifier which identifies the node where
-            the method is run.
+            the method is run, or fails with ``UnknownInstanceID`` if we cannot
+            determine the identifier.
         """
 
     def create_volume(dataset_id, size):
@@ -879,6 +889,8 @@ class IBlockDeviceAPI(Interface):
         to determine which volumes are locally attached and it will be used
         with ``attach_volume`` to locally attach volumes.
 
+        :raise UnknownInstanceID: If we cannot determine the identifier of the
+            node.
         :returns: A ``unicode`` object giving a provider-specific node
             identifier which identifies the node where the method is run.
         """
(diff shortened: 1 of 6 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -1206,6 +1218,8 @@ def allocation_unit(self):
         return self._allocation_unit
 
     def compute_instance_id(self):
+        if not self._compute_instance_id:
+            raise UnknownInstanceID(self)
         return self._compute_instance_id
```
exarkun: This part of the change should be unit-testable.

flocker/node/agents/cinder.py
```diff
@@ -449,13 +449,14 @@ def compute_instance_id(self):
...
         # But we don't currently test this directly. See FLOC-2281.
-        if len(matching_instances) == 1:
+        if len(matching_instances) == 1 and matching_instances[0]:
             return matching_instances[0]
         # If there was no match, or if multiple matches were found, log an
         # error containing all the local and remote IPs.
         COMPUTE_INSTANCE_ID_NOT_FOUND(
             local_ips=local_ips, api_ips=api_ip_map
         ).write()
+        raise UnknownInstanceID(self)
```
exarkun: It would be pretty nice to have automated test coverage for this case, particularly if we anticipate ever meaningfully handling `UnknownInstanceID` in the deployer.  A verified fake of nova might make it possible to unit test this (I hear there is one, I don't think we use it anywhere for anything yet).

flocker/node/agents/ebs.py
```diff
@@ -609,7 +609,10 @@ def compute_instance_id(self):
         """
         Look up the EC2 instance ID for this node.
         """
-        return get_instance_metadata()['instance-id'].decode("ascii")
+        instance_id = get_instance_metadata().get('instance-id', None)
+        if not instance_id:
```
exarkun: Oughtn't this be something more like `instance_id is None`?  Or a handler for `KeyError`?

Any idea if this case is actually possible or is it just that it looks possible due to symmetry with OpenStack?

exarkun: To expand on that last, it would seem quite odd to me that `get_instance_metadata` would ever return a structure where `instance-id` is missing entirely.  I might expect a failure where the instance id supplied is actually inconsistent with some other part of AWS and attempting to use it in another API call produces an error like "no such instance" - which this change would not guard against.

exarkun:
The motivation for the change seems reasonable.  The resulting failures should at least be a little easier to interpret.  It's a slight bummer that we won't be automatically exercising these new code paths in a demonstration that the failures actually look like what we're hoping they'll now look like.  I think it's at least reasonable to test the loopback case.  The cinder and ebs cases may be too hard to reproduce for it to be worthwhile.  And since we should be able to reason about them via their parallelism with the loopback code maybe that's okay.

So please add a test demonstrating the new exception provides the debugging experience you're seeking, address the other inline comments that you think are worthy, and then merge.  Thanks.

flocker/node/agents/ebs.py
jml: Last job had v. strong style constraints against `foo is None` (preferring `not foo`) unless there was a particular reason for it. Am happy to change to `is None` here, but don't see the merit of one over the other. 

Likewise, I don't see much meaningful distinction between `get`-with-out-of-band-default and a `KeyError` handler.

jml: Depends on what you mean by "possible". I have no evidence that it's occurred, and if it did, it would be a bug in AWS. However, it's just a streams of bits from some computer on the Internet, anything is possible. 

Having the check here helps by:
* catching the problem as soon as we're aware of it
* providing symmetry with other implementations

and harms by:
* increasing complexity of code in this function
* doing work up-front that might not be necessary

Truly, I don't know what's best. I guess I put it here because in other languages I use, in any lookup we'd be forced to locally handle the case of the lookup failing, which I think reduces mental overhead.

jml: Changed to `is None`. Didn't make other changes  here.

jml:
* Added loopback test
* Did not add ebs or cinder tests (I don't think going down the fake nova client rabbithole is the right thing to do now)

PR #2025, opened 2015-10-05
[FLOC-3046] Cleanups for virtio stuff

jml:
Based on @jongiddy's comments in #1922. 
 
I've implemented most of his suggestions, but chickened out on re-using the minimum device sizes. It looks like the code is a great big tangle there (see [FLOC-2584](https://clusterhq.atlassian.net/browse/FLOC-2584)), and although I *do* think it's a good idea to disentangle it, I personally don't want to do it right now.

I pulled the comment on the symlink resolution from @wallrj's [insights on the original PR](https://github.com/ClusterHQ/flocker/pull/1921).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2025)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -342,6 +342,9 @@ def configured_cluster_for_nodes(
         configured.
     """
     default_volume_size = GiB(1)
+    # XXX: TODO: This looks like it should come from
+    # node.agents.test.blockdevicefactory.MINIMUM_ALLOCATABLE_SIZES. Or maybe
+    # they should both come from a non-test file.
     if dataset_backend_configuration.get('auth_plugin') == 'rackspace':
         default_volume_size = GiB(100)
 
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -341,10 +341,23 @@ def configured_cluster_for_nodes(
     :returns: A ``Deferred`` which fires with ``Cluster`` when it is
         configured.
     """
+    # XXX: There is duplication between the values here and those in
+    # f.node.agents.test.blockdevicefactory.MINIMUM_ALLOCATABLE_SIZES. We want
+    # the default volume size to be greater than or equal to the minimum
+    # allocatable size.
+    #
+    # Ideally, the minimum allocatable size (and perhaps the default volume
+    # size) would be something known by an object that represents the dataset
+    # backend. Unfortunately:
+    #  1. There is no such object
+    #  2. There is existing confusion in the code around 'openstack' and
+    #     'rackspace'
+    #
+    # Here, we special-case Rackspace (presumably) because it has a minimum
+    # allocatable size that is different from other Openstack backends.
+    #
+    # FLOC-2584 also discusses this.
     default_volume_size = GiB(1)
-    # XXX: TODO: This looks like it should come from
-    # node.agents.test.blockdevicefactory.MINIMUM_ALLOCATABLE_SIZES. Or maybe
-    # they should both come from a non-test file.
     if dataset_backend_configuration.get('auth_plugin') == 'rackspace':
         default_volume_size = GiB(100)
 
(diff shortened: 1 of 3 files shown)


flocker/node/agents/cinder.py
```diff
@@ -602,6 +602,17 @@ def _get_device_path_virtio_blk(self, volume):
         expected_path = FilePath(
             "/dev/disk/by-id/virtio-{}".format(volume.id[:20])
         )
+        # Return the real path to avoid two problems:
```
adamtheturtle: :question: Make this clearer "instead of the symbolic link"

flocker/node/agents/functional/test_cinder.py
```diff
@@ -49,7 +49,33 @@
 
 # Tests requiring virsh can currently only be run on a devstack installation
 # that is not within our CI system. This will be addressed with FLOC-2972.
-require_virsh = skipIf(
+#
+# In the meantime, you can do the following (provided you have access):
```
adamtheturtle: :question: I would prefer it if this were in the contributing docs

adamtheturtle:
Thanks, the only comments are minor and optional. Let's keep in mind to address the minimum device sizes next time it is a bother, else it will grow into even more of a mess. Resubmit if there are major changes and merge when you are happy.

admin/acceptance.py
```diff
@@ -341,6 +341,22 @@ def configured_cluster_for_nodes(
...
+    # XXX: There is duplication between the values here and those in
+    # f.node.agents.test.blockdevicefactory.MINIMUM_ALLOCATABLE_SIZES. We want
+    # the default volume size to be greater than or equal to the minimum
+    # allocatable size.
+    #
+    # Ideally, the minimum allocatable size (and perhaps the default volume
+    # size) would be something known by an object that represents the dataset
+    # backend. Unfortunately:
+    #  1. There is no such object
+    #  2. There is existing confusion in the code around 'openstack' and
```
adamtheturtle: Is this something to do with https://clusterhq.atlassian.net/browse/FLOC-3017?

flocker/node/agents/functional/test_cinder.py
```diff
@@ -49,7 +49,33 @@
 
 # Tests requiring virsh can currently only be run on a devstack installation
```
adamtheturtle: "requiring virtio"?

flocker/node/agents/cinder.py
jml: Done.

flocker/node/agents/functional/test_cinder.py
jml: Not done.

What's the reason for your preference?

I put it in the comments because:
 * it's temporary documentation until FLOC-2972 (mentioned in the comment) is fixed
 * no one outside ClusterHQ has access to the machine
 * the only time when people will want to follow these instructions is when they are doing stuff related to this code

admin/acceptance.py
jml: Possibly! I'll link the two tickets.

flocker/node/agents/functional/test_cinder.py
jml: Fixed.

jml:
Thanks!

flocker/node/agents/functional/test_cinder.py
adamtheturtle: > What's the reason for your preference?

That's where I look when I want to know how to run tests. We have the acceptance test docs in the contributing docs and this seems similar so I thought it fit into a similar place.

jml: > > What's the reason for your preference?

> That's where I look when I want to know how to run tests. We have the acceptance test docs in the contributing docs and this seems similar so I thought it fit into a similar place.

Interesting. I look at rendered documentation as a last resort. Probably mostly because of habit.  Maybe also to avoid the context switch and the inevitable intermediate problem of finding the relevant documentation.

adamtheturtle: > the inevitable intermediate problem of finding the relevant documentation.

In some places we have links in the code to doc-dev

PR #2026, opened 2015-10-05
[FLOC-3195] Use boto for provisioning AWS acceptance test nodes.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3195

This allows us to take advantage of boto retrying calls that fail due to `RequestLimitExceeded`.

In theory, this should reduce the incidents of getting that error, but I don't have data to back that up.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2026)

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -63,6 +63,7 @@ def _wait_until_running(instance):
                 instance_state=instance.state,
             ):
                 sleep(1)
+            instance.update()
 
 
 @implementer(INode)
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -21,6 +21,8 @@ from ._install import (
     task_install_ssh_key,
 )
 
+from eliot import start_action
+
 from ._ssh import run_remotely, run_from_args
 from ._effect import sequence
 
@@ -40,6 +42,29 @@ IMAGE_NAMES = {
 }
 
 
+def _wait_until_running(instance):
+    """
+    Wait until a instance is running.
+
+    :param boto.ec2.instance.Instance instance: The instance to wait for.
+    """
+    with start_action(
+        action_type=u"flocker:provision:aws:wait_until_running",
+        instance_id=instance.id,
+    ):
+        state = instance.state
+        while state != 'running':
+            with start_action(
+                action_type=u"flocker:provision:aws:wait_until_running:sleep",
+                instance_state=instance.state,
+            ):
+                sleep(1)
+            instance.update()
+            new_state = instance.state
+            if new_state != state:
+                state = new_state
+
+
 @implementer(INode)
 class AWSNode(PClass):
     _provisioner = field(mandatory=True)
@@ -56,7 +81,11 @@ class AWSNode(PClass):
         return self._instance.private_ip_address.encode('ascii')
 
     def destroy(self):
-        self._instance.terminate()
+        with start_action(
+            action_type=u"flocker:provision:aws:destroy",
+            instance_id=self._instance.id,
+        ):
+            self._instance.terminate()
 
     def get_default_username(self):
         """
@@ -112,15 +141,12 @@ class AWSNode(PClass):
         """
 
         def do_reboot(_):
-            self._node.reboot()
-            state = self.instance.state
-            while state != 'running':
-                sleep(1)
-                self.instance.update()
-                new_state = self.instance.state
-                if new_state != state:
-                    state = new_state
-                    print 'Instance', state
+            with start_action(
+                action_type=u"flocker:provision:aws:reboot",
+                instance_id=self._instance.id,
+            ):
+                self._instance.reboot()
+                _wait_until_running(self._instance)
 
         return run_remotely(
             username="root",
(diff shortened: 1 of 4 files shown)


flocker/provision/_aws.py
```diff
@@ -102,45 +271,18 @@ def aws_provisioner(access_key, secret_access_token, keyname,
...
-            # disable this, so we can use sudo with conch.
-            "ex_userdata": dedent("""\
-                #!/bin/sh
-                sed -i '/Defaults *requiretty/d' /etc/sudoers
-                """)
-        }
-
-    provisioner = LibcloudProvisioner(
-        driver=driver,
+    # Import these here, so that this can be imported without installng
```
adamtheturtle: No longer relevant comment.

Can we just expect that all dev requirements will be available?

Existing typo "installng" > "installing".

flocker/provision/_aws.py
```diff
@@ -27,70 +34,232 @@
...
+                _instance=instance,
+                distribution=distribution,
+            )
 
 
 def aws_provisioner(access_key, secret_access_token, keyname,
                     region, zone, security_groups):
     """
-    Create a LibCloudProvisioner for provisioning nodes on AWS EC2.
+    Create a IProvisioner for provisioning nodes on AWS EC2.
```
adamtheturtle: nit: a > an

jongiddy:
Acceptance tests are failing to use of leading underscores in `AWSProvisioner` class field names, but not in `aws_provisioner` function.

flocker/provision/_common.py
```diff
@@ -69,3 +72,74 @@ class Cluster(PRecord):
...
+
+        :return Effect:
+        """
+
+
+class IProvisioner(Interface):
+    """
+    A provisioner for creating nodes to run acceptance tests agasint.
+    """
+    default_size = InterfaceAttribute(
```
jongiddy: It doesn't seem that this attribute is required to be part of the interface. It is only used by the implementation of `create_node`. Leave it out until it's actually needed by external code.

flocker/provision/_aws.py
```diff
@@ -27,70 +34,232 @@
...
+        username = self.get_default_username()
+
+        commands = []
 
+        # cloud-init may not have allowed sudo without tty yet, so try SSH key
+        # installation for a few more seconds:
+        start = []
 
-def provision_aws(node, package_source, distribution, variants):
+        def for_ten_seconds(*args, **kwargs):
```
jongiddy: This function could be more self-contained by defining it as `for_ten_seconds(exc_info, start=[])`.  This also sets us up for identifying any "permanent" exceptions which should not be retried.

flocker/provision/_aws.py
```diff
@@ -27,70 +34,232 @@
...
+            if not start:
+                start.append(time())
+            return Effect(Constant((time() - start[0]) < 30))
```
jongiddy: This seems to wait 30 seconds rather than 10?  Definitely change the function name. Maybe a better solution is something like:
```
def for_time_period(seconds):
    start = []
    def check(exc_info):
        if not start:
            start.append(time())
        return Effect(Constant((time() - start[0]) < seconds))
    return check

... commands=retry(task_install_ssh_key(), for_time_period(30))
```
This also sets us up for checking any permanent exceptions that should not be retried.

jongiddy:
Please address comments, and resubmit. What are the plans for applying the new interfaces to the Rackspace code?

tomprince:
> What are the plans for applying the new interfaces to the Rackspace code?

Depending on how you look at it, either never or it has already been done. The rackspace code is a thin wrapper around the libcloud code which already has the interfaces applied.

flocker/provision/_aws.py
tomprince: I've renamed the function. I'm not going to rewrite this, since this is just existing code that got moved.

tomprince:
The ubuntu acceptance test failures are due to https://github.com/docker/docker/issues/16820

thaJeztah:
@tomprince should be resolved now!

flocker/provision/_aws.py
```diff
@@ -27,70 +39,227 @@
...
+    ):
+        state = instance.state
+        while state != 'running':
+            with start_action(
+                action_type=u"flocker:provision:aws:wait_until_running:sleep",
+                instance_state=instance.state,
+            ):
+                sleep(1)
+            instance.update()
+            new_state = instance.state
```
itamarst: :question: `state = instance.state`? `new_state` seems unnecessary (left over from earlier version of code maybe.)

flocker/provision/_common.py
itamarst: Seems like this wasn't addressed?

itamarst:
:question: We had to configure boto somehow for EBS driver so we got better results out of retry logic, looks like that isn't done here.

itamarst:
Address and merge.

tomprince:
>  :question: We had to configure boto somehow for EBS driver so we got better results out of retry logic, looks like that isn't done here.

It looks like the only thing done is to boost the number of retries? I think it makes sense to try with the defaults, and see if that is sufficient.

PR #2027, opened 2015-10-06
fix test_resize_volume_invalid_max_size

avg-I:
First, resized_filesystem() callback was unused.
Second and more important, if the error-callback maximum_too_small()
was not called, e.g. because a backend did not try to resize
a volume at all, then the test would just pass.
Now we ensure that the resize operation fails with
MaximumSizeTooSmall exception.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2027)


jongiddy:
:+1: OK to merge when all tests green.

flocker/volume/filesystems/memory.py
```diff
@@ -177,8 +178,19 @@ def destroy(self, volume):
...
+            for f in root.walk():
+                if not f.isdir():
+                    total_size += f.getsize()
+            return total_size
+
+        current_size = get_size()
         size_path = root.child(b".size")
         if volume.size.maximum_size is not None:
+            if volume.size.maximum_size < current_size:
+                raise MaximumSizeTooSmall
```
exarkun: Hm.  I wish the style checker pointed this out because it probably is just a matter of style - but typically we raise instances, not types (ie, `raise MaximumSizeTooSmall()` instead).  The fact that both of these are allowed is a weird quirk resulting from an internal CPython implementation detail.

But as I said it is basically just a style thing.  I can't think of any behavior that differs based on which you use.  And since it's apparently not part of the style the pep8 tool checks for it's not really enforced.

avg-I: Actually not sure why I wrote that line that way. I also have a habit of raising an instance of an exception. So, I will fix this if only for the consistency sake.

PR #2029, opened 2015-10-06
use more portable /dev/fd instead of /proc/self/fd

avg-I:
The latter seems to be Linux specific while the former is available
on FreeBSD and illumos as well as on Linux (via a symbolic link).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2029)


jml:
* Code looks good to me
* Note that without having some sort of builder for this, we're likely to regress on our compatibility

But no changes required by me. There are a bunch of test failures, if any are caused by this change, then please resubmit for review.

PR #2030, opened 2015-10-06
[FLOC-3182] Only attempt to configure the firewall on operating systems that have a firewall installed

wallrj:
Fixes: 
 * https://clusterhq.atlassian.net/browse/FLOC-3182

Also fixes:
 * https://clusterhq.atlassian.net/browse/FLOC-1550

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2030)


wallrj:
Hmm. I should probably still only run the Docker related firewall commands when provisioning Rackspace nodes.

jongiddy:
Many system admins would object to having a package open firewall ports automatically.  I think that is a non-starter.

I suggest looking at PR #2026 which introduces an `INode` interface, and an `AWSNode`.  If this was added to Rackspace, then `INode` could have an `open_firewall` method to return the correct firewall incantation for different platforms.

exarkun:
> Many system admins would object to having a package open firewall ports automatically. I think that is a non-starter.

Should we get input from the product team about this?

exarkun:
(If I recall correctly, we developed the current system of firewall interactions without any input from the product team.  If that's the case then we don't have any clear evidence that the current system is particularly more desirable than the proposed change from their perspective.)

jongiddy:
As a quick fix to the fact that the firewall is not enabled on AWS, you could change the `open_firewalld` to use `run` and then test for existence. See the `task_install_docker` command.

wallrj:
After discussion with @jongiddy @Azulinho and @tomprince I reverted the previous approach -- packaged post installation hooks for configuring the firewall.

Instead I've now figured out how to handle process error exit codes in sequences of `Effect` s so I can test for ``firewall-cmd`` and only run the firewall steps if that's present on the target system.

That also allows me to run the ``flocker-control`` configuration only on the control node and the docker API firewall configuration on all nodes while hiding the docker API steps from the documentation.

wallrj:
There's a cleaner version of this branch in https://github.com/ClusterHQ/flocker/pull/2042

PR #2033, opened 2015-10-07
[FLOC 3132] Update the Installing Overview page

davidoliverSP2:
Fixes 3132

This PR moves the Flocker demo info (TrueAbility and Vagrant) into a section at the end of "Introduction to Flocker", and rewrites the overview of Installation to talk about the Labs Installer, and a simplified overview of the full install steps.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2033)

### DIFF
--- a/docs/introduction/index.rst
+++ b/docs/introduction/index.rst
@@ -160,12 +160,14 @@ Configuration details for each of the backends can be found in the :ref:`Configu
 
 .. XXX FLOC 3144 - add a link here to the instructions on how to create your own backend driver
 
+.. _get-started:
+
 Get Started with Flocker
 ========================
 
 Before you begin to install Flocker, there are a couple of ways in which you can try out Flocker without having to go through the full installation steps:
 
-Flocker in a Live Hosted Environment
+Flocker in a Live Hgosted Environment
 --------------------------------------
 
 You can take Flocker for a spin using a free, live hosted environment. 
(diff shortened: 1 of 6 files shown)


jongiddy:
As an overall design issue, I think clicking on "Installing Flocker" to find a "No installation" option, which is not actually a practical option, to be confusing. To my mind, a "No installation" option in this section would mean a "cloud" option (e.g. our JIRA is a "no installation" installation of JIRA). This option clearly belongs to the "Introduction to Flocker" section.

jongiddy:
The text is fine, if we decide to structure it this way. However, I would prefer the Installing Flocker section to contain only instructions for installing practical Flocker installations, and that is either the Quick Start Installer or the manual process.

The No Installation and Vagrant "play" installations would be better in the Introduction to Flocker section, where people are evaluating whether to use Flocker.  In fact, after "Flocker Basics", it seems to be a big jump into the Flocker Plugin description and demo, with no further reference to Flocker. I think the "play" Flocker installations should go here, either before or after the plugin stuff.

docs/introduction/index.rst
```diff
@@ -159,3 +159,37 @@ The following backends can be used with Flocker:
...
+* To give you time to play, you’ll have access to the environment for 3 hours.
+
+For more information, see the `Try Flocker`_ page.
+
+.. _vagrant-install:
+
+Flocker Installed Locally Using Vagrant
+---------------------------------------
+
+If you don’t want to use real servers, you can set up a cluster locally on virtual machines using our Vagrant image.
```
jongiddy: Out of its previous context, the clause "If you don't want to use real servers" seems wrong. Perhaps "For quick experiments, you can set up..."

jongiddy:
Looks good to me. Please address one comment and merge.

PR #2035, opened 2015-10-07
[FLOC 3228] lower the number of visible links from the left hand docs menu

davidoliverSP2:
Fixes 3228

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2035)


jongiddy:
Looks good. Apart from the length, it is also much less overwhelming.  `docs` tests are OK, and failing acceptance tests are usual suspects, and not related to this HTML-only change. Please merge.

PR #2039, opened 2015-10-08
[FLOC-3227] Allow managed runner to operate on AWS nodes

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3227

This allows managed runner to be configured with a dictionary mapping ``private_address`` to ``public_address``.
This is required when nodes do not have the public IP address configured locally, as is the case with AWS.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2039)


docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -223,6 +235,70 @@ The nodes should be configured to allow key based SSH connections as user ``root
...
+   managed:
+     addresses: {
+       "10.230.191.121": "54.158.225.35",
+       "10.69.174.223": "54.159.119.143"
+     }
+
+
+And then run the acceptance tests on those nodes using the following command:
+
+.. code-block:: yaml
```
tomprince: This isn't YAML.

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -223,6 +235,70 @@ The nodes should be configured to allow key based SSH connections as user ``root
...
+   To run acceptance tests against these nodes, set the following environment variables:
+
+   export FLOCKER_ACCEPTANCE_DEFAULT_VOLUME_SIZE=1073741824;
+   export FLOCKER_ACCEPTANCE_CONTROL_NODE=54.159.119.143;
+   export FLOCKER_ACCEPTANCE_HOSTNAME_TO_PUBLIC_ADDRESS='{"10.230.191.121": "54.158.225.35", "10.69.174.223": "54.159.119.143"}';
+   export FLOCKER_ACCEPTANCE_VOLUME_BACKEND=aws;
+   export FLOCKER_ACCEPTANCE_NUM_AGENT_NODES=2;
+   export FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH=/tmp/tmpfvb3xV;
+
+In this case you can copy and paste the ``FLOCKER_ACCEPTANCE_HOSTNAME_TO_PUBLIC_ADDRESS`` value directly into the configuration file. E.g.
```
tomprince: Perhaps a good follow-up would be to get `run-acceptance-tests --keep` to dump out an appropriate `managed` stanza?

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -204,9 +204,21 @@ Managed
...
+* re-installs the ``clusterhq-flocker-node`` and ``clusterhq-python-flocker`` packages,
+* creates and distributes new certificates and keys to all the nodes,
+* restarts the ``flocker-control`` on the first ``managed`` node, and
+* restarts the ``flocker-*-agent`` processes on all the nodes.
+
+This means that the ``managed`` provider can be used to quickly test different package versions and packages built from different branches.
+
+To use the ``managed`` provider, the configuration file should include:
+
+- **addresses**: A ``list`` of IP addresses of the nodes or a ``dict`` of ``{"<private_address>": "<public_address>"}`` if the public addresses are not configured on the node (see below).
```
tomprince: Thinking forward, I wonder how much more information we'll want to have about nodes. Eventually, perhaps we want 
```yaml
managed:
    nodes:
    - private: 10.x.x.x
      public: 8.8.x.x
      ...
    - ...
```

When I first started thinking about this comment, I was slightly concerned about overriding the meaning of `addresses` and taking over the meaning of a `dict` (and so the new configuration should be `private_address_mapping`. But on further reflection, if we are extending this much further, `nodes` seems like a better key that `addresses` anyway.

flocker/acceptance/obsolete/test_containers.py
```diff
@@ -38,6 +38,8 @@ def _create_container(self, cluster, script):
             }, script)
 
         def check_result(response):
+            # FLOC-3227 With the managed runner, this public_address is the
+            # private address.
```
tomprince: This now seems obsolete?

admin/acceptance.py
```diff
@@ -205,10 +206,27 @@ class ManagedRunner(object):
...
-            for address in node_addresses
-        )
+        """
+        :param list or dict node_addresses: A ``list`` of public IP addresses
+            or a ``dict`` of ``private_address`` -> ``public_address``.
+
+        See ``ManagedRunner`` and ``ManagedNode`` for other parameter
+        documentation.
+        """
+        if isinstance(node_addresses, Mapping):
```
tomprince: `Mapping`?

tomprince:
This looks good, but I haven't tested that it works. Please merge after addressing the above points to your satisfaction, and buildbot is happy, and you have verified that the changes work as desired.

docs/gettinginvolved/acceptance-testing.rst
wallrj: Yep, good idea, but we're not creating and linking to follow up issues any more, right?

admin/acceptance.py
wallrj: I was trying to be clever. Now using `dict` instead.

flocker/acceptance/obsolete/test_containers.py
wallrj: Removed.

docs/gettinginvolved/acceptance-testing.rst
wallrj: Fixed.

docs/gettinginvolved/acceptance-testing.rst
wallrj: I thought about  forcing addresses to be a dict in all cases and just documenting that the hostname and private address should be same in the case of Rackspace....but that seemed ugly.

`{"1.2.3.4": "1.2.3.4"}`

I have a slight concern about ordering when using the dictionary form. With the list form, the first listed node is always the control node. With dicts that's not the case.

docs/gettinginvolved/acceptance-testing.rst
tomprince: I thought we are still leaving comments about possible follow-ups, even if there isn't an associated ticket though

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -204,9 +204,21 @@ Managed
...
 
-- **addresses**: The IP addresses of two running nodes.
+This is a quicker way to run the acceptance tests as it avoids the slow process of provisioning new acceptance testing nodes.
+
+The ``managed`` provider:
+
+* re-installs the ``clusterhq-flocker-node`` and ``clusterhq-python-flocker`` packages,
+* creates and distributes new certificates and keys to all the nodes,
+* restarts the ``flocker-control`` on the first ``managed`` node, and
+* restarts the ``flocker-*-agent`` processes on all the nodes.
```
exarkun: Are these details important?  They'll have to be kept up-to-date with the code if they're not to become incorrect details.

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -223,6 +235,70 @@ The nodes should be configured to allow key based SSH connections as user ``root
 
    admin/run-acceptance-tests --distribution centos-7 --provider managed --config-file config.yml
 
+If you are using the ``managed`` provider with ``AWS`` nodes, you  will need to supply both the private and public IP addresses for each node.
+AWS nodes do not have public IP addresses configured in the operating system; instead Amazon routes public IP traffic using NAT.
+In this case the acceptance tests need a hint in order to map the private IP address reported by the Flocker ``/state/nodes`` API to the public node address.
```
exarkun: Maybe (next, not instead of this) we should teach Flocker on AWS to discover and report public addresses on its own.  We already kinda know most clients of Flocker are going to have to do this public/private mapping thing - and this is even the second time we're doing it ourselves.  Are we going to have to repeat this for, say, the benchmarking tool?

docs/gettinginvolved/acceptance-testing.rst
wallrj: I removed the details.

docs/gettinginvolved/acceptance-testing.rst
wallrj: I'm changing it to a list of lists which ensures that the first node is always the control node and also simplifies the code a little.

docs/gettinginvolved/acceptance-testing.rst
wallrj: There's 
 * https://clusterhq.atlassian.net/browse/FLOC-2137 and
 * https://github.com/ClusterHQ/flocker/compare/public-ips-FLOC-2137 

...where @tomprince introduced a new node hostname item to `agent.yml`.

PR #2040, opened 2015-10-08
[FLOC-3124] Enforce bytes on public address returned from libcloud

dwgebler:
Fixes FLOC-3124. On PyPy, sometimes the resulting IP addresses returned from libcloud's rackspace driver (`wait_until_running`) are `unicode` instead of bytes. Our `run_remotely` helper uses a Pyrsistent record type that accepts `bytes`, therefore if the result from libcloud is unicode, encode it to ASCII.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2040)

### DIFF
--- a/flocker/provision/_libcloud.py
+++ b/flocker/provision/_libcloud.py
@@ -240,6 +240,9 @@ class LibcloudProvisioner(object):
         else:
             private_address = None
 
+        if isinstance(private_address, unicode):
+            private_address = private_address.encode("ascii")
+
         return LibcloudNode(
             provisioner=self,
             node=node, address=public_address,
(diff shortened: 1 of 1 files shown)


exarkun:
This seems like it might be a bug in libcloud that's worth reporting.

itamarst:
Given the randomness involved, quite possibly the private address might have same issue. So may as well fix those too. Do that and merge. Thanks!

jongiddy:
I'm running this on a PyPy-enabled Buildbot. If it works, then the Rackspace acceptance tests should (and the AWS acceptance tests might) pass, assuming nothing else goes wrong.

See http://52.89.99.185/boxes-flocker?branch=pypy-pyrsistent-error-FLOC-3124

PR #2028, opened 2015-10-06
[FLOC-2770] Check release packages

jongiddy:
Note, this is a resubmission of FLOC-2770, which was previously merged (#1933) and then reverted (#1973), due to regression of the (undocumented) ability to provide an empty string for the `--flocker-version` flag in order to test the latest version of a branch.  This was necessary for both client tests and acceptance tests because no `--flocker-version` flag always tried to use the latest installable version, which only worked for non-branch builds.

The code has been updated so that either `--flocker-version=''` or the absence of the `--flocker-version` flag cause the latest version of a branch to be tested.

The remainder of this comment is copied from PR #1933:

Currently, we do not check that the artifacts published during release are correct. This has lead to problems only observed when not using Buildbot repos, hence usually undetected by any of our automated tests. (e.g. https://clusterhq.atlassian.net/browse/FLOC-2904 and https://clusterhq.atlassian.net/browse/FLOC-3008).

This PR adds an additional verification step to the release process, to check that the client packages can be installed in a Docker container OK. Many of the issues that could affect our packaging (S3 permissions, errors in packaging format) should be obvious from these tests.  Although this test only detects problems after a release, at least we're ahead of our users in awareness of problems with the release, as opposed to waiting for them to complain.

The test script is created as a Python unittest, allowing it to perform all tests and summarise the results.  Error outputs are saved to a file for later perusal.

Tests for the client now not only print the version number, but also check it against the expected version number.  Since this is inappropriate for our documentation, the client tests have been removed as task directives, and replaced in the docs.

To test the new code, run `admin/test-artifacts`.

~~At this moment, the CentOS RPM tests fails, due to FLOC-3008.  This is useful to demonstrate the error output.~~ Since FLOC-3008 is resolved, merging this branch with `master` eliminates this problem. [The branch has now had master and hence FLOC-3008 merged in].

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2028)

PR #2041, opened 2015-10-08
[FLOC-3233] Make os_version a generated property of PackageSource

jongiddy:
PR #2028 introduces a third location where `os_version` is generated from `version` and then both values are used to create a `PackageSource` object.

This PR makes `os_version` a property of `PackageSource`, allowing it to be generated on demand in one place, rather than by each piece of code that creates a `PackageSource` instance.

The code that uses `os_version` has been modified to call it once, to prevent frequent re-generation of the value.

Note, this changes the two existing locations where `PackageSource` instances are created.  The latter merge of this PR and #2028 should be updated to modify the third `PackageSource` creation.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2041)

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -29,7 +29,6 @@ from effect import parallel
 from effect.twisted import perform
 
 from admin.vagrant import vagrant_version
-from flocker.common.version import make_rpm_version
 from flocker.provision import PackageSource, Variants, CLOUD_PROVIDERS
 import flocker
 from flocker.provision._ssh import (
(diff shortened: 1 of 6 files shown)


tomprince:
> The code that uses os_version has been modified to call it once, to prevent frequent re-generation of the value.

If we actually care about the cost (which I'm not sure we do), then it probably makes sense to turn it into a method, rather than a property. Otherwise, I wouldn't bother changing the logic around its use.

There should probably be a couple of tests for the new property/method.

Address those two points, then merge once buildbot is happy.

PR #2042, opened 2015-10-08
[FLOC-3182] Only attempt to configure the firewall on operating systems that have a firewall installed

wallrj:
Fixes: 
 * https://clusterhq.atlassian.net/browse/FLOC-3182
 * https://clusterhq.atlassian.net/browse/FLOC-1550

Currently if you attempt to use the master version of ``run-acceptance-tests --provider=managed `` with Rackspace nodes, the tests fail because the firewall isn't reconfigured after re-installing Flocker so the agents can never connect to the control service.

We need to configure the firewall after re-installing Flocker in the managed runner, but AWS Centos and Ubuntu images don't have a firewall installed.
We don't want to have to specify the cloud in the configuration file.

So in this branch I:
 * Add a test for the presence of a firewall command line interface before attempting to configure the firewall on each node. Which means that I can,
 * move the firewall configuration step in to the ``configure_cluster`` function which is called by all acceptance test providers. And by configuring the firewall there I can, 
 * only run the `flocker-control`` related firewall rules on the control node, and
 * run the Docker API related firewall configuration steps on *all* nodes, which has the added benefit of
 * hiding those acceptance-test-specific configuration steps from our docs.

If you run `iptables-save  | grep 452` on nodes created from this branch you should only see the rules on the control node.
And if you run `iptables-save  | grep 2376` you should see rules on every node.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2042)


flocker/provision/_install.py
```diff
@@ -800,6 +802,57 @@ def task_open_control_firewall(distribution):
...
+        firewall_command = b'firewall-cmd'
+    elif distribution == 'ubuntu-14.04':
+        firewall_command = b'ufw'
+    else:
+        raise DistributionNotSupported(distribution=distribution)
+
+    # Only run the commands if the firewall command is available.
+    return run_from_args([b'which', firewall_command]).on(
+        success=lambda result: commands,
+        # XXX: Is this the exit code of the SSH command or the remote command?
```
jml: :grey_question: This seems like a relatively easy thing to find out by experimentation. I'll leave it up to you whether it's worth doing so or not before landing.

flocker/provision/_install.py
```diff
@@ -800,6 +802,57 @@ def task_open_control_firewall(distribution):
...
+    else:
+        raise DistributionNotSupported(distribution=distribution)
+    return sequence([
+        open_firewall(service)
+        for service in ['flocker-control-api', 'flocker-control-agent']
+    ])
+
+
+def catch_exit_code(expected_exit_code):
+    """
```
jml: :grey_question: I was going to ask for a sentence here describing what the function does, but reading the rest of the docstring it seems redundant. No action required.

jml:
Thanks Richard! PR looks good to me. No changes required.

flocker/provision/_install.py
tomprince: It isn't, but I don't think it would be difficult to make it so.

wallrj:
```
man ssh
...
EXIT STATUS
     ssh exits with the exit status of the remote command or with 255 if an error occurred.
```

wallrj: But this is conch, so it'll be more like:
```
(3182)(reopen-firewall-FLOC-3182-2 ? ▴1)[~/.../HybridLogic/flocker]$ python ~/Downloads/echoclient_ssh.py -h 119.9.108.42 -u centos
[27] > /home/richard/Downloads/echoclient_ssh.py(58)connectionLost()
-> self.finished.callback(None)
(Pdb++) reason.value
ProcessTerminated('process ended with exit code 2',)
(Pdb++) reason.value.exitCode
2

```

PR #2043, opened 2015-10-08
[FLOC-3173] Document eliot/journald support

itamarst:
1. Install eliot-tree package.
2. Include eliot-tree and eliot-prettyprint as public binaries installed by the shared Python package so that they are always available.
3. Document how to use journalctl and eliot to extract useful information from Flocker logs.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2043)

### DIFF
--- a/docs/control/administering/debugging.rst
+++ b/docs/control/administering/debugging.rst
@@ -9,8 +9,11 @@ Debugging
 Logging
 -------
 
-Flocker processes use `eliot`_ for logging.
-These logs can be rendered as an ASCII tree using `eliottree`_.
+Flocker processes use the `Eliot`_ framework for logging.
+Eliot structures logs as a tree of actions, which means given an error you can see what Flocker actions caused the errors by finding the other messages in the tree.
+The tree of actions can also span processes; thus you can trace API calls from within the Docker plugin and see the effects in the control service logs.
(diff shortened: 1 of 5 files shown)


docs/control/administering/debugging.rst
```diff
@@ -9,8 +9,11 @@ Debugging
 Logging
 -------
 
-Flocker processes use `eliot`_ for logging.
-These logs can be rendered as an ASCII tree using `eliottree`_.
+Flocker processes use the `Eliot`_ framework for logging.
+Eliot structures logs as a tree of actions, which means given an error you can see what Flocker actions caused the errors by finding the other messages in the tree.
+The tree of actions can also span processes; thus you can trace API calls from within the Docker plugin and see the effects in the control service logs.
+Messages can be rendered into a human-readable tree using the `eliot-tree`_ command-line tool, which is pre-installed with Flocker.
```
davidoliverSP2: We're pretty inconsistent when we use "command-line" and "command line" throughout the docs - I think we should be using it without the dash, but happy to pick this up myself in FLOC 3237.

itamarst: I can drop the dash, just tell me which to do :grinning:

davidoliverSP2: cool - go for it, cheers!

I'm (somewhat slowly) building a style guide for this kind of stuff. I'll add this to it.

jml:
Lovely, thanks.

PR #2044, opened 2015-10-08
[FLOC-3185] Reset reconnection delay upon succesful connection

itamarst:
Before this the reconnect delay would grow on every reconnect and never go back down.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2044)


jongiddy:
LGTM. Please merge when green.

PR #2045, opened 2015-10-08
[FLOC-2907] Human-readable volume names

itamarst:
When possible we want UIs displaying information about volumes we've created to include something a human can easily use to correlate with our logs. I.e. we want a user to be able to say "oh that's dataset XYZ and I see that detached earlier that's what's going on". Right now there's a whole bunch of extra work trying to correlate volumes with Flocker datasets.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2045)


sarum90:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[flocker/node/agents/functional/test_ebs.py, line 116 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2045#-K08xAoY7UWM8KBBDru7)** ([raw file](https://github.com/clusterhq/flocker/blob/c0d6f32a5bdfc8af15e034cd97f344b4a41f0ceb/flocker/node/agents/functional/test_ebs.py#L116)):</sup>
I'm sure this is covered somewhere, but I'm just curious: Presumably this is run against real AWS since we don't have a full fake aws implementation (AFAIK). How do we keep this from just adding volumes and not deleting them? Presumably that could eventually be expensive if we just keep creating volumes and never remove them. Same with the Cinder implementation.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2045)

itamarst:
If you follow the code there's a `flocker.node.agents.test.blockdevicefactory.get_blockdeviceapi_with_cleanup` that is used to construct IBlockDeviceAPI used by tests.

exarkun:
Thanks.

Build failures look like known issues (but a JIRA search doesn't turn up an issue for the `DatasetExists` failures ...  File one, please).

I don't think the issue this PR fixes is defined very well.  A user story giving some sense of how this change actually helps users would probably be nice.  It's not clear how presenting the information in this way makes the user's life easier - or how the user will learn about this feature in order to benefit from it.

In the future I'd like to see more thought (reflected somewhere in our trail of artifacts) put into the user experience for changes that are primarily user-experience-oriented.

Last thought - there's no upgrade code here.  Anyone who already has volumes won't benefit from this and once they upgrade Flocker, they'll have a mix of annotated and un-annotated volumes.  That seems undesirable to me.  If the feature is useful, we should provide it to everyone?  At the very minimum, warning existing users of the inconsistency that will result seems necessary.

Address those points and then merge.

itamarst:
Sorry, user story was somewhat implicit, yeah. Goal is "I can easily identify volumes as belonging to Flocker and to which dataset they correspond, by glancing at EC2 web console, Cinder command line, etc." So user doesn't have to do anything special, they just use thier normal tools and life is easier.

Upgrading... yeah, maybe. On the one hand maybe not worth the time since user base is so small and this is not a semantic change, on the other hand would help current user who is main motivation (for easier debugging).

itamarst:
I'm going to say that upgrade isn't worth it; better to have functionality for new users *the vast majority) and not spend time on upgrading something that is merely a nice to have.  I will document the new feature and its impact on existing volumes, though.

itamarst:
OK, mergeable as soon as buildbot is green.

PR #2047, opened 2015-10-09
[FLOC-3044] OS X client test fixes

jongiddy:
Interacting with Boot2Docker / Docker-machine on OS X has a few foibles compared to Linux, due to the use of Virtual Machine.  The changes in this PR should fix the client tests for consistent behaviour between platforms.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2047)

### DIFF
--- a/admin/client.py
+++ b/admin/client.py
@@ -121,9 +121,6 @@ class DockerRunner:
         # http://doc-dev.clusterhq.com/gettinginvolved/client-testing.html
         # for details.
         params = docker.utils.kwargs_from_env(assert_hostname=False)
-        tls_config = params.get('tls')
-        if tls_config:
-            tls_config.verify = False
         self.docker = docker.Client(version='1.16', **params)
         self.image = image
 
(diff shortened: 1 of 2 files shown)


adamtheturtle:
Thanks @jongiddy this works for me so merge when the build is green.

admin/client.py
```diff
@@ -116,14 +116,25 @@ class DockerRunner:
     """
 
     def __init__(self, image):
-        self.docker = docker.Client(version='1.18')
+        # Getting Docker to work correctly on any client platform can
+        # be tricky.  See
+        # http://doc-dev.clusterhq.com/gettinginvolved/client-testing.html
+        # for details.
+        params = docker.utils.kwargs_from_env(assert_hostname=False)
```
adamtheturtle: `assert_hostname=False` is not necessary, please remove it.

adamtheturtle: or it seems the `tls` stuff. It does need one of the two changes but seemingly not both.

jongiddy: OK, `verify=False` is usually a bad thing to do, so I will remove that, and keep the `assert_hostname` change.

PR #2048, opened 2015-10-09
[FLOC-2977] Docker plugin tests

dwgebler:
Hopefully this fixes FLOC-2977; if we get a read timeout from docker-py when awaiting a response from the Docker daemon, let's just continuously retry starting the container again, since making multiple requests in this case won't hurt.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2048)

### DIFF
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -4,6 +4,8 @@
 Tests for the Flocker Docker plugin.
 """
 
+from requests.exceptions import ReadTimeout
+
 from twisted.internet import reactor
 from twisted.trial.unittest import TestCase
 
@@ -94,10 +96,19 @@ class DockerPluginTests(TestCase):
             ["python", "-c", script.getContent()] + list(script_arguments),
             volume_driver="flocker", **docker_arguments)
         cid = container["Id"]
-        client.start(container=cid)
         if cleanup:
             self.addCleanup(client.remove_container, cid, force=True)
-        return cid
+
+        def start_python_container():
+            try:
+                client.start(container=cid)
+                return True
+            except ReadTimeout:
+                return False
+
+        d = loop_until(start_python_container)
+        d.addCallback(lambda _: cid)
+        return d
 
     @require_cluster(1)
     def test_volume_persists_restart(self, cluster):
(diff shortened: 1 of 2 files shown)


itamarst:
I don't quite understand why this would help lacking any understanding of why it's happening. If a request takes more than 100 seconds to get a response, which is what we're seeing, why would retrying fix it? Why do we not get this error in other docker interactions?

The image is supposed to be pre-pulled by acceptance tests, so it's not that we're waiting for docker pull... or perhaps it isn't? Worth checking. Do Docker logs (included in remote_logs in buildbot UI) indicate any reason for lack of response?

itamarst:
This is the remote logs as linked from build: http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fcentos-7%2Faws/builds/2913/steps/trial/logs/remote_logs.log

If this only happens during releases might also be worth trying to figure out how that differs from regular builds.

dwgebler:
At first I thought it might be a (unknown) bug in Docker that meant the daemon sometimes hung on a request and failed to send a response and that it could be fixed by retrying the request as needed. But further testing is indicating that it's probably hanging because of waiting on the Flocker plugin to return its response, which is a known Docker API behaviour with plugins. So on further investigation, this change is unlikely to solve the issue and I'm still working on it.

itamarst:
Oh boy. I remember Luke talking about something like that, yeah.

sarum90:
I successfully reproduced hanging the docker daemon with the labs implementation of the flocker docker plugin just by skipping step 7 on https://github.com/ClusterHQ/volume-plugins-demo. Looking through logs it looked like docker was waiting for the flocker plugin before it would respond any other api call (at least for docker ps to respond), and the flocker plugin was endlessly retrying a docker API call presumably because it needed the result in order to unblock.

I figured it was just a limitation of the labs implementation, especially given commits like https://github.com/ClusterHQ/flocker/commit/d9eb76e7fc9293f4c6a55def4f63e75807ee6434 in the new implementation. I can dig out the logs pretty quickly if they would be of any help here.

Sort of doubt it will be of any help since it was the labs implementation and it sounds like you might have a good testing repro, but it might give insight into how docker interacts with the plugins.

dwgebler:
Closing this as I have another branch for the ticket now.

PR #2049, opened 2015-10-09
[FLOC-3226] Get rid of intersphinx so we don't get false positives on doc builds.

itamarst:
Fallback to intersphinx has caused builds that should've broken to pass, so get rid of it.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2049)

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -30,7 +30,6 @@ on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
 # coming with Sphinx (named 'sphinx.ext.*') or your custom ones.
 extensions = [
     'sphinx.ext.extlinks',
-    'sphinx.ext.intersphinx',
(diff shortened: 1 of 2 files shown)


adamtheturtle:
Sorry I didn't consider this before. Surely this will have the same problem as intersphinx - if we change the docs so that these links break, the links will pass in the docs builder and then if we merge master will pass and then the next build of master. I think that intersphinx is neater and that this problem is not avoidable (maybe with some very advanced checking but let's ignore that). I'd like to close the issue without any change. Do you agree?

itamarst:
My impression from discussion in bug was that intersphinx impacts *all* `ref`? In which case it seems too dangerous period? Maybe I'm misunderstanding.

adamtheturtle:
Ah yes! Friday mistake. Merging because the docs builder is green.

PR #2053, opened 2015-10-20
[FLOC-3256] Fix typo in `_send_state_to_connections`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3256

I think this might be causing a bunch of our intermittent failures.

I'm not sure what the best way to test this is, though.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2053)

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -560,7 +560,7 @@ class ControlAMPService(Service):
                 for connection in elided_update:
                     AGENT_UPDATE_ELIDED(agent=connection).write()
 
-                for conection in delayed_update:
+                for connection in delayed_update:
                     self._delayed_update_connection(connection)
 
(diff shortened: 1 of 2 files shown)


sarum90:
Awesome find!

Sad panda :panda_face: about python letting iterating variables escape scope. I'm 90% certain that we will not be able to figure out why we added the confounding agent in the future when we come back to this code without looking at the git history, but your comment in concise and correct, and I certainly can't think of anything better that proves your fix.


---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2053)

tomprince:
Merging with failures, since this will hopefully address some intermittent causes of them.

PR #2057, opened 2015-10-21
Reduce API list_volume in DestroyBlockDeviceDataset

sarum90:
Now that we have the result of list_volumes() being passed from discover_state to calculate_changes, it is possible to use that information to map dataset_ids to blockdevice_ids prior to the construction of the IStateChanges.

This is helpful for the IStateChanges that start out by calling list_volumes just to map from dataset_id to blockdevice_id. This PR adds a blockdevice_id field to all of the IStateChanges that are used in DestroyBlockDeviceDataset, and removes many of the calls to list_volumes in that state change. The goal of this is to reduce over all API calls in the IStateChanges.

During development I added a hacky test to verify that list_volume calls were decreasing, but the test seemed brittle, so I have removed it for merging.

While I was in the code I did a drive by fix to FLOC-1818 (making DestroyVolume async), which was trivial given current infrastructure.
    
I also removed comment referencing  FLOC-1756 was already marked as 'Done' and certainly the comment was no longer accurate.
    
In addition I removed comment referencing FLOC-1616, which was also already marked as 'Done', and the comment was no longer accurate.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2057)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -214,7 +214,7 @@ UNMOUNT_BLOCK_DEVICE = ActionType(
 
 UNMOUNT_BLOCK_DEVICE_DETAILS = MessageType(
     u"agent:blockdevice:unmount:details",
-    [VOLUME, BLOCK_DEVICE_PATH],
+    [BLOCK_DEVICE_ID, BLOCK_DEVICE_PATH],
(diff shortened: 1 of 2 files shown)


jml:
:+1: Lovely, thanks.

PR #2058, opened 2015-10-22
[FLOC-3238] Fix argument type for Attached unexpected device exception

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3238

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2058)


wallrj:
I forced a build http://build.clusterhq.com/boxes-flocker?branch=AttachedUnexpectedDevice_FLOC-3238

wallrj:
Thanks @myechuri 

I don't think it makes sense to return `/dev/null` from `_next_device` and `_wait_for_new_device`.

And I don't think it'd make sense to see 

```
AttachedUnexpectedDevice(
    requested="/dev/null",
    discovered="/dev/sdc",
)

AttachedUnexpectedDevice(
    requested="/dev/sdc",
    discovered="/dev/null",
)
```

`None` is a more natural return value in both those functions. 

 * `_next_device` > `None` > "All the devices on this node have been used. No more volumes can be attached here."
* `_wait_for_new_device` > `None` > "A new device did not appear in the OS within X seconds"

`/dev/null` doesn't make sense in either of those cases.

And neither of those situations should lead to `AttachedUnexpectedDevice`, right?

Perhaps when we requested the attachment there was a free device, but by the time the attach takes place, that device has been taken. 

If we try and attach a volume and its device doesn't appear within a time limit we should probably try and clean up by making an API request to detach it again. 

What do you think?

Perhaps we

flocker/node/agents/test/test_ebs.py
```diff
@@ -4,10 +4,32 @@
 Tests for ``flocker.node.agents.ebs``.
 """
 
+from string import lowercase
+from uuid import uuid4
+
+from hypothesis import given
```
sarum90: :smiley_cat: I love reviewing hypothesis PRs.

flocker/node/agents/test/test_ebs.py
```diff
@@ -4,10 +4,32 @@
...
+    AttachedUnexpectedDevice, _expected_device,
+    _attach_volume_and_wait_for_device, _get_blockdevices,
+    _get_device_size,
+)
+from ..blockdevice import BlockDeviceVolume
+
+from ....testtools import CustomException
+
+
+# A Hypothesis strategy for generating /dev/sd?
```
sarum90: :grey_question: For bonus points you could also make this generate /dev/sd?? (e.g. /dev/sdaa) or /dev/sd??? I believe thats where the kernel puts >26 SCSI disks. You can't attach that many with ebs, but I'm guessing it'd be tempting to just copy this for other drivers, and richardw@ was working on a system the other day that was way off into the /dev/sdbx range with attached devices.

flocker/node/agents/ebs.py
```diff
@@ -309,6 +311,7 @@ def _enable_boto_logging():
 _enable_boto_logging()
 
 
+@with_cmp(["requested", "discovered"])
```
sarum90: :grey_question: Is it important that comparison works here (just curious, not suggesting any change)?

Also, sad that you couldn't use more of the characteristic library, looks like you basically had to roll your own type checking and __repr__ for other reasons :(.

flocker/node/agents/ebs.py
```diff
@@ -683,6 +701,87 @@ def _is_cluster_volume(cluster_id, ebs_volume):
...
+):
+    """
+    Attempt to attach an EBS volume to an EC2 instance and wait for the
+    corresponding OS device to become available.
+
+    :param BlockDeviceVolume volume: The Flocker representation of the volume
+        to attach.
+    :param unicode attach_to: The EC2 instance id to which to attach it.
+    :param attach_volume: A function like ``EC2Connection.attach_volume``.
+    :param attach_volume: A function like ``EC2Connection.detach_volume``.
```
sarum90: s/attach_volume/detach_volume/g

flocker/node/agents/ebs.py
```diff
@@ -683,6 +701,87 @@ def _is_cluster_volume(cluster_id, ebs_volume):
...
+    :param list blockdevices: The OS device paths which are already present on
+        the system before this operation is attempted (primarily useful to make
+        testing easier).
+
+    :raise: Anything ``attach_volume`` can raise.  Or
+        ``AttachedUnexpectedDevice`` if the volume appears to become attached
+        to the wrong OS device file.
+
+    :return: ``True`` if the volume is attached and accessible via the expected
+        OS device file.  ``False`` if the attempt times out without succeeding.
```
sarum90: Hrm... seems more like, "``False`` if attach_volume fails with an InvalidParameterValue and attach_attempt was less than MAX_ATTACH_RETRIES."

This seems like a confusing interface to me. I think I would rather have the:
```
+        if e.code == u'InvalidParameterValue':
+            if attach_attempt == MAX_ATTACH_RETRIES:
+                raise
+            return False
+        else:
+            raise
```
block 1 layer up in the stack, or have attach_attempt be a boolean instead of a count, and be named something like return_false_instead_of_raise_invalid_parameter_value - or something more manageable.  To me those semantics indicate "this logic belongs a layer up".

flocker/node/agents/test/test_ebs.py
```diff
@@ -31,6 +53,133 @@ def test_repr(self):
...
+            raise CustomException("Fake failure generated by test")
+
+        def detach_volume(*a, **kw):
+            pass
+
+        self.assertRaises(
+            CustomException,
+            _attach_volume_and_wait_for_device,
+            self.volume, self.compute_id, attach_volume, detach_volume,
+            device,
```
sarum90: slight preference for using kwargs here rather than positional (as you do for blockdevices) I had to go back to the definition a lot to determine what each of these arguments were.

flocker/node/agents/ebs.py
sarum90: Looks like the retry isn't tested at this level anyway, so shouldn't be too much work to put it up a layer.

sarum90:
Request a tiny refactor, but in general this looks great! And an excuse to learn another cool python library for me (I hadn't reviewed things that used characteristic before).

Address and merge.

flocker/node/agents/ebs.py
exarkun: > Is it important that comparison works here (just curious, not suggesting any change)?

Yea.  One of the unit tests wants to make sure an exception with certain attribute values gets raised and the nicest spelling of that in the tests is with `assertEqual`.

exarkun: > Also, sad that you couldn't use more of the characteristic library, looks like you basically had to roll your own type checking and __repr__ for other reasons :(.

Indeed.  Also sad is that I couldn't use `pyrsistent` here.  That would done the type checking for me, at least (probably not the `__repr__` bit though).  Unfortunately it seems that `PClass` and `Exception` are incompatible base classes so I guess it's not possible to make any exception use `pyrsistent`.

flocker/node/agents/test/test_ebs.py
exarkun: Several of these will look like `attach_volume=attach_volume`.  I'm guessing that doesn't tell you anything more than `attach_volume` - except that you don't know there's nothing more to know until you know the name of the parameter, which still involves looking back at the signature?

I guess the slight redundancy is worth it for the increased readability (in the "can I understand this easily" sense rather than the "does this read nicely" sense).

flocker/node/agents/ebs.py
exarkun: Hmm.  Mixed feelings about this.  On one hand, yes this code is gross and mixes some layers.  I did intentionally keep the exception handling down here because I don't want to introduce new exception handling around the `detach_volume` call which is what happens if I do the obvious refactoring to move this up.

While thinking about this I did implement something that introduces a couple new exception types for the control flow necessary to put the retry logic up a layer but make sure it still only applies to attach failures.  At the end of that exercise I realized the reason I did the refactoring at all was to avoid changing logic in a function I couldn't unit test - which is exactly what the next layer up is.

So I'm inclined to add tests for the retry logic where it is now.  The downside is that the layering remains kind of gross.  The upside is I can have unit tests for it.

exarkun: Ah, on the gripping, maybe the existing counter logic led me off track and there's a different factoring that gets the best of both worlds.

exarkun:
Only known, tracked test failures in the build.  Merging now.

PR #2059, opened 2015-10-22
[FLOC-2137] Allow the public IP address of the node to be statically configured

wallrj:
Design / Implementation for https://clusterhq.atlassian.net/browse/FLOC-2137
- Allow an optional `hostname` value in `agent.yml`
- Use that hostname when reporting NodeState.
- Fall back to the existing `socket` probing technique.
- Have run-acceptance-tests set the `hostname` configuration option when configuring agents on acceptance nodes.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2059)


docs/config/configuring-nodes-storage.rst
```diff
@@ -31,12 +31,15 @@ The optional ``port`` variable is the port on the control node to connect to.
 This value must agree with the configuration for the control service telling it on what port to listen.
 Omit the ``port`` from both configurations and the services will automatically agree.
 
+If node has a public address that is different than its local address (for example, on an EC2 instance),
```
itamarst: This isn't strictly necessary though, right? We after all manage without right now. And it adds an extra burden of configuration on users that they don't necessarily need. Maybe indicate it's useful but not required? Or the caveats of not doing it?

flocker/node/script.py
```diff
@@ -312,7 +312,7 @@ def get_service(self, reactor, options):
         configuration = get_configuration(options)
         host = configuration['control-service']['hostname']
         port = configuration['control-service']['port']
-        ip = self.get_external_ip(host, port)
+        ip = configuration.get('hostname', self.get_external_ip(host, port))
```
itamarst: I suggest only running get_external_ip if configuration doesn't have an entry, given it adds noise to control service logs.

flocker/node/test/test_script.py
```diff
@@ -226,6 +230,21 @@ def test_initialized(self):
...
+    def test_initialized_hostname(self):
+        host = b"192.0.2.13"
+        port = 2314
+        name = u"from_config-test"
+
+        setup_config(self, control_address=host, control_port=port, name=name)
+        options = DatasetAgentOptions()
+        options.parseOptions([b"--agent-config", self.config.path])
+        config = get_configuration(options)
+        config['hostname'] = 'hostname.example'
```
itamarst: So.... everywhere else we use IPs. This changes means we're OK with DNS hostnames appearing in e.g. REST API output? That might violate our schema.

docs/config/configuring-nodes-storage.rst
itamarst: One thing we need to keep in mind: bandwidth costs. Data sent within AWS private network is free, data sent over external IPs costs money. I would hope that DNS lookups for AWS hosts *within* AWS get turned into the private IP, in which case documenting we want explicit hostname (as opposed to "address" which implies IP) is important. If hostnames always resolve to external IPs then it's possible that following this advice may cost our users a bunch of money on AWS, depending how AWS does bandwidth accounting...

itamarst:
I am happy with the implementation mostly, other than some comments here and there and whatever more detailed code review would come up with. What worries me is change from IP to hostnames (are there consequences we haven't thought of beyond REST API?) and impact on users re bandwidth costs. So please address only those two points and resubmit.

sarum90:
Just FYI, I just killed a buildbot builder for this branch that had been running for 205 hrs (http://build.clusterhq.com/builders/flocker-centos-7/builds/7521). Hopefully it didn't need more time :stuck_out_tongue:. Probably worth investigating as part of review.

wallrj:
Thanks @itamar for the review.

I put this up for review because @exarkun and I thought we might need it for our benchmarking tool on AWS, but I'm working on something else now.

I'll put the issue into the backlog and @jongiddy may decide to pick this up while working on the benchmarking project.

PR #2061, opened 2015-10-22
[FLOC-3277] Add ``lshw`` output to the diagnostics bundle

wallrj:
Design for https://clusterhq.atlassian.net/browse/FLOC-3277

* A thin wrapper around `lshw` which we can call from `flocker-diagnostics` and `flocker-benchmark`
* A new benchmark module and `flocker-benchmark` script.
* Made `flocker.node.diagnostics` public so that it's functions can be shared.

`flocker-benchmark hardware-report` will be invoked via SSH from https://github.com/ClusterHQ/flocker/tree/benchmark-prototype 

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2061)


itamarst:
Neat, never heard of `lshw` before.

itamarst:
Please proceed.

jongiddy:
Putting into code review. Now filters the information for benchmarking to processor and memory hardware.

Renamed `lshw` function to generic `list_hardware` in case we need to implement it for non-Linux or using a different command later.

wallrj:
Looks pretty good to me @jongiddy .

1. The flocker-centos-7 and flocker%2Funit-test%2Fcentos-7%2Fzfs-head are failing because those build slaves haven't `lshw`. Add a `require_lshw` skip decorator (or something) to those tests and create a followup issue to install `lshw` on those slaves ...or jenkins.
2. Looks like the output doesn't include block device info or full ip address info so we still need all the other diagnostic commands...fine.

Answer or address the points above then merge when green unless you want another review.

jongiddy:
1. Added unittest skip. Couldn't see similar errors on Jenkins, and I believe we're not making any changes to Buildbot with pending decommissioning.
2. Added network, disk and volume classes, which should provide that info.

PR #2062, opened 2015-10-22
[FLOC-2819] Docker 1.9 plugin compatibility

dwgebler:
Fixes FLOC-2819. Looking at https://github.com/docker/docker/blob/master/docs/extend/plugins_volume.md versus https://docs.docker.com/extend/plugins_volume/ it seems that the only difference in how the plugin API needs to work is that `/VolumeDriver.Create` now takes an `Opts` mapping of custom options passed through from Docker, e.g. as a result of `docker volume create --driver flocker --opt foo=bar --opt maxsize=10G`

It is unclear whether this `Opts` map is optional or not, I am inclined to suspect Docker will make the POST request to our plugin with `Opts` as an empty mapping `{}` if no `opt` flags were specified through the CLI/API on Docker's end - we also want compatibility with our existing previous Docker versions, so this design makes `Opts` on our endpoint optional with a default of `None`. We ignore `Opts` at this stage anyway.

A new test has been written requiring Docker >=1.9.0 that exercises the case that `Opts` can be passed to `VolumeDriver.Create` and that a dataset is still created

A new end-to-end acceptance test also exercises that we are able to successfully create and run a container with a volume provisioned via the Flocker plugin.

NOTE: We do not have CI for Docker 1.9, this test will be skipped on Buildbot. I am raising a separate ticket to address the CI side.

I have decided not to explicitly test passing `Opts` through `docker-py`, as the test we have sufficiently exercises that we have the basic Docker 1.9 compatibility asked for by the JIRA issue - and `docker-py` doesn't have a way of passing an options dictionary at this stage anyway, plus we have a unit test establishing the plugin code will safely accept and ignore `Opts`.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2062)

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -13,7 +13,7 @@ You can learn more about where we might be going with future releases by:
 Next Release
 ============
 
-* The :ref:`Flocker plugin for Docker<docker-plugin>` is now compatible with Docker 1.9
+* The :ref:`Flocker plugin for Docker<docker-plugin>` is now compatible with Docker 1.9.
 * New EBS and OpenStack Cinder volumes created by Flocker will now have ``flocker-<dataset ID>`` as their name, to make it easier to find them in their respective cloud administration UIs.
   Existing volumes created by older versions of Flocker will continue to have no name.
 
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -13,6 +13,7 @@ You can learn more about where we might be going with future releases by:
 Next Release
 ============
 
+* The :ref:`Flocker plugin for Docker<docker-plugin>` is now compatible with Docker 1.9
 * New EBS and OpenStack Cinder volumes created by Flocker will now have ``flocker-<dataset ID>`` as their name, to make it easier to find them in their respective cloud administration UIs.
   Existing volumes created by older versions of Flocker will continue to have no name.
 
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -29,6 +29,9 @@ class DockerPluginTests(TestCase):
     Tests for the Docker plugin.
     """
     def require_docker(self, required_version, cluster):
+        """
+        Check for a specific minimum version of Docker on a remote node.
+        """
         client = get_docker_client(cluster, cluster.nodes[0].public_address)
         client_version = LooseVersion(client.version()['Version'])
         minimum_version = LooseVersion(required_version)
@@ -113,6 +116,11 @@ class DockerPluginTests(TestCase):
         return cid
 
     def _test_create_container(self, cluster):
+        """
+        Create a container running a simple HTTP server that writes to
+        its volume on POST and reads the same data back from the volume
+        on GET.
+        """
         data = random_name(self).encode("utf-8")
         node = cluster.nodes[0]
         http_port = 8080
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -112,15 +112,39 @@ class DockerPluginTests(TestCase):
             self.addCleanup(client.remove_container, cid, force=True)
         return cid
 
+    def _test_create_container(self, cluster):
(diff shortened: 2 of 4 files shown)


exarkun:
Thanks!

Questions:

  * Are you planning to implement any particular options?  Since none are described on the issue, I would guess not.
  * What's the desired behavior if an unrecognized option is received?
  * Do we have CI for Docker 1.9 yet?  If not, do we have a plan for setting it up?
  * Supposing docker-py doesn't support the new options, what could we do?  Monkey-patch?  Issue an API call ourselves?  What would those things look like?

FWIW, I think this issue would probably be just fine with no explicit design/design review step - but what it is missing is a better up-front (shared) understanding of what the goals of the work are.  I think this starts with the JIRA issue which is vague - it leans heavily on "compatibility" but it probably should have gone into more detail about what that means.

Do we have a product motivation to do more than just not be broken?  Does the product owner even know about this new feature in Docker 1.9 so they they could make a decision about this?

The code changes implied here seem quite reasonable to me.  Once you have answers to all of the above questions I think it should be fine to go ahead with the implementation.

itamarst:
I can answer some of these, since I've been talking to some other people:

The profiles team is planning on using an option passed in by Docker as way of indicating which profile a dataset should be created with. Other obvious candidates (not for this issue, but in future) is maximum_size and Flocker's concept of metadata.

The goal for this issue is basically "don't become completely unusable when using Docker 1.9". Given that goal ignoring unknown options is fine for now.

itamarst:
I've updated the JIRA issue to reflect minimal requirement of "don't break".

dwgebler:
All these questions had come up for me as well, some of my answers are "best guess" in the absence of more information in the JIRA issue.

> Are you planning to implement any particular options? Since none are described on the issue, I would guess not.

Indeed, no, the plan at this stage is that we completely ignore the value of `Opts`, I'm not even planning to explicitly validate that `Opts` is a map rather than some scalar at this stage. (we have no input schema defined at the moment)

> What's the desired behavior if an unrecognized option is received?

At this stage, in reference to the answer to question 1, I'm not concerned with a concept of recognised vs unrecognised options. In the future, `Err` in the response will describe any failure to validate `Opts`.

> Do we have CI for Docker 1.9 yet? If not, do we have a plan for setting it up?

We do not. I am intending to create a separate issue for that and at this stage, looking at manually testing against Docker 1.9. I assume that will not be considered sufficient for merging, but I don't know what's involved in setting up CI for a particular Docker version, particularly one that is not even considered stable release yet, so it's out of scope for this ticket.

> Supposing docker-py doesn't support the new options, what could we do? Monkey-patch? Issue an API call ourselves? What would those things look like?

Worst-case scenario I envision is calling out to the Docker API directly with manually constructed requests, or more likely patching `docker-py`'s `create_container` with an implementation copied and modified to support the `Opts` map - not a great solution, but we can check if `docker-py` has an existing issue to support this and maybe even upstream our modification. That's assuming this hasn't already been addressed in `docker-py`, though I'd consider it my lucky day if it had!

itamarst:
Quick look at docker-py suggests latest version does have this.

itamarst:
Or perhaps only `master`. But definitely being worked on.

itamarst:
I suggest a news entry in the documentation.

docs/releasenotes/index.rst
```diff
@@ -13,6 +13,7 @@ You can learn more about where we might be going with future releases by:
 Next Release
 ============
 
+* The :ref:`Flocker plugin for Docker<docker-plugin>` is now compatible with Docker 1.9
```
itamarst: Missing period at end for consistency.

itamarst:
This looks good, thanks for getting this fixed.

1. One minor comment above to fix.
2. You need to force build to ensure results show in GitHub, so I can't tell if tests are passing.
3. Did you test this manually with Docker 1.9 daily build, and see that it didn't work before but works now?

Once buildbot is green and manual testing has been done (if it hasn't already) please merge.

dwgebler:
> Did you test this manually with Docker 1.9 daily build, and see that it didn't work before but works now?

Ah, forgot to mention, yes this was tested manually against the latest experimental Docker 1.9 on AWS nodes.

itamarst:
:heart:

PR #2066, opened 2015-10-22
[FLOC-3270] Add a new interface for IBlockDeviceAPIs with storage profiles.

sarum90:
Introduce a new interface for drivers that are capable of creating profiles.

In the long term drivers can choose to provide IProfiledBlockDeviceAPI, and if they do, then flocker will call create_volume_with_profile if the user attempts to create a volume with a specified storage profile over the control API.

I suppose this might be arguably more of a design review disguised as a tiny code review.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2066)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -944,6 +944,42 @@ class IBlockDeviceAPI(Interface):
         """
 
 
+class IProfiledBlockDeviceAPI(IBlockDeviceAPI):
+    """
+    An interface for ``IBlockDeviceAPI`` implementers that are capable of
+    creating volumes with a specific profile.
+    """
+
+    def create_volume_with_profile(name, size, profile_name):
+        """
+        Create a new volume with the specified profile.
+
+        When called by ``IDeployer``, the supplied size will be
+        rounded up to the nearest
+        ``IBlockDeviceAPI.allocation_unit()``
+
+
+        :param UUID dataset_id: The Flocker dataset ID of the dataset on this
+            volume.
+        :param int size: The size of the new volume in bytes.
+        :param unicode profile_name: The name of the storage profile for this
+            volume.
+
+        :returns: A ``BlockDeviceVolume`` of the newly created volume.
+        """
+
+    def get_profile_for_volume(blockdevice_id):
+        """
+        Determine the name of the storage profile for a given blockdevice_id.
+
+        :param unicode blockdevice_id: The blockdevice_id of the volume to
+            query.
+
+        :returns unicode: The name of the profile that the volume currently
+            matches.
+        """
+
+
 @implementer(IBlockDeviceAsyncAPI)
 @auto_threaded(IBlockDeviceAPI, "_reactor", "_sync", "_threadpool")
 class _SyncToThreadedAsyncAPIAdapter(PRecord):
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -985,17 +985,6 @@ class IProfiledBlockDeviceAPI(Interface):
         :returns: A ``BlockDeviceVolume`` of the newly created volume.
         """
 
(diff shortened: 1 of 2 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -944,6 +944,42 @@ def get_device_path(blockdevice_id):
         """
 
 
+class IProfiledBlockDeviceAPI(IBlockDeviceAPI):
+    """
+    An interface for ``IBlockDeviceAPI`` implementers that are capable of
+    creating volumes with a specific profile.
+    """
+
```
myechuri: How about adding generation/version attribute to the interface?

generation = Attribute('1')

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2281,6 +2281,45 @@ def setUp(self):
...
+        self.assertTrue(
+            verifyObject(IProfiledBlockDeviceAPI, self.api)
+        )
+
+    def test_profile_respected(self):
+        """
+        Verify that the profile for a volume created with one of the mandatory
+        profiles is respected.
+        """
+        for profile in [u'gold', u'silver', u'bronze']:
```
myechuri: Should we define below in blockdevice.py?

class MandatoryProfiles(Values):
    """
    Mandatory Storage Profiles to be implemented by ``IProfiledBlockDeviceAPI``
    """
    GOLD = ValueConstant(u'gold')
    SILVER = ValueConstant(u'silver')
    BRONZE = ValueConstant(u'bronze')
    DEFAULT = ValueConstant(u'silver')

flocker/node/agents/blockdevice.py
sarum90: If it's an interface version number do we want it as an attribute or in the name?

IProfiledBlockDeviceAPIV1 or something.

itamarst: Version numbers as attributes of implementations would allow us to... change `IProfiledBlockDeviceAPI` within our code rather than copy/paste alter for new versions. But that means we don't have good way to refer to old version in any adapter code, so seems better to just copy/paste for this particular use case.

flocker/node/agents/blockdevice.py
```diff
@@ -944,6 +944,42 @@ def get_device_path(blockdevice_id):
         """
 
 
+class IProfiledBlockDeviceAPI(IBlockDeviceAPI):
```
itamarst: I don't think needs to inherit from `IBlockDeviceAPI`. It can just be an additional interface a class implements in addition to `IBlockDeviceAPI`, with additional methods.

flocker/node/agents/test/test_blockdevice.py
itamarst: I suggest default belongs outside the class, but yeah having explicit constants is a good idea.

itamarst: Or hm, maybe it does belong there. But perhaps `DEFAULT = SILVER`.

itamarst:
Thanks, address and merge; I'm assuming a decent chance we'll end up tweaking these a bit once we start using them so we'll have at least one review with actual usage in mind.

flocker/node/agents/blockdevice.py
sarum90: Sticking as is for now. Will re-visit before we have people outside the company try to implement one of these.

flocker/node/agents/test/test_blockdevice.py
sarum90: Added.

sarum90:
Addressed comments. I also removed 1 of the methods from the interface. (We can add it back later if we need it - I will implement that). If this code was in-use I might re-submit it for code review... given that it is not in use any decisions can be revisited and questioned in subsequent reviews, so I intend to submit once I get greens on buildbot.

PR #2067, opened 2015-10-23
[FLOC 3198] Use correct name for the plugin

davidoliverSP2:
Fixes 3198

The Flocker plugin for Docker was named Flocker Docker plugin when it was part of Labs. Now that this is supported, the rest of the docs should use the new name.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2067)

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -80,7 +80,7 @@ Upgrading is strongly recommended for all users of v1.0.0.
 
 * The EBS storage driver now more reliably selects the correct OS device file corresponding to an EBS volume being used.
 * Additional safety checks were added to ensure only empty volumes are formatted.
-* ClusterHQ Labs projects, including the Flocker Docker Plugin and an experimental Volumes CLI and GUI are now documented in the :ref:`Labs section <labs-projects>`.
+* ClusterHQ Labs projects, including the Flocker plugin for Docker and an experimental Volumes CLI and GUI are now documented in the :ref:`Labs section <labs-projects>`.
 
 v1.0
 ====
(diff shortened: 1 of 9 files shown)


docs/releasenotes/index.rst
```diff
@@ -80,7 +80,7 @@ Upgrading is strongly recommended for all users of v1.0.0.
 
 * The EBS storage driver now more reliably selects the correct OS device file corresponding to an EBS volume being used.
 * Additional safety checks were added to ensure only empty volumes are formatted.
-* ClusterHQ Labs projects, including the Flocker Docker Plugin and an experimental Volumes CLI and GUI are now documented in the :ref:`Labs section <labs-projects>`.
+* ClusterHQ Labs projects, including the Flocker plugin for Docker and an experimental Volumes CLI and GUI are now documented in the :ref:`Labs section <labs-projects>`.
```
adamtheturtle: Don't change this one, it was correct at the time and changing old release notes is confusing.

adamtheturtle:
Thanks @davidoliverSP2 . I think that almost all of these would be better with `:ref:`docker-plugin`. This renders as the text of the section title linked to the section. This would help avoid issues like this cropping up.

Please address and resubmit.

exarkun:
Thanks David.

There are some non-link occurrences of "the Flocker plugin for Docker" left - eg in `labs/installer.rst` - which duplicate the name of the thing being discussed.  If this is intentional (because you don't want links in those places, I guess?), that's fine, just wanted to point it out in case it wasn't.

Merge once you're happy.  Thanks again.

PR #2070, opened 2015-10-23
[FLOC 2744] Remove Fig and replace with docker-compose

davidoliverSP2:
Fixes 2744

When changing the Fig-compatible application config doc to use Docker-compose, it was simpler to split the large CLI topic into stand alone docs, so there was a little additional work here rather than just the change.

Other than the replacement of "Fig" with "docker-compose", no other content was changed, only moved.


davidoliverSP2:
Going to split the actions in this PR to smaller changes for ease of reviewing

PR #2071, opened 2015-10-23
[FLOC-3262] Retry removing container if there is a 500

adamtheturtle:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3262

There is no direct test for this. I made the call not to add a test because:
* Our test suite tells already us relatively often already that this is a problem
* I hope that this code can be removed soonish
* I can't see tests for the other similar workarounds (Ok, that is a terrible reason)
* What the user cares about is that the container is removed. This is tested for. The intent is specified in a comment for future developers.

Manual testing I've done: I have 16 tabs open in a Vagrant dev box running the test simultaneously. Almost every time on master this test fails on at least one of the 16 attempts. With this branch I have tried a few times and seen no failures. This makes me quite confident that this fix is suitable.

Alternatives include:
* Drop support for Docker <1.9.0 soon and live with this as-is until then.
* Only run this test on >=1.9.0 - this might cause users on <1.9.0 to see errors when trying to remove containers which they have presumably been seeing for a while.

I chose to work around the issue in the code because it didn't seem particularly difficult. I guess that if this fix isn't deemed suitable, then the alternatives can be discussed with the product team.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2071)


flocker/node/_docker.py
```diff
@@ -752,39 +752,51 @@ def _remove():
...
+                    # remove a container which it is already removing.
+                    # This has been resolved in https://github.com/docker/docker/commit/c4e49d10143178c718178ff7ce857f4f8ed46a0b  # noqa
+                    # This code should be removed once we drop support for
+                    # Docker versions older than 1.9.0. See FLOC-3284.
+                    if e.response.status_code == INTERNAL_SERVER_ERROR:
+                        Message.new(
+                            message_type="flocker:docker:container_remove_internal_error",  # noqa
+                            container=container_name
+                        ).write()
+                        continue
```
adamtheturtle: Docker's fix is the equivalent of having `return` here, and not putting this in a loop. I chose to put this in a loop and then hit `return` when the container doesn't exist. I'm not confident about what the trade-offs are.

flocker/node/_docker.py
```diff
@@ -752,39 +752,51 @@ def _remove():
...
+                            message_type="flocker:docker:container_not_found",
+                            container=container_name
+                        ).write()
+                        return
+                    # Sometimes Docker returns a 500 error when trying to
+                    # remove a container which it is already removing.
+                    # This has been resolved in https://github.com/docker/docker/commit/c4e49d10143178c718178ff7ce857f4f8ed46a0b  # noqa
+                    # This code should be removed once we drop support for
+                    # Docker versions older than 1.9.0. See FLOC-3284.
+                    if e.response.status_code == INTERNAL_SERVER_ERROR:
```
adamtheturtle: I'm interested in hearing review feedback on whether this should have a more specific check, e.g. of the response message

sarum90:
Nice find and dedication to testing the fix!

I'm a little concerned about some of the details of the fix - check my logic on one of the comments below.


---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[flocker/node/_docker.py, line 774 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1LMoJ95uMgOl64R-F0)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L774)):</sup>
I think there might be something unintentional here, but please double check my logic.
In the happy case where this doesn't raise any exception, this is now done twice after this PR (the second time we'll get the 404 and exit nicely), whereas it was only done once prior to this PR. Is this intentional? I realize perhaps that was even the intent of the PR, but the description makes it sound like you only intended on retrying on 500 errors rather than, "always retry until 404".


---

<sup>**[flocker/node/_docker.py, line 793 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1Kfmu3KZe72xoP0SlH-r1-793)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L793)):</sup>
See comment below about concern with infinite loop. I think retrying a couple times regardless of what caused the 500 makes sense. retrying infinite times on 500 sounds more scary to me.


---

<sup>**[flocker/node/_docker.py, line 798 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1Kfmu3KZe72xoP0SlH-r1-798)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L798)):</sup>
I think my bigger concern would be if docker gets in some bad state where it starts 500 on every request, we would be stuck in an infinite loop here. I'd be more comfortable with something along the lines of retry twice then raise the error.

I think it's probably nicer on our end to retry 500s (in general) rather than assuming we know what caused this 500 and we can assume it completed successfully.

Also I'm surprised that the fix is just a return here - I would have figured it was a wait_til_removal_completed and then a return.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2071)

sarum90:
Also I flunk at using reviewable... trying to leave the comments open rather than leaving them resolved...


---

Review status: 0 of 1 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/_docker.py, line 774 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1LMoJ95uMgOl64R-F0)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L774)):</sup>
leave open


---

<sup>**[flocker/node/_docker.py, line 793 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1Kfmu3KZe72xoP0SlH-r1-793)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L793)):</sup>
leave open


---

<sup>**[flocker/node/_docker.py, line 798 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2071#-K1Kfmu3KZe72xoP0SlH-r1-798)** ([raw file](https://github.com/clusterhq/flocker/blob/a36ddbfa76660eb4998ae840637015e6420229b2/flocker/node/_docker.py#L798)):</sup>
leave open


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2071)

adamtheturtle:
Thanks for the review @sarum90 . I've discussed this with @jml and the next round will replace "while True" with a loop on the new check (`loop_until` with some parameters for a maximum wait, or some modification of `retry_failure`). Then we can somehow choose when to stop checking. The balance is between: don't loop unnecessarily long, don't raise an error when the removal is going to happen soon.

adamtheturtle:
Testing setup:

I used iTerm to split a console into 16 sections and broadcast input to all.
I `vagrant up`, `vagrant ssh` and then I go to a virtualenv with Flocker's dev dependencies.
I run:

```
(flocker)[vagrant@localhost vagrant]$ sudo $(type -p trial) flocker.node.functional.test_docker.NamespacedDockerClientTests.test_dead_is_removed
```

and see that the test passes in its 16 runs  (takes up to 2.5 mins usually).

What remains (at least):

* Move loop_until out of testtools
* Give loop_until some kind of parameters for the delay between tries, and the max number of loops
* Somehow choose those parameters

jongiddy:
Closing this in favor of #2107

PR #2072, opened 2015-10-24
[FLOC-3260] Update AWS base images.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3260?jql=text%20~%20%22update%20centos%22

Both CentOS and Ubuntu have newer cloud images in AWS. Since we install all updates anyway, upgrade the base image to reduce the updates we need to install.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2072)

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -40,10 +40,9 @@ _usernames = {
 
 
 IMAGE_NAMES = {
-    'centos-7': 'CentOS 7 x86_64 (2014_09_29) EBS HVM'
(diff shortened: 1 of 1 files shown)


adamtheturtle:
Thanks @tomprince . To confirm, this is just to make the builds quicker because they have to download / install less, right? If that is the case then merge if you and Buildbot are happy. If not, please explain what I'm missing and resubmit.

Separately, it would be good to have some kind of policy on how often we update these.

tomprince:
Also,
- this is closer to what users will be using (since they mostly likely use the latest images at install time
- if there are any security holes in distribution packages that are default exploitable, we close the window where the test machines are being updated.

PR #2076, opened 2015-10-26
[FLOC-3077] Retry `list` in client tests to avoid spurious failures

jml:
Our centos-7 tests often fail with `docker.errors.APIError: 500: Internal Server Error: Unknown device`

This is *probably* due to a race condition in Docker. See [FLOC-3077](https://clusterhq.atlassian.net/browse/FLOC-3077) for discussion. 

The work-around is to retry the list operation in the test in order to avoid the race condition. 

Rather than putting all of that logic in the test, I added a new `retry_failure` helper which does this. I considered trying to shoe-horn `loop_until` to fit this purpose, but it seemed cleaner to do it this way.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2076)

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -789,6 +789,7 @@ class DockerClient(object):
         return d
 
     def list(self):
+        # XXX: This should use flocker.common.auto_threaded
         def _list():
             result = set()
             ids = [d[u"Id"] for d in
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -1,4 +1,4 @@
-# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# Copyright ClusterHQ Ltd.  See LICENSE file for details.
 # -*- test-case-name: flocker.node.test.test_docker -*-
 
 """
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -789,7 +789,6 @@ class DockerClient(object):
         return d
 
     def list(self):
-        # XXX: This should use flocker.common.auto_threaded
         def _list():
             result = set()
             ids = [d[u"Id"] for d in
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -1,4 +1,4 @@
-# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# Copyright ClusterHQ Ltd.  See LICENSE file for details.
 
 """
 Functional tests for :module:`flocker.node._docker`.
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -527,6 +527,7 @@ class GenericDockerClientTests(TestCase):
         """
         The Docker image is pulled if it is unavailable locally.
         """
+        # XXX: This test calls 'list'. But how?
         client = Client()
 
         path = FilePath(self.mktemp())
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -32,7 +32,7 @@ from pyrsistent import PClass, pvector, field
 
 from ...testtools import (
     loop_until, find_free_port, DockerImageBuilder, assertContainsAll,
-    random_name)
+    random_name, retry_failure)
 
 from ..test.test_docker import ANY_IMAGE, make_idockerclient_tests
 from .._docker import (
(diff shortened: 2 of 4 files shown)


sarum90:
Awesome little test wrapper.

Your tests look like they want to be written in pymock :grin:. A nit and a question on your drive-by added XXX. Address and merge at will.


---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[flocker/node/_docker.py, line 792 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1_ESL0O1z6kO0zvWmh)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/node/_docker.py#L792)):</sup>
Should this be here, or at the class level? looks like auto_threaded is used with classes rather than functions. Also see optional point below about potentially removing:

:grey_question: Fine as is... but I'm never sure what to think of comments like this. Presumably while you were coding this PR, you noticed this, left the TODO/XXX.

The proposed value-add of this comment is later, if someone comes across this and is wondering why it isn't using this nice wrapper, they'll see your comment, and feel vindicated that it should use the nice wrapper and will go on coding it.

I don't think we ever will schedule a "reduce TODOs" epic, so I don't think this comment puts work on any engineer work-queue.

Comments always have the overhead of adding maintenance and risk becoming stale. Some of the methods of this class do some synchronous stuff before calling deferToThread. Perhaps it'd be okay for that work to be delayed til the other thread, but it might intentionally be done synchronously before scheduling async work, or that might be the case in the future. In that case this might misdirect a future reader.

Potentially consider just removing the comment and letting future maintainers make the assessment if the wrapper is useful? I'll defer to your judgement on if it's value-add is worth it.

Side note: I always get a little sad perusing a codebase with a bunch of TODOs littered throughout :stuck_out_tongue:. 


---

<sup>**[flocker/testtools/__init__.py, line 261 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1Zn6h8UjC5haUz5c-i)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/testtools/__init__.py#L261)):</sup>
:grey_question: Cool! Are you familiar with https://pypi.python.org/pypi/retry/? looks like this is just the twisted implementation of that. Just thought I'd note that here in case you wanted to look at that for API ideas. Looks like you've got it covered though.


---

<sup>**[flocker/testtools/__init__.py, line 268 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1_NaNwSfh_EvtN_V-1)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/testtools/__init__.py#L268)):</sup>
:grey_question: iterable of Exceptions rather than just "the exceptions" is clearer I think. Or I guess anything that can be unpacked to be arguments to Failure.check.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2076)

jml:
Thanks! It will take more than that to overcome my postjudice (is that a word? is now) against mocking.  ;)


---

Review status: 0 of 4 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[flocker/node/_docker.py, line 792 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1_ESL0O1z6kO0zvWmh)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/node/_docker.py#L792)):</sup>
Done. That's a good analysis of TODO/XXX comments.


---

<sup>**[flocker/testtools/__init__.py, line 261 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1Zn6h8UjC5haUz5c-i)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/testtools/__init__.py#L261)):</sup>
I wasn't, actually. Decorator and built-in backoff & jitter are both nice features (although I am perhaps over-fond of the iterator approach I've taken here). 


---

<sup>**[flocker/testtools/__init__.py, line 268 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2076#-K1_NaNwSfh_EvtN_V-1)** ([raw file](https://github.com/clusterhq/flocker/blob/1046ff01f4d91b6cc31dcf8374fc080faeb1c0bb/flocker/testtools/__init__.py#L268)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2076)

jml:
Have re-run failing tests manually on buildbot.

PR #2079, opened 2015-10-26
[FLOC-1853] Added timeout to destroy volume

mpgomez:
Design for https://clusterhq.atlassian.net/browse/FLOC-1853

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2079)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -883,6 +883,8 @@ class IBlockDeviceAPI(Interface):
 
         :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
             exist.
+            Note that the ebs and cinder implementation can raise a
+            TimeoutException
 
         :return: ``None``
         """
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -883,8 +883,6 @@ class IBlockDeviceAPI(Interface):
 
         :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
             exist.
-            Note that the ebs and cinder implementation can raise a
-            TimeoutException
 
         :return: ``None``
         """
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -403,7 +403,8 @@ class CinderBlockDeviceAPI(object):
                  cinder_volume_manager,
                  nova_volume_manager, nova_server_manager,
                  cluster_id,
-                 timeout=CINDER_VOLUME_DESTRUCTION_TIMEOUT):
+                 timeout=CINDER_VOLUME_DESTRUCTION_TIMEOUT,
+                 time_module=None):
         """
         :param ICinderVolumeManager cinder_volume_manager: A client for
             interacting with Cinder API.
@@ -420,6 +421,9 @@ class CinderBlockDeviceAPI(object):
         self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
         self._timeout = timeout
+        if time_module is None:
+            time_module = time
+        self._time = time_module
 
     def allocation_unit(self):
         """
@@ -583,20 +587,20 @@ class CinderBlockDeviceAPI(object):
             self.cinder_volume_manager.delete(blockdevice_id)
         except CinderNotFound:
             raise UnknownVolume(blockdevice_id)
-        start_time = time.time()
+        start_time = self._time.time()
         # Wait until the volume is not there or until the operation
         # timesout
-        while(time.time() - start_time < self._timeout):
+        while(self._time.time() - start_time < self._timeout):
             try:
                 self.cinder_volume_manager.get(blockdevice_id)
             except CinderNotFound:
                 return
-            time.sleep(1.0)
+            self._time.sleep(1.0)
         # If the volume is not deleted, raise an exception
         raise TimeoutException(
             unicode(blockdevice_id),
             None,
-            time.time() - start_time
+            self._time.time() - start_time
         )
 
     def _get_device_path_virtio_blk(self, volume):
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -236,6 +236,10 @@ class INovaServerManager(Interface):
 class TimeoutException(Exception):
     """
     A timeout on waiting for volume to reach destination end state.
+    :param expected_volume: the volume we were waiting on
+    :param desired_state: the new state we wanted the volume to have
+    :param elapsed_time: how much time we had been waiting for the volume
+        to change the state
     """
     def __init__(self, expected_volume, desired_state, elapsed_time):
         self.expected_volume = expected_volume
@@ -398,7 +402,8 @@ class CinderBlockDeviceAPI(object):
     def __init__(self,
                  cinder_volume_manager,
                  nova_volume_manager, nova_server_manager,
-                 cluster_id):
+                 cluster_id,
+                 timeout=CINDER_VOLUME_DESTRUCTION_TIMEOUT):
         """
         :param ICinderVolumeManager cinder_volume_manager: A client for
             interacting with Cinder API.
@@ -414,7 +419,7 @@ class CinderBlockDeviceAPI(object):
         self.nova_volume_manager = nova_volume_manager
         self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
-        self.timeout = CINDER_VOLUME_DESTRUCTION_TIMEOUT
+        self._timeout = timeout
 
     def allocation_unit(self):
         """
@@ -581,14 +586,18 @@ class CinderBlockDeviceAPI(object):
         start_time = time.time()
         # Wait until the volume is not there or until the operation
         # timesout
-        while(time.time() - start_time < self.timeout):
+        while(time.time() - start_time < self._timeout):
             try:
                 self.cinder_volume_manager.get(blockdevice_id)
             except CinderNotFound:
                 return
             time.sleep(1.0)
         # If the volume is not deleted, raise an exception
-        raise TimeoutException(unicode(blockdevice_id), None, self.timeout)
+        raise TimeoutException(
+            unicode(blockdevice_id),
+            None,
+            time.time() - start_time
+        )
 
     def _get_device_path_virtio_blk(self, volume):
         """
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -563,19 +563,35 @@ class CinderBlockDeviceAPI(object):
             transient_states=(u'in-use', u'detaching')
         )
 
-    def destroy_volume(self, blockdevice_id):
+    def destroy_volume(self, blockdevice_id, timeout=300):
+        """
+        Detach Cinder volume identified by blockdevice_id.
+
+        :raises TimeoutException: If the volume is not deleted
+        within the expected time
+        """
         try:
             self.cinder_volume_manager.delete(blockdevice_id)
         except CinderNotFound:
             raise UnknownVolume(blockdevice_id)
-
-        while True:
-            # Don't loop forever here.  FLOC-1853
+        start_time = time.time()
+        #NOTE: maybe we could share the ''VOLUME_STATE_CHANGE_TIMEOUT''
+        #in the ebs implementation
+        #TODO do we need to check the state instead of waiting until the volume
+        #is deleted?
+        #Does it go to "deleting" state or something?
+        #Is it worth it? (how much time is there between going to "deleting"
+        #state and actually deleting the volume?
+        #We would need to find out all the possible states of a Cinder volume
+        #and put a list here
+        while(time.time() - start_time < timeout):
             try:
                 self.cinder_volume_manager.get(blockdevice_id)
             except CinderNotFound:
-                break
+                return
             time.sleep(1.0)
+        #If the volume is not deleted, raise an exception
+        raise TimeoutException(unicode(blockdevice_id), None, timeout)
 
     def _get_device_path_virtio_blk(self, volume):
         """
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -579,9 +579,13 @@ class CinderBlockDeviceAPI(object):
     def destroy_volume(self, blockdevice_id):
         """
         Detach Cinder volume identified by blockdevice_id.
+        It will loop listing the volume until it is no longer there. If the volume is still there
+        after the defined timeout, the function will just raise an exception and do nothing.
 
         :raises TimeoutException: If the volume is not deleted
-            within the expected time
+            within the expected time. If that happens, it will be because after the timeout,
+            the volume can still be listed. The volume will not be deleted unless further action
+            is taken.
         """
         try:
             self.cinder_volume_manager.delete(blockdevice_id)
@@ -598,7 +602,7 @@ class CinderBlockDeviceAPI(object):
             self._time.sleep(1.0)
         # If the volume is not deleted, raise an exception
         raise TimeoutException(
-            unicode(blockdevice_id),
+            blockdevice_id,
             None,
             self._time.time() - start_time
         )
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -579,13 +579,14 @@ class CinderBlockDeviceAPI(object):
     def destroy_volume(self, blockdevice_id):
         """
         Detach Cinder volume identified by blockdevice_id.
-        It will loop listing the volume until it is no longer there. If the volume is still there
-        after the defined timeout, the function will just raise an exception and do nothing.
+        It will loop listing the volume until it is no longer there.
+        if the volume is still there after the defined timeout,
+        the function will just raise an exception and do nothing.
 
         :raises TimeoutException: If the volume is not deleted
-            within the expected time. If that happens, it will be because after the timeout,
-            the volume can still be listed. The volume will not be deleted unless further action
-            is taken.
+            within the expected time. If that happens, it will be because after
+            the timeout, the volume can still be listed. The volume will not
+            be deleted unless further action is taken.
         """
         try:
             self.cinder_volume_manager.delete(blockdevice_id)
@@ -754,8 +755,10 @@ def _blockdevicevolume_from_cinder_volume(cinder_volume):
 
 
 @auto_openstack_logging(ICinderVolumeManager, "_cinder_volumes")
-class _LoggingCinderVolumeManager(PRecord):
-    _cinder_volumes = field(mandatory=True)
+class _LoggingCinderVolumeManager(object):
+
+    def __init__(self, cinder_volumes):
+        self._cinder_volumes = cinder_volumes
 
 
 @auto_openstack_logging(INovaVolumeManager, "_nova_volumes")
@@ -894,8 +897,9 @@ def cinder_from_configuration(region, cluster_id, **config):
     nova_client = get_nova_v2_client(session, region)
 
     logging_cinder = _LoggingCinderVolumeManager(
-        _cinder_volumes=cinder_client.volumes
+        cinder_client.volumes
     )
+
     logging_nova_volume_manager = _LoggingNovaVolumeManager(
         _nova_volumes=nova_client.volumes
     )
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -55,6 +55,7 @@ CINDER_TIMEOUT = 600
 # The longest time we're willing to wait for a Cinder volume to be destroyed
 CINDER_VOLUME_DESTRUCTION_TIMEOUT = 300
 
+
 def _openstack_logged_method(method_name, original_name):
     """
     Run a method and log additional information about any exceptions that are
@@ -413,7 +414,7 @@ class CinderBlockDeviceAPI(object):
         self.nova_volume_manager = nova_volume_manager
         self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
-        self.timeout=CINDER_VOLUME_DESTRUCTION_TIMEOUT
+        self.timeout = CINDER_VOLUME_DESTRUCTION_TIMEOUT
 
     def allocation_unit(self):
         """
@@ -579,14 +580,14 @@ class CinderBlockDeviceAPI(object):
             raise UnknownVolume(blockdevice_id)
         start_time = time.time()
         # Wait until the volume is not there or until the operation
-        #timesout
+        # timesout
         while(time.time() - start_time < self.timeout):
             try:
                 self.cinder_volume_manager.get(blockdevice_id)
             except CinderNotFound:
                 return
             time.sleep(1.0)
-        #If the volume is not deleted, raise an exception
+        # If the volume is not deleted, raise an exception
         raise TimeoutException(unicode(blockdevice_id), None, self.timeout)
 
     def _get_device_path_virtio_blk(self, volume):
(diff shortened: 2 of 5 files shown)


itamarst:
In general, looking good!

Specific design review comments: you can have much less detail in designs; it can just be a simple sketch with details to be filled out. Alternatively, if you've gotten this far along in the implementation in a short period of time, you could just skip straight to code review and just note in the PR that there was no design review.

flocker/node/agents/cinder.py
```diff
@@ -563,19 +563,35 @@ def detach_volume(self, blockdevice_id):
...
             self.cinder_volume_manager.delete(blockdevice_id)
         except CinderNotFound:
             raise UnknownVolume(blockdevice_id)
-
-        while True:
-            # Don't loop forever here.  FLOC-1853
+        start_time = time.time()
+        #NOTE: maybe we could share the ''VOLUME_STATE_CHANGE_TIMEOUT''
+        #in the ebs implementation
+        #TODO do we need to check the state instead of waiting until the volume
```
itamarst: That would be another legitimate alternative. I think there is a "deleting" state, and it seems reasonable to just wait until that shows up. There is however some expectation that deleted volumes don't show up in list_volume(), so we may need to wait until it's actually not there. So this is fine.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,39 @@ def test_verify_false_ca_path(self):
             'verify_ca_path': '/a/path'
         }
         self.assertEqual(_openstack_verify_from_config(**config), False)
+
+
+@implementer(ICinderVolumeManager)
```
itamarst: These classes will need docstrings for final code review.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,39 @@ def test_verify_false_ca_path(self):
...
+        """
+        A no-op get.
+        """
+
+
+class CinderDestroyTests(SynchronousTestCase):
+    def test_timeout(self):
+        """
+        ``CinderBlockDeviceAPI.destroy_volume`` raises
+        ``TimeoutException`` if the volume is not removed in 60s.
```
itamarst: This doesn't match the actual timeout hard-coded below (1s) nor the timeout in the implementation.  Maybe "if the volume is not deleted within some time."

(As a side-note, it's also not an deal test insofar as it's somewhat different than real-world usage. On the other hand it's probably the best test one *could* do without massive infrastructure work on OpenStack error injection :grinning: so this is good approach.)

itamarst:
Ok, please do final cleanups, force buildbot run and put up for code review. Thanks!

flocker/node/agents/test/test_cinder.py
mpgomez: the 1 second thing was because if this tests is going to run periodically, maybe we don't want it to get stuck for 300 seconds. I still need to find out how to do a proper acceptance test. Any ideas are more than welcome!

mpgomez:
Changes:
Added extra documentation
Removed the timeout from the destroy_volume function, so the signature actually matches the interface (IBlockDeviceAPI) that CinderBlockDeviceAPI is implementing. The timeout is now an attribute of CinderBlockDeviceAPI
Removed useless variables from test that were only used to pass them down to the CinderBloockDevice constructor. Instantiating them directly now.

QUESTION:
Is now the moment to do a bigger (with more functionality) FakeCinderClient class? - I removed the @implementer bit as it is not even implementing all the function of the IBlockDeviceAPI interface, only the two it needs.

flocker/node/agents/blockdevice.py
```diff
@@ -883,6 +883,8 @@ def destroy_volume(blockdevice_id):
 
         :raises UnknownVolume: If the supplied ``blockdevice_id`` does not
             exist.
+            Note that the ebs and cinder implementation can raise a
```
itamarst: I'm not sure this should be on interface documentation.

In general in Python (and really most languages with exceptions AFAIK, even Java with supposedly explicit exceptions) you have to assume that there are additional unexpected random exceptions that will get thrown for some error cases. It's worth documenting this in the relevant implementation docstrings, though.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +65,59 @@ def test_verify_false_ca_path(self):
             'verify_ca_path': '/a/path'
         }
         self.assertEqual(_openstack_verify_from_config(**config), False)
+
+
+class FakeCinderClient(object):
+    """"
+    Mock implementation of the cinder volume manager to use in the Cinder
```
itamarst: I'd say "fake". Mock usually implies standard Python `mock` library.

itamarst: Not actionable, just FYI: for fakes we try to have verified fakes. That is, define interface that both fake and real client implement, and then run same tests against both implementation to ensure they act in same way. See e.g. `flocker.node._docker`, `flocker.node.test.test_docker` and `flocker.node.functional.test_docker`, where we have real and fake Docker clients tested by same interface compliance suite.

In this case I don't think it's worth the effort unless we decide OpenStack backend is higher priority and needs a lot more polish and in-depth testing.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +65,59 @@ def test_verify_false_ca_path(self):
...
+    Destroy tests.
+    This class will actually do nothing as we want to emulate a non-responsive
+    cinder volume.
+    """
+    def delete(self, volume_id):
+        """
+        A no-op delete.
```
itamarst: Minor nit: typically we'd have extra line break between first sentence and rest of docstring.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +65,59 @@ def test_verify_false_ca_path(self):
...
+        (see FLOC-1853)
+        """
+
+
+class CinderDestroyTests(SynchronousTestCase):
+    """
+    Tests for destroy_volume.
+    """
+
+    # Test added with the fix of the issue FLOC-1853
```
itamarst: Put this in docstring, I think. Or maybe just omit and rely on merge history?

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +65,59 @@ def test_verify_false_ca_path(self):
...
+    def test_timeout(self):
+        """
+        ``CinderBlockDeviceAPI.destroy_volume`` raises
+        ``TimeoutException`` if the volume is not removed within
+        some time (``_currentTimeout`` will define how long it will wait).
```
itamarst: ``_currentTimeout`` isn't the actual name of the attribute you're setting. And maybe rely on test body to explain this?

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +65,59 @@ def test_verify_false_ca_path(self):
...
+    """
+    def delete(self, volume_id):
+        """
+        A no-op delete.
+        We do not need the delete to do anything for the timeout tests.
+        """
+
+    def get(self, volume_id):
+        """
+        A no-op get.
```
itamarst: This might be more robust if it returned something that matched the interface, e.g. a hardcoded `BlockDeviceVolume` or something. As it is this will break the test if implementation of delete changes slightly.

flocker/node/agents/cinder.py
```diff
@@ -411,6 +413,7 @@ def __init__(self,
         self.nova_volume_manager = nova_volume_manager
         self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
+        self.timeout=CINDER_VOLUME_DESTRUCTION_TIMEOUT
```
itamarst: 1. Maybe `delete_timeout` or `_delete_timeout` for more accurate explanation of how it's used.
2. Document the attribute in the class docstring.

itamarst:
Thanks for working on this!

1. There's a bunch of minor comments to address.
2. I don't think this is time to make complete new fake, no, see inline comments.
3. I don't see any test results; I suspect there'll be some lint failures. You can force a build by visiting http://build.clusterhq.com/boxes-flocker , typing in branch name in form at top-right and hitting "FORCE" button.

It occurs to me (sorry only at this late point) that there is one alternative testing strategy: use a real client, but just make its `delete()` method a no-op via subclassing or setting attribute directly:

```
c = RealCinderClient()
c.delete = lambda *args, **kwargs: None
# Now c.delete() does nothing.
```

Perhaps that would be simpler?

Please address comments above and resubmit.

mpgomez:
Thanks Itamar!

flocker/node/agents/blockdevice.py
exarkun: Reinforcing this.  I'd say instead that details about particular implementations of the interface belong on the implementations.  If a consumer of the interface *has* to know about those implementation details in order to use the interface correctly then the implementations don't actually conform to the interface (ie, they're broken).

Maybe the thing to say here instead is that `TimeoutException` is just another exception this method should be expected to raise.  Ideally that would come with tests and some changes to the existing consumers of the interface (so perhaps a follow-up issue).

flocker/node/agents/cinder.py
```diff
@@ -411,6 +414,7 @@ def __init__(self,
         self.nova_volume_manager = nova_volume_manager
         self.nova_server_manager = nova_server_manager
         self.cluster_id = cluster_id
+        self.timeout = CINDER_VOLUME_DESTRUCTION_TIMEOUT
```
exarkun: This doesn't need to be public.  Private by default is a good policy.

flocker/node/agents/cinder.py
```diff
@@ -564,18 +568,27 @@ def detach_volume(self, blockdevice_id):
         )
 
     def destroy_volume(self, blockdevice_id):
+        """
+        Detach Cinder volume identified by blockdevice_id.
+
+        :raises TimeoutException: If the volume is not deleted
+            within the expected time
```
exarkun: Anything more we can expect in this case?  Is the volume going to be on its way to being deleted?  Or if this exception is raised then the volume won't be deleted unless further action is taken?

flocker/node/agents/cinder.py
```diff
@@ -564,18 +568,27 @@ def detach_volume(self, blockdevice_id):
...
+        # timesout
+        while(time.time() - start_time < self.timeout):
             try:
                 self.cinder_volume_manager.get(blockdevice_id)
             except CinderNotFound:
-                break
+                return
             time.sleep(1.0)
+        # If the volume is not deleted, raise an exception
+        raise TimeoutException(unicode(blockdevice_id), None, self.timeout)
```
exarkun: The interface says `blockdevice_id` is unicode - so is this conversion required?  That suggests a caller is doing something wrong.

exarkun: The parameters and attributes of `TimeoutException` are undocumented.  Adding a new use of the type is probably a good time to correct that mistake.

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -34,6 +38,17 @@ def cinder_volume_manager():
     return get_cinder_v1_client(session, region).volumes
 
 
+def fakeGet(volume_id):
+    """
+
```
exarkun: missing docstring

exarkun: `fake_get` is probably more in line with the naming convention

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +127,39 @@ def test_updated_metadata_is_listed(self):
...
+    def test_destroy_timesout(self):
+        """
+        If the cinder cannot delete the volume, we should timeout
+        after waiting some time
+        """
+        # Adding a fake get that will always return a volume, so we
+        # timeout waiting for the volume to be deleted
+        self.patch(self.cinder_volumes, "get", fakeGet)
+        # Using a fake no-op delete so it doens't actually delete anything
```
exarkun: typo, `doens't`

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +127,39 @@ def test_updated_metadata_is_listed(self):
...
+
+        api = CinderBlockDeviceAPI(
+            cinder_volume_manager=self.cinder_volumes,
+            nova_volume_manager=object(),
+            nova_server_manager=object(),
+            cluster_id=uuid4()
+            )
+        # Setting the timeut to 1, as the default is quite high, and we do not
+        # want to wait that much in a test
+        api.timeout = 1
```
exarkun: The tragedy of blocking APIs... A fake clock on which you controlled the passage of time would make this test run about 1 second faster.  It's hard to do that in this case, though, since you'd need to use an extra thread to control the clock (or run the delete, one or the other).

Separately, making the timeout a parameter to `__init__` would be better.  Setting attributes on other objects is always risky (maybe  you misspell the name, maybe the implementation changes so the attribute isn't used anymore, etc) and the failures are often hard to track down.

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +127,39 @@ def test_updated_metadata_is_listed(self):
...
+        self.assertRaises(
+            TimeoutException,
+            api.destroy_volume,
+            blockdevice_id=unicode(uuid4())
+        )
```
exarkun: This demonstrates a timeout happens but doesn't demonstrate anything about when it happens.  The specified timeout value might be ignored and the timeout raised after 0.1 seconds or 1000 seconds... The test passes either way.

It would be nice to make some assertion about time here.  This is another place where a fake clock is nice since it's rather difficult to make any reliable assertions about the result of `time.time()`.  Even if you assert something like "the time after the exception is no later than 10 seconds after the time before the call was made" you still run the risk that someone will run the test on a loaded machine where 10 seconds actually do elapse in between those events even though the timeout of 1 second is being properly respected by the implementation.

One option (not something we've tried in flocker before afaik) might be to add some logging and try making the time assertions based on the contents of the log messages.  For example, if the loop logged the Eliot equivalent to "giving up because at current time=x is after planned timeout time=y, deletion not achieved within timeout=z seconds of starting time=a" and then you can check the logged values and see if the logging logic at least reflects the right timeout behavior.

Another option would be to try really using a fake clock.  Maybe you can avoid needing threads if you use something like a fake that reports real time but records the values it reports back to a caller.  Then the timeout logic is satisfied if the last reported time is the first time that's beyond the desired timeout.  Still a bit fragile (it encodes a lot of logic about the implementation strategy for the timeout - if the implementation changes to use the clock differently, maybe it is still correct but the test starts failing) but maybe it's better than nothing.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
             'verify_ca_path': '/a/path'
         }
         self.assertEqual(_openstack_verify_from_config(**config), False)
+
+
+class FakeCinderClient(object):
+    """"
+    Fake implementation of the cinder volume manager to use in the Cinder
+    Destroy tests.
+    Right now, we don't need a full mock of the class, we just need the
```
exarkun: in general, we're probably going to try not to use mocks anywhere in Flocker's tests (at least, that's my inclination).  I find other strategies almost always produce better tests.

I point this out because it's easy for people to mix up the different kinds of test fake strategies so wherever possible we should be clear on the terms we use and make sure we're all on the same page about what they mean.

In this case, I'd say "we don't need a full fake" (or a "complete fake", reads slightly more easily).

It may be that the deletion timeout logic never needs something more complete than this.  Perhaps this isn't just "FakeCinderClient", perhaps this is "SlowDeletionCinderClient" ("Slow" doesn't quite capture it since it actually never deletes things you ask it to... "SilentDeletionFailureCinderClient" is maybe too verbose a class name though).

So maybe it's worth saying (and committing to) this is only a fake suitable for those particular deletion tests and if we need some other kind of behavior we'll build another fake for it.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
...
+    get not to raise any exception so the destroy volume timeout test
+    actually times out - it will be trying to get the volume in a loop
+    after deleting it, expecting a ``CinderNotFound`` exception that will
+    never happen, so we can verify that the timeout exception is rised
+    (see FLOC-1853)
+    """
+    def delete(self, volume_id):
+        """
+        A no-op delete.
+        We do not need the delete to do anything for the timeout tests.
```
exarkun: Even stronger than this, we need for deletion to not succeed (at least not quickly) for the timeout tests.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
...
+        """
+        A no-op get.
+        Return a fake BlockDeviceVolume
+        """
+        return BlockDeviceVolume(
+            size=100, attached_to=None,
+            dataset_id=uuid4(),
+            blockdevice_id=unicode(volume_id),
+        )
+
```
exarkun: This looks a lot like `fake_get`.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
...
+        api = CinderBlockDeviceAPI(
+            cinder_volume_manager=FakeCinderClient(),
+            nova_volume_manager=object(),
+            nova_server_manager=object(),
+            cluster_id=uuid4()
+        )
+        # Timeout = 1 because the default timeout is 300 secods, and
+        # it is a bit too long for a test that runs regularly under CI
+        returnedBlockDevice = FakeCinderClient().get(uuid4())
+        if(not isinstance(returnedBlockDevice, BlockDeviceVolume)):
```
exarkun: It's not a style violation but we generally don't use parens in this case (unless spanning multiple lines).  If we did, we'd put a space after the `if`, I think.

If the linter doesn't complain, though, this isn't actually something you have to fix.

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
             'verify_ca_path': '/a/path'
         }
         self.assertEqual(_openstack_verify_from_config(**config), False)
+
+
+class FakeCinderClient(object):
```
exarkun: Also worth pointing out there's a project for building a test-friendly fake of the OpenStack API, https://github.com/rackerlabs/mimic

I think we're not using it because it doesn't do Cinder.  But maybe we should consider contributing...

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +127,39 @@ def test_updated_metadata_is_listed(self):
         missing_items = expected_items - actual_items
 
         self.assertEqual(set(), missing_items)
+
+
+class BlockDeviceAPIDestroyTests(SynchronousTestCase):
+    """
+    Test for ``cinder.CinderBlockDeviceAPI.destroy_volume``
+    """
```
exarkun: These tests belong in `test_cinder.py`.  Tests for behavior of our code goes there.  `test_cinder_behavior.py` is more like "hey, does Cinder actually work like we think it does?"

flocker/node/agents/test/test_cinder.py
```diff
@@ -61,3 +67,65 @@ def test_verify_false_ca_path(self):
...
+        # it is a bit too long for a test that runs regularly under CI
+        returnedBlockDevice = FakeCinderClient().get(uuid4())
+        if(not isinstance(returnedBlockDevice, BlockDeviceVolume)):
+            self.fail("returned block device is not a BlockDeviceVolume")
+        api.timeout = 1
+        self.assertRaises(
+            TimeoutException,
+            api.destroy_volume,
+            blockdevice_id=unicode(uuid4())
+        )
```
exarkun: This test is a lot like the other one.  It would be cool if some code could be shared between these two somehow.

flocker/node/agents/test/test_cinder.py
exarkun: What's the point of checking the result of `FakeCinderClient().get`?  It's a test helper and it does return a `BlockDeviceVolume`...

exarkun:
Thanks Pilar!  This looks pretty good overall.  I left some comments about minor points inline.  Also, I think I saw a couple comments from Itamar's previous review that were unaddressed.  Please have a look and then re-submit.

flocker/node/agents/blockdevice.py
itamarst: This still needs to be addressed.

flocker/node/agents/cinder.py
itamarst: This hasn't been addressed, needs another sentence I guess.

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +115,69 @@ def test_updated_metadata_is_listed(self):
...
+        # the timeout)
+        self.patch(self.cinder_volumes, "delete", lambda *args, **kwargs: None)
+        # Now try to delete it
+        time_module = FakeTime(initial_time=0)
+        api = CinderBlockDeviceAPI(
+            cinder_volume_manager=self.cinder_volumes,
+            nova_volume_manager=object(),
+            nova_server_manager=object(),
+            cluster_id=uuid4(),
+            # Setting the timeut to 1, as the default is quite high,
```
itamarst: No longer relevant comment. In fact you don't need that argument at all given fake time module I guess?

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -112,3 +115,69 @@ def test_updated_metadata_is_listed(self):
...
+            time_module=time_module
+            )
+
+        exception = self.assertRaises(
+            TimeoutException,
+            api.destroy_volume,
+            blockdevice_id=listed_volume.id
+        )
+
+        self.assertEqual(
```
itamarst: It's nice to only have on assertion call per test (perhaps excepting `assertRaises`, and custom assertion utility functions). One thing we've been doing:

```
self.assertEqual(dict(a=one_thing, b=another_thing),
                           dict(a=1, # we expect this because reasons
                                  b=3)) # we expect this because other reasons
```

This gives full info on test failures: you can see all the things that were wrong in result.

flocker/node/agents/test/test_cinder.py
itamarst: Still needs addressing.

flocker/node/agents/test/test_cinder.py
itamarst: Still needs addressing.

itamarst:
1. There's a bunch of minor things that need addressing / haven't been addressed from previous reviews.
2. Address those, make sure buildbot is all green, and merge.

Thanks for all the work on this!

flocker/node/agents/functional/test_cinder_behaviour.py
mpgomez: The comment is irrelevant (removed now!), but we kind of need the expect timeout. It will be making an API call every loop, so waiting until it makes 300 calls (the default timeout) results in a rather slow test. 
I'll document why we set it to 8

flocker/node/agents/functional/test_cinder_behaviour.py
mpgomez: As I cannot combine the assertRises with the assertEqual, I guess I cannot use the dictionary approach, so, should I just leave it as it is?

mpgomez:
Green build! 
http://build.clusterhq.com/boxes-flocker?branch=destroy-volume-timeout-FLOC-1853-2&num_builds=10
Merging :dancer:

PR #2081, opened 2015-10-26
[FLOC-3310] Switch to explicit SLEEPING state in convergence loop

itamarst:
The convergence loop needs to be improved such that sleeps in between iterations can be modified (different durations, for exampe) and interrupted (e.g. config updates should perhaps interrupt sleep). A preliminary step is making sleep an explicit state machine step.

http://build.clusterhq.com/boxes-flocker?branch=explicit-convergence-loop-sleep-state-FLOC-3310

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2081)

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -202,7 +202,7 @@ class ConvergenceLoopInputs(Names):
     # iteration of the convergence loop:
     ITERATION_DONE = NamedConstant()
     # Sleep has ended due to hitting its timeout:
-    SLEEP_TIMEOUT = NamedConstant()
+    WAKEUP = NamedConstant()
 
 
 @attributes(["client", "configuration", "state"])
@@ -254,7 +254,7 @@ class ConvergenceLoopOutputs(Names):
     # Start an iteration of the covergence loop:
     CONVERGE = NamedConstant()
     # Schedule timeout for sleep so we don't sleep forever:
-    SCHEDULE_SLEEP_TIMEOUT = NamedConstant()
+    SCHEDULE_WAKEUP = NamedConstant()
     # Cancel the sleep timeout:
     CANCEL_TIMEOUT = NamedConstant()
 
@@ -432,10 +432,10 @@ class ConvergenceLoop(object):
         d.addCallback(lambda _: self.fsm.receive(_IterationDone()))
         d.addActionFinish()
 
-    def output_SCHEDULE_SLEEP_TIMEOUT(self, context):
+    def output_SCHEDULE_WAKEUP(self, context):
         self._sleep_timeout = self.reactor.callLater(
             context.delay_seconds,
-            lambda: self.fsm.receive(ConvergenceLoopInputs.SLEEP_TIMEOUT))
+            lambda: self.fsm.receive(ConvergenceLoopInputs.WAKEUP))
 
     def output_CANCEL_TIMEOUT(self, context):
         self._sleep_timeout.cancel()
@@ -462,7 +462,7 @@ def build_convergence_loop_fsm(reactor, deployer):
         S.CONVERGING, {
             I.STATUS_UPDATE: ([O.STORE_INFO], S.CONVERGING),
             I.STOP: ([], S.CONVERGING_STOPPING),
-            I.ITERATION_DONE: ([O.SCHEDULE_SLEEP_TIMEOUT], S.SLEEPING),
+            I.ITERATION_DONE: ([O.SCHEDULE_WAKEUP], S.SLEEPING),
         })
     table = table.addTransitions(
         S.CONVERGING_STOPPING, {
@@ -471,7 +471,7 @@ def build_convergence_loop_fsm(reactor, deployer):
         })
     table = table.addTransitions(
         S.SLEEPING, {
-            I.SLEEP_TIMEOUT: ([O.CONVERGE], S.CONVERGING),
+            I.WAKEUP: ([O.CONVERGE], S.CONVERGING),
             I.STOP: ([O.CANCEL_TIMEOUT], S.STOPPED),
             # At some point in FLOC-3205 might want to make this interrupt
             # sleep, but at the moment that would increase polling which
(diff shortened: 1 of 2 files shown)


itamarst:
I'd appreciate suggestions on making the machinist interaction less hacky-seeming (I don't enjoy storing the `IDelayedCall`).

tomprince:
I'm not sure if this is any better, but `deferLater` is cancellable. (Avoiding the `IDelayedCall` but not otherwise affecting the code structure).

itamarst:
Hmmmmm cancellable deferreds hmmmmm. I wonder if machinist can build some sort of state machine timeout infrastructure using those?

exarkun:
Thanks.

Only minor comments inline, address as you see fit.  Sorry, no big picture ideas about how to model these parts of the state machine more faithfully as explicit state machine components using machinist.  I think machinist needs some new features to help out here.

The changes proposed here don't seem particularly hackish to me, though.  There are other worse offenses in this state machine already.  The sleep delayed call is nicely encapsulated and has a pretty well-defined lifetime, so ...

Some test failures on buildbot.  I think maybe they're all known issues, maybe even with fixes in-progress?  Please verify that, address inline comments as appropriate, then merge.


---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

<sup>**[flocker/node/_loop.py, line 205 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1deOG7Q6RhUSuD-CRY)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/_loop.py#L205)):</sup>
Total nit ... `WAKEUP`?


---

<sup>**[flocker/node/_loop.py, line 223 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1dWL9MrzhRELQqcNxp)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/_loop.py#L223)):</sup>
The sleeping behavior is really a property of the output and next state associated with this input in the state table.  Do you anticipate making `delay_seconds` a parameter to this input that the instantiator might know how to compute interestingly different values for?  I can sort of see how that might work - perhaps an exponential backoff when discovering no changes, for example - but maybe that logic could just as easily be put into the `SCHEDULE_SLEEP_TIMEOUT` output method?

Or ... something.  I think you're definitely running into a shortcoming of machinist here.  So apart from making machinist better I'm not sure what the best solution would be.  It does feel like a mistake to mix sleep durations together with the iteration done input, though.


---

<sup>**[flocker/node/_loop.py, line 474 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1dYwf3n9teZm0AJ_DF)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/_loop.py#L474)):</sup>
It would be nice to also throw away `_sleep_timeout` in this transition so that we can say it's always `None` outside of the `SLEEPING` state.


---

<sup>**[flocker/node/test/test_loop.py, line 652 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1dbqGbeYSQ2ejlWhYe)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/test/test_loop.py#L652)):</sup>
Pretty hard to read this sentence.  Could it be a couple commas should be added here? :smile:   Similar comments for other `If ...` test method docstrings below.


---

<sup>**[flocker/node/test/test_loop.py, line 659 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1dcCoEwI_ZHPOxz_E0)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/test/test_loop.py#L659)):</sup>
Someone suggested an alternate form that produces more readable failures,

```
self.assertEqual(
    dict(state=loop.state, calls=self.reactor.getDelayedCalls()),
    dict(state=ConvergenceLoopStates.STOPPED, calls=[]),
)
```


---

<sup>**[flocker/node/test/test_loop.py, line 668 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1ddQ43aRmUfJN8Dxnp)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/test/test_loop.py#L668)):</sup>
I don't find that `2` suffix conveys much information.  Why is there another configuration (is it even "another" in this case?  Where's `configuration` and `state`, anyway)?

A name like `changed_configuration` does a better job explaining what's going on.  Or a comment saying what the point of this new thing is or how it's different from some existing thing.

Also I've started to like using pyrsistent transformations to generate new slightly different versions of things ... but maybe that's not applicable in this case since the version of configuration and state these two variables are meant to differ from aren't apparent in this test.


---

<sup>**[flocker/node/test/test_loop.py, line 676 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2081#-K1deBQSu5iY-YDk8sPC)** ([raw file](https://github.com/clusterhq/flocker/blob/cd6e46350017649d900d8f19fe96c2cff45ae2ef/flocker/node/test/test_loop.py#L676)):</sup>
Would be nice not to duplicate magic number here (I know it's already duplicated in a lot of places in this module ...).


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2081)

itamarst:
I left number of seconds to sleep as input to ITERATION_DONE, but changed it to be called SLEEP instead since it's more accurate name. It may be that this is, as you indicate, not quite the right way to do it, but that'll be easier to see once I do followup branch and actually have code that wants to control sleep duration.

wallrj:
I forced a build.

PR #2082, opened 2015-10-27
[FLOC-3276] IProfiledBlockDeviceAPI implementation for EBS driver

myechuri:
First step in enabling Storage Profile functionality on EBS. Allow EBS Flocker driver users to create volumes with gold/silver/bronze profiles.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2082)


flocker/node/agents/ebs.py
```diff
@@ -51,6 +52,108 @@
 NOT_FOUND = u'InvalidVolume.NotFound'
 
 
+class EBSVolumeTypes(Values):
+    """
+    Constants for the different types of volumes that can be created on EBS.
+    These are taken from the documentation at:
+    http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html
+
+    ivar STANDARD: Magnetic
```
itamarst: Missing colon, here and in later ones.

flocker/node/agents/ebs.py
```diff
@@ -51,6 +52,108 @@
...
+    These constants are the ``EBSProfileAttributes`` for the mandatory
+    profiles. Many of the values for these were gotten from the documentation
+    at:
+    http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/EBSVolumeTypes.html
+
+    :ivar GOLD: The high performing Provisioned IOPS disks.
+    :ivar SILVER: The medium performing SSD disks.
+    :ivar BRONZE: The cheap magnetic disks.
+    """
+    # Values here are gotten from:
```
itamarst: Stray comment.

flocker/node/agents/ebs.py
```diff
@@ -682,13 +786,38 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
+            requested_volume = self.connection.create_volume(
+                size=requested_size,
+                zone=self.zone,
+                volume_type=volume_type,
+                iops=iops)
+        except EC2ResponseError as e:
+            CREATE_VOLUME_FAILURE(dataset_id=dataset_id,
+                                  aws_code=e.code,
+                                  aws_message=e.message).write()
+            volume_type, iops = _volume_type_and_iops_for_profile_name(
```
itamarst: Why are you retrying? What specific error do you expect? Could you catch it explicitly? If not, at least document why.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+
+    def test_create_largest_volume_gold_profile(self):
+        """
+        Largest volume with ``gold`` profile.
+        """
+        raise SkipTest(u'This test is expensive (money).')
+        self._assert_create_volume_with_mandatory_profile(
+            MandatoryProfiles.GOLD, size_GiB=16*1024)
+
+    def test_create_too_large_volume_gold_profile(self):
```
itamarst: Not sure this test tells us much.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+        """
+        self._assert_create_volume_with_mandatory_profile(
+            MandatoryProfiles.GOLD, created_profile=MandatoryProfiles.DEFAULT,
+            size_GiB=1)
+
+    def test_create_largest_volume_gold_profile(self):
+        """
+        Largest volume with ``gold`` profile.
+        """
+        raise SkipTest(u'This test is expensive (money).')
```
itamarst: I assume we delete volumes as soon as test is over; is it really that expensive?

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+    def test_create_too_large_volume_silver_profile(self):
+        """
+        Too large volume (> 16TiB) for ``silver`` profile.
+        """
+        self.assertRaises(EC2ResponseError,
+                          self._assert_create_volume_with_mandatory_profile,
+                          MandatoryProfiles.SILVER,
+                          size_GiB=1024*1024)
+
+    def test_create_largest_volume_silver_profile(self):
```
itamarst: Maybe just delete this test.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+        self._assert_create_volume_with_mandatory_profile(
+            MandatoryProfiles.BRONZE)
+
+    def test_create_too_large_volume_bronze_profile(self):
+        """
+        Too large volume for ``bronze`` profile.
+        """
+        self.assertRaises(EC2ResponseError,
+                          self._assert_create_volume_with_mandatory_profile,
+                          MandatoryProfiles.SILVER,
```
itamarst: That's silver, not bronze. And again not sure this test tells us anything useful.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+        """
+        raise SkipTest(u'This test is expensive (money).')
+        self._assert_create_volume_with_mandatory_profile(
+            MandatoryProfiles.BRONZE, size_GiB=1024)
+
+    def _assert_create_volume_with_mandatory_profile(self, profile,
+                                                     created_profile=None,
+                                                     size_GiB=4):
+        """
+        Volume created with ``gold`` profile has the attributes
```
itamarst: given profile, not gold profile.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
...
+        expected from the profile.
+
+        :param ValueConstant profile: Name of profile to use for creation.
```
itamarst: Document other parameters.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,132 @@ def test_next_device_in_use(self):
                                        {u"/dev/sdf"})
         self.assertEqual(result, u"/dev/sdg")
 
+    def test_create_volume_gold_profile(self):
```
itamarst: Most of these tests seem to just duplicate the standard interface tests. The exception is the fallback to DEFAULT on errors, which should be tested so it has coverage.

flocker/node/agents/ebs.py
itamarst: Reading tests suggests this is here to cover cases where size doesn't match profile requirements. I don't think we should be retrying in other cases. Possibly it's better to not retry at all, and just document this restriction for now? Would users prefer outright failure or would they prefer wrong profile? I suspect the former but maybe I'm wrong.

itamarst:
Excited to see this feature happening.

Some minor cleanups, some suggestions about reducing tests, and some questions about error handling. Address and resubmit.

flocker/node/agents/functional/test_ebs.py
sarum90: $4/ a run was my estimate.

flocker/node/agents/ebs.py
myechuri: Sorry, fixed.

flocker/node/agents/ebs.py

flocker/node/agents/ebs.py
myechuri: > Why are you retrying? Reading tests suggests this is here to cover cases where size doesn't match profile requirements. 

Sorry for the poor naming of ``test_create_too_small_volume_gold_profile``. Retrying in case volume creation failed because profile attributes corresponding to the requested profile cannot be satisfied for the volume that is being created.

As an example, ``test_create_too_small_volume_gold_profile`` attempts to create volume with:
size: 1 GiB
profile (``GOLD``):
volume_type: ``io1``
iops:size ratio: 30:1 (thats the best iops ratio for ``io1`` volume type (not documented, but noticed during testing of hack day project)).

With these parameters, EBS volume creation fails with the following error:

```
<Response><Errors><Error><Code>InvalidParameterValue</Code><Message>Volume iops of 30 is too low; minimum is 100.</Message></Error></Errors><RequestID>3f9facc0-7e77-4b7c-9655-69e6db504f7f</RequestID></Response>
```

The retry is saying that we have tried our best to honor profile request during volume creation. If we failed, we will make one more attempt to create your volume with default profile. Profile assignment can be a best effort attempt. If we create the profile and move on, profile compliance module (to be built later on) can detect and act on profile violations at run time (similar to https://pubs.vmware.com/vsphere-51/index.jsp?topic=%2Fcom.vmware.vsphere.storage.doc%2FGUID-46345561-C7ED-4296-91AE-8F6C5D8FF552.html).

Apologies for not documenting the reason for retry. Will add it.

myechuri: > profile compliance module

https://clusterhq.atlassian.net/browse/FLOC-3275

flocker/node/agents/functional/test_ebs.py
myechuri: The aim was to try to create a volume so large that retry failed volume creation with downgraded profile does not help make progress. You are right in pointing out little value in these set of tests. Combined all three test_create_largest_volume_{gold,silver,bronze}_profile into one test. Please let me know what you think of the new version. Thanks!

flocker/node/agents/functional/test_ebs.py
myechuri: Deleted (combined with gold/siver cases, as mentioned above).

flocker/node/agents/functional/test_ebs.py
myechuri: Addressed above. Thanks!

flocker/node/agents/functional/test_ebs.py
myechuri: Sorry, fixed.

flocker/node/agents/functional/test_ebs.py
myechuri: Fixed.

flocker/node/agents/functional/test_ebs.py
myechuri: > Most of these tests seem to just duplicate the standard interface tests.

Slight variation: ``test_profile_respected`` tests that ``create_volume_with_profile`` with given profile succeeds in creating a volume. The EBS tests actually check that profile attributes reported by EBS backend match profile attributes published for ``MandatoryProfiles`` by EBS driver. Please let me know if that sounds reasonable? Thanks!

myechuri:
Thanks a lot for the quick review, @itamarst !

flocker/node/agents/functional/test_ebs.py
itamarst: I'm not sure the "Create largest" tests add much value, so I'd suggest just deleting them.

flocker/node/agents/functional/test_ebs.py
itamarst: Ah, yes. OK yeah that makes sense.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,128 @@ def test_next_device_in_use(self):
...
+        Too large volume (> 16TiB) for ``silver`` profile.
+        """
+        self.assertRaises(EC2ResponseError,
+                          self._assert_create_volume_with_mandatory_profile,
+                          MandatoryProfiles.SILVER,
+                          size_GiB=1024*1024)
+
+    def test_create_largest_volume_silver_profile(self):
+        """
+        Largest volume with ``silver`` profile.
```
itamarst: Again, "create largest" seems unuseful test, unless you have specific motivation.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,128 @@ def test_next_device_in_use(self):
...
+            profile_name=profile.value)
+
+        cannonical_profile = MandatoryProfiles.lookupByValue(
+            created_profile.value)
+        A = EBSMandatoryProfileAttributes.lookupByName(
+            cannonical_profile.name).value
+        ebs_volume = self.api._get_ebs_volume(volume1.blockdevice_id)
+        self.assertEqual(ebs_volume.type, A.volume_type.value)
+        requested_iops = A.requested_iops(ebs_volume.size)
+        if requested_iops is not None:
```
itamarst: Conditional assertions are problematic: it's too easy to end up never calling the branch and not noticing. If `iops` is some arbitrary value then you could do:

```
self.assertEqual(ebs_volume.iops if requested_iops is not None else None, requested_iops)
```

Which is slightly better. Or maybe:

```
if requested_iops is None:
     requested_iops = ebs_volume.iops  # don't care about value
self.assertEqual(ebs_volume.iops, requested_iops)
```

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2310,7 +2310,12 @@ def test_profile_respected(self):
         mandatory profiles.
         """
         for profile in (c.value for c in MandatoryProfiles.iterconstants()):
-            self.api.create_volume_with_profile(profile)
+            size = GiB(1).to_Byte().value
+            dataset_id = uuid4()
+            self.addCleanup(detach_destroy_volumes, self.api)
+            self.api.create_volume_with_profile(dataset_id=dataset_id,
+                                                size=size,
+                                                profile_name=profile)
```
itamarst: It would be good if the standard interface tests checked that something created as GOLD is returned as GOLD; that after all is the point of a backend saying they implement the interface. Possibly the test class should be constructed with constants indicating size to use per profile so you don't have to hard code a value that works for all profiles on all backends.

flocker/node/agents/ebs.py
```diff
@@ -682,13 +785,43 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
+                size=requested_size,
+                zone=self.zone,
+                volume_type=volume_type,
+                iops=iops)
+        except EC2ResponseError as e:
+            # If we failed to create a volume with attributes complying
+            # with requested profile, make a second attempt at volume
+            # creation with default profile.
+            # Compliance violation of the volume's requested profile
+            # will be detected and remediated in a future release (FLOC-3275).
```
itamarst: This will result in volumes being created with wrong profile due to unrelated intermittent errors, e.g. rate limit exceeded. So it would be good to at least check for a specific EC2 error indicating "couldn't create with these options" rather than doing this unconditionally for all errors. That could be done in separate issue if you want.

itamarst:
1. Delete a few tests.
2. Create an additional interface-specific test(s).
3. One optional error handling task that should be done either here or in followup.

Address and resubmit, and I'll try to do next review as soon as it's up so you're not blocked.

flocker/node/agents/test/test_blockdevice.py
sarum90: This is definitely a more subtle problem that it originally appears. We should probably do a hangout tomorrow to discuss some of the finer points.

Specifically:

- This suggests the API should have a "get profile for this blockdevice_id"
 -- In situations where desired profile can differ from actual profile (due to availability or changing profiles) what should this return?
 -- What should implementations look like for dumb backends like EBS, and smart backends like CIO?
 -- If profiles are more of a suggestion, does that complicate any answers to the former?

Whenever I've wandered down that thought path I've always landed on some things that seem open-ended. I feel if we put it in the interface and it's open ended, we are going to get storage parters implementing multiple interpretations, and that might be a hard place to come out of. So for now, we've sort of been biasing toward, "just put it in EBS and 'know' that its EBS in tests to try out the current API" rather than putting it in the interface. Hopefully later we can ask other driver writers about their experiences and discuss some of the above questions to have a good approach before we extend the interface to be more featureful.

Anyway just something to think about. Hopefully tomorrow we can find some time to do a hangout. It's possible you, Madhuri, or Ryan has more concrete answers to my open questions above and we can make the interface a little larger.

sarum90: Also that said... we should probably put more of this logic in the backend... even if we don't add it to the interface yet.

flocker/node/agents/functional/test_ebs.py
myechuri: Removed.

flocker/node/agents/functional/test_ebs.py
myechuri: Removed all "create largest" tests (gold/silver/bronze).

flocker/node/agents/functional/test_ebs.py
myechuri: Thanks for pointing out!

i like this better:
```
self.assertEqual(ebs_volume.iops if requested_iops is not None else None, requested_iops)
```
because requested_iops are not meant to specified for certain poor profiles. Changed. Thanks!

flocker/node/agents/test/test_blockdevice.py
myechuri: Thanks for your thoughts, @itamarst @sarum90 ! 

Made a few enhancements to interface based on your feedback and expected usage of profiles. Will schedule a meeting thursday morning to hear your thoughts on the new revision. Thanks!

flocker/node/agents/ebs.py
myechuri: The closest narrowing possible is checking for ``e.code == u'InvalidParameterValue'``. Added the check. Thanks for suggesting!

flocker/node/agents/functional/test_ebs.py
```diff
@@ -182,6 +188,106 @@ def test_next_device_in_use(self):
...
+        cannonical_profile = MandatoryProfiles.lookupByValue(
+            created_profile.value)
+        A = EBSMandatoryProfileAttributes.lookupByName(
+            cannonical_profile.name).value
+        ebs_volume = self.api._get_ebs_volume(volume1.blockdevice_id)
+        self.assertEqual(ebs_volume.type, A.volume_type.value)
+        requested_iops = A.requested_iops(ebs_volume.size)
+        self.assertEqual(ebs_volume.iops if requested_iops is not None
+                         else None, requested_iops)
+        if requested_iops is not None:
```
itamarst: I think there was previous review comment that didn't get addressed here: conditional asserts are a bad idea since they can easily just never happen.

itamarst: You could instead do:

```
if requested_iops is None:
    requested_iops = ebs_volume.iops
self.assertEqual(ebs_volume.iops, requested_iops)
```

itamarst:
1. A couple of minor comments.
2. Buildbot is red for relevant storage driver tests. Might just be coincidence, but make sure it's all green.

Then merge. Thanks for all the work on this in between PoC meetings!

flocker/node/agents/functional/test_ebs.py
myechuri: Sorry, had addressed the review comment, but missed removing the original line (``if requested_iops is not None``) while doing partial revert of last two commits. Thanks for pointing out - fixed.

flocker/node/agents/blockdevice.py
```diff
@@ -956,7 +956,7 @@ class MandatoryProfiles(Values):
     GOLD = ValueConstant(u'gold')
     SILVER = ValueConstant(u'silver')
     BRONZE = ValueConstant(u'bronze')
-    DEFAULT = SILVER
+    DEFAULT = ValueConstant(BRONZE.value)
```
wallnerryan: mentioned this in original profiled implementation, but gold, sillver, bronze suggests some type of limit. Where as ```default``` typically zeros out values and allows "unlimited" style for default profile. Suggesting we might want to keep default actually default, instead of assigning it to gold, silver or bronze.

myechuri: Thanks for the note, @wallnerryan ! In the interest of moving ahead, will address ``default`` handling in a follow up branch. Please let me know if that does not sound reasonable. Thanks!

sarum90:
Finally got a green run of the tests 3 days later, so I'll be merging this.

PR #2087, opened 2015-10-27
[FLOC-3318] Wakeup convergence loop if changes matter

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-3318
http://build.clusterhq.com/boxes-flocker?branch=wakeup-loop-if-changes-matter-FLOC-3318

1. Update sequentially/in_parallel so they indicate a no-op if they're not going to take any action.
2. Whenever the convergence loop receives an update from the control service while sleeping, it calculates what changes need to be made given that update, without doing any local discovery. If the result is not a no-op this means we should wake up, since the update apparently required some action on our part. Since `calculate_changes` is supposed to be purely in-memory operation this additional check should have limited no impact on system load, e.g. no cloud API calls involved.
3. Demonstrate that individual deployers return `NoOp` when no changes are needed.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2087)

### DIFF
--- a/flocker/node/__init__.py
+++ b/flocker/node/__init__.py
@@ -5,7 +5,7 @@ Local node manager for Flocker.
 """
 
 from ._change import (
-    IStateChange, in_parallel, sequentially, run_state_change
+    IStateChange, in_parallel, sequentially, run_state_change, NoOp,
 )
 
 from ._deploy import (
@@ -21,6 +21,7 @@ from .script import BackendDescription, DeployerType
 
 __all__ = [
     'IDeployer', 'ILocalState', 'NodeLocalState', 'IStateChange',
+    'NoOp',
     'P2PManifestationDeployer',
     'ApplicationNodeDeployer',
     'run_state_change', 'in_parallel', 'sequentially',
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -121,6 +121,9 @@ def in_parallel(changes):
     The order in which execution of the changes is started is unspecified.
     Comparison of the resulting object disregards the ordering of the changes.
     """
+    changes = [c for c in changes if c != NoOp()]
+    if not changes:
+        return NoOp()
     return _InParallel(changes=changes)
 
 
@@ -142,4 +145,13 @@ def sequentially(changes):
 
     Failures in earlier changes stop later changes.
     """
+    changes = [c for c in changes if c != NoOp()]
+    if not changes:
+        return NoOp()
     return _Sequentially(changes=changes)
+
+
+@implementer(IStateChange):
+class NoOp(object):
+    def run(self, deployer):
+        return succeed(None)
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -16,11 +16,12 @@ changes.
 
 from zope.interface import Interface, Attribute, implementer
 
-from pyrsistent import PVector, PRecord, pvector, field
+from pyrsistent import PVector, PRecord, pvector, field, PClass
 
 from twisted.internet.defer import maybeDeferred, succeed
 
 from eliot.twisted import DeferredContext
+from eliot import ActionType
 
 from ..common import gather_deferreds
 
@@ -121,8 +122,7 @@ def in_parallel(changes):
     The order in which execution of the changes is started is unspecified.
     Comparison of the resulting object disregards the ordering of the changes.
     """
-    changes = [c for c in changes if c != NoOp()]
-    if not changes:
+    if not any(c for c in changes if c != NoOp()):
         return NoOp()
     return _InParallel(changes=changes)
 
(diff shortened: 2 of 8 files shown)


wallrj:
Thanks @itamarst 

After staring at the code for a while and remembering Machinist I think this makes sense:

* When the loop is SLEEPING it can be woken by a STATE_UPDATE from the control service  (via AMP connection) and maybe moves on to a full CONVERGE if the last discovered local state is different than the desired configuration for this node.
* And (for now) we schedule a delayed call to WAKEUP the loop every 1s which leads to a `discover_state` and a `calculate_actions` every 1s. The current behaviour.

In the follow up branch the dataset agent will only wake up and `discover_state` every 60s while ZFS and container agents will continue to discover_state every 1s.

Please proceed.

sarum90:
Good stuff. Mostly nits to address and merge as you see fit.

One larger question on the legitimacy of calling fsm.receive inside an output function. If you choose to do any major refactors around that perhaps re-submit for review, but otherwise looks solid.

Sorry for the slow review, had a lot of code grepping to do to figure out how machinist worked.


---

Review status: 0 of 8 files reviewed at latest revision, 7 unresolved discussions, some commit checks pending.

---

<sup>**[flocker/node/_change.py, line 125 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1jppC9Epeh5ccvvAOI)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/_change.py#L125)):</sup>
:grey_question: I think the following reads better, and is equivalent:

```
if all(c == NoOp() for c in changes):
  return NoOp()
```

Feel free to keep as is if you have your reasons.


---

<sup>**[flocker/node/_change.py, line 149 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1jrQ9Mavk0wJu3Kfef)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/_change.py#L149)):</sup>
:grey_question: ditto


---

<sup>**[flocker/node/_loop.py, line 358 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1k31qgLPwRKg4qZCWu)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/_loop.py#L358)):</sup>
:grey_question: Passing concern.  self.fsm.receive in an output seems scary to me.

If I'm reading machinist correctly, you couldn't add a second output that wants to call  self.fsm.receive(ConvergenceLoopInputs.WAKEUP) to the outputs of   (S.SLEEPING, I.STATUS_UPDATE) because then you would be giving it a WAKEUP to an FSM that is already in the CONVERGING state.

Alternatively it seems like this ouptput both assumes it knows what the current state is _and_ changes it. Which seems incompatible with the fact that we can chain outputs and expect them to be called in series.

Maybe it's a non-issue, and in this case it's safe as this is the final output of the transition... but it seems like this is setting us up on something to trip over in the future. I don't feel qualified to assess how large of a tripwire this is because I don't have extensive experience working with this style of FSM code, but it definitely looks fishy.


---

<sup>**[flocker/node/_loop.py, line 514 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1jw1bJmaMBjpCdvkOE)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/_loop.py#L514)):</sup>
I might be wrong, but it looks to me like this PR precisely addresses this comment. I'm guessing we get to remove this now :grinning: 


---

<sup>**[flocker/node/test/test_change.py, line 183 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1jt5-bW_4E29U0Bb-I)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/test/test_change.py#L183)):</sup>
:grey_question:  Perhaps mention overriding because the base class implements this test and it will not work for this IStateChange in the docstring. You also could let test_equality fall through to the base class test, but I recognize that is just a stylistic decision.

Also, this required overriding is unfortunate. Would there be a cleaner way to fix this? optional parameter to make_istate_change_tests?

Fine as is, but certainly non-obvious.


---

<sup>**[flocker/node/test/test_loop.py, line 614 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1kBQBRUrP5H5ryN1yw)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/test/test_loop.py#L614)):</sup>
:grey_question: ControllableAction has state that changes, slight preference for using None as the default argument and instantiating on every call. Not an issue for the usage of ControllableAction here, but seems like that'd be better practice for more future-proof code.


---

<sup>**[flocker/node/test/test_loop.py, line 684 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2087#-K1kJHQeZBPQIw3CNOEE)** ([raw file](https://github.com/clusterhq/flocker/blob/13d43790cfc990c9d3451a230a85e0b8b56fe4df/flocker/node/test/test_loop.py#L684)):</sup>
:grey_question: Do you want any assertions around the number of calls to discover_state around this resolution?

You specifically call it out in a comment:
> Notably we do **not** do a discovery of local state when an update is received while sleeping

As I was still getting used to how machinist works I wanted to verify that the above comment was true and hacked together the following branch to double check:

https://github.com/ClusterHQ/flocker/compare/wakeup-loop-if-changes-matter-mewert-add~1...wakeup-loop-if-changes-matter-mewert-add?expand=1

It occurred to me that it might also be a useful expansion of this test. I did not see it covered in other tests, and seems worth it since extra discover_state calls mean extra API calls. Feel free to incorporate it or not at your discretion.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2087)

itamarst:
Addressed all but the input-in-output-method issue, which I don't see any solution to.

PR #2088, opened 2015-10-27
[FLOC-3326] Convergence sleeps appropriately

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-3326
http://build.clusterhq.com/boxes-flocker?branch=convergence-sleeps-appropriately-FLOC-3326

After each iteration of convergence loop we decide how long to sleep:
1. If we've done nothing this iteration, sleep for max interval set by the deployer. (If configuration/cluster state changes in way that requires action the loop will wake up thanks to work in previou branches.)
2. If we've done something, sleep for short period of time. The theory is that many configuration changes (e.g. "move a dataset here") require multiple steps so it's worth doing another iteration soon after.

Previously pattern of interactions was:
* Iteration.
* Sleep 1.0 seconds
* Iteration
* Sleep 1.0 seconds
* ... etc.

New pattern will be like this for block device deployer:
* Iteration. Did some stuff.
* Sleep 0.1 seconds
* Iteration. Did some stuff.
* Sleep 0.1 seconds
* Iteration. Did nothing.
* Sleep 60 seconds (with some noise)

So while doing work we iterate rather more, but then we sleep for much longer after the work is done.

0.1 was picked out of a hat.

I used Marcus' script (https://clusterhq.atlassian.net/browse/FLOC-3178) to get an approximation of EBS API calls. This number isn't an exact number of calls, but should strongly correlate with it. Before this set of convergence loop changes the number was 1660 (as measured by Marcus). For this branch the number is 330. So that seems like a success.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2088)


wallrj:
Thanks @itamarst 

Would be good to grep the acceptance test logs for API call count before and after this change and include that when you put it up for review.

Please proceed.

flocker/node/_deploy.py
```diff
@@ -96,9 +96,17 @@ class IDeployer(Interface):
...
+    :ivar float poll_interval: Number of seconds to delay between
+        iterations of convergence loop that call ``discover_state()``, to
+        reduce impact of polling external resources. The actual delay may
+        be smaller if the convergence loop decides more work is necessary
+        in order to converge.
     """
     node_uuid = Attribute("The UUID of thise node, a ``UUID`` instance.")
     hostname = Attribute("The public IP address of this node.")
+    poll_interval = Attribute(
+        "Time to sleep between convergence loop iterations.")
```
exarkun: Seems silly to document all these attributes twice :question:

flocker/node/_deploy.py
```diff
@@ -96,9 +96,17 @@ class IDeployer(Interface):
     :ivar UUID node_uuid: The UUID of the node this deployer is running.
     :ivar unicode hostname: The hostname (really, IP) of the node this
         deployer is managing.
+
+    :ivar float poll_interval: Number of seconds to delay between
```
exarkun: datetime.timedelta is probably a better data type than float :question:

flocker/node/_loop.py
```diff
@@ -429,6 +437,17 @@ def got_local_state(local_state):
             action = self.deployer.calculate_changes(
                 self.configuration, self.cluster_state, local_state
             )
+            if action == NoOp():
```
exarkun: I think it might be good to make it more explicit in the interface that if you *don't* compute some non-NoOp action then the system is going to sleep for a while.  ie, you may not want an implementation that passively waits for some condition to be discovered and then acts because the delay between iterations in this case may be much greater than you'd like. :question:

flocker/node/test/test_loop.py
```diff
@@ -724,6 +726,43 @@ def test_status_update_while_sleeping_no_discovery(self):
...
 
+    def test_longer_sleep_when_converged(self):
+        """
+        When a convergence loop results in a ``NoOp`` the sleep is based on
+        that configured on the ``Deployer``.
+        """
+        loop = self.convergence_iteration(initial_action=NoOp(),
+                                          later_actions=[NoOp(), NoOp()])
+
+        # Another update received while sleeping:
```
exarkun: Another?

flocker/node/test/test_loop.py
```diff
@@ -724,6 +726,43 @@ def test_status_update_while_sleeping_no_discovery(self):
...
+            client=self.make_amp_client([self.local_state]),
+            configuration=self.configuration, state=self.cluster_state))
+
+        # No additional discovery done due to update:
+        pre_sleep = len(self.deployer.local_states)
+
+        # Sleep until right before lowest possible random sleep interval:
+        self.reactor.advance(
+            self.deployer.poll_interval * (1 - ConvergenceLoop._JITTER_RANGE)
+            - 0.05)
```
exarkun: 0.05?

flocker/node/test/test_loop.py
```diff
@@ -724,6 +726,43 @@ def test_status_update_while_sleeping_no_discovery(self):
...
+        mid_sleep = len(self.deployer.local_states)
+        # Sleep until right after highest possible sleep interval
+        self.reactor.advance(
+            self.deployer.poll_interval + (2 * ConvergenceLoop._JITTER_RANGE)
+            + 0.05
+            # Extra bit of sleep to ensure we don't break on floating
+            # point rounding errors:
+            + 0.00001)
```
exarkun: Not necessary for the mid case?

flocker/node/test/test_loop.py
```diff
@@ -350,7 +350,7 @@ def test_convergence_done_unchanged_notify(self):
         loop = build_convergence_loop_fsm(reactor, deployer)
         loop.receive(_ClientStatusUpdate(
             client=client, configuration=configuration, state=state))
-        reactor.advance(1.0)
+        reactor.advance(ConvergenceLoop._UNCONVERGED_DELAY)
```
exarkun: If the sleep values were parameters to `build_convergence_loop_fsm` then the tests could avoid using this private attribute. :question:

flocker/node/test/test_loop.py
```diff
@@ -724,6 +726,43 @@ def test_status_update_while_sleeping_no_discovery(self):
             remaining_discover_calls - len(self.deployer.local_states),
             0)
 
+    def test_longer_sleep_when_converged(self):
+        """
+        When a convergence loop results in a ``NoOp`` the sleep is based on
+        that configured on the ``Deployer``.
+        """
```
exarkun: This test gets pretty deep into specific randomization implementation details of the loop which aren't really that important.  Some abstraction over those details would help the test be clearer and less fragile.  It might also help write a test that exercises the behavior we actually care about from the randomization (that the sleep interval varies to avoid load spikes due to herding) which don't appear to be tested currently.

:question:

exarkun:
Thanks.  Some optional inline comments.  Lots of build errors, unfortunately.  Merge when you're satisfied with the code and you get buildbot into the right shape.

flocker/node/test/test_loop.py
itamarst: Nothing leaped to mind at the time. I'll think about it more. Any suggestions would be welcome.

flocker/node/test/test_loop.py
itamarst: Good point. Not feeling motivated enough to fix this though since it doesn't leak out into public API.

flocker/node/test/test_loop.py
itamarst: Hm, maybe a `_JitterySleep` input type.

flocker/node/test/test_loop.py
itamarst: `convergence_iteration()` involves an update, so this is the second one. But that is a little unclear, I'll change it.

flocker/node/test/test_loop.py
itamarst: No, with mid case it's about not going too far, with post case you want to make sure you go far enough.

exarkun: That still sounds to me like it would at least make sense to subtract some tiny value.  But I don't know what the 0.05 is for, maybe that already accounts for this.

PR #2091, opened 2015-10-28
[FLOC-2603] Retry downloads in brew

thepeak99:
Helps mitigate: https://clusterhq.atlassian.net/browse/FLOC-2603 

Also adds a generic function `retry_effect_with_timeout` that can be used in flaky tests to mitigate network (or other) issues.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2091)


flocker/testtools/__init__.py
```diff
@@ -299,6 +302,28 @@ def loop(failure):
     return d
 
 
+def retry_effect_with_timeout(effect, timeout):
+    """
+    If ``effect`` fails, retry it until ``timeout`` expires.
+
+    :param Effect effect: The Effect to retry.
+    :param int timeout: Timeout in seconds.
```
jml: Please document the return value, e.g. `:return: An Effect that does what ``effect`` does, but retrying.`

flocker/testtools/__init__.py
```diff
@@ -299,6 +302,28 @@ def loop(failure):
     return d
 
 
+def retry_effect_with_timeout(effect, timeout):
+    """
+    If ``effect`` fails, retry it until ``timeout`` expires.
```
jml: Please add documentation noting that it does exponential backoff.

flocker/testtools/__init__.py
```diff
@@ -299,6 +302,28 @@ def loop(failure):
...
+
+    :param Effect effect: The Effect to retry.
+    :param int timeout: Timeout in seconds.
+    """
+
+    end_time = time.time() + timeout
+
+    def ret(e):
```
jml: `ret` here stands for "retry"? Please rename it to `retry` to avoid confusion (`ret` is often used to mean "return value").

flocker/testtools/__init__.py
```diff
@@ -299,6 +302,28 @@ def loop(failure):
...
+        if time.time() > end_time:
+            return Effect(Constant(False))
+        else:
+            time.sleep(ret.wait_secs)
+            ret.wait_secs *= 2
+            return Effect(Constant(True))
+    ret.wait_secs = 1
```
jml: Please make the initial retry period a documented parameter. It's OK to make it default to 1.

flocker/testtools/__init__.py
```diff
@@ -299,6 +302,28 @@ def loop(failure):
...
+    :param int timeout: Timeout in seconds.
+    """
+
+    end_time = time.time() + timeout
+
+    def ret(e):
+        if time.time() > end_time:
+            return Effect(Constant(False))
+        else:
+            time.sleep(ret.wait_secs)
```
jml: I'm not super familiar with effect, but I think that rather than explicitly sleeping here, we should be using `effect.Delay` (see http://effect.readthedocs.org/en/latest/apidocs.html?highlight=time#effect.Delay) and presumably sequencing it somehow (I don't know and the docs don't say, but I'd start by looking for other examples. If you'd like we can pair on this).

Once you've done _that_, then it becomes feasible to write a few quick unit tests for this, which you really should.

jml:
Thanks!

It's great to see this code made more robust, and the addition of a new helper. There's some work that's required, some of which might be fiddly. I'm happy to help with it. Please re-submit for review when you're done.

flocker/common/__init__.py
```diff
@@ -16,4 +17,5 @@
 from ._thread import auto_threaded
 from ._interface import interface_decorator
 from ._net import get_all_ips, ipaddress_from_string
-from ._retry import loop_until, poll_until, retry_failure
+from ._retry import loop_until, poll_until, retry_failure, \
+    retry_effect_with_timeout
```
jml: Use parens, rather than backslash. e.g. 

```python
from ._retry import (loop_until, poll_until, retry_failure,
    retry_effect_with_timeout)
```

flocker/common/_retry.py
```diff
@@ -175,3 +178,39 @@ def poll_until(predicate, steps, sleep=None):
...
+
+
+def retry_effect_with_timeout(effect, timeout, retry_wait=1, exp_backoff=True,
+                              time=time.time):
+    """
+    If ``effect`` fails, retry it until ``timeout`` expires.
+
+    To avoid excessive retrying, this function by default (it can be disabled)
+    uses the exponential backoff algorithm, waiting twice the time between
+    each retry.
```
jml: Thanks for the helpful comment. Minor English points:

* Drop "(it can be disabled)", since it's implied by "by default"
* s/twice/double/ (I think twice is technically correct, but it's weird)
* move "by default" to after "algorithm"

flocker/common/_retry.py
```diff
@@ -175,3 +178,39 @@ def poll_until(predicate, steps, sleep=None):
...
+
+    :param Effect effect: The Effect to retry.
+    :param int timeout: Keep retrying until timeout.
+    :param int retry_wait: The wait time between retries
+    :param bool exp_timeout: Whether we should use exponential backoff
```
jml: s/exp_timeout/exp_backoff/

flocker/common/_retry.py
```diff
@@ -175,3 +178,39 @@ def poll_until(predicate, steps, sleep=None):
...
+        else:
+            effect = Effect(Delay(should_retry.wait_secs)).on(
+                success=lambda x: Effect(Constant(True)))
+
+            if exp_backoff:
+                should_retry.wait_secs *= 2
+
+            return effect
+
+    should_retry.wait_secs = retry_wait
```
jml: Storing this on `should_retry` (i.e. assigning an attribute to a function) is a bit weird, but OK. No change required.

flocker/common/test/test_retry.py
```diff
@@ -386,3 +396,102 @@ def test_polls_one_last_time(self):
...
+
+        retrier = retry_effect_with_timeout(Effect(Func(tester)), 1, 5, False)
+        result = perform_sequence(seq, retrier)
+        self.assertEqual(result, 1 / 1)
+
+    def test_timeout(self):
+        """
+        If the timeout expires, the effect should fail with the same exception.
+        """
+
```
jml: Nit: delete this blank line.

jml:
Thanks! 

Only a few minor style things to clean up.

In a later patch, we may want to change this to have a more similar interface to the other retry functions.

jml:
Please make the suggested changes and merge.

PR #2094, opened 2015-10-28
[FLOC-3271] Acceptance test for Flocker for storage profiles.

sarum90:
This is the codified acceptance test for Flocker storage profiles basic support on EBS

It verifies that if a volume is created with the specific metadata that forces a profile for the volume, then the volume that is created has the correct EBS type for the given profile.

In the interest of keeping the acceptance tests lean, I've only added a test for gold rather than also testing for silver and bronze. I'm open to feedback on that decision. I figured this was enough to show the plumbing was there even if we aren't exhaustively testing every length of pipe at this level.

In order to validate the volume's performance characteristics from the acceptance test, this adds a backdoor to get the backend configuration from the process that sets up the acceptance test rig.  It then uses those credentials to verify the type of the volume created.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2094)

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -209,13 +209,10 @@ def get_backend_api(test_case, cluster_id):
             'in order to verify construction. Please set '
             'FLOCKER_ACCEPTANCE_TEST_VOLUME_BACKEND_CONFIG to a yaml filepath '
             'with the dataset configuration.')
-    print backend_config_filename
     backend_config_filepath = FilePath(backend_config_filename)
     full_backend_config = yaml.safe_load(
         backend_config_filepath.getContent())
-    print full_backend_config
     backend_config = full_backend_config.get(backend_name)
-    print backend_config
     if 'backend' in backend_config:
         backend_config.pop('backend')
     return aws_from_configuration(cluster_id=cluster_id, **backend_config)
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -7,11 +7,12 @@ from functools import wraps
 from json import dumps
 from os import environ, close
 from unittest import SkipTest, skipUnless
-from uuid import uuid4
+from uuid import uuid4, UUID
 from socket import socket
 from contextlib import closing
 from tempfile import mkstemp
 
+import yaml
 import json
 import ssl
 
@@ -39,9 +40,10 @@ from ..common import gather_deferreds
 from ..common.runner import download_file
 
 from ..control.httpapi import REST_API_PORT
-from ..ca import treq_with_authentication
+from ..ca import treq_with_authentication, UserCredential
 from ..testtools import loop_until, random_name
 from ..apiclient import FlockerClient, DatasetState
+from ..node.agents.ebs import aws_from_configuration
 
 try:
     from pymongo import MongoClient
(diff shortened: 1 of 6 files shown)


sarum90:
I ran the tests once before adding the SkipTest (http://build.clusterhq.com/boxes-flocker?branch=flocker-storage-profiles-acceptance-FLOC-3271) if we're lucky they'll fail for all the right reasons, specifically that the "standard" volume type is not "io1" the type expected for gold (they did on my local runs). Then I'll re-kick them and we should see this new test being skipped. I'll take care of that in address_and_merge. I don't think that should block review though. If you have a different opinion let me know and I'll just sit on this until that testing is complete.

flocker/acceptance/testtools.py
```diff
@@ -789,6 +839,15 @@ def decorator(test_method):
         """
         def call_test_method_with_cluster(cluster, test_case, args, kwargs):
             kwargs['cluster'] = cluster
+            if required_backend:
+                backend_type = get_dataset_backend(test_case)
+                if backend_type != required_backend:
+                    raise SkipTest(
+                        'This test requires backend type {} but is being run '
+                        'on {}.'.format(backend_type.name,
```
sarum90: These are backwards: http://build.clusterhq.com/builders/flocker%2Facceptance%2Frackspace%2Fcentos-7%2Frackspace/builds/3001/steps/trial/logs/stdio

I'll fix that when I address other review comments.

flocker/acceptance/testtools.py
```diff
@@ -866,10 +928,12 @@ def create_dataset(test_case, cluster, maximum_size=None, dataset_id=None):
         maximum_size = get_default_volume_size()
     if dataset_id is None:
         dataset_id = uuid4()
-
+    if metadata is None:
+        metadata = {}
+    metadata[u"name"] = metadata.get(u"name", u"my_volume")
```
itamarst: Mutating inputs isn't a great idea, tests might accidently pass-in something that then gets mutated. Make a copy and mutate that.

flocker/provision/_common.py
```diff
@@ -71,6 +71,8 @@ class Cluster(PRecord):
     :ivar FilePath certificates_path: Directory where certificates can be
         found; specifically the directory used by ``Certificates``.
     :ivar Certificates certificates: Certificates to for the cluster.
+    :ivar unicode dataset_backend_config_file: File with the backend
```
itamarst: FilePath instead?

itamarst:
Thinking out loud:

1. Access to backend is useful, I guess. Sometimes the effects we care about are in external systems.
2. I'm not sure this is best way to do it, but then I don't have any better ideas.
3. This will need to support other backends.

But... I guess this is interim step and long term plan is to expose profile info in REST API, and then we don't need to talk to the backend and can test that way?

itamarst:
So... address and merge. Still not sure if this is a good idea or a bad idea, but it's definitely an interesting idea! :grinning:

flocker/acceptance/testtools.py
sarum90: Fixed

flocker/acceptance/testtools.py
sarum90: Thanks for catching that! Fixed

flocker/provision/_common.py
sarum90: Done

sarum90:
+1 on it maybe not being the best idea, still seemed better than not having codified acceptance criteria.

Re:supporting other backends. If and when we need that should be simple to code. We would have to rework a bit of the node agent so that the "instantiate IBlockDeviceAPI" code could be re-used. Figured this was good enough for now for the current acceptance criteria. On the other hand... if we get to the point where we no longer need this... then we can just remove it.

Didn't just work on Buildbot though... going to look into that.

sarum90:
Enabled the test, saw it fail on Buildbot for the right reason. Disabled the test, got a clean buildbot run.

Merging.

PR #2095, opened 2015-10-28
[FLOC-3340] Move retry helpers to flocker.common

jml:
While looking at https://github.com/ClusterHQ/flocker/pull/2071, Adam & I independently noticed that it would be useful to have common logic for retrying things.

`loop_until` and `retry_failure` already exist in testtools. `poll_until` exists in a cinder module. This PR moves them all to one place, which makes the subtle variations in behavior more obvious (and hopefully easier to fix!). It's my hope that having a place to look for such things will prevent further ad hoc re-implementations from sprouting up.

Because the notion of doing this branch as a separate thing was late in coming, it also includes a change to `loop_until`, which now also takes a `steps` parameter, so callers can decide to loop a finite number of times, or exponentially, or whatever.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2095)

PR #2098, opened 2015-10-29
[FLOC-3343] No default reactor in retry helpers

jml:
**This PR depends on FLOC-3340 (#2095)**

One of the review comments in FLOC-3340 (#2095) was that it's unsafe to have a reactor parameter that defaults to the global reactor. 

This patch addresses that concern by:
1. making `reactor` a non-optional parameter
2. making it the _first_ parameter of `loop_until` and `retry_failure`
3. updating all call sites to pass `reactor` explicitly

As a general principle, I've tried to only pass a concrete reactor in at the leafs. Thus, I've updated `flocker.provision._ssh._conch` to pass `reactor` through from its entry point, `make_dispatcher`, and also `flocker.node.testtools.wait_for_unit_state` to take a reactor.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2098)

### DIFF
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -6,6 +6,7 @@ Tests for the datasets REST API.
 
 from uuid import UUID
 
+from twisted.internet import reactor
 from twisted.trial.unittest import TestCase
 
 from ...common import loop_until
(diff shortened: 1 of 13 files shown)


sarum90:
LGTM.

After grepping around I couldn't find any references that you missed. Hopefully this didn't induce  any carpal tunnel. Looks like it might have been one of those sorts of refactors.

Tests are still running, so I'll put this on your queue to merge tomorrow. Also 1 of your tests hit an EBS flake, so I'll manually kick it now.

PR #2100, opened 2015-10-29
[FLOC 2246] Remove volume manager documentation

davidoliverSP2:
FIxes 2246

In discussion with @robhaswell about FLOC 953 we agreed that all of the content in the Volume Manager documentation is out of date: https://docs.clusterhq.com/en/1.5.0/concepts/volume.html 

Specifically, it uses ZFS heavily and covers features that are no longer supported. It was agreed that this document is no longer required. Details relating to Flocker architecture are now covered in the Introduction topics, and FLOC 953 will include a re-write of the Cluster Architecture docs to replace the ZFS example with EBS.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2100)


docs/concepts/clustering.rst
```diff
@@ -7,14 +7,14 @@ Data-Oriented Clustering
 Minimal Downtime Volume Migration
```
sarum90: :grey_question: This page seems optimistic to me. Or rather... like aspirational. Do we have a functioning ZFS backend right now? I thought that was sort of work-in progress.  Is there more work to clean up these pages?

davidoliverSP2: thanks @sarum90 - yep, this PR just involves removing the Volumes doc. 

The changes to clustering.rst, (which was deemed relevant but out of date), will be covered in FLOC 953

sarum90:
Looks good to merge as is. As far as I'm concerned the removed page was potentially more harmful than good since most of our users use blockdevice deployers anyway (as of now anyway), and I don't think the information was accurate for anything that we currently support (although I'm new here, what do we support?).

It looks like you removed the page and all references to it correctly. I think there are some other concepts pages we might not want long term (Do we want to be in the routing business?).

I think that's all you were looking for in a review? I'm going to shy away from merging myself in case you were looking for a stronger +1 on the removal of the page. As far as I'm concerned though, Ship it!

PR #2101, opened 2015-10-29
[FLOC-3339] Add boto result logging

dwgebler:
Fixes FLOC-3339, this simply adds messages to the logs for the results of boto calls to create, tag and list volumes via the EC2 API, as well as the filtered results of volumes identified as belonging to the cluster on which this is running. This makes it easier to track and corroborate the EC2 EBS actions against the actions expected from current state and desired configuration / calculated changes in the cluster.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2101)


jml:
Thanks!

It might make sense to do these as full eliot Actions, but what you've got here now is an improvement on base, so I'm OK for this to merge once it passes tests.

dwgebler:
> It might make sense to do these as full eliot Actions

I'll do a follow-up to improve this logging once I've managed to figure out the original intermittent test failure ticket I needed these result logs for :)

PR #2102, opened 2015-10-29
[FLOC 2443] Add an introduction to choosing a backend

davidoliverSP2:
Fixes 2443

Added info to the list of storage backends in the introduction, which points to the table in our marketing website, so that the user has some more information about their backend options.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2102)


jml:
Glorious, thanks.

PR #2103, opened 2015-10-29
FLOC-3273 Modify the Node agent to plumb flocker profiles to the driver.

sarum90:
This PR modifies the Node agent on create requests.

After this PR, if the dataset being created has a metadata key "clusterhq:flocker:profile" and the backend supports creating volumes with a profile, then the value of that metadata will be handed down as the profile for the volume.

If that metadata entry exists but the backend does not support creating volumes with profiles, the agent will log that the profile was dropped, and will simply call create_volume on the backend without passing the profile down.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2103)

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -50,6 +50,9 @@ _logger = Logger()
 # XXX: Make this configurable. FLOC-2679
 DEFAULT_DATASET_SIZE = int(GiB(100).to_Byte().value)
 
+# The metadata key for flocker profiles.
+PROFILE_METADATA_KEY = u"clusterhq:flocker:profile"
+
 
 class VolumeException(Exception):
     """
@@ -733,7 +736,7 @@ class CreateBlockDeviceDataset(PRecord):
         except:
             return fail()
 
-        profile_name = self.dataset.metadata.get(u"clusterhq:flocker:profile")
+        profile_name = self.dataset.metadata.get(PROFILE_METADATA_KEY)
         dataset_id = UUID(self.dataset.dataset_id)
         size = allocated_size(allocation_unit=api.allocation_unit(),
                               requested_size=self.dataset.maximum_size)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -186,7 +186,7 @@ BLOCK_DEVICE_PATH = Field(
 
 PROFILE_NAME = Field(
     u"profile_name",
-    lambda dataset_id: unicode(dataset_id),
+    identity,
     u"The name of a profile for a volume."
 )
 
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -1031,6 +1031,8 @@ class ProfiledBlockDeviceAPIAdapter(PClass):
         backends that do not implement ``IProfiledBlockDeviceAPI``, but do
         implement ``IBlockDeviceAPI``.
         """
+        CREATE_VOLUME_PROFILE_DROPPED(dataset_id=dataset_id,
+                                      profile_name=profile_name).write(_logger)
         return self._blockdevice_api.create_volume(dataset_id=dataset_id,
                                                    size=size)
 
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -184,9 +184,9 @@ BLOCK_DEVICE_PATH = Field(
     u"The system device file for an attached block device."
 )
 
-PROFILE_NAME = Field(
+PROFILE_NAME = Field.forTypes(
     u"profile_name",
-    identity,
+    [unicode],
     u"The name of a profile for a volume."
 )
 
@@ -696,7 +696,10 @@ def check_allocatable_size(allocation_unit, requested_size):
 
 
 # Get rid of this in favor of calculating each individual operation in
-# BlockDeviceDeployer.calculate_changes.  FLOC-1771
+# BlockDeviceDeployer.calculate_changes. Also consider splitting the
+# CreateVolume portion of the IStateChange into 2 IStateChanges, one for
+# creating a volume with a profile, and one for creating a volume without a
+# profile. FLOC-1771
 @implementer(IStateChange)
 class CreateBlockDeviceDataset(PRecord):
     """
@@ -715,6 +718,32 @@ class CreateBlockDeviceDataset(PRecord):
             _logger,
             dataset=self.dataset, mountpoint=self.mountpoint
         )
+
+    def _create_volume(self, deployer):
+        """
+        Create the volume using the backend API. This method will create the
+        volume with a profile if the metadata on the volume suggests that we
+        should
+
+        :param deployer: The deployer to use to create the volume.
+
+        :returns: The created ``BlockDeviceVolume``.
+        """
+        api = deployer.block_device_api
+        profile_name = self.dataset.metadata.get(PROFILE_METADATA_KEY)
+        dataset_id = UUID(self.dataset.dataset_id)
+        size = allocated_size(allocation_unit=api.allocation_unit(),
+                              requested_size=self.dataset.maximum_size)
+        if profile_name:
+            return (
+                deployer.profiled_blockdevice_api.create_volume_with_profile(
+                    dataset_id=dataset_id,
+                    size=size,
+                    profile_name=profile_name
+                )
+            )
+        else:
+            return api.create_volume(dataset_id=dataset_id, size=size)
 
     def run(self, deployer):
         """
@@ -736,20 +765,7 @@ class CreateBlockDeviceDataset(PRecord):
         except:
             return fail()
 
-        profile_name = self.dataset.metadata.get(PROFILE_METADATA_KEY)
-        dataset_id = UUID(self.dataset.dataset_id)
-        size = allocated_size(allocation_unit=api.allocation_unit(),
-                              requested_size=self.dataset.maximum_size)
-        if profile_name:
-            volume = (
-                deployer.profiled_blockdevice_api.create_volume_with_profile(
-                    dataset_id=dataset_id,
-                    size=size,
-                    profile_name=profile_name
-                )
-            )
-        else:
-            volume = api.create_volume(dataset_id=dataset_id, size=size)
+        volume = self._create_volume(deployer)
 
         # This duplicates AttachVolume now.
         volume = api.attach_volume(
@@ -761,7 +777,7 @@ class CreateBlockDeviceDataset(PRecord):
         create = CreateFilesystem(volume=volume, filesystem=u"ext4")
         d = run_state_change(create, deployer)
 
-        mount = MountBlockDevice(dataset_id=dataset_id,
+        mount = MountBlockDevice(dataset_id=volume.dataset_id,
                                  blockdevice_id=volume.blockdevice_id,
                                  mountpoint=self.mountpoint)
         d.addCallback(lambda _: run_state_change(mount, deployer))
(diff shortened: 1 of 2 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2706,6 +2717,78 @@ def test_remove_inode(self):
...
+        loopback_blockdevice_api=loopbackblockdeviceapi_for_test(
+            test_case, allocation_unit=allocation_unit))
+
+
+_fake_profiled_loopback_block_device_api = partial(
+    fakeprofiledloopbackblockdeviceapi_for_test,
+    allocation_unit=LOOPBACK_ALLOCATION_UNIT)
+
+
+class FakeProfiledLoopbackBlockDeviceIBlockDeviceTests(
```
sarum90: I'm not convinced this test adds much. If we are to assume that "proxyForInterface" works, it's hard to see what this adds over LoopbackBlockDeviceAPITests. Love to get review feedback on this.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2310,22 +2314,29 @@ def test_profile_respected(self):
         mandatory profiles.
         """
         for profile in (c.value for c in MandatoryProfiles.iterconstants()):
-            self.api.create_volume_with_profile(profile)
+            self.api.create_volume_with_profile(dataset_id=uuid4(),
```
sarum90: For the reviewer:

Unfortunately when I merged these changes in they were un-used, and unfortunately incomplete. Upon request I can separate maybe 10 lines of this change to a different PR. didn't seem substantial enough for a new PR to me, but at 300 lines this PR I'm sympathetic that this might be hard to digest.

flocker/node/agents/blockdevice.py
```diff
@@ -719,13 +736,20 @@ def run(self, deployer):
...
+        if profile_name:
+            volume = (
+                deployer.profiled_blockdevice_api.create_volume_with_profile(
+                    dataset_id=dataset_id,
+                    size=size,
+                    profile_name=profile_name
+                )
+            )
+        else:
+            volume = api.create_volume(dataset_id=dataset_id, size=size)
```
exarkun: It might help readability a bit to factor all of this `create_volume` related code out into a private helper method, I think.

flocker/node/agents/test/test_blockdevice.py
exarkun: That seems fine.

Unrelatedly, there are a couple things that ideally would be different about this test, I think.

There are some drawbacks to having loops in tests.  If one the call fails for one of the values, the test won't exercise the rest of the values.  So you can't fix the failing case and be confident the test will pass on the next run.  Also, typically failures are such that you can't tell which iteration of the loop the test failed on so you don't actually have a great idea of what the problem is.

Also, this particular test doesn't make any assertions so, in xUnit lingo, it can't fail, it can only "error".  This isn't a huge deal but it's usually not something to aim for: failures are more likely to convey something meaningful to a developer than errors are.

Also, the test is satisfied by a no-op implementation, so not a ton of value there. :smile:

flocker/node/agents/test/test_blockdevice.py
exarkun: I guess there's a chance `FakeProfiledLoopbackBlockDeviceAPI` could break some aspect of `LoopbackBlockDeviceAPI`.

Doesn't seem terribly likely, and the former is test-only code so we generally verify that it is correct by using it in tests for something else and making sure the tests fail and pass in the ways we expect.  So I could definitely see dropping this, particularly if it adds noticably to the suite run time.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3395,23 +3478,99 @@ class CreateBlockDeviceDatasetInterfaceTests(
...
+            self.api.compute_instance_id()
+        )
+
+
+def make_createblockdevicedataset_mixin(profiled_api):
+    """
+    Constructs a base class for tests that verify the implementation of
+    ``CreateBlockDeviceDataset``.
+
+    This ``IStateChange`` needs to be tested in two configurations:
```
exarkun: Hm.  Alternatively, there should be two different `IStateChange` implementations...  Consider this for follow-up work (maybe add a note to the issue for breaking `CreateBlockDeviceDataset` up into pieces).

Ideally, there shouldn't be branches in an `IStateChange.run` method.  Decisions should all have been made by the time the object was created.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3395,23 +3478,99 @@ class CreateBlockDeviceDatasetInterfaceTests(
...
+
+        [volume] = self.api.list_volumes()
+        device_path = self.api.get_device_path(volume.blockdevice_id)
+        return (
+            volume, device_path, expected_mountpoint,
+            self.api.compute_instance_id()
+        )
+
+
+def make_createblockdevicedataset_mixin(profiled_api):
```
exarkun: missing param docs for this

flocker/node/agents/blockdevice.py
```diff
@@ -983,6 +1007,33 @@ def create_volume_with_profile(name, size, profile_name):
...
+        invariant=lambda i: (IBlockDeviceAPI.providedBy(i),
+                             '_blockdevice_api must provide IBlockDeviceAPI'))
+
+    def create_volume_with_profile(self, dataset_id, size, profile_name):
+        """
+        Reverts to constructing a volume with no profile. To be used with
+        backends that do not implement ``IProfiledBlockDeviceAPI``, but do
+        implement ``IBlockDeviceAPI``.
+        """
+        CREATE_VOLUME_PROFILE_DROPPED(dataset_id=dataset_id,
```
exarkun: In new code, use the implicit global Eliot logger, unless some interaction with existing code makes it trickier to do so.

In this case I think you can just `.write()` and update the tests to remove the monkey-patching.

flocker/node/agents/blockdevice.py
```diff
@@ -181,6 +184,12 @@ def __init__(self, blockdevice):
     u"The system device file for an attached block device."
 )
 
+PROFILE_NAME = Field(
+    u"profile_name",
+    identity,
```
exarkun: Probably better to use `Field.for_type` and limit this to unicode.  Anything else is a bug, right?

flocker/node/agents/blockdevice.py
```diff
@@ -1456,6 +1507,18 @@ class BlockDeviceDeployer(PRecord):
     mountroot = field(type=FilePath, initial=FilePath(b"/flocker"))
 
     @property
+    def profiled_blockdevice_api(self):
+        """
+        Get an ``IProfiledBlockDeviceAPI`` provider which can create profiled
+        block devices for this deployer.
```
exarkun: "profiled block devices" is confusing - partly in a new way, partly because of existing naming confusing in this module.

What the thing can really create is volumes (and then it can attach them and expose the result via a Linux block device).

And "profiled" seems to hint at some kind of instrumented volume that will report back usage or performance information.

Maybe "create volumes configured based on pre-defined profiles" (though that's quite a mouthful; it would be great to have a convenient shorthand for this concept.  Maybe "profiled volume" is it, but maybe there's a less ambiguous option somewhere out there.)

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3395,23 +3478,99 @@ class CreateBlockDeviceDatasetInterfaceTests(
...
+        """
+        Call ``CreateBlockDeviceDataset.run`` with a ``BlockDeviceDeployer``.
+
+        :param UUID dataset_id: The uuid4 identifier for the dataset which will
+            be created.
+        :param int maximum_size: The size, in bytes, of the dataset which will
+            be created.
+        :param pmap(unicode, unicode) metadata: The metadata for the dataset.
+
+        :returns: A 3-tuple of:
```
exarkun: Tuples are a sad interface type.  Consider a `PClass`.
:question:

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -3395,23 +3478,99 @@ class CreateBlockDeviceDatasetInterfaceTests(
...
+            dataset_id=unicode(dataset_id),
+            maximum_size=maximum_size,
+            metadata=metadata
+        )
+
+        change = CreateBlockDeviceDataset(
+            dataset=dataset, mountpoint=expected_mountpoint
         )
 
+        run_state_change(change, self.deployer)
```
exarkun: This supposes that the state change is implemented as a synchronous operation - which `CreateBlockDeviceDataset` happens to be, but isn't necessarily the case.  I see this is just refactored code, though, so pre-existing problem, not something to fix right now - just thought I'd mention it.

flocker/node/agents/test/test_blockdevice.py
exarkun: Also it seems to be a four tuple :wink:

exarkun:
Thanks.  Reviewed this primarily with two things in mind:

  * Avoid introducing problems for users not interacting with the profiling feature
  * Make useful progress towards demoable profiling feature

I think both of those goals are pretty well satisfied (with one exception, see below).  I left some comments inline about very minor issues, all very localized points, nothing really earth shaking I think.

The EBS functional builder has several errors, though.  Could just be coincidence - no EBS changes here, after all.  I think it would be a good idea to dig into them a little bit so we can be confident they're definitely not related to this change.  Once that possibility is ruled out and the inline comments are addressed, please merge.

If it turns out the failures are due to this change somehow, please re-submit the fix for review.  Thanks again.

flocker/node/agents/blockdevice.py
sarum90: Yup, Thanks! I just wasn't familiar enough with Eliot to look for that :).

flocker/node/agents/blockdevice.py
sarum90: Done

flocker/node/agents/blockdevice.py
sarum90: Thanks! much better.

flocker/node/agents/blockdevice.py
sarum90: Used your language for now.

flocker/node/agents/test/test_blockdevice.py
sarum90: Acknowledged. Yea I imagine with time (once the interface is expanded) this will add more value, but for now with the "create and hope" interface I think this is the best we can do. I figured this was better than nothing for now, as I figure there is some value in verifying we don't "error" even if the value is less than  something that has proper "failure" conditions.

flocker/node/agents/test/test_blockdevice.py
sarum90: adds 17 seconds to runtime. That seems like a lot. Given both of our doubts on the value of the test I am removing it.

flocker/node/agents/test/test_blockdevice.py
sarum90: :grinning: yea this was just moving code. Although I agree that PClass is better, and I like small incremental improvements to code so I'll go back and change all callsites before merging.

flocker/node/agents/test/test_blockdevice.py
sarum90: +1.

flocker/node/agents/test/test_blockdevice.py
sarum90: added.

flocker/node/agents/test/test_blockdevice.py
sarum90: Added to comment.

sarum90:
Finished review comments. Now digging into the EBS functional builder and re-running all tests.

sarum90:
rebased on top of master to get the latest EC2 logging (it looks like EBS is just particularly eventually consistent for me right now, and master has better logs for these). Re-running tests.

sarum90:
Ah! I'm getting `flocker:node:agents:blockdevice:aws:no_available_device` all over the place. Further investigation showed that one of our buildbot slaves has been stuck in a state where it had like 10 volumes attached to it since yesterday. Re-running tests now.

sarum90:
So another bug in our buildbot. ubuntu on rackspace was broken (FLOC-3358). I was able to repro all morning, but at 4pm it seemed to stop being reproducible.  at that time I got a green run on this PR, so I guess I'll be submitting presently.

PR #2106, opened 2015-10-30
[FLOC 977] add elk download file

davidoliverSP2:
Fixes 977

An old issue which adds a downloadable elk file where the text suggests one should be. A few tweaks to the wording around this.

I've not manually tested this, as it was an obvious error (the contents of the elk file was always provided - this just adds a link for users to download the file)

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2106)


jongiddy:
I haven't actually tested that the code works, but it appears to be more likely to work than what was there initially.  There don't appear to be any acceptance tests for this code.  Please add an issue to create one.

OK to merge this.

davidoliverSP2:
thanks @jongiddy

PR #2107, opened 2015-10-30
[FLOC-3262] Loop to destroy container

jongiddy:
Container removal fails intermittently with a 500 error. To make this more reliable, any 500 error will be logged and the removal will be repeated. If the removal can be fixed simply by retrying, this will make the code more robust. If retries do not help, at least we have more logging info.

The container stopping and removal both run synchronously in a separate thread, so the `poll_until` function (synchronous version of `loop_until`) is modified to take a `steps` iterable for retry intervals (matching recent changes to `loop_until`).

Tests for `poll_until` are added, and to support the testing, the `sleep` function used in `poll_until` can be set to ensure tests do not take too much time.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2107)

### DIFF
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -161,6 +161,8 @@ def poll_until(predicate, steps, sleep=None):
     :param callable sleep: called with the interval to delay on.
         Defaults to `time.sleep`.
     :returns: the non-false result from the final call.
+    :raise LoopExceeded: If given a finite sequence of steps, and we exhaust
+        that sequence waiting for predicate to be truthy.
     """
     if sleep is None:
         sleep = time.sleep
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -169,3 +169,7 @@ def poll_until(predicate, steps, sleep=None):
         if result:
             return result
         sleep(step)
+    result = predicate()
+    if result:
+        return result
+    raise LoopExceeded(predicate, result)
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -148,7 +148,7 @@ def retry_failure(reactor, function, expected=None, steps=None):
     return d
 
 
-def poll_until(predicate, interval):
+def poll_until(predicate, interval, sleep=None):
     """
     Perform steps until a non-false result is returned.
 
@@ -158,10 +158,14 @@ def poll_until(predicate, interval):
     :param predicate: a function to be called until it returns a
         non-false result.
     :param interval: time in seconds between calls to the function.
+    :param callable sleep: called with the interval to delay on.
+        Defaults to `time.sleep`.
     :returns: the non-false result from the final call.
     """
+    if sleep is None:
+        sleep = time.sleep
     result = predicate()
     while not result:
-        time.sleep(interval)
+        sleep(interval)
         result = predicate()
     return result
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -148,24 +148,24 @@ def retry_failure(reactor, function, expected=None, steps=None):
     return d
 
 
-def poll_until(predicate, interval, sleep=None):
+def poll_until(predicate, steps, sleep=None):
     """
     Perform steps until a non-false result is returned.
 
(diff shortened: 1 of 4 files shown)


exarkun:
Thanks!  Looks good, hopefully this new retry loop will address the problem - and I dunno how to evaluate whether it does or not except to run the code a bunch and see if the problem goes away.  It might be interesting to do some builds pre-merge and see if the logs indicate that the loop is having an effect (ie, try to find a (green!) build where the remove has to be retried at least once).  I dunno how often this problem comes up, though, so I dunno how practical it is to do that pre-merge testing.

If you can do that and get verification that the problem is really solved, that'd be awesome.  If not, oh well, we'll do the testing on master.  Apart from that, just address the other minor inline comments, then go ahead and merge.  Thanks again.


---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

<sup>**[flocker/common/_retry.py, line 163 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2107#-K1tDDZqCjqgIHSID6ep)** ([raw file](https://github.com/clusterhq/flocker/blob/bcb663be4e35fa75ac50a773f6ea8b334d6fdc65/flocker/common/_retry.py#L163)):</sup>
Raises something now, too.


---

<sup>**[flocker/common/test/test_retry.py, line 358 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2107#-K1tE0Rdi39BZ_JSvM2U)** ([raw file](https://github.com/clusterhq/flocker/blob/bcb663be4e35fa75ac50a773f6ea8b334d6fdc65/flocker/common/test/test_retry.py#L358)):</sup>
missing docstring


---

<sup>**[flocker/common/test/test_retry.py, line 365 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2107#-K1tE3sTzor5V6MFelNp)** ([raw file](https://github.com/clusterhq/flocker/blob/bcb663be4e35fa75ac50a773f6ea8b334d6fdc65/flocker/common/test/test_retry.py#L365)):</sup>
s/generator/iterable/


---

<sup>**[flocker/node/_docker.py, line 807 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2107#-K1tITQzCRSCVH_RHP8C)** ([raw file](https://github.com/clusterhq/flocker/blob/bcb663be4e35fa75ac50a773f6ea8b334d6fdc65/flocker/node/_docker.py#L807)):</sup>
Probably also log as much detail about the exception as possible here, too?  Maybe it's also available in the Docker logs but those are less easily accessible from CI failures at the moment.


---

<sup>**[flocker/node/_docker.py, line 818 \[r2\]](https://reviewable.io:443/reviews/clusterhq/flocker/2107#-K1tIkRhjjzBqLxdhl6E)** ([raw file](https://github.com/clusterhq/flocker/blob/bcb663be4e35fa75ac50a773f6ea8b334d6fdc65/flocker/node/_docker.py#L818)):</sup>
Ah, this method is so much easier to read now.  Thank you.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2107)

jongiddy:
There are instances in http://build.clusterhq.com/builders/flocker-ubuntu-14.04/builds/7082/steps/trial/logs/test.log
where the additional logging makes it clear that the actual problem in FLOC-3262 is that `test_dead_is_removed` gets the error because `remove` is called twice, once as part of the test, and again as part of the cleanup. 

The cleanup call usually fails with 404 (not found), which is ignored (since removal is supposed to be idempotent). The reported issue would appear to suggest that occasionally the second remove runs while the container is still in the RemovalInProgress state, and returns 500.  This could probably be ignored safely as well, but 500 might be used to indicate other problems, so safer to retry until removal is confirmed.  


---

Review status: 1 of 4 files reviewed at latest revision, all discussions resolved.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2107)

PR #2109, opened 2015-11-02
[FLOC-3352] Global transition tables

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3352

Also applied the treatment to the other transition table in this module so it's also available for inspection by tools.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2109)


mpgomez:
It makes sense to me; I don't have any constructive comments about it though :)

PR #2110, opened 2015-11-02
Adds lint flake8 checks jenkins job FLOC-2858

Azulinho:
This PR adds a new Jenkins Job for the missing flake8 lint checks.
It simply runs tox -e lint just like the BB builder

Jenkins job:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/add_lint_FLOC-2858/job/run_lint/7/console

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2110)

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -717,6 +717,10 @@ common_cli:
     wget https://raw.githubusercontent.com/AndrewDryga/vagrant-box-osx/master/Vagrantfile
     vagrant up
 
+  run_lint: &run_lint |
+    # run flake8 lint tests on Flocker source code
+    tox -e lint
+
 
 #-----------------------------------------------------------------------------#
 # Job Definitions below this point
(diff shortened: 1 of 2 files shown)


build.yaml
```diff
@@ -717,6 +717,18 @@ common_cli:
...
+    # we need to unset PIP_INDEX_URL since it causes tox to fail.
+    # in order to use our pip caching we need to append --trusted-hosts
+    # which we currently can't do through tox.
+    unset PIP_INDEX_URL
+    # run flake8 lint tests on Flocker source code
+    tox -e lint
+
+  install_flake8: &install_flake8 |
+    # flake8 is used by lint tests on Flocker source code
+    \${PARSE_LOGS} pip install flake8 \${PIP_ADDITIONAL_OPTIONS}
```
jml: It's unfortunate that if ever we change the tox.ini file to do linting differently we'll have to remember to change this file to install a different set of dependencies. i.e. the knowledge of how to lint is spread across multiple places, when it should be in just one place.

I don't think it needs to be addressed before landing (for all I know, buildbot is just as bad), but it would be nice to address it _sometime_.

jml:
Thanks. You might say "we can live without it", but automated lint checking is incredibly useful. This is OK to land. I've raised a more general design concern that probably doesn't need to be addressed yet. 

If you *do* decide to address it, then please resubmit for review.

Otherwise, merge away.

build.yaml
Azulinho: do we need tox ?
I think this might be the only place where we call tox for running tests

Azulinho:
@jml FYI the "we can live without" was agreed on a triage meeting of issues for this particular EPIC.
you can see the list and priorities set for the next EPIC here:
https://clusterhqteam.slack.com/archives/jenkins/p1445526526000011

Azulinho:
@jml  danka berry berry much for da review

PR #2111, opened 2015-11-02
Added maybeDeferred and test

mpgomez:
[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2111)


flocker/control/test/test_protocol.py
```diff
@@ -732,6 +732,31 @@ def test_configuration_change(self):
             dict(configuration=agent.desired, state=agent.actual),
         )
 
+    def test_unexpected_exception_raised(self):
+        """
+        A configuration change results in connected protocols being notified
```
itamarst: This docstring looks wrong.

flocker/control/test/test_protocol.py
```diff
@@ -732,6 +732,31 @@ def test_configuration_change(self):
...
+        service.startService()
+        server = LoopbackAMPClient(client.locator)
+        self.patch(server,
+                   "callRemote",
+                   raise_unexpected_exception
+                   )
+
+        service.connected(server)
+
+        service.configuration_service.save(TEST_DEPLOYMENT)
```
itamarst: This does test that the exception isn't raised, but it doesn't test the expected behavior which is that if connection A gets exception raised that doesn't impact updates to connection B. There should be a test case for that in the case where callRemote returns a `Failure`, so a variant of that is probably what you want (perhaps with shared code for most of it?).

itamarst:
Thanks for working on this!

The test is half-way there, but should probably go the whole way to really make sure we're not losing updates to other connections. So do that and resubmit.

flocker/control/test/test_protocol.py
```diff
@@ -732,6 +732,43 @@ def test_configuration_change(self):
             dict(configuration=agent.desired, state=agent.actual),
         )
 
+    #TODO delete
```
itamarst: You can delete this test, yes.

flocker/control/test/test_protocol.py
```diff
@@ -778,6 +815,69 @@ def test_second_configuration_change_waits_for_first_acknowledgement(self):
...
+        configuration = service.configuration_service.get()
+        modified_configuration = arbitrary_transformation(configuration)
+
+        server = LoopbackAMPClient(client.locator)
+        delayed_server = DelayedAMPClient(server)
+        # Send first update
+        service.connected(delayed_server)
+        # Because the first agent fails to update, we do not
+        # expect the original configuration to change
+        first_agent_desired = configuration
```
itamarst: Later on in test you compare this to `configuration`. And it matches, unsurprisingly.

It seems like only thing you need to actually assert is that post-save() the second agent gets the modified_configuration? Which should fail before your fix and succeed after.

flocker/control/test/test_protocol.py
```diff
@@ -778,6 +815,69 @@ def test_second_configuration_change_waits_for_first_acknowledgement(self):
...
+        If the first update fails, we want to ensure that following updates
+        won't get stuck waiting, and will get updated.
+        """
+
+        def raise_unexpected_exception(self,
+                                       commandType=None,
+                                       *a, **kw):
+            raise Exception("I'm an unexpected exception")
+
+        agent = FakeAgent()
```
itamarst: It's a little confusing that these are constructed here and then used later on; maybe put all setup of this client in one place.

flocker/control/test/test_protocol.py
```diff
@@ -778,6 +815,69 @@ def test_second_configuration_change_waits_for_first_acknowledgement(self):
...
+        first_agent_desired = configuration
+
+        # Send second update
+        service.configuration_service.save(modified_configuration)
+        second_agent_desired = agent.desired
+
+        delayed_server.respond()
+        third_agent_desired = agent.desired
+
+        self.assertEqual(
```
itamarst: As mentioned above, comparing only the third value should suffice.

itamarst:
Doesn't look like forced a build: http://build.clusterhq.com/boxes-flocker?branch=catch-all-exception-FLOC-3221

itamarst:
1. Address the above.
2. Force a build.
3. Merge when all builders in Buildbot are green. (it will add info here in the PR).

PR #2114, opened 2015-11-03
[FLOC 3334] Split up the Labs/Installer document

davidoliverSP2:
Fixes 3334

Splits out the Labs/Installer document, so that the context, steps and tutorial are 3 separate topics. This makes the Installer doc easier to follow, and easier for the user to recognise the start and end of each activity.

On reviewing, please observe the Cleaning up your Cluster section on Short Tuturial: Deploy and Migrate a Stateful App - is this the right place for these instructions? It seemed logical to me

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2114)


docs/labs/installer-getstarted.rst
```diff
@@ -0,0 +1,130 @@
...
+==================================
+Getting Started with the Installer
+==================================
+
+#. Installing the Installer:
+
+   First we install the installer on your workstation.
+   This will work on Linux or OS X machines with Docker installed.
+
+   * If you don't have Docker installed, install it now (`Mac       <https://docs.docker.com/mac/started/>`_, `Linux <https://docs.docker.com/linux/started/>`_).
```
sarum90: Do the additional spaces after Mac have a purpose? they seem to be new.

docs/labs/installer.rst
```diff
@@ -57,207 +57,14 @@ You may also be interested in the long-form documentation if you like to see exa
...
-.. note::
-
-    If you wish to clean up your cluster manually, be sure to delete the instances that were created in your AWS console and the ``flocker_rules`` security group.
+Get Started
+===========
+
+.. toctree::
+   :maxdepth: 1
+   
+   installer-getstarted
```
sarum90: :grey_question: Curious what the motivation was to move the installer instructions to its own page. To me it reads slightly better inline, although I definitely agree that the short tutorial should be a new page. Fine as is, it just seems as if those directions are really first-class to the installer

docs/labs/installer-tutorial.rst
```diff
@@ -0,0 +1,72 @@
...
+      uft-flocker-volumes list
+
+   At the end you should see the volume has moved to the second node.
+
+   This may take up to a minute since Flocker is ensuring the volume is on the second host before starting the container.
+
+#. Now visit ``http://<node 2 public IP>/`` and you’ll see that the location of the Moby Docks has been preserved.
+   Nice.
+
+Cleaning up your Cluster
```
sarum90: :grey_question: Should this be separate as well? Seems to me that it would be conceivable for someone to want to set up a cluster, not do our tutorial but kick the tires in some other manner, and then clean up the cluster.

Although... that would take a bit of changing, since half of this cleanup is for the demo and half of this cleanup is for the cluster.

docs/labs/installer-getstarted.rst
```diff
@@ -0,0 +1,130 @@
...
+
+   The tools will create some configuration files and certificate files for your cluster.
+   It is convenient to keep these in a directory, so let's make a directory on your workstation like this:
+
+   .. prompt:: bash $
+
+      mkdir -p ~/clusters/test
+      cd ~/clusters/test
+
+   Now we'll put some files in this directory.
```
sarum90: :grey_question: There was a note here: https://clusterhq.atlassian.net/browse/FLOC-3334 about addressing the comments here: https://github.com/ClusterHQ/unofficial-flocker-tools/issues/38, one of which was about this line. Looks like you are not addressing those points here, which is fine, but I just wanted to call your attention to that in case you wanted to make a new JIRA issue for those adjustments or otherwise put those on your work queue. Concern is that those notes might fall out of our list of things to do if we just close 3334 without addressing those notes.

sarum90:
Looks good to me! A couple notes here and there, but nothing large. Address and Merge when you get around to it.

davidoliverSP2:
thanks @sarum90 

I'm going to merge this, despite some broken links in the doc builder - I've opened FLOC 3371 to fix.

docs/labs/installer.rst
davidoliverSP2: the thinking was to keep the main page as the concept of the Installer, and its benefits, then the user links down to more information - to get started, then a tutorial, then clean up.

this will be tested before the installer gets moved into the main supported docs, so it might get concatenated again when that happens.

docs/labs/installer-getstarted.rst
davidoliverSP2: Thanks for spotting this @sarum90 - I've addressed a few of the issues in this list, but some are suggestions for the next iteration of the installer. 

@alicesowerby - I think we should review outstanding items on https://github.com/ClusterHQ/unofficial-flocker-tools/issues/38 and create a story to implement these changes.

PR #2116, opened 2015-11-03
[FLOC 3361] add introduction to Generating API Certs doc

davidoliverSP2:
Fixes 3361

Added an introduction, which clearly states that the API user certificates is mandatory, regardless of how you intend to communicate with the control service - please could someone in the know about this confirm this.

Many thanks

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2116)

### DIFF
--- a/docs/config/generate-api-certificates.rst
+++ b/docs/config/generate-api-certificates.rst
@@ -4,6 +4,8 @@
 Generating an API User Certificate
 ==================================
 
+If you wish to send instructions to the control service, by using the API directly, via the CLI, or any other method, you will need to follow the instructions below to generate an API user certificate:
+
 #. Generate an API user certificate.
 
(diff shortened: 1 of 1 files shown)


exarkun:
Thanks.

Using the API definitely requires a user certificate.  This applies whether you're trying to make changes to the cluster (kinda like "sending instructions", I guess?) or if you just want to read the cluster configuration or state.

The CLI is just a wrapper around the API so all the same things are true for that.

It's tricky to evaluate "or any other method".  There are no other methods right now, as far as I know.  I guess some are planned but work hasn't been started on them yet.  And they'll really just be more wrappers around the API.  So all the same should apply to them ... Or I could be mispredicting the future and someone may come up with some other scheme that works differently.

Happy to have you decide if/how any of that information needs to be incorporated into the docs.  Merge when you're satisfied and the doc builder is green.

PR #2119, opened 2015-11-03
[FLOC 3362] add context to Enabling control service doc

davidoliverSP2:
Fixes 3362

Adds an explanation of why you need to enable the control service.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2119)

### DIFF
--- a/docs/config/enabling-control-service.rst
+++ b/docs/config/enabling-control-service.rst
@@ -1,7 +1,7 @@
 .. _enabling-control-service:
 
 ====================================
-Enabling the Flocker control service 
+Enabling the Flocker Control Service 
 ====================================
 
 The control service is the brain of Flocker; it can live anywhere in your cluster, and enabling it is an essential step in setting up your cluster.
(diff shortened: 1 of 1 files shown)


docs/config/enabling-control-service.rst
```diff
@@ -4,6 +4,8 @@
 Enabling the Flocker control service 
 ====================================
 
+The control service is the brain of Flocker; it can live anywhere in your cluster, and enabling it is an essential step in setting up your cluster.
```
exarkun: If you wanted to go into any more detail, you could also say that the control service is what serves the API for the cluster and you could say that all of the convergence agents coordinate their activities by communicating with the control service.

Also I think there's an extra comma in this sentence.  And if you want, you could put a newline after the `;` (making the lines a bit shorter).

exarkun:
Thanks.  Just one minor suggestion inline.  Address as you like and then merge when the doc builder is green.

PR #2120, opened 2015-11-03
[FLOC 3363] add context to Enabling agent service doc

davidoliverSP2:
Fixes 3363

Adds an explanation of why you need to enable the agent service.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2120)

### DIFF
--- a/docs/config/enabling-agent-service.rst
+++ b/docs/config/enabling-agent-service.rst
@@ -2,7 +2,7 @@
 Enabling the Flocker Agent Service
 ==================================
 
-The Flocker agents, the `flocker-dataset-agent` and the `flocker-container-agent`, are the workhorses of Flocker; you have them on each node in your cluster, and enabling them is an essential step in setting up your cluster:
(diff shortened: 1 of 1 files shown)


docs/config/enabling-agent-service.rst
```diff
@@ -2,6 +2,8 @@
 Enabling the Flocker agent service
 ==================================
 
+The Flocker agents are the workhorses of Flocker; you have one on each node in your cluster, and enabling them is an essential step in setting up your cluster:
```
exarkun: There are two on each node right now.  There's the dataset convergence agent and the container convergence agent.

The plan is to get rid of the container convergence agent or at least to make it optional (so if you're using something like kubernetes to manage containers, you don't have to run our container management software as well).  We're not there yet, though, and it's not clear when we will be.  Also it's not clear what the installation story will look like at that point.  It could be drastically different from what we have now.

If you want some examples of things the dataset agent does, they're things like "create the filesystem for a newly created dataset" and "mount and unmount filesystems to support movement of datasets between nodes".

exarkun:
Thanks.  One comment inline with some extra info.  Merge when you're happy and the doc builder is green.

PR #2126, opened 2015-11-04
[FLOC-3357] sequentially() and in_parallel() now have eliot actions

itamarst:
1. Added explicit eliot actions for in-parallel and sequential runs.
2. Minorly refactored the code to be more object-oriented.
3. Fixed the bug which is the point of the exercise: sub-statechanges of `sequentially` are now run in correct Eliot context so it's possible to see them in appropriate place when reading logs.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2126)

### DIFF
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -31,9 +31,15 @@ class IStateChange(Interface):
     An operation that changes local state.
     """
     eliot_action = Attribute(
-        "An ``eliot.ActionType`` within the context of which to execute this "
-        "state change's run method."
-    )
+        """
+        A hack whereby getting this attributes has a side-effect: a
+        ``eliot.ActionType`` is started and return. This state change's
+        run method should be run within the context of the returned
+        action.
+
+        At some point we should fix this so it's a method instead of a
+        attribute-which-must-always-be-a-property.
+        """)
 
     def run(deployer):
         """
@@ -70,30 +76,16 @@ def run_state_change(change, deployer):
 
     :return: ``Deferred`` firing when the change is done.
     """
-    if isinstance(change, _InParallel):
-        return gather_deferreds(list(
-            run_state_change(subchange, deployer)
-            for subchange in change.changes
-        ))
-    if isinstance(change, _Sequentially):
-        d = succeed(None)
-        for subchange in change.changes:
-            d.addCallback(
-                lambda _, subchange=subchange: run_state_change(
-                    subchange, deployer
-                )
-            )
-        return d
-
     with change.eliot_action.context():
         context = DeferredContext(maybeDeferred(change.run, deployer))
         context.addActionFinish()
         return context.result
 
 
-# run_state_change doesn't use the IStateChange implementation provided by
-# _InParallel and _Sequentially but those types provide it anyway because
-# certain other areas of the test suite depend on it.
+LOG_SEQUENTIALLY = ActionType("flocker:node:sequentially", [], [])
+LOG_IN_PARALLEL = ActionType("flocker:node:in_parallel", [], [])
+
+
 @implementer(IStateChange)
 class _InParallel(PRecord):
     changes = field(
@@ -105,12 +97,15 @@ class _InParallel(PRecord):
         mandatory=True
     )
 
-    # Supplied so we technically conform to the interface.  This won't actually
-    # be used.
-    eliot_action = None
+    @property
+    def eliot_action(self):
+        return LOG_IN_PARALLEL()
 
     def run(self, deployer):
-        return run_state_change(self, deployer)
+        return gather_deferreds(list(
+            run_state_change(subchange, deployer)
+            for subchange in self.changes
+        ))
 
 
 def in_parallel(changes):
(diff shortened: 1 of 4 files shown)


flocker/node/_change.py
```diff
@@ -31,9 +31,15 @@ class IStateChange(Interface):
...
-    )
+        """
+        A hack whereby getting this attributes has a side-effect: a
+        ``eliot.ActionType`` is started and return. This state change's
+        run method should be run within the context of the returned
+        action.
+
+        At some point we should fix this so it's a method instead of a
+        attribute-which-must-always-be-a-property.
+        """)
```
jml: Can't that point be now? Please?

flocker/node/test/test_change.py
```diff
@@ -521,3 +523,47 @@ def test_failed(self, logger):
...
+
+    def assert_nested_logging(self, combo, action_type, logger):
+        """
+        All the underlying ``IStateChange`` will be run in Eliot context in
+        which the sequential ``IStateChange`` is run, even if they are not
+        run immediately.
+
+        :param combo: ``sequentially`` or ``in_parallel``.
+        :param action_type: ``eliot.ActionType`` we expect to be parent of
+            sub-changes logs.
```
jml: Nit: "log entries" clearer than "logs"? :grey_question:

flocker/node/test/test_change.py
```diff
@@ -521,3 +523,47 @@ def test_failed(self, logger):
...
+        :param logger: A ``MemoryLogger`` where messages go.
+        """
+        actions = [ControllableAction(result=Deferred()),
+                   ControllableAction(result=succeed(None))]
+        for action in actions:
+            self.patch(action, "_logger", logger)
```
jml: Why `patch` rather than straight-up assignment? Also, why not make `logger` an optional constructor parameter for `ControllableAction`?

itamarst: `patch` is better insofar it complains if attribute isn't there in first place.

As far as constructor: honestly we don't need this at all given Eliot capacity for catching global log messages via `capture_logging`. See comment about next-step refactor.

flocker/node/_change.py
itamarst: Should I change the JIRA issue to "clean up IStateChange logging"?

flocker/node/test/test_change.py
itamarst: I could fix that in this branch. Or maybe we want follow up branch where we cleanup that and the property thing.

flocker/node/test/test_change.py
```diff
@@ -521,3 +523,47 @@ def test_failed(self, logger):
...
+                    logger.messages, CONTROLLABLE_ACTION_TYPE),
+                length=2))
+
+    @capture_logging(None)
+    def test_sequential_logging(self, logger):
+        """
+        All the underlying ``IStateChange`` will be run in Eliot context in
+        which the sequential ``IStateChange`` is run.
+        """
+        self.assert_nested_logging(sequentially, LOG_SEQUENTIALLY, logger)
```
jml: :grey_question: No action required, but wanted to get this off my chest. 

I'm not a fan of the pattern of making one big helper that does the set-up, operation to test and assertions. What is saved in repetition, is often lost in ease of manual analysis and  cyclomatic simplicity. In cases like line 545, code that's not relevant to one test is run as a part of another, which (in other circumstances) invites unexpected dependencies of behaviour. 

I'm all for helpers for set-up and for domain-specific assertion methods, but I think these should be called explicitly from within `test_` methods, and that manipulation of the system under test should likewise be direct from test methods.

I realize though that "one big helper" is a very common pattern in flocker, and that this review isn't the place to change that. So, no change required.

flocker/node/_change.py
jml: Sorry, I don't see how that's relevant.

flocker/node/test/test_change.py
jml: > `patch` is better insofar it complains if attribute isn't there in first place.

Is that particularly relevant here? We assign to instance variables elsewhere without `patch`. 

> I could fix that in this branch. Or maybe we want follow up branch where we cleanup that and the property thing.

Well, my preference is that you fix them here, but I'm also happy for them to be fixed in a follow-up branch.

jml:
Couple of small changes suggested (logs -> log entries; patch -> assignment).

I really think it's a mistake to have the property hack even temporarily—ideally we should be clearing the weeds as we go, not sowing new ones—but I'm OK with this landing if there's a follow-up that does the cleanup.

If you decide to do the cleanup in this branch, please resubmit for review.

itamarst:
Oh, I see, it's a misunderstanding.

The hack with `@property` is pre-existing - I just updated the docstring to reflect reality.

flocker/node/test/test_change.py
itamarst: OK, will keep in mind for the future.

flocker/node/_change.py
itamarst: It's a pre-existing issue. I just updated docs to reflect reality.

PR #2127, opened 2015-11-04
[FLOC-3261] Design review: use node eras to identify nodes, so reboots invalidate node state

itamarst:
The idea is that if each reboot you have new era for a node, and clients of the REST API identify the node using its era (easily obtainable from `/proc`), the clients will never get stale data from previous reboots.

This also means clients don't have to read our TLS certificates to find node identity, a minor benefit.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2127)

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -226,6 +226,15 @@ class NoOp(Command):
     requiresAnswer = False
 
 
+class SetNodeEra(Command):
+    """
+    Update the node's era.
+
+    Sent by agent upon first connecting. Eras should reset on each reboot.
+    """
+    arguments = [('era', Unicode())]
+
+
 class ClusterStatusCommand(Command):
     """
     Used by the control service to inform a convergence agent of the
@@ -316,7 +325,7 @@ class ControlServiceLocator(CommandLocator):
 
     @VersionCommand.responder
     def version(self):
-        return {"major": 1}
+        return {"major": 2}
 
     @NodeStateCommand.responder
     def node_changed(self, eliot_context, state_changes):
@@ -325,6 +334,12 @@ class ControlServiceLocator(CommandLocator):
                 self._source, state_changes,
             )
             return {}
+
+    @SetNodeEra.responder
+    def set_node_era(self, era):
+        # 1. get node UUID from certificate
+        # 2. tell cluster state service to wipe any NodeState for that node UUID which has different era
+        # 3. if NodeState with given node UUID and era does not exist, add one to cluster state service
 
 
 class ControlAMP(AMP):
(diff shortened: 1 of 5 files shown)


flocker/control/_protocol.py
```diff
@@ -316,7 +325,7 @@ def noop(self):
 
     @VersionCommand.responder
     def version(self):
-        return {"major": 1}
+        return {"major": 2}
```
exarkun: A nice issue for the debugging epic would probably be "fail with a clear error when mixed protocol versions are encountered".

flocker/control/httpapi.py
```diff
@@ -868,6 +868,19 @@ def replace_configuration(self, applications, deployment):
         except ConfigurationError as e:
             raise make_bad_request(code=BAD_REQUEST, description=unicode(e))
 
+    @app.route("/state/node_eras", methods=['GET'])
+    @user_documentation(
+        u"""
+        List the currently known nodes and their eras.
+
+        A node's era can be looked up by reading
+        /proc/sys/kernel/random/boot_id. If possible use eras to identify
```
exarkun: Since we're going to encourage people to look up this value, should we provide an abstraction over `boot_id` so we can make changes without disrupting everyone, if necessary?

itamarst: `flocker-era-id` command line tool, maybe?

flocker/control/_protocol.py
```diff
@@ -734,6 +749,7 @@ def connectionMade(self):
         AMP.connectionMade(self)
         self.agent.connected(self)
         self._pinger.start(self, PING_INTERVAL)
+        # Send SetNodeEra using value of /proc/sys/kernel/random/boot_id
```
exarkun: If someone continues to deploy using our container agent, then is it the case that this design fixes the container agent misbehavior because in the previously problematic case where the container agent connected first and saw the dataset agent state from the previous boot, it would proceed to act incorrectly whereas now when it first connects it will send `SetNodeEra` with the new value which will wipe the dataset agent state from the previous boot, preventing it from seeing it and taking incorrect actions as a result?

I expected to see some era comparison in the `IDeployer.calculate_actions` implementation for the container agent too.  Perhaps those would be redundant with the fix here and that's why they're not in the design?  OTOH perhaps that redundancy wouldn't necessarily be bad.

Anyhow, have I understood correctly?

flocker/control/httpapi.py
exarkun: Yea, that's roughly what I had in mind.

exarkun:
Thanks.  Looks good, unless I've seriously misunderstood.  Please proceed (unless I've seriously misunderstood :wink: ).

flocker/control/_protocol.py
itamarst: Right, initial connection will result in old state being wiped. Might be a race condition there with updates from control service, so yeah, might want some check in the agent itself if the race condition isn't fixable.

It's worth noting that the worse symptom of the race condition, where we use a non-existent path, will be solved separately by FLOC-3254 if we can manage to implement it. My main focus in this branch is the Docker plugin and other REST API users since they are our main use case going forward.

PR #2128, opened 2015-11-04
[FLOC-3372] Vagrant tutorial box fix for Docker 1.9.0

sarum90:
It appears that docker 1.9.0 broke the Vagrant 'docker' provisioner on our tutorial's CentOS 7 box.

Currently the buildbot run for `flocker/vagrant/build/tutorial` has been red with failures that always end in:
```
==> default: Running provisioner: docker...
    default: Configuring Docker to autostart containers...
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

sed -i.bak 's/docker -d/docker -d -r=true/' /etc/sysconfig/docker 

Stdout from the command:



Stderr from the command:

sed: can't read /etc/sysconfig/docker: No such file or directory
Failed 1: vagrant up
```
This seems to indicate that the Vagrant 'docker' provisioner assumes that docker will have `/etc/sysconfig/docker`, which is evidently no longer true with docker 1.9.0 on CentOS 7.  Since we were only using the docker provisioner for the sake of starting docker, and pre-caching docker images, we can just do those things in a shell provisioner rather than relying on the Vagrant 'docker' provisioner.

Interestingly this is effectively just a flip-flop of https://github.com/ClusterHQ/flocker/commit/c4e34185327a101cc309055e4094615691fdc03d, although I cannot find any documentation as to why we switched to using the 'docker' provider. Hopefully we only started using it because it was so tempting and not because the shell script idea is fundamentally flawed.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2128)

### DIFF
--- a/vagrant/tutorial/Vagrantfile
+++ b/vagrant/tutorial/Vagrantfile
@@ -29,6 +29,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
                         ENV.fetch('FLOCKER_BUILD_SERVER', '')
                       ]
 
+  config.vm.provision :shell,
+                      path: 'setup_docker.sh',
+                      privileged: true
+
   # Cleanup
   config.vm.provision :shell, path: '../cleanup.sh', privileged: true
   #
--- a/vagrant/tutorial/Vagrantfile
+++ b/vagrant/tutorial/Vagrantfile
@@ -29,14 +29,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
                         ENV.fetch('FLOCKER_BUILD_SERVER', '')
                       ]
 
-  config.vm.provision 'docker', images: [
-    'busybox',
-    'clusterhq/mongodb',
-    'redis',
-    'python:2.7-slim',
-    'clusterhq/flask'
-  ]
-
   # Cleanup
   config.vm.provision :shell, path: '../cleanup.sh', privileged: true
   #
--- a/vagrant/tutorial/Vagrantfile
+++ b/vagrant/tutorial/Vagrantfile
@@ -29,6 +29,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
                         ENV.fetch('FLOCKER_BUILD_SERVER', '')
                       ]
 
+  # Despite the fact that the Vagrant 'docker' provider is tempting to use
+  # here, we instead use a shell script to start docker and pull down docker
+  # images. In the past we've had bad luck with the Vagrant 'docker' provider
+  # being fragile relative to new releases of docker running on CentOS
+  # (FLOC-3372).
   config.vm.provision :shell,
                       path: 'setup_docker.sh',
                       privileged: true
--- a/vagrant/tutorial/setup_docker.sh
+++ b/vagrant/tutorial/setup_docker.sh
@@ -13,7 +13,7 @@ systemctl restart docker
 docker --version
 
 # Pull docker images so they are cached for later.
-IMAGES=(busybox clusterhq/mongodb redis python:2.7-slim clusterhq/flask)
+IMAGES="busybox clusterhq/mongodb redis python:2.7-slim clusterhq/flask"
 for I in ${IMAGES}
 do
     echo Pulling image ${I}...
--- a/vagrant/tutorial/setup_docker.sh
+++ b/vagrant/tutorial/setup_docker.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+#
+# Small script to start the docker daemon and pull docker images so they are
+# cached for use in the tutorial later.
+
+# Error on any failure of a simple command.
+set -e
+
+# Start or restart the docker daemon.
+systemctl restart docker
+
+# Record the version of docker for the logs.
+echo "Docker Version:"
+docker --version
+
+# Pull docker images so they are cached for later.
+IMAGES="busybox"
+IMAGES+=" clusterhq/mongodb"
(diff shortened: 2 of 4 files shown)


vagrant/tutorial/setup_docker.sh
```diff
@@ -0,0 +1,26 @@
...
+# Record the version of docker for the logs.
+echo "Docker Version:"
+docker --version
+
+# Pull docker images so they are cached for later.
+IMAGES="busybox"
+IMAGES+=" clusterhq/mongodb"
+IMAGES+=" redis"
+IMAGES+=" python:2.7-slim"
+IMAGES+=" clusterhq/flask"
```
exarkun: IMAGES=(busybox clusterhq/mongodb redis python:2.7-slim clusterhq/flask)

vagrant/tutorial/setup_docker.sh
```diff
@@ -0,0 +1,26 @@
...
+for I in $IMAGES
```
exarkun: ${IMAGES}

vagrant/tutorial/setup_docker.sh
```diff
@@ -0,0 +1,26 @@
...
+do
+    echo Pulling image $I...
```
exarkun: ${I}

vagrant/tutorial/setup_docker.sh
```diff
@@ -0,0 +1,26 @@
...
+    docker pull $I
```
exarkun: ${I}

exarkun:
c4e34185327a101cc309055e4094615691fdc03d seems noteworthy (or the set up for the joke to which this is the punchline, not sure which)

vagrant/tutorial/Vagrantfile
```diff
@@ -29,13 +29,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
                       ]
 
-  config.vm.provision 'docker', images: [
-    'busybox',
-    'clusterhq/mongodb',
-    'redis',
-    'python:2.7-slim',
-    'clusterhq/flask'
-  ]
+  config.vm.provision :shell,
```
exarkun: Perhaps worth some commentary that the Vagrant Docker provisioner is fragile and we're intentionally not using it even though it's obviously the correct thing to use.

exarkun:
Thanks.  Please twiddle the shell script as suggested and add some warnings to try to keep us from flip-flopping like this /again/... then merge.

vagrant/tutorial/setup_docker.sh
sarum90: Done.

vagrant/tutorial/setup_docker.sh
sarum90: Done.

vagrant/tutorial/setup_docker.sh
sarum90: Done.

vagrant/tutorial/setup_docker.sh
sarum90: Done.

vagrant/tutorial/Vagrantfile
sarum90: Added.

sarum90:
Addressed comments (including adding referrence to the flip-flop to the pull request description), going to re-run the tests at least until the point where the shell script is actually run, then merge.

sarum90:
Not submitting: Looking at the logs: http://build.clusterhq.com/builders/flocker%2Fvagrant%2Fbuild%2Ftutorial/builds/2292/steps/build-base-box/logs/stdio

We are not pulling all of the images now. Doing some local hacking to figure out why.

vagrant/tutorial/setup_docker.sh
sarum90: Using quotes instead.

sarum90:
```
IMAGES=(busybox clusterhq/mongodb redis python:2.7-slim clusterhq/flask)
for I in ${IMAGES}
do
    echo HI ${I}
done
```
yields `HI busybox`

whereas:
```
IMAGES="busybox clusterhq/mongodb redis python:2.7-slim clusterhq/flask"
for I in ${IMAGES}
do
    echo HI ${I}
done
```
yields:
```
HI busybox
HI clusterhq/mongodb
HI redis
HI python:2.7-slim
HI clusterhq/flask
```

I'm going to use quotes instead of parens.

vagrant/tutorial/setup_docker.sh
sarum90: Decided to grep through the bash man page to figure out what was going on. We could have done this, but then below we would need ${IMAGES[*]} ( or "${IMAGES[@]}" might be even better). I'll keep that in mind for next time, although honestly the fact that ${IMAGES} just returns ${IMAGES[0]} makes me sort of not want to use bash Arrays.

For now leaving as quotes given that I've already restarted the tests.

sarum90:
Okay, latest run looks better: http://build.clusterhq.com/builders/flocker%2Fvagrant%2Fbuild%2Ftutorial/builds/2293/steps/build-base-box/logs/stdio

I'll be merging shortly.

exarkun:
> yields HI busybox

Oops, sorry.  Too late o'clock for writing shell.  I gave you the wrong expansion expression.  `X=(...)` defines an array.  To expand the full array you need `${X[*]}` instead of `${X}`.

Not a big deal.  Just posting for the sake of completeness.  Also, to have a chance to point out that shell sucks.

PR #2132, opened 2015-11-05
[FLOC-3378] Nodes have eras

itamarst:
Each node has an era, a unique identifier per node per boot. This information is sent to the control service when the agent connects.

Followup issues will use this to establish NodeState as being only valid in the context of a particular era (i.e. reboot).

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2132)


flocker/common/_era.py
```diff
@@ -0,0 +1,65 @@
...
+
+@flocker_standard_options
+class EraOptions(Options):
+    """
+    Command line options for ``flocker-node-era``.
+    """
+    longdesc = (
+        "Print the current node's era to stdout. The era is a unique"
+        "identifier per reboot per node, and can be used to discover the"
+        "current node's state safely using Flocker's REST API.\n"
```
jml: Nit: I think "safely" raises more questions than it answers, and should thus be removed from the description.

flocker/common/test/test_era.py
```diff
@@ -0,0 +1,39 @@
...
+from twisted.trial.unittest import SynchronousTestCase
+from twisted.python.runtime import platform
+from .._era import get_era
+
+
+class EraTests(SynchronousTestCase):
+    """
+    Tests for ``get_era``
+    """
+    @skipUnless(platform.isLinux(), "Only possible on Linux.")
```
jml: I'd say "get_era only supported on Linux". This sounds too much like a challenge that will only end with the BSD hordes descending upon you, awash in the glory of the their microkernel.

flocker/control/_protocol.py
```diff
@@ -240,6 +243,19 @@ class ClusterStatusCommand(Command):
     response = []
 
 
+class SetNodeEraCommand(Command):
+    """
+    Set the era for a node.
```
jml: As in, "Tell the control service that this is the era for a node" or as in, "Tell a node what it's era is"? Please update the docs to clarify.

flocker/control/_protocol.py
```diff
@@ -240,6 +243,19 @@ class ClusterStatusCommand(Command):
...
+
+    This should clear any previous NodeState that has a different
+    era. Updates to the node should only be sent after this command, to
+    ensure it doesn't get stale pre-reboot information (i.e. NodeState
+    with wrong era).
```
jml: Does this mean that if we send updates to the node before setting the era, we'll silently corrupt the data? Is there any way we can make this more fool-proof?

flocker/common/functional/test_era.py
```diff
@@ -0,0 +1,31 @@
...
+
+class FlockerNodeEraTests(make_script_tests(EXECUTABLE)):
+    """
+    Tests for ``flocker-node-era``.
+    """
+    @skipUnless(which(EXECUTABLE), EXECUTABLE + " not installed")
+    def setUp(self):
+        pass
+
+    def test_output(self):
```
jml: This is only supported on Linux, and should be skipped otherwise. Currently causing failing tests: http://build.clusterhq.com/builders/flocker-osx-10.10/builds/6116/steps/trial/logs/problems

flocker/common/functional/test_era.py
```diff
@@ -0,0 +1,31 @@
...
+from ...testtools import make_script_tests
+
+EXECUTABLE = b"flocker-node-era"
+
+
+class FlockerNodeEraTests(make_script_tests(EXECUTABLE)):
+    """
+    Tests for ``flocker-node-era``.
+    """
+    @skipUnless(which(EXECUTABLE), EXECUTABLE + " not installed")
```
jml: We're installing `flocker-node-era` on non-Linux platforms where it is near guaranteed not to work, e.g. http://build.clusterhq.com/builders/flocker-osx-10.10/builds/6116/steps/trial/logs/problems

We should either:
1. Not install it
2. Have it fail gracefully with a 'not supported on this platform' error message

2. feels cleaner to me.

jml:
Nice. This era stuff promises to be a solid foundation. 

In summary:
* couple of doc / comment tweaks
* handle `flocker-node-era` command on non-Linux platforms
* design concern about `SetNodeEraCommand` which doesn't need to be addressed in this PR.

The second one is probably enough for a quick resubmit.

flocker/control/_protocol.py
itamarst: Yep. See https://clusterhq.atlassian.net/browse/FLOC-3380, and parent issue for the full plan.

flocker/common/functional/test_era.py
itamarst: Flocker doesn't work on non-Linux. The fact you can install it is merely a convenience for developers. This seems like a waste of time to me.

itamarst:
We're going to be using more and more Linux-specific functionality. I really don't see why it's worth any effort at all to make Flocker even pretend it might work elsewhere. `flocker-node` packages, where `flocker-node-era` will go, will only be installed on Linux. Skipping tests is fine, since developers use Linux. Trying to make Flocker fail nicely elsewhere is silly.

flocker/control/_protocol.py
jml: Thanks!

jml:
I'm sorry that you think I'm being silly and asking you to waste your time.

Please fix the documentation and skip the failing test on non-linux platforms so homebrew & osx-10.10 builds can pass. 

The failure at http://build.clusterhq.com/builders/flocker%2Ffunctional%2Faws%2Fubuntu-14.04%2Fstorage-driver/builds/3369 is a new one to me. I don't _think_ it's related to your branch.

jml:
Thanks!

In future, please feel free to ping me for follow-up reviews. I'm more than happy to be interrupted if it unblocks someone.

jml:
Ugh sorry forgot to hit the Approved button. Multitasking is the mindkiller.

PR #2133, opened 2015-11-06
[FLOC-3272] Flongle profiles + acceptance test

sarum90:
This also completes FLOC-3274.

With this PR, the Flocker plugin for Docker can now accept a "profile" option if it is being used with docker version 1.9.0+. The added test verifies that if the 'silver' profile is chosen with the create_volume request using the EBS backend, we get the expected volume type for EBS u'silver' profile.

Unfortunately docker-py 1.5.0 does not have full support for create_volume (it is merged at head), so there is some odd code that explicitly uses internals to that library to construct the correct request to docker.

Also this turns on the acceptance test, which revealed that the proxy BlockDeviceAPI was hiding the fact that the EBS driver implemented IProfiledBlockDeviceAPI, so that is fixed as well.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2133)

### DIFF
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -5,7 +5,6 @@ Tests for the datasets REST API.
 """
 
 from uuid import UUID
-from unittest import skipUnless
 
 from twisted.internet import reactor
 from twisted.trial.unittest import TestCase
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -16,8 +16,6 @@ from ..testtools import (
     require_cluster, require_moving_backend, create_dataset, DatasetBackend
 )
 
-STORAGE_PROFILES_IMPLEMENTED = False
-
 
 class DatasetAPITests(TestCase):
     """
@@ -30,8 +28,6 @@ class DatasetAPITests(TestCase):
         """
         return create_dataset(self, cluster)
 
-    @skipUnless(STORAGE_PROFILES_IMPLEMENTED,
-                "Flocker storage profiles are not implemented yet.")
     @require_cluster(1, required_backend=DatasetBackend.aws)
     def test_dataset_creation_with_gold_profile(self, cluster, backend):
         """
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -19,7 +19,7 @@ from ...testtools import (
 )
 from ..testtools import (
     require_cluster, post_http_server, assert_http_server,
-    get_docker_client, verify_socket, check_http_server,
+    get_docker_client, verify_socket, check_http_server, DatasetBackend
 )
 from ..scripts import SCRIPTS
 
@@ -153,6 +153,46 @@ class DockerPluginTests(TestCase):
         """
         self.require_docker('1.9.0', cluster)
         return self._test_create_container(cluster)
+
+    @require_cluster(1, required_backend=DatasetBackend.aws)
+    def test_create_profiled_container_with_v2_plugin_api(self, cluster,
+                                                          backend):
+        """
+        Docker >=1.9, using the v2 plugin API, can create a volume with the
+        volumes API, and pass a profile argument which is represented in the
+        backing volume created.
+        """
+        self.require_docker('1.9.0', cluster)
+        node = cluster.nodes[0]
+        docker = get_docker_client(cluster, node.public_address)
+        volume_name = random_name(self)
+        docker.create_volume(volume_name, driver=u'flocker',
+                             driver_opts={'profile': 'silver'})
+
+        datasets_deferred = cluster.client.list_datasets_configuration()
+
+        def _evaluate_datasets(datasets):
+            dataset = None
+            for d in datasets:
+                for key, value in d.metadata.iteritems():
+                    if key.lower() == u"name" and value == volume_name:
+                        dataset = d
+                        break
+                if dataset:
+                    break
+            self.assertTrue(dataset is not None)
+
+            volumes = backend.list_volumes()
+
+            for volume in volumes:
+                if volume.dataset_id == dataset.dataset_id:
+                    break
+            ebs_volume = backend._get_ebs_volume(volume.blockdevice_id)
+            self.assertEqual(u'gp2', ebs_volume.type)
+
+        datasets_deferred.addCallback(_evaluate_datasets)
+
+        return datasets_deferred
 
     @require_cluster(1)
     def test_volume_persists_restart(self, cluster):
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -128,7 +128,7 @@ class DockerPluginTests(TestCase):
         """
         # XXX: replace with:
         #
-        # client.create_volume(name, u'flocker', driver_opts)
+        # return client.create_volume(name, u'flocker', driver_opts)
         #
         # Once version 1.5.1+ of docker-py is released. It is currently fixed
         # at head, but version 1.5.0 unfortunately has the wrong endpoint for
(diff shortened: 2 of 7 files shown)


jml:
Thanks Marcus!

Looks good. I've made a few comments about the test, because I found it a bit hard to follow. Please address them and merge.


---

Reviewed 7 of 7 files at r1.

---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 201 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2REiZDTgdHys13y318)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L201)):</sup>
I'm not 100% clear on what this is evaluating. 

1. From the dataset configuration, get the dataset with the given name.
2. From the backend, get the volume associated with that dataset
3. Assert that the volume type is equal to `gp2`

OK, now that I write it out like that it's a bit more clear—maybe that could go into a comment or docstring? Also, I don't know what `gp2` is or why it's the right answer—could you please add commentary explaining why (or linking to somewhere that does)? 


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 207 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RCWYa1g27wszmAOy3)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L207)):</sup>
I _think_ you can replace this inner loop with:

```python
if d.metadata.get(u'name', None) == volume_name:
    dataset = d
    break
```

Unless I'm wrong, you should probably do that.

Optionally, if you want to get all crazy functional you could replace both loops with:

```python
dataset = (d for d in datasets if d.metadata.get(u'name', None) == volume_name).next()
```

Which I personally & subjectively think is more clear.


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 210 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RDSYOZUfyuaB22-aR)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L210)):</sup>
`self.assertIsNotNone(dataset)`


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 218 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RDXcJ4aKVCk_fW6fs)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L218)):</sup>
:grey_question: Again optionally change this to a generator expression and call to `.next()`. Also has the advantage of not polluting the namespace. 

Vaguely surprised there's not an existing helper for this.


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 222 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RET2YoSXT-XTe4xoq)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L222)):</sup>
:grey_question: This smells like a deficiency in some API somewhere. Why can't we get the volume type using public methods?

Probably shouldn't be fixed in this branch.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2133)

sarum90:
Thanks for the review, will be addressing shortly. Although I'm afraid I will also have to address the failing tests. I'm suspect that I inadvertently coded a race-condition in my test which I won when I tested it locally, but lost overnight on buildbot. I'm going to try to remove that race condition along with the address and merge.

sarum90:
Pushing updates and re-running tests.


---

Review status: 0 of 7 files reviewed at latest revision, 5 unresolved discussions.

---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 201 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2REiZDTgdHys13y318)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L201)):</sup>
Added comment, changed function name, used a constant from a different file so whoever looks at this file doesn't even have to see "gp2".


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 207 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RCWYa1g27wszmAOy3)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L207)):</sup>
Yea I like that more as well.


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 210 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RDSYOZUfyuaB22-aR)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L210)):</sup>
Unneeded with above construction I think :)


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 218 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RDXcJ4aKVCk_fW6fs)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L218)):</sup>
Done, also next(x for x in Y if z(x)) is another nice way I ripped from http://stackoverflow.com/questions/9542738/python-find-in-list


---

<sup>**[flocker/acceptance/endtoend/test_dockerplugin.py, line 222 \[r1\]](https://reviewable.io:443/reviews/clusterhq/flocker/2133#-K2RET2YoSXT-XTe4xoq)** ([raw file](https://github.com/clusterhq/flocker/blob/831b0677322c5be9fcae632f8c05d4aee943e3da/flocker/acceptance/endtoend/test_dockerplugin.py#L222)):</sup>
+1 and known future work :smile:  The problem is it's actually not obvious what that API should look like, so we didn't want to rush into forcing the wrong API on our driver vendors.

Specifically, the naive implementation might be, "what profile does volume X have?" but considering we want to be able to let people change what a given profile means in the future, there is then the follow up question, does that mean, "What profile was volume X created with?" or "What is the set of profiles that volume X currently meets?".

I think the correct API endpoint is "Does volume X meet profile Y" but needed more time between launch and dockerCon to realize that and go through design review, etc.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/clusterhq/flocker/2133)

sarum90:
Also I wanted to post here about why the tests failed on buildbot.

They failed either from AWS eventual consistency or because docker /volumes/create is an async endpoint (i.e. it might respond before the volume is fully created). I thought about digging into the nitty gritty of it, but I realized that just in case /volumes/create is async or eventually becomes async, we might as well create a docker container using that volume to ensure it really is created before we start inspecting datasets.

I ran the test 5+ times with this addition and didn't see any failures. So even if it was AWS eventual consistency, it seems waiting for a container to be brought up is time enough to be eventual enough for AWS.

sarum90:
Tests are green so I'm going to merge. Thanks again for the review Jono.

wallnerryan:
FYI, I know this is merged already by just for your info attempts to call a method on a plugin are retried with an exponential backoff for up to 30 seconds

sarum90:
Cool, thanks @wallnerryan! I'm sure I could also grep through docker code for this... but do you happen to know off hand if "docker create volume ..." returns immediately or only after the volume plugin has returned (potentially after 30 seconds if the above mentioned exponential backoff has concluded)?

I guess I'd have to dig deeper anyway since I'm not sure if the flongle returns from a create request before or after the volume actually manifests... Either way it seems to me this is a safer test if I actually use the volume in a container before I start to query it. It must be manifest before it's used in a container so there's no chance any of the layers don't have it existing yet if we wait for a container to use it.

PR #2137, opened 2015-11-06
[FLOC-3391] Docs feedback

msh100:
The form works the same way as the ones on the marketing website do.

When a user submits a form it sends an HTTP request with Javascript to Formstack. We have a formstack form made for documentation feedback and the fields are the IDs defined in the html file.
If the user doesn't have Javascript, the form will just post as normal.

Once the entry is at Formstack, Zapier detects this and creates a new Zendesk ticket.

Fixes FLOC-3391

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2137)


jml:
:+1: Thanks!

PR #2144, opened 2015-11-09
[FLOC 3251] add key pair step to AWS instructions

davidoliverSP2:
Fixes 3251

Adds a step to our instructions on how to create nodes using AWS. Previously we did not mention there is a step on choosing an existing key pair or create a new one.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2144)


jml:
:+1: Thanks!

PR #2145, opened 2015-11-09
added missing tests to run on jenkins - FLOC-3405

Azulinho:
https://clusterhq.atlassian.net/browse/FLOC-3405

A number of tests are missing from Jenkins, these were created after
we took the initial snapshot of jobs running on BuildBot.
This PR adds those missing tests.

jenkins builds running here:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/add_missing_tests_to_jenkins_FLOC-3405/

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -767,14 +767,12 @@ run_trial_modules: &run_trial_modules
   - flocker.cli
   - flocker.common
   - flocker.control
+  - flocker.restapi
+  - flocker.docs
   - flocker.dockerplugin
   - flocker.node.agents
   - flocker.node.test
-  # TODO:
-  # one of the functional tests is hanging, so we split the functional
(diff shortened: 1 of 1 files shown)


jml:
:+1: OK for this code to land.

However, separately, this patch teaches me that in Jenkins, we'll have to add test modules in two separate places: once as the file itself and once in `build.yaml`. This is going to be a source of errors in the future, as people will add the file and forget to update `build.yaml`. Do we have a plan for fixing / mitigating this?

(Note that this patch can land independently of whether we have such a thing).

Azulinho:
@jml we were aware of that issue from the start, no resolution yet.
will ping you on slack to brainstorm how that could be tackled

thanks for the review

PR #2146, opened 2015-11-09
[FLOC 3250] Make the AMI step of AWS instructions clearer

davidoliverSP2:
Fixes 3250

We discovered that the AMI links we provided were CentOS 7 examples, which were correct, but not the typical process of creating instances on EC2.

This change directs users to the EC2 page they'd expect, and moves the CentOS 7 instructions into a note, highlighting that this is slightly different to what they'd might expect.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2146)

### DIFF
--- a/docs/install/setup-aws.rst
+++ b/docs/install/setup-aws.rst
@@ -4,14 +4,23 @@
 Setting Up Nodes Using Amazon Web Services
 ==========================================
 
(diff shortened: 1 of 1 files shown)


jml:
Lovely, thanks.

PR #2149, opened 2015-11-09
[FLOC-3385] Explicitly discover state in BlockDeviceDeployer

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3385

This just changes `discover_state` to explicitly discover state. There will be a follow up to switch calculate changes to use the new functions.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2149)


flocker/node/agents/blockdevice.py
```diff
@@ -55,6 +58,45 @@
...
 
 
+class DatasetStates(Names):
+    NON_MANIFEST = NamedConstant()
+    ATTACHED = NamedConstant()
+    MOUNTED = NamedConstant()
+    DELETED = NamedConstant()
+
+
+class DiscoveredDataset(PClass):
```
tomprince: This and the above need better docstrings.

flocker/node/agents/blockdevice.py
```diff
@@ -55,6 +58,45 @@
...
+    maximum_size = field(type=int, mandatory=True)
+    blockdevice_id = field(type=unicode, mandatory=True)
+    device_path = field(FilePath)
+    mount_point = field(FilePath)
+
+    def __invariant__(self):
+        expected_attributes = [
+            ((DatasetStates.ATTACHED, DatasetStates.MOUNTED),
+             "device_path",
+             "device_path only for attached and mounted xDataset."),
```
tomprince: This needs better messages.

flocker/node/agents/blockdevice.py
```diff
@@ -1173,50 +1282,14 @@ def async_block_device_api(self):
...
-            FilePath(partition.mountpoint):
-                device_to_dataset_id[FilePath(partition.device)]
-            for partition
-            in partitions
-            if FilePath(partition.device) in device_to_dataset_id
-        }
-
-    def discover_state(self, node_state):
+    def _discover_raw_state(self):
         """
```
tomprince: docstring

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -391,32 +549,74 @@ def assert_discovered_state(case,
...
+                for dataset in expected_discoved_datasets
+            },
+            volumes=expected_volumes,
+        )
+    )
+
+
+class BlockDeviceDeployerDiscoverRawStateTests(SynchronousTestCase):
+    """
+    Tests for ``BlockDeviceDeployer._discover_raw_state``.
```
tomprince: I'm not sure how much (if any) tests the warrants.

One difficulty with testing this, is that it reports state that isn't under our control as well.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -865,6 +1026,76 @@ class DiscoverVolumesMethod(Names):
     INFER_VOLUMES_FROM_STATE = NamedConstant()
 
 
+@implementer(ILocalState)
+class FakeBlockDeviceDeployerLocalState(PClass):
```
tomprince: This is definitely not ideal. But the tests below were copied from the real tests and just adjusted enough to construct this. I wanted to avoid making this PR bigger by adjusting the `calculate_changes` tests as well. That will be done in a follow up.

flocker/node/agents/blockdevice.py
```diff
@@ -55,6 +58,70 @@
...
+    state = field(
+        invariant=lambda state: (state in DatasetStates.iterconstants(),
+                                 "Not a valid state"),
+        mandatory=True,
+    )
+    dataset_id = field(type=UUID, mandatory=True)
+    maximum_size = field(type=int, mandatory=True)
+    blockdevice_id = field(type=unicode, mandatory=True)
+    device_path = field(FilePath)
+    mount_point = field(FilePath)
```
itamarst: Maybe set these two to `None` if unavailable, instead of not being set at all?

flocker/node/agents/blockdevice.py
```diff
@@ -1116,7 +1224,46 @@ def shared_state_changes(self):
...
+            if dataset.state == DatasetStates.MOUNTED:
+                manifestations[unicode(dataset_id)] = Manifestation(
+                    dataset=Dataset(
+                        dataset_id=dataset_id,
+                        maximum_size=dataset.maximum_size,
+                    ),
+                    primary=True,
+                )
+                paths[unicode(dataset_id)] = dataset.mount_point
+            if dataset.state in (
```
itamarst: `elif` for clarity.

flocker/node/agents/blockdevice.py
```diff
@@ -1226,77 +1341,78 @@ def is_existing_block_device(dataset_id, path):
...
-                        volume
-                    )
-            elif volume.attached_to is None:
-                # XXX: Looks like we don't attempt to report the size
-                # of non-manifest datasets.
-                # Why not? The size is available from the volume.
-                # https://clusterhq.atlassian.net/browse/FLOC-1983
-                nonmanifest[u_dataset_id] = Dataset(dataset_id=dataset_id)
+                else:
+                    # XXX We will detect this as NON_MANIFEST
```
itamarst: Was this XXX handled?

flocker/node/agents/blockdevice.py
```diff
@@ -1092,12 +1175,35 @@ def _manifestation_from_volume(volume):
...
+class RawState(PClass):
+    """
+    The raw state of a node.
+
+    :param unicode compute_instance_id:
+    :param volumes: List of volumes attached to this node or non-manifest.
+    :type volumes: ``pvector`` of ``BlockDeviceVolume:
+    :param devices: Mapping from datasets to block device path containing
+        filesystem of that dataset.
+    :type devices: ``pmamp`` of ``UUID`` to ``FilePath``
```
itamarst: pmap, and maybe mention it's node UUID.

flocker/node/agents/blockdevice.py
```diff
@@ -1092,12 +1175,35 @@ def _manifestation_from_volume(volume):
...
+    :param system_mounts: Mapping of block device path to mount point of all
+        mounts on the system.
+    :type system_mounts: ``pmap`` of ``FilePath`` to ``FilePath``.
```
itamarst: What are the two FilePaths?

flocker/node/agents/blockdevice.py
```diff
@@ -1116,7 +1224,46 @@ def shared_state_changes(self):
         These are the only parts of the state that need to be sent to the
         control service.
         """
-        return (self.node_state, self.nonmanifest_datasets)
+        manifestations = {}
```
itamarst: So this function creates `devices` and `paths` from low-level data even though `discover_state()` already had that information in the `RawState` object. Couldn't `discover_state` just pass those in?

itamarst:
I'm excited to have explicit states exposed!

Please:
1. Address comments above, in particular idea of simplifying design by having `discover_state` just do the low-level "raw" discovery and moving all the object construction logic into `BlockDeviceDeployerState`. In addition to removing need for code to reconstruct `paths` and `devices`, insofar as the state object will have direct access to them, I suspect this would also allow getting rid of `RawState` class.
2. Make the tests pass.
3. Resubmit.

flocker/node/agents/blockdevice.py
tomprince: The corresponding data in `RawState` isn't in the same format as expected here.

tomprince: Actually, I guess `devices` is. But, as far as I can tell `NodeState.devices` is just used to pass data between `discover_state` and `calculate_changes`. I didn't want to get rid of `NodeState.devices` here, but I think that we should eventually do that. Given that, I think keeping the separation we have here seems reasonable.

flocker/node/agents/blockdevice.py
tomprince: I wonder if it is better to immediately raise an exception if these are accessed when they shouldn't, or later when trying to manipulate `None` as a `FilePath`. My inclination is that the former is better.

itamarst:
Tests are still failing. Or is that the previous run?

flocker/node/agents/blockdevice.py
itamarst: Ah, so it's a temporary step. I guess add a comment to that effect then.

flocker/node/agents/blockdevice.py
tomprince: No. I've expanded the comment, though. (out-of-scope)

tomprince: No. I've expanded the comment, though. (out-of-scope)

tomprince:
Putting this back up for review, on the assumption that the tests will pass. (The affected CentOS tests have already passed).

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -275,31 +277,187 @@ class BlockDeviceDeployerLocalStateTests(SynchronousTestCase):
...
+            )
+        )
+        self.assertEqual(
+            local_state.shared_state_changes(),
+            expected_changes,
+        )
+
+    def test_non_manifest_dataset(self):
+        """
+        When there is a a dataset in the ``NON_MANIFEST`` state,
```
sarum90: s/a a/a/g

flocker/node/agents/blockdevice.py
sarum90: Do you still intend on fixing these up before merging? Just pointing it out.

I think at least documenting the constants above makes sense.

flocker/node/agents/blockdevice.py
sarum90: +1 for raising an exception on access rather than manipulating None as a FilePath. Sort of seems like we could have different PClasses for datasets in different states.... but I think I like your approach better.

flocker/node/agents/blockdevice.py
```diff
@@ -1078,6 +1145,22 @@ def get_blockdevice_volume(api, blockdevice_id):
     raise UnknownVolume(blockdevice_id)
 
 
+def get_system_mounts():
+    """
+    Load information about mounted filesystems.  volumes.
```
sarum90: I'm guessing you wanted this to be "filesystems and volumes" rather than "filesystems.  volumes." but I could be wrong.

flocker/node/agents/blockdevice.py
```diff
@@ -1092,12 +1175,35 @@ def _manifestation_from_volume(volume):
     return Manifestation(dataset=dataset, primary=True)
 
 
+class RawState(PClass):
+    """
+    The raw state of a node.
+
+    :param unicode compute_instance_id:
+    :param volumes: List of volumes attached to this node or non-manifest.
+    :type volumes: ``pvector`` of ``BlockDeviceVolume:
```
sarum90: s/ ``BlockDeviceVolume:/ ``BlockDeviceVolume``./

flocker/node/agents/blockdevice.py
```diff
@@ -1092,12 +1175,35 @@ def _manifestation_from_volume(volume):
...
     """
     An ``ILocalState`` implementation for the ``BlockDeviceDeployer``.
 
-    :ivar NodeState node_state: The current ``NodeState`` for this node.
+    :ivar unicode hostname: The IP address of the node that has this is the
+        state for.
+    :ivar UUID node_uuid: The UUID of the node that this is the state for.
+    :ivar datasets: The datasets discovered from this node.
 
     :ivar NonManifestDatasets nonmanifest_datasets: The current
```
sarum90: This is also gone now, yea?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -275,31 +277,187 @@ class BlockDeviceDeployerLocalStateTests(SynchronousTestCase):
...
+        )
+        self.assertEqual(
+            local_state.shared_state_changes(),
+            expected_changes,
+        )
+
+    def test_mounted_dataset(self):
+        """
+        When there is a a dataset in the ``MOUNTED`` state,
+        it is reported as a non-manifest dataset.
```
sarum90: This comment doesn't look accurate, but the test does.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -275,31 +277,187 @@ class BlockDeviceDeployerLocalStateTests(SynchronousTestCase):
...
+                dataset_id: DiscoveredDataset(
+                    state=DatasetStates.NON_MANIFEST,
+                    dataset_id=dataset_id,
+                    blockdevice_id=ARBITRARY_BLOCKDEVICE_ID,
+                    maximum_size=LOOPBACK_MINIMUM_ALLOCATABLE_SIZE,
+                ),
+            },
+        )
+        expected_changes = (
+            NodeState(
```
sarum90: :grey_question: Consider putting this empty state in a member variable, and explicitly modifying it in future tests to what you are expecting. Concern about maintenance burden of adding a new parameter to NodeState, this would add 4 more separate places that we would have to run around to and add the parameter, instead of just 1.

I recognize that's probably just a matter of personal preference though so I'll leave it up to you.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -512,10 +769,12 @@ def _incorrect_device_path_test(self, bad_value):
         that must be wrong, the corresponding manifestation is not included in
         the discovered state for the node.
         """
+        # XXX This discovers volums as NON_MANIFEST, but we should
```
sarum90: s/volums/volumes/g

flocker/node/agents/test/test_blockdevice.py
sarum90: Thanks for keeping the PR small :) I agree it doesn't make sense to bite off more testing work til calculate_changes is modified. And this may not be ideal, but you'll probably be deleting it then anyway.

sarum90:
This looks great! Much cleaner and more accurate view of state. Seems like it'll be much easier to think about in the long run.

Mostly just doc nits here and there. 

I do have an overarching comment though, this feels like it should be more testable in smaller units and cleaner to think about, so I was surprised that the lines of code in the test file increases with this PR. Maybe I'm wrong - I guess just look for tests that are no longer meaningful as you go forward so we can remove them.

tomprince:
>  I do have an overarching comment though, this feels like it should be more testable in smaller units and cleaner to think about, so I was surprised that the lines of code in the test file increases with this PR. 

That is probably true in the fullness of time, but for the moment we are still interacting with things that expect the old model, so there is extra code to deal with that. Also this introduces some new model things, which also takes up lines.

PR #2141, opened 2015-11-06
[FLOC-3379] Control service learns about eras

itamarst:
1. The control service now handles node era update.
2. Sending of node era updates was moved elsewhere so we can easily get at the node UUID.

For full plan see https://clusterhq.atlassian.net/browse/FLOC-3261

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2141)

PR #2150, opened 2015-11-09
[FLOC-3381] Allow looking up node identity by era

itamarst:
This builds on #2141 which will hopefully be merged soon.

This adds a REST API for looking up nodes by era. The idea is that this is the new mechanism for figuring out "which node is this", and since eras are reboot-specific it's a mechanism that ensures clients of the REST API don't get stale node state.

For example, in the next stacked branch the Docker plugin will be changed to use this new API to discover its node identity, instead of extracting it from the node TLS certificate. This will mean the Docker plugin won't ever get stale pre-reboot state, since it won't be able to lookup node identity until an agent has updated the control service post-reboot.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2150)

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -7,7 +7,7 @@ Client for the Flocker REST API.
 from uuid import UUID, uuid4
 from json import dumps
 from datetime import datetime
-from subprocess import check_output
+from os import environ
 
 from ipaddr import IPv4Address, IPv6Address, IPAddress
 
@@ -23,12 +23,14 @@ from eliot.twisted import DeferredContext
 from twisted.internet.defer import succeed, fail
 from twisted.python.filepath import FilePath
 from twisted.web.http import CREATED, OK, CONFLICT, NOT_FOUND
-from twisted.internet.task import deferLater
+from twisted.internet.utils import getProcessOutput
 
 from treq import json_content, content
 
 from ..ca import treq_with_authentication
 from ..control import Leases as LeasesModel, LeaseError
+from ..common import retry_failure
+
 from .. import __version__
 
 _LOG_HTTP_REQUEST = ActionType(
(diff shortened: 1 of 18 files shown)


flocker/apiclient/_client.py
```diff
@@ -536,3 +556,17 @@ def to_nodes(result):
...
+            b"GET", b"/state/nodes/by_era/" + era, None, {OK},
+            {NOT_FOUND: NotFound},
+        )
+
+        def handle_error(failure):
+            failure.trap(NotFound)
+            return deferLater(self._reactor, 0.1, self.this_node_uuid)
+        request.addCallbacks(lambda result: UUID(result["uuid"]), handle_error)
+        return request
+
```
jml: This blank line causes lint failures.

flocker/apiclient/_client.py
```diff
@@ -536,3 +556,17 @@ def to_nodes(result):
         request.addCallback(to_nodes)
 
         return request
+
+    def this_node_uuid(self):
+        era = check_output(["flocker-node-era"])
```
itamarst: Possibly this should use a Twisted API.

On the other hand it only needs to be called once per process, at startup time.

jml: I think it would surprise most callers to learn that this method blocks. AFAICT, it's the only such method on the class. Please change to use an asynchronous subprocess runner.

flocker/apiclient/_client.py
```diff
@@ -536,3 +556,17 @@ def to_nodes(result):
...
+        request = self._request(
+            b"GET", b"/state/nodes/by_era/" + era, None, {OK},
+            {NOT_FOUND: NotFound},
+        )
+
+        def handle_error(failure):
+            failure.trap(NotFound)
+            return deferLater(self._reactor, 0.1, self.this_node_uuid)
+        request.addCallbacks(lambda result: UUID(result["uuid"]), handle_error)
```
jml: Please use `flocker.common.retry_failure` here, e.g.

```python
retry_failure(self._reactor, lambda: self._request(...), [NotFound], repeat(0.1))
```

Also, is it correct to retry calls to `flocker-node-era` if it's not found on the API?

itamarst: Cool. The subprocess result can be cached, actually, so no need to rerun it.

flocker/apiclient/_client.py
jml: Wrote my comment before I saw yours. I still think that it should use Twisted APIs here, since that's the least surprising thing.

jml:
Nice. Couple of changes required & then merge.

PR #2151, opened 2015-11-09
[FLOC-3380] No updates with wrong era

itamarst:
Builds on #2141.

When an agent receives an update it checks to see if the state ascribed to the current node has the appropriate era. If not that particular piece of state is obviously stale (pre-reboot) and should be discarded.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2151)


jml:
Nice, thanks!

PR #2152, opened 2015-11-09
[FLOC-3382] Better node identity for plugin

itamarst:
Now that we have a better API for getting node identity, switch the Docker plugin to use it.

This builds on #2150.

The only way to see if this actually works is passing acceptance tests, which will take a while. In particular `test_move_volume_different_node` relies on correct node identification.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2152)


flocker/control/httpapi.py
```diff
@@ -836,6 +837,45 @@ def list_current_nodes(self):
                 for node in
                 self.cluster_state_service.as_deployment().nodes]
 
+    @app.route("/state/nodes/by_era/<era>", methods=['GET'])
+    @user_documentation(
+        u"""
+        If you are talking to the API from a process on a node then this
```
sarum90: Do you want to actually define era explicitly? I think can figure it out from context, but if I was just looking at the docs as an external user this would be at best just noise that I'd have to filter out.

flocker/control/_model.py
```diff
@@ -890,12 +909,13 @@ def _field_missing(fields):
...
-        if "uuid" not in kwargs and "_precord_buckets" not in kwargs:
-            # To be removed in https://clusterhq.atlassian.net/browse/FLOC-1795
-            warn("UUID is required, this is for backwards compat with existing"
-                 " tests only. If you see this in production code that's "
-                 "a bug.", DeprecationWarning, stacklevel=2)
-            kwargs["uuid"] = ip_to_uuid(kwargs["hostname"])
+        if "_precord_buckets" not in kwargs:
+            if "uuid" not in kwargs:
+                # See https://clusterhq.atlassian.net/browse/FLOC-1795
+                warn("UUID is required, this is for backwards compat with "
```
sarum90: :grey_question: Net difference here is just changing the warning text and comment by a couple words, right? I think it's a cleaner diff with this reverted. (I'm guessing you were originally going to remove, but wound up not removing?)

flocker/control/_model.py
```diff
@@ -930,6 +950,38 @@ def get_information_wipe(self):
         return _WipeNodeState(node_uuid=self.uuid, attributes=attributes)
 
 
+@implementer(IClusterStateChange)
+class UpdateNodeStateEra(PClass):
+    """
+    Update a node's era.
```
sarum90: Perhaps mention the wiping the former era's state in the docstring as well.

flocker/control/test/test_model.py
```diff
@@ -1418,42 +1418,51 @@ def test_applying_removes_node(self):
...
         """
-        Applying the ``IClusterStateWipe`` does nothing to the cluster state.
+        Applying the ``NoWipe`` does nothing to the cluster state.
         """
-        # Cluster has some non-manifested datasets:
-        cluster_state = self.NON_MANIFEST.update_cluster_state(
-            DeploymentState())
+        non_manifest = NonManifestDatasets(datasets={MANIFESTATION.dataset_id:
+                                                     MANIFESTATION.dataset})
+        cluster_state = non_manifest.update_cluster_state(DeploymentState())
```
sarum90: :grey_question: The above three lines are now just a construction of an arbitrary non-null cluster_state, correct? I feel like this might be more maintainable if that chunk was separated out as a global or otherwise labelled as an arbitrary cluster_state. (So future maintainers realize there is nothing special about using NonManifestDatasets in this example.

Obviously ignore if I'm misreading this.

flocker/control/_model.py
```diff
@@ -930,6 +950,38 @@ def get_information_wipe(self):
...
+            cluster_state = cluster_state.set(
+                nodes={n for n in cluster_state.nodes
+                       if n.uuid != self.uuid})
+        cluster_state = cluster_state.transform(
+            ["node_uuid_to_era", self.uuid], self.era)
+        return cluster_state
+
+    def get_information_wipe(self):
+        """
+        Since we just deleted some information, there's nothing to wipe.
```
sarum90: Is this accurate? One could imaging a wipe implementation that removed the uuid -> era mapping. I think we're not doing that here because we expect the next incarnation to wipe it out for us instead of using the wiping mechanism here.

flocker/control/_model.py
```diff
@@ -930,6 +950,38 @@ def get_information_wipe(self):
...
+    uuid = field(type=UUID, mandatory=True)
+    era = field(type=UUID, mandatory=True)
+
+    def update_cluster_state(self, cluster_state):
+        """
+        Record the node's era and discard the ``NodeState`` if it doesn't
+        match the era.
+        """
+        if cluster_state.node_uuid_to_era.get(self.uuid) != self.era:
+            # Discard the NodeState:
```
sarum90: :grey_question: Question on future work: With this here, do we still have use for the wiping mechanism? it seems that this will let us remove that. (Also, if we can get rid of wiping a big +1 to that idea, as a newbie it always takes me awhile to wrap my head around why we need wiping, and it's very confusing that we only implement it for the 1 specific state change that needs it).

sarum90:
Looks good, Thanks! Took awhile to get my head around the era concept, but it looks like that all comes from earlier reviews anyway.

Is the long term plan to get rid of the Wipeout stuff once all this lands? I find that code rather confusing as a newbie to the codebase. Specifically because we simply don't implement it, even for ClusterStateChanges that could easily have reverse implementations. We don't implement it for good reason, but it's a confusing policy decision that is non-obvious to me as a newbie to the codebase.

Mild concern about https://github.com/ClusterHQ/flocker/blob/b5bb83a/flocker/control/_model.py#L953 blowing up with too many values to unpack in combination with FLOC-3380.

Otherwise just nits. Apologies if they are actually relevant to earlier PRs, with stacked PRs it's a bit harder to tell what was meant to be reviewed at this layer.

itamarst:
We do actually implement wipes, on `NodeState` specifically.

itamarst:
Ah, but would era+reboots mean we can get rid of expiration altogether... I would say no. A machine can go offline and never come back, so we need some way to get rid of its state.

itamarst:
The comments were on the stacked branches which were already merged, and none of them seemed like blocker so just going to merge this.

PR #2154, opened 2015-11-09
[FLOC-3414] No-op @flaky decorator

jml:
Part of the work to add a `@flaky` decorator to flocker.

This patch adds a decorator that does absolutely nothing, and then applies that decorator to all of the tests for which we have bugs saying they are flaky.

It has been a worthwhile exercise in that it has taught me about which tests are actually flaky (I already had a sense of it, but it's surprisingly clear once you look at what actually needs to _import_ `flaky`), and also because it has made the uses for the `@flaky` decorator more clear (see `XXX` notes in the diff).

I think having these markers in the code will help people with debugging (esp. @Azulinho with the Jenkins migration), so even though there's no behavioural change, I think there's enough added value to land.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2154)

### DIFF
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -10,6 +10,7 @@ from twisted.internet import reactor
 from twisted.trial.unittest import TestCase
 
 from ...common import loop_until
+from ...testtools import flaky
 
 from ..testtools import (
     require_cluster, require_moving_backend, create_dataset, DatasetBackend
@@ -20,6 +21,8 @@ class DatasetAPITests(TestCase):
     """
     Tests for the dataset API.
     """
+
+    @flaky('FLOC-3207')
     @require_cluster(1)
     def test_dataset_creation(self, cluster):
         """
@@ -51,6 +54,7 @@ class DatasetAPITests(TestCase):
         waiting_for_create.addCallback(confirm_gold)
         return waiting_for_create
 
+    @flaky('FLOC-3341')
     @require_moving_backend
     @require_cluster(2)
     def test_dataset_move(self, cluster):
@@ -78,6 +82,7 @@ class DatasetAPITests(TestCase):
         waiting_for_create.addCallback(move_dataset)
         return waiting_for_create
 
+    @flaky('FLOC-3196')
     @require_cluster(1)
     def test_dataset_deletion(self, cluster):
         """
--- a/flocker/acceptance/obsolete/test_containers.py
+++ b/flocker/acceptance/obsolete/test_containers.py
@@ -11,7 +11,7 @@ from twisted.internet.defer import gatherResults
 from twisted.trial.unittest import TestCase
 
 from ...common import loop_until
-from ...testtools import random_name
+from ...testtools import flaky, random_name
 from ..testtools import (
     require_cluster, require_moving_backend, create_dataset,
     create_python_container, verify_socket, post_http_server,
@@ -118,6 +118,7 @@ class ContainerAPITests(TestCase):
         )
         return d
 
+    @flaky('FLOC-2488')
     @require_moving_backend
     @require_cluster(2)
     def test_move_container_with_dataset(self, cluster):
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -40,7 +40,7 @@ from ..test.blockdevicefactory import (
     get_blockdeviceapi_with_cleanup, get_device_allocation_unit,
     get_minimum_allocatable_size, get_openstack_region_for_test,
 )
-from ....testtools import run_process
+from ....testtools import flaky, run_process
 
 from ..cinder import (
     get_keystone_session, get_cinder_v1_client, get_nova_v2_client,
@@ -170,6 +170,11 @@ class CinderBlockDeviceAPIInterfaceTests(
             cinder_client.volumes.get(
                 flocker_volume.blockdevice_id).display_name,
             u"flocker-{}".format(dataset_id))
+
+    @flaky('FLOC-3347')
+    def test_get_device_path_device(self):
+        super(CinderBlockDeviceAPIInterfaceTests,
+              self).test_get_device_path_device()
 
 
 class CinderHttpsTests(SynchronousTestCase):
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -173,8 +173,9 @@ class CinderBlockDeviceAPIInterfaceTests(
 
     @flaky('FLOC-3347')
     def test_get_device_path_device(self):
-        super(CinderBlockDeviceAPIInterfaceTests,
-              self).test_get_device_path_device()
+        return super(
+            CinderBlockDeviceAPIInterfaceTests,
+            self).test_get_device_path_device()
 
 
 class CinderHttpsTests(SynchronousTestCase):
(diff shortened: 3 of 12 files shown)


flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -10,6 +10,7 @@
 from twisted.trial.unittest import TestCase
 
 from ...common import loop_until
+from ...testtools import flaky
 
 from ..testtools import (
     require_cluster, require_moving_backend, create_dataset, DatasetBackend
```
Azulinho: this is confusing,
there are now two 'testtools' modules

is the intent to depreceate flocker/flocker/acceptance/testtools.py in future JIRAs ?

flocker/acceptance/obsolete/test_containers.py
```diff
@@ -11,7 +11,7 @@
 from twisted.trial.unittest import TestCase
 
 from ...common import loop_until
-from ...testtools import random_name
+from ...testtools import flaky, random_name
 from ..testtools import (
     require_cluster, require_moving_backend, create_dataset,
```
Azulinho: same comment as above,
having two modules with the same name is confusing, my first look didn't see the extra '.' on my laptop screen and assumed it was two different imports from the same module
I

flocker/test/test_testtools.py
```diff
@@ -1,4 +1,4 @@
-# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+# Copyright ClusterHQ Ltd.  See LICENSE file for details.
```
Azulinho: about time, innit?

flocker/node/agents/functional/test_cinder.py
```diff
@@ -171,6 +171,11 @@ def test_name(self):
                 flocker_volume.blockdevice_id).display_name,
             u"flocker-{}".format(dataset_id))
 
+    @flaky('FLOC-3347')
+    def test_get_device_path_device(self):
+        super(CinderBlockDeviceAPIInterfaceTests,
+              self).test_get_device_path_device()
+
```
Azulinho: it doesn't bother me personally, but I would assume someone else doing this review would ask for new tests to be added in a separate PR, as this PR is essentially about adding the @flaky decorator functionality.

flocker/testtools/__init__.py
```diff
@@ -47,11 +47,15 @@
...
+from ._flaky import flaky
 from .. import __version__
 from ..common.script import (
     FlockerScriptRunner, ICommandLineScript)
 
 
+# Export for other places to import. Be still, flake8.
+flaky
+
 # This is currently set to the minimum size for a SATA based Rackspace Cloud
```
Azulinho: confused, what would this do when called like this ?

flocker/testtools/test/test_flaky.py
```diff
@@ -0,0 +1,104 @@
...
+from .._flaky import flaky
+
+
+class FlakyTests(SynchronousTestCase):
+    """
+    Tests for ``@flaky`` decorator.
+    """
+
+    @given(integers())
+    def test_decorated_function_executed(self, x):
```
Azulinho: you just forced me to readthedocs on hipothesis....
reading...

Azulinho: sold.
thanks for forcing me to read on hipothesis

flocker/testtools/_flaky.py
```diff
@@ -0,0 +1,31 @@
...
+
+    # TODO (see FLOC-3281):
+    # - allow specifying which exceptions are expected
+    # - retry on expected exceptions
+    #   - parametrize on retry options
+    # - provide interesting & parseable logs for flaky tests
+
+    def wrapper(test_method):
+        return test_method
+    return wrapper
```
Azulinho: was expecting something exciting...

flocker/acceptance/endtoend/test_dataset.py
exarkun: More than two

```
$ find flocker/ -name testtools.py
flocker/node/testtools.py
flocker/node/agents/testtools.py
flocker/volume/testtools.py
flocker/ca/testtools.py
flocker/restapi/testtools.py
flocker/acceptance/testtools.py
flocker/acceptance/integration/testtools.py
```

flocker/testtools/__init__.py
exarkun: It doesn't do anything except trick the static checker (flake8) into not emitting a particular warning (that a name is imported but unused).

I think putting the name in `__all__` is a better solution - since this is /also/ how you trick programmers into thinking that the name is being imported just so it can be re-exported. :wink: 

In this case, because of the massive pile of code also in this `__init__`, defining an `__all__` is probably more annoying.  Maybe the massive pile should move, though.  But that doesn't sound like the kind of thing that should block this change.

flocker/acceptance/endtoend/test_dataset.py
Azulinho: okay, if you're happy reading that everyday and don't get confused than I am happy too.
my untrainned eyes to flocker code combined with a small laptop screen raised some doubts

exarkun: I think my point was more that it's even more confusing than you realized. :grinning:

flocker/testtools/__init__.py
Azulinho: gotcha, which explains the comment just above.
my Syntastic vim checker seems to eat it too.
if vim is happy I am happy

Azulinho:
thanks for @exarkun for stepping in and clarified some of my doubts
thanks for @jml for starting some work on this well needed @flaky
thanks @azulinho for looking into hypothesis (cool stuff)

quite a few failures due to timeouts and friends, I clicked rebuild on all the failed ones.
gerge it when mreen
or when you're happy with it

Azulinho:
oh, and your jenkins builds is here btw:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/no-op-flaky-decorator-FLOC-3414/job/__main_multijob/2/

flocker/acceptance/endtoend/test_dataset.py
jml: I agree that the relative imports make this more confusing than it needs to be. It's not too different from other large projects with multiple `utils` modules scattered everywhere.

flocker/acceptance/obsolete/test_containers.py
jml: Ack.

flocker/test/test_testtools.py
jml: Yep.

flocker/node/agents/functional/test_cinder.py
jml: This doesn't actually add a new test: the `super` means that it's calling an existing test method that's inherited from the base class. Normally, you wouldn't do this in Python. The reason we do it here is so that we have something to decorate: `test_get_device_path_device` isn't flaky for blockdeviceapi tests.

flocker/testtools/__init__.py
jml: Moved to `__all__`. 

Splitting up testtools would be a fun thing to do another time. I want to reduce the cost per PR a little first, though.

flocker/testtools/test/test_flaky.py
jml: It's awesome, isn't it?

flocker/testtools/_flaky.py
jml: Soon!

jml:
Sole remaining failure is a known flaky test. Merging.

PR #2162, opened 2015-11-11
[FLOC-3288] Benchmark driver

jongiddy:
This PR creates a basic benchmarking command. Using an existing Flocker cluster, the command allows the user to combine a scenario (e.g. specified rate of external load on the control server), an operation to be measured (read request, write request), and a form of measurement (wallclock time, CPU time, log file size).

The driver will setup the load scenario, and then run the operation, taking measurements before and after the operation runs.  The measurement is currently sampled 3 times.  The sample results are reported to stdout as a JSON structure.

The current PR provides the basic framework and one or two examples of each parameter. The current scenario is "no load", the current operations are "no-op" and "read-request", and the current measurement is "wallclock" time.

Additional issues in FLOC-2748 will introduce more values for these parameters. Untested versions of these additional values can be seen in the spike branch https://github.com/ClusterHQ/flocker/tree/benchmark-prototype-with-scenarios.

With no Flocker cluster, run the benchmark using:
```
python -m benchmark.benchmark_control --operation nop
```

This will display a JSON stanza like:
```json
{"client": {"username": "jon", "flocker_version": "1.6.1+111.g4cb5b53", "working_directory": "/home/jon/projects/clusterhq/flocker", "nodename": "ecuador", "platform": "Linux-3.13.0-66-generic-x86_64-with-Ubuntu-14.04-trusty", "date": "2015-11-11T18:11:19.173954"}, "measurement": "wallclock", "metric_name": "nop", "samples": [{"value": 1.0967254638671875e-05, "success": true}, {"value": 6.9141387939453125e-06, "success": true}, {"value": 7.152557373046875e-06, "success": true}], "server": {"flocker_version": "1.6.1+111.g4cb5b53"}}
```

Note, as well as the final JSON output to stdout, the script displays logging information to stderr.

If you have a Flocker cluster handy (and who doesn't), to get a genuine benchmark result, run:
```
python -m benchmark.benchmark_control --control=${FLOCKER_ACCEPTANCE_CONTROL_NODE} --certs=${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH}
```
to get a genuine benchmark result.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2162)


benchmark/benchmark_control.py
```diff
@@ -0,0 +1,49 @@
...
+
+from eliot import to_file
+
+from twisted.internet.task import react
+from twisted.python.usage import Options, UsageError
+
+from benchmark_driver import driver
+from benchmark_scenarios import supported_scenarios, default_scenario
+from benchmark_operations import supported_operations, default_operation
+from benchmark_measurements import supported_measurements, default_measurement
```
jml: Any particular reason for all of these modules to be prefixed w/ `benchmark_`? Since this is a package already, why not call them `_driver', `_scenarios`, etc.?

benchmark/benchmark_control.py
```diff
@@ -0,0 +1,49 @@
+import sys
```
jml: Please add module docstrings, here & in all other added files. Don't have to be anything fancy.

benchmark/benchmark_driver.py
```diff
@@ -0,0 +1,190 @@
...
+from benchmark_measurements import get_measurement
+from benchmark_scenarios import get_scenario
+
+
+def benchmark(measure, operation, scenario, n=3):
+    """
+    Perform sampling of the operation.
+
+    :param measure: A quantity to measure.
+    :param operation: An operation to perform.
```
jml: Does it really _perform_ the operation? AFAICT, it only calls `get_probe` on it. Maybe I'll understand better as I get further in.

benchmark/benchmark_driver.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def delete_container(self, *a, **kw):
+        result = self.original.delete_container(*a, **kw)
+        self.original.synchronize_state()
+        return result
+
+
+def driver(reactor, control_service_address=None, cert_directory=b"certs",
+           metric_name=b"read-request", measurement_name=b"wallclock",
+           scenario_name=b'ten_ro_req_sec'):
```
jml: Please add a docstring.

benchmark/benchmark_driver.py
```diff
@@ -0,0 +1,190 @@
...
+    metric = get_operation(client=client, name=metric_name)
+    measurement = get_measurement(
+        clock=reactor, client=client, name=measurement_name,
+    )
+    scenario = get_scenario(
+        clock=reactor, client=client, name=scenario_name,
+    )
+    version = client.version()
+
+    d = gather_deferreds((metric, measurement, scenario, version))
```
jml: :grey_question: Looks like `driver` takes human-inputtable values, turns them into domain objects and then actually starts driving, I'd consider factoring out a separate function that takes the actual domain objects, and having `driver` call that.

benchmark/benchmark_driver.py
```diff
@@ -0,0 +1,190 @@
...
+        d.addCallback(
+            record_samples,
+            version[u"flocker"],
+            metric_name,
+            measurement_name
+        )
+        return d
+
+    d.addCallback(got_parameters)
+    # d.addErrback(lambda ignored: None)
```
jml: Please delete commented out code.

benchmark/benchmark_measurements.py
```diff
@@ -0,0 +1,34 @@
+from pyrsistent import PClass, field
```
jml: Since there's no actual thing in Python (e.g. a class) that defines what a measurement is, the docstring for this module should define it (or maybe there should be an interface). 

AFAICT, a measurement is a callable that takes a function and parameters, runs the function, discards the return value and returns the, umm, measurement, which has to be JSON-serializable. If the function raises an error, then that bubbles up. 

Is that right?

benchmark/benchmark_driver.py
```diff
@@ -0,0 +1,190 @@
...
+
+def benchmark(measure, operation, scenario, n=3):
+    """
+    Perform sampling of the operation.
+
+    :param measure: A quantity to measure.
+    :param operation: An operation to perform.
+    :param scenario: A load scenario.
+    :param int n: Number of samples to take.
+    """
```
jml: Please document the return value.

benchmark/benchmark_measurements.py
```diff
@@ -0,0 +1,34 @@
...
+
+_measurements = {
+    'wallclock': _WallClock,
+}
+
+supported_measurements = _measurements.keys()
+default_measurement = 'wallclock'
+
+
+def get_measurement(clock, client, name):
```
jml: Please add a docstring. I can see what it does, but I can't really tell what it's for (esp. what's 'client'?)

benchmark/benchmark_measurements.py
```diff
@@ -0,0 +1,34 @@
...
+
+        start = self.clock.seconds()
+        d = f(*a, **kw)
+        d.addCallback(finished)
+        return d
+
+
+_measurements = {
+    'wallclock': _WallClock,
+}
```
jml: Maybe make this a PMap? :grey_question:

benchmark/benchmark_measurements.py
```diff
@@ -0,0 +1,34 @@
...
+
+supported_measurements = _measurements.keys()
+default_measurement = 'wallclock'
```
jml: Shouldn't this & others be `CONSTANT_CASE`? I genuinely don't know.

benchmark/benchmark_operations.py
```diff
@@ -0,0 +1,110 @@
...
+    'nop': _NoOperation,
+    'read-request': _ReadRequestOperation,
+}
+
+supported_operations = _operations.keys()
+default_operation = 'read-request'
+
+
+def get_operation(client, name):
+    return maybeDeferred(_operations[name], client=client)
```
jml: :grey_question: This pattern (registry, supported list, default, get by name) is repeated three times, which is normally my trigger point for refactoring. I think it's simple enough to leave as-is, but if you wanted to make a `Registry` object or something I wouldn't stop you. (If you do though, I'd like a quick look before merging).

benchmark/benchmark_scenarios.py
```diff
@@ -0,0 +1,55 @@
+from twisted.internet.defer import maybeDeferred, Deferred, succeed
+
+
+class RunningScenario:
```
jml: Please add a docstring.

benchmark/benchmark_scenarios.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(self):
+        self.scenario_established = Deferred()
+        self.scenario_maintained = Deferred()
```
jml: Why prefix these with `scenario`? Shouldn't they be `_established` and `_maintained`? :grey_question:

benchmark/benchmark_operations.py
jml: If you make the PMap and/or `CONST_NAME` changes above, please make them here & in scenarios too.

benchmark/benchmark_measurements.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class _WallClock(PClass):
+    """
+    Measure the elapsed wallclock time during an operation.
+    """
+    clock = field(mandatory=True)
+    client = field()
+
+    def __call__(self, f, *a, **kw):
```
jml: :grey_question: Officially registering my unwarranted prejudice against `__call__`

benchmark/test/test_operations.py
```diff
@@ -0,0 +1,23 @@
+from zope.interface.verify import verifyObject
+
+from twisted.trial.unittest import TestCase
+
+from benchmark.benchmark_operations import _operations, IProbe, IOperation
+
+
+def check_interfaces(factory):
```
jml: Please add a docstring.

benchmark/test/test_operations.py
```diff
@@ -0,0 +1,23 @@
...
+            verifyObject(IOperation, operation)
+            probe = operation.get_probe()
+            verifyObject(IProbe, probe)
+
+    testname = '{}InterfaceTests'.format(factory.__name__)
+    OperationTests.__name__ = testname
+    globals()[testname] = OperationTests
+
+for factory in _operations.values():
+    check_interfaces(factory)
```
jml: Is this cleverness really needed? I guess it's a placeholder for future tests that will refine the interface of `IOperation` and `IProbe`? Maybe worth putting that in a comment.

docs/gettinginvolved/benchmarking.rst
```diff
@@ -0,0 +1,64 @@
+.. _benchmarking:
```
jml: I can't figure out how to make a comment on a "no changes" file.  `test_scenarios.py` appears to be missing.

benchmark/test/test_driver.py
```diff
@@ -0,0 +1,152 @@
...
+        """
+        self.measurements = measurements
+
+    def __call__(self, f, *a, **kw):
+        def finished(ignored):
+            return next(self.measurements)
+
+        d = f(*a, **kw)
+        d.addCallback(finished)
+        return d
```
jml: Maybe this should go into `measurements` itself, and become a verified fake? :grey_question:

benchmark/test/test_driver.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def __init__(self, succeeds):
+        """
+        :param succeeds: An iterator providing a boolean indicating
+            whether the probe will succeed.
+        """
+        self.succeeds = succeeds
+
+    def get_probe(self):
+        return FakeProbe(next(self.succeeds))
```
jml: If this pair were to go into `operations`, then I would suggest splitting `FakeProbe` in two: a trivial succeeding probe and a failing probe. `FakeOperation` would then take an iterable of probes, and `get_probe` would just return whatever was necessary.

I think that's cleaner than passing a boolean around, and so it might be worthwhile doing here. What's here is fine though.

:grey_question:

benchmark/benchmark_scenarios.py
```diff
@@ -0,0 +1,55 @@
...
+class _NoLoadScenario(object):
+
+    def __init__(self, clock, client):
+        self.clock = clock
+        self.client = client
+
+    def start(self):
+        running = RunningScenario()
+        running.established().callback(None)  # no setup needed
+        return running
```
jml: I just got rather confused when reading the tests. There's a thing called a "scenario" which just has a `start` method, and there's another thing called a "running scenario" which isn't a scenario but actually has a large chunk of the logic for a scenario? Is that right?

I'd strongly consider adding interfaces here, and maybe changing the names (although I don't have any good ideas right now). At least, please provide a module docstring that explains the core concepts.

benchmark/benchmark_driver.py
jml: I do understand better, the operation docs are very helpful in this regard. Please update the docstring to say that operation is an `IOperation`.

Please also consider renaming `n` to `num_samples`.

jml:
My caffeine levels dropped precipitously toward the end!

This is looking really good. I think most of my comments are around docstrings & renaming variables, with some optional ones around refactoring flagged with a grey question mark that I leave entirely up to you. 

It looks like test_scenarios is missing. I'd like a chance to look at that before merging.

benchmark/benchmark_control.py
jongiddy: Done.

benchmark/benchmark_operations.py
jongiddy: Structure is simpler now, so keeping as a dictionary, but have uppercased the names

benchmark/benchmark_measurements.py
jongiddy: Call to `__call__` eliminated with extreme prejudice.

benchmark/test/test_operations.py
jongiddy: I simplified it a bit, but this is how zope interfaces are meant to be tested, and the cleverness attempts to create a test for each registered operation, without burdening future developers.

jongiddy:
I think I have addressed the issues from the previous review. In order to make things a bit clearer, I restructured the files quite a bit. The scenarios, operations, and metrics are now in their own sub-directories, with each class in its own file.  I think this will keep things cleaner as we add more of these parameters.

I have moved more of the configuration parsing into the top-level file (`benchmark_control.py`) in order to make the lower modules more generic.

Lots more docstrings and some renaming of functions to try to make the code clearer.

The `nop` instruction has changed to `no-op`, so the no-Flocker-cluster test is now
```
python -m benchmark.benchmark_control --operation no-op
```

benchmark/scenarios/no_load.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        return succeed(None)  # no setup needed
+
+    def maintained(self):
+        """
+        :return: A Deferred that fires with an errback if the desired
+            scenario fails to hold between being established and being
+            stopped.  This Deferred never fires with a callback.
+        """
+        return self._maintained
```
jml: AFAICT, we don't need to store `self._maintained` at all? Would it be simpler / clearer to return `Deferred()`?

benchmark/test/test_driver.py
```diff
@@ -0,0 +1,145 @@
...
+        self.scenario_maintained = Deferred()
+
+    def start(self):
+        return succeed(None)
+
+    def maintained(self):
+        return self.scenario_maintained
+
+    def stop(self):
+        return succeed(None)
```
jml: This looks identical to `NoLoadScenario`? Could / should we use it instead?

benchmark/test/test_driver.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class FakeCollapsingScenario:
+
+    def __init__(self):
+        self.scenario_maintained = fail(RuntimeError('collapse'))
```
jml: Alternatively, we could combine `FakeScenario` & this, having them take a parameter that determines the result of `maintained`. :grey_question:

benchmark/test/test_driver.py
```diff
@@ -0,0 +1,145 @@
...
+        samples_ready = benchmark(
+            FakeMetric(count(5)),
+            FakeOperation(repeat(True)),
+            FakeScenario(),
+            3)
+
+        def check(samples):
+            self.assertEqual(
+                samples, [{'success': True, 'value': x} for x in [5, 6, 7]])
+        samples_ready.addCallback(check)
```
jml: AFAICT, we're not doing anything with the reactor. In which case, this should be `SynchronousTestCase`, and the tests should use `successResultOf` & friends.

benchmark/test/test_metrics.py
```diff
@@ -0,0 +1,23 @@
...
+
+    def test_wallclock(self):
+        """
+        Returns the difference in time from before and after call to
+        function.
+        """
+        clock = Clock()
+        wallclock = WallClock(clock=clock)
+        d = wallclock.measure(maybeDeferred, clock.advance, 1.23)
+        d.addCallback(self.assertEqual, 1.23)
```
jml: This needs to either `return d` or, if this isn't using the reactor, use `SynchronousTestCase`, which is faster & safer.

benchmark/test/test_operations.py
```diff
@@ -0,0 +1,102 @@
...
+    testname = '{}InterfaceTests'.format(factory.__name__)
+    OperationTests.__name__ = testname
+    globals()[testname] = OperationTests
+
+for factory in (NoOperation, ReadRequest):
+    check_interfaces(factory)
+
+
+class FastConvergingFakeFlockerClient(
+    proxyForInterface(IFlockerAPIV1Client)
```
jml: Please add a class docstring.

docs/gettinginvolved/benchmarking.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Benchmarking
+============
+
+Flocker includes a tool for benchmarking operations.
+It needs the appropriate ssh-key added to a running ssh-agent.
```
jml: How is a user going to figure out which SSH key is appropriate?

benchmark/test/test_operations.py
```diff
@@ -0,0 +1,102 @@
...
+    def delete_container(self, *a, **kw):
+        result = self.original.delete_container(*a, **kw)
+        self.original.synchronize_state()
+        return result
+
+
+class ReadRequestTests(TestCase):
+
+    def test_read_request(self):
+        control_service = FastConvergingFakeFlockerClient(FakeFlockerClient())
```
jml: Please add a class & method docstring

benchmark/test/test_operations.py
```diff
@@ -0,0 +1,102 @@
...
+            d.addCallback(cleanup)
+            return d
+        d.addCallback(run_probe)
+
+        # Check the cluster state has one dataset with the correct primary
+        def handle_results(states):
+            self.assertEqual([state.primary for state in states], [primary])
+        d.addCallback(handle_results)
+
+        return d
```
jml: If possible, please change this to use `SynchronousTestCase`.

docs/gettinginvolved/benchmarking.rst
```diff
@@ -0,0 +1,64 @@
...
+   Specifies the quantity to measure while the operation is performed.
+   Supported values include:
+
+      ``wallclock``
+         Actual clock time elapsed.
+         This is the default.
+
+To see the supported values for each option, run:
+
+.. prompt:: bash $
```
jml: :grey_question: Not necessary for this patch, but do we have a plan for keeping this (excellent!) documentation up to date?

benchmark/scenarios/no_load.py
```diff
@@ -0,0 +1,44 @@
...
+
+from zope.interface import implementer
+
+from twisted.internet.defer import Deferred, succeed
+
+from .._interfaces import IScenario
+
+
+@implementer(IScenario)
+class NoLoadScenario(object):
```
jml: This appears to be untested. I realize it's very simple, but because our static checks are so limited, it really is worth having direct coverage so we can be sure we're avoiding common mistakes. Please write some tests.

benchmark/operations/read_request.py
```diff
@@ -0,0 +1,43 @@
...
+from twisted.web.client import ResponseFailed
+
+from .._interfaces import IProbe, IOperation
+
+
+def _report_error(failure):
+    failure.trap(ResponseFailed)
+    for reason in failure.value.reasons:
+        reason.printTraceback()
+    return reason
```
jml: Is this debugging code? Is there a reason we're printing this traceback and not others?

Please either delete the code or add a comment explaining why we want to do this here.

benchmark/operations/read_request.py
```diff
@@ -0,0 +1,43 @@
+from pyrsistent import PClass, field
+from zope.interface import implementer
+
+from twisted.internet.defer import succeed
+from twisted.web.client import ResponseFailed
+
+from .._interfaces import IProbe, IOperation
+
+
+def _report_error(failure):
```
jml: If you keep this, please add a docstring.

benchmark/_driver.py
```diff
@@ -0,0 +1,139 @@
...
+            host=config['control'],
+            port=4523,
+            ca_cluster_path=cert_directory.child(b"cluster.crt"),
+            cert_path=cert_directory.child(b"user.crt"),
+            key_path=cert_directory.child(b"user.key"),
+        )
+
+        d = control_service.version()
+    else:
+        # Only valid for operation 'no-op'
```
jml: I guess this is a convenience for some sort of testing?

benchmark/_driver.py
```diff
@@ -0,0 +1,139 @@
...
+                    ),
+                    lambda reason: samples.append(
+                        dict(success=False, reason=reason.getTraceback()),
+                    ),
+                )
+                probing.addCallback(lambda ignored: probe.cleanup())
+                return probing
+            sampling.addCallback(run_probe)
+            sampling.addActionFinish()
+            return sampling.result
```
jml: `benchmark` is a fairly large function. I think it'd be clearer if `sample` were extracted to a top-level, private function that is called by `collect_samples`. Maybe (i.e. optionally) it would benefit from direct tests?

benchmark/_interfaces.py
```diff
@@ -0,0 +1,100 @@
...
+        :return: A Deferred that fires when the desired scenario is
+            established (e.g. that a certain load is being applied).
+        """
+
+    def maintained(self):
+        """
+        :return: A Deferred that fires with an errback if the desired
+            scenario fails to hold between being established and being
+            stopped.  This Deferred never fires with a callback.
+        """
```
jml: I don't quite get how the word "maintained" is relevant in this context. I can't exactly think of a better name though.

docs/gettinginvolved/benchmarking.rst
```diff
@@ -0,0 +1,64 @@
...
+Benchmarking
+============
+
+Flocker includes a tool for benchmarking operations.
+It needs the appropriate ssh-key added to a running ssh-agent.
+It is called like this:
+
+.. prompt:: bash $
+
+   python benchmark/benchmark_control.py <options>
```
jml: I guess we've gone for this approach (having the executable as a module in the package itself rather than a separate `bin` executable or an `entry_point`) because we don't want benchmarking distributed with the rest of flocker. Is that right?

benchmark/_interfaces.py
jongiddy: I did think of calling it `collapsed` and making it use a callback if the scenario collapses. In that case, errBack would indicate a problem in monitoring the status.  Since both a collapsing scenario and a problem monitoring whether the scenario has collapsed are problematic for the benchmark, the calling code would probably be expected to use `addBoth` to handle. (That's why I didn't bother in the end). But maybe it makes more sense?

jml:
Thanks! A few comments, but nothing major, I think. Please address & merge.

jml:
Feel free to ask me questions out-of-band.

benchmark/_driver.py
jongiddy: Yes. I found that it was very easy to leave off the `control` option, and not realise that you are not testing a real cluster, but it is useful to be able to run the benchmark without a cluster.  Since benchmarking `no-op` makes no real sense, you must be testing in this case.

A better solution might be to have an explicit name for a fake cluster (e.g. `--control=test`)?

benchmark/_interfaces.py
jml: Oh, I see. I think `maintained` is better than that. Maybe it's worth adding an explanatory comment in the docstring?

docs/gettinginvolved/benchmarking.rst
jongiddy: Nah, I'm just lazy. :laughing: 

Rather than an entrypoint, I think the code should follow the model used in `admin` directory, so the command would invoked as `benchmark/benchmark_control <options>`.

However, I might put that question out for discussion. Should benchmarking be available on deployed clusters?

benchmark/scenarios/no_load.py
jongiddy: For real scenarios, there will be a monitoring "thread" that will fire the Deferred, and the client that adds callbacks.  These two users of the Deferred need to get the same Deferred.  It doesn't matter here so much, but it also doesn't seem a big problem.

benchmark/test/test_driver.py
jongiddy: I tried converting to `SynchronousTestCase`, but the `cooperate` function in `benchmark` uses the global reactor: http://twistedmatrix.com/trac/browser/tags/releases/twisted-15.4.0/twisted/internet/task.py#L337

benchmark/_interfaces.py
```diff
@@ -0,0 +1,100 @@
...
+
+class IScenario(Interface):
+    """
+    A scenario in which an operation occurs.
+    """
+
+    def __init__(self, clock, control_service):
+        """
+        Provide the scenario with a clock and control service.
+        """
```
exarkun: Couple things about zope.interface
 - the convention is to leave the `self` argument out of the method signatures (since the caller isn't expected to pass this argument)
 - `__init__` is typically not part of the interface (ie, creation and initialization of the object that provides the interface is not part of the interface - though sometimes it's part of some other interface)

jml: Yes, good point.

benchmark/scenarios/no_load.py
jongiddy: Test added

benchmark/test/test_driver.py
jongiddy: Combined the fakes.

docs/gettinginvolved/benchmarking.rst
jongiddy: I've added a comment to the options area of the code, to remind people to update the docs.  More automated solutions may be possible, but could be significant work, so would deserve their own issue.

PR #2163, opened 2015-11-11
[FLOC-3412] New base TestCase 

jml:
In order to implement retries for flaky tests, we're going to have to customize the way tests are run, because we want to retry `setUp`, the test method itself, and `tearDown`.

[testtools](http://testtools.readthedocs.org/en/latest/) has a built-in mechanism for customizing the behaviour of how a test is run without retrying (using the [RunTest abstraction](http://testtools.readthedocs.org/en/latest/for-framework-folk.html#controlling-test-execution)). Because these are composable, we can combine our (future, not yet implemented) `RetryFlakyTest` runner with the runners that testtools provides for Twisted tests. 

The implementation strategy here is to create two new "official" base classes for flocker tests: `flocker.testtools.TestCase` and `flocker.testtools.AsyncTestCase`. In later branches, all existing tests will be migrated to use these base cases. For now, I've focused only on the acceptance tests, because they are the most flaky ones.

A later branch will add a runner for retrying flaky tests. 

In an attempt to actually _add_ value, I've done a little bit of extra work to provide pretty logs with test failures. Unfortunately, I haven't yet got a good example of how it looks. 

Some context:

testtools has a feature for attaching arbitrary stuff to a test and having that stuff show up when the test fails. For Twisted tests, it attaches the entire log that occurred while the test was run, e.g.
http://build.clusterhq.com/builders/flocker%2Facceptance%2Frackspace%2Fubuntu-14.04%2Frackspace/builds/3324/steps/trial/logs/problems

Because we log a lot of stuff to Twisted in our acceptance tests, I thought it would be nice to separate out the eliot logs & pretty print them. This will hopefully make failures easier to debug. When I get an example I'll link it here.

As usual, I've added `XXX` comments where I'd particularly like feedback about a trade-off I made or a decision that I'm not entirely comfortable about.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2163)


flocker/testtools/_base.py
```diff
@@ -0,0 +1,147 @@
...
+    # XXX: Looks like the acceptance tests (which were the first tests that we
+    # tried to migrate) aren't cleaning up after themselves even in the
+    # successful case. Use the RunTest that loops the reactor a couple of
+    # times after the test is done.
+    return AsynchronousDeferredRunTestForBrokenTwisted.make_factory(
+        timeout=timeout)
+
+
+# By default, asynchronous tests are timed out after 2 minutes.
+DEFAULT_ASYNC_TIMEOUT = 120
```
jml: The default timeout in testtools is 0.005 seconds. I'd like to do a later pass where we lower this significantly and add custom timeouts for all the slow tests.

jml:
> When I get an example I'll link it here.

I was assuming that at least one of the acceptance tests would fail due to flakiness. After three builds, none of them did. 

Then I wrote a test that demonstrated the post-processing code, and it doesn't work for obscure, easily-fixable reasons. So, I've killed all the post-processing logic, leaving the branch nice & simple.

flocker/testtools/_base.py
exarkun: Does testtools support a kind of timeout where progress towards completion in the test can be detected/reported to bump the eventual timeout forward, making it so a timeout can be imposed on individual steps rather than the test method as a whole?

I see the timeout feature as a way to catch hung tests which will never finish.  As long as the test is making progress towards completion, I don't see much point in aborting it early and calling it a failure.  Calling out tests that take too long to run overall seems like the job for some kind of test monitoring system, not the test runner.

exarkun:
Just out of curiosity ... why all the extra unrelated commits in the history of this branch, as rendered on this PR?

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+                raise unittest.SkipTest(reason)
+
+        test = SkippingTest('test_skip')
+        result = Python27TestResult()
+        test.run(result)
+        self.assertEqual([
+            ('startTest', test),
+            ('addSkip', test, reason),
+            ('stopTest', test),
+        ], result._events)
```
exarkun: Private API from a third-party library ... File an upstream feature request for adding an equivalent public API and add a link to that issue from here?

jml:
No good reason. Want me to fix it?

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+        If tests raise the ``SkipTest`` exported by Trial, then that's
+        recorded as a skip.
+        """
+
+        class SkippingTest(AsyncTestCase):
+            def test_skip(self):
+                raise unittest.SkipTest(reason)
+
+        test = SkippingTest('test_skip')
+        result = Python27TestResult()
```
exarkun: First I wondered by Python 2.7 specifically.  Then I wondered why a test double at all.  Looks like the answer to the latter is because this adds "what happened" logging behavior that the stdlib test result doesn't have and the answer to the former is that testtools doesn't offer a "current version of Python" test result double.

I guess Python 2.7 is the explictly supported version of Python for Flocker.  And I guess the chance that we'll switch to another version in the foreseeable future is basically nil.  So this is probably a non-problem.

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+    def mktemp(self):
+        """
+        Create a temporary directory that will be deleted on test completion.
+
+        Provided for compatibility with Twisted's ``TestCase``.
+
+        :return: Path to the newly-created temporary directory.
+        """
+        # XXX: Should we provide a cleaner interface for people to use? One
+        # that returns FilePath? One that returns a directory?
```
exarkun: I always want a `FilePath` when I call `mktemp`.  That sounds like a follow-up change, since if we don't want to provide this one,we have to fix a bunch of tests.

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+
+        # XXX: Actually belongs in a mixin or something, not actually specific
+        # to async.
+        temp_dir = FilePath(self.useFixture(fixtures.TempDir()).path)
```
exarkun: I tried to figure out where these temporary directories are created by reading the fixtures documentation but I couldn't.  I'd like them to stay inside `_trial_temp`.

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+
+    run_tests_with = async_runner(timeout=DEFAULT_ASYNC_TIMEOUT)
+
+    def __init__(self, *args, **kwargs):
+        super(AsyncTestCase, self).__init__(*args, **kwargs)
+        self.exception_handlers.insert(-1, (unittest.SkipTest, _test_skipped))
+
+    def mktemp(self):
+        """
+        Create a temporary directory that will be deleted on test completion.
```
exarkun: Deleting them makes debugging harder.

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+)
+from testtools.testresult.doubles import Python27TestResult
+from twisted.python.filepath import FilePath
+from twisted.trial import unittest
+
+from .._base import AsyncTestCase
+
+
+class AsyncTestCaseTests(TestCase):
+
```
exarkun: Class docstring missing

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+        does.
+        """
+
+        class SomeTest(AsyncTestCase):
+            def test_pass(self):
+                pass
+
+        test = SomeTest('test_pass')
+        temp_path = FilePath(test.mktemp())
+        self.addCleanup(_remove_dir, temp_path.parent())
```
exarkun: The docs for the new `mktemp` method say the deletion happens automatically.  This cleanup is redundant, then?  Belt and suspenders or ...?

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+
+        self.expectThat(temp_path.parent().path, DirExists())
```
exarkun: Yea.

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+        self.assertThat(temp_path.path, Not(PathExists()))
```
exarkun: *Oh* yea.

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,71 @@
...
+from testtools.testresult.doubles import Python27TestResult
+from twisted.python.filepath import FilePath
+from twisted.trial import unittest
+
+from .._base import AsyncTestCase
+
+
+class AsyncTestCaseTests(TestCase):
+
+    @given(binary())
```
exarkun: If the test ever fails and I have to read the (one list) != (another list) output including the generated value, does this mean I might end up looking at 80kB of binary junk?  Would configuring a smallish average here be a reasonable use of Hypothesis?  Something around 30, say.

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+class AsyncTestCase(testtools.TestCase):
+    """
+    Base class for asynchronous test cases.
+    """
+
+    run_tests_with = async_runner(timeout=DEFAULT_ASYNC_TIMEOUT)
+
+    def __init__(self, *args, **kwargs):
+        super(AsyncTestCase, self).__init__(*args, **kwargs)
+        self.exception_handlers.insert(-1, (unittest.SkipTest, _test_skipped))
```
exarkun: Smells like `append`?

exarkun: Pft except of course it isn't.  What comes after the skip handler, then, I wonder...

exarkun: (Indexes into a list is probably not a great API for supporting third-party extensions... Maybe out of scope, thugh.)

flocker/testtools/_base.py
exarkun: :question: Also timedeltas are cool

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+    Make a ``RunTest`` instance for asynchronous tests.
+
+    :param float timeout: The maximum length of time (in seconds) that a test
+        is allowed to take.
+    """
+    # XXX: Looks like the acceptance tests (which were the first tests that we
+    # tried to migrate) aren't cleaning up after themselves even in the
+    # successful case. Use the RunTest that loops the reactor a couple of
+    # times after the test is done.
+    return AsynchronousDeferredRunTestForBrokenTwisted.make_factory(
```
exarkun: :cry:

flocker/testtools/_base.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+    Base class for synchronous test cases.
+    """
+
+
+def async_runner(timeout):
+    """
+    Make a ``RunTest`` instance for asynchronous tests.
+
+    :param float timeout: The maximum length of time (in seconds) that a test
```
exarkun: Turns out it's an int right now :wink:

exarkun:
> No good reason. Want me to fix it?

The diff is reasonable so it's not a problem for the review.  Just wanted to understand so I could recognize it on my own in the future or avoid bumping into whatever the circumstance is myself.

exarkun:
Tricky review!  Simple code change but since what's going on here is "stop using the large bag of partially-explicit, partially-implicit, partially-understood, partially-mysterious functionality represented by Twisted's TestCase and start using this other different (probably smaller? though I'm not sure) bag of sort of the same type" there's lots of room for surprising results, I think.

Of course in an ideal world the test suite would tell us everything we need to know about any unaccounted-for differences between those two bags - or would it?  If one difference were to quietly make all of our tests pass even when they should fail (say, by running the wrong method when trying to run the test method...) that would be hard to pick out of a green build results.  That's pretty far down the line towards a worst case scenario, though, and somehow I doubt it's actually of practical interest.

I did get to see a failed test run for this branch, in fact.  The entire acceptance test suite failed on one of the Rackspace test runs - with timeout errors from each test.  Looks like the cluster got into that state where it can't destroy a dataset.  Not the most interesting failure, but it demonstrates at least the first test ran and when it ran out of time the test was marked as failed.  So that's good.

I like the idea of switching from Twisted's barely-good-enough TestCase to something offered by a project which exists solely for the purpose of building testing tools.  I like the *specific* idea of using testtools' Twisted support a little bit less after starting to read the testtools documentation for it.  Quoting:

    testtools provides *highly experimental* support for running Twisted tests

Also, it's a bit discouraging that I didn't find any documentation for `AsynchronousDeferredRunTestForBrokenTwisted`.  And when I read the source code, I accidentally read the implementation of `make_factory` :cry: .

I suppose the worst that can happen is that this stuff turns out to either not solve our problems or introduce worse problems and we have to switch back - which should pose no greater challenge than switching did in the first place.  I'm comfortable with taking on that risk in exchange for the expected reward.  And presumably you are as well since you reached for this particular tool. :smile: 

Please address inline comments, make peace with buildbot, and then merge.

flocker/testtools/_base.py
jml: testtools doesn't have such a feature, but that could be added. Not 100% sure if it's a feature for a test framework or a feature for a thing that runs a Twisted reactor until a Deferred fires.

Changed to a timedelta.

flocker/testtools/test/test_base.py
jml: Already fixed: https://github.com/testing-cabal/testtools/commit/c51fdb854ea1f7b0e473756fc42bfcf927c9652d#diff-3a626e856299020b1600f83f27d0a299L57

Just need a release.

flocker/testtools/test/test_base.py
jml: Right. stdlib provides poor infrastructure for testing test-related stuff. testtools is better there, but could also use some improvement there.

testtools doesn't provide a "current version of Python" test result double because the API varies hilariously by Python version, and because we try to stay out of the DWIM business by design.

I've added a comment here.

flocker/testtools/_base.py
jml: Yes, follow-up change. My meta-goal here (which I can clarify by adding a comment if you think it would be wise) is to provide a compatibility layer to ease transition.

flocker/testtools/test/test_base.py
jml: Added. 

(We should add a linter for this check)

flocker/testtools/test/test_base.py
jml: Happens automatically if the test is _run_. There isn't any logic to clean it up in `__del__` or anything like that. In any case, is a moot point because I'll have to fix this up to have more trial-like behaviour. Will circle back to see if there's anything else to be resolved here.

flocker/testtools/test/test_base.py
jml: :dancer:

flocker/testtools/test/test_base.py
jml: > If the test ever fails and I have to read the (one list) != (another list) output including the generated value, does this mean I might end up looking at 80kB of binary junk?

No, because hypothesis has awesome shrinking logic. If it finds a failing case, it'll do its level best to find a smaller failing case. 

> Would configuring a smallish average here be a reasonable use of Hypothesis? Something around 30, say.

Yes. We could also restrict the alphabet, but I think it's best to wait until it's a problem.

flocker/testtools/_base.py
jml: Yeah, sorry. At least this wart is documented.

flocker/testtools/_base.py
jml: It's now a timedelta. Docs fixed.

flocker/testtools/test/test_base.py
exarkun: > No, because hypothesis has awesome shrinking logic.

Oh yea, I forgot about that feature.  Awesome.

jml:
> Tricky review! Simple code change but since what's going on here is "stop using the large bag of partially-explicit, partially-implicit, partially-understood, partially-mysterious functionality represented by Twisted's TestCase and start using this other different (probably smaller? though I'm not sure) bag of sort of the same type" there's lots of room for surprising results, I think.

Thanks! I really appreciated you taking the time. I agree that it's tricky, and mostly agree with the "bag" analysis (except I'm not sure what "explicit functionality" means). 

When I wrote the Twisted support in testtools, I tried hard to at least draw clean lines around the mystery. Thus, `_spinner` does all of the reactor stuff, is [pretty well tested](https://github.com/testing-cabal/testtools/blob/master/testtools/tests/test_spinner.py) and doesn't actually know anything about testing. If I had had a time machine I might have traveled into the future and used crochet.

It does some horrible stuff to mitigate the globalness and the unrestartability of the reactor, but the idea is that it's all in a well-tested black box.

> Of course in an ideal world the test suite would tell us everything we need to know about any unaccounted-for differences between those two bags - or would it? If one difference were to quietly make all of our tests pass even when they should fail (say, by running the wrong method when trying to run the test method...) that would be hard to pick out of a green build results. That's pretty far down the line towards a worst case scenario, though, and somehow I doubt it's actually of practical interest.

Actually, it did genuinely concern me when I had a run of "bad" luck trying to get an example of my nice log formatting. 

If it becomes a more live concern, we can add an integration test. 

> I did get to see a failed test run for this branch, in fact. The entire acceptance test suite failed on one of the Rackspace test runs - with timeout errors from each test. Looks like the cluster got into that state where it can't destroy a dataset. Not the most interesting failure, but it demonstrates at least the first test ran and when it ran out of time the test was marked as failed. So that's good.

Right. You might have also noticed that it didn't fail the rest of the tests due to leaked garbage.

> I like the idea of switching from Twisted's barely-good-enough TestCase to something offered by a project which exists solely for the purpose of building testing tools. I like the specific idea of using testtools' Twisted support a little bit less after starting to read the testtools documentation for it. Quoting:
>
> > testtools provides *highly experimental* support for running Twisted tests
>
> Also, it's a bit discouraging that I didn't find any documentation for AsynchronousDeferredRunTestForBrokenTwisted. And when I read the source code, I accidentally read the implementation of make_factory :cry: .

When I wrote that, all the other parts of testtools were well-proven, known-good solutions to problems that had been pioneered elsewhere, so this part was comparatively experimental. Since then, testtools started pioneering more of its own stuff (details, matchers, `exception_handlers`)

Anyway, the lack of docs is a bit of a consequence of that. After we start using it, I'll do a sweep to remove the scares: https://bugs.launchpad.net/testtools/+bug/1515684

> I suppose the worst that can happen is that this stuff turns out to either not solve our problems or introduce worse problems and we have to switch back - which should pose no greater challenge than switching did in the first place. I'm comfortable with taking on that risk in exchange for the expected reward. And presumably you are as well since you reached for this particular tool. :smile:

Hopefully less of a challenge. Channeling all of our tests through base cases that we control in our code base means that if we want to move from testtools to something else, we only have to change the one place.

flocker/testtools/test/test_base.py
jml: Circling back: Now that we don't auto-delete at all, this is even more important.

flocker/testtools/_base.py
jml: Fixed. Now also creates under current working directory, which is normally `_trial_temp`.

jml:
Some non-trivial updates for making behaviour of `mktemp` match trial a bit more. Feedback welcome, but per your original review I'll land once I'm happy w/ the test results.

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,193 @@
...
+                created_files.append(path)
+                open(path, 'w').write('hello')
+
+        SomeTest('test_create_file').run(Python27TestResult())
+        [path] = created_files
+        self.addCleanup(os.unlink, path)
+        self.assertThat(path, FileContains('hello'))
+
+
+identifier_characters = string.ascii_letters + string.digits + '_'
```
exarkun: `.decode("ascii")`?  (Seeing as next line uses this for `text(...)`)

flocker/testtools/test/test_base.py
```diff
@@ -0,0 +1,193 @@
...
+identifiers = text(average_size=20, min_size=1, alphabet=identifier_characters)
+fqpns = lists(
+    identifiers, min_size=1, average_size=5).map(lambda xs: '.'.join(xs))
```
exarkun: `u'.'`?

flocker/testtools/test/test_base.py
exarkun: Oh but they're ... Python identifier strings ...

nevermind, way too much of a mess.  Please disregard this.

exarkun:
Great changes, thanks.  Please ignore all of my new comments on this PR.

PR #2165, opened 2015-11-12
FLOC-3441: Readd skip_tests method

stmcginnis:
Using skip_tests was very useful for backend driver developers. Typically the
make_iblockdeviceapi_tests function is called to get all relevant functional
tests to run against the backend driver. As the driver is being developed, or
when trying to isolate a specific failure condition, it can be distracting to
have all tests run.

skip_tests allows the driver developer to restrict these functional tests to
the subset or single test that is needed to help troubleshoot. This makes it a
lot easier to grep through logs and inspect specific actions of the driver.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2165)


robhaswell:
Hi! Thanks for this. Sorry it took a while to get back to you, we have been running around trying to prepare for DockerCon last week! Somebody will review this shortly and get back to you.

exarkun:
Hi stmcginnis,

Right now we're not in a good position to offer much functionality from Flocker in its role as a Python library.  The fact that this function you were using was removed with no fanfare or warning is a symptom of this.  A few parts of Flocker are unquestionably part of a public interface to Python programmers - the `IBlockDeviceAPI` interface, for example.  I'd put the test-generating helpers like `make_iblockdeviceapi_tests` close to this but since these are part of Flocker's own test suite they're still not really set in stone - we just don't treat the test suite implementation as something that requires the kind of care and thought necessary to maintain a public interface.  Test support code that's less specifically about Flocker is probably going to be even more subject to incompatible changes without notice.  Providing better tools here would probably be a really great thing and I hope we'll look into it but I can't really say when that might happen.

I think my recommendation for things like `skip_tests` will be that if it's useful, please copy it into your code base - it's already Apache licensed which is pretty permissive, but if that's not enough then I can find a way to give it away even more permissively.

stmcginnis:
I think issues like this need to be worked out sooner than later. As more vendors add third party Flocker drivers it needs to be clear what is considered "public" and what is an internal piece of Flocker.

Documentation for implementing a Flocker driver instruct to use make_iblockdeviceapi_tests, so that one is clearly public at this point, so if the stance on that is going to change then it will somehow need to be clearly communicated.

PR #2171, opened 2015-11-13
[FLOC-3387] Managed acceptance for existent nodes 

mpgomez:
Hacky fix that may or may not work as well as I expect (I keep getting unclean reactor errors that I don't know if they are caused by my code or not).
Open to suggestions. This code may very well had been in design review instead of code review...

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2171)


admin/acceptance.py
```diff
@@ -1046,8 +1050,11 @@ def main(reactor, args, base_path, top_level):
     cluster = None
     try:
         yield runner.ensure_keys(reactor)
+        # try:
+        #     cluster = yield runner.start_cluster(reactor)
+        # except SequenceFailed as e:
+        #     pass
```
jml: Please delete this commented out code.

flocker/provision/_install.py
```diff
@@ -27,7 +27,7 @@
     put,
     run_remotely,
 )
-from ._effect import sequence
+from ._effect import sequence, Sequence
```
jml: "Sequence" is not being used. Please delete the import.

flocker/provision/_install.py
```diff
@@ -817,11 +817,17 @@ def task_enable_flocker_control(distribution):
             ),
             run("echo 'flocker-control-api\t4523/tcp\t\t\t# Flocker Control API port' >> /etc/services"),  # noqa
             run("echo 'flocker-control-agent\t4524/tcp\t\t\t# Flocker Control Agent port' >> /etc/services"),  # noqa
-            run_from_args(['service', 'flocker-control', 'start']),
+            #run_from_args(['service', 'flocker-control', 'start']).on(error=managed_errors_handler),
```
jml: Please delete commented out code.

flocker/provision/_install.py
```diff
@@ -817,11 +817,17 @@ def task_enable_flocker_control(distribution):
...
+            run_from_args(['service', 'flocker-control', action]),
         ])
+
     else:
         raise DistributionNotSupported(distribution=distribution)
 
+# def managed_errors_handler(argument):
+#     print "Error starting flocker control. Ignoring it."
+#     sleep(10)
+
```
jml: Please delete this too :smile:

flocker/provision/_install.py
```diff
@@ -1000,8 +1006,10 @@ def task_enable_flocker_agent(distribution):
         ])
     elif distribution == 'ubuntu-14.04':
         return sequence([
-            run_from_args(['service', 'flocker-dataset-agent', 'start']),
-            run_from_args(['service', 'flocker-container-agent', 'start']),
+            # run_from_args(['service', 'flocker-dataset-agent', 'start']).on(error=managed_errors_handler),
+            # run_from_args(['service', 'flocker-container-agent', 'start']).on(error=managed_errors_handler),
```
jml: Please delete this too :smile:

flocker/provision/_install.py
```diff
@@ -1393,6 +1401,10 @@ def configure_cluster(cluster, dataset_backend_configuration):
     :param dict dataset_backend_configuration: Configuration parameters to
         supply to the dataset backend.
     """
+    setup_action="start"
+    if(provider == "managed"):
+        setup_action="restart"
```
jml: Please put spaces around the assignment operator, e.g.`setup_action = "start"`).

flocker/provision/_install.py
```diff
@@ -1393,6 +1401,10 @@ def configure_cluster(cluster, dataset_backend_configuration):
     :param dict dataset_backend_configuration: Configuration parameters to
         supply to the dataset backend.
     """
+    setup_action="start"
+    if(provider == "managed"):
```
jml: No need for parentheses around conditions in Python. Just `if provider == "managed"`.

admin/acceptance.py
```diff
@@ -51,6 +51,7 @@
 )
 
 from flocker.common.runner import run, run_ssh
+from flocker.provision._effect import SequenceFailed
```
jml: I don't _think_ `SequenceFailed` is being used in this file?

jml:
I've just done a quick pass for style. Will do another, deeper pass now.

flocker/provision/_install.py
```diff
@@ -793,7 +793,7 @@ def open_ufw(service):
         ])
 
 
-def task_enable_flocker_control(distribution):
+def task_enable_flocker_control(distribution, action="start"):
     """
     Enable flocker-control service.
```
jml: It's a bit unfair asking you to do this, since previous committers haven't but... could you please document the parameters to this function?

flocker/provision/_install.py
```diff
@@ -817,11 +817,17 @@ def task_enable_flocker_control(distribution):
             ),
             run("echo 'flocker-control-api\t4523/tcp\t\t\t# Flocker Control API port' >> /etc/services"),  # noqa
             run("echo 'flocker-control-agent\t4524/tcp\t\t\t# Flocker Control Agent port' >> /etc/services"),  # noqa
-            run_from_args(['service', 'flocker-control', 'start']),
+            #run_from_args(['service', 'flocker-control', 'start']).on(error=managed_errors_handler),
+            run_from_args(['service', 'flocker-control', action]),
```
jml: This is interesting. `action` is only used by the Ubuntu branch, because the Centos branch *always* does a restart (See `START = "restart"` above).

I wonder if it makes sense for Ubuntu to always do a restart too?

jml:
So the basic idea is that if it's a managed run, we restart the nodes (when we're on Ubuntu), rather than just starting them?

That seems like a reasonable step forward.

As noted in my inline comment, I wonder whether we should _always_ restart. You probably know better than I do whether that's a good idea. Happy to talk about it in some other forum if you'd like.

Please make the suggested (mostly stylistic) changes. If you think we should always restart, please make that change too. 

The build errors you're seeing are almost certainly not due to the changes in this branch, except for the "lint" failures, which should be fixed after you address all the review comments. The "acceptance" failures  are known flaky.

jml:
Forgot a couple of things:
1. Thanks for doing this! It's important that our tools are reliable.
2. Please resubmit for review when you're done, assigning to me

PR #2174, opened 2015-11-13
[FLOC-3450] Bump testtools to 1.8.1

jml:
We were using a private API for testtools. Now we're not.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2174)


sarum90:
Took a look at the 1.8.0 vs 1.8.1 diff for kicks, seems pretty tame. Looks good to me, although probably wait til tests are green to merge. Both of the rackspace tests failing makes me a little nervous.

sarum90:
Also +1 on removing dependencies on private APIs :+1:

PR #2175, opened 2015-11-13
[FLOC-3386] Calculate changes explicitly

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3386

There are a couple of tests missing about verifying the fake ICalculater, but there is enough here to review.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2175)


tomprince:
I *think* that there is pretty much equivalent test coverage, but there is enough lines that it is hard to be sure.

A couple that I dropped:
- that metadata changes don't trigger convergence
- that default size doesn't trigger repeated convergence

but the code isn't shaped in such a way that those could ever happen anyway.

flocker/node/agents/blockdevice.py
```diff
@@ -87,6 +89,7 @@ class DiscoveredDataset(PClass):
         mandatory=True,
     )
     dataset_id = field(type=UUID, mandatory=True)
+    # XXX int or *None*
```
itamarst: If it's discovered then it should always have a size?

flocker/node/agents/blockdevice.py
```diff
@@ -62,6 +62,8 @@ class DatasetStates(Names):
     """
     States that a ``Dataset`` can be in.
     """
+    NON_EXISTANT = NamedConstant()
```
itamarst: NON_EXISTENT I think.

flocker/node/agents/blockdevice.py
```diff
@@ -117,11 +120,73 @@ def __invariant__(self):
                         )
                     )
                 return (False, message)
-        if self.state in (DatasetStates.DELETED,):
+        if self.state in (DatasetStates.DELETED, DatasetStates.NON_EXISTANT):
             return (False, "DesiredDataset can't be in state DELETED.")
```
itamarst: `NON_EXISTENT` too.

flocker/node/agents/blockdevice.py
```diff
@@ -117,11 +120,73 @@ def __invariant__(self):
...
+        value_type=unicode,
+    )
+    mount_point = field(FilePath)
+
+    def __invariant__(self):
+        expected_attributes = [
+            ((DatasetStates.NON_MANIFEST, DatasetStates.MOUNTED),
+             "maximum_size"),
+        ]
+        expected_attributes = [
```
itamarst: This overwrites previous one? Does `DiscoveredDataset` have same bug? Perhaps a dictionary would be better.

flocker/node/agents/blockdevice.py
```diff
@@ -117,11 +120,73 @@ def __invariant__(self):
...
+            ((DatasetStates.MOUNTED,),
+             "mount_point"),
+        ]
+        for states, attribute in expected_attributes:
+            if (self.state in states) != hasattr(self, attribute):
```
itamarst: `if self.state not in states: continue` would simplify this whole chunk a bit. Ditto for DiscoveredDataset if has same issues.

flocker/node/agents/blockdevice.py
```diff
@@ -117,11 +120,73 @@ def __invariant__(self):
...
+                            attr=attribute,
+                            states=','.join(map("`{0.name}`".format, states)),
+                        )
+                    )
+                return (False, message)
+        if self.state in (
+            DatasetStates.ATTACHED, DatasetStates.NON_EXISTANT,
+            DatasetStates.ATTACHED_ELSEWHERE,
+        ):
+            return (False, "DesiredDataset can't be in state ATTACHED.")
```
itamarst: Inaccurate string.

flocker/node/agents/blockdevice.py
```diff
@@ -431,6 +496,13 @@ class DestroyBlockDeviceDataset(PRecord):
     dataset_id = field(type=UUID, mandatory=True)
     blockdevice_id = field(type=unicode, mandatory=True)
 
+    @classmethod
+    def from_dataset(cls, discovered_dataset, desired_dataset):
```
itamarst: Maybe these should be called... `from_discovered_and_desired`? `from_state_and_config`?

flocker/node/agents/blockdevice.py
```diff
@@ -763,6 +870,16 @@ class CreateBlockDeviceDataset(PRecord):
     dataset = field(mandatory=True, type=Dataset)
     mountpoint = field(mandatory=True, type=FilePath)
 
+    @classmethod
+    def from_dataset(cls, discovered_dataset, desired_dataset):
+        return cls(
+            dataset=Dataset(
+                dataset_id=desired_dataset.dataset_id,
+                maximum_size=desired_dataset.maximum_size,
```
itamarst: `primary=True` probably? or maybe it doesn't care.

flocker/node/agents/blockdevice.py
```diff
@@ -1264,6 +1381,72 @@ def shared_state_changes(self):
         )
 
 
+# Mapping from desired and discovered dataset state to
+# IStateChange factory. (The factory is expected to take
+# ``desired_dataset`` and ``discovered_dataset``.
+DATASET_TRANSITIONS = {
```
itamarst: :heart_eyes:

itamarst: This might be more readable if you did:
1. `Desired = DatasetStates; Current = DatasetStates`
2. define a `DO_NOTHING`
3. add comments where necessary.

E.g.

```
Desired.MOUNTED: {
    Current.ATTACHED_ELSEWHERE: DO_NOTHING,  # other node will need to detach first
    Current.ATTACHED: MountBlockDevice.from_dataset,
}
```

flocker/node/agents/blockdevice.py
```diff
@@ -1264,6 +1381,72 @@ def shared_state_changes(self):
...
+            )
+        else:
+            return NoOp()
+
+    def calculate_changes_for_datasets(
+        self, discovered_datasets, desired_datasets
+    ):
+        actions = []
+        # If a dataset isn't in the configuration, we don't act on it.
+        for dataset_id in set(discovered_datasets) | set(desired_datasets):
```
itamarst: Maybe clearer with with `set(discovered_datasets.keys())`?

flocker/node/agents/blockdevice.py
```diff
@@ -1442,234 +1626,93 @@ def _mountpath_for_dataset_id(self, dataset_id):
...
+        desired_datasets = {}
+        for manifestation in this_node_config.manifestations.values():
+            dataset_id = UUID(manifestation.dataset.dataset_id)
+            # XXX: Make this configurable. FLOC-2679
+            maximum_size = manifestation.dataset.maximum_size
+            if maximum_size is None:
+                maximum_size = DEFAULT_DATASET_SIZE
+
+            if (
+                manifestation.dataset.deleted is True
```
itamarst: maybe just omit `is True`?

flocker/node/agents/blockdevice.py
```diff
@@ -1442,234 +1626,93 @@ def _mountpath_for_dataset_id(self, dataset_id):
...
-                 attached to this node.
-        """
-        for manifestation in configured.values():
-            dataset_id = UUID(manifestation.dataset_id)
-            if dataset_id in devices:
-                # It's already attached here.
+            if dataset.state not in (
+                DatasetStates.ATTACHED, DatasetStates.MOUNTED
+            ):
+                # A lease doesn't force a mount.
```
itamarst: 1. Where's the bit where you calculate desired NOT_MANIFEST? Seems like that never happens.
2. What is the purpose of this whole block of logic? I don't see any fitering by not_in_use above so it seems redundant.

itamarst:
* Add docstring for `assert_desired_datasets`, and maybe use defaults that aren't mutable.
* The expected `DesiredDataset` in CalculateDesiredStateTests are pretty repetitive, can we just have one and reuse it?

itamarst:
```
                ScenarioMixin.MANIFESTATION.transform(
                    ["dataset"], lambda d: d.set(deleted=True)
                ),
```

can be

```
                ScenarioMixin.MANIFESTATION.transform(
                    ["dataset", "deleted"], True,
                ),
```

itamarst:
There's a reference to FLOC-1172 that ought to be FLOC-1772.

itamarst:
Docstring for `test_deleted_attached_manifest_dataset` erroneously talks about wanting `DELETED`, we really want `MOUNTED` until 1772 is fixed.

itamarst:
`test_deleted_non_manifest_dataset` should probably mention 1772 too since I believe it will fix that too.

flocker/node/agents/blockdevice.py
itamarst: Oh, I see, it's "if this is in use we want it to be `MOUNTED`". That isn't really how leases work right now though; currently they just *prevent* anything from happening... so that seems not quite right. Maybe just prevent NOT_MANIFEST from happening and otherwise don't change what's already there?

itamarst: Also still need to calculate NOT_MANIFEST.

itamarst:
I'm out of steam, but pretty much reached the calculate_changes tests but that's a good stopping point since as we discussed they're less important.

OK, so:
1. Address the above.
2. Address remaining XXX/things lacking docstrings.
3. Probably merge forward to get 1771 & 1576? I assume they'll be merged real soon now.

I'm really excited about getting this in, thanks for figuring this out.

flocker/node/agents/blockdevice.py
tomprince: I don't see how. We want to error when the attribute is present in other states, and not present in the given states.

flocker/node/agents/blockdevice.py
tomprince: I don't think it cares, but I've changed it anyway.

flocker/node/agents/blockdevice.py
tomprince: I agree, but I thought that having it on a single line was even clearer. :disappointed:

flocker/node/agents/blockdevice.py
tomprince: >  That isn't really how leases work right now though; currently they just prevent anything from happening... so that seems not quite right. 

Well, the only real change is that this causes a lease to mount an `ATTACHED` dataset. If the dataset is in any other state, we won't set the desired state to `MOUNTED`.

tomprince: Regarding `NON_MANIFEST`, I was going to comment that while it wasn't necessary (since missing `DesiredDataset`s get treated as `NON_MANIFEST`) it makes sense to have them explicit. But, we don't have any configuration for non-manifest nodes. I could loop through every other node and mark all the manifestations there, but I'll just leave a comment as to why we don't discover `NON_MANIFEST` right now.

flocker/node/agents/blockdevice.py
tomprince: Actually, `primary` isn't an attribute of `Datasset`, just `Manifestation`.

tomprince:
I think I've addressed all the review comments, except for the missing tests of the fake. I also dropped some tests merging forward (about changes from `ATTACHED_NO_FILESYSTEM`) but plan to address those tests differently.

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+DO_NOTHING = lambda **kwargs: NoOp()
+DATASET_TRANSITIONS = {
+    Desired.MOUNTED: {
+        Discovered.NON_EXISTENT:
+            CreateBlockDeviceDataset.from_state_and_config,
+        # Other node will need to deatch first
+        Discovered.ATTACHED_ELSEWHERE: DO_NOTHING,
+        Discovered.ATTACHED_NO_FILESYSTEM:
+            CreateFilesystem.from_state_and_config,
+        Discovered.NON_MANIFEST: AttachVolume.from_state_and_config,
```
itamarst: In `master` this code path has a `ActionNeeded` in some cases. It may be that is not necessary in this version of calculation, but you'll need to check if that is actually the case.

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+        DatasetStates.ATTACHED: MountBlockDevice.from_state_and_config,
+    },
+    Desired.NON_MANIFEST: {
+        # Can't create non-manifest datasets yet.
```
itamarst: 1771 was merged, so in fact we can now.

itamarst: Or... maybe not? File a followup issue.

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+    Desired.DELETED: {
+        Discovered.NON_EXISTENT: DO_NOTHING,
+        # Other node will destroy
+        Discovered.ATTACHED_ELSEWHERE: DO_NOTHING,
+        # Can't pick node that will do destruction yet.
+        Discovered.NON_MANIFEST: DO_NOTHING,
+        # This state won't occur until FLOC-1772
+        Discovered.ATTACHED_NO_FILESYSTEM: DetachVolume.from_state_and_config,
+        # This state won't occur until FLOC-1772
+        Discovered.ATTACHED: DetachVolume.from_state_and_config,
```
itamarst: I think this should be MountVolume -^

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+    },
+    Desired.DELETED: {
+        Discovered.NON_EXISTENT: DO_NOTHING,
+        # Other node will destroy
+        Discovered.ATTACHED_ELSEWHERE: DO_NOTHING,
+        # Can't pick node that will do destruction yet.
+        Discovered.NON_MANIFEST: DO_NOTHING,
+        # This state won't occur until FLOC-1772
+        Discovered.ATTACHED_NO_FILESYSTEM: DetachVolume.from_state_and_config,
+        # This state won't occur until FLOC-1772
```
itamarst: This should be CreateFilesystem (until 1772) -^

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+        Discovered.ATTACHED: DetachVolume.from_state_and_config,
+        Discovered.MOUNTED: UnmountBlockDevice.from_state_and_config,
+    },
+    Desired.DELETED: {
+        Discovered.NON_EXISTENT: DO_NOTHING,
+        # Other node will destroy
+        Discovered.ATTACHED_ELSEWHERE: DO_NOTHING,
+        # Can't pick node that will do destruction yet.
+        Discovered.NON_MANIFEST: DO_NOTHING,
+        # This state won't occur until FLOC-1772
```
itamarst: This should be AttachVolume (until 1772).

flocker/node/agents/blockdevice.py
```diff
@@ -1308,6 +1460,98 @@ def shared_state_changes(self):
...
+        Discovered.ATTACHED_NO_FILESYSTEM: DetachVolume.from_state_and_config,
+        # This state won't occur until FLOC-1772
+        Discovered.ATTACHED: DetachVolume.from_state_and_config,
+        Discovered.MOUNTED: DestroyBlockDeviceDataset.from_state_and_config,
+    },
+}
```
itamarst: SInce comments above are slightly confusing: DestroyBlockDeviceDataset does unmount-detach-delete, so we need to make sure volume is mounted before deletion will work.

flocker/node/agents/blockdevice.py
```diff
@@ -1333,6 +1579,12 @@ class BlockDeviceDeployer(PRecord):
     _async_block_device_api = field(mandatory=True, initial=None)
     mountroot = field(type=FilePath, initial=FilePath(b"/flocker"))
     poll_interval = timedelta(seconds=60.0)
+    calculater = field(
+        # XXX We should abstact this invariant out.
```
itamarst: I think maybe Marcus already did?

flocker/node/agents/blockdevice.py
```diff
@@ -1470,6 +1722,13 @@ def discover_state(self, node_state):
                         maximum_size=volume.size,
                         blockdevice_id=volume.blockdevice_id,
                     )
+                else:
```
itamarst: This will record datasets with `attached_to == None` as ATTACHED_ELSEWHERE, which is wrong.

flocker/node/agents/blockdevice.py
itamarst: unaddressed comment -^

flocker/node/agents/blockdevice.py
tomprince: I don't *think* that case should ever occur here. That case is we have stale non-manifest data. But we now never actually consult external state (except for applications).

flocker/node/agents/blockdevice.py
```diff
@@ -801,6 +943,16 @@ class CreateBlockDeviceDataset(PRecord):
     dataset = field(mandatory=True, type=Dataset)
     mountpoint = field(mandatory=True, type=FilePath)
 
+    @classmethod
+    def from_state_and_config(cls, discovered_dataset, desired_dataset):
+        return cls(
+            dataset=Dataset(
```
itamarst: The implementation of `run` relies on the metadata to figure out if there's a profile, so you need that as well.

flocker/node/agents/blockdevice.py
```diff
@@ -1503,274 +1751,98 @@ def _mountpath_for_dataset_id(self, dataset_id):
...
+            dataset_id = UUID(manifestation.dataset.dataset_id)
+            # XXX: Make this configurable. FLOC-2679
+            maximum_size = manifestation.dataset.maximum_size
+            if maximum_size is None:
+                maximum_size = DEFAULT_DATASET_SIZE
+
+            if (
+                manifestation.dataset.deleted is True
+                # XXX FLOC-1772
+                and dataset_id in local_datasets
```
itamarst: I don't think you need to check local state or location; I think just checking for `deleted == True` is sufficient. See suggestions I made for changing the state transition table above.

flocker/node/agents/blockdevice.py
```diff
@@ -1503,274 +1751,98 @@ def _mountpath_for_dataset_id(self, dataset_id):
...
-
-        :return: A generator of ``AttachVolume`` instances, one for each
-                 dataset which exists, is unattached, and is configured to be
-                 attached to this node.
-        """
-        for manifestation in configured.values():
-            dataset_id = UUID(manifestation.dataset_id)
-            if dataset_id in devices:
-                # It's already attached here.
+            if dataset.state != DatasetStates.MOUNTED:
```
itamarst: No action needed. 

Hm. Leases are supposed to prevent things from being moved out from under an application. So in theory if it is ATTACHED you still don't want it detached. On the other if it's not mounted then no one is using it in practice so maybe that's OK? Except sooner or later we'll have some application that wants a raw block device without a filesystem.

But this is probably OK for now, I guess.

itamarst:
1. Merge forward, with particular attention to 1771 and to code path that does `ActionNeeded`.
2. Address the above.
3. Resubmit, or better yet submit in smaller chunks.

I never got to the tests but am out of time (and ability to focus for that matter).

itamarst:
I believe this has been superceded.

PR #2177, opened 2015-11-13
[FLOC-3377] Drop sleep workaround

itamarst:
This adds a test for datasets being attached *before* container is restarted by container agent, and removes 120 second sleep on startup which is no longer necessary because of previous work.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2177)


itamarst:
See http://build.clusterhq.com/boxes-flocker?branch=drop-sleep-workaround-FLOC-3377

flocker/acceptance/obsolete/test_containers.py
```diff
@@ -335,3 +336,59 @@ def test_traffic_routed(self, cluster):
...
+
+    @require_cluster(2)
+    def test_reboot(self, cluster):
+        """
+        After a reboot the containers are only started once all datasets are
+        available locally.
+        """
+        # Find a node which is not running the control service.
+        # If the control node is rebooted, we won't get stale dataset state.
+        node = [node for node in cluster.nodes if
```
sarum90: :grey_question: Totally optional in this case, but for large lists this is slightly faster using a inline generator rather than a list. I came across a sort of cool construction for that on stackoverflow the other day and thought I'd share:

next(node for node in cluster.nodes if
        node.public_address != cluster.control_node.public_address)

flocker/acceptance/scripts/reboot.py
```diff
@@ -0,0 +1,35 @@
+"""
```
sarum90: Rename file, potentially to boot_id_service or somesuch.

I expected this to be a python file which rebooted the host it was run on, and was rather surprised that it merely reported boot_ids.

flocker/acceptance/obsolete/test_containers.py
```diff
@@ -335,3 +336,59 @@ def test_traffic_routed(self, cluster):
...
+        def got_initial_result(initial_result):
+            call([b"ssh", b"root@{}".format(node.public_address),
+                  b"shutdown", b"-r", b"now"])
+            changed = loop_until(
+                reactor, lambda: query().addCallback(
+                    lambda result: result != initial_result))
+            changed.addCallback(lambda _: query())
+
+            def result_changed(new_result):
+                self.assertEqual(
```
sarum90: I've got a question. If we rewind time to <commit where 120s sleep was introduced>~1 how would this test fail? presumably you need to interact with the cluster at the end to verify it's in a healthy state?

I feel as if we should ensure that this test fails when run against a cluster built at that commit. (Perhaps you have?)

sarum90:
Cool test idea, hopefully it works on buildbot without raising havoc :+1: 

I do have 1 request to rename `reboot.py` as it seems to not actually do any rebooting. Also some concern that this test might not verify that the original problem that forced us to add the 120s sleep is tested? I don't know the specifics of that issue, but I was surprised there weren't any cluster health checks at the end. Actually thinking about it again now perhaps the mere presence of the reboot.py job being brought back up is evidence enough that the cluster is healthy? Maybe just add a comment in that case.

Probably wait for tests to be green and address and merge at will. Thanks!

itamarst:
The original issue was that after reboot it takes some amount of time reattach cloud volume to machine. Meanwhile, container agent might rely on stale data from the control service that says dataset is already there, and so start the container too soon.

itamarst:
If I disable 120 second sleep *and* add a sleep to dataset agent this fails on older versions. So possibly the test is too hard to write in a reasonable way that actually prove we're doing the right thing... Ideas welcome.

itamarst:
Additionally the test is somewhat flaky.

itamarst:
Probably going to add per-OS code for disabling dataset agent before reboot, re-enabling after test finishes.

itamarst:
I have verified updated test fails consistently on 1.6.1 with the 120 second sleep removed.

itamarst:
Oh, except this test won't work on latest code since container won't ever start. So... I need to loop until reboot is done and then wait until container agent is up and then start the dataset agent in the test itself. Brilliant.

flocker/acceptance/node_scripts/disable_service.py
```diff
@@ -0,0 +1,20 @@
...
+if os.path.exists("/etc/redhat-release"):
+    # Redhat-based system:
+    check_call(["systemctl", "disable", service])
+    check_call(["systemctl", "stop", service])
+else:
+    # Ubuntu 14.04
+    override = "/etc/init/%.override" % (service,)
+    with file(override, "w") as f:
+        f.write("manual\n")
+    check_call(["service", service, "start"])
```
sarum90: I'm guessing this is supposed to be "stop" instead of "start" - probably add a comment as to why this isn't naively just the ubuntu version of the redhad commands if I'm wrong.

flocker/acceptance/node_scripts/disable_service.py
```diff
@@ -0,0 +1,20 @@
...
+
+service = sys.argv[1]
+
+if os.path.exists("/etc/redhat-release"):
+    # Redhat-based system:
+    check_call(["systemctl", "disable", service])
+    check_call(["systemctl", "stop", service])
+else:
+    # Ubuntu 14.04
+    override = "/etc/init/%.override" % (service,)
```
sarum90: I think this format string needs some work:

```
>>> "/etc/init/%.override" % ("cat",)
TypeError: %o format: a number is required, not str
```

Same in the other script.

flocker/acceptance/node_scripts/enable_service.py
```diff
@@ -0,0 +1,20 @@
...
+if os.path.exists("/etc/redhat-release"):
+    # Redhat-based system:
+    check_call(["systemctl", "enable", service])
+    check_call(["systemctl", "start", service])
+else:
+    # Ubuntu 14.04
+    override = "/etc/init/%s.override" % (service,)
+    if os.path.exists(override):
+        os.remove(override)
+    check_call(["service", service, "start"])
```
sarum90: This looks good to me. Specifically it looks like it will fail and exit in an error condition if it is run on a system that either doesn't have `service` installed or doesn't support `systemctl` or the `service` command. Exitting on the error condition should make ssh exit in an error condition, and consequently should result in run_ssh calling its errback, and the test should error out in that case.

sarum90:
No issues with the code, I'm fine with the enable_service script. :+1:

I do have some concern that this test has decently high probability of adding flakiness to our test suite. It's probably a waste of time to wait for two green columns on http://build.clusterhq.com/boxes-flocker?branch=drop-sleep-workaround-FLOC-3377&num_builds=10... but this feels like the sort of test that almost merits that kind of caution.

I trust your judgment. Certainly wait for green test runs, if you get them--go ahead and merge.

itamarst:
I'm going to require green because this supposed to fix the most common flaky acceptance test.

itamarst:
But on the plus side the only failing acceptance test on previous run on centos was a bug I introduced and then fixed, so I'm hopeful.

sarum90:
http://build.clusterhq.com/builders/flocker%2Facceptance%2Frackspace%2Fubuntu-14.04%2Frackspace/builds/3429/steps/trial/logs/stdio :( - I know you're done for the day, so I might dive in and see if I can figure out what happened  a little later so I can kick off  another round of tests before the end of the day.

PR #2179, opened 2015-11-16
[FLOC-3299] Flocker client support for Container API

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3299

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2179)


flocker/apiclient/_client.py
```diff
@@ -27,8 +27,10 @@
 
 from treq import json_content, content
 
+# XXX These dependencies all make splitting the client out into a separate
+# package harder.
```
sarum90: :grey_question: I'd prefer not to add this `XXX`. I'm not sure if it adds much value, and `TODOs` make me sad (they can get stale, they can be a bit overwhelming if your code is full of them, etc.).

Is it supposed to deter people from adding dependencies? That seems like the wrong answer--we don't want unnecessary duplication. Things that are added here presumably should be put in a separate package that both flocker and the client library would have as a dependency, regardless of how many there are.

Optional as I recognize this is not the most objective view of the world, but I'd prefer not adding this particular `XXX`.

flocker/apiclient/_client.py
```diff
@@ -226,6 +248,31 @@ def list_nodes():
         :return: ``Deferred`` firing with a ``list`` of ``Node``.
         """
 
+    def create_container(node_uuid, name, image):
+        """
+        :param UUID node_uuid: The ``UUID`` of the node where the container
+            will be started.
+        :param unicode name: The name to assign to the container.
+        :param DockerImage image: The ``Docker`` image which the container will
```
sarum90: :grey_question: nit: 
```inconsistent ``Docker`` vs Docker relative to line 106.```

File: flocker/apiclient/_client.py [hunk 3]
```diff
@@ -538,6 +606,44 @@ def version(self):
...
+    def _parse_configuration_container(self, container_dict):
+        """
+        Convert a dictionary decoded from JSON with a container's
+        configuration.
+
+        :param container_dict: Dictionary describing a container.
+        :return: ``Container`` instance.
+        """
+        return Container(
+            node_uuid=UUID(hex=container_dict[u"node_uuid"], version=4),
```
sarum90: :grey_question: Probably okay as is, just noting that this can generate `KeyError`'s if the container_dict doesn't meet the expected schema. Not for this PR, but perhaps we should re-use some of our fancy API schema yaml files to do proper type-checking on things returned from the server to make nicer errors. Or maybe I'm just missing where we do that.

flocker/apiclient/test/test_client.py
```diff
@@ -391,6 +473,28 @@ def test_this_node_uuid(self):
     return InterfaceTests
 
 
+def create_container_for_test(case, client, name=None):
+    """
+    :param TestCase case: The currently running test.
+    :param IFlockerClient client: The client for creating containers.
+    :param unicode name: The name to be assigned to the container or ``None``
+        to assign a random name.
```
sarum90: :grey_question: I don't know if we have a rule for what is required of docstrings... but this one is missing a summary and a returns section. I don't think the presence or absence of those docstrings will enhance or inhibit my personal development in future, so I'm indifferent, but I noticed they were missing.

I'll defer to your judgement on "what's normal" as you've worked in this codebase for longer.

sarum90:
Just some nits in documentation. Code all looks great, Thanks!

I'm going to leave for someone else to address and merge solely because one of my comments requests judgement from someone who has worked on flocker for longer than I have.

PR #2180, opened 2015-11-16
[FLOC 3420] Integrate plugin install steps to the correct places in the docs

davidoliverSP2:
Fixes 3420

Several users had contacted us about the difficulty in following the Flocker plugin for Docker installation instructions. On review, this document had introduced some steps, including the generation of API user certificates, out of sync with the rest of the docs. The way it presented also meant that users were installing the node services on each of their nodes, then having to revisit each node to install the plugin.

This PR moves the plugin install step to the Installing the Node Services page. There is a new page in Configuring for generating API certs for the plugin, and the enabling step (and testing the plugin) have been moved to the Enabling the Flocker Agent Services page.

I've also updated the docs in several areas that referenced these docs.
Note - this change means there is no longer a page titled "Installing the Flocker Plugin for Docker", as these steps have been integrated into the doc architecture as described above.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2180)


docs/install/install-node.rst
```diff
@@ -58,9 +60,17 @@ Installing on CentOS 7
...
       :prompt: [root@centos]#
 
-#. **Repeat steps 1 and 2 for all other nodes:**
+#. **Install the** ``clusterhq-flocker-docker-plugin`` **package:**
+
+   Run the following command as root on the target node:
+
+   .. prompt:: bash [root@centos]#
+   
+      yum install -y clusterhq-flocker-docker-plugin
```
jongiddy: This step depends on the commands for installing `flocker-node` having been run. Emphasise that the flocker-node steps are required, but the Flocker plugin is an optional additional step.

docs/install/install-node.rst
```diff
@@ -88,9 +98,18 @@ Installing on Ubuntu 14.04
...
       :prompt: [root@ubuntu]#
 
-#. **Repeat steps 1 and 2 on all other nodes:**
+#. **Install the** ``clusterhq-flocker-docker-plugin`` **package:**
+
+   Run the following command as root on the target node:
+
+   .. prompt:: bash [root@ubuntu]#
+   
+      apt-get install -y clusterhq-flocker-docker-plugin
```
jongiddy: As above, depends on the steps run for flocker-node

docs/install/install-node.rst
jongiddy: Also, please file an issue to replace the explicit command with a task directive and to add to installation tests.

jongiddy:
The first page in "Configuring Flocker" is the "Configuring Cluster Authentication" page, which says at the bottom that "The next topic is Generating an API User Certificate, which is used to identify yourself when sending instructions to the control service (by any method)."

This links to the second page: "Generating an API User Certificate".

However, the third page is now "Generating an API User Certificate for the Flocker Plugin for Docker". So it doesn't sound true that the second page is for sending instructions **by any method**.

Can a user skip page 2 and only perform page 3 if they only want to use the Flocker Plugin? Maybe the line at the bottom of the first page should remove the "(by any method)", and perhaps even add an explicit comment about preparing to use the Flocker Plugin for Docker.

docs/install/install-node.rst
davidoliverSP2: Created FLOC-3454 to create a task directive for installing the plugin

PR #2181, opened 2015-11-16
[FLOC 3453] Don't build storage-profiles-content.rst

davidoliverSP2:
Fixes 3453

The content in storage-profiles-content.rst is reused in a couple of places (concepts and config), so to avoid duplicating and maintaining the content in 2 places, a holding file was created, and then brought into the docs where it was required. 

However, this content was only hidden in the concepts toctree, so sphinx was still building this content, and it was viewable in a browser. 

Following a discussion with @tomprince, I've removed the content-only file, moving the text into the main storage profiles concept. I've added body tags, which are then referenced in the configuration document. There is now no hidden file.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2181)


exarkun:
It looks like the page can still be found by a search?  http://build.clusterhq.com/results/docs/hide-storageprofiles-FLOC-3453/build-11195/concepts/_storage-profiles-content.html?highlight=profiles

davidoliverSP2:
thanks @exarkun - I'll do some more research

davidoliverSP2:
@exarkun - an update:

I'm having trouble replicating this problem in 1.7.2 - when I search for `storage profiles` the <no title> page is not appearing in the search: 
https://docs.clusterhq.com/en/1.7.2/search.html?q=storage+profiles&check_keywords=yes&area=default

You can get to it directly though:
https://docs.clusterhq.com/en/1.7.2/concepts/storage-profiles-content.html

I've asked @tomprince for some assistance as I can't find any Sphinx documentation about hiding this page.

itamarst:
I suspect the search thing is due to Sphinx caching of builds.

itamarst:
Looking good.

PR #2182, opened 2015-11-16
[FLOC 2390] add favicon to docs

davidoliverSP2:
Fixes 2390

The docs pages have been missing the favicon that is present for the rest of the website. This adds it.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2182)

### DIFF
--- a/docs/_themes/clusterhq/layout_docs.html
+++ b/docs/_themes/clusterhq/layout_docs.html
@@ -65,6 +65,8 @@
             })();
         </script>
     {% endif %}
+	
+    <link rel="shortcut icon" href="https://clusterhq.com/assets/favicon.ico"/>
 </head>
 <body>
     <!-- Menu area -->


docs/_themes/clusterhq/layout_docs.html
```diff
@@ -65,6 +65,8 @@
             })();
         </script>
     {% endif %}
+	
+    <link rel="shortcut icon" href="https://clusterhq.com/assets/favicon.ico"/>
```
exarkun: Hm.  Considering clusterhq.com is maintained separately from this, do we want this foreign dependency?  What does @msh100 think about docs.clusterhq.com reaching over and grabbing clusterhq.com's favicon.ico?

exarkun: Also, do we have a policy about mixed content (http and https)?  Should a URL like this be protocol-relative (`//domain/path`)?

A quick search of the internet suggest people don't like this anymore and everything should just be `https` all the time, so maybe not relevant (except maybe the doc standard should point this out, if this is in fact what we want to do - I note that immediately above we use the ga blob that picks http or https for you to avoid mixed content, maybe we want to fix that).

exarkun:
Thanks.  I suspect it'd be better if we imported the favicon into the Flocker docs and referred to it there - but if @msh100 doesn't think it's a problem to do it this way, or if he has a suggestion for some other way to do it, I'd just follow his lead.

Address that and maybe the other point if you think it's valid, then merge.

davidoliverSP2:
thanks @exarkun 
@msh100 was with me when we put this PR together, but I'll see what he says before merging.

docs/_themes/clusterhq/layout_docs.html
msh100: The favicon living in two different places doesn't really make sense in my opinion unless we were to have two different images. I guess it makes sense to put some policy on the website however to ensure this favicon doesn't move.

As for `//`, the website will **always** redirect to https so I guess it doesn't matter so much. Also if the docs are built locally and the client uses something like `file://`, the favicon would be broken (but that's maybe a case for adding the favicon locally?).

For the time I think this PR is fine to merge, if we choose that in fact the favicon should be the bird logo (something not the same as the website) or for some other reason it should be enclosed in this repo, it's a quick change that I am happy to handle.

PR #2183, opened 2015-11-16
[FLOC 3419] add a margin to the bottom of lists, except if it is part of the toctree

davidoliverSP2:
Fixes 3419

In bulleted lists, nested in a numbered list, the padding at the bottom of the last bullet was missing padding. This PR adds 10px to the bottom of the list, but limits this change so as not to impact the toctree listings used throughout the docs.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2183)

### DIFF
--- a/docs/_themes/clusterhq/static/css/docs.css
+++ b/docs/_themes/clusterhq/static/css/docs.css
@@ -305,6 +305,11 @@ a.download:before {
 .page ul, ol {
     color: #646570;
     font-family: lato;
+    margin-bottom: 10px;
+}
+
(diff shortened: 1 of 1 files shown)


exarkun:
Thanks.  I can verify that the spacing looks more consistent now (in Firefox 42).

I can't say much about the css changes beyond that.  I'm left with the same feeling I usually have after reading css changes which is that in month no one will have any clue why these particular rules are here.  Maybe that's just how css is meant to be, though.

Go ahead and merge.  Thanks again.

PR #2184, opened 2015-11-16
[FLOC-3457] Merge 1.7.2 docs into master

jongiddy:
Flocker 1.7.2 contains back-ported changes from `docs` in `master`. These need to be re-merged into `master` to make the 1.7.2 release PR only contain changes to release notes.

Please check that this branch does no work when merged into master.
`git merge remerge-docs-FLOC-3457` runs OK, but changes no files.

You can also check that the changes are only to the `docs` directory.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2184)


itamarst:
I'm seeing a whole bunch of diffs, so doesn't really seem like this does no work.

jongiddy:
Are you merging into an up-to-date master?  The PR contains changes but they are copies of existing changes, so the result of a merge on the command line should be no files changed.

PR #2185, opened 2015-11-16
[FLOC-3447] Calculate changes for wakeups

itamarst:
calculate_changes() is sometimes called with cached local state to decide if the sleeping converence loop should wake up.

1. If it throws an exception (indicating a bug) then wake up; better to be responsive than to sleep when there is work to be done.
2. Fix a case where it threw an exception due to using cached local state that was missing information

The latter case was happening in acceptance tests and may have been causing timeouts since the loop would not wake up.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2185)

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -23,7 +23,9 @@ from random import uniform
 
 from zope.interface import implementer
 
-from eliot import ActionType, Field, writeFailure, MessageType
+from eliot import (
+    ActionType, Field, writeFailure, MessageType, write_traceback,
+)
 from eliot.twisted import DeferredContext
 
 from characteristic import attributes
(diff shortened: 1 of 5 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -688,6 +688,32 @@ def got_compute_id(compute_instance_id):
 
 
 @implementer(IStateChange)
+class AttachNeeded(PClass):
+    """
+    We need to attach an unattached volume to this node (the node of the
+    deployer it is run with) but lack the necessary information to do so.
+
+    Creating this is still useful insofar as it may let the convergence
```
itamarst: Well, it's better information than "too stale". If you've reached this point you know that you probably would've attached, i.e. actions would be taken. So maybe `ActionNeeded` or something.

itamarst:
There's separately the 120 second timeout + 120 second startup which I suspect, which I am fighting in a different issue.

flocker/node/_loop.py
```diff
@@ -391,8 +393,16 @@ def output_UPDATE_MAYBE_WAKEUP(self, context):
...
-                self.configuration, self.cluster_state, discovered) != NoOp():
+        try:
+            changes = self.deployer.calculate_changes(
+                self.configuration, self.cluster_state, discovered)
+        except:
+            # Something went wrong in calculation due to a bug in the
+            # code. We should wake up just in case in order to be more
+            # responsive.
+            write_traceback()
+            changes = None
```
itamarst: Longer term I think we want any `eliot:traceback` message to make build go red for acceptance tests, so I don't think we need to do anything special h

PR #2186, opened 2015-11-16
[FLOC-3456] Release Flocker 1.7.2

jongiddy:
Release adds doc changes from `master` that do not have any related changes in code.  In particular, this was done to fix FLOC-3420, which was being reported by users.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2186)


sarum90:
Verified that this release has the docs from master, and everything else identical to 1.7.1

Release notes look good.

Tests look good, earlier failures look like flakes. Given that this is identical to 1.7.1 and that you have passing runs on both rackspace and aws and centos and ubuntu, I think you can proceed even before current outstanding tests conclude.

Thanks!

PR #2188, opened 2015-11-17
[FLOC-3452] Docker plugin looks up volume by name

itamarst:
As a temporary hack the docker plugin creates volumes with a hashed dataset id in order to prevent duplicate creates. When looking up datasets, however, there is no need to rely on this mechanism and we should just use the name in metadata so we are fully compatible with volumes created by Flocker outside the plugin.

In addition this involved improving the error handling in that code path, since previously we just blew up mysteriously.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2188)


flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -310,6 +312,18 @@ def test_run_container_with_volume(self, cluster):
...
+    def test_run_container_with_preexisting_volume(self, cluster):
+        """
+        Docker can run a container with a volume provisioned by Flocker before
+        it is mentioned to the Docker plugin.
+        """
+        name = random_name(self)
+        d = create_dataset(self, cluster, metadata={u"name": name})
+        d.addCallback(lambda _: self._test_create_container(
+            cluster, volume_name=name))
+        return d
```
sarum90: :smiley: Feature work has such obvious and simple tests.

flocker/dockerplugin/_api.py
```diff
@@ -60,11 +66,29 @@ def decorator(f):
             schema_store=SCHEMAS,
             ignore_body=ignore_body)
         def wrapped(*args, **kwargs):
-            return f(*args, **kwargs)
+            d = maybeDeferred(f, *args, **kwargs)
```
sarum90: :grey_question: Potentially add to the docstring above the new functionality that this PR is adding to the decorator.

sarum90:
Thanks, code looks great - a couple nits, nothing major. Address and merge.

A fairly large number of test failures though. Might just be the state of things these days though.

itamarst:
Well, I'll merge forward and see. We really ought to be getting greener builds :crying_cat_face:

itamarst:
Going to update this to return 200 OK instead of 404/500 based on Marcus' research that suggests Docker only tests for error reporting in the 200 OK case.

itamarst:
This passed completely before tweaking response codes; it's stil passing on both Ubuntu and CentOS, and the change of response codes is completely unrelated to cloud. Given that Rackspace is borked I'm just going to merge this.

PR #2191, opened 2015-11-18
[FLOC-3199] Better random port assignment

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3199

This changes a handful of acceptance tests to let Docker assign a random external port number for a container.  Previously the tests tried to assign the port number using `find_free_port`.  `find_free_port` only works most of the time even in ideal circumstances.  Any code using it should also be prepared to retry, calling `find_free_port` to get a new port number.

However, the acceptance tests aren't the ideal circumstances for `find_free_port` - at least not where they're picking a port number for a container to use.  The containers run on a different system; `find_free_port` only attempts to find a port that's free on the local system.  Basically, `find_free_port` should not be used in the acceptance tests.

Docker's random port number assignment *also* only works most of the time and code relying on it should be prepared to retry.  In the interest of incremental improvement, I haven't implemented those retries here.  I'll do that as a follow-up.


flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -325,12 +339,11 @@ def _test_move(self, cluster, origin_node, destination_node):
         client = get_docker_client(cluster, origin_node.public_address)
         data = "hello world"
         http_port = 8080
-        host_port = find_free_port()[1]
         volume_name = random_name(self)
         container_args = {
             "host_config": client.create_host_config(
                 binds=["{}:/data".format(volume_name)],
-                port_bindings={http_port: host_port}),
```
jml: Hooray! I am genuinely surprised we didn't do this from day 1.

flocker/acceptance/testtools.py
```diff
@@ -928,6 +928,35 @@ def created(response):
...
 
+def extract_external_port(
+    client, container_identifier, internal_port
+):
+    """
+    Inspect a running container for the external port number on which a
+    particular internal port is exposed.
+
+    :param docker.Client client: The Docker client to use to perform the
+        inspect.
```
jml: Why is this in terms of the docker client, rather than our own interface around the docker client?

jml:
LGTM, thanks.

flocker/acceptance/testtools.py
exarkun: The tests that want to use it use docker-py and not our client.  I suppose because they don't want to have the extra complexity of our layer involved? Like, it's more code to generate failures that aren't relevant to what's being tested.  Maybe also because docker-py is synchronous and ours is asynchronous and it's more convenient to use a synchronous API than an asynchronous one.  Also ours is private inside `flocker.node`.

exarkun:
BuildBot says this actually makes things worse, perhaps by replacing an interaction that's not very likely to fail (create and start a container with a port I hope is free) with an interaction that seems very prone to failure (create and start a container and then inspect it to see what port it was assigned).

So perhaps this needs some kind of retry logic at the Docker interface.  I think we need that in general so maybe I'll implement it generally and see if it helps in this case specifically.

exarkun:
One problem with the branch in its current form is that it assumes ephemeral ports assigned by Docker will remain the same across container restarts.  Sometimes this happens but it's not guaranteed.

exarkun:
Most recent revision fixes the problem of not re-extracting the ephemeral port.  I'm slightly concerned that there's a race condition between the container starting up and the port being extracted, though.

exarkun:
Well, as it turns out, it's even worse than that.  There's an earlier race condition between when Docker says the container exists and has given back a unique identifier for it and when Docker will respond to a container inspect request with container information instead of a 404.

exarkun:
This seems to be in better shape now.  Re-submitting.

flocker/acceptance/testtools.py
itamarst: And ideally we're throwing ours out once we stop supporting the container agent.

itamarst:
Seems reasonable. Merge when appropriately green.

PR #2192, opened 2015-11-18
[FLOC-3443] Retry tests marked as flaky 

jml:
This branch adds support for retrying flaky tests.

It adds thing that wraps an arbitrary `RunTest` object with something that retries tests until they pass a minimum number of times, and updates our `async_runner` to use that, which means that any test case which inherits from `AsyncTestCase` and has a `@flaky` decorator will automatically retry.

The implementation approach means that it should also work for synchronous test cases, once we switch those over to testtools.

An earlier version of this branch used the [flaky](https://github.com/box/flaky) library. Unfortunately, there is very little that it provides which is useful to us.

You might notice the `_run_prepared_result` call. This is an infelicity in testtools, which uses underscore prefixes for both private and "protected" methods. 

Slight diff bloat due to changing all existing `@flaky` markers to use unicode literals. Tradeoffs called out w/ `XXX` comments.

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -79,8 +79,7 @@ class IDockerClientNamespacedTests(make_idockerclient_tests(
     def setUp(self):
         pass
 
-    # XXX: First time we've needed multiple JIRA keys. Is this the right API?
-    @flaky(['FLOC-2628', 'FLOC-2874'])
+    @flaky([u'FLOC-2628', u'FLOC-2874'])
     def test_added_is_listed(self):
         return super(IDockerClientNamespacedTests, self).test_added_is_listed()
 
@@ -532,7 +531,7 @@ class GenericDockerClientTests(TestCase):
     # XXX: This is subclassed, and the subclass decorated with a different
     # JIRA key. When @flaky actually does something, we should have tests that
     # cover that case.
-    @flaky('FLOC-3077')
+    @flaky(u'FLOC-3077')
     def test_pull_image_if_necessary(self):
         """
         The Docker image is pulled if it is unavailable locally.
@@ -1082,7 +1081,7 @@ class GenericDockerClientTests(TestCase):
         d.addCallback(self.assertEqual, "1")
         return d
 
-    @flaky('FLOC-2840')
+    @flaky(u'FLOC-2840')
     def test_restart_policy_always(self):
         """
         An container with a restart policy of always is restarted
@@ -1326,7 +1325,7 @@ class NamespacedDockerClientTests(GenericDockerClientTests):
         d.addCallback(self.assertEqual, set())
         return d
 
-    @flaky('FLOC-1657')
+    @flaky(u'FLOC-1657')
     def test_pull_image_if_necessary(self):
         return super(
             NamespacedDockerClientTests, self).test_pull_image_if_necessary()
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -528,9 +528,6 @@ class GenericDockerClientTests(TestCase):
         d.addCallback(started)
         return d
 
-    # XXX: This is subclassed, and the subclass decorated with a different
-    # JIRA key. When @flaky actually does something, we should have tests that
-    # cover that case.
     @flaky(u'FLOC-3077')
     def test_pull_image_if_necessary(self):
         """
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -4,7 +4,7 @@
 Logic for handling flaky tests.
 """
 
-from functools import wraps
+from functools import partial
 
 import testtools
 
@@ -39,15 +39,27 @@ class _RetryFlaky(object):
     """
     ``RunTest`` implementation that retries tests that fail.
     """
+    # XXX: Make this a pyrsistent object
 
     # XXX: This should probably become a part of testtools:
     # https://bugs.launchpad.net/testtools/+bug/1515933
 
-    def __init__(self, run_test):
-        self._run_test = run_test
+    # TODO: I think the "right" thing to do is to take a RunTest factory and
+    # construct RunTest once per test run, each time giving it a newly-created
+    # TestResult.
+    #
+    # When we're "done", we aggregate all of the results and output that to
+    # the result we were given.
+
+    def __init__(self, run_test_factory, case, *args, **kwargs):
+        self._run_test_factory = run_test_factory
+        self._case = case
+        self._args = args
+        self._kwargs = kwargs
 
     def run(self, result=None):
-        return self._run_test.run(result)
+        return self._run_test_factory(
+            self._case, *self._args, **self._kwargs).run(result)
 
 
 def retry_flaky(run_test_factory=None):
@@ -57,8 +69,4 @@ def retry_flaky(run_test_factory=None):
     if run_test_factory is None:
         run_test_factory = testtools.RunTest
 
-    # XXX: I feel as if there's a simpler, more standard way of doing this.
-    @wraps(run_test_factory)
-    def wrapped(*args, **kwargs):
-        return _RetryFlaky(run_test_factory(*args, **kwargs))
-    return wrapped
+    return partial(_RetryFlaky, run_test_factory)
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -6,7 +6,7 @@ Logic for handling flaky tests.
 
 from functools import partial
 
-from pyrsistent import PClass, field, pmap
+from pyrsistent import PClass, field, pmap, pset, pset_field
 import testtools
 from testtools.testcase import gather_details
 
@@ -16,7 +16,6 @@ _FLAKY_ATTRIBUTE = '_flaky'
 
 # TODO:
 # - make sure that JIRA data is included in flaky annotations
-#   - do "either text or sequence" UI for jira_keys
 # - handle tests with many different kinds of results
 
 
@@ -30,7 +29,7 @@ def flaky(jira_keys, max_runs=5, min_passes=2):
 
     :param unicode jira_keys: The JIRA key of the bug for this flaky test,
         e.g. 'FLOC-2345'. Can also be a sequence of keys if the test is flaky
-        fr multiple reasons.
+        for multiple reasons.
     :param int max_runs: The maximum number of times to run the test.
     :param int min_passes: The minimum number of passes required to treat this
         test as successful.
@@ -43,7 +42,11 @@ def flaky(jira_keys, max_runs=5, min_passes=2):
     # - allow specifying which exceptions are expected
     # - provide interesting & parseable logs for flaky tests
 
-    annotation = _FlakyAnnotation(max_runs=max_runs, min_passes=min_passes)
+    if isinstance(jira_keys, basestring):
+        jira_keys = [jira_keys]
+
+    annotation = _FlakyAnnotation(
+        jira_keys=pset(jira_keys), max_runs=max_runs, min_passes=min_passes)
 
     def wrapper(test_method):
         setattr(test_method, _FLAKY_ATTRIBUTE, annotation)
@@ -71,9 +74,18 @@ class _FlakyAnnotation(PClass):
                      invariant=lambda x: (x > 0, "must run at least once"))
     min_passes = field(int, mandatory=True,
                        invariant=lambda x: (x > 0, "must pass at least once"))
+    jira_keys = pset_field(unicode, optional=False)
 
-    __invariant__ = lambda x: (x.max_runs >= x.min_passes,
-                               "Can't pass more than we run")
+    __invariant__ = lambda x: (
+        x.max_runs >= x.min_passes and len(x.jira_keys) > 0,
+        "Can't pass more than we run and must provide a jira key")
+
+    def to_dict(self):
+        return {
+            'max_runs': self.max_runs,
+            'min_passes': self.min_passes,
+            'jira_keys': set(self.jira_keys),
+        }
 
 
 def retry_flaky(run_test_factory=None):
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -123,7 +123,10 @@ class _RetryFlaky(testtools.RunTest):
         else:
             # XXX: Need to actually provide data about the errors.
             # XXX: How are we going to report on tests that sometimes fail,
-            # sometimes error.
+            # sometimes error. Probably "if all failures, failure; otherwise,
+            # error"
+            # XXX: Consider extracting this aggregation into a separate
+            # function.
             result.addError(case, details={})
         result.stopTest(case)
 
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -160,9 +160,12 @@ class _RetryFlaky(testtools.RunTest):
                 successes += 1
             results.append(details)
 
+        flaky_data = flaky.to_dict()
+        flaky_data.update({'runs': len(results), 'passes': successes})
         flaky_details = {
-            'flaky': text_content(pformat(flaky.to_dict())),
+            'flaky': text_content(pformat(flaky_data)),
         }
+
         details = _combine_details([flaky_details] + results)
         result.startTest(case)
         if successes >= flaky.min_passes:
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -244,7 +244,7 @@ class _RetryFlaky(testtools.RunTest):
         if result_type == _ResultType.skip:
             # XXX: Work around a testtools bug where it reports stack traces
             # for skips that aren't passed through its supported
-            # SkipException.
+            # SkipException: https://bugs.launchpad.net/testtools/+bug/1518100
             [reason] = list(tmp_result.skip_reasons.keys())
             details = details.discard('traceback').set(
                 'reason', text_content(reason))
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -14,7 +14,7 @@ from testtools.testcase import gather_details
 _FLAKY_ATTRIBUTE = '_flaky'
 
 
-def flaky(jira_key, max_runs=5, min_passes=2):
+def flaky(jira_keys, max_runs=5, min_passes=2):
     """
     Mark a test as flaky.
 
@@ -22,8 +22,9 @@ def flaky(jira_key, max_runs=5, min_passes=2):
     test as flaky means both failures and successes are expected, and that
     neither will fail the test run.
 
-    :param unicode jira_key: The JIRA key of the bug for this flaky test,
-        e.g. 'FLOC-2345'
+    :param unicode jira_keys: The JIRA key of the bug for this flaky test,
+        e.g. 'FLOC-2345'. Can also be a sequence of keys if the test is flaky
+        fr multiple reasons.
     :param int max_runs: The maximum number of times to run the test.
     :param int min_passes: The minimum number of passes required to treat this
         test as successful.
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -14,6 +14,14 @@ from testtools.testcase import gather_details
 _FLAKY_ATTRIBUTE = '_flaky'
 
 
+# TODO:
+# - actually run a flaky test manually to see what everything looks like
+# - make sure that JIRA data is included in flaky annotations
+#   - do "either text or sequence" UI for jira_keys
+# - attach details to success
+# - handle tests with many different kinds of results
+
+
 def flaky(jira_keys, max_runs=5, min_passes=2):
     """
     Mark a test as flaky.
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -240,14 +240,14 @@ class _RetryFlaky(testtools.RunTest):
         tmp_result = testtools.TestResult()
         self._run_test(case, tmp_result)
         result_type = _get_result_type(tmp_result)
+        details = pmap(case.getDetails())
         if result_type == _ResultType.skip:
             # XXX: Work around a testtools bug where it reports stack traces
             # for skips that aren't passed through its supported
             # SkipException.
             [reason] = list(tmp_result.skip_reasons.keys())
-            details = pmap({'reason': text_content(reason)})
-        else:
-            details = pmap(case.getDetails())
+            details = details.discard('traceback').set(
+                'reason', text_content(reason))
         _reset_case(case)
         return (tmp_result.wasSuccessful(), result_type, details)
 
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -63,18 +63,6 @@ class _RetryFlaky(testtools.RunTest):
         self._args = args
         self._kwargs = kwargs
 
-    def _run_test(self, case, result):
-        """
-        Create an instance of the ``RunTest`` we are wrapping.
-
-        :param testtools.TestCase case: The test to run.
-        :param testtools.TestResult result: The test result to report to.
-            Must conform to testtools extended test result interface.
-        :return: The modified ``result``.
-        """
-        run_test = self._run_test_factory(case, *self._args, **self._kwargs)
-        return run_test._run_prepared_result(result)
-
     def _run_prepared_result(self, result):
         """
         Run the test with a result that conforms to testtools' extended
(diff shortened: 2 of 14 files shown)


jml:
http://build.clusterhq.com/builders/flocker-centos-7/builds/7929 failed with a `@flaky` test. Retry logic is not yet enabled for `flocker.node.functional.test_docker.GenericDockerClientTests`, because it does not yet inherit from `AsyncTestCase`

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
     # - provide interesting & parseable logs for flaky tests
 
+    if isinstance(jira_keys, basestring):
+        jira_keys = [jira_keys]
+
+    annotation = _FlakyAnnotation(
+        jira_keys=pset(jira_keys), max_runs=max_runs, min_passes=min_passes)
+
     def wrapper(test_method):
+        setattr(test_method, _FLAKY_ATTRIBUTE, annotation)
```
sarum90: :grey_question: Do you want to do anything special if someone added the flaky decorator twice? looks like this would just stomp on the inner one presently. That's probably okay, although we might lose some jira_keys. Okay as is  / at least we won't be retrying a retry loop (by my reading anyway)

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
+
+def _get_flaky_annotation(case):
+    """
+    Get the flaky decoration detail from ``case``.
+
+    :param TestCase case: A test case that might have been decorated with
+        @flaky.
+    :return: ``None`` if not flaky, or a ``pmap`` of the flaky test details.
+    """
+    # XXX: Is there a public way of doing this?
```
sarum90: :grey_question: Looks like you are doing it in the standard way, or at least the way it's done within testtools:
https://github.com/testing-cabal/testtools/blob/master/testtools/testcase.py#L216

Might I suggest making this TODO a little more future-actionable by openning an issue against testtools to make this method public, and referencing that issue here?

flocker/testtools/_flaky.py
```diff
@@ -4,8 +4,19 @@
...
+from pyrsistent import PClass, field, pmap, pset, pset_field
+import testtools
+from testtools.content import text_content
+from testtools.testcase import gather_details
+
+
+_FLAKY_ATTRIBUTE = '_flaky'
+
+
+def flaky(jira_keys, max_runs=5, min_passes=2):
```
sarum90: :grey_question: min_passes of 2 will always at least double the time to run for some of our large, considerably more flaky test suites, right? Personally I think defaults here of (max=3, min=1) would more accurately depict automating my current practices as an engineer.

Marking as optional because I'm guessing you've probably already had slack conversations and have some pretty good reason for believing 5/2 are good defaults for these values. I'm also fine with trying those out, but the wait time for AWS runners is already really long. Hopefully this decreases net wait time, but doubling the time for some of the tests seems as if it might threaten that.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
+
+class _FlakyAnnotation(PClass):
+
+    max_runs = field(int, mandatory=True,
+                     invariant=lambda x: (x > 0, "must run at least once"))
+    min_passes = field(int, mandatory=True,
+                       invariant=lambda x: (x > 0, "must pass at least once"))
+    jira_keys = pset_field(unicode, optional=False)
+
+    __invariant__ = lambda x: (
```
sarum90: :grey_question: nit:

From http://pyrsistent.readthedocs.org/en/latest/intro.html#invariants:
```
>>> class EvenX(PRecord):
...     x = field(invariant=lambda x: ((x > 0, 'x negative'), (x % 2 == 0, 'x odd')))
...
```

so:
```
 __invariant__ = lambda x: (
         (x.max_runs >= x.min_passes, "Can't pass more than we run"),
         (len(x.jira_keys) > 0, "Must provide a jira key"))
```
or the like should do the trick.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
+        self._args = args
+        self._kwargs = kwargs
+
+    def _run_prepared_result(self, result):
+        """
+        Run the test with a result that conforms to testtools' extended
+        ``TestResult`` interface.
+
+        This overrides a method in base ``RunTest`` which is intended to be
+        overwritten.
```
sarum90: Oh, good old unittest using all the design patterns that makes things impossible to follow :)

This is just a drive-by thank you for this line in the docstring! Very helpful while reading the code.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
 
     # TODO (see FLOC-3281):
     # - allow specifying which exceptions are expected
-    # - retry on expected exceptions
-    #   - parametrize on retry options
     # - provide interesting & parseable logs for flaky tests
 
+    if isinstance(jira_keys, basestring):
```
exarkun: :question: This may be a losing battle ... but ... I don't think `basestring` is what you mean here.  Based on all the byte->unicode changes in this branch, I think you mean `unicode` here :smile: At worst, I think you mean `(bytes, unicode)` though.  `basestring` used to mean "The C struct layout of this object is compatible with PyString_Type / PyUnicode_Type" and then CPython changed the C struct layout of PyUnicode_Type so it was impossible to be compatible with both so now `basestring` doesn't mean anything afaict.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
+    """
+    Reset ``case`` so it can be run again.
+    """
+    # XXX: Alternative approach is to use clone_test_with_new_id, rather than
+    # resetting the same test case.
+    # Don't want details from last run.
+    case.getDetails().clear()
+    # https://github.com/testing-cabal/testtools/pull/165/ fixes this.
+    case._TestCase__setup_called = False
+    case._TestCase__teardown_called = False
```
exarkun: Does this fix https://clusterhq.atlassian.net/browse/FLOC-3462 ?

flocker/testtools/test/test_flaky.py
```diff
@@ -42,63 +82,138 @@ def test_successful_flaky_test(self):
...
+        )
+
+    def test_intermittent_flaky_test(self):
         """
-        As of FLOC-3414, the @flaky decorator doesn't actually treat failed
-        tests in special in any way - it just acts as a structured comment.
+        A @flaky test that fails sometimes and succeeds other times counts as a
+        pass, as long as it passes more than the given min_passes threshold.
         """
+        # XXX: Do we have a 'shuffled' strategy?
```
sarum90: I think https://github.com/DRMacIver/hypothesis/blob/49d860822639c2426a9bcd9ebf55c0b6e0ee472f/src/hypothesis/strategies.py#L706 might be what you are looking for?

sarum90:
Is there any way for this to indicate what it's doing in the log? I think that would be almost essential for knowing if we can remove some decorators.

By looking at http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fubuntu-14.04%2Faws/builds/3513/steps/trial/logs/stdio I could not tell that a bunch of the tests were run twice.

but http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fubuntu-14.04%2Faws/builds/3513 took 34 minutes whereas your random run on master right now takes 27 minutes (http://build.clusterhq.com/builders/flocker%2Facceptance%2Faws%2Fubuntu-14.04%2Faws/builds/3511) so... I think the flaky decorator was working?

flocker/node/functional/test_docker.py
```diff
@@ -532,7 +531,7 @@ def started(_):
     # XXX: This is subclassed, and the subclass decorated with a different
```
sarum90: I don't think you tested this, did you want to do that?

flocker/testtools/_flaky.py
```diff
@@ -22,10 +37,179 @@ def flaky(jira_key):
...
+            testtools extended result interface.
+        :param _FlakyAnnotation flaky: A description of the conditions of
+            flakiness.
+
+        :return: A ``TestResult`` with the result of running the flaky test.
+        """
+        successes = 0
+        results = []
+
+        while successes < flaky.min_passes and len(results) < flaky.max_runs:
```
sarum90: :grey_question: Optional: change the second half of the condition to: (len(results) + flaky.min_passes - successes <= flaky.max_runs)
It'll save us some runs/time/money/energy if a flaky test configured for 5/2 has already failed 4 times.

sarum90:
This look great! :+1: I'm excited to get this turned on and stop having to kick buildbot so much.

Although.... right now I think I'm going to reject, mostly for 1 reason. I can't actually tell for which tests on buildbot the flaky decorator was actually working (I think from your comment above that it actually is turned on for some tests?). I feel like for this work to be effective, engineers will have to be able to tell which tests were automatically retried and what the results of those retries were. I feel like if I can't tell as the reviewer, then your average ClusterHQian almost certainly can't tell.

If I'm crazy and it's just not turned on feel free to ignore me. All of my other comments were just optional nits and potential marginal improvements that I saw. Either way if you are looking for second review I'll try to make it a quick one - just ping me on slack when it's ready.

flocker/testtools/_flaky.py
jml: TBH, I was ambivalent about providing this DWIM support at all. I'd rather each parameter have just one type. It's not _that_ much syntactic burden for callers to wrap things in a list.

Changed to `unicode` for now, but all I need is the slightest breath of a suggestion that it'd be OK to always take a sequence and I'll change it & update the call sites.

flocker/testtools/_flaky.py
jml: testtools actually strives for composition instead of inheritance, but they don't always get it right. I think they'd be open to patches.

flocker/testtools/_flaky.py
jml: Changed, thanks!

flocker/testtools/_flaky.py
jml: testing-cabal/testtools#165 as currently written will at least mitigate the problem. I hope that before it lands it will fix it entirely.

flocker/testtools/test/test_flaky.py
jml: Yes, thanks!

flocker/testtools/_flaky.py
jml: Changed. It's possible that we'll feel the lack of information, but if we do it's simple enough to change back.

jml: Had to update tests, because this change combined with `permutations` usage means number of runs is non-deterministic.

flocker/testtools/_flaky.py
jml: Actually, we haven't had those conversations. I picked those numbers based on the upstream box/flaky defaults.

Changed to your suggested defaults.

flocker/testtools/_flaky.py
jml: Done: https://bugs.launchpad.net/testtools/+bug/1517867

We have moderately good relations with the maintainers of the testtools project, so there's a good chance that bugs like this can get fixed.

flocker/testtools/_flaky.py
jml: We don't retry a retry loop. I've updated this to merge jira_keys and use the max of the other values.

flocker/node/functional/test_docker.py
jml: I've done it. Not 100% sure it adds value. Either way the comment can go.

jml:
Signalling flakiness for failed tests is easy: the attached `flaky` dict will be included in the output, and all of the tracebacks will be included as well. Signalling flakiness for successful tests is harder than it should be. 

Currently, we swallow twisted logs and only emit them on failures ([FLOC-3469](https://clusterhq.atlassian.net/browse/FLOC-3469) and [lp:942785](https://bugs.launchpad.net/testtools/+bug/942785)). This means that even if we add logging, it won't actually appear on success.

I would very much like to fix that bug. After that bug is fixed, I would revisit this code to add logging. 

Other options:
* just print something to stdout or stderr
    * using `warnings` for the last
* rename the flaky test, recording it as `test_foo_bar_FLAKY`
* record flaky successes as "unexpected success", which doesn't mark the test run as failed, but _is_ recorded faithfully in trial, and output to the user. There are compatibility issues that make this complicated, as the idea of "expected failure" was re-invented by stdlib after trial grew the feature.

jml:
"Unexpected success" solution fails given https://twistedmatrix.com/trac/ticket/4811 (filed in testtools as https://bugs.launchpad.net/twisted/+bug/1517916 but it's really a bug in trial)

jml:
I've settled for "just print something to stdout". Here's what it looks like:

```
 FooTests
    test_something ... @flaky(flocker.testtools.test.test_flaky.FooTests.test_something): passed 1 out of 1 runs (min passes: 1; max runs: 3)                                                    [OK]
```

jml:
The lint check is fixed. I'm not going to bother triggering a whole run for 5662891, and will just stick to 0ddd411 for the test results.

sarum90:
Playing with this a bit locally because I wanted to see what failures looked like (all of the tests on buildbot seemed to have passed). Realized that skipped tests just come out as passed - I think putting this in master as-is has high probability of confusing new engineers. My workflow as a new starter was "try to run test X, see skiptest message, fix skiptest message, actually run test X". Chatted with @jml offline and we think we should at least plumb the skiptest message out. Putting off next review until after that is added.

jml:
Here's a thing that maybe addresses the skip test issue. `_run_flaky_test` is just about at my personal threshold for complexity, but I'm too tired right now to think of clean abstractions. Obvious points are moving the bit that does the running into a separate thing, or moving the bit that translates the result of the run into something appropriate for the main `TestResult`.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +43,274 @@ def flaky(jira_key):
...
+
+
+def _combine_flaky_annotation(flaky1, flaky2):
+    """
+    Combine two flaky annotations.
+    """
+    return _FlakyAnnotation(
+        jira_keys=flaky1.jira_keys | flaky2.jira_keys,
+        max_runs=max(flaky1.max_runs, flaky2.max_runs),
+        min_passes=max(flaky1.min_passes, flaky2.min_passes),
```
sarum90: :grey_question: So, I think this PR has pretty much blown it's complexity budget so don't necessarily take action on this point, but I have some interesting observations here.

```
@flaky('XXX', min_passes=1, max_runs=100) # If this ever passes it means the code is probably fine.
@flaky('YYY', min_passes=3, max_runs=3) # This test is quick, and I think all the flakes are fixed, run it a bit more frequently for the next week to verify.
def test_...
```
Resolves kinda funny, but shame on us if we ever set up a situation like that.

More interestingly:
```
@flaky('XXX')
@flaky('YYY', min_passes=1, max_runs=2) # test is almost stable, and failures take 5 min to time out.
```
resolves oddly as well.

My hunch is using the values from whichever one has the higher standards is the right call:
```
def ratio(f):
   return float(f.min_passes)/float(f.max_runs)

higher_standards = flaky1 if ratio(flaky1) > ratio(flaky2) else flaky2
```
but... as I said complexity budget running low and this is just test infrastructure, so what you have here certainly is good enough.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +43,274 @@ def flaky(jira_key):
...
+        raise ValueError('%r has run %d tests, 1 expected' % (
+            result, result.testsRun))
+
+    if len(result.errors) > 0:
+        return _ResultType.error
+    elif len(result.failures) > 0:
+        return _ResultType.failure
+    elif len(result.expectedFailures) > 0:
+        return _ResultType.expected_failure
+    elif len(result.unexpectedSuccesses) > 0:
```
sarum90: :grey_question: I've never understood what expectedFailures and unexpectedSuccesses are for... but should unexpectedSuccesses be a failure condition and thus higher priority than expectedFailures? This is at most a nit, and more probably a casual musing.

sarum90:
Seeing a bunch of errors like:
```
Failure: testtools.testresult.real._StringException: Empty attachments:
  twisted-log
  twisted-log-1
  twisted-log-2
```
When I run tests that I make fail 3 times in a row. You know what those are about?

flocker/testtools/_flaky.py
```diff
@@ -22,10 +43,274 @@ def flaky(jira_key):
...
+            results.append((result_type, details))
+        successful = successes >= flaky.min_passes
+
+        flaky_data = flaky.to_dict()
+        flaky_data.update({'runs': len(results), 'passes': successes})
+        flaky_details = {
+            'flaky': text_content(pformat(flaky_data)),
+        }
+        combined_details = _combine_details(
+            [flaky_details] + list(r[1] for r in results))
```
sarum90: :grey_question: nit: forming a list out of a generator looks funny when you could just make a list in place:
`[r[1] for r in results]`. Obviously just a style thing, do so or not at your own discretion.

flocker/testtools/_flaky.py
```diff
@@ -22,10 +43,274 @@ def flaky(jira_key):
...
+            [reason] = list(tmp_result.skip_reasons.keys())
+            details = details.discard('traceback').set(
+                'reason', text_content(reason))
+        _reset_case(case)
+        return (tmp_result.wasSuccessful(), result_type, details)
+
+
+class _ResultType(Names):
+    """
+    Different kinds of test results.
```
sarum90: :grey_question: I feel like as a company we usually document `NamedConstant`'s. In my opinion there is no reason to do that in this case, as these are just constants to describe states that come from a dependency and are probably as self-describing as any documentation that we would add.

Just pointing it out in case you are of the opinion that we should document them and just forgot.

flocker/testtools/test/test_flaky.py
```diff
@@ -42,63 +133,167 @@ def test_successful_flaky_test(self):
...
+        self.assertThat(run_test(test), has_results(tests_run=Equals(1)))
+
+    @given(jira_keys, num_runs, num_runs,
+           streaming(text(average_size=10)).map(iter))
+    def test_flaky_skipped_test(self, jira_keys, max_runs, min_passes,
+                                reasons):
+        """
+        If a test is skipped and also marked @flaky, we report it as skipped.
+        """
+        [min_passes, max_runs] = sorted([min_passes, max_runs])
```
sarum90: cute

sarum90:
This is going to be great for development cycles. :dancing_banana: Thanks for fixing the skip messages, they're great.

A few nits throughout - none that are really that important.

I also downloaded this branch and played around with it a bit manually. If I were to have 1 complaint it's that tests that ultimately wind up passing or skipping don't log the error / traceback from the times that they failed or errorin out. I don't think that's worth holding this up though, so go ahead and merge whenever!

Also a few thoughts about future work in no particular order:
- We might only want to allow `Error`'s to be retried, hopefully `Failure`'s indicate assertions that are actually based around our software breaking. Maybe that's not that big of a deal though, probably better to have you hacking on fixing flaky tests rather than worrying about little differences like this.
- Recently I've notice _trial_tmp/test.log to be somewhat emptier than expected after I run tests, is that related to this? or is there any way that log could get the tracebacks from failed runs?

flocker/testtools/_flaky.py
jml: I think you're right: both that max strictness is the right decider and that it's probably not necessary. I'll leave it as-is for now.

flocker/testtools/_flaky.py
jml: They are for enshrining todos as tests. You can mark a test as being "expected" to fail, and then if it fails according to those conditions, the suite still passes. If it succeeds, that is reported as an unexpected success.

The `TestResult` implementation gets to decide whether or not an unexpected success fails the test suite. I think trial's `Reporter` treats it as a pass.

In any case, I've swapped these around. I've also added a check to make sure that we only have one kind of result. I _really_ want to be able to trust what I get back from this function.

flocker/testtools/_flaky.py
jml: Actually, not _just_ a style thing. 

```
In [1]: xs = [1, 2, 3]

In [2]: list(x + 1 for x in xs)
Out[2]: [2, 3, 4]

In [3]: print x
NameError                                 Traceback (most recent call last)
<ipython-input-3-2d264e11d975> in <module>()
----> 1 print x

NameError: name 'x' is not defined

In [4]: [x + 1 for x in xs]
Out[4]: [2, 3, 4]

In [5]: print x
3
```

flocker/testtools/_flaky.py
jml: Thanks. I'm going to skip documenting these for now.

flocker/testtools/test/test_flaky.py
jml: More-or-less recommended practice: http://hypothesis.readthedocs.org/en/latest/details.html#how-good-is-assume

(I had a change where I had a strategy for `_FlakyAnnotation` itself, and then used that for populating the decorator, but thought that would diminish the utility of tests as examples)

jml:
> Seeing a bunch of errors like:
> 
> ```
> Failure: testtools.testresult.real._StringException: Empty attachments:
>   twisted-log
>   twisted-log-1
>   twisted-log-2
> ```
> When I run tests that I make fail 3 times in a row. You know what those are about?

This is a feature, not a bug. The `AsynchronousDeferredRunTest` object captures the relevant bits of the Twisted logs and displays them when there's an error or failure. If there are no logs, it indicates that they are empty.

I'll grant that the message could be more informative. If you're keen, could you please file a bug at https://bugs.launchpad.net/testtools/+filebug or at https://github.com/testing-cabal/testtools?

> I also downloaded this branch and played around with it a bit manually. If I were to have 1 complaint it's that tests that ultimately wind up passing or skipping don't log the error / traceback from the times that they failed or errorin out. I don't think that's worth holding this up though, so go ahead and merge whenever!

To me, the tradeoff is that this branch makes debugging flaky tests a little harder, but the benefit is that team members who are trying to do other things will be bothered by flaky tests less.

Fixing the actual problem is possible (of course), but relatively complicated: it involves providing a customized trial reporter and somehow making sure we use that. Or, possibly, fixing trial to support unexpected success in the same way the rest of the world does it.

The best interim solution is to restore logging functionality. See below.

> Also a few thoughts about future work in no particular order:
>
> * We might only want to allow Error's to be retried, hopefully Failure's indicate assertions that are actually based around our software breaking. Maybe that's not that big of a deal though, probably better to have you hacking on fixing flaky tests rather than worrying about little differences like this.

Agreed on both counts. The `@flaky` decorator at box/flaky actually allows users to provide a predicate that determines whether an error is worthy of retrying. 

I'd like to try to get the guts of `_RetryFlaky` incorporated into upstream testtools. At which point, we'd need to focus on things like this.

> * Recently I've notice _trial_tmp/test.log to be somewhat emptier than expected after I run tests, is that related to this? or is there any way that log could get the tracebacks from failed runs?

This is https://clusterhq.atlassian.net/browse/FLOC-3469, tracked upstream at https://bugs.launchpad.net/testtools/+bug/942785 and intimately related https://bugs.launchpad.net/testtools/+bug/1515362.

Fixing this is my next order of business, right after https://clusterhq.atlassian.net/browse/FLOC-3462.

Once we've got the logs fixed, we'll be better able to debug the flaky tests that are now passing. 

Thanks for the reviews! I've really appreciated them.

flocker/testtools/_flaky.py
sarum90: Oh, Thanks I was unaware +1 to not polluting our variable namespace.

sarum90:
>>Seeing a bunch of errors like:
>>
>>Failure: testtools.testresult.real._StringException: Empty attachments:
>>  twisted-log
>>  twisted-log-1
>>  twisted-log-2
>>When I run tests that I make fail 3 times in a row. You know what those are about?
>This is a feature, not a bug. The AsynchronousDeferredRunTest object captures the relevant bits of >the Twisted logs and displays them when there's an error or failure. If there are no logs, it indicates >that they are empty.
>
>I'll grant that the message could be more informative. If you're keen, could you please file a bug at >https://bugs.launchpad.net/testtools/+filebug or at https://github.com/testing-cabal/testtools?

I'm probably not going to do this right now, because I'll have to dig in a bit to write a comprehensive bug report, but I'll keep that in the back of my mind for the next time I see it in some logs. And do leave it for me - I wouldn't mind contributing to testtools, but for now I think there are a few bigger fish to fry during work hours :smile: 

Thanks for going through all the review cycles and putting this together! This will absolutely speed up my development cycle.

PR #2194, opened 2015-11-19
[FLOC 2824] Correct irregular title fonts in Acceptance/Functional test docs

davidoliverSP2:
Fixes 2824

We had some strange header tags being used in Acceptance testing doc - This has been corrected, and the Functional Testing content has been moved to its own page (it previously was embedded in the Acceptance test page, but the use of headers was inconsistent)


sarum90:
Looks better now, and :+1: to Functional tests being split out.

PR #2195, opened 2015-11-19
[FLOC 2128] Add installing Vagrant/VirtualBox step to tutorial

davidoliverSP2:
Fixes 2128

The Using Vagrant to Create Virtual Machines tutorial provides no details about installing the products. This PR adds an installation section with links to Vagrant and VirtualBox docs, but also adds a useful example for OS X users of installing using `brew cask`

### DIFF
--- a/docs/tutorials_examples/tutorial/vagrant-setup.rst
+++ b/docs/tutorials_examples/tutorial/vagrant-setup.rst
@@ -36,6 +36,20 @@ These two IP addresses are used throughout the :ref:`MongoDB tutorial<tutorial-m
    This will also require you to start the node services manually.
    For more information, see :ref:`authentication`.
 
+Installing Vagrant and VirtualBox
+=================================
+
+Follow the instructions specific to your platform that are provided in the `Vagrant`_ and `VirtualBox`_ documentation.
+
+Alternatively, on OS X you can use ``brew-cask`` to install the packages using the `Homebrew`_ package manager.
+For example:
+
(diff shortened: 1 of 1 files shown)


docs/tutorials_examples/tutorial/vagrant-setup.rst
```diff
@@ -36,6 +36,20 @@ These two IP addresses are used throughout the :ref:`MongoDB tutorial<tutorial-m
...
 
+Installing Vagrant and VirtualBox
+=================================
+
+Follow the instructions specific to your platform that are provided in the `Vagrant`_ and `VirtualBox`_ documentation.
+
+Alternatively, on OS X you can use ``brew-cask`` to install the packages using the `Homebrew`_ package manager.
+For example:
+
+.. prompt:: bash $
```
sarum90: :grey_question: The rest of this tutorial uses `alice@mercury` for the prompt. Do we have policy on when docs are supposed to have which prompt? Totally your call which to use here, just thought I'd point out that it was different.

sarum90:
Definitely +1 on the installation section. I'm sort of meh-y on the cask lines (and the virtualbox one didn't work for me, but I think that's because I already have virtualbox installed).

But... they do seem rather convenient, so let's keep them.

Otherwise 1 question, I'll defer to you on the answer and let you address and merge when you get a chance.

davidoliverSP2:
thanks @sarum90 - I should have used `alice@mercury` so I made this amendment!

PR #2198, opened 2015-11-19
[FLOC 2236] Add extra info to pre-tag introduction

davidoliverSP2:
Fixes 2236

The Pre-tag review section of the release docs simply said that not deleting tags pushed to GitHub was a policy. This PR adds an explanation of why we don't do that.


sarum90:
Looks good, Thanks for fixing that up!

PR #2199, opened 2015-11-19
[FLOC-3473] Try to reduce control service load

itamarst:
Instead of a cache that is tied to a particular context, use a LRU cache so delayed updates also benefit from the cache.

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -27,14 +27,13 @@ Eliot contexts are transferred along with AMP commands, allowing tracing
 of logged actions across processes (see
 http://eliot.readthedocs.org/en/0.6.0/threads.html).
 
-:var _caching_encoder: ``CachingEncoder`` used by
-    ``SerializableArgument``, allowing for cached serialization.
+:var _wire_encode_cache: ``LRUCache`` mapping serializable objects to
+    their ``wire_encode`` output.
 """
 
 from datetime import timedelta
 from io import BytesIO
 from itertools import count
-from contextlib import contextmanager
 from twisted.internet.defer import maybeDeferred
 from uuid import UUID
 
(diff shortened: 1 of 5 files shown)


flocker/control/_protocol.py
```diff
@@ -123,46 +124,27 @@ def fromBox(self, name, strings, objects, proto):
         self.another_argument.fromBox(name, strings, objects, proto)
 
 
-class CachingEncoder(object):
-    """
-    Cache results of ``wire_encode`` and re-use them, relying on the fact
-    we're encoding immutable objects.
+# The configuration and state can get pretty big, so don't want too many:
+_wire_encode_cache = LRUCache(50)
```
dwgebler: What's defined as pretty big and too many? I guess we can watch this value over time and modify the cache size if necessary.

flocker/control/_protocol.py
```diff
@@ -123,46 +124,27 @@ def fromBox(self, name, strings, objects, proto):
...
-        self._cache = None
-
-_caching_encoder = CachingEncoder()
+    :param obj: Object to encode.
+    :return: Resulting ``bytes``.
+    """
+    result = _wire_encode_cache.get(obj)
+    if result is None:
+        result = wire_encode(obj)
+        _wire_encode_cache.put(obj, result)
```
dwgebler: As you've noted, this relies on the cached objects' immutability. That makes this look a little bit flaky, but actually we already rely on the type of objects we serialize being immutable in our existing context cache, which I think are mostly ``Deployment`` and maybe one or two other similar models anyway.

dwgebler:
Thanks @itamarst I think this is good to merge when you can get a green build. Cause and solution in the issue make sense, though it'd be interesting to see some metrics on how much of a performance benefit this delivers.

There was a minor merge conflict with ``master`` I have fixed.

dwgebler:
Also you probably have to do something to fix the homebrew recipe builder's failure where it's complaining about not finding `repoze.lru` - I don't know why this happens, it's already a dependency we use elsewhere and specify in ``requirements.txt``.

itamarst:
Weird, it does seem to have installed repoze.lru...

PR #2201, opened 2015-11-19
FLOC-3487] Dont create volumes found to exist.

tomprince:
We have a list of volumes that we think exist. Rather than checking to see if anybody reported that a dataset with a given id exists, just check if there is a volume that exists, before deciding to create a volume.

PR #2202, opened 2015-11-19
[FLOC-3493] Add ATTACHED_ELSEWHERE state.

tomprince:
#2201 is an example of where we will need access to this state.

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -896,7 +896,14 @@ class BlockDeviceDeployerDiscoverStateTests(SynchronousTestCase):
         )
         assert_discovered_state(
             self, self.deployer,
-            expected_discovered_datasets=[],
+            expected_discovered_datasets=[
+                DiscoveredDataset(
+                    state=DatasetStates.ATTACHED_ELSEWHERE,
+                    dataset_id=volume.dataset_id,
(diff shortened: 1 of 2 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -896,7 +896,14 @@ def test_only_remote_device(self):
         )
         assert_discovered_state(
             self, self.deployer,
-            expected_discovered_datasets=[],
+            expected_discovered_datasets=[
```
itamarst: Need to change the test's docstring too.

itamarst:
Seems like you'd want these to go in NonManifestDatasets for backwards compat with the current state of the logic (pre-start of refactor)? Even if deployer is going to stop looking at that, the volume catalog is still going to rely on that for a while until we have better model in control service.

itamarst:
Or, hm. We're going to end up with duelling updates to non-manifest datasets which is going to result in confusing user experience, until the control service data model is updated. How did we handle this before? Or did we have the same issue?

itamarst:
Apparently current state of things is non-manifest changing randomly and some datasets disappearing. Luckily datasets usually don't end up in this edge case since REST API just has destination...

So, merge when green (one minor comment above) and if there's no extra exceptions in acceptance test logs.

PR #2203, opened 2015-11-20
[FLOC-3465] Improve poll length and add flongle timeout.

sarum90:
This PR addresses 2 issues:

- Currently the Flocker plugin for Docker polls the Flocker API way fast (20x a second).
- Currently the Flocker plugin for Docker polls forever waiting for a dataset to be present on its node.

To address the first issue I just increased the poll time rather than doing a fancier solution like exponential backoff. It seems to me that without good data on how long convergence normally takes fancier solutions would be premature.

To address the second issue I implemented a timeout that cancels the polling deferred after two minutes. When it cancels it returns a dict to Docker that has "Err" set to a reasonable error string.

It might be tempting to return a non 200/OK response, but by inspecting the current docker code it seems that they will take precisely the string you pass in Err for 200s: https://github.com/docker/docker/blob/877fe61f7538a36c5fe312221356f13b83abbf64/volume/drivers/proxy.go#L120, they even have a test for this API: https://github.com/docker/docker/blob/877fe61f7538a36c5fe312221356f13b83abbf64/volume/drivers/proxy_test.go#L61 and they did not have explicit tests for anything that returned non-200.

Our tests for the flocker plugin for docker use a loopback networking device which made it slightly difficult to inject a Clock as the reactor for just half of the test. The approach I took was to patch the Klein endpoint execution to inject fake clock advancement after the endpoint was run.

My door is open for ways to actually test this in combination with a running instance of docker. It'd be nice to gain some confidence that this response works as expected.

Manually ran docker with a hacked up version of the code to force a timeout:

```
root@ip-172-31-25-58:~# docker run -it -v cat:/data --volume-driver flocker busybox /bin/sh
Error response from daemon: Cannot start container e8d68324dd5de31f267faa23c9d42736760106c930eb53685a1651757e796318: Timed out waiting for dataset to mount.
```


flocker/dockerplugin/_api.py
```diff
@@ -235,6 +237,24 @@ def got_path(path):
...
+                delayed_timeout.cancel()
+            return passthrough
+        d.addBoth(abort_timeout)
+
+        def handleCancel(failure):
+            failure.trap(CancelledError)
+            return {u"Err": u"Timed out waiting for dataset to mount.",
+                    u"Mountpoint": u""}
+
+        d.addErrback(handleCancel)
```
jml: Is it possible to express some of this with `flocker.common._retry.loop_until` instead? 

At the least, we would need a new generator to be passed `steps` that stopped yielding once an absolute timeout was hit.

flocker/dockerplugin/test/test_api.py
```diff
@@ -8,16 +8,110 @@
...
+    This is an automatic clock, which schedules itself to advance whenever
+    there is a pending call.
+
+    :ivar Clock clock: The underlying clock that gets all calls proxied to it.
+        This clock is advanced with every call to callLater.
+    :ivar IReactorTime reactor: This ``IReactorTime`` provider is used to
+        provide an alternate context to advance the clock from. It is not
+        realistic to advance the clock from the context that calls
+        ``callLater``, so we require a different context to advance this clock
+        from.
```
jml: I feel I must be running low on brain. I'm having trouble understanding this comment and also making suggestions for how it could be better. I guess it's because I don't know what "context to advance the clock from" (and variants) mean.

flocker/dockerplugin/test/test_api.py
```diff
@@ -8,16 +8,110 @@
...
+    :ivar bool _scheduled: boolean indicating if the clock is currently
+        scheduled to advance.
```
jml: What does it mean to be scheduled to advance?

flocker/dockerplugin/test/test_api.py
```diff
@@ -8,16 +8,110 @@
...
+    """
+    def __init__(self):
+        self._scheduled = False
+        pass
```
jml: `pass` redundant

jml:
Thanks!

This looks like it was a thorny thing to figure out. I can't think of anything better than `AutomaticClock`, alas. I feel that there should be something better though. 

The patch was kind of tricky to follow. My guess is that most of that complexity is intrinsic, but even so I'd appreciate another look before landing.

sarum90:
>This looks like it was a thorny thing to figure out. I can't think of anything better than AutomaticClock, alas. I feel that there should be something better though.
 
^^ I think that's an understatement -- I'm going to try using `flocker.common._retry.loop_until` instead. I think it might lend itself to an easier testing strategy.

sarum90:
Also if I do wind up doing `AutomaticClock` I'll try to provide some tests to show why it has its complexity. Specifically, tests that fail if you do the naive implementation.

flocker/dockerplugin/_api.py
sarum90: So I'm slightly partial to using cancellation for the timeout because it works precisely even if the timeout is not a multiple of poll_interval, but I'm thinking that loop_until might make the tricky parts of testing better.

flocker/dockerplugin/test/test_api.py
sarum90: Yea I wasn't proud of this.... if I do have to use this, I'll be adding some tests to demonstrate what I'm talking about here.

flocker/dockerplugin/_api.py
sarum90: So... I changed to code to use it, but I didn't give up any of the testing burden... looks cooler now though.

sarum90:
Refactored the test to just inject clock advancement at a different location. Hopefully this is a bit easier to follow.

flocker/common/_retry.py
```diff
@@ -112,6 +112,32 @@ def loop(result):
...
+
+    :param reactor: The reactor implementation to schedule the timeout.
+    :type reactor: ``IReactorTime``.
+
+    :param deferred: The deferred to cancel at a later point in time.
+    :type deferred: ``Deferred``.
+
+    :param timeout_sec: The number of seconds to wait before the deferred
+                        should time out.
+    :type timeout_sec: ``float``.
```
jml: Nit: I think it's OK to incorporate the type into the param declaration, e.g. `param IReactorTime reactor:`

flocker/common/test/test_retry.py
```diff
@@ -191,6 +193,72 @@ def predicate():
...
+        self._timeout = 1.0
+        self._clock = Clock()
+
+    def _execute_timeout(self):
+        """Execute the timeout."""
+        timeout(self._clock, self._deferred, self._timeout)
+
+    def test_times_out(self):
+        """
+        Verify that a deferred that never fires is timed out at the correct
```
jml: Nit: no need for "Verify that" or "Assert that" in test docstrings. https://jml.io/test-docstrings/

jml:
This looks much better, thanks. Please address nits & merge.

flocker/common/_retry.py
sarum90: Done, Agreed.

(I was trying to mimic the documentation for "loop_until")

flocker/common/test/test_retry.py
sarum90: Fixed throughout.

sarum90:
This is causing new failures for reasons I don't understand. Going to understand the reasons, fix them, kickoff tests, wait for greens, and merge.

sarum90:
Waiting for tests and merging.

PR #2204, opened 2015-11-20
[FLOC-3495] Cleanup discovered dataset invariant

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3495

### DIFF
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -112,8 +112,8 @@ def tagged_union_strategy(type, attr_strategies):
     Create a strategy for building a type with a ``TaggedUnionInvariant``.
 
     :param type: Type to generate a strategy for.
-    :param attr_strategies: Mapping of attributes to stratgeies to
-        dgenerate corresponding attributes.
+    :param attr_strategies: Mapping of attributes to strategies to
+        generate corresponding attributes.
     :type attr_strategies: ``dict`` mapping ``str`` to ``SearchStrategy`s.
     """
     invariant = type.__invariant__
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -7,6 +7,8 @@ Invariants for algebraic data types.
 
 from pyrsistent import PClass, field
 
+from hypothesis.strategies import sampled_from, fixed_dictionaries, just
+
 __all__ = ["TaggedUnionInvariant"]
 
 
@@ -23,6 +25,16 @@ class TaggedUnionInvariant(PClass):
 
     allowed_states = field(set, mandatory=True)
     expected_attributes = field(dict, mandatory=True)
+
+    def _get_attrs(self, tag):
+        return {
+            attribute
+            for attribute, tags in self.expected_attributes.items()
+            if tag in tags
+        }
+
+    def _get_states(self, attr):
+        return self.expected_attributes[attr]
 
     def __call__(self, value):
         for attribute, states in self.expected_attributes.items():
@@ -47,3 +59,20 @@ class TaggedUnionInvariant(PClass):
                                      self.allowed_states)),
             ))
         return (True, "")
+
+
+def tagged_union_strategy(type, attr_strategies):
+    invariant = type.__invariant__
+
+    def build(tag):
+        args = {
+            'state': just(tag),
+        }
+        args.update({
+            attribute: strategy
+            for attribute, strategy in attr_strategies.items()
+            if attribute in invariant._get_attrs(tag)
+        })
+        return fixed_dictionaries(args).map(lambda kwargs: type(**kwargs))
+
+    return sampled_from(type.__invariant__.allowed_states).flatmap(build)
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -19,43 +19,44 @@ class TaggedUnionInvariant(PClass):
     present if and only if the object is in one of the corresponding states.
 
     :param set allowed_states: Set of allowed states.
-    :param dict expected_attributes: Dictionary mapping attribute names to
-        the set of states that attribute should be present in.
+    :param dict attributes: Dictionary mapping states to the set of
+        attributes allowed in that state.
     """
 
+    tag_attribute = field(str, mandatory=True)
     attributes = field(dict, mandatory=True)
 
     def _get_attrs(self, tag):
         return self.attributes[tag]
 
     @property
-    def allowed_states(self):
+    def allowed_tags(self):
         return set(self.attributes.keys())
 
     @property
     def _all_attributes(self):
         return {
             attribute
-            for state, attributes in self.attributes.items()
+            for tag, attributes in self.attributes.items()
             for attribute in attributes
         }
 
     def __call__(self, value):
-        if value.state not in self.allowed_states:
-            return (False, "can only be in states {states}.".format(
-                states=', '.join(map("`{0.name}`".format,
-                                     self.allowed_states)),
+        tag = getattr(value, self.tag_attribute)
+        if tag not in self.allowed_tags:
+            return (False, "can only be in {tag_name}s {tags}.".format(
+                tag_name=self.tag_attribute,
+                tags=', '.join(map("`{0.name}`".format,
+                                   self.allowed_tags)),
             ))
-        print value
-        print self._get_attrs(value.state)
-        for attribute in self._get_attrs(value.state):
+        for attribute in self._get_attrs(tag):
             if not hasattr(value, attribute):
-                return (False, "`{attr}` must be specified in state `{state}`"
-                        .format(attr=attribute, state=value.state.name))
-        for attribute in self._all_attributes - self._get_attrs(value.state):
+                return (False, "`{attr}` must be specified in {tag_name} `{tag}`"
+                        .format(attr=attribute, tag_name=self.tag_attribute, tag=tag.name))
+        for attribute in self._all_attributes - self._get_attrs(tag):
             if hasattr(value, attribute):
-                return (False, "`{attr}` can't be specified in state `{state}`"
-                        .format(attr=attribute, state=value.state.name))
+                return (False, "`{attr}` can't be specified in {tag_name} `{tag}`"
+                        .format(attr=attribute, tag_name=self.tag_attribute, tag=tag.name))
         return (True, "")
 
 
@@ -64,7 +65,7 @@ def tagged_union_strategy(type, attr_strategies):
 
     def build(tag):
         args = {
-            'state': just(tag),
+            invariant.tag_attribute: just(tag),
         }
         args.update({
             attribute: strategy
@@ -73,4 +74,4 @@ def tagged_union_strategy(type, attr_strategies):
         })
         return fixed_dictionaries(args).map(lambda kwargs: type(**kwargs))
 
-    return sampled_from(type.__invariant__.allowed_states).flatmap(build)
+    return sampled_from(type.__invariant__.allowed_tags).flatmap(build)
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -14,13 +14,15 @@ __all__ = ["TaggedUnionInvariant"]
 
 class TaggedUnionInvariant(PClass):
     """
-    An invariant that ensure the given object has a ``state`` attribute
-    in the given states, and that all the other specified attributes are
-    present if and only if the object is in one of the corresponding states.
+    An invariant that ensure the given object has an allowd tag
+    attribute, and that all the other specified attributes are
+    present if and only if the object has the appropriate tag.
 
-    :param set allowed_states: Set of allowed states.
-    :param dict attributes_for_tag: Dictionary mapping states to the
-        set of attributes allowed in that state.
+    Note: Attributes that aren't specified by any tag are ignored.
+
+    :param str tag_attribute: The attribute that contains the tag.
+    :param dict attributes_for_tag: Dictionary mapping tags to the
+        set of attributes allowed by that tag.
     """
 
     tag_attribute = field(str, mandatory=True)
@@ -28,10 +30,16 @@ class TaggedUnionInvariant(PClass):
 
     @property
     def _allowed_tags(self):
+        """
+        The set of all allowed tags.
+        """
         return set(self.attributes_for_tag.keys())
 
     @property
     def _all_attributes(self):
+        """
+        The set of all attributes controlled by the invariant.
+        """
         return {
             attribute
             for tag, attributes in self.attributes_for_tag.items()
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -7,10 +7,10 @@ Invariants for algebraic data types.
 
 from pyrsistent import PClass, field
 
-__all__ = ["StateInvariant"]
+__all__ = ["TaggedUnionInvariant"]
 
 
-class StateInvariant(PClass):
+class TaggedUnionInvariant(PClass):
     """
     An invariant that ensure the given object has a ``state`` attribute
     in the given states, and that all the other specified attributes are
--- a/flocker/common/algebraic.py
+++ b/flocker/common/algebraic.py
@@ -3,6 +3,40 @@
 
 """
 Invariants for algebraic data types.
+
+This currently provides an invariant for defining a tagged union. For
+example, a volume can either be detached, attached (with an
+associated block device) or mounted (with an associated block device
+and mount point). The invariant ensures that all and only the
+appropriate attributes are set in each state.
+
+.. code::
+
+    class VolumeStates(Names):
+       # Not attached
+       DEATTACHED = NamedConstant()
+       # Attached to this node but no filesystem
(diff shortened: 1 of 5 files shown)


flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
+# -*- test-case-name: flocker.common.test.test_algebraic -*-
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
jml: ITYM "ClusterHQ"

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+Invariants for algebraic data types.
+"""
+
+from pyrsistent import PClass, field
+
+__all__ = ["StateInvariant"]
+
+
+class StateInvariant(PClass):
```
jml: It's a bit confusing to use the term "state" (here & elsewhere), because it overlaps with the abstract concept of programming state, and also with states in our state machine, which is where I assume this is derived from. It would be confusing in a different way to use the term "constructor", which is more common in Haskell. 

Looking at https://doc.rust-lang.org/book/enums.html for inspiration, `TaggedUnionInvariant`, with `states` renamed to `tags` and `state` to `tag` might be more clear. 

Alternatively, `EnumInvariant`, either with the "tag" nomenclature, or `variant`.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    An invariant that ensure the given object has a ``state`` attribute
+    in the given states, and that all the other specified attributes are
+    present if and only if the object is in one of the corresponding states.
+
+    :param set allowed_states: Set of allowed states.
+    :param dict expected_attributes: Dictionary mapping attribute names to
+        the set of states that attribute should be present in.
```
jml: I think it would be more clear to invert this dictionary, mapping states to the set of attributes that are allowed, since that's customary for ADTs in most other languages.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+
+    allowed_states = field(set, mandatory=True)
+    expected_attributes = field(dict, mandatory=True)
+
+    def __call__(self, value):
+        for attribute, states in self.expected_attributes.items():
+            if (value.state in states) != hasattr(value, attribute):
```
jml: Please add a comment explaining what this conditional is actually checking.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
...
+                    )
+                else:
+                    message = (
+                        "`{attr}` can only be specified in states {states}"
+                        .format(
+                            attr=attribute,
+                            states=', '.join(map("`{0.name}`".format, states)),
+                        )
+                    )
+                return (False, message)
```
jml: :grey_question: Since you're doing early return elsewhere, I'd probably do early return in both branches of the `value.state in states` conditional.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
+# -*- test-case-name: flocker.common.test.test_algebraic -*-
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
jml: ClusterHQ. The data container people. :wink:

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
+# -*- test-case-name: flocker.common.test.test_algebraic -*-
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
+
+"""
+Invariants for algebraic data types.
+"""
```
jml: A lot of people will have to look up what "algebraic data type" means. It would really help to have an example in the docstring. Please add one.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+ARGS_STRATEGEY = st.fixed_dictionaries({
+    'state': st.sampled_from(AlgebraicType.__invariant__.allowed_states),
+    'one': st.booleans(),
+    'two': st.booleans(),
+})
+
+
+class MakeStateInvariantTests(SynchronousTestCase):
```
jml: Please add a docstring.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @given(
+        args=ARGS_STRATEGEY,
+        extra_attribute=st.sampled_from(
+            AlgebraicType.__invariant__.expected_attributes
+        ),
+    )
+    def test_extra_attributes(self, args, extra_attribute):
```
jml: Please add a docstring.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+        for attribute, states in (
+                AlgebraicType.__invariant__.expected_attributes.items()
+        ):
+            if attribute == extra_attribute:
+                assume(args['state'] not in states)
+                continue
```
jml: `assume` raises an exception, so there's no need for `continue`

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+            if args['state'] not in states:
+                del args[attribute]
```
jml: It took me a while to figure out what this `for` loop was doing. I think it's trying to generate a state with an extra argument that isn't valid in that state.

Rather than doing post-processing in the tests, it's better to have some helpers to generate the right thing. Sketch follows:

```python
states = st.sampled_from(AlgebraicType.__invariant__.allowed_states)
valid_attributes = st.sampled_from(AlgebraicType.__invariant__.expected_attributes)

def args_for_state(state):
    # XXX: This would be simpler if `expected_attributes` were inverted.
    args = {'state': st.just(state)}
    for arg, attrs in AlgebraicType.__invariant__.expected_attributes.items():
        if state in attrs:
            args[arg] = st.booleans()
    return st.fixed_dictionaries(args)

valid_arguments = states.flatmap(args_for_state)

@given(
  args=valid_arguments,
  extra_attribute=valid_attributes,
  extra_arg=st.booleans(),
)
def test_extra_attributes(self, args, extra_attribute, extra_arg):
    assume(extra_attribute not in args)
    args[extra_attribute] = extra_arg
    exc = ...
```

another approach might be to use the new, semi-documented `choices` strategy (http://www.drmaciver.com/2015/11/let-hypothesis-making-your-choices-for-you/)

```python
@given(
  args=valid_arguments,
  extra_arg=st.booleans(),
  choice=st.choices(),
)
def test_extra_attributes(self, args, extra_arg, choice):
    extra_attribute = choice(set(AlgebraicType.__invariant__.allowed_states.keys()) - set(args.keys()))
    args[extra_attribute] = extra_arg
    ...
```

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+            exc.invariant_errors[0],
+        )
+
+    @given(
+        args=ARGS_STRATEGEY,
+        missing_attribute=st.sampled_from(
+            AlgebraicType.__invariant__.expected_attributes
+        ),
+    )
+    def test_missing_attributes(self, args, missing_attribute):
```
jml: Please add a docstring.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+        for attribute, states in (
+            AlgebraicType.__invariant__.expected_attributes.items()
+        ):
+            if attribute == missing_attribute:
+                assume(args['state'] in states)
+                del args[attribute]
+            if args['state'] not in states:
+                del args[attribute]
```
jml: Similar thing about the loop & conditional. Instead, please generate valid arguments to begin with, and then use `st.choices` to pick something from them to delete.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+            States.WITH_ATTRIBUTE,
+            States.WITH_TWO_ATTRIBUTES,
+        },
+        expected_attributes={
+            'one': {States.WITH_ATTRIBUTE, States.WITH_TWO_ATTRIBUTES},
+            'two': {States.WITH_TWO_ATTRIBUTES},
+        },
+    )
+
+ARGS_STRATEGEY = st.fixed_dictionaries({
```
jml: Spelling: "strategy", not "strategey".

Although if you make the changes I suggest below (essentially making a strategy to generate valid arguments), then I think this probably won't be necessary any more.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+    def test_invalid_states(self, state, args):
+        """
+        When constructed with a state that isn't allowed,
+        """
+        args['state'] = state
+        for attribute in (
+            AlgebraicType.__invariant__.expected_attributes.keys()
+        ):
+            del args[attribute]
```
jml: I think this is always going to leave `args` equal to `{"state": state}`, i.e. it will delete all the non-`state` attributes. 

Instead, please either:
 1. pass in an empty dict
 2. generate an arbitrary dictionary of kwargs, and `.map()` it to delete the expected attributes.

Please also add a comment explaining why we're doing this. i.e. that if we pass in invalid arguments, we'll get errors about those rather than about the invalid state.

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,117 @@
...
+        exc = self.assertRaises(InvariantException, AlgebraicType, **args)
+        self.assertIn(
+            'can only be in states',
+            exc.invariant_errors[0],
+        )
```
jml: Please add a test demonstrating that we can construct an `AlgebraicType` with valid arguments.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,49 @@
...
+    in the given states, and that all the other specified attributes are
+    present if and only if the object is in one of the corresponding states.
+
+    :param set allowed_states: Set of allowed states.
+    :param dict expected_attributes: Dictionary mapping attribute names to
+        the set of states that attribute should be present in.
+    """
+
+    allowed_states = field(set, mandatory=True)
+    expected_attributes = field(dict, mandatory=True)
```
jml: Should these be `pset` and `pmap`? Perhaps even with constraints of their own?

flocker/common/algebraic.py
jml: As interim steps, I suggest the following:

* Make `allowed_states` and `expected_attributes` both private
* Add helpers: `get_states_for_attribute(self, attribute)`, `get_attributes_for_state(self, state)` (or "tag" or "iter_" or whatever)
* Write the tests and possibly `__call__` in terms of those helpers
* See what's remaining in terms of access to private access and proceed accordingly

At the very least, I think a helper to get valid attributes for a given state would simplify a good chunk of code.

jml:
Thanks Tom! 

This is really interesting code, and potentially useful elsewhere. TBH, it took me quite a while to reason through the logic and the tests, so I've made a few suggestions of how they could be made clearer—at least to me. Please resubmit for review after you've made the suggested changes. I'm happy to talk further about the review & comments in real-time (although maybe after the weekend).

Do you have any ideas about pattern matching?

tomprince:
I think I have addressed all the review comments.

Having written all the tests, I wonder if they would be clearer if they explicitly tested the cases, instead of using hypothesis (or in addition to using hypothesis). Having a hypothesis strategy for a tagged union is definitely going to be valuable, but using it in the tests of the tagged union code seems like it might be confusing what is being tested between the strategy and the code under test.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+
+    Note: Attributes that aren't specified by any tag are ignored.
+
+    :param str tag_attribute: The attribute that contains the tag.
+    :param dict attributes_for_tag: Dictionary mapping tags to the
+        set of attributes allowed by that tag.
+    """
+
+    tag_attribute = field(str, mandatory=True)
+    attributes_for_tag = field(dict, mandatory=True)
```
jml: :question: Should this be a `pmap`? Or even `pmap_field(object, pset)`?

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+        return set(self.attributes_for_tag.keys())
+
+    @property
+    def _all_attributes(self):
+        """
+        The set of all attributes controlled by the invariant.
+        """
+        return {
+            attribute
+            for tag, attributes in self.attributes_for_tag.items()
```
jml: :question: Could just use `values()` instead of `items()`

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+            for attribute in attributes
+        }
+
+    def __call__(self, value):
```
jml: Please add a docstring.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+                            tag=tag.name))
+        for attribute in self._all_attributes - self.attributes_for_tag[tag]:
+            if hasattr(value, attribute):
+                return (
+                    False,
+                    "`{attr}` can't be specified in {tag_name} `{tag}`"
+                    .format(attr=attribute,
+                            tag_name=self.tag_attribute,
+                            tag=tag.name))
+        return (True, "")
```
jml: :question: I guess if we could get a reliable list of fields that have been set on `value` then we could do this in one loop, iterating over those fields rather than the schema. I don't see a way to do this though.

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+                            tag_name=self.tag_attribute,
+                            tag=tag.name))
+        for attribute in self._all_attributes - self.attributes_for_tag[tag]:
+            if hasattr(value, attribute):
+                return (
+                    False,
+                    "`{attr}` can't be specified in {tag_name} `{tag}`"
+                    .format(attr=attribute,
+                            tag_name=self.tag_attribute,
+                            tag=tag.name))
```
jml: It looks like we don't check for fields that have been set that are not valid for any tag? Is this deliberate?

flocker/common/algebraic.py
```diff
@@ -0,0 +1,133 @@
...
+        return (True, "")
+
+
+def tagged_union_strategy(type, attr_strategies):
+    """
+    Create a strategy for building a type with a ``TaggedUnionInvariant``.
+
+    :param type: Type to generate a strategy for.
+    :param attr_strategies: Mapping of attributes to stratgeies to
+        dgenerate corresponding attributes.
```
jml: Nit: "generate"

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,150 @@
...
+        },
+    )
+
+ALGEBRAIC_TYPE_STRATEGY = tagged_union_strategy(AlgebraicType, {
+    'one': st.booleans(),
+    'two': st.booleans(),
+    'extra': st.booleans(),
+})
+
+ALGEBRAIC_TYPE_ARGUMENTS_STRATGEY = ALGEBRAIC_TYPE_STRATEGY.map(
```
jml: Nit: `STRATEGY`, not `STRATGEY`

flocker/common/test/test_algebraic.py
```diff
@@ -0,0 +1,150 @@
...
+        :param extra_value: A value to provide to the exta attribute
+        """
+        state = args['state']
+        invariant = AlgebraicType.__invariant__
+        extra_attributes = (
+            invariant._all_attributes
+            - invariant.attributes_for_tag[state]
+        )
+        assume(extra_attributes)
+        extra_attribute = choice(sorted(extra_attributes))
```
jml: :question: Any particular reason for `sorted` here?

jml:
Thanks!

Only a few minor things to do. Please address those & merge when the tests pass.

As for testing strategy, it might be a good idea to _add_ a couple of explicit non-hypothesis test cases, if only as extra documentation. I'll leave it up to you as to whether to do this, and to decide on whether a follow-up review is needed. 

There's probably scope for future work here, esp. with factoring out some top-level functions to extract data & metadata about tagged unions: what's the tag attribute? what's the value of the tag attribute? what are valid tags? However, presently I think all that's YAGNI.

flocker/common/algebraic.py
tomprince: Yes. The motivating example, for example, has a bunch of attributes that are required in every state, so I took (and documented) the decision to only care about attributes that should be in some tag (and let pyrsistent's `mandatory` deal with others).

flocker/common/test/test_algebraic.py
tomprince: Yes. `choice` doesn't accept `set`s and since hypothesis keeps examples around, I thought it would be good to have a stable ordering to pick from.

tomprince:
> As for testing strategy, it might be a good idea to add a couple of explicit non-hypothesis test cases, if only as extra documentation.

Revisiting this, I realized that I could add explicit hypothesis examples to test, that has about the same effect.

PR #2205, opened 2015-11-20
[FLOC-3410] Close silent AMP connections

dwgebler:
Fixes FLOC-3410.


flocker/control/_protocol.py
```diff
@@ -355,7 +380,25 @@ def set_node_era(self, era, node_uuid):
...
+    Base AMP protocol that provides a method to forcefully close the
+    transport connection.
+    """
+    def __init__(self, reactor, *args, **kwargs):
+        self._timeout = Timeout(
+            reactor, 2 * PING_INTERVAL.seconds, self.close_connection)
+        AMP.__init__(self, *args, **kwargs)
+        locator = kwargs.pop('locator', None)
+        if locator:
+            locator.timeout_reset = self._timeout.reset
```
jml: `self._timeout` doesn't get used anywhere else, so it doesn't need to be an attribute.

jml:
Sorry, going to have to abort this review. Someone else should feel free to take it up.

flocker/control/test/test_protocol.py
```diff
@@ -493,6 +506,33 @@ def setUp(self):
...
+        """
+        self.protocol.makeConnection(StringTransportWithAbort())
+        advance_some(self.reactor)
+        self.client.callRemote(NoOp)
+        self.assertFalse(self.protocol.transport.aborted)
+        self.reactor.advance(PING_INTERVAL.seconds * 1.9)
+        # This NoOp will reset the timeout.
+        self.client.callRemote(NoOp)
+        self.reactor.advance(PING_INTERVAL.seconds * 1.9)
+        self.assertEqual(self.protocol.transport.aborted, False)
```
itamarst: :question: 

You could do single assert at end, which is better since it centralizes the decision of success vs. failure:

```
initially_aborted = self.protocol.transport.aborted
...
later_aborted = self.protoco..transport.aborted
self.assertEqual(dict(initially=initially_aborted, later=later_aborted),
                           dict(initally=False, later=False))
```

flocker/control/test/test_protocol.py
```diff
@@ -493,6 +506,33 @@ def setUp(self):
...
+    def test_connection_closed_on_no_activity(self):
+        """
+        If no communication has been received for long enough that we expire
+        cluster state, the silent connection is forcefully closed.
+        """
+        self.protocol.makeConnection(StringTransportWithAbort())
+        advance_some(self.reactor)
+        self.client.callRemote(NoOp)
+        self.assertFalse(self.protocol.transport.aborted)
+        advance_rest(self.reactor)
```
itamarst: The relationship between advance_rest and the timeout value is not immediately clear. Maybe add a comment?

flocker/control/test/test_protocol.py
```diff
@@ -942,12 +983,38 @@ class AgentClientTests(SynchronousTestCase):
     """
     def setUp(self):
         self.agent = FakeAgent()
-        self.client = AgentAMP(Clock(), self.agent)
+        self.reactor = Clock()
+        self.client = AgentAMP(self.reactor, self.agent)
```
itamarst: Do the makeConnection here so it doesn't have to be added to all tests.

itamarst:
Thanks! Address, and merge when green.

PR #2206, opened 2015-11-20
[FLOC 3488] Add a note to agent service doc about Docker running

davidoliverSP2:
Fixes 3488

The `flocker-container-agent` service must have Docker running before it is enabled (there is currently a bug on Ubuntu that forces a hang if Docker is not running - see FLOC 3490)

This PR adds a note to the Configuring the Agent Services page so the user is aware that they should have Docker running at this point in configuring Flocker.


jml:
LGTM. Thanks.

PR #2208, opened 2015-11-20
[FLOC-3290] Add CPU time metric to benchmark

jongiddy:
Add CPU time as a metric for benchmarking.

SSH to each node in the cluster, and run a `ps` command to see the elapsed cpu time of named processes (the Flocker services). Perform some operation, and then repeat the `ps` command to see how much the elapsed CPU time has increased for each service.

If you have Flocker cluster available, this allows you to run
```
benchmark/benchmark --control=${FLOCKER_ACCEPTANCE_CONTROL_NODE} --certs=${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} --metric cputime
```
to get output like:
```
{
  "scenario": "no-load", 
  "timestamp": "2015-11-20T15:52:15.126751", 
  "metric": "cputime", 
  "client": {
    "username": "jon", 
    "platform": "Linux-3.13.0-68-generic-x86_64-with-Ubuntu-14.04-trusty", 
    "working_directory": "/home/jon/projects/clusterhq/flocker", 
    "nodename": "ecuador", 
    "flocker_version": "1.7.2+342.g6feb850"
  }, 
  "samples": [
    {
      "value": {
        "104.130.239.240": {
          "flocker-control": 0, 
          "flocker-dataset": 1, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }, 
        "104.130.239.127": {
          "flocker-dataset": 1, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }
      }, 
      "success": true
    }, 
    {
      "value": {
        "104.130.239.240": {
          "flocker-control": 0, 
          "flocker-dataset": 0, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }, 
        "104.130.239.127": {
          "flocker-dataset": 0, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }
      }, 
      "success": true
    }, 
    {
      "value": {
        "104.130.239.240": {
          "flocker-control": 0, 
          "flocker-dataset": 0, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }, 
        "104.130.239.127": {
          "flocker-dataset": 0, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }
      }, 
      "success": true
    }
  ], 
  "control_service": {
    "flocker_version": "1.7.1+175.g031bfed", 
    "host": "104.130.239.240", 
    "port": 4523
  }, 
  "operation": "read-request"
```

Only 1-second resolution, so the results won't be very interesting until we implement a long-wait operation, to put a significant gap between the two readings.

### DIFF
--- a/benchmark/metrics/cputime.py
+++ b/benchmark/metrics/cputime.py
@@ -66,25 +66,25 @@ class _CPUParser(LineOnlyReceiver):
 
 class SSHRunner:
 
-    def __init__(self, reactor, node, user=b'root'):
+    def __init__(self, reactor, user=b'root'):
         self.reactor = reactor
-        self.node = node
         self.user = user
 
-    def run(self, command_args, handle_stdout):
+    def run(self, node, command_args, handle_stdout):
         d = run_ssh(
             self.reactor,
             self.user,
-            self.node.public_address.exploded,
+            node.public_address.exploded,
             command_args,
             handle_stdout=handle_stdout,
         )
         return d
 
 
-def get_node_cpu_times(runner, processes):
+def get_node_cpu_times(runner, node, processes):
     parser = _CPUParser()
     d = runner.run(
+        node,
         _GET_CPUTIME_COMMAND + [b",".join(processes)],
         handle_stdout=parser.lineReceived,
     )
@@ -94,7 +94,7 @@ def get_node_cpu_times(runner, processes):
 
 def _get_cluster_cpu_times(clock, nodes, runner, processes):
     return gather_deferreds(list(
-        get_node_cpu_times(runner(clock, node), processes)
+        get_node_cpu_times(runner, node, processes)
         for node in nodes
     ))
 
@@ -109,14 +109,22 @@ def _compute_change(labels, before, after):
 
 
 @implementer(IMetric)
-class CPUTime(PClass):
+class CPUTime(object):
     """
     Measure the elapsed CPU time during an operation.
     """
-    clock = field(mandatory=True)
-    control_service = field(mandatory=True)
-    runner = field(initial=SSHRunner)
-    processes = field(initial=_FLOCKER_PROCESSES)
+
+    def __init__(
+        self, clock, control_service, runner=None,
+        processes=_FLOCKER_PROCESSES
+    ):
+        self.clock = clock
+        self.control_service = control_service
+        if runner is None:
+            self.runner = SSHRunner(clock)
+        else:
+            self.runner = runner
+        self.processes = processes
 
     def measure(self, f, *a, **kw):
         nodes = []
--- a/benchmark/metrics/cputime.py
+++ b/benchmark/metrics/cputime.py
@@ -1,0 +1,151 @@
+# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
+"""
+CPU time metric for the control service benchmarks.
+"""
+
+from characteristic import Attribute, attributes
+from pyrsistent import PClass, field, pset
+from zope.interface import implementer
+
+from twisted.protocols.basic import LineOnlyReceiver
+
+from flocker.common import gather_deferreds
+from flocker.common.runner import run_ssh
+
+from .._interfaces import IMetric
+
+_FLOCKER_PROCESSES = _JOURNAL_UNITS = pset({
+    u'flocker-control',
+    u'flocker-dataset-agent',
+    u'flocker-container-agent',
+})
+
+
+_GET_CPUTIME_COMMAND = [
+    # Use system ps to collect the information
+    b'ps',
+    # Output the command name (truncated) and the cputime of the process.
+    # `=` provides a header.  Making all the headers blank prevents the header
+    # line from being written.
+    b'-o',
+    b'comm=,cputime=',
+    # Output lines for processes with names matching the following (values
+    # supplied by invoker)
+    b'-C',
+]
+
+
+class _CPUParser(LineOnlyReceiver):
+
+    def __init__(self):
+        self.result = {}
+
+    def lineReceived(self, line):
+        # Lines are like:
+        #
+        # flocker-control 1-00:03:41
+        # flocker-dataset 00:18:14
+        # flocker-contain 01:47:02
+        if not line.strip():
+            # ignore blank lines
+            return
+        try:
+            name, formatted_cputime = line.split()
+            parts = formatted_cputime.split(b'-', 1)
+            # The last item is always HH:MM:SS. Split it and convert to
+            # integers.
+            h, m, s = [int(t) for t in parts.pop().split(b':')]
+            # Anything left is the number of days
+            days = int(parts[0]) if parts else 0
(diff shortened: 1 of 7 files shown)


benchmark/operations/read_request.py
```diff
@@ -2,7 +2,6 @@
 from zope.interface import implementer
 
 from twisted.internet.defer import succeed
-from twisted.web.client import ResponseFailed
```
jongiddy: Functionally unrelated to PR - just a lint cleanup

benchmark/metrics/wallclock.py
```diff
@@ -1,6 +1,6 @@
 # Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
 """
-Metrics for the control service benchmarks.
+Wallclock time metric for the control service benchmarks.
```
jongiddy: Disambiguate this file from the CPU time metric file created for this PR.

benchmark/metrics/cputime.py
```diff
@@ -0,0 +1,216 @@
...
+    """
+    return gather_deferreds(list(
+        get_node_cpu_times(runner, node, processes)
+        for node in nodes
+    ))
+
+
+def compute_change(labels, before, after):
+    """
+    Compute the difference between times retrieved ast different times.
```
zubron: This comment could be clearer:
`Compute the difference between CPU times retrieved at different wallclock times.`

zubron:
I've just run the benchmarking script on a cluster on AWS and it's failing to run as the internal IP being used rather than the external public IP.

```
{  
  "task_uuid":"6610a769-aac6-4c49-a406-651005be129d",
  "command":[  
    "ssh",
    "-C",
    "-q",
    "-o",
    "StrictHostKeyChecking=no",
    "-o",
    "ControlMaster=no",
    "-o",
    "PreferredAuthentications=publickey",
    "-l",
    "root",
    "172.31.41.218",
    "ps -o comm=,cputime= -C flocker-docker-plugin,flocker-control,flocker-dataset-agent,flocker-container-agent"
  ],
  "task_level":[  
    1
  ],
  "action_type":"flocker.common.runner:run",
  "timestamp":1448360542.707636,
  "action_status":"started"
}
```

zubron:
The IP addresses for the cluster are returned from `flocker.apiclient.FlockerClient.list_nodes()` which may return private IP addresses. After running the acceptance tests using AWS, the private to public mapping is available with the `FLOCKER_ACCEPTANCE_HOSTNAME_TO_PUBLIC_ADDRESS` environment variable. We could use this to get the public IP addresses for the nodes in the cluster.

benchmark/metrics/cputime.py
```diff
@@ -0,0 +1,216 @@
...
+    # Output the command name (truncated) and the cputime of the process.
+    # `=` provides a header.  Making all the headers blank prevents the header
+    # line from being written.
+    b'-o',
+    b'comm=,cputime=',
+    # Output lines for processes with names matching the following (values
+    # supplied by invoker)
+    b'-C',
+]
+
```
zubron: The version of `ps` on OS X has different switches compared to the version of `ps` on Linux. As a result, tests which use this command will fail to run on OS X. Given that the tests are designed to run on systems that can run Flocker (CentOS 7, Ubuntu 14.04), should we disable these tests on unsupported systems?

benchmark/metrics/test/test_cputime.py
```diff
@@ -0,0 +1,219 @@
...
+    def test_cpu_time(self):
+        """
+        Fake Flocker cluster gives expected results.
+        """
+        node1 = Node(uuid=uuid4(), public_address=IPAddress('10.0.0.1'))
+        node2 = Node(uuid=uuid4(), public_address=IPAddress('10.0.0.2'))
+        metric = CPUTime(
+            Clock(), FakeFlockerClient([node1, node2]),
+            _LocalRunner(), processes=['init'])
+        d = metric.measure(lambda: 1)
```
zubron: Could you add a comment to explain that the function being passed here is essentially a no-op?

benchmark/metrics/test/test_cputime.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def test_compute_change_new_proc(self):
+        """
+        Process that only appears in ``after`` is ignored.
+        """
+        labels = ['node1']
+        before = [{'foo': 500}]
+        after = [{'foo': 555, 'bar': 5}]
+        result = compute_change(labels, before, after)
+        self.assertEqual(result, {'node1': {'foo': 55}})
```
zubron: Given that we are testing that processes that appear in the after list are ignored, would it be better to test for the absence of `bar` rather than also checking that the value for `foo` has been calculated correctly?

benchmark/metrics/test/test_cputime.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def test_compute_change_lost_proc(self):
+        """
+        Process that only appears in ``before`` is ignored.
+        """
+        labels = ['node1']
+        before = [{'foo': 555, 'bar': 5}]
+        after = [{'foo': 600}]
+        result = compute_change(labels, before, after)
+        self.assertEqual(result, {'node1': {'foo': 45}})
```
zubron: Same as the question for the test above.

zubron:
Thanks @jongiddy! Please address the comments above. If you make changes to skip the tests on OS X, please resubmit and I can run them again.

benchmark/metrics/cputime.py
jongiddy: Fixed to run tests only on Linux. Added FLOC-3522 to test these commands on all supported node platforms, independent of developer's personal platform.

benchmark/metrics/test/test_cputime.py
jongiddy: `assertEqual` will fail if `result` contains a `bar` attribute inside the `node1` value, so its absence is being tested.

benchmark/metrics/test/test_cputime.py
jongiddy: Same answer :smile:

benchmark/metrics/test/test_cputime.py
zubron: That was a poorly worded question :smile:  I meant it might be better for the assertion to only check that `bar` doesn't exist as this test could also fail if `compute_change` results in an incorrect value for `foo`.

jongiddy: OK, I understand now. Yes, I should be testing just that `bar` is missing.

jongiddy:
Review comments addressed. The handling of public addresses is a bit of a hack, requiring the acceptance testing environment variables to be set, but should be better addressed in FLOC-3514, which I plan to work on next.

Please test at least using OS X and AWS cluster.

zubron:
Thanks again @jongiddy! The tests that are using `ps` are now skipped on OS X and it works with an AWS cluster. Please merge when the test pass.

PR #2209, opened 2015-11-20
[FLOC 1877] add the Ubuntu log directory

davidoliverSP2:
Fixes 1877

Ubuntu logs are in the var/log/flocker directory. This PR adds this location to the debugging page.

There was some inconsistent titles, so this was cleaned up also.

FLOC-1936 will add more information about requiring the user to be root in order to view logs, so this PR simply adds the location.

### DIFF
--- a/docs/control/administering/debugging.rst
+++ b/docs/control/administering/debugging.rst
@@ -7,7 +7,7 @@ Debugging
 .. _flocker-logging:
 
 Logging
--------
+=======
 
 Flocker processes use the `Eliot`_ framework for logging.
 Eliot structures logs as a tree of actions, which means given an error you can see what Flocker actions caused the errors by finding the other messages in the tree.
@@ -20,7 +20,7 @@ Logs from the Docker containers can be viewed using `the Docker CLI <https://doc
 Ubuntu
 ^^^^^^
 
-XXX This should be documented, see :issue:`1877`.
+Logs from the Flocker processes running on the nodes can be found in the :file:`var/log/flocker` directory.
 
 CentOS 7
 ^^^^^^^^
@@ -86,7 +86,7 @@ We can then find the full set of actions leading up to this decision, as well as
 .. _flocker-bug-reporting:
 
 Bug Reporting
--------------
+=============
 
 When reporting issues with Flocker please include:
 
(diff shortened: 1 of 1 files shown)


docs/control/administering/debugging.rst
```diff
@@ -20,7 +20,7 @@ Logs from the Docker containers can be viewed using `the Docker CLI <https://doc
 Ubuntu
 ^^^^^^
 
-XXX This should be documented, see :issue:`1877`.
+Logs from the Flocker processes running on the nodes can be found in the :file:`var/log/flocker` directory.
```
sarum90: nit: should be `/var/log/flocker` rather than just `var/log/flocker`

docs/control/administering/debugging.rst
```diff
@@ -20,7 +20,7 @@ Logs from the Docker containers can be viewed using `the Docker CLI <https://doc
...
```
sarum90: So right now these docs are really CentOS based, when a lot of the actions are not specifically CentOS actions below. At the very least this block should have a comment like:

`ubuntu$ cat /var/log/flocker/flocker-dataset-agent.log` is the ubuntu equivalent of `centos$ journalctl --all --output cat -u flocker-dataset-agent`.

A more complete solution would be to put these in environment variables, set them up 1 way for ubuntu, a different way for CentOS, and then have a new section for processing.  Right now that's not perfect though, because we seem to sort of willy-nilly add `--priority=err` to some of the commands below but not all :/ I'll leave it up to you if you want to do a big refactor or just add the one line, but I think we should at least give our ubuntu users something more.

sarum90:
Accepting review for now... if you wind up doing a larger refactor maybe re-submit for review, if you just add a line to the ubuntu docs just submit.

davidoliverSP2:
Thanks @sarum90 

I'll add the information suggested to close out this PR, and raise a new issue to expand the debugging section for Ubuntu users.

sarum90:
Thanks, reword looks good!

davidoliverSP2:
Created https://clusterhq.atlassian.net/browse/FLOC-3506 in response to @sarum90 comments that we're not addressed in this PR

PR #2210, opened 2015-11-20
[FLOC 1936] Root access required for viewing logs

davidoliverSP2:
Fixes 1936

The code blocks in the debugging page specify root access on nodes, but this PR adds a sentence in the introduction so users do not miss this.


sarum90:
Looks good, Thanks!

PR #2211, opened 2015-11-21
[FLOC-3417] Add size opt

wallnerryan:
- Add size opt as a option that the docker flocker plugin accepts
- Split up old test that ignores opts to test `profile` and `size` separately
- Add acceptance test that creates sized volume inside of container and verifies it with lsblk
- add `testtools` to `dockerplugin` to parse size expressions.


flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -118,7 +123,7 @@ def run_python_container(self, cluster, address, docker_arguments, script,
             self.addCleanup(client.remove_container, cid, force=True)
         return cid
 
-    def _create_volume(self, client, name, driver_opts):
+    def _create_volume(self, client, name, driver_opts, cleanup=True):
```
tomprince: I think this may have been for debugging?

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -143,9 +148,83 @@ def _create_volume(self, client, name, driver_opts):
...
+        :param cluster: flocker cluster
+        :param node: node the contianer and volume will be on.
+        """
+        client = get_docker_client(cluster, node.public_address)
+        volume_name = random_name(self)
+        # Because flocker/control/schema/types.yml#L250 > 64MiB
+        # "k" and "t" are too big and too small for test
+        # but we to keep some randomness to this test.
+        size_expressions = list(chain.from_iterable((x, x+"b", x+"ib")
+                                for x in ["g"]))
```
tomprince: :question: I'd really like to suggest `hypothesis` here, except that we probably don't want to run this multiple times. I guess maybe `strategy.example()` would be a reasonable thing to do here?

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -143,9 +148,83 @@ def _create_volume(self, client, name, driver_opts):
...
+
+        def _get_datasets(unused_arg):
+            return cluster.client.list_datasets_configuration()
+        d.addCallback(_get_datasets)
+
+        def _verify_created_volume_is_size(datasets):
+            dataset = next(d for d in datasets
+                           if d.metadata.get(u'name') == volume_name)
+            self.assertEqual(int(dataset.metadata.get(u'maximum_size')),
+                             parse_num(size))
```
tomprince: You shouldn't be parsing this multiple times. In fact, I'd suggest starting from the `int` and converting that to the string passed to docker, rather than the other way around.

flocker/acceptance/scripts/lsblkhttp.py
```diff
@@ -0,0 +1,36 @@
...
+from subprocess import check_output
+
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+
+
+class Handler(BaseHTTPRequestHandler):
+    def do_GET(s):
+        s.send_response(200)
+        s.end_headers()
+        if "is_http" in s.path:
```
tomprince: The name isn't clear what this is for, from the path. I guess it is to determine if the server is ready yet. It would be clearer if the path reflected that.

flocker/acceptance/scripts/lsblkhttp.py
```diff
@@ -0,0 +1,36 @@
...
+            try:
+                mounts = {}
+                for line in check_output(['mount', '-l']).split('\n'):
+                    parts = line.split(' ')
+                    if len(parts) > 2:
+                        mounts[parts[2]] = parts[0]
+                device_path = mounts["/sizedvoldata"]
+
+                command = [b"/bin/lsblk", b"--noheadings", b"--bytes",
+                           b"--output", b"SIZE", device_path.encode("ascii")]
```
tomprince: The code here seems somewhat confused about encodings. Here, it treats the return value of `check_output` as `unicode` (since it encodes it). Below, it treats it as `bytes` (i.e. `str`), since it decodes it.

I would suggest:
- explicitly mark all the strings here as bytes with `b""`
- don't encode or decode (experimenting with python3, where this would matter, `check_output` returns bytes)
- Use `"%d" % (size,)` or the like to convert from an int to a string (on python3, `bytes(5)` returns a sequence of 5 NUL bytes, instead of `b"5"`)

flocker/acceptance/scripts/lsblkhttp.py
```diff
@@ -0,0 +1,36 @@
...
+                command_output = check_output(command).split(b'\n')[0]
```
tomprince: A comment about the format of the output here would be helpful.

flocker/acceptance/scripts/lsblkhttp.py
```diff
@@ -0,0 +1,36 @@
...
+                device_size = str(int(command_output.strip().decode("ascii")))
+            except Exception as e:
+                s.wfile.write(str(e.__class__) + ": " + str(e))
```
tomprince: This should probably have an error response, rather than the `200` sent above.

flocker/dockerplugin/_api.py
```diff
@@ -28,7 +28,7 @@
 from ..control._config import dataset_id_from_name
 from ..apiclient import DatasetAlreadyExists
 from ..node.agents.blockdevice import PROFILE_METADATA_KEY
-
+from .testtools import parse_num
```
tomprince: If you are going to be using this in non-test code, then the implementation should be moved out of `testtools`.

flocker/dockerplugin/_api.py
```diff
@@ -231,9 +231,15 @@ def got_configured(configured):
         if profile:
             metadata[PROFILE_METADATA_KEY] = profile
 
+        SIZE = DEFAULT_SIZE
```
tomprince: `ALL_CAPS` is only for constants, so this should probably be `size` instead. Also, consider writing this like
```python
if ...:
   size = ...
else:
   size = ....
```
instead of
```
size =
if ...:
  size =
```
The former makes it clearer that you are picking between two alternatives for what `size` can be.

flocker/dockerplugin/test/test_api.py
```diff
@@ -64,15 +66,17 @@ def test_unmount(self):
...
-    def test_create_with_opts(self):
+    def test_create_with_profile(self):
         """
         Calling the ``/VolumerDriver.Create`` API with an ``Opts`` value
-        in the request body JSON ignores this parameter and creates
-        a volume with the given name.
+        of "profile=[gold,silver,bronze] in the request body JSON create a
+        volume with a given name with [gold,silver,bronze] profile.
         """
+        default_profiles = ["gold", "silver", "bronze"]
```
tomprince: This would probably be better using `hypothesis`. Something like
```python
@given(profile = sampled_from(["gold", "silver", "bronze"])
def test_...(self, profile):
    ...
```

flocker/dockerplugin/test/test_api.py
```diff
@@ -80,7 +84,35 @@ def test_create_with_opts(self):
...
+    )
+    def test_create_with_size(self):
+        """
+        Calling the ``/VolumerDriver.Create`` API with an ``Opts`` value
+        of "size=<somesize> in the request body JSON create a Volumes
+        with a given name and random size between 1-100G
+        """
+        name = u"testvolume"
+        # Cant be below 64 MiB flocker/control/schema/types.yml#L250
+        random_size = str(randint(1, 100))+"G"
```
tomprince: Again, this should use hypothesis.

flocker/dockerplugin/testtools.py
```diff
@@ -0,0 +1,34 @@
+# Copyright ClusterHQ Ltd.  See LICENSE file for details.
+
+"""
+Testing utilities for ``flocker.dockerplugin`.
+"""
+
+from bitmath import TiB, GiB, MiB, KiB
+
+
+def parse_num(expression):
```
tomprince: This needs tests and better error handling. A number of cases that this will fail unintelligibly on:
- Anything that isn't ascii (e.g. `u"\N{snowman}"`)
- Things with non-digits interspersed (e.g. `u"123GiB456")
- Things with multiple runs of non-digits (e.g. `u"glabrazu456GiB"`)
- Things with unknown units (e.g. `u"123parsec"`).
As a general rule, when parsing things, it is better to be explicit in what you accept and reject everything else, rather than trying to strip away what isn't understood and interpreting the rest.

But, probably a better to try and use an existing library (`bitmath.parse_string`?). Unless there is already a specific format that we are parsing (I don't think this is the case).

flocker/acceptance/scripts/lsblkhttp.py
```diff
@@ -0,0 +1,36 @@
...
+class Handler(BaseHTTPRequestHandler):
+    def do_GET(s):
+        s.send_response(200)
+        s.end_headers()
+        if "is_http" in s.path:
+            s.wfile.write(b"hi")
+        else:
+            try:
+                mounts = {}
+                for line in check_output(['mount', '-l']).split('\n'):
```
tomprince: :question: I was curious if `mount` had a stable output format, so I looked at mount(8), which pointed me to `findmnt`, which looks like it can get all the information we need in one call (presuming it is available).

tomprince:
This looks good, but there are a few things that should be cleaned up. Please address the above comments and resubmit.

flocker/acceptance/endtoend/test_dockerplugin.py
wallnerryan: it was initially, but thought it could be useful in the future. Will remove since nothing would use it. tx!

flocker/acceptance/endtoend/test_dockerplugin.py
wallnerryan: parse_num expects a size expression, which is a str and returns a byte value integer. So, we need to get the integer byte value at least once, maybe we can do this once instead of call parse_num twice, but not sure what you mean by start with int here.

flocker/acceptance/scripts/lsblkhttp.py
wallnerryan: its used by `assert_http_server` hence names the path `is_http` to do the assertion. Other scripts return `ok` for this, and thought it was more vague. I can either make it do the same, `return "ok"` or return the path `is_http` to something like `is_http_server` or somthing. Im indefferent and was just looking to get an easy way to use `assert_http_sever` which accepts a path.

flocker/acceptance/scripts/lsblkhttp.py
wallnerryan: agreed, I think the issue was i took code form another script we had as an example, and found the mount output example and never matched up how to treated encoding. Will make it consistent.

flocker/dockerplugin/_api.py
wallnerryan: tx, normally i would put this in something like `utils` but havent seen that convention, i guess i can add it to `_api`, but seems odd.

flocker/dockerplugin/_api.py
wallnerryan: ack

flocker/dockerplugin/test/test_api.py
wallnerryan: ack

flocker/dockerplugin/test/test_api.py
wallnerryan: ack

flocker/dockerplugin/testtools.py
wallnerryan: the format comes from https://github.com/ClusterHQ/unofficial-flocker-tools/blob/master/design.md which is slightly different than `bitmath.parse_string`, but i can look into if would be applicable.

flocker/acceptance/scripts/lsblkhttp.py
wallnerryan: ack, thanks

flocker/acceptance/endtoend/test_dockerplugin.py
wallnerryan: @tomprince im not familiar with `strategy.example()`. 

On the other hand, if we keep the strategy sample small enough maybe we could run this multiple times, but then again maybe not. Open to other ways.

wallnerryan:
thanks for the review! Added a few comments/question on a few of yours. Mostly makes sense, will work on changes.

flocker/acceptance/endtoend/test_dockerplugin.py
tomprince: I was suggesting starting with a number, which would be used here, and converting that to a size expression to pass to plugin via docker.

That is, instead of generating a string and then parsing it, start with a number and format it as a string.

flocker/dockerplugin/testtools.py
tomprince: Looking at that document, it doesn't appear to suggest a rigid interface that needs to be supported exactly unchanged. 

A further consideration is that it is much easier to compatibility accept more stuff than less stuff. That is, if something was originally an error, than making it valid is a compatible change. But if it was previously valid, making it an error is an incompatible change.

flocker/acceptance/scripts/lsblkhttp.py
tomprince: I guess I wasn't clear with what I meant. Looking at the test, this is query to ensure that the container is running (as far as I can tell), so is seems like the path should reflect that.

Looking at the test a bit closer, though; if you reversed the order of the assertions, (so the the size reported by the container is checked first), then the that assertion could perhaps take the place of the one involving this path.

flocker/acceptance/endtoend/test_dockerplugin.py
wallnerryan: ack

exarkun:
Have you considered writing a simpler test using the in-memory Flocker client implementation rather than an acceptance test?  I think you might find that to be easier and to result in a better test.

wallnerryan:
I have that already here
https://github.com/ClusterHQ/flocker/blob/add-size-opt-FLOC-3417/flocker/dockerplugin/test/test_api.py#L148
i beleive, but was under the impression we needed an acceptance test for
this.

PR #2215, opened 2015-11-23
[FLOC-3504] Benchmark wait operation

jongiddy:
Add an operation to benchmarks to wait for 10 seconds between measurements.

This allows us to take better measurements for idle cluster conditions.  CPU time, for example, has a resolution of 1 second, so we need to run for 10 seconds just to get the nearest 10% of CPU load for each process.

This is probably a precursor to running operations repeatedly for a period of time, in order to ensure measurements are useful in general.

To test with a Flocker cluster, run:
```
$ benchmark/benchmark --control=${FLOCKER_ACCEPTANCE_CONTROL_NODE} --certs=${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} --operation wait-10
```

This uses the wallclock time metric and shows that each sample runs for at least (and very close to) 10 seconds.

To test the CPU time metric:
```bash
$ git merge benchmark-cpu-metric-FLOC-3290
$ benchmark/benchmark --control=${FLOCKER_ACCEPTANCE_CONTROL_NODE} --certs=${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} --operation wait-10 --metric cputime
```

### DIFF
--- a/benchmark/_driver.py
+++ b/benchmark/_driver.py
@@ -148,7 +148,7 @@ def driver(reactor, config, scenario, operation, metric, result, output):
     def run_benchmark(ignored):
         return benchmark(
             scenario(reactor, control_service),
-            operation(control_service=control_service),
+            operation(clock=reactor, control_service=control_service),
             metric(clock=reactor, control_service=control_service),
         )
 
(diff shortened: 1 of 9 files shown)


jongiddy:
@mpgomez The control service is not used for this operation, but the instance creation for all operations needs to have the same signature `(clock, control_service)`. I will add a comment to explicitly state it is unused.  Thanks for the review.

PR #2217, opened 2015-11-23
[FLOC-2820] Sketch of conditional dataset operations.

itamarst:
Probably best to leave updating the `flocker.apiclient` code to a followup branch.

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -26,6 +26,36 @@ For more information, see the :ref:`API authentication guide <generate-api>`.
 
 For more information read the :ref:`cluster architecture<architecture>` documentation.
 
+
+Using the API
+=============
+
+Conditional requests
+--------------------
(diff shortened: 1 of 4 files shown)


exarkun:
Thanks.

To the general idea here, of providing a way to make requests that are conditional on the state being modified, makes sense and sounds good (an I think we've talked about this idea before so no real surprise there).

To the specific implementation approach, seems reasonable and straightforward.  I won't claim to have 100% of the details of the etag specification in my brain right now so I won't claim there isn't anything here that doesn't run afoul of the spec (the main thing I can imagine is that supporting `If-Match` on some endpoints and not supporting any other `If-*` header might confuse some people or somehow not quite conform to the spec; so that might be worth double-checking, and maybe we should actually generate errors if people use `If-Match` or another `If-*` header on endpoints where we're ignoring them... not a hard requirement, just something to consider).

So, yea, thanks and please proceed.

itamarst:
Hm. Instead of erroring I could also... add support to all methods? Or osmething.

PR #2220, opened 2015-11-23
Adjust copyright headers.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3519


sarum90:
Looks like you did this correctly as far as I can tell -- Thanks! nice to clean up low hanging fruit like this.

I think this is a fine thing to do with all of these files? I've certainly seen it done on several in other code reviews. I'm still not 100% I know the ramifications of changing these, so I will accept the review, but I'll leave it for you to merge.

tomprince:
I've checked with the appropriate people, and will merge once I hear back from them.

jongiddy:
Shouldn't the copyright notice include the years in which the file has been modified? https://en.wikipedia.org/wiki/Copyright_notice

tomprince:
> Shouldn't the copyright notice include the years in which the file has been modified? https://en.wikipedia.org/wiki/Copyright_notice

I've asked about this, but am going to merge this as an incremental improvement.

PR #2222, opened 2015-11-23
[FLOC-3517] Sketch of client-side support for conditional dataset operations.

itamarst:
This would require FLOC-2820 (see #2217) to be implemented first. The goal is to make it easy to do conditional operations using the client; see example at the end.

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -8,6 +8,7 @@ from uuid import UUID, uuid4
 from json import dumps
 from datetime import datetime
 from os import environ
+from itertools import repeat
 
 from ipaddr import IPv4Address, IPv6Address, IPAddress
 
@@ -24,6 +25,7 @@ from twisted.internet.defer import succeed, fail
 from twisted.python.filepath import FilePath
 from twisted.web.http import CREATED, OK, CONFLICT, NOT_FOUND
 from twisted.internet.utils import getProcessOutput
+from twisted.internet import reactor
 
 from treq import json_content, content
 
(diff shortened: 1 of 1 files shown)


flocker/apiclient/_client.py
```diff
@@ -140,12 +142,36 @@ class ContainerAlreadyExists(Exception):
...
+class DatasetsConfiguration(PClass):
+    """
+    Currently configured datasets.
+
+    :ivar etag: The current version of the configuration, suitable for
+        passing as conditional to operations like creation.
+    :ivar datasets: The current ``Dataset`` on the server; maps ``UUID``
+        to ``Dataset``.
+    """
+    etag = field(mandatory=True)
```
tomprince: We probably want this to be an opaque type.

flocker/apiclient/_client.py
```diff
@@ -187,7 +219,7 @@ def list_datasets_configuration():
         Return the configured datasets, excluding any datasets that
         have been deleted.
 
-        :return: ``Deferred`` firing with iterable of ``Dataset``.
+        :return: ``Deferred`` firing with a ``DatasetsConfiguration``.
```
tomprince: :question: Do we care about backwards compatibility for the client (since this changes isn't compatible)? Although, `DatasetsConfiguration` could be made iterable to be backwards compatible with the documented interface.

flocker/apiclient/_client.py
```diff
@@ -686,3 +718,44 @@ def got_era(era):
...
+    :param client: ``IFlockerAPIV1Client`` provider.
+    :param condition: Callable which will be called with the current set
+        of ``Dataset`` on the server. If this raises an exception then the
+        create will be aborted.
+    """
+    def create():
+        d = client.list_datasets_configuration()
+
+        def got_config(config):
+            condition(config.datasets)
```
tomprince: :question: `maybe_deferred`?

tomprince:
The RFC allows sending multiple `Etags` in `If-Match`; I wonder if it makes sense to expose that in the API? I guess I can't think of a compelling use case for it.

This looks good, modulo the comments on #2225 regarding the fact that using an Etag of one resource for a different resource isn't supported by the RFC.

I also think we should consider what happens if we later have a more specific etag for the `/configuration/datasets/<dataset-id>` endpoint, will code that is passing the entire configuration etag still work? I think it is probably possible to make it work, but it would be good to think through the details before committing to this API.

flocker/apiclient/_client.py
itamarst: It's a private interface for now, so probably not.

flocker/apiclient/_client.py
itamarst: Can add it later, for now synchronous seems fine.

PR #2224, opened 2015-11-24
[FLOC-3247] Retry the Docker installation step.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3247

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -1306,12 +1306,16 @@ def task_install_docker(distribution):
     else:
         update = b""
 
-    return run(command=(
-        b"[[ -e /usr/bin/docker ]] || { " + update +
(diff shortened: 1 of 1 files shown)


exarkun:
Some tests failed but only after the code this PR changes ran.  However, none of the tests actually triggered the retry logic either so I wouldn't say CI has demonstrated this change is achieving the desired result.

tomprince:
Yes. I was going to say, it would be nice to see this code take effect. However, the code does work in the non-error case, so presumably it could only be incorrect in the failure case. Since that is no worse than currently, and the code change looks correct, I think this is okay to merge once this is green.

exarkun:
Woo.  Your logic is lovely.  Merging.  Thank you.

PR #2225, opened 2015-11-24
[FLOC-2820] Allow conditional updates of the dataset configuration

itamarst:
This will be used to ensure uniqueness of names of datasets created by the Docker plugin.


wallrj:
Forced a build: http://build.clusterhq.com/boxes-flocker?branch=compare-and-set-rest-api-FLOC-2820

flocker/restapi/_infrastructure.py
```diff
@@ -210,6 +222,22 @@ def deco(original):
...
+            # If-Matches support:
+            if request.requestHeaders.hasHeader(b"if-matches"):
+                if get_etag is None:
+                    raise BadRequest(
+                        BAD_REQUEST,
+                        {u"description":
+                         u"If-Matches not supported by this method."})
+                etag = get_etag(self)
+                if_matches = request.requestHeaders.getRawHeaders(
+                    b"if-matches")
```
tomprince: I don't think this is parsing the headers correctly. The [RFC](https://tools.ietf.org/html/rfc7232#section-3.1) describes this header as a comma separated list of strings, with `"` surrounding the bodies.

flocker/control/httpapi.py
```diff
@@ -165,12 +178,21 @@ def get_dataset_configuration(self):
...
             that is configured to exist anywhere on the cluster.
         """
-        return list(datasets_from_deployment(self.persistence_service.get()))
+        # Perhaps exposing etag should be done in @structured
+        # implementation, if we end up doing it in more than one
+        # place. For now this is simpler:
+        etag = get_configuration_etag(self)
+        return EndpointResponse(
+            OK, list(datasets_from_deployment(self.persistence_service.get())),
+            headers={b"etag": etag})
```
tomprince: `Etag` should be a string surrounded by double quotes (see [RFC](https://tools.ietf.org/html/rfc7232#section-2.3)).

flocker/control/httpapi.py
```diff
@@ -280,7 +306,8 @@ def saved(ignored):
         outputSchema={
             '$ref':
             '/v1/endpoints.json#/definitions/configuration_datasets'},
-        schema_store=SCHEMAS
+        schema_store=SCHEMAS,
+        get_etag=get_configuration_etag,
```
tomprince: Using the etag from `/configuration/datasets` on `/configuration/datasets/<dataset_id>` isn't supported by the RFC (but I guess is probably a compatible extension to it?)

flocker/control/httpapi.py
```diff
@@ -339,7 +369,8 @@ def saved(ignored):
...
```
tomprince: Same comment here about the applicability of the RFC.

tomprince:
It occurs to me that etags are supposed to be valid forever. Do we want to include an version in the Etag, to make it easy to change how we calculate them?

itamarst:
Maybe we should just change headers to be `X-Configuration-Version` and `X-If-Configuration-Matches` so wedon't have to worry about etag semantics?

tomprince:
> Maybe we should just change headers to be X-Configuration-Version and X-If-Configuration-Matches so wedon't have to worry about etag semantics?

Maybe. But, I *think* the semantics we are providing (but not currently the syntax) complies with those for Etag, we just provide a stronger guarantee.

itamarst:
We don't need versioning, i think. If we change the format then if-matches will result in 419 error, client will retry, and then we're good.

tomprince:
> We don't need versioning, i think. If we change the format then if-matches will result in 419 error, client will retry, and then we're good.

My concern was clashes in old and new etag causing the opposite behavior.

itamarst:
SHA256 ought to prevent that?

flocker/restapi/_infrastructure.py
```diff
@@ -25,6 +23,10 @@
 from ._logging import LOG_SYSTEM, REQUEST, JSON_REQUEST
 from ._schema import getValidator
 
+__all__ = [
+    "EndpointResponse", "structured", "user_documentation",
+    ]
+
```
wallrj: :question: This is a  private module. Does it need an `__all__` ...at all? We do this in the package ``__init__.py`` file right?

flocker/restapi/_infrastructure.py
```diff
@@ -210,6 +222,22 @@ def deco(original):
         @_logging
         @_serialize(outputValidator)
         def loadAndDispatch(self, request, **routeArguments):
+            # If-Matches support:
+            if request.requestHeaders.hasHeader(b"if-matches"):
```
wallrj: :question: I see that `hasHeader` is case-insensitive, but maybe supply `If-Matches` for consistency. I had to look it up in the docs. Others probably will too: http://twistedmatrix.com/documents/current/api/twisted.web.http_headers.Headers.html#hasHeader

flocker/restapi/test/test_infrastructure.py
```diff
@@ -580,6 +619,53 @@ def test_mixedArgumentsFromRoute(self, logger):
...
+    @validateLogging(_assertRequestLogged(b"/baz/quux", b"POST"))
+    def test_if_matches_no_get_etag(self, logger):
+        """
+        If an ``If-Matches`` header is sent to an endpoint without a
+        ``get_etag`` function then an error is returned.
+        """
+        request = dummyRequest(
+            b"POST", b"/baz/quux",
+            Headers({b"content-type": [b"application/json"],
+                     b"If-matches": [b"456"]}),
```
wallrj: `If-Matches`

wallrj:
Thanks @itamarst

1. I agree with @tomprince that if you're going to use ETag for this we need to return quoted Etags and
2. parse multiple quoted Etags in the request header.
3. Seems to me that we ought to implement a GET `/configuration/datasets/<dataset_id>` endpoint that returns the single dataset and a unique ETag derived from the `/configuration/datasets` ETag ....perhaps "<etag>+<dataset_id>"
4. And we should respond to POST  `/configuration/datasets/<dataset_id>` with an updated ETag in that same format...to avoid an extra GET between successive configuration change requests.
5. I wonder if you could return the latest ETag value in the body of a 412 response...again, to avoid an extra GET request.
6. ...or, as you say, perhaps we should just roll our own custom headers for now.
7. Or why not add support for optional Etag style hashes in the JSON blobs that we accept and return from the API eg

```
curl -d@- http://192.0.2.1:4523/v1/configuration/datasets <<EOF
{
  "primary": "8f93806e-4886-4959-8aa4-6f7a6d20d18f",
  "etag": "1adb68e3db9145ecacd60b5c88c82556019f48ed9359ed6a0ef8a749e743d5e3",
}
EOF
```

```
curl -d@- http://192.0.2.1:4523/v1/configuration/datasets/61eb2b86-6f99-4d71-b19a-c0dc6432e887 <<EOF
{
  "primary": "5f8bc90c-37c7-4810-8461-90c88ec84dd4",
  "etag": "c2f75dbae14e7cffb49e97ce6c7d9d44adf880f55d86f63cd65b0561770b8953",
}
EOF
```

Please answer or address the points above and resubmit.

itamarst:
1. Adding etags to specific dataset configs doesn't seem to be useful at the moment.
2. I don't particularly want to start trying to match those specs exactly.

So seems like something non-etag is the way to go.

Adding some more info to the JSON might work, except it wouldn't be backwards compatible. So it has to be HTTP headers. So I'll just switch to custom HTTP headers.

flocker/control/httpapi.py
```diff
@@ -90,9 +92,44 @@
...
+    """
+    Decorator that compares X-If-Configuration-Matches header to result of
+    ``get_configuration_tag``.
+
+    :param original: Original function.
+    :return: Wrapped function.
+    """
+    @wraps(original)
+    def render_if_matches(self, request, **route_arguments):
+        if request.requestHeaders.hasHeader(b"X-If-Configuration-Matches"):
```
wallrj: :question: Maybe define a constant for this custom header name and use it here and in the tests.

flocker/control/httpapi.py
```diff
@@ -90,9 +92,44 @@
...
+            tag = get_configuration_tag(self)
+            if_matches = request.requestHeaders.getRawHeaders(
+                b"X-If-Configuration-Matches")
+            if tag not in if_matches:
+                request.setResponseCode(PRECONDITION_FAILED)
```
wallrj: :question: Could we add the actual tag here too to save having to make a further request?

flocker/control/httpapi.py
```diff
@@ -90,9 +92,44 @@
...
+                request.responseHeaders.setRawHeaders(
+                    b"content-type", [b"application/json"])
+                return dumps({"description": "Tag doesn't match."})
```
wallrj: :question: Maybe `X-Configuration-Tag doesn't match. Requested: aaa-bbb-ccc, Actual: xxx-yyy-zzz.`

wallrj:
Thanks @itamarst 

I tested it with Curl and it works well.

```
...
> X-If-Configuration-Matches: x5f181bad9f14f2aaf5a2426aa0ca4790514054c61947bdb6ec61e1c0e3f35e75
> Content-Length: 30
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 30 out of 30 bytes
< HTTP/1.1 412 Precondition Failed
< Transfer-Encoding: chunked
< Date: Tue, 01 Dec 2015 01:02:22 GMT
< Content-Type: application/json
< Server: TwistedWeb/15.1.0
< 
* Connection #0 to host 119.9.108.114 left intact
{"description": "Tag doesn't match."}

```

Please answer or address the suggestions above and merge when you get clean builds.

flocker/control/httpapi.py
itamarst: That isn't particularly useful, unfortunately, at least for the current use case: ensuring uniqueness of names. That is, if you're told of a precondition failure that still doesn't tell you if the names conflict, so you have to do another retrieval of config anyway.

PR #2226, opened 2015-11-24
Proposed desing

mpgomez:
draft code ready for design review

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -950,6 +950,39 @@ def eliot_output(message):
     sys.stdout.write(format % message)
     sys.stdout.flush()
 
+def capture_upstart(reactor, host, output_file):
+    """
+    SSH into given machine and capture relevant logs, writing them to
+    output file.
+
+    :param reactor: The reactor.
+    :param bytes host: Machine to SSH into.
+    :param file output_file: File to write to.
+    """
+    # TODO: check the format we need to use here (if we need it at all)
+    output_file.write("[[HOST]]"+host+"[[/HOST]]\n\n")
+    # Using tail instead of cat becase it will also print the name of the file so we
+    # can use it later. Using the whole filename instead of smtg like flocker*
+    # because I can't make it work inside the command. It has a funny way of
+    # parsing the commands. Suggestions are welcome :D
+    formatter = dummy_formatter(output_file)
+    ran = run_ssh(
+        reactor=reactor,
+        host=host,
+        username='root',
+        command=[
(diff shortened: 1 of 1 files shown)


admin/acceptance.py
```diff
@@ -984,6 +1017,16 @@ def capture_journal(reactor, host, output_file):
     ran.addCallback(lambda ignored: formatter(b""))
 
 
+def dummy_formatter(output_file):
+    """
+    Create an output handler which turns journald's export format back into
```
itamarst: There's no journald involved? This would also also need to take host and process name as arguments, since unlike journald it's not available in the message itself.

admin/acceptance.py
```diff
@@ -950,6 +950,39 @@ def eliot_output(message):
...
+def capture_upstart(reactor, host, output_file):
+    """
+    SSH into given machine and capture relevant logs, writing them to
+    output file.
+
+    :param reactor: The reactor.
+    :param bytes host: Machine to SSH into.
+    :param file output_file: File to write to.
+    """
+    # TODO: check the format we need to use here (if we need it at all)
```
itamarst: Printing this line isn't useful; you're writing logs from multiple nodes to same file, so it'll just be interleaved and add no useful info.

admin/acceptance.py
```diff
@@ -950,6 +950,39 @@ def eliot_output(message):
...
+            b'tail',
+            b'/var/log/flocker/flocker-control.log',
+            b'/var/log/flocker/flocker-dataset-agent.log',
+            b'/var/log/flocker/flocker-container-agent.log',
+            b'/var/log/flocker/flocker-docker-plugin.log',
+        ],
+        handle_stdout=formatter,
+    )
+    ran.addErrback(write_failure, logger=None)
+    # Deliver a final empty line to process the last message
```
itamarst: This seems unnecessary?

admin/acceptance.py
```diff
@@ -950,6 +950,39 @@ def eliot_output(message):
...
+    # because I can't make it work inside the command. It has a funny way of
+    # parsing the commands. Suggestions are welcome :D
+    formatter = dummy_formatter(output_file)
+    ran = run_ssh(
+        reactor=reactor,
+        host=host,
+        username='root',
+        command=[
+            b'tail',
+            b'/var/log/flocker/flocker-control.log',
```
itamarst: It'd good to get Docker logs as well.

itamarst:
Please proceed.

PR #2227, opened 2015-11-24
[FLOC-3077] Automatic retry on certain Docker failures

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3077

This adds a retrying layer on top of `docker.Client` and uses that layer in the container agent, the functional tests, and the acceptance tests (most places, at least).  There's a remaining use of `docker.Client` without the retry layer in `flocker/testtools/__init__.py` that I'm not fixing because of a dependency direction problem (probably need to move the docker retry stuff to `flocker/common/`?).

The layer retries two Docker 500 errors which we know are helped by retrying (ie, they go away if you try again).  It's easy to add detection for more cases as we discover them.

This should solve both tracebacks on the linked JIRA issue (one in the description, one in a comment).  It also supersedes retry logic that was added to the `run_registry` helper in `test_docker.py` which added retries for any `APIError` on a particular `DockerClient.list` call.  Since that code should now be using the retrying layer, retries should already be applied to that case (and, happily, many others).

Note this *doesn't* appear to solve the problem I was hoping it would also solve, the acceptance test failures for FLOC-3199 / #2191.

Submitting anyway because hopefully it solves some other problems.

http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/no-such-device-FLOC-3077/job/__main_multijob/


flocker/common/_retry.py
```diff
@@ -244,3 +250,175 @@ def should_retry(e):
...
+            Message.new(
+                message_type=_TRY_SUCCESS,
+                result=result,
+            ).write()
+            saved_result.append(result)
+            return True
+
+    poll_until(pollable, repeat(0.0), sleep=sleep)
+
+    return saved_result[0]
```
jml: It might be best to have a voice call to talk this patch over, or perhaps even pair on it—it's very different from how I was imagining the retry system. I don't understand why the patch introduces a new way to manage the delay intervals.

It would have seemed more straightforward to me to make a synchronous equivalent to `retry_failure` (modifying the API to allow for arbitrary predicates on exceptions rather than type checks), and then to `tee` the steps iterator for each of the wrapped methods. But probably I'm missing something. Am keen to talk.

exarkun:
Local testing produces good results but master has moved on in a way that breaks one of the new test cases.  I think I need to port `failureResultOf` to make it work again.  Apart from that this seems mergeable now.

flocker/acceptance/testtools.py
```diff
@@ -110,8 +110,11 @@ def get_path(name):
         # do verify certificate authority signed the server certificate:
         assert_hostname=False,
         verify=get_path(b"cluster.crt"))
-    return Client(base_url="https://{}:{}".format(address, DOCKER_PORT),
-                  tls=tls, timeout=100, version='1.21')
+
+    return dockerpy_client(
+        base_url="https://{}:{}".format(address, DOCKER_PORT),
+        tls=tls, timeout=100, version='1.21',
```
jml: :question: Can we easily change this `timeout` parameter to take `timedelta`?

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+            steps=[timedelta(seconds=3)] * 5,
+        )
+        obj_with_three_retries = decorate_methods(obj, retry_three_times)
+
+    :param callable decorator: A unary callable that takes a method and returns
+        a method.
+
+    :return: An object like ``obj`` but with all the methods decorated.
+    """
+    return _DecoratedInstance(obj, decorator)
```
jml: :question: This should probably be moved to a different module, since it's not really about retries.

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+    Call a function until it does not raise an exception or ``should_retry``
+    says it shouldn't be tried anymore, whichever comes first.
+
+    :param should_retry: A three-argument callable which determines whether
+        further retries are attempted.  If ``None`` or a ``timedelta`` is
+        returned, another retry is attempted (immediately or after sleeping for
+        the indicated interval, respectively).  If an exception is raised,
+        further tries are not attempted and the exception is allowed to
+        propagate.
+    :param steps: XXX
```
jml: Please document this parameter.

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+    Return a new version of ``method`` that retries.
+
+    :param callable method: A method to retry.
+    :param callable should_retry: A one-argument callable which accepts a
+        three-tuple of exception state and returns a ``timedelta`` or raises
+        an exception. If the call should be retried, the ``timedelta`` gives a
+        time to delay before the retry. If an exception is raised, no further
+        retries are attempted and the exception is propagated from the method
+        call.
+    :param XXX steps: XXX
```
jml: Please document this parameter.

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+    saved_result = [None]
+
+    def pollable():
+        Message.new(
+            message_type=_TRY_RETRYING,
+        ).write()
+        try:
+            result = function(*args, **kwargs)
+        except Exception as e:
+            # XXX: Should this return bool?
```
jml: What is the resolution of this XXX? Please update the comment.

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+    """
+    Return a new version of ``method`` that retries.
+
+    :param callable method: A method to retry.
+    :param callable should_retry: A one-argument callable which accepts a
+        three-tuple of exception state and returns a ``timedelta`` or raises
+        an exception. If the call should be retried, the ``timedelta`` gives a
+        time to delay before the retry. If an exception is raised, no further
+        retries are attempted and the exception is propagated from the method
+        call.
```
jml: The `timedelta` part of the documentation for `should_retry` is no longer valid. Please fix.

flocker/common/_retry.py
```diff
@@ -275,3 +279,178 @@ def should_retry(e):
...
+    """
+    Call a function until it does not raise an exception or ``should_retry``
+    says it shouldn't be tried anymore, whichever comes first.
+
+    :param should_retry: A three-argument callable which determines whether
+        further retries are attempted.  If ``None`` or a ``timedelta`` is
+        returned, another retry is attempted (immediately or after sleeping for
+        the indicated interval, respectively).  If an exception is raised,
+        further tries are not attempted and the exception is allowed to
+        propagate.
```
jml: Please update the documentation of `should_retry` to account for the fact that `timedelta` isn't used.

flocker/common/_retry.py
jml: :question: Maybe it's better to change the interface so that instead of passing `should_retry`, callers would just pass the predicate on `exc_value`.

flocker/common/test/test_retry.py
```diff
@@ -674,3 +689,218 @@ def tester():
...
+        """
+        class Original(object):
+            class_attribute = object()
+
+            def __init__(self):
+                self.instance_attribute = object()
+
+        original = Original()
+        wrapper = decorate_methods(original, self.noop_wrapper)
+        self.assertThat(
```
jml: Probably want `expectThat` here.

flocker/common/test/test_retry.py
```diff
@@ -674,3 +689,218 @@ def tester():
...
+        sleep = time.append
+
+        expected = object()
+        another = object()
+        results = [another, expected]
+
+        wrapper = with_retry(results.pop, sleep=sleep)
+        actual = wrapper()
+
+        self.assertThat(actual, Equals(expected))
```
jml: Probably want `expectThat` here.

flocker/common/test/test_retry.py
```diff
@@ -674,3 +689,218 @@ def tester():
...
+        )
+        # XXX testtools ``raises`` helper generates a crummy message when this
+        # assertion fails
+        self.assertRaises(CustomException, wrapper)
+        self.assertThat(
+            next(counter),
+            # The number of times we demonstrated (above) that retry_some_times
+            # retries - plus one more for the initial call.
+            Equals(EXPECTED_RETRY_SOME_TIMES_RETRIES + 1),
+        )
```
jml: `expectThat`, probably.

flocker/common/test/test_retry.py
```diff
@@ -674,3 +689,218 @@ def tester():
...
+            should_retry=retry_if(
+                lambda exception: (
+                    isinstance(exception, CustomException) and
+                    exception.args[0] < 10
+                ),
+            ),
+            sleep=lambda interval: None,
+        )
+
+        self.assertThat(wrapped, raises(CustomException))
```
jml: `expectThat`

flocker/testtools/_base.py
```diff
@@ -54,7 +54,22 @@ def mktemp(self):
...
+
+    This is provided for compatibility with Twisted's TestCase.  New code
+    should use matchers instead.
+    """
+    successResultOf = unittest.SynchronousTestCase.successResultOf.__func__
+    failureResultOf = unittest.SynchronousTestCase.failureResultOf.__func__
+    assertNoResult = unittest.SynchronousTestCase.assertNoResult.__func__
+
+    # Not related to Deferreds but required by the implementation of the above.
+    assertIdentical = unittest.SynchronousTestCase.assertIdentical.__func__
```
jml: This is a pretty gross hack. I can't think of a productive suggestion though, so no action required.

flocker/testtools/test/test_base.py
```diff
@@ -393,3 +396,40 @@ def test_success(self):
...
             result,
             has_results(tests_run=Equals(1)))
+
+
+class ResultOfAssertionsTests(TestCase):
+    """
+    Bare minimum of testing for the Deferred result-related assertions borrowed
+    from Twisted.
+
+    These tests aren't more thorough because we don't implementat of them in
```
jml: "implement"

jml:
Thanks. 

A few suggestions, please address & merge.

sarum90:
Merging as we actually got a green run of the tests.

Side note, the few tests that are red if you go to the jenkins page for this branch are both:

- No longer run by main_multijob.
- Clearly Jenkins errors, erroring out before they really start.

PR #2231, opened 2015-11-25
[FLOC-3542] Release 1.8.0.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3542

Review Process: http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/release-process.html#pre-tag-review-process

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -13,10 +13,10 @@ You can learn more about where we might be going with future releases by:
 v1.8.0
 ======
 
-* The Docker plugin is now able to use datasets created directly via Flocker so long as the metadata has a matching ``"name"`` value.
-* Better error reporting for the Docker plugin.
-* Added a new REST API for looking up node identity by era; eras are reset after every reboot.
-  This allows interacting with Flocker in a robust way across reboots without getting stale data.
+* The :ref:`Flocker Plugin for Docker<docker-plugin>` is now able to use datasets created directly via Flocker so long as the metadata has a matching ``"name"`` value.
+* Better error reporting for the Flocker Plugin for Docker.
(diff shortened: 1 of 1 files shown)


docs/releasenotes/index.rst
```diff
@@ -10,14 +10,16 @@ You can learn more about where we might be going with future releases by:
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
 
-Next Release
-============
+v1.8.0
+======
 
 * The Docker plugin is now able to use datasets created directly via Flocker so long as the metadata has a matching ``"name"`` value.
```
davidoliverSP2: Could we use "The Flocker plugin for Docker" here instead of "The Docker plugin"?
We could even link directly to the plugin page..... (using `:ref:`docker-plugin``)

docs/releasenotes/index.rst
```diff
@@ -10,14 +10,16 @@ You can learn more about where we might be going with future releases by:
...
 * Better error reporting for the Docker plugin.
 * Added a new REST API for looking up node identity by era; eras are reset after every reboot.
   This allows interacting with Flocker in a robust way across reboots without getting stale data.
```
davidoliverSP2: Small change (optional - I think this just reads a bit easier):
From: This allows interacting with Flocker in a robust way across reboots without getting stale data.
To: This allows robust interaction with Flocker across reboots without getting stale data.

Is it worth specifying what the new REST API is called?

davidoliverSP2:
Couple of small recommended changes (not essential, and not to the new content - this is on the content that was previously under "next release")

Cheers @tomprince

wallrj:
Looks good to me. 
 * Checked the release notes links in the built documentation and they all work.
 * Couldn't see any missing  features since the 1.7.2 release.

Please proceed.

PR #2232, opened 2015-11-26
[FLOC-3530] Log the Cinder volume id in some of the Cinder functional tests

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3530

This will let us compare logs and volumes that got leaked into the cloud and see how often these tests are the culprit.

http://build.clusterhq.com/boxes-flocker?branch=log-cinder-volume-id-FLOC-3530

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -21,6 +21,8 @@ from bitmath import Byte
 import netifaces
 import psutil
 
+from fixtures import Fixture
+
 from keystoneclient.openstack.common.apiclient.exceptions import Unauthorized
 
 from twisted.python.filepath import FilePath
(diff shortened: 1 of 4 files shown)


flocker/node/agents/functional/test_cinder.py
```diff
@@ -396,7 +415,19 @@ def test_get_device_path_no_attached_disks(self):
 
         self.assertEqual(device_path.realpath(), new_device)
 
-    @require_virtio
+
+@require_virtio
```
exarkun: The tests that had this skip were soooooooooooooooo sloooooooooooooooow to discover they should skip it was absurd.  I couldn't stop myself from fixing them so they skip super fast instead.

jml:
LGTM. Thanks!

PR #2237, opened 2015-11-26
[FLOC-3538] sometimes ec2 instance gets public ip address _after_ becoming running

avg-I:
So now we wait until the instance is running and its ip_address attribute
is not None.

https://clusterhq.atlassian.net/browse/FLOC-3538

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -90,8 +90,9 @@ def _check_if_running(instance):
     Message.new(
         message_type=u"flocker:provision:aws:check_if_running:update",
         instance_state=instance.state,
+        ip_address=instance.ip_address,
     ).write()
-    return instance.state == 'running'
+    return instance.state == 'running' and instance.ip_address is not None
 
 
 def _wait_until_running(instance):


exarkun:
Thanks.  Sadly, I couldn't find any AWS EC2 API documentation that confirms this is an expected case.  Waiting a little seems sensible.  Particularly if there's some extra logging added to make it clear what's going on here.

This is going to conflict with some of the other changes you're working on.  Please re-submit with logging and with those conflicts resolved.

avg-I:
Or I could add logging first, merge this change to master and then resolve conflicts in the other two changes?

avg-I:
@exarkun would it enough of logging to report the ip address in the sleep action (along the state)?

exarkun:
> would it enough of logging to report the ip address in the sleep action (along the state)?

Yea that sounds fine.

avg-I:
@exarkun I've updated the code based on the latest master and your suggestions.

flocker/provision/_aws.py
```diff
@@ -90,8 +90,9 @@ def _check_if_running(instance):
     Message.new(
         message_type=u"flocker:provision:aws:check_if_running:update",
         instance_state=instance.state,
+        ip_address=instance.ip_address,
     ).write()
```
itamarst: For future reference, you can do `Message.log(message_type=...)` instead of `Message.new(message_type=...).write()`.

itamarst:
Wow, a green build. Can't argue with that!

flocker/provision/_aws.py
avg-I: Thank you.

PR #2238, opened 2015-11-26
[FLOC-3541] pending aws instance may transition to terminated without running

avg-I:
Previously we would wait until the terminated instance is completely
removed (could take a long time) and then failed with an exception.

Now we retry the creation of an instance.
For additional convenience _wait_until_running() returns True if the
instance has become running, False otherwise.

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -58,6 +58,13 @@ BOTO_INSTANCE_NOT_FOUND = u'InvalidInstanceID.NotFound'
 INSTANCE_TIMEOUT = 300
 
 
+class FailedToRun(Exception):
+    """
+    Raised if a pending AWS failed to become running.
+    See: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
+    """     # noqa
+
+
 def _check_response_error(e, message_type):
     """
     Check if an exception is a transient one.
@@ -74,9 +81,11 @@ def _check_response_error(e, message_type):
     ).write()
 
 
-def _check_if_running(instance):
+def _node_is_booting(instance):
     """
-    Check if an instance is running.
+    Check if an instance is still booting, where booting is defined
+    as either a pending or rebooting instance that is expected to
+    become running.
 
     :param boto.ec2.instance.Instance instance: The instance to check.
     """
@@ -85,14 +94,25 @@ def _check_if_running(instance):
     except EC2ResponseError as e:
         _check_response_error(
             e,
-            u"flocker:provision:aws:check_if_running:retry"
+            u"flocker:provision:aws:node_is_booting:retry"
         )
     Message.new(
-        message_type=u"flocker:provision:aws:check_if_running:update",
+        message_type=u"flocker:provision:aws:node_is_booting:update",
         instance_state=instance.state,
         ip_address=instance.ip_address,
     ).write()
-    return instance.state == 'running' and instance.ip_address is not None
+
+    # Sometimes an instance can be reported as running but without a public
+    # address being set, we consider that instance to be still pending.
+    return (instance.state == u'pending' or instance.state == u'rebooting' or
+            (instance.state == u'running' and instance.ip_address is None))
+
+
+def _poll_while(predicate, steps, sleep=None):
+    """
+    Like common.poll_until, but with the reverse meaning of the predicate.
+    """
+    return poll_until(lambda: not predicate(), steps, sleep)
 
 
 def _wait_until_running(instance):
@@ -104,9 +124,17 @@ def _wait_until_running(instance):
     with start_action(
         action_type=u"flocker:provision:aws:wait_until_running",
         instance_id=instance.id,
-    ):
-        poll_until(lambda: _check_if_running(instance),
-                   repeat(1, INSTANCE_TIMEOUT))
+    ) as context:
+        # Since we are refreshing the instance's state once in a while
+        # we may miss some transitions.  So, here we are waiting until
+        # the node has transitioned out of the original state and then
+        # check if the new state is the one that we expect.
+        _poll_while(lambda: _node_is_booting(instance),
+                    repeat(1, INSTANCE_TIMEOUT))
+        context.add_success_fields(instance_state=instance.state)
+        context.add_success_fields(instance_state_reason=instance.state_reason)
+    if instance.state != u'running':
+        raise FailedToRun(instance.state_reason)
 
 
 @implementer(INode)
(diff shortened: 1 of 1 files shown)


flocker/provision/_aws.py
```diff
@@ -220,33 +222,34 @@ def create_node(self, name, distribution,
...
+                            """),
+                    )
+
+                    instance = reservation.instances[0]
+                    context.add_success_fields(instance_id=instance.id)
+
+                self._connection.create_tags([instance.id], metadata)
+
+                if _wait_until_running(instance):
+                    break
```
exarkun: Probably wouldn't hurt to make the body of this loop a separate function now to help with the readability of `create_node`.

Also, I'd probably put a pretty small retry limit on this.  If something totally weird is going on, I wouldn't like it if this code ends up booting 50 or 100 instances before I realize what's happening and kill it.  Also, worse, buildbot isn't smart and could take up to an hour to get bored.  That could be a lot of instances.

flocker/provision/_aws.py
```diff
@@ -60,14 +60,16 @@ def _wait_until_running(instance):
     with start_action(
         action_type=u"flocker:provision:aws:wait_until_running",
         instance_id=instance.id,
-    ):
-        while instance.state != 'running':
+    ) as context:
+        while instance.state != u'running' and instance.state != u'terminated':
```
exarkun: Once the state is `shutting-down` is there much hope of this succeeding?  It's a little more complex to look for three states instead of two might it might be worthwhile to make progress faster.

Also ... it should perhaps be even more complicated: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html

I guess "stopping" and "stopped" are technically recoverable... The instance might get started again (and so go back to pending and then running).  But this isn't *supposed* to happen as part of this process.  Likewise for "rebooting".

So it might make sense to only allow "pending" and "running".  Anything else means something unexpected has happened to the instance and making another one is likely the safest thing to do (apart from giving up entirely).

Also, EC2 instances come with `state-reason-code` and `state-reason-message` which tell you what Amazon thinks the reason for the instance being in the state it is in.  This might be interesting information to log as well - perhaps it will give a clue as to why things are in the wrong state.  Off the top of my head, I'm not sure how you get this using boto.  I know that at the AWS EC2 API level, it's part of the `DescribeInstances` response, though.

flocker/provision/_aws.py
```diff
@@ -60,14 +60,16 @@ def _wait_until_running(instance):
...
             with start_action(
                 action_type=u"flocker:provision:aws:wait_until_running:sleep",
                 instance_state=instance.state,
             ):
                 sleep(1)
             instance.update()
+        context.add_success_fields(instance_state=instance.state)
+        return instance.state == u'running'
```
exarkun: Consider an exception for this case.  The function has failed to do what it is primarily meant to do (block further execution until the instance is running).  An exception is a good way to encourage the caller to handle that unexpected outcome (whereas return values can be ignored by a careless caller).

exarkun:
Thanks!  Pretty good start, here.  I think there are enough suggested changes that it would be good to have this re-reviewed when the inline points have all been addressed.

flocker/provision/_aws.py
avg-I: I am implementing both of your suggestion. Using a limit of 10 for now.

flocker/provision/_aws.py
avg-I: I agree.  And given that we poll for the state changes the picture is even more complicated for us because we can miss some intermediate transitions. So, for now I decided to wait until an instance is out of `pending` or `rebooting` state (`wait_until_running` is used with `reboot()` as well) and then check if the new state is actually `running`.
Also, I am adding logging of `instance.state_reason` field. I think that it should be either `state-reason-code`  or `state-reason-message`.

flocker/provision/_aws.py
avg-I: That's a good suggestion, I'll do that.

flocker/provision/_aws.py
```diff
@@ -58,6 +58,13 @@
 INSTANCE_TIMEOUT = 300
 
 
+class FailedToRun(Exception):
+    """
+    Raised if a pending AWS failed to become running.
+    See: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
+    """     # noqa
```
exarkun: Blech.  `noqa` really bothers me.  It might be okay if all it did was suppress "line too long" warnings.  Instead, it suppresses any and all warnings. :(

I don't have a good suggestion about what to do instead, though.  It just seems like the cure is worse than the disease here.

flocker/provision/_aws.py
```diff
@@ -104,9 +124,17 @@ def _wait_until_running(instance):
...
+        # Since we are refreshing the instance's state once in a while
+        # we may miss some transitions.  So, here we are waiting until
+        # the node has transitioned out of the original state and then
+        # check if the new state is the one that we expect.
+        _poll_while(lambda: _node_is_booting(instance),
+                    repeat(1, INSTANCE_TIMEOUT))
+        context.add_success_fields(instance_state=instance.state)
+        context.add_success_fields(instance_state_reason=instance.state_reason)
+    if instance.state != u'running':
+        raise FailedToRun(instance.state_reason)
```
exarkun: Cool.  I suggest putting this exception case in the function docstring, too.

exarkun:
Thanks!  Looking great.  Just those couple minor comments, the one about `noqa` isn't even really actionable.

Jenkins looks pretty good.  The two failures seems to be problems with Jenkins itself, not the branch, and they happened on builds that don't seem to have anything to do with the code being changed here.

I wonder if you filed issues for them in JIRA (or found existing ones) already?  If not, that'd be nice.

Otherwise, great.  Please merge. :)

flocker/provision/_aws.py
avg-I: Yeah, but I do not want to break the lint build.
Perhaps I could remove the URL, it does not seem to be that informative to me now (in the context of the exception).

avg-I:
Thank you for the review!

> I  wonder if you filed issues for them in JIRA (or found existing ones) already?

There were existing issues. Also, it seems that "main multijob" build no longer triggers those builds.
Not sure if that's intended or by accident.

flocker/provision/_aws.py
avg-I: I've done just that.

flocker/provision/_aws.py
avg-I: Done.

PR #2239, opened 2015-11-26
aws may transiently refuse to recognize a new pending instance

avg-I:
We are making two calls right after the instance creation,
`boto.ec2.instance.update()` and `boto.ec2.connection.create_tags()`,
so these could be affected by the problem.

Now, if we get `InvalidInstanceID.NotFound` error we simply retry the
affected calls.

https://clusterhq.atlassian.net/browse/FLOC-3537

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -4,8 +4,9 @@
 AWS provisioner.
 """
 
+from itertools import repeat
 from textwrap import dedent
-from time import time, sleep
+from time import time
 
 from pyrsistent import PClass, field
 
@@ -19,6 +20,8 @@ from boto.ec2.blockdevicemapping import (
     EBSBlockDeviceType, BlockDeviceMapping,
 )
 from boto.exception import EC2ResponseError
+
+from ..common import poll_until
 
 from ._common import INode, IProvisioner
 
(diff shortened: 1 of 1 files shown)


flocker/provision/_aws.py
```diff
@@ -67,7 +79,13 @@ def _wait_until_running(instance):
...
             ):
                 sleep(1)
-            instance.update()
+            try:
+                instance.update()
+            except EC2ResponseError as e:
+                _check_response_error(
+                    e,
+                    u"flocker:provision:aws:wait_until_running:retry"
+                )
```
exarkun: Likewise - `flocker.common.poll_until` with a timeout might be a good thing to use here.

exarkun:
Thanks.  Please give the retrying helpers a try and then merge.

flocker/provision/_aws.py
avg-I: JP, you said likewise, but I am not sure likewise to what :-)
In any case, I''ll try `flocker.common.poll_until`

exarkun: Uh. :/  Looks like github lost my other comment.  I'll try to re-create it.

flocker/provision/_aws.py
```diff
@@ -242,10 +260,17 @@ def create_node(self, name, distribution,
...
+            while True:
+                try:
+                    self._connection.create_tags([instance.id], metadata)
+                    break
+                except EC2ResponseError as e:
+                    _check_response_error(
+                        e,
+                        u"flocker:provision:aws:create_node:retry"
+                    )
+                    sleep(1)
```
exarkun: I guess it was about using `poll_until` or a similar helper here.

avg-I:
@exarkun I've implemented your suggestion about `poll_until`. The code changed in a slightly non-trivial way, so could you please take another look just in case?
Thanks!

PR #2240, opened 2015-11-26
[FLOC 3545] Add missing content from Functional Test doc

davidoliverSP2:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3545

The instructions for setting up functional tests was missing key environment variables, and then the instructions for each backend were thin.

Working with @jongiddy, we've added the extra environment variables, and provided the commands necessary to run the tests on each backend.

### DIFF
--- a/docs/gettinginvolved/functional-testing.rst
+++ b/docs/gettinginvolved/functional-testing.rst
@@ -10,15 +10,17 @@ The tests look for the following environment variables:
      # FLOC-2090 This is yet another configuration file.
      # Make it just be the same as the acceptance testing configuration file.
 
-- ``FLOCKER_FUNCTIONAL_TEST``: This variable must be set to ``TRUE``.
+- ``FLOCKER_FUNCTIONAL_TEST``: This variable must be set.
 - ``FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE``: This variable points at a YAML file with the credentials.
 - ``FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER``: This variable must be the name of a top-level key in the configuration file.
+- ``FLOCKER_FUNCTIONAL_TEST_AWS_AVAILABILITY_ZONE`` (AWS only): The AWS backend also requires that the availability zone that the test is running in to be specified.
+  This is specified separately from the credential file, so that the file can be reused in different regions.
 
-The credentials are read from the stanza specified by the ``CLOUD_PROVIDER`` environment variable.
+The credentials are read from the stanza specified by the ``FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER`` environment variable.
 The supported block-device backend is specified by a ``provider`` key in the stanza, 
 or the name of the stanza, if the ``provider`` key is missing.
 
-If the environment variables aren't present, the tests will be skipped.
+If the environment variables are not present, the tests will be skipped.
 The tests that do not correspond to the configured provider will also be skipped.
 
 AWS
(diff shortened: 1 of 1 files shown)


docs/gettinginvolved/functional-testing.rst
```diff
@@ -4,19 +4,23 @@ Functional Testing
...
+.. XXX
      # FLOC-2090 This is yet another configuration file.
      # Make it just be the same as the acceptance testing configuration file.
 
-- ``FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE``: This points at a YAML file with the credentials.
-- ``FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER``: This is the name of a top-level key in the configuration file.
+- ``FLOCKER_FUNCTIONAL_TEST``: This variable must be set.
+- ``FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE``: This variable points at a YAML file with the credentials.
+- ``FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER``: This variable must be the name of a top-level key in the configuration file.
+- ``FLOCKER_FUNCTIONAL_TEST_AWS_AVAILABILITY_ZONE`` (AWS only): The AWS backend also requires that the availability zone that the test is running in to be specified.
```
exarkun: There's a similar variable for OpenStack, `FLOCKER_FUNCTIONAL_TEST_OPENSTACK_REGION` I think.

docs/gettinginvolved/functional-testing.rst
```diff
@@ -45,6 +52,12 @@ The configuration stanza for the OpenStack backend running on Rackspace looks as
      username: <rackspace username>
      key: <access key>
 
+To run the functional tests, run the following command:
```
exarkun: I think the example configuration above is wrong as well.

exarkun:
Thanks.  Changes look good but maybe a few more are needed as well.

More generally, it would be really nice if we had some test coverage for these instructions.  Without that, I suspect they're going to be constantly ruined by changes to the implementation and we'll always be playing catch up.

Any thoughts on that?

docs/gettinginvolved/functional-testing.rst
davidoliverSP2: presumably its just removing the region field, as its set in the ``FLOCKER_FUNCTIONAL_TEST_OPENSTACK_REGION`` variable?

davidoliverSP2:
Thanks @exarkun 

I agree that test coverage for this would be great. I've created FLOC 3555 to track this (although wasn't sure on labels/epics values for test related issues)

wallrj:
Thanks @davidoliverSP2 

I set up nodes on AWS and Rackspace and tried the latest instructions.

There were still things missing.

I've pushed a change (https://github.com/ClusterHQ/flocker/commit/e57a9e7f82cd72b0fe7d6c8129e6b80dde7e5071) with the configurations that I used.

The private Open Stack instructions also look out of date. 
I'd be tempted to remove those altogether.
I think working instructions will be almost identical to the Rackspace instructions except with `auth_plugin: password` or something.

Running the functional tests is way too complicated and there's https://clusterhq.atlassian.net/browse/FLOC-2090 for fixing that.

Please merge when you've got new green documentation build and when you've decided whether to remove the private openstack instructions.

davidoliverSP2:
Thanks @wallrj 

I removed the OpenStack example, and just left a bit of content which explains the instructions are the same as Rackspace, but to include the auth_plugin field.

I'll merge on a green build.

PR #2241, opened 2015-11-26
[FLOC 3548] Make it easier for users to reach the Labs Installer docs

davidoliverSP2:
Fixes 3548

@Stephenitis raised an issue that it was at least 3 clicks from the main docs before a user would find the Labs Installer docs. This PR reduces that overhead, and streamlines the Installation docs so the user can quickly decide to either use the Installer, or complete the full installation process.


exarkun:
Can confirm it's now two clicks.  Please merge.  Thanks.

PR #2242, opened 2015-11-26
[FLOC-1831] Change cinder backend to retrieve expected ID only

dwgebler:
Fixes FLOC-1831. Note this change is sufficiently tested by a lack of regression in tests in `flocker.node.agents.functional.test_cinder` that utilise `wait_for_volume_state`.

### DIFF
--- a/flocker/node/agents/cinder.py
+++ b/flocker/node/agents/cinder.py
@@ -22,6 +22,7 @@ from keystoneclient.auth import get_plugin_class
 from keystoneclient.session import Session
 from keystoneclient_rackspace.v2_0 import RackspaceAuth
 from cinderclient.client import Client as CinderClient
+from cinderclient.exceptions import NotFound as CinderClientNotFound
 from novaclient.client import Client as NovaClient
 from novaclient.exceptions import NotFound as NovaNotFound
 from novaclient.exceptions import ClientException as NovaClientException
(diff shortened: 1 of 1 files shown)


flocker/node/agents/cinder.py
```diff
@@ -311,28 +312,25 @@ def reached_desired_state(self):
         Raise an exception if a non-valid state is reached or if the
         desired state is not reached within the supplied time limit.
         """
-        # Could be more efficient.  FLOC-1831
-        for listed_volume in self.volume_manager.list():
-            if listed_volume.id == self.expected_volume.id:
-                # Could miss the expected status because race conditions.
-                # FLOC-1832
```
exarkun: It would be nice not to lose this comment.  As far as I can tell, it's still relevant.

flocker/node/agents/cinder.py
```diff
@@ -311,28 +312,25 @@ def reached_desired_state(self):
...
+            elif current_state in self.transient_states:
+                # Once an intermediate state is reached, the prior
+                # states become invalid.
+                idx = self.transient_states.index(current_state)
+                if idx > 0:
+                    self.transient_states = self.transient_states[idx:]
+            else:
+                raise UnexpectedStateException(
+                    self.expected_volume, self.desired_state, current_state)
+        except CinderClientNotFound:
```
exarkun: This exception handling is just for the `get` call so it should only be applied to the `get` call.  Generally, try to keep the `try` block as short as possible.

exarkun:
Thanks.

> Note this change is sufficiently tested by a lack of regression in tests in `flocker.node.agents.functional.test_cinder` that utilise `wait_for_volume_state`.

For correctness, could be.  This is basically an optimization so hopefully it shouldn't need new unit tests.  Ideally there would be some kind of automated tests demonstrating how it improves system performance - otherwise we don't really know that it has, and even if it has, we don't know if we regress the improvement via a future change.

This might be something the benchmarking work will be able to help with but I don't know if that's yet in a state you can use it for this.

Anyway, long way of saying that in an ideal world, performance improvements are not exempt from automated testing and hopefully soon we'll be at least a little closer to that.

Please address other inline comments and merge.

PR #2243, opened 2015-11-26
[FLOC 3546] Explain why the plugin install instructions are optional

davidoliverSP2:
Fixes 3546

Explains why the plugin is an optional install step in the node services install doc, linking back to the introduction document on the plugin.

### DIFF
--- a/docs/install/install-node.rst
+++ b/docs/install/install-node.rst
@@ -6,6 +6,8 @@ Installing the Flocker Node Services
 
 The following instructions describe how to install the ``clusterhq-flocker node`` package, and the optional ``clusterhq-flocker-docker-plugin`` package on each of the nodes in your cluster.
 
+Installing the :ref:`Flocker plugin for Docker <docker-plugin>` at the same time as the other node services is not essential, but it will save you from revisiting each of the nodes in your cluster again if you were to install it at a later time.
+
 .. _installing-flocker-node-prereq:
 
 Prerequisites
--- a/docs/install/install-node.rst
+++ b/docs/install/install-node.rst
@@ -4,9 +4,10 @@
 Installing the Flocker Node Services
 ====================================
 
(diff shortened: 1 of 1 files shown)


docs/install/install-node.rst
```diff
@@ -6,6 +6,8 @@ Installing the Flocker Node Services
 
 The following instructions describe how to install the ``clusterhq-flocker node`` package, and the optional ``clusterhq-flocker-docker-plugin`` package on each of the nodes in your cluster.
 
+Installing the :ref:`Flocker plugin for Docker <docker-plugin>` at the same time as the other node services is not essential, but it will save you from revisiting each of the nodes in your cluster again if you were to install it at a later time.
```
exarkun: Why isn't it essential?

I think this addition is trivially true.  I'm not sure it is useful, though.  Why would I choose to install the plugin now, later, or never at all?  Does it have to do with what I plan to do with Flocker?  Or maybe the choice is based on my OS or Docker version or license constraints ...  This doesn't really tell me.

exarkun:
Thanks.  I'm not sure this quite does the job as-is.  What do you think?

davidoliverSP2:
Thanks @exarkun 

I agree - I wasn't totally happy with using words like optional or essential. I've amended the text to highlight the scenario where users would want to install the plugin without having to say that anything is optional or essential - hopefully this is a bit clearer.

wallrj:
How about top-level, self contained documentation for Flocker Docker Plugin?

* The Flocker Docker Plugin
 * Introduction
 * Installation
 * Configuration
 * Usage examples

I know I'm not supposed to say stuff like that in review so put this back up for review if you disagree.

Sorry to be a pain but it seems to me that this still doesn't address @Stephenitis 's original issue and the plugin documentation does seem...difficult to navigate.

davidoliverSP2:
Thanks @wallrj 

I think it depends on whether we see the plugin as a part of Flocker, or a separate product. I'll get a second opinion.

davidoliverSP2:
I've closed this PR, and put FLOC 3546 back on the backlog. I believe it will no longer be required, or will be covered by FLOC 3718

PR #2244, opened 2015-11-26
[FLOC-3540] The vervet is leaving 

mpgomez:
Had to build a new docker image in https://hub.docker.com/u/clusterhqci/
It looks like this (it is not checked in anywhere)
{code}
FROM alanfranz/fwd-ubuntu-wily:latest
MAINTAINER ClusterHQ <contact@clusterhq.com>
RUN apt-get update
RUN apt-get install --no-install-recommends -y \
git ruby-dev libffi-dev libssl-dev build-essential python-pip \
python-netaddr \
wget curl python2.7-dev lintian
RUN curl https://raw.githubusercontent.com/ClusterHQ/flocker/master/requirements.txt  > /tmp/requirements.txt
RUN pip install -r /tmp/requirements.txt
{code}

I also found a few issues with lintian:
https://lintian.debian.org/tags/package-installs-ieee-data.html
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=785662

I added them to the errors we ignore, with a couple of extra ones that I am not that sure we should be ignoring.

The ubuntu stuff is running in the staging buildbot :
http://build.staging.clusterhq.com/builders/flocker-omnibus-ubuntu-15.10

In order for those changes to work, we need to merge the changes in buildbot too.

We also need a new issue to make Jenkins work with 15.10


admin/packaging.py
```diff
@@ -662,6 +662,16 @@ def __str__(self):
 
         # Cryptography hazmat bindings
         'package-installs-python-pycache-dir opt/flocker/lib/python2.7/site-packages/cryptography/hazmat/bindings/__pycache__/',  # noqa
+
```
mpgomez: I had to add those errors to the ignored errors list. 
I know that the install-ieee-data belongs here, but I am not that sure about
+        "package-contains-timestamped-gzip",
+        "systemd-service-file-outside-lib",
The first one seems like a reasonable things to add to the ignore list, but I know they were adding systemd in this new 15.10 release, so we may need some extra refactoring to use systemd instead of service?

admin/build_targets/ubuntu-15.10/Dockerfile
```diff
@@ -0,0 +1,12 @@
+# Copyright Hybrid Logic Ltd. See LICENSE file for details.
```
mpgomez: PLEASE NOTE: 
clusterhqci/fpm-ubuntu-wily is an image I have created manually. It looks ok, but I haven't checked it in anywhere.
This manually-built docker image is temporary, as we need to update Jenkins so it builds it periodically the same way it does with Ubuntu14.04 and CentOS (and Ubuntu15.04, but we need to delete that bit as it is not useful any more)

jml: Please update to say "ClusterHQ Ltd", not "Hybrid Logic Ltd"

admin/packaging.py
jml: The first does seem OK.

The second is a bit more complicated, and I don't know the full answer. I'm pretty sure that in a lot of cases we assume that the ubuntu distro implies "not systemd". I think the best way to find out is by building this test image and (somehow!) running the relevant tests against it. I don't know the exact details of how to do this, but will offer what help I can. 

I think this is worth doing soon, even if we don't make any changes to actual systems until next week.

build.yaml
```diff
@@ -573,6 +573,21 @@ common_cli:
...
+    # https://github.com/alanfranz/fpm-within-docker
+    echo "FROM alanfranz/fwd-ubuntu-wily:latest" > Dockerfile
+    echo "MAINTAINER ClusterHQ <contact@clusterhq.com>" >> Dockerfile
+    echo "RUN apt-get update" >> Dockerfile
+    echo "RUN apt-get install --no-install-recommends -y \
+          git ruby-dev libffi-dev libssl-dev build-essential python-pip \
+          python2.7-dev lintian" >> Dockerfile
+    echo "COPY requirements.txt /tmp/" >> Dockerfile
+    echo "RUN pip install -r /tmp/requirements.txt" >> Dockerfile
+
```
jml: No change required, but worth noting. 

The `wget`, `apt-get` and `pip` steps are all sources of intermittent build failure, as they assume that the network is reliable.

Since we're using them everywhere else in this file, there's no need to fix it here.

docs/releasenotes/index.rst
```diff
@@ -9,6 +9,7 @@ You can learn more about where we might be going with future releases by:
 
 * Stopping by the ``#clusterhq`` channel on ``irc.freenode.net``.
 * Visiting our GitHub repository at https://github.com/ClusterHQ/flocker.
+* Now supporting Ubuntu 15.10 insteand of Ubuntu 15.04 for the flocker client.
```
jml: s/insteand/instead/

jml:
Thanks Pilar! Only a couple of small suggestions re code itself. However, I think we need to have more testing to check that we actually work with Wily, which should _probably_ take place before this is merged.

admin/packaging.py
mpgomez: We are running them in Jenkins (full run) and BuildBot (only new ubuntu wily stuff):
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/the-vervet-is-leaving-FLOC-3540/
http://build.staging.clusterhq.com/boxes-flocker?branch=the-vervet-is-leaving-FLOC-3540

jml:
Looking at http://build.staging.clusterhq.com/boxes-flocker?branch=the-vervet-is-leaving-FLOC-3540 (which is an old build), it looks like there are quite a few test failures. I don't know how many of them have been fixed by changes since f95e25e.

Can you please fix the failures and run the tests again on the latest code, giving a link to the results here in Github? Once we've got green builds on buildbot staging and Jenkins, I'd like to do another quick review before merging.

Thanks!

jml:
We talked about this on Slack, and I think the conclusion was that we'll wait to merge this until Monday, after the transition to Jenkins. At that point, we'll not worry about getting buildbot working. 

If it turns out we need to roll back from Jenkins to Buildbot (unlikely, because we're committed to fixing what errors we find), then we can fix 15.10 on buildbot.

PR #2245, opened 2015-11-26
[FLOC-3336] Update commit status on GitHub with Jenkins build result

zubron:
This change adds support for updating a commit status with the build result from Jenkins.

After being unable to provide this support using Jenkins plugins, I have chosen to update the commit status using build steps within the configuration for the __main_multijob project. When the __main_multijob project on Jenkins is triggered for a branch, the commit that is being tested will be updated with the status that the build has started. Two conditional post-build steps have been added to update the commit status with the build result of the __main_multijob. This means the status will not be updated until all the builds within this project have been completed.

To update the status on GitHub, we need to obtain the correct SHA for the commit. This is usually available as an environment variable within Jenkins, however, as we are performing a merge with master before the build, this environment variable is set to the SHA of the merge commit instead. To get the correct SHA, we get the latest commit for the branch using `git rev-list`. We then use the [GitHub Status API](https://developer.github.com/v3/repos/statuses/) to update the status for that commit.

### DIFF
--- a/jobs.groovy.j2
+++ b/jobs.groovy.j2
@@ -92,26 +92,26 @@
 
 
 {% macro updateGitHubStatus(status)                                         -%}
-  shell("""
-REMOTE_GIT_COMMIT=`git rev-list --no-merges --max-count=1 HEAD`
-echo \$REMOTE_GIT_COMMIT > gitcommit
-  """)
-
 {# Read the commit and github credentials                                    #}
 {# Update github with the build status                                       #}
-  if ("{{status}}" == "pending" ) {
+{% if status == "pending"                                                    %}
     shell("""
-      REMOTE_GIT_COMMIT=\$(cat gitcommit)
-      echo \$REMOTE_GIT_COMMIT
+REMOTE_GIT_COMMIT=`git rev-list --no-merges --max-count=1 HEAD`
+echo \$REMOTE_GIT_COMMIT
+echo 'Job started'
     """)
-
-  } else {
+{% endif %}
+{% if status == "complete" %}
     conditionalSteps {
       condition {
         status('FAILURE', 'FAILURE')
       }
       steps {
-        shell("echo 'Job failed'")
+        shell("""
+REMOTE_GIT_COMMIT=`git rev-list --no-merges --max-count=1 HEAD`
+echo \$REMOTE_GIT_COMMIT
+echo 'Job failed'
+        """)
       }
     }
     conditionalSteps {
@@ -119,10 +119,14 @@ echo \$REMOTE_GIT_COMMIT > gitcommit
         status('SUCCESS', 'SUCCESS')
       }
       steps {
-        shell("echo 'Job succeeded'")
+        shell("""
+REMOTE_GIT_COMMIT=`git rev-list --no-merges --max-count=1 HEAD`
+echo \$REMOTE_GIT_COMMIT
+echo 'Job succeeded'
+        """)
       }
     }
-  }
+{% endif %}
 {%- endmacro                                                                -%}
 
 {% macro folder(folder, display_name)                                       -%}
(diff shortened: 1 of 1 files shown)


tomprince:
Github provides two SHA1s for each pull request: The tip of the branch being merged (`refs/pull/<NNN>/head`), and a merge commit, merging it into the target branch (`refs/pull/<NNN>/merge`). I wonder if reporting status against that commit would correctly display the status on github? In that case, it seems like using a custom code might be avoided.

jobs.groovy.j2
```diff
@@ -90,6 +90,28 @@
...
+{# Send a request to Github to update the commit with the build status       #}
+REMOTE_GIT_COMMIT=`git rev-list --max-count=1 upstream/{{branch}}`
+BUILD_URL="\$JENKINS_URL/job/{{dashProject}}/job/{{dashBranchName}}/"
+PAYLOAD="{\\"state\\": \\"{{ status }}\\", \\"target_url\\": \\"\$BUILD_URL\\", \\"description\\": \\"{{ description }}\\", \\"context\\": \\"jenkins-multijob\\" }"
+curl \\
+  -H "Content-Type: application/json" \\
+  -H "Authorization: token \$GITHUB_STATUS_API_TOKEN" \\
+  -X POST \\
+  --data "\$PAYLOAD" \\
+  https://api.github.com/repos/ClusterHQ/flocker/statuses/\$REMOTE_GIT_COMMIT
```
Azulinho: I think you might be able to use some jinja here {{ cfg.project }}
from: https://github.com/ClusterHQ/flocker/blob/master/build.yaml#L56

something along the lines of:
https://api.github.com/repos/{{ cfg.project }}/statuses/\$REMOTE_GIT_COMMIT

jobs.groovy.j2
```diff
@@ -644,6 +672,25 @@ branches.each {
             failUnstable(true)
             failNoReports(false)
           }
+{# Update the commit status on GitHub with the build result                  #}
+          flexiblePublish {
+            condition {
```
Azulinho: Can you had a link to the plugin page for future reference?: https://wiki.jenkins-ci.org/display/JENKINS/Flexible+Publish+Plugin 

and another for the conditions block?
https://wiki.jenkins-ci.org/display/JENKINS/Run+Condition+Plugin

the flexiblePublish is not very well documented upstream (well most jenkins plugins have little documention, but this one is a bit worse than most). 
The job DSL api:
https://jenkinsci.github.io/job-dsl-plugin/#method/javaposse.jobdsl.dsl.helpers.publisher.PublisherContext.flexiblePublish
states that the build step capability used below is part of the 'any build steps' plugin.
https://wiki.jenkins-ci.org/display/JENKINS/Any+Build+Step+Plugin

jobs.groovy.j2
```diff
@@ -90,6 +90,28 @@
 {# JINJA2 MACROS DEFINED BELOW                                               #}
 {# --------------------------------------------------------------------------#}
 
+{% macro updateGitHubStatus(status, description, branch, dashProject,
+                            dashBranchName)                                 -%}
+"""
+#!/bin/bash
+set +x
```
Azulinho: maybe state why we force a +x

# Jenkins runs jobs using 'set -x'
# we set '+x' so that we don't leak credentials to the logs
set +x

Azulinho:
thanks for this work, I am happy with what it does which is to update github with the status of the main multijob.
Tom's comment might be useful, but I will let you decide if you want to rework this or not.

Only have a couple of suggestions which I commented above, but from my side please address those, run a final build test and merge.

zubron:
Thanks for the info @tomprince. The reason we had to add the code to do this manually is that we don't have control over the commit that Jenkins reports that it is using for the build and the merge commit that it reports is one that is actually created locally before the build. Also, when we're performing a build, it might not necessarily be from a PR as builds for a branch can also be triggered manually.

If we decide to change how we handle PRs then knowing about what GitHub provides will definitely be useful. It's just a shame that we can't use that here :(

PR #2249, opened 2015-11-27
[FLOC-3556] split jobs into tabs

Azulinho:
This adds multiple tabs at the top of the job list to help with the large number of jobs in the branch.

It filters jobs based on:
Cloud Provider
Client
Acceptance
Trial
OperatingSystem
other bits and pieces

view with the added tabs:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/split-jobs-into-different-tabs-FLOC-3556/

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -738,7 +738,6 @@ common_cli:
     # flake8 is used by lint tests on Flocker source code
     \${PARSE_LOGS} pip install flake8 \${PIP_ADDITIONAL_OPTIONS}
 
-
 #-----------------------------------------------------------------------------#
 # Job Definitions below this point
 #-----------------------------------------------------------------------------#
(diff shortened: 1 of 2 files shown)


build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
+    regex: '.*_(?i)trusty_.*.*'
+  Vivid:
+    description: 'All Ubuntu Vivid jobs'
+    regex: '.*_(?i)vivid_.*.*'
+  Wily:
+    description: 'All Ubuntu Wily jobs'
+    regex: '.*_(?i)wily_.*.*'
+  cron:
+    description: 'All Nightly Cron jobs'
+    regex: '^_[^_](?i)[a-z]*.*'
```
jml: I don't really know what these actually are, but I think this would be a great place for a longer (possibly even multi-paragraph) description of what they are and how they relate to the rest of the build.

build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
+    regex: '.*_(?i)centos_.*.*'
+  Trusty:
+    description: 'All Ubuntu Trusty jobs'
+    regex: '.*_(?i)trusty_.*.*'
+  Vivid:
+    description: 'All Ubuntu Vivid jobs'
+    regex: '.*_(?i)vivid_.*.*'
+  Wily:
+    description: 'All Ubuntu Wily jobs'
+    regex: '.*_(?i)wily_.*.*'
```
jml: Can we please use release numbers, rather than code names, for the Ubuntu releases? (e.g. 14.04 LTS, 15.04, 15.10)

build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
+
+views:
+  trial:
+    description: 'All trial jobs'
+    regex: 'run_(?i)trial_.*.*'
+  acceptance:
+    description: 'All Acceptance jobs'
+    regex: 'run_(?i)acceptance_.*.*'
+  client:
+    description: 'All client instalation jobs'
```
jml: "installation"

build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
+#-----------------------------------------------------------------------------#
+# View definitions below this point
+#-----------------------------------------------------------------------------#
+
+views:
+  trial:
+    description: 'All trial jobs'
+    regex: 'run_(?i)trial_.*.*'
+  acceptance:
+    description: 'All Acceptance jobs'
```
jml: Would it make sense here to say "All acceptance tests" :grey_question:

build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
       timeout: 120
+
+#-----------------------------------------------------------------------------#
+# View definitions below this point
+#-----------------------------------------------------------------------------#
+
+views:
+  trial:
+    description: 'All trial jobs'
+    regex: 'run_(?i)trial_.*.*'
```
jml: What does it mean for something to be a trial job?

build.yaml
```diff
@@ -1204,3 +1203,55 @@ job_type:
...
+    regex: 'run_(?i)lint.*.*'
+  docs:
+    description: 'Documentation jobs'
+    regex: 'run_(?i)sphinx.*.*'
+  cinder:
+    description: 'cinder jobs'
+    regex: '.*_(?i)cinder_.*.*'
+  ebs:
+    description: 'ebs jobs'
+    regex: '.*_(?i)ebs_.*.*'
```
jml: I was surprised to see that these jobs _don't_ appear on "AWS jobs". I don't know if that's a bug in the regexes, or a bug in my understanding (which might be better fixed by more documentation). :grey_question:

jml:
Looking at http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/split-jobs-into-different-tabs-FLOC-3556/view/, I see that the "AWS" tab appears to the left of "All". Is it possible to make it so that "All" is always the left-most?

jml:
Thanks @Azulinho! This is a very useful change, especially since it gives us a hook for adding more documentation for our tests—having a build system that's more self-documenting is one of the advantages of switching to Jenkins.

I've got a few questions and one or two minor English glitches. Feel free to bounce back for another review if you'd like (probably esp. if you add lots of text), but not strictly necessary.

build.yaml
Azulinho: that would mean renaming all the jobs, and lose history.
I also think this might be a bit of a bike shed issue, as nothing prevents a future PR from suggesting the opposite

build.yaml
Azulinho: well... its a filter for jenkins 'jobs'.

build.yaml
Azulinho: or are you refering to only the name of the Tab?

build.yaml
Azulinho: "All jobs that run 'trial' " <-- how does that sound ?

build.yaml
Azulinho: fixed

build.yaml
Azulinho: let me try it out.

When browsing do you find the 'AWS' and 'Rackspace' usefull ?
or could I simply drop them ?

Azulinho:
"I see that the "AWS" tab appears to the left of "All". Is it possible to make it so that "All" is always the left-most?" <<- not sure, let me find out.

build.yaml
jml: I meant only the name of the tab and the description. Later we should probably change the jobs in a future PR.

build.yaml
jml: OK

build.yaml
jml: Sounds good. I guess the thing is that as a flocker developer, I don't know what I'd use it for. (I have no objections against having it).

build.yaml
jml: Yes, I do find them useful. Please keep them.

build.yaml
Azulinho: well... they are jobs that are executed directly with 'trial' vs 'run-<something>-tests'

Azulinho: I will leave them there, this will eventually get refactored as people work on it and find out better ways to refer to these 'things'

Azulinho:
fixed and addressed all requests above,

PR #2250, opened 2015-11-30
[FLOC-2560] Download logs from Ubuntu acceptance testing nodes

mpgomez:
It has been manually tested, but I am not sure if we add tests for the acceptance test runner or not. I found some of them when trying to add a test at least for the regex.

If tests are needed (not sure anyway if testing the test runner may be a bit overtesting), any advice on how and where is welcome :D


admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+            b'/var/log/flocker/flocker-dataset-agent.log',
+            b'/var/log/flocker/flocker-container-agent.log',
+            b'/var/log/flocker/flocker-docker-plugin.log',
+            b'/var/log/upstart/docker.log',
+        ],
+        handle_stdout=formatter.handle_output_line,
+    )
+    ran.addErrback(write_failure, logger=None)
+    # Deliver a final empty line to process the last message
+    ran.addCallback(lambda ignored: formatter.handle_output_line(b""))
```
jml: You almost certainly want to return `ran` from here, so that anything that calls `capture_upstart` can wait for the results.

admin/acceptance.py
```diff
@@ -1055,6 +1151,10 @@ def main(reactor, args, base_path, top_level):
             remote_logs_file = open("remote_logs.log", "a")
             for node in cluster.all_nodes:
                 capture_journal(reactor, node.address, remote_logs_file)
+        elif options['distribution'] in ('ubuntu-14.04', 'ubuntu-15.04'):
+            remote_logs_file = open("remote_logs.log", "a")
+            for node in cluster.all_nodes:
+                capture_upstart(reactor,node.address,remote_logs_file)
```
jml: Please put a space after the commas.

mpgomez:
Jenkins build:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/ubuntu-logs-FLOC-2560-2/

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+
+class tail_formatter(object):
+    """
+    Formatter for the output of the ``tail`` commands that will produce logs
+    with Eliot messages with the same format as the ones produced when
+    parsing journalctl output
+
+    :ivar file output_file: log file where we want to write our log
+    :ivar bytes _host: ip address or identifier of our host to be
+        added to the Elliot messages
```
jml: `%s/Elliot/Eliot/g`

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+    ran.addCallback(lambda ignored: formatter.handle_output_line(b""))
+
+
+class tail_formatter(object):
+    """
+    Formatter for the output of the ``tail`` commands that will produce logs
+    with Eliot messages with the same format as the ones produced when
+    parsing journalctl output
+
+    :ivar file output_file: log file where we want to write our log
```
jml: `s/output_file/_output_file/`

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+    :ivar bytes _host: ip address or identifier of our host to be
+        added to the Elliot messages
+    :ivar bytes service: optional initial name of the service. This initial
+        name shouldn't appear anywhere unless there is an error, as the first
+        line of the output of trial will be a file name, that will be used
+        to set the name of the service we are currently parsing
+    """
+    def __init__(self, output_file, host, service = "unknown"):
```
jml: No spaces around `=` in default arguments.

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+            b'/var/log/upstart/docker.log',
+        ],
+        handle_stdout=formatter.handle_output_line,
+    )
+    ran.addErrback(write_failure, logger=None)
+    # Deliver a final empty line to process the last message
+    ran.addCallback(lambda ignored: formatter.handle_output_line(b""))
+
+
+class tail_formatter(object):
```
jml: Classes should be CamelCase, e.g. `TailFormatter`

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+    """
+    Formatter for the output of the ``tail`` commands that will produce logs
+    with Eliot messages with the same format as the ones produced when
+    parsing journalctl output
+
+    :ivar file output_file: log file where we want to write our log
+    :ivar bytes _host: ip address or identifier of our host to be
+        added to the Elliot messages
+    :ivar bytes service: optional initial name of the service. This initial
```
jml: `s/service/_service/`

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+        name shouldn't appear anywhere unless there is an error, as the first
+        line of the output of trial will be a file name, that will be used
+        to set the name of the service we are currently parsing
+    """
+    def __init__(self, output_file, host, service = "unknown"):
+        self._output_file = output_file
+        self._host = host
+        self._service = service
+        self.service_regexp = re.compile(r"[/var/log/flocker/]|[/log/upstart/](.*)\.log")
```
jml: Probably should make `service_regexp` private as well.

jml: I don't think this regex is exactly right.

```
In [1]: import re

In [2]: some_re = re.compile(r"[/var/log/flocker/]|[/log/upstart/](.*)\.log")

In [3]: some_re.search('v.log')
Out[3]: <_sre.SRE_Match at 0x1048fae40>
```

The square brackets indicate character classes. In order to match literal `[` and `]`, we need to escape them with `\`.

admin/acceptance.py
```diff
@@ -950,6 +951,101 @@ def eliot_output(message):
...
+    output file.
+
+    :param reactor: The reactor.
+    :param bytes host: Machine to SSH into.
+    :param file output_file: File to write to.
+    """
+    # TODO: check the format we need to use here (if we need it at all)
+    # can use it later. Using the whole filename instead of smtg like flocker*
+    # because I can't make it work inside the command. It has a funny way of
+    # parsing the commands. Suggestions are welcome :D
```
jml: I don't really know what we need to do regarding formatting, so unfortunately I can't help. 

If you want to use `flocker-*` then you need to run the command via a shell, e.g.

```python
command = [
    b'/bin/sh',
    b'tail',
    b'/var/log/flocker/flocker-*.log',
    b'/var/log/upstart/docker.log',
]
```

This might need some tweaking though.

jml:
Hi Pilar, 

Thanks for this! I've mostly got a few minor pedantic comments re Python style. The regex probably needs looking at, and I think the right thing to do is to add some tests for the `TailFormatter` object. These should go in `admin.test.test_acceptance`.

jml

admin/acceptance.py
```diff
@@ -950,6 +951,98 @@ def eliot_output(message):
...
+def capture_upstart(reactor, host, output_file):
+    """
+    SSH into given machine and capture relevant logs, writing them to
+    output file.
+
+    :param reactor: The reactor.
+    :param bytes host: Machine to SSH into.
+    :param file output_file: File to write to.
+    """
+    formatter = TailFormatter(output_file, host)
```
itamarst: You're not passing in `service` here, so it'll always be "unknown". You should be able to get service by tailing each file separately and having a TailFormatter for each.

admin/acceptance.py
```diff
@@ -950,6 +951,98 @@ def eliot_output(message):
...
+class TailFormatter(object):
+    """
+    Formatter for the output of the ``tail`` commands that will produce logs
+    with Eliot messages with the same format as the ones produced when
+    parsing journalctl output
+
+    :ivar file output_file: log file where we want to write our log
+    :ivar bytes _host: ip address or identifier of our host to be
+        added to the Eliot messages
+    :ivar bytes service: optional initial name of the service. This initial
```
itamarst: This is incorrect; the input files are different, but on the output side all the logs are being interleaved into a single output file, so there's way to tell from which service an individual message comes from. This should be a required input.

itamarst: The regex thing is unnecessary and seems very fragile to me; you should just require this and have the caller pass the service name in since it knows it, by having a tail per-service (see above).

admin/acceptance.py
```diff
@@ -1055,6 +1148,10 @@ def main(reactor, args, base_path, top_level):
             remote_logs_file = open("remote_logs.log", "a")
             for node in cluster.all_nodes:
                 capture_journal(reactor, node.address, remote_logs_file)
+        elif options['distribution'] in ('ubuntu-14.04', 'ubuntu-15.04'):
```
itamarst: Does 15.04 actually use upstart logs anymore?

admin/test/test_acceptance.py
```diff
@@ -191,3 +191,41 @@ def test_non_json_message(self):
             )],
             self._convert(NON_JSON_JOURNAL_EXPORT),
         )
+
+
+class TailFormatterRegex(SynchronousTestCase):
+    """
+    Tests for ``TailFormatter``
+    """
+    def setUp(self):
```
itamarst: Tests for the JSON/non-JSON cases?

wallrj:
Thanks @mpgomez 

I clicked the review button on this issue before realising that @itamarst  and @jml had already given feedback.
I probably shouldn't get involved, but since I'd already checked out the branch I decided to give it a try.
It works but I noted a couple of things.
 1. The remote_logs.log  lines contain `_SYSTEMD_UNIT` fields which seems odd because those aren't present in the log files on the nodes and Ubuntu doesn't use systemd.

```
(2560)(ubuntu-logs-FLOC-2560-2 ? =)[~/.../HybridLogic/flocker]$ ssh root@119.9.108.114 'egrep -i SYSTEMD /var/log/flocker/*' | wc 
      0       0       0
```

And yet remote_logs contains the following:

```
{
  "_SYSTEMD_UNIT": "flocker-control",
  "timestamp": 1448923624.874645,
  "action_status": "succeeded",
  "task_uuid": "9c360485-253b-4dc9-b9f0-c3167f89d79f",
  "_HOSTNAME": "119.9.108.114",
  "action_type": "eliot:remote_task",
  "task_level": [
    2,
    2,
    2,
    2,
    2,
    2,
    2,
    2,
    3
  ]
}
```

I'll keep the logs and the nodes in case you want to take a look.

1. I agree with @itamarst 's comments about the use of RegExp for identifying the source of each log. Seems like it'd be simpler and much more reliable to tail them separately.

 1. And secondly, I think it might be wise to store the deferreds (the `ran` variable) returned by `capture_upstart` and check their status at the end of the acceptance test run. Use `gather_deferreds` for that.
  1. I notice the journalctl code doesn't do that, and I'm not sure why.

Please answer or address the points above and resubmit for a final review.

mpgomez:
Moving now to help out with Jenkins so I can verify the changes properly :D

jml:
Jenkins should _probably_ be working now, more or less. Let me know when / if you want further input.

PR #2251, opened 2015-11-30
[FLOC-3558] Fixed lint

thepeak99:


jml:
This is OK to land once the tests pass on Jenkins.

jongiddy:
Note, the `run_trial_on_AWS_CentOS_7_benchmark` will fail on Jenkins. This will be fixed by FLOC-3561, but you should merge this PR even though this test fails. Otherwise, we will have a deadlock.

PR #2252, opened 2015-11-30
[FLOC 3518] Move Container API documentation

davidoliverSP2:
Fixes 3518

To make it clearer that Flocker focuses on volumes, not containers, this PR moves the container API endpoints to a new document: Deprecated Endpoints, and makes it easier for the user to locate the API documentation. (this is in response to user feedback gathered at DockerCon EU 2015)

Note: 
Step 2 of FLOC 3518 handles the addition of sections to other documentation such as tutorials. This will be handled in FLOC 3559


docs/reference/index.rst
```diff
@@ -1,11 +1,62 @@
...
+
+For more information read the :ref:`cluster architecture<architecture>` documentation.
+
+REST API Endpoints
+==================
+
+.. contents::
+        :local:
+
+Common API Endpoints
```
jongiddy: If we only have the single "dataset" API, then having a "common" API doesn't make sense. Does it work better to remove the headers for the the two sections?

jongiddy:
Since the container API still exists, and may be in use, we should still provide a reference to it. Is it possible to keep the container API, but "footnote" it?

jongiddy:
Please consider how we could continue to provide the container API in a way that makes it available for users looking for it, but makes it less obvious for people browsing the docs for ways to build new systems.

docs/reference/deprecated_api.rst
```diff
@@ -0,0 +1,22 @@
...
+
+.. XXX: As part of FLOC 3518 the Container API Endpoints below have been deprecated, as Flocker is not a container framework:
+
+.. warning:: 
+   The endpoints listed in this document are deprecated, and will be removed in a future version of Flocker.
+
+The Flocker Container API enables you to manage containers in your cluster.
+However, ClusterHQ is no longer developing any future features in this area, as you can now use Flocker with 3rd party orchestration tools.
+
+The deprecated endpoints listed below are still available for use, but please be aware that they will be removed in a future version of Flocker:
```
exarkun: Hmm!  Forward looking statements.  Oh my.  Always a bit out on a limb with these...

The philosophy behind the URL hierarchy used with these endpoints - specifically the `v1` prefix - is that as long as we provide these at all, we won't change how they work or what they do.  And if we're being really precise, I think that means that the whole collection of v1 endpoints should remain unchanged.  This probably means that what we're likely to do is introduce a `v2` which has the dataset endpoints but not the container endpoints.  Then we'll carry on with v1 and v2 for a while and then eventually we'll remove all of v1 - at which point, indeed, the container endpoints will be entirely unavailable.

However, we're very inexperienced in this area (we are at v1, after all - ie, we've never *actually* gone through such a transformation) so there's some chance when we actually go and try to do this we'll discover some reason it's a bad idea (or maybe even an infeasible idea) and have to do something else.

My best suggestion is to change the phrasing here to be something like "they will not be available in future versions of the Flocker API".  Pretty similar and maybe the subtlety will be lost on most of our audience without a more complex explanation of our compatibility policy and change workflow (which I suspect is currently missing from the docs) - so I could be making too much of this point - but also easy to do so maybe worth it.

Maybe @itamarst has some thoughts about this as well.

exarkun:
Thanks.  Looks cool to me apart from the one minor wording change that I wrote three whole paragraphs about inline.

It might be nice if we had a simple way to consistently markup/style deprecated endpoints for the future times when we decide to deprecate stuff.  Making sure we convey the deprecation consistently will make it easier for our users to notice and react appropriately.

I think what you've done here is just fine for now, though.  I also like how the index is a bit easier to navigate as a result of this change.

Feel free to merge (or wait for @itamarst to weigh in if you want).

PR #2253, opened 2015-11-30
[FLOC-3514] YAML configuration of benchmark scenarios, operations, and metrics

jongiddy:
It is useful to perform the same benchmark for a series of parameters (e.g. measured with 10, 20, 30 request/sec). Command line options for benchmarking make it difficult to parametrize the various options. 

This PR allows the scenarios, operations, and metrics provided to the benchmarking script to be read from a YAML file, allowing each one to contain additional fields for parameterization.

An example YAML file is added, in which the `wait` operation has been parameterized to wait 10 seconds.

This can be tested by running:
```
./benchmark --control ${FLOCKER_ACCEPTANCE_CONTROL_NODE} --certs ${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} --operation wait-10 --metric cputime
```

The options to the command now name a scenario, operation, or metric in the YAML file, which maps to a type, which selects the class to use, and a set of keyword parameters to that class.  Classes are also provided with reactor and control service parameters.

The creation of the classes is moved from `script.py` to `_driver.py` to allow creation of the class with the reactor parameter.

The above command returns output as before, but with the operation field containing the parameters:
```
{
  "scenario": {
    "type": "no-load"
  }, 
  "timestamp": "2015-11-30T12:17:11.961664", 
  "metric": {
    "type": "cputime"
  }, 
  "client": {
    "username": "jon", 
    "platform": "Linux-3.13.0-68-generic-x86_64-with-Ubuntu-14.04-trusty", 
    "working_directory": "/home/jon/projects/clusterhq/flocker-alt/benchmark", 
    "nodename": "ecuador", 
    "flocker_version": "1.8.0+84.g9943194"
  }, 
  "samples": [
    {
      "value": {
        "172.99.68.188": {
          "flocker-dataset": 1, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }, 
        "172.99.67.245": {
          "flocker-control": 0, 
          "flocker-dataset": 0, 
          "flocker-docker-": 0, 
          "flocker-contain": 0
        }
      }, 
      "success": true
    }, 
    [...]
  ], 
  "control_service": {
    "flocker_version": "1.8.0+82.g7e996e7", 
    "host": "172.99.67.245", 
    "port": 4523
  }, 
  "operation": {
    "wait_seconds": 10, 
    "type": "wait"
  }
}
```

### DIFF
--- a/benchmark/operations/no_op.py
+++ b/benchmark/operations/no_op.py
@@ -21,10 +21,9 @@ class NoOpProbe(object):
 @implementer(IOperation)
 class NoOperation(object):
     """
-    An nop operation.
+    A no-op operation.
     """
 
-    # attributes unused, but required for __init__ signature
     def __init__(self, clock, control_service):
         self.clock = clock
         self.control_service = control_service
--- a/benchmark/operations/no_op.py
+++ b/benchmark/operations/no_op.py
@@ -1,4 +1,3 @@
-from pyrsistent import PClass, field
 from zope.interface import implementer
 
 from twisted.internet.defer import succeed
@@ -7,7 +6,7 @@ from .._interfaces import IProbe, IOperation
 
 
 @implementer(IProbe)
-class NoOpProbe(PClass):
+class NoOpProbe(object):
     """
     A probe that performs no operation.
     """
@@ -20,14 +19,15 @@ class NoOpProbe(PClass):
 
 
 @implementer(IOperation)
-class NoOperation(PClass):
+class NoOperation(object):
     """
     An nop operation.
     """
 
     # attributes unused, but required for __init__ signature
-    clock = field(mandatory=True)
-    control_service = field(mandatory=True)
+    def __init__(self, clock, control_service):
+        self.clock = clock
+        self.control_service = control_service
 
     def get_probe(self):
         return NoOpProbe()
--- a/benchmark/operations/read_request.py
+++ b/benchmark/operations/read_request.py
@@ -28,7 +28,6 @@ class ReadRequest(object):
     An operation to perform a read request on the control service.
     """
 
-    # `clock` unused, but required for __init__ signature
     def __init__(self, clock, control_service):
         self.clock = clock
         self.control_service = control_service
--- a/benchmark/operations/read_request.py
+++ b/benchmark/operations/read_request.py
@@ -1,4 +1,3 @@
-from pyrsistent import PClass, field
 from zope.interface import implementer
 
 from twisted.internet.defer import succeed
@@ -7,12 +6,13 @@ from .._interfaces import IProbe, IOperation
 
 
 @implementer(IProbe)
-class ReadRequestProbe(PClass):
+class ReadRequestProbe(object):
     """
     A probe to perform a read request on the control service.
     """
 
-    control_service = field(mandatory=True)
+    def __init__(self, control_service):
+        self.control_service = control_service
 
     def run(self):
         d = self.control_service.list_datasets_state()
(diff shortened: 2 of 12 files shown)


benchmark/script.py
```diff
@@ -56,33 +31,48 @@ class BenchmarkOptions(Options):
...
+operation_name = options['operation']
+metric_name = options['metric']
 
-scenario = _SCENARIOS[config['scenario']]
-operation = _OPERATIONS[config['operation']]
-metric = _METRICS[config['metric']]
+scenario_config = scenarios.get(scenario_name) or usage(
+    'No such scenario: {!r}'.format(scenario_name))
+operation_config = operations.get(operation_name) or usage(
+    'No such operations: {!r}'.format(operation_name))
```
jongiddy: `operations` -> `operation`

benchmark/operations/no_op.py
```diff
@@ -20,14 +19,15 @@ def cleanup(self):
 
 
 @implementer(IOperation)
-class NoOperation(PClass):
+class NoOperation(object):
     """
     An nop operation.
     """
 
     # attributes unused, but required for __init__ signature
```
zubron: We no longer need this comment.

benchmark/operations/read_request.py
```diff
@@ -23,14 +23,15 @@ def cleanup(self):
 
 
 @implementer(IOperation)
-class ReadRequest(PClass):
+class ReadRequest(object):
     """
     An operation to perform a read request on the control service.
     """
 
     # `clock` unused, but required for __init__ signature
```
zubron: We no longer need this comment.

benchmark/script.py
```diff
@@ -95,14 +85,14 @@ class BenchmarkOptions(Options):
         nodename=node(),
         platform=platform(),
     ),
-    scenario=config['scenario'],
-    operation=config['operation'],
-    metric=config['metric'],
+    scenario=scenario_config.copy(),
+    operation=operation_config.copy(),
+    metric=metric_config.copy(),
 )
```
zubron: Could you add a comment to explain why we choose to copy the dictionaries at this point? Either that or provide some documentation for `driver` that explains that the different `*_config` parameters will be modified.

benchmark/script.py
```diff
@@ -56,33 +31,48 @@ class BenchmarkOptions(Options):
...
+
+if not options['control'] and options['operation'] != 'no-op':
+    # No-op is OK with no control service
+    usage('Control service required')
+
+with open(options['config'], 'rt') as f:
+    config = yaml.safe_load(f)
+    scenarios = config['scenarios']
+    operations = config['operations']
+    metrics = config['metrics']
```
zubron: It would be good to add validation of this configuration file. If a particular section in the file has an invalid configuration, it will result in a `KeyError` when it is used. When this error is encountered, it's unclear what the cause is. For example, if the key `type` is missing for any of the `scenarios`, `operations`, or `metrics` then the error will not be encountered until the `driver` has started.

zubron: We do validation of yaml configuration elsewhere. See https://github.com/ClusterHQ/flocker/blob/eb26ab07c0902757099a7113abdef52580ec36cc/flocker/node/script.py#L169-L217

zubron:
Thanks for this @jongiddy. I've requested a few changes, please address the comments and then resubmit.

jongiddy:
Addressed all comments, and updated documentation to describe configuration file.

Minor fix to OS detection code, to ensure tests are _not_ skipped on Ubuntu.

jongiddy:
Also, moved configuration parsing back to main script file, which allows nicer error messages, and removes the need to copy the config dicts.

zubron:
This look good. Thanks for making the additional changes :) Please merge when the tests pass.

PR #2254, opened 2015-11-30
[FLOC-3561] Select standard test process based on OS

jongiddy:
Benchmarking tests were working on Ubuntu, but not on CentOS.

Some tests assumed that a process named `init` always existed. Fixed by using `systemd` if OS is CentOS.  Tests are skipped if not on Ubuntu or CentOS.

This problem was highlighted by change from Buildbot to Jenkins, which adds tests for the `benchmark` directory.


jml:
LGTM. Thanks!

PR #2256, opened 2015-11-30
By default clean up builds; previously some jobs didn't.

itamarst:
We're running out of disk space, so make sure we clean up as much as possible.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -764,7 +764,7 @@ common_cli:
 # Toggles:
 #   * archive_artifacts: ( define if there are files to be archived)
 #   * coverage_report: ( enable if this job produces a coverage report file)
-#   * clean_repo: (enable if we need to clean old files owned by root)
+#   * dont_clean_repo: (enable if we don't want to delete old files owned by root)
 #
 
 # run_trial_modules contains a list of all the modules we want to execute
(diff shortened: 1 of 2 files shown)


exarkun:
How much does this actually clean up?  "Files owned by root" seems like a strange criterion to apply for cleanup.  How about all the non-root owned files still taking up space?  Presumably they'll eventually lead to an out-of-space condition too.

So maybe this it a good change, but it's hard to evaluate without knowing how much space it's going to save.  But it's probably not sufficient to resolve the issue.

exarkun:
> But it's probably not sufficient to resolve the issue.

Therefore, clicking "changes requested".

Azulinho:
looks like this one here:
https://wiki.jenkins-ci.org/display/JENKINS/Workspace+Cleanup+Plugin

build.yaml
```diff
@@ -1138,7 +1124,7 @@ job_type:
                    *run_build_sh_script_on_vagrant_box,
                    *destroy_vagrant_box]
           }
-      clean_repo: false
+      dont_clean_repo: true
```
jml: Can you please add a comment explaining why we don't want to clean the repo here?

jml:
Failures look intermittent + FLOC-3564.

Please add the comment & merge when ready.

build.yaml
itamarst: I have no idea, it was there when I got here.

PR #2257, opened 2015-11-30
Use failureResultOf to verify failed Deferred in tests.

sarum90:
Fixes FLOC-3574 where I rolled my own instead of using the twisted provided built-in.


wallrj:
Thanks @sarum90 

Looks good to me.

* `failureResultOf` raises an assertion error unless the supplied deferred has fired. I can't tell  whether the home grown version had the same behaviour.

sarum90:
All builds have passed on Jenkins at some point in time, all current failures look like rackspace is unhealthy at the moment. Separately I'm looking down that issue, but almost certainly unrelated to this change, so I'm merging.

PR #2260, opened 2015-12-01
[FLOC 3577] Temporarily remove the TrueAbility links from the documentation

davidoliverSP2:
Fixes 3577

Due to TA not working, it has been taken offline to be updated. While this is down, this PR suppresses 2 locations in the documentation which referenced it.

### DIFF
--- a/docs/install/index.rst
+++ b/docs/install/index.rst
@@ -6,7 +6,9 @@ Installing Flocker
 
 There are several installation options with which you can get up and running with Flocker.
 
-You can also try Flocker without installation, either in our live hosted environment or on virtual machines using our Vagrant image. For more information, see :ref:`get-started`.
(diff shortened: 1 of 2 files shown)


exarkun:
Thanks.

I think I'd prefer to see the documentation just removed rather than commented out.  Leaving a link to this PR or the issue will make it easy to recover the text if, when we want to bring this back, we want to see the text for reference or restore it verbatim.

Leaving it there commented out clutters the file and hampers further maintenance.

Or at least, that's the case for source code.  I tend to think it makes sense for documentation as well.  Since you're the primary maintainer here, happy to hear ideas to the contrary, though.

So please either:

  * delete instead of commenting out or
  * leave as-is and explain the motivation for doing so

and then merge.  Thanks!

davidoliverSP2:
thanks @exarkun 

I like option 1. I'll dust off my delete button.....

PR #2261, opened 2015-12-01
Skip test_rpm on Ubuntu

jml:
This probably fixes the failing Jenkins test at http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/master/job/run_trial_on_AWS_Ubuntu_Trusty_admin/lastCompletedBuild/testReport/admin.test.test_packaging/LintPackageTests/test_rpm/

It does so by skipping the test on Ubuntu, which is unsatisfactory, but might lead to us having passing Jenkins builds on master.

The next step is to actually understand and fix the problem.

### DIFF
--- a/admin/test/test_packaging.py
+++ b/admin/test/test_packaging.py
@@ -5,6 +5,7 @@ Tests for ``admin.packaging``.
 """
 
 from glob import glob
+import platform
 from subprocess import check_output
 from textwrap import dedent
 from unittest import skipIf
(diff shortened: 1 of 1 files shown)


admin/test/test_packaging.py
```diff
@@ -41,6 +42,9 @@
 require_dpkg = skipIf(not which('dpkg'), "Tests require the ``dpkg`` command.")
 require_lintian = skipIf(not which('lintian'),
                          "Tests require the ``lintian`` command.")
+require_not_ubuntu = skipIf(
+    platform.linux_distribution()[0] == 'Ubuntu',
+    "rpmlint returns spurious results on Ubuntu: FLOC-3564.")
```
exarkun: It would be exactly equivalent to make this "require centos" since we only have two supported platforms.  I would have spelled it that way ... but since it's exactly equivalent I guess I shouldn't say anything.

exarkun:
Thanks.  Looks right.  Please merge when you see it skip on the right CI jobs.

PR #2262, opened 2015-12-01
[FLOC-3517] Conditional actions in REST API client

itamarst:
This is the corresponding client-side API for doing conditional queries against the REST API.

It builds on unmerged PR #2225.


flocker/apiclient/_client.py
```diff
@@ -446,8 +514,13 @@ def _request(self, method, path, body, success_codes, error_codes=None):
         :param set success_codes: Expected success response codes.
         :param error_codes: Mapping from HTTP response code to exception to be
             raised if it is present, or ``None`` to set no errors.
+        :param with_headers: Boolean indicating whether headers are required.
+        :param configuration_tag: If not ``None``, include value as
+            ``X-If-Configuration-Matches`` header.
 
-        :return: ``Deferred`` firing with decoded JSON.
+        :return: ``Deferred`` firing with decoded JSON, or if
```
sarum90: :grey_question: This API looks a little gross, but seems to be internal only so probably okay as is.

Slight preference for making a new method `_request_with_headers` that has the body of this function and with_headers set to True. Then changing the content of this function to be a call to `d = self._request_with_headers(...); d.addCallback(lambda x: x[0]); return d` or equivalent.

flocker/apiclient/_client.py
```diff
@@ -522,25 +603,32 @@ def create_dataset(self, primary, maximum_size=None, dataset_id=None,
...
+                                with_headers=True)
         request.addCallback(
-            lambda results:
-            [
-                self._parse_configuration_dataset(d)
-                for d in results if not d['deleted']
-            ]
+            lambda (results, headers):
+            DatasetsConfiguration(
+                tag=headers.getRawHeaders('X-Configuration-Tag')[0],
```
sarum90: :grey_question: IIUC, using this client with a version of flocker before #2225 is merged would result in a `list index out of range` error instead of some nice version mismatch error, or just handling the cross version incompatability.

I guess given that the flocker client is not its own package people should not be expecting any consumer of this to be cross version compatible. I guess if you think the odds of someone using this client with a pre #2225 server is low fine as is. Otherwise it'd be nice to at least have a nicer error here.

sarum90:
This looks good to me. Seems to provide a nice API to handle changes.

Resolve merge conflicts, land the precursor, get a green run of the tests, address, and merge. :smile:

flocker/apiclient/_client.py
itamarst: Good idea!

flocker/apiclient/_client.py
itamarst: Yeah, I'm fine with only supporting latest version.

PR #2265, opened 2015-12-01
[FLOC-2963] No more choosing dataset_id via hashing

itamarst:
Get rid of a hack that likely caused issues with e.g. deleted datasets.

Builds on #2262 and #2225.


jml:
Great, thanks. Merge when you've made your peace with the tests.

PR #2266, opened 2015-12-01
[FLOC-3455] Better exception logging

itamarst:
1. Upgrade Eliot, and therefore Pyrsistent, so we can get access to new Eliot feature.
2. use the new Eliot feature to make sure all BadRequest exceptions get logged with machine readable response code.
3. Improve default string rendering of a few exceptions so they get logged more informatively by Eliot.


jml:
Looks good, thanks!

PR #2269, opened 2015-12-02
[FLOC 3429] Add a concept topic about Flocker Leases

davidoliverSP2:
Fixes 3429

As part of the updates on the Flocker Concepts section of the docs, this addition explains the concept of Flocker Leases.


jml:
LGTM, thanks.

PR #2271, opened 2015-12-02
[FLOC 3599] Ensure header style is consistent

davidoliverSP2:
Fixes 3599

The installing pages used inconsistent style on similar headers in the Client installation page, and the Node installation page. This is a fix.


jml:
Glorious. Thanks.

PR #2272, opened 2015-12-02
[FLOC-3602] Design sketch: make the container agent optional.

itamarst:
Currently the container agent is required, even if one is not using Flocker's container API. This uses resources (CPU, disk, RAM) without offering any benefit. In the medium term we want to remove the agent completely, but for now it would be good to simply allow people to disable it.

The way to do that is to make sure the dataset agent no longer requires the container agent be running in order to operate. This means that if the container agent becomes unresponsive somehow for more than two minutes and a container is still using a dataset we might end up moving a dataset that is still in use. In practice this is unlikely to happen due to fact one can't unmount a filesystem that is in use, so allowing this edge case seems reasonable.

This issue does not cover documentation changes to make it clear that the container agent is optional, except perhaps a release notes comment; the relevant docs are already quite complex. This will however allow us to improve the docs in a separate issue as part of planned docs changes in this area.

### DIFF
--- a/flocker/acceptance/endtoend/test_diagnostics.py
+++ b/flocker/acceptance/endtoend/test_diagnostics.py
@@ -20,7 +20,7 @@ class DiagnosticsTests(AsyncTestCase):
     Tests for ``flocker-diagnostics``.
     """
     # This only requires the container agent to check
-    # that it's log is collected. We still care about
+    # that its log is collected. We still care about
     # that working, so we run it. We should stop
     # running it for this test when we get closer
     # to never running it in production.
--- a/flocker/acceptance/endtoend/test_diagnostics.py
+++ b/flocker/acceptance/endtoend/test_diagnostics.py
@@ -19,7 +19,12 @@ class DiagnosticsTests(AsyncTestCase):
     """
     Tests for ``flocker-diagnostics``.
     """
-    @require_cluster(1)
+    # This only requires the container agent to check
+    # that it's log is collected. We still care about
+    # that working, so we run it. We should stop
+    # running it for this test when we get closer
+    # to never running it in production.
+    @require_cluster(1, require_container_agent=True)
     def test_export(self, cluster):
         """
         ``flocker-diagnostics`` creates an archive of all Flocker service logs
--- a/flocker/acceptance/integration/testtools.py
+++ b/flocker/acceptance/integration/testtools.py
@@ -62,7 +62,9 @@ def make_dataset_integration_testcase(image_name, volume_path, internal_port,
                     cluster.remove_container, name))
             return created
 
-        @require_cluster(1)
+        # TODO: this test don't actually require the container agent, it just
+        # uses it do to the setup. It should be ported to the docker API.
+        @require_cluster(1, require_container_agent=True)
         def test_start(self, cluster):
             """
             The specified application can be started with a Docker dataset
@@ -86,7 +88,9 @@ def make_dataset_integration_testcase(image_name, volume_path, internal_port,
                 lambda _: assert_inserted(self, host, port))
             return creating_dataset
 
-        @require_cluster(1)
+        # TODO: this test don't actually require the container agent, it just
+        # uses it do to the setup. It should be ported to the docker API.
+        @require_cluster(1, require_container_agent=True)
         def test_restart(self, cluster):
             """
             The specified application can be started with a Docker dataset
--- a/flocker/acceptance/integration/testtools.py
+++ b/flocker/acceptance/integration/testtools.py
@@ -62,7 +62,7 @@ def make_dataset_integration_testcase(image_name, volume_path, internal_port,
                     cluster.remove_container, name))
             return created
 
-        # TODO: this test don't actually require the container agent, it just
+        # TODO: this test doesn't actually require the container agent, it just
         # uses it do to the setup. It should be ported to the docker API.
         @require_cluster(1, require_container_agent=True)
         def test_start(self, cluster):
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -18,12 +18,14 @@ import ssl
 
 from docker.tls import TLSConfig
 
+from twisted.internet import defer
 from twisted.web.http import OK, CREATED
 from twisted.python.filepath import FilePath
 from twisted.python.constants import Names, NamedConstant
 from twisted.python.procutils import which
 from twisted.internet import reactor
 from twisted.internet.error import ProcessTerminated
+from twisted.internet.task import deferLater
 
 from eliot import start_action, Message, write_failure
 from eliot.twisted import DeferredContext
@@ -1029,7 +1031,8 @@ def _get_test_cluster(reactor):
     )
 
 
-def require_cluster(num_nodes, required_backend=None):
+def require_cluster(num_nodes, required_backend=None,
+                    require_container_agent=False):
     """
     A decorator which will call the supplied test_method when a cluster with
     the required number of nodes is available.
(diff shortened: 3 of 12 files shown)


jml:
Hey Itamar,

Making container agent optional seems like a good next step. 

Not changing the documentation in this change also seems like the right decision.

I have to confess that I don't really understand how the changes in this patch begin to make that happen. The changed methods both refer to leases as being a better solution, but the documentation in those methods implies that they don't currently use leases even though the code does.

We should have some plan for testing this change beyond our test suite, I think. For example, will this allow a cluster that has a container agent on some nodes and not others? If we have a cluster that _does_ have container agents, and one of those drops off somehow (or becomes very slow), would that trigger fallback behaviour that would be subtly incompatible? What if the dropped container agent comes back?

I don't feel that this is a great design review, so I'm more than happy to talk about it further, or for you to seek out someone else.

jml

jml:
Notes from conversation w/ Itamar:

* Currently dataset agent will not do things if there's ignorance of applications (`applications is None`), which is what happens when the container agent isn't running
* This design makes ignorance of applications equivalent to _no_ applications 
* If a container agent goes down for less than two minutes, nothing bad will happen, because we'll rely on the control service cache 
* If a container agent goes down for longer, then it is possible for the volume to get moved out from under a running container if there are no open file descriptors (here we rely on Linux)
* If there are no open file descriptors, it's possible the container could open a new file and have it write to the local directory, rather than the mount. [FLOC-3254](https://clusterhq.atlassian.net/browse/FLOC-3254) addresses this.
* Docker plugin doesn't use leases, which causes similar problems, FLOC-3254 mitigates these too
* If the container agent comes back, and we've managed to delete or move the dataset, then the container agent will stop the container
* If when launching a cluster for the first time the dataset agents start well before the container agents, then behaviour after this change will be equivalent to behaviour now.

james-w:
Hi,

My proposed changes to complete this are now ready for review.

  * There is one XXX that needs to be removed before merging. Either becasue the code is right, or because the code has been changed to be right.
  * I will file an issue for the tests that should be ported to the docker client if the approach is agreed upon.
  * I deferred any changes that would affect non-test clusters (e.g. not starting the container agent by default.) That can be done as a followup.
  * The test isolation isn't great, adding a reboot to the acceptance test run isn't great. However the cost of doing it another way seems fairly high. I'm happy to reconsider though. In addition these changes actually improve test isolation from what we had before. If a test stopped the container agent and then timed out then nothing would ever start it again.

james-w:
I think there are real test failures on the current branch, so you might want to hold off on the review
until they are fixed, but please tell me about any problems you see if you do look.

mdevilliers:
If you want to reboot the test node we will have to revert a change weade to the Jenkins plugin to tell AWS to terminate the node on user initiated shutdown.

Ping me to get it done

james-w:
Hi @mdevilliers this doesn't reboot the Jenkins slave, but one of the instances that it spawns,
so we'll only need to make that change if it also changed the acceptance test nodes, not
just the Jenkins slaves.

flocker/node/_deploy.py
```diff
@@ -135,6 +129,8 @@ def __init__(self, node_uuid, local_applications, leases):
         :param applications: Applications running on the node.
         :param Leases leases: The current leases on datasets.
         """
+        if local_applications is None:
+            local_applications = []
```
jml: You might have told me before, in which case I'm sorry for asking again. How is this change related to making container API optional?

flocker/acceptance/testtools.py
```diff
@@ -1091,6 +1108,100 @@ def clean(cluster):
...
+    """
+    Ensure the container agent is enabled/disabled as specified.
+
+    Doesn't make any changes if the agent is already in the
+    desired state.
+
+    :param Node node: the node on which to ensure the container
+        agent's state
+    :param bool to_enable: True to ensure the container agent
+        is enabled and running, false to ensure the opposite.
```
jml: Nit: `False`

flocker/acceptance/testtools.py
```diff
@@ -1091,6 +1108,100 @@ def clean(cluster):
...
+            # errback chain.
+            command = [b'pidof', b'-x', b'flocker-dataset-agent']
+            d = node.run_as_root(command)
+
+            def not_existing(failure):
+                failure.trap(ProcessTerminated)
+                return False
+            d.addCallbacks(lambda result: True, not_existing)
+            return d
+        d.addCallback(lambda _: loop_until(reactor, dataset_agent_running))
```
jml: :question: Maybe factor out this code, since it's quite similar to that which checks if the container agent is running.

flocker/acceptance/testtools.py
```diff
@@ -1091,6 +1108,100 @@ def clean(cluster):
...
+        the container agent is in the desired state.
+    """
+    if enabled:
+        d = node.run_script("enable_service", "flocker-container-agent")
+    else:
+        d = node.run_script("disable_service", "flocker-container-agent")
+    # If the agent was disabled We have to reboot to clear the control cache.
+    # If we want to avoid the reboot we could add an API to do this.
+    # XXX: does this need to reboot the control node? the node that
+    # changed?
```
jml: You mention this `XXX` in your cover letter, but I'm not clear on what the action / uncertainty is.

jml:
Thanks! I'm glad that you're seeing this through.

I've made some minor comments. I agree with the overall approach. Probably should have another review before it goes in, once tests are fixed & XXXs are resolved.

flocker/node/_deploy.py
james-w: I don't entirely know. This was one of the functional changes in Itamar's
original sketch.

I believe it works because an absent container agent will mean
this function gets called with None. This then treats that as []
so it carries on as if there are no containers.

flocker/acceptance/testtools.py
james-w: This reboots the node on which the container agent was disabled. Is that the right node
to reboot? Should we instead reboot the node running the controller? If the aim
is to clear the controller cache then it seems like it should be the latter, but I
don't know.

itamarst: It should reboot node that container agent is on. We have concept of eras, rebooting gives you a new era so for practical purposes the rebooted node is a completely new node with no cached history (even though the old data is still in control service, it's not used.)

james-w:
This has had some fairly substantial changes to pass all the tests, so should be re-reviewed.

Also it's my first real flocker change, so I think it deserves some extra scrutiny.

flocker/acceptance/endtoend/test_diagnostics.py
```diff
@@ -19,7 +19,12 @@ class DiagnosticsTests(AsyncTestCase):
     """
     Tests for ``flocker-diagnostics``.
     """
-    @require_cluster(1)
+    # This only requires the container agent to check
+    # that it's log is collected. We still care about
```
jml: nit: its

flocker/acceptance/integration/testtools.py
```diff
@@ -62,7 +62,9 @@ def _start_container(self, name, dataset_id, external_port, cluster,
                     cluster.remove_container, name))
             return created
 
-        @require_cluster(1)
+        # TODO: this test don't actually require the container agent, it just
```
jml: nit: doesn't.

jml:
Sorry, got interrupted mid-review and didn't resume. These changes LGTM. Please resolve conflicts, get tests to pass to your satisfaction, then merge.

james-w:
This is green now, going to merge it.

PR #2273, opened 2015-12-02
[FLOC-3603] Get rid of `FakeBlockDeviceDeployerLocalState`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3603

Also, get rid of some indirection in constructing local state.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1177,144 +1197,124 @@ def _infer_volumes_for_test(node_uuid, node_hostname, compute_instance_id,
...
+            datasets[dataset_id] = DiscoveredDataset(
+                state=DatasetStates.MOUNTED,
+                dataset_id=dataset_id,
+                maximum_size=manifestation.dataset.maximum_size or 0,
+                device_path=node_state.devices[dataset_id],
+                blockdevice_id=_create_blockdevice_id_for_test(dataset_id),
+                mount_point=node_state.paths[unicode(dataset_id)],
+            )
 
+    return BlockDeviceDeployerLocalState(
```
itamarst: Perhaps as a sanity check we could have an `assert` here that the resulting object's `shared_state_changes()` match the input ones?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1177,144 +1197,124 @@ def _infer_volumes_for_test(node_uuid, node_hostname, compute_instance_id,
...
-                uuid=self.node_uuid,
-                hostname=self.hostname,
-                manifestations={},
-                paths={},
-                devices={},
-                applications=None
+DISCOVERED_DATASET_STRATEGY = tagged_union_strategy(
+    DiscoveredDataset,
+    {
+        'dataset_id': uuids(),
```
itamarst: Doesn't this need a state? And... it's only creating MOUNTED ones? why is that? At least document comment explaining why.

itamarst:
This all vaguely makes sense to me and seems like an improvement, especially insofar as we have sanity checks and tests for the things that construct data. So proceed and and merge when green.

flocker/node/agents/test/test_blockdevice.py
tomprince: `tagged_union_strategy` is a strategy that will generate things from all possible states. You just need to pass the possible attributes, and it will generate possibilities from there.

tomprince:
Since this is only test changes, and all the changed tests pass, I'm merging this, despite unrelated failures.

PR #2274, opened 2015-12-02
[FLOC-3605] Test against smaller volumes on Rackspace

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3605

The lower bound used to be 100GiB but now it is 75GiB.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -30,7 +30,7 @@ from effect import parallel
 from txeffect import perform
 
 from admin.vagrant import vagrant_version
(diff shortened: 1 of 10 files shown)


mpgomez:
Looks good to me. Is that going to fix all the red we have in Jenkins in Rackspace? :DDD

exarkun:
> Looks good to me. Is that going to fix all the red we have in Jenkins in Rackspace? :DDD

Ha ha. :smile:

PR #2077, opened 2015-10-26
[FLOC-3248] Encode args as well as kwargs 

jml:
Applies almost all of the suggestions from #2052 and [FLOC-3248](https://clusterhq.atlassian.net/browse/FLOC-3248)

* uses `wire_encode` instead of `json.dumps`
* catches `ValueError`, which can be raised on circular references and also invalid dates
* checks `args` for serializability as well as kwargs

Does this by extracting an `_ensure_encodeable` function, which in turn allows for a more functional implementation within `log_method`.

Did not implement the "force all arguments to be keyword arguments" suggestion. That can be done as a separate clean-up ticket, outside of the Green CI epic.

[Review on Reviewable](https://reviewable.io/reviews/clusterhq/flocker/2077)

PR #2276, opened 2015-12-02
[FLOC-2997] Log structured information rather than strings for configuration and state

itamarst:
Structured information is easier to pretty print, e.g. with json formatting tools or `eliot-prettyprint` (the latter requires some follow up coding though).

### DIFF
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -213,7 +213,18 @@ def wire_decode(data):
     return loads(data, object_hook=decode)
 
 
-_DEPLOYMENT_FIELD = Field(u"configuration", repr)
+def to_unserialized_json(obj):
+    """
+    Convert a wire encodeable object into structured Python objects that
+    are JSON serializable.
+
+    :param obj: An object that can be passed to ``wire_encode``.
+    :return: Python object that can be JSON serialized.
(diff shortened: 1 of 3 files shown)


jml:
The code is clear & this is an incremental improvement, so happy for this to land once you've made peace with the tests. 

FWIW, I think the "right" approach to this is to use an adapter registry.

jml:
I just remembered this: e1a0800e3e7ee6cfda2d5dea5a543c57255acd1a which is part of https://github.com/ClusterHQ/flocker/pull/2077 

In it, I explicitly moved away from `wire_encode` to `json.dumps` because eliot had problems serializing the results of `wire_encode`. 

I don't know if this is relevant to your patch, but worth flagging in case it saves you some time.

itamarst:
Eliot just takes anything JSON-encodable. `wire_encode` was the problem in that patch: `wire_encode` doesn't take any arbitrary object, it's only designed for things you can send over AMP. So I only used it for those things in this patch.

As far as registry, yeah, I'm starting to think that's the way to go; I just added support for exception serialization adapters in latest Eliot. https://github.com/ClusterHQ/eliot/issues/224 is the relevant issue for non-exceptions.

PR #2277, opened 2015-12-03
[FLOC-3608] Replace ``PRecord`` with ``PClass``.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3608

There are a couple of references to `PRecord` left in code that deals explicitly with both (persistence).

Also, `NodeState` was left as a `PRecord`, since `.items()` is called on it.

### DIFF
--- a/flocker/node/test/test_deploy.py
+++ b/flocker/node/test/test_deploy.py
@@ -519,7 +519,10 @@ class StartApplicationTests(SynchronousTestCase):
         StartApplication(application=application,
                          node_state=EMPTY_NODESTATE).run(deployer)
 
-        self.assertEqual(policy, RestartNever())
+        [unit] = self.successResultOf(fake_docker.list())
+        self.assertEqual(
+            unit.restart_policy,
+            RestartNever())
 
     def test_command_line(self):
         """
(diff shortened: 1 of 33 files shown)


flocker/node/test/test_deploy.py
```diff
@@ -1038,7 +1041,7 @@ def test_discover_manifestation_with_size(self):
         )
         d = api.discover_state(self.EMPTY_NODESTATE)
 
-        self.assertItemsEqual(
```
jml: The changes in this file look unrelated to the purpose of the patch. Given the scope of the patch, I'd strongly recommend making this change separately.

jml:
Thanks Tom, it's great to see this finally happen!

One comment about a change that looks unrelated. Please back that out & merge the rest once the tests pass / fail reasonably.

flocker/node/test/test_deploy.py
tomprince: This change is necessary, since `Manifestation` was a mapping (so `assertItemsEqual` would do something with it) but now it isn't. That is, the tests started failing with the change to a `PClass` here.

jml: How do the other changes (`RestartNever`, `self.DATASET_ID`) relate to the `PClass` change?

tomprince: - `assertItemsEqual` actually compares the keys of dictionaries, not the contents (?!?) so the test was incorrect (i.e. comparing against a manifestation with the wrong dataset id.
- `RestartNever` compared equal to `RestartAlways` (since they both had no fields). The test was trying to say that `StartApplication` always passed `RestartNever` to the docker client, but instead just said that the policy defined above (which was `RestartAlways`) equaled `RestartNever`. The mismatch was noticed because they did in fact compare equal.

PR #2278, opened 2015-12-03
[FLOC-3606] Cleanup volume created by test_destroy_timesout

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3606

The reason it didn't work before is that the directly-created volume didn't appear to belong to the `IBlockDeviceAPI` implementation being used for cleanup and so it wasn't cleaned up.

You can verify this is fixed by running the test and then looking for the created blockdevice_id in the test log and then using the Rackspace Storage web UI to search for it.  On master, you'll always find it.  With the branch, you never should.

Downside is that the test takes twice as long now because it actually has to destroy the volume. :wink: 

Perhaps also noteworthy is that this change should also cause the periodic batch cleanup to be capable of cleaning up this volume should the test runner itself somehow fail to clean it up (because the destroy operation hits a network hiccup or because the test suite crashes before cleanup can be run or whatever).  Since the `IBlockDeviceAPI` implementation is now creating the volume, it has metadata associating it with a cluster that can be recognized as belonging to a test suite run.  That should be sufficient to cause the batch cleanup job to recognize it.

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -40,7 +40,7 @@ from ..test.blockdevicefactory import (
     get_blockdeviceapi_with_cleanup, get_device_allocation_unit,
     get_minimum_allocatable_size, get_openstack_region_for_test,
 )
-from ....testtools import flaky, run_process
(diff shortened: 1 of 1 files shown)


jml:
Looks good, thanks!

PR #2281, opened 2015-12-03
Eliot contexts now preserved across threads using our auto-threading support

itamarst:
This should allow us to see in the logs the logged Cinder or EBS operations as children of the IStateChange running in the main Twisted thread.

### DIFF
--- a/flocker/common/_thread.py
+++ b/flocker/common/_thread.py
@@ -6,7 +6,41 @@ Some thread-related tools.
 
 from twisted.internet.threads import deferToThreadPool
 
+from eliot import Action
+
 from ._interface import interface_decorator
+
+
+# This will be moved into Eliot:
+# https://github.com/ClusterHQ/eliot/issues/225
+def preserve_context(f, *args, **kwargs):
+    """
+    Package up the given function and arguments with the current Eliot
+    context, and then restore context and call given function when the
+    resulting callable is run.
+
+    The result should only be used once, presumably in a thread.
+
+    :param action_context: Current Eliot context.
+    :param f: A callable.
+    :param args: Positional arguments.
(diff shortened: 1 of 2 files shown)


flocker/common/_thread.py
```diff
@@ -6,9 +6,43 @@
...
+    from eliot._action import currentAction
+    action = currentAction()
+    if action is None:
+        return lambda: f(*args, **kwargs)
+    task_id = action.serialize_task_id()
+
+    def restore_eliot_context():
+        with Action.continue_task(task_id=task_id):
+            return f(*args, **kwargs)
+    return restore_eliot_context
```
jml: :question: Maybe implement this as a decorator, rather than taking `*args` and `**kwargs`? 

It (subjectively) feels more surprising to capture the arguments and return a nullary callable than to take a function and return a new function with the same arity.

jml:
Looks good. I've got an optional suggestion which if followed would produce a non-trivial diff, but mostly mechanical, so no need for follow-up review.

flocker/common/_thread.py
itamarst: A decorator doesn't work because the result will get called in the thread, by which point you don't have access to the action anymore.

itamarst: Unless I misunderstood the suggestion. Could you show me a code sample?

jml: Not with any promptness. Feel free to go ahead as it is & I'll try to look at it later.

jml: I told a lie. Here's a diff: https://gist.github.com/jml/bdf471879f8affbc2055. It passes tests.

itamarst: Oh, I see. Yeah that seems better.

PR #2283, opened 2015-12-04
[FLOC-2889] Don't assign sdX when the equivalent xvdX is already in use

wallrj:
[dwgebler comment, I took over this branch]
Fix for FLOC-2889.

I've switched `_next_device` to examine local state only, which basically assumes we're never trying to attach a volume to a different instance ID than the one this backend is running on. This is hopefully a safe assumption, else we still have an issue that we can't reliably map EBS devices to OS devices and I'm not sure what we can do about that. (actually, we're assuming a straightforward `/dev/sd*` - `/dev/xvd*` mapping in either case, but local state at least prevents the `AttachedUnexpectedDevice` error when non-Flocker volumes are present on an instance)

The problem as described by @wallrj at https://clusterhq.atlassian.net/browse/FLOC-2889?focusedCommentId=21035&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-21035 can be reproduced against `master` by attaching a volume with a `/dev/xv*` device name, then attempting to create and attach a volume via the Flocker EBS driver (using a `/dev/sd*` device name). Worth noting though, that while this is **a** way of reproducing the described issue, I don't know that it's the **only** circumstance in which it happens. The explanation seems consistent with observed behaviour.


flocker/node/agents/functional/test_ebs.py
```diff
@@ -68,6 +68,76 @@ class EBSBlockDeviceAPIInterfaceTests(
     """
     Interface adherence Tests for ``EBSBlockDeviceAPI``.
     """
+    def test_attach_when_foreign_device_has_next_device(self):
```
itamarst: This test seems far far too brittle and tied to implementation details. It would be better to just use public interfaces to recreate this:

1. Create volume with EC2 APIs directly.
2. Create volume with Flocker API, attempt to attach it, see that it fails. Or... doesn't fail, in which case diagnosis of problem is wrong or needs refining.

flocker/node/agents/ebs.py
```diff
@@ -875,6 +875,7 @@ def _next_device(self, instance_id, volumes, devices_in_use):
         IN_USE_DEVICES(devices=sorted_devices).write()
 
         for suffix in b"fghijklmonp":
+            # check for /dev/xvdX here too.
```
itamarst: Well... maybe? That might be the issue, but... we already presume that the device names are *always* mapped. So perhaps a better way to describing this issue is that this method mixes local devices with untranslated EC2-reported devices. Which is definitely a bug, it seems like the retry logic is wrong somehow since "ignored devices" is never actually useful.

But... we're getting all devices from EC2. Just because a third party attached one doesn't mean this should fail. It'll only fail if a third party added it recently enough that we hit eventual consistency issues (or in between list and attach). Is that what actually happened in the case you saw?

Since I believe we're fundamentally always assuming a mapping from EC2-reported `/dev/sd*` to local `/dev/xvd*`, perhaps the correct answer is to just look at local devices, filter to `/dev/xvd*` and only use those?

itamarst: I.e. ignore EC2 reporting altogether and just look at local state.

itamarst:
Besides the changes to design suggested above (which include a reproducer test that isn't so brittle so we can be sure we've actually diagnosed correctly), this will likely conflict with the boto3 port Dave is working on if you continue to use EC2 calls. Which I guess the test will need to. So maybe return to this when he's done.

So:
1. Re-implement test using public APIs, see if you can still reproduce, and explain better what happened.
2. Do actual fix implementation once FLOC-3508 is merged.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -65,6 +65,63 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+        _wait_for_volume_state_change(VolumeOperations.ATTACH, created_volume)
+
+        # Now create a volume via Flocker EBS backend.
+        blockdevice_volume = self.api.create_volume(
+            dataset_id=dataset_id, size=ONE_GIB)
+
+        flocker_volume = ec2_client.connection.Volume(
+            blockdevice_volume.blockdevice_id)
+        self.addCleanup(clean_volume, flocker_volume)
+        _wait_for_volume_state_change(VolumeOperations.CREATE, flocker_volume)
```
itamarst: `IBlockDeviceAPI.create_volume` is supposed to block until the creation is done, so why do you need this wait?

flocker/node/agents/functional/test_ebs.py
```diff
@@ -65,6 +65,63 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+
+        # Attach the blockdevice volume to this instance.
+        # The assertion in this test is that this operation does not
+        # raise an ``AttachedUnexpectedDevice`` error.
+        self.api.attach_volume(
+            blockdevice_volume.blockdevice_id, instance_id)
+        _wait_for_volume_state_change(VolumeOperations.ATTACH, flocker_volume)
```
itamarst: `IBlockDeviceAPI.attach_volume` is supposed to block until the attach is done, so why do you need this wait?

flocker/node/agents/functional/test_ebs.py
```diff
@@ -65,6 +65,63 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+            raise SkipTest(str(e))
+
+        ONE_GIB = 1073741824
+        dataset_id = uuid4()
+        ec2_client = get_ec2_client_for_test(config)
+        meta_client = ec2_client.connection.meta.client
+
+        # Create a volume directly using boto.
+        requested_volume = meta_client.create_volume(
+            Size=int(Byte(ONE_GIB).to_GiB().value),
```
itamarst: or just `Size=1`

flocker/node/agents/ebs.py
```diff
@@ -1022,20 +1016,30 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
         sorted_devices = sorted(list(thaw(devices)))
         IN_USE_DEVICES(devices=sorted_devices).write()
 
         for suffix in b"fghijklmonp":
+            next_local_device = b'xvd' + suffix
+            next_local_sd_device = b'sd' + suffix
             file_name = u'/dev/sd' + suffix
-            if file_name not in devices:
+            possible_devices = [
+                next_local_device, next_local_sd_device, file_name
```
itamarst: We really shouldn't have `possible_devices` contain all three of `/dev/sda`, `xvda` and `sda`. We should pick one and stick to it.

flocker/node/agents/ebs.py
```diff
@@ -1181,8 +1185,7 @@ def attach_volume(self, blockdevice_id, attach_to):
         ignore_devices = pset([])
         for attach_attempt in range(3):
```
itamarst: If we're going to limit ourselves to attaching to local machine (a reasonable assumption for now) then we should throw an exception if `blockdevice_id != self.compute_instance_id()` saying attaching elsewhere is unsupported. Otherwise when the assumption is broken we'll get some weird behavior.

flocker/node/agents/ebs.py
```diff
@@ -1022,20 +1016,30 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
-            for attachment in volume_attachments:
-                if attachment['InstanceId'] == instance_id:
-                    volume_devices.append(attachment['Device'])
-        devices = pset(volume_devices)
-        devices = devices | devices_in_use
+
+        command = [b"/bin/lsblk", b"--output", b"KNAME"]
+        command_result = check_output(command)
+        local_devices = pset(filter(
+            lambda d: d.startswith(b"xvd") or d.startswith('sd'),
```
itamarst: I think we're pretty much stuck assuming Linux *always* maps `sd` -> `xvd`.

flocker/node/agents/functional/test_ebs.py
itamarst: Above comment was addressed.

itamarst:
Glad you were able to reproduce a particular issue, and write a test for it.

Please address comments above and resubmit. In particular, I think we should just standardize on a particular device name (`/dev/sda` seems the reasonable default) and then convert to it throughout in this particular calculation, instead of mixing different kinds in same data structure.

flocker/node/agents/ebs.py
```diff
@@ -997,23 +1019,16 @@ def _attach_ebs_volume(self, volume_id, instance_id, device):
...
+        Devices available for EBS volume usage are ``/dev/sd[f-p]``.
+        Find the first device from this set that is currently not
+        in use.
         XXX: Handle lack of free devices in ``/dev/sd[f-p]`` range
         (see https://clusterhq.atlassian.net/browse/FLOC-1887).
 
-        :param unicode instance_id: EC2 instance ID.
-        :param volumes: Collection of currently known
-            ``BlockDeviceVolume`` instances.
         :param set devices_in_use: Unicode names of devices that are
```
itamarst: The passed in value is a set of devices in the form `"/dev/sda"`. Which doesn't match the internally used format of `"sdb"`.

1. This suggests insufficient test coverage.
2. However... the reason we do the ignore thing is in case we attach and it fails. Since we're looking at all local devices in this code path it seems like we don't need to know that an attach failed because of an attached volume, we'll see it next time we do `lsblk`. So maybe this parameter can just be dropped?

flocker/node/agents/ebs.py
```diff
@@ -1022,20 +1037,27 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
-            for attachment in volume_attachments:
-                if attachment['InstanceId'] == instance_id:
-                    volume_devices.append(attachment['Device'])
-        devices = pset(volume_devices)
-        devices = devices | devices_in_use
-        sorted_devices = sorted(list(thaw(devices)))
+
+        command = [b"/bin/lsblk", b"--output", b"KNAME"]
+        command_result = check_output(command)
+        local_devices = pset(filter(
```
itamarst: Add comment with format of output.

flocker/node/agents/ebs.py
```diff
@@ -1022,20 +1037,27 @@ def _next_device(self, instance_id, volumes, devices_in_use):
...
+            lambda d: d.startswith(b"xvd") or d.startswith('sd'),
+            command_result.split("\n")[1:]
+        ))
+        existing_devices = local_devices | devices_in_use
+        sorted_devices = sorted(list(thaw(existing_devices)))
         IN_USE_DEVICES(devices=sorted_devices).write()
 
         for suffix in b"fghijklmonp":
+            next_local_device = b'xvd' + suffix
+            next_local_sd_device = b'sd' + suffix
```
itamarst: Minor nit : we're mixing unicode and bytes in line below.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -65,6 +67,84 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+        try:
+            config = get_blockdevice_config(ProviderType.aws)
+        except InvalidConfig as e:
+            raise SkipTest(str(e))
+        ec2_client = get_ec2_client_for_test(config)
+        dataset_id = uuid4()
+        volume = self.api.create_volume(dataset_id=dataset_id, size=ONE_GIB)
+
+        flocker_volume = ec2_client.connection.Volume(volume.blockdevice_id)
+        self.addCleanup(flocker_volume.delete)
```
itamarst: Seems better to do cleanup using the API object, in case it needs to do something else we don't know about.

itamarst:
One final change, hopefully: address and resubmit.

flocker/node/agents/ebs.py
dwgebler: Hmm, yes that's correct, these come in from the previous returned results of `_next_device` if an attach failed. So yes, I agree with your logic, we can drop this parameter entirely and the use of `ignored_devices` in the `attach_volume` method. Intriguingly, it looks to me like `ignored_devices` might not have worked properly even prior to this branch.

itamarst:
Great, merge when Jenkins is happy.

PR #2287, opened 2015-12-04
Ebs use boto3 floc 3508

dwgebler:


flocker/node/agents/ebs.py
```diff
@@ -244,6 +256,22 @@ class AttachFailed(Exception):
...
+    """
+    A volume is not in an appropriate state to perform an operation.
+
+    :param boto3.resources.factory.ec2.Volume volume: The volume.
+    :param str state: The known volume state at the time of the exception.
+    :param list valid_states: A ``list`` of ``str`` representing states
+        that would have been valid for this operation.
+    """
+    def __init__(self, volume, state, valid_states):
+        Exception.__init__(self)
```
itamarst: Please do `Exception.__init__(self, volume, state, valid_states)` so that the `reason` field in Eliot logs has this information in it.

flocker/node/agents/ebs.py
```diff
@@ -306,13 +330,9 @@ def _enable_boto_logging():
...
     Make boto log activity using Eliot.
     """
-    logger = logging.getLogger("boto")
-    logger.addHandler(EliotLogHandler())
-
-    # It seems as though basically all boto log messages are at the same
-    # level.  Either we can see all of them or we can see none of them.
-    # We'll do some extra filtering in the handler.
+    logger = logging.getLogger("boto3")
     logger.setLevel(logging.DEBUG)
```
itamarst: Given we're only logging `INFO` messages, presumably we don't need to set level to DEBUG but rather to INFO?

flocker/node/agents/ebs.py
```diff
@@ -390,122 +410,221 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
-    return _EC2(zone=zone,
-                connection=_LoggedBotoConnection(connection=connection))
-
-
-def _boto_logged_method(method_name, original_name):
+    connection = boto3.session.Session(
+        aws_access_key_id=access_key_id,
+        aws_secret_access_key=secret_access_key
+    )
+    connection._session.set_config_variable(
```
itamarst: 1. The deleted comment about BOTO_NUM_RETRIES seems informative and should be preserved.
2. Is there no metadata service retry thingie anymore?

dwgebler: `num_retries` doesn't exist as a `botocore` config option and I can't find any documented equivalent. I have a question on Stack Overflow open about that.

flocker/node/agents/ebs.py
```diff
@@ -390,122 +410,221 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
+    def _run_with_logging(*args, **kwargs):
         """
-        Run given boto.ec2.connection.EC2Connection method with exception
-        logging for ``EC2ResponseError``.
+        Run given boto3.ec2.ServiceResource method with exception
+        logging for ``ClientError``.
         """
-        original = getattr(self, original_name)
-        method = getattr(original, method_name)
+        def log_boto3_error(e):
```
itamarst: When an exception is raised inside an Eliot action context and not caught the exception is automatically logged. And in latest version of Eliot, as used by Flocker, you can register additional field extraction for specific exceptions.  So:

1. On module level register an exception extractor that extracts those fields from `ClientError` (see http://eliot.readthedocs.org/en/0.10.1/generating/errors.html#custom-exception-logging for details).
2. You can then get rid of `BOTO_EC2RESPONSE_ERROR` and the need to catch and reraise `ClientError`.

flocker/node/agents/ebs.py
```diff
@@ -390,122 +410,221 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
-        representing connection to an EC2 instance.
+    :ivar boto3.resources.factory.ec2.ServiceResource: Object
+        representing an EC2 resource.
     """
     zone = field(mandatory=True)
     connection = field(mandatory=True)
 
+    def get_volume_tag(self, volume, name):
+        """
+        Retreive the tag from the specified volume with the specified name.
```
itamarst: Typo: "retreive"

flocker/node/agents/ebs.py
```diff
@@ -390,122 +410,221 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
-
-    :ivar boto.ec2.connection.EC2Connection connection: Object
-        representing connection to an EC2 instance with logged
-        ``EC2ConnectionError``.
-    """
-    connection = field(mandatory=True)
-
-
-class _EC2(PClass):
+class EC2Client(PClass):
```
itamarst: Is this class useful at all, as opposed to just having the EBS interactions directly within the IBlockDeviceAPI implementation? It's true that the logging tells you something... but it *doesn't* map directly to individual AWS interactions anyway so maybe not as much as we'd like. And "retry on SSL error" can happen half-way through a series of cloud interactions, so the whole *series* is restarted if an *individual* operation fails... which is somewhat unexpected and may lead to bugs.

flocker/node/agents/ebs.py
```diff
@@ -390,122 +410,221 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
 
     :return: Input volume in BlockDeviceVolume format.
     """
+    if ebs_volume.attachments:
+        attached_to = unicode(ebs_volume.attachments[0]['InstanceId'])
+    else:
+        attached_to = None
+
+    volume_dataset_id = None
+    for tag in ebs_volume.tags:
```
itamarst: Isn't there a utlity method for this above?

dwgebler: :) oh yeah, think I wrote this before the utility method

flocker/node/agents/ebs.py
```diff
@@ -907,17 +1037,19 @@ def create_volume_with_profile(self, dataset_id, size, profile_name):
...
             # creation with default profile.
             # Compliance violation of the volume's requested profile
             # will be detected and remediated in a future release (FLOC-3275).
-            if e.code != INVALID_PARAMETER_VALUE:
+            if e.response['Error']['Code'] != INVALID_PARAMETER_VALUE:
                 raise e
-            CREATE_VOLUME_FAILURE(dataset_id=dataset_id,
-                                  aws_code=e.code,
-                                  aws_message=e.message).write()
+            CREATE_VOLUME_FAILURE(
```
itamarst: This logging shouldn't be necessary with exception field extraction, since presumably higher-level context includes the dataset ID and the fields will be handled by auto-extraction.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -352,7 +357,7 @@ def _create_template_ebs_volume(self, operation):
         :returns: Suitable volume in the right start state for input operation.
         :rtype: boto.ec2.volume.Volume
```
itamarst: No longer correct.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -315,11 +312,19 @@ class EBSProfiledBlockDeviceAPIInterfaceTests(
     pass
 
 
+class VolumeStub(object):
```
itamarst: Even if they're immutable, can you just *create* the real thing with the appropriate attributes? Creating and then setting attributes is kinda meh anyway, but with completely custom class it's even less useful  as a realistic test.

flocker/node/agents/ebs.py
itamarst: `max_attempts` maybe? https://github.com/boto/botocore/blob/082a8b8e564b12500ff3de4d048f89d56d33d67d/botocore/retryhandler.py#L98

flocker/node/agents/functional/test_ebs.py
dwgebler: You can create a ``Volume`` with an ``id`` before calling out to AWS, but you can't instantiate with any other attributes.

dwgebler: This at least allows the state table tests to call the real implemented methods on stub objects that can have their state manipulated.

flocker/node/agents/ebs.py
dwgebler: Ah, good spot, I'll give that a go :)

itamarst:
There's a bunch of comments to address. I think the additional intermediate client is unnecessary, and that the SSL error retry logic is a potential source of bugs.

As far as the SSL error, I think we should just prevent it from happening. https://github.com/boto/boto3/issues/220 is likely the issue. Options include:

1. Disable persistent connections. This will slow operations down somewhat.
2. Decrease timeout on pooled HTTP connections so they time out faster.
3. Trick urllib3 into not using pyOpenSSL.

itamarst:
Thanks for figuring all this out. Please address the above and resubmit. I'm happy to talk about details of implementing these changes.

dwgebler:
Ready for review again. Removing pyOpenSSL from urllib3 has solved the problem of `OpenSSL.SSL.Error` coming up when we re-connect.

flocker/node/agents/ebs.py
```diff
@@ -289,16 +334,12 @@ def __init__(self, blockdevice_id, operation,
...
 class EliotLogHandler(logging.Handler):
-    # Whitelist ``"msg": "Params:`` field for logging.
-    _to_log = {"Params"}
-
     def emit(self, record):
-        fields = vars(record)
         # Only log certain things.  The log is massively too verbose
         # otherwise.
-        if fields.get("msg", ":").split(":")[0] in self._to_log:
+        if record.levelname == 'INFO':
```
itamarst: This means you won't log things with e.g. ERROR level. Since you already limit to INFO or higher below you can just omit this if and log everything.

flocker/node/agents/ebs.py
itamarst: You discovered max_attemps is unnecessary or has a good limit already?

flocker/node/agents/ebs.py
```diff
@@ -390,103 +427,64 @@ def ec2_client(region, zone, access_key_id, secret_access_key):
...
-        representing connection to an EC2 instance with logged
-        ``EC2ConnectionError``.
+    :return: A ``str`` representing the value of the tag, or ``None`` if
+        no tag with this name can be found in the volume.
     """
-    connection = field(mandatory=True)
+    for tag in volume.tags:
+        if tag['Key'] == name:
+            return tag['Value']
+    return None
```
itamarst: It's better to raise an exception rather than return `None`. Probably custom `NoSuchTag` exception. That way if e.g. there's a bug in the code or unexpected edge case where tag is missing then you get a meaningful exception instead of getting (five lines later) a TypeError about how `UUID` can't deal with `None` or errors like that.

flocker/node/agents/ebs.py
dwgebler: `max_attempts` can't be configured this way, however the default limit is 5 retries with an exponential backoff between tries. This may be sufficient, or if not, more research is needed in how to set the retry config dynamically.

itamarst: OK, we can probably live with that.

itamarst:
Thanks for getting this done!

1. Address comments above.
2. There's a bunch of acceptance test failures. We really want AWS acceptance tests and functional EBS tests passing before this is merged.

Merge when all unit tests builders and acceptance tests and functional tests on EBS are green.

Issue #2284, opened 2015-12-04
UNIXAddress' object has no attribute 'host'


johnjelinek:
```
$ docker version
Client:
 Version:      1.10.0-dev
 API version:  1.22
 Go version:   go1.5.1
 Git commit:   ee3e07d
 Built:        Thu Dec  3 00:48:27 2015
 OS/Arch:      linux/amd64

Server:
 Version:      swarm/1.0.0
 API version:  1.21
 Go version:   go1.5.1
 Git commit:   6aa6f76
 Built:
 OS/Arch:      linux/amd64
```

```
$ docker info
Containers: 24
Images: 58
Role: replica
Primary: xxx.xxx.xxx.xxx:3376
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 5
 aboveandbeyond: xxx.xxx.xxx.xxx:12375
  └ Containers: 5
  └ Reserved CPUs: 0 / 7
  └ Reserved Memory: 0 B / 3.371 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-229.20.1.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), provider=generic, storagedriver=devicemapper
 adventureclub: xxx.xxx.xxx.xxx:12375
  └ Containers: 4
  └ Reserved CPUs: 0 / 7
  └ Reserved Memory: 0 B / 3.371 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-229.20.1.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), provider=generic, storagedriver=devicemapper
 knifeparty: xxx.xxx.xxx.xxx:12375
  └ Containers: 4
  └ Reserved CPUs: 0 / 7
  └ Reserved Memory: 0 B / 3.371 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-229.20.1.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), provider=generic, storagedriver=devicemapper
 skrillex: xxx.xxx.xxx.xxx:12375
  └ Containers: 6
  └ Reserved CPUs: 0 / 7
  └ Reserved Memory: 0 B / 3.371 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-229.20.1.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), provider=generic, storagedriver=devicemapper
[...]

PR #2288, opened 2015-12-04
[FLOC-3618] Fix another incompatibility between UNIX sockets and Twisted's HTTP implementation

itamarst:
When the `Host` header was empty or missing the the address is used to calculate the host ... and `UNIXAddress` has no `host` attribute. Apparently Docker 1.10 (dev release) stops sending the `Host` header in HTTP requests, or maybe sends it as empty string, triggering this bug; haven't actually looked at Docker source code.

This addresses user-reported issue #2284.

### DIFF
--- a/flocker/dockerplugin/__init__.py
+++ b/flocker/dockerplugin/__init__.py
@@ -5,3 +5,12 @@ Docker plugin allowing use of Flocker to manage Docker volumes.
 
 This may eventually be a standalone package.
 """
+
+from twisted.internet.address import UNIXAddress
+
+# Many places in both twisted.web and Klein are unhappy with listening on
+# Unix socket, e.g.  https://twistedmatrix.com/trac/ticket/5406.  The
+# Docker plugin needs to listen using a Unix socket, so "fix" that by
+# pretending we have a port number. Yes, I feel guilty.
+UNIXAddress.port = 0
+del UNIXAddress
--- a/flocker/dockerplugin/__init__.py
+++ b/flocker/dockerplugin/__init__.py
@@ -11,6 +11,7 @@ from twisted.internet.address import UNIXAddress
 # Many places in both twisted.web and Klein are unhappy with listening on
 # Unix socket, e.g.  https://twistedmatrix.com/trac/ticket/5406.  The
 # Docker plugin needs to listen using a Unix socket, so "fix" that by
-# pretending we have a port number. Yes, I feel guilty.
+# pretending we have a port number and host. Yes, I feel guilty.
 UNIXAddress.port = 0
+UNIXAddress.host = b"127.0.0.1"
 del UNIXAddress
--- a/flocker/dockerplugin/test/test_api.py
+++ b/flocker/dockerplugin/test/test_api.py
@@ -442,15 +442,15 @@ class APITestsMixin(APIAssertionsMixin):
 
     def test_unsupported_method(self):
         """
-        If an unsupported method is requested the appropriate response code is
-        returned.
+        If an unsupported method is requested the 405 Not Allowed response
+        code is returned.
         """
         return self.assertResponseCode(
             b"BAD_METHOD", b"/VolumeDriver.Path", None, NOT_ALLOWED)
 
     def test_unknown_uri(self):
         """
-        If an unknown URI path is requested the appropriate response code is
+        If an unknown URI path is requested the 404 Not Found response code is
         returned.
         """
         return self.assertResponseCode(
@@ -459,7 +459,8 @@ class APITestsMixin(APIAssertionsMixin):
     def test_empty_host(self):
         """
         If an empty host header is sent to the Docker plugin it does not blow
-        up.
+        up, instead operating normally. E.g. for ``Plugin.Activate`` call
+        returns the ``Implements`` response.
         """
         return self.assertResult(b"POST", b"/Plugin.Activate", 12345, OK,
                                  {u"Implements": [u"VolumeDriver"]},
--- a/flocker/dockerplugin/test/test_api.py
+++ b/flocker/dockerplugin/test/test_api.py
@@ -6,7 +6,7 @@ Tests for the Volumes Plugin API provided by the plugin.
 
 from uuid import uuid4
 
-from twisted.web.http import OK
+from twisted.web.http import OK, NOT_ALLOWED, NOT_FOUND
 from twisted.internet.task import Clock, LoopingCall
 
 from pyrsistent import pmap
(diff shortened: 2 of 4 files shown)


flocker/dockerplugin/__init__.py
```diff
@@ -5,3 +5,13 @@
...
+
+from twisted.internet.address import UNIXAddress
+
+# Many places in both twisted.web and Klein are unhappy with listening on
+# Unix socket, e.g.  https://twistedmatrix.com/trac/ticket/5406.  The
+# Docker plugin needs to listen using a Unix socket, so "fix" that by
+# pretending we have a port number and host. Yes, I feel guilty.
+UNIXAddress.port = 0
+UNIXAddress.host = b"127.0.0.1"
+del UNIXAddress
```
exarkun: I'm actually a little concerned about what effect this might have on the validity of the rest of the Flocker test suite.  This code will be loaded into a `trial flocker` process but not into an *actual* Flocker process.  Are we going to mask other bugs in Flocker with this?

Fortunately I don't think we run any other Flocker services over a UNIX socket ... at the moment ... But if we did, maybe these lines would make it appear as though it works when it really won't.

flocker/dockerplugin/test/test_api.py
```diff
@@ -438,10 +440,34 @@ def error():
             {u"Name": u"whatever"}, 423,
             {u"Err": "no good"})
 
+    def test_unsupported_method(self):
+        """
+        If an unsupported method is requested the appropriate response code is
```
exarkun: Which response code is that?

flocker/dockerplugin/test/test_api.py
```diff
@@ -438,10 +440,34 @@ def error():
...
+        returned.
+        """
+        return self.assertResponseCode(
+            b"BAD_METHOD", b"/VolumeDriver.Path", None, NOT_ALLOWED)
+
+    def test_unknown_uri(self):
+        """
+        If an unknown URI path is requested the appropriate response code is
```
exarkun: Which response code is that?

flocker/dockerplugin/test/test_api.py
```diff
@@ -438,10 +440,34 @@ def error():
...
+        returned.
+        """
+        return self.assertResponseCode(
+            b"BAD_METHOD", b"/xxxnotthere", None, NOT_FOUND)
+
+    def test_empty_host(self):
+        """
+        If an empty host header is sent to the Docker plugin it does not blow
```
exarkun: What *does* it do?

flocker/dockerplugin/test/test_api.py
```diff
@@ -438,10 +440,34 @@ def error():
...
+                                 additional_headers={b"Host": [""]})
+
 
 def _build_app(test):
     test.initialize()
     return VolumePlugin(
         test.volume_plugin_reactor, test.flocker_client, test.NODE_A).app
-RealTestsAPI, MemoryTestsAPI = buildIntegrationTests(
-    APITestsMixin, "API", _build_app)
+RealTestsAPI = buildUNIXIntegrationTests(APITestsMixin, "API", _build_app)
```
exarkun: No support for running over a TCP socket, then?  If this is because you never use Docker plugins like that, cool.

Any particular reason not to maintain the in-memory version?  Because it's useless (ie, it has never found a problem more quickly or identified the cause more clearly than the real-socket kind)?

flocker/restapi/testtools.py
```diff
@@ -211,6 +211,42 @@ def setUp(self):
     MemoryTests.__module__ = mixinClass.__module__
     return RealTests, MemoryTests
 
+
+def buildUNIXIntegrationTests(mixinClass, name, fixture):
+    """
+    Build L{TestCase} class that runs the tests in the mixin class with
+    real queries over a UNIX socket.
+
+    @param mixinClass: A mixin class for L{TestCase} that relies on having a
```
exarkun: I do this sometimes too.  sigh.

`s/@/:/`

exarkun: Oh, and ``TestCase``.

exarkun: Bah, markup fight.  I mean \`\`TestCase\`\`

flocker/restapi/testtools.py
```diff
@@ -211,6 +211,42 @@ def setUp(self):
...
+        C{self.scenario}.
+
+    @param name: A C{str}, the name of the test category.
```
exarkun: and so on

flocker/restapi/testtools.py
exarkun: Oh wait ... maybe?  Or does the plugin actually use epytext everywhere?  Whaaat?

flocker/dockerplugin/__init__.py
exarkun: So, if fixing this without this global state mashing isn't feasible, I suggest that fixing this problem upstream is a good candidate for what to work on next.

exarkun:
Thanks.

I think that this change isn't *inherently* valuable.  What's valuable is allowing users to use Docker 1.10 with the plugin.  As you pointed out elsewhere, this seems necessary but we don't know if it is sufficient.

Maybe we'll be lucky and this is the only change required.  In that case, nice cheap win.  If not, maybe we should identify the acceptance criteria necessary to ensure we deliver something useful to users and then prioritize the work of achieving that like we prioritize other feature development work.

Rest of my comments inline.  Please address and merge when you're satisfied.

flocker/dockerplugin/__init__.py
itamarst: I'm not sure why you think this won't happen in an actual Flocker process. This is only necessary for Docker plugin (previously it run in the docker plugin script's `main()` method) and anything involving docker plugin (test or real use) will import this package.

flocker/dockerplugin/test/test_api.py
itamarst: 1. Yeah, Docker plugin is always UNIX AFAIK.
2. I've never seen the in-memory behave any differently, yeah. Either both fail or neither fail... except in this case where only Unix one fails. So not sure memory one adds anything.

flocker/restapi/testtools.py
itamarst: This code was inherited from HybridCluster...

flocker/dockerplugin/__init__.py
itamarst: Oh, I see, the "non-docker-plugin Flocker service" case. If that ever happens then we'd have acceptance tests that fail so that's probably OK. Especially if we upstream a (better) fix.

itamarst:
At the very minimum this will be useful in helping users diagnose 1.10 errors, insofar as they will now get more meaningful response from the plugin.

PR #2291, opened 2015-12-04
[FLOC-3613] Log discovery as its own action.

itamarst:


flocker/node/test/test_loop.py
```diff
@@ -353,6 +365,7 @@ def test_convergence_done_notify(self, logger):
         )
         self.assertEqual(client.calls, [(NodeStateCommand,
                                          dict(state_changes=(local_state,)))])
+        action.result.callback(None)
```
exarkun: How about `action = ControllableAction(result=succeed(None))` instead?

flocker/node/_loop.py
```diff
@@ -332,6 +332,10 @@ class ConvergenceLoopOutputs(Names):
     [_FIELD_CLUSTERSTATE, _FIELD_CONFIGURATION], [],
     "The convergence action within the loop.")
 
+LOG_DISCOVERY = ActionType(
+    u"flocker:agent:discovery", [], [],
+    "The deployer is doing discovery of local state.")
```
exarkun: `u"The deployer ..."` right?

flocker/node/_loop.py
```diff
@@ -453,8 +457,12 @@ def output_CONVERGE(self, context):
...
         with LOG_CONVERGE(self.fsm.logger, cluster_state=self.cluster_state,
                           desired_configuration=self.configuration).context():
-            d = DeferredContext(maybeDeferred(
-                self.deployer.discover_state, known_local_state))
+            d = DeferredContext(succeed(None))
+            with LOG_DISCOVERY(self.fsm.logger).context():
+                discover = DeferredContext(maybeDeferred(
+                    self.deployer.discover_state, known_local_state))
+                discover.addActionFinish()
+            d.addCallback(lambda _: discover.result)
```
exarkun: I can't say it's obvious to me how this `d` / `discover` chaining is different from just defining `d` as `discover.result` after the `with` block.  It strikes me that this might be an idiom for doing something pretty useful ... but what?

I would have like to have answered this question on my own by reading the resulting logs but since the trial logs are all mangled and unparseable and actually deploying the software would take half an hour I didn't bother.

exarkun:
Thanks.  Build results look about as good as can be expected right now.  Just the few minor inline comments.  I would like to understand the `DeferredContext(succeed(None))` thing so please explain that wherever you think is appropriate, address other points to your satisfaction, and then merge.

itamarst:
I think that was actually unnecessary, I can probably simplify it.

PR #2293, opened 2015-12-04
[FLOC-3525] Upgrade testing 

sarum90:
This enables a simple acceptance level test to verify happy-case upgradability of Flocker.

Note that in general this does not ensure that upgrades will work, and if you are writing something that has more tricky upgrade ramifications you also need to have lower level tests that attempt to verify upgrading all different states that a cluster could be in. Instead, this is supposed to at least give us some amount of confidence that we don't regress or add any blatant errors in upgradability.

The general approach:

- Re-use our existing provisioning code to uninstall flocker from one version and install a different version.
- Pass additional env vars from run_acceptance_test to trial in order to get the required information into the trial process to do that work.
- Install the most recently released installable version of flocker
- Make a dataset and write some data to it.
- Uninstall the old version of flocker and install HEAD
- verify the new version of flocker is running and then read the data previously written.

This also includes a workaround for a twisted bug, and a workaround for the fact that the most recent release did not actually create any installable targets.


jml:
LGTM. Merge when it passes tests in a way that brings justice, propriety, and benefit.

sarum90:
Going to re-run tests on this.

I also think I'm going to queue up a bunch of test_dataset runs so I can try to catch flakes in the test and see the impact on runtime before merging.

sarum90:
Hrm... so I'm not sure how to get traction on this PR... for the past 2 days I've been trying to use Jenkins to  run acceptance tests, but I consistently keep hitting flakes (I think).

I've got some other coding tasks that I should be concentrating on, and this does not block me for awhile yet, so I'm sort of inclined to let this sit until the acceptance test environment is a bit healthier and then run tests on it for awhile before merging.

I'll keep attempting to run tests on it at a low priority -- but I'd still like to see like 3-4 passes of the endtoend tests somewhere before merging, and that doesn't seem to be likely to be done before next week.

sarum90:
So this is hard for me to merge right now, because while doing upgrade testing I first downgrade, and when I downgrade to 1.8.0 on CentOS I hit FLOC-3712 and can no longer communicate with the control agent.... we maybe should do a release again sometime.

sarum90:
But that did help me catch that I was unnecessarily causing cascading failures in this branch... perhaps with a flaky decorator this might just work now...

sarum90:
Been awhile since this was accepted. After it was accepted I got some red on Jenkins, and what I thought would be just a few more iterations turned into more of a saga of building confidence in this PR. So, sending out for a second review.

Changes since last review:

- Hardened cleanup code. Previously it was possible to get stuck with an old version of flocker on one of the nodes.
- Added code to make an HTTP GET request to a specific S3 bucket every time we install flocker if we install it from one of our releases using this provisioning code. This lets us track our internal downloads.
- Changed some things that are no longer required now that 1.9.0 is released.
- Ran the test a _lot_, no flakes relating to the upgrading flocker in 13+* runs:
  - http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/upgrade-testing-FLOC-3525/job/_run_acceptance_on_AWS_CentOS_7_with_EBS/
  - http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/upgrade-testing-FLOC-3525/job/_run_acceptance_on_AWS_Ubuntu_Trusty_with_EBS/

\* Failures centos/6 and ubuntu/9 were because of using persistent http pools with treq, note that the actual upgrade still was fine. Failure ubuntu/7 was a flake because the jenkins slave was randomly terminated.

sarum90:
One more note on tests: I believe acceptance tests will fail on master tonight, which is why they are failing on this branch right now. Looking at remote_logs there is a suspicious traceback that is unrelated to any of my work in this branch.

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,114 @@ def test_dataset_creation(self, cluster):
...
+        # Verify that the file still has its contents.
+        def cat_and_verify_file(dataset):
+            output = []
+
+            def append_output(line):
+                output.append(line)
+            file_catting = node.run_as_root(
+                ['bash', '-c', 'cat %s' % (
+                    os.path.join(dataset.path.path, 'test.txt'))],
+                handle_stdout=append_output)
```
jml: :question: I'd personally just use `output.append` and forget about the `append_output` helper.

flocker/acceptance/testtools.py
```diff
@@ -89,6 +90,13 @@
 SOCKET_TIMEOUT_FOR_POLLING = 2.0
 
 
+class FailureToUpgrade(Exception):
+    """
+    Exception raised to indicate a failure to install a new version of Flocker.
+    """
+    pass
```
jml: No need for `pass`, given docstring.

jml:
LGTM. Minor comments. I've hit restart on the two aborted tests.

sarum90:
Thanks for hitting restart. I was aborting things last night cause I was hitting requestLimitExceeded all over the place.

I'll be addressing comments and then running acceptance tests a few more times to convince myself we aren't breaking anything before merging.

flocker/acceptance/endtoend/test_dataset.py
sarum90: Decided to do `lambda line: append_output(line)` instead.

...

Just kidding, obviously your suggestion was better and I'm using that.

flocker/acceptance/testtools.py
sarum90: Removed

I don't understand when to or not to use pass :(

Would you mind teaching me in the form of a suggested edit to https://docs.google.com/document/d/1WyPosnJuIyAdPHnI5E0D5QYzcwhyBqVYIEZoHeh3ndQ/edit ?  Might not be the best use of your time though... maybe I'll look it up and add a section later.

sarum90:
Thanks for the review!

Got green runs of the acceptance tests now that master is fixed, so 15 successes with no upgrade related failures. Waiting for all of the other tests to be green before merging.

sarum90:
With all green I'll be merging now :)

PR #2299, opened 2015-12-07
[FLOC-3498] Use ClusterHQ fork of testtools

jml:
It's taking days to get patches landed in testtools and these delays are having a direct & significant impact on our development velocity (flaky tests; debugging info from failing tests). This patch circumvents  that by using a temporary, dedicated fork: ClusterHQ/testtools#clusterhq-fork

The trade-off is that we'll probably have to do some rework in exchange for advancing our learning from the future to today.

The fork currently incorporates the latest versions of testing-cabal/testtools#165 and testing-cabal/testtool#171.

The astute will note that we've _already_ learned something, which is that testing-cabal/testtools#165 isn't enough for a `RunTest` implementation to be able to run a test case twice. I plan to address that in a follow-up patch on testtools, since 165 is already huge.

### DIFF
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -238,8 +238,9 @@ class _RetryFlaky(testtools.RunTest):
             result was and ``details`` is a dictionary of testtools details.
         """
         tmp_result = testtools.TestResult()
-        # XXX: jml: Argh! Of course putting _reset in TestCase.run() doesn't
-        # help runners that want to re-run.
+        # XXX: Still using internal API of testtools despite improvements in
+        # #165. Will need to do follow-up work on testtools to ensure that
+        # RunTest.run(case); RunTest.run(case) is supported.
         case._reset()
         self._run_test(case, tmp_result)
         result_type = _get_result_type(tmp_result)
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -238,8 +238,9 @@ class _RetryFlaky(testtools.RunTest):
             result was and ``details`` is a dictionary of testtools details.
         """
         tmp_result = testtools.TestResult()
-        # XXX: Work around https://bugs.launchpad.net/testtools/+bug/1517879
-        _reset_case(case)
+        # XXX: jml: Argh! Of course putting _reset in TestCase.run() doesn't
+        # help runners that want to re-run.
+        case._reset()
         self._run_test(case, tmp_result)
         result_type = _get_result_type(tmp_result)
         details = pmap(case.getDetails())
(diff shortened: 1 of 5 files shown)


flocker/testtools/_flaky.py
```diff
@@ -238,8 +238,9 @@ def _attempt_test(self, case):
             result was and ``details`` is a dictionary of testtools details.
         """
         tmp_result = testtools.TestResult()
-        # XXX: Work around https://bugs.launchpad.net/testtools/+bug/1517879
-        _reset_case(case)
+        # XXX: jml: Argh! Of course putting _reset in TestCase.run() doesn't
```
sarum90: I don't understand this XXX.

setup.py
```diff
@@ -108,6 +108,12 @@
 
     cmdclass=versioneer.get_cmdclass(),
 
+    dependency_links = [
+        # Use our fork of testtools until #165, #171, and #172 are merged and
+        # released. See FLOC-3498.
+        "git+git://github.com/ClusterHQ/testtools@clusterhq-fork#egg=testtools-1.8.2chq1",  # noqa
```
sarum90: is git+git a thing? It's a hard thing to google search for so I couldn't verify.

sarum90:
This looks fine to me. As far as I'm concerned if this makes your work go faster and it doesn't break any clusterHQ developers dev environment or the jenkins tests it should be submitted. That said... it seems to still make Jenkins be completely red, so I'm rejecting only because I'm unsure how complex the fix to that will be.

Also I wanted to double check:
-Does this work on the Vagrant Dev box?
-Does this work on an OSX dev environment?

^^ I'm willing to test one or both of those for you as part of the review.

I did not attempt to understand every part of this PR, instead basically checking it for typos and the like. It looks reasonable to me, and I'm not worried about running our own fork of a testing library.

flocker/testtools/_flaky.py
jml: Fixed.

setup.py
jml: Yes, it is. Added a comment explaining what's going on & linked to the docs.

jml:
* Fixed Jenkins being completely red (now completely green, although not yet seen results of last push, which only changed comments)
* Confirm it works on OS X dev environment (that's what I used to develop)
* No idea whether it works on vagrant dev box—we're only half supporting it now—but feel free to test and I'll do any shallow fixes required.
* I've updated the bug to have more explanation of why we're forking.

Thanks!

sarum90:
Working on OSX and in the automated tests is good enough for me.

So long as this isn't breaking peoples builds go ahead and merge! :)

PR #2300, opened 2015-12-07
[FLOC-3628] Add elapsed wallclock time to CPU time samples.

jongiddy:
Add wallclock times to CPU time samples to enable calculation of CPU load % from raw CPU times.

Wallclock times are calculated per node, to handle any delays for an individual node's network.

The wallclock time is measured at the start of parsing the output. This would seem to be the point on the local side with the fewest opportunities for added latency from external conditions.

### DIFF
--- a/benchmark/metrics/cputime.py
+++ b/benchmark/metrics/cputime.py
@@ -93,7 +93,7 @@ class CPUParser(LineOnlyReceiver):
         self.result[name] = self.result.get(name, 0) + cputime
 
 
-class SSHRunner:
+class SSHRunner(object):
     """
     Run a command using ssh.
 
--- a/benchmark/metrics/cputime.py
+++ b/benchmark/metrics/cputime.py
@@ -44,10 +44,15 @@ WALLCLOCK_LABEL = '-- WALL --'
 class CPUParser(LineOnlyReceiver):
     """
     Handler for the output lines returned from the cpu time command.
+
+    After parsing, the ``result`` attribute will contain a dictionary
+    mapping process names to elapsed CPU time.  Process names may be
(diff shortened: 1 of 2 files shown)


benchmark/metrics/cputime.py
```diff
@@ -36,19 +36,28 @@
...
 
 class CPUParser(LineOnlyReceiver):
     """
     Handler for the output lines returned from the cpu time command.
     """
 
-    def __init__(self):
+    def __init__(self, reactor):
+        self.reactor = reactor
         self.result = {}
```
jml: :question: I'd suggest prefixing both of these attributes with `_` since they aren't (AIUI) meant to be accessed externally.

benchmark/metrics/test/test_cputime.py
```diff
@@ -204,7 +213,8 @@ def filter(node_cpu_times):
             for process_times in node_cpu_times.values():
                 if process_times:
                     for process in process_times:
-                        process_times[process] = 0
+                        if process != WALLCLOCK_LABEL:
+                            process_times[process] = 0
```
jml: :question: If you're bored, it might be fun to try expressing this w/ pmap.transform (http://pyrsistent.readthedocs.org/en/latest/intro.html#transformations)

jml:
Great stuff. No particularly pressing suggestions. Act on them at your leisure and merge when you've made your peace with the tests. :)

benchmark/metrics/test/test_cputime.py
jongiddy: Since it's test code, I'd rather keep it verbose and obvious and independent of external libraries.

PR #2301, opened 2015-12-07
[FLOC-3631] Fix `test_leases.LeaseAPITests.test_lease_prevents_move`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3631


exarkun:
This is fixed in FLOC-3077 or FLOC-3199 I forget which.

exarkun:
Aha.  I spoke too quickly.  It's *like* something fixed in another branch but it's a different case.

PR #2302, opened 2015-12-07
[FLOC-3614] IBlockDeviceAPI implementation tips and hints

itamarst:
1. Some hints were added to interface.
2. One was added to plugin docs.
3. Plugin docs were re-ordered slightly so instructions make more sense (docs now finish talking about necessary code before talking about how to publish the result).

### DIFF
--- a/docs/gettinginvolved/plugins/building-driver.rst
+++ b/docs/gettinginvolved/plugins/building-driver.rst
@@ -134,14 +134,14 @@ Here's what the module could look like:
 The ``cluster_id`` parameter is a Python :py:obj:`uuid.UUID` instance uniquely identifying the cluster.
 This is useful if you want to build a system that supports multiple Flocker clusters talking to a shared storage backend.
 
-Make sure that your factory function raises an exception if it is given incorrect or insufficient parameters, so that users can easily they have mis-configured your backend.
+Make sure that your factory function raises an exception if it is given incorrect or insufficient parameters, so that users can easily see when they have mis-configured your backend.
 
 .. XXX FLOC-3461 might suggest using ``UsageError`` exceptions, or some other more specific suggestion.
 
 Publishing Your Driver
 ======================
 
-Once your CI tests are running and passing successfully, you are ready to publish your driver and assert that is certified to work with Flocker.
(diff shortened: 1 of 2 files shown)


docs/gettinginvolved/plugins/building-driver.rst
```diff
@@ -154,6 +134,30 @@ Here's what the module could look like:
 The ``cluster_id`` parameter is a Python :py:obj:`uuid.UUID` instance uniquely identifying the cluster.
 This is useful if you want to build a system that supports multiple Flocker clusters talking to a shared storage backend.
 
+Make sure that your factory function raises an exception if it is given incorrect or insufficient parameters, so that users can easily they have mis-configured your backend.
```
wallrj: typo: easily *see when* they

docs/gettinginvolved/plugins/building-driver.rst
```diff
@@ -154,6 +134,30 @@ Here's what the module could look like:
...
+
+.. XXX FLOC-3461 might suggest using ``UsageError`` exceptions, or some other more specific suggestion.
+
+Publishing Your Driver
+======================
+
+Once your CI tests are running and passing successfully, you are ready to publish your driver and assert that is certified to work with Flocker.
```
wallrj: assert that *it* is

wallrj:
Thanks @itamarst 

Looks like a good improvement.

Some other things that occur to me:
 * Should `attach_volume` be idempotent to some extent. e.g.
    1. vnx driver calls naviseccli to configure the storage array so that a LUN is available to a host, then
    2. calls scsi-rescan on the host to that the now available LUN is assigned a device in the OS.

If step 2 fails first time round (which it may) should the driver clean up and detach_volume? 
(we do that in the EBS driver if the new device path isn't what's expected)
And if so should detach_volume operate on partially attached volumes?
Or should the cleanup be performed by calling the driver specific detachment API?
Or should it leave the partially attached volume behind and list it as unattached, so that a subsequent attach_volume call gets an opportunity to complete the attachment?

If you think any of those things are worth mentioning add them.
Fix the couple of typos.
Then merge.

itamarst:
Should probably explain implications of convergence loop, yeah.

PR #2304, opened 2015-12-08
[FLOC-3637] Removing Ubuntu 15.04 from supported distros for nodes in acceptance …

mpgomez:
…test runner

And that is what happens when working in two different branches in the same area of code.
Moving to 15.10 in one branch, pulling the logs from ubuntu nodes in another branch. Merge. And bad code there.
I had removed 15.04 instead of changing it to 15.10 because we do not support 15.10 for nodes as far as I know, only for the client.


itamarst:
Merge when sufficiently green, thanks.

PR #2306, opened 2015-12-08
[FLOC-3641] Improve Docker plugin logs

itamarst:
The Docker plugin, and utilities it relied on, don't preserve Eliot context. Fix that so it's easier to read a trace of what happened.


jml:
LGTM, thanks.

PR #2311, opened 2015-12-09
[FLOC 3535] cluster setup tool

avg-I:
A tool "branched off" of `run-acceptance-tests`.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -205,9 +205,10 @@ class ManagedRunner(object):
         use when they are "started".
     :ivar dict dataset_backend_configuration: The backend-specific
         configuration the nodes will be given for their dataset backend.
+    :ivar str purpose: The purpose of the cluster.
     """
     def __init__(self, node_addresses, package_source, distribution,
-                 dataset_backend, dataset_backend_configuration):
+                 dataset_backend, dataset_backend_configuration, purpose):
         """
         :param list: A ``list`` of public IP addresses or
             ``[private_address, public_address]`` lists.
@@ -236,6 +237,7 @@ class ManagedRunner(object):
         self.package_source = package_source
         self.dataset_backend = dataset_backend
         self.dataset_backend_configuration = dataset_backend_configuration
+        self.purpose = purpose
 
     def _upgrade_flocker(self, reactor, nodes, package_source):
         """
@@ -288,20 +290,22 @@ class ManagedRunner(object):
         else:
             upgrading = succeed(None)
 
+        cluster_name, cluster_id = make_cluster_name_and_id(
+            self.purpose,
+            self.dataset_backend
+        )
+
         def configure(ignored):
             return configured_cluster_for_nodes(
                 reactor,
-                generate_certificates(
-                    make_cluster_id(
-                        TestTypes.ACCEPTANCE,
-                        _provider_for_cluster_id(self.dataset_backend),
-                    ),
-                    self._nodes),
+                generate_certificates(cluster_name, cluster_id, self._nodes),
                 self._nodes,
                 self.dataset_backend,
                 self.dataset_backend_configuration,
-                _save_backend_configuration(self.dataset_backend,
-                    self.dataset_backend_configuration),
+                _save_backend_configuration(
+                    self.dataset_backend,
+                    self.dataset_backend_configuration
+                ),
                 provider="managed"
             )
         configuring = upgrading.addCallback(configure)
@@ -328,12 +332,13 @@ def _provider_for_cluster_id(dataset_backend):
     return Providers.UNSPECIFIED
 
 
-def generate_certificates(cluster_id, nodes):
+def generate_certificates(cluster_name, cluster_id, nodes):
     """
     Generate a new set of certificates for the given nodes.
 
     :param UUID cluster_id: The unique identifier of the cluster for which to
-        generate the certificates.
+        generate the certificates.  If ``None`` then a new random identifier
+        is generated.
     :param list nodes: The ``INode`` providers that make up the cluster.
 
     :return: A ``Certificates`` instance referring to the newly generated
@@ -345,9 +350,34 @@ def generate_certificates(cluster_id, nodes):
         certificates_path,
         nodes[0].address,
         len(nodes),
+        cluster_name=cluster_name,
         cluster_id=cluster_id,
     )
     return certificates
+
+
+def make_cluster_name_and_id(purpose, dataset_backend):
+    """
+    Generate a cluster name and cluster UUID based on the cluster's purpose.
+
+    :param str purpose: The purpose of the cluster.
+                        "acceptance-testing" is a special value.
+    :param DatasetBackend dataset_backend: The dataset backend used by
+                        the cluster.
+
+    :rtype: tuple of bytes, UUID
+    :returns: The generated name and UUID of the cluster as a tuple.
+    """
+    if purpose == u'acceptance-testing':
+        cluster_name = b'acceptance-cluster'
+        cluster_id = make_cluster_id(
+            TestTypes.ACCEPTANCE,
+            _provider_for_cluster_id(dataset_backend),
+        )
+    else:
+        cluster_name = b'{}-cluster'.format(purpose.encode("ascii"))
+        cluster_id = None       # let a random UUID be generated
+    return (cluster_name, cluster_id)
 
 
 def _save_backend_configuration(dataset_backend_name,
@@ -626,14 +656,14 @@ class LibcloudRunner(object):
 
         yield perform(make_dispatcher(reactor), commands)
 
+        cluster_name, cluster_id = make_cluster_name_and_id(
+            self.purpose,
+            self.dataset_backend
+        )
+
         cluster = yield configured_cluster_for_nodes(
             reactor,
-            generate_certificates(
-                make_cluster_id(
-                    TestTypes.ACCEPTANCE,
-                    _provider_for_cluster_id(self.dataset_backend),
-                ),
-                self.nodes),
+            generate_certificates(cluster_name, cluster_id, self.nodes),
             self.nodes,
             self.dataset_backend,
             self.dataset_backend_configuration,
@@ -849,6 +879,7 @@ class CommonOptions(Options):
             distribution=self['distribution'],
             dataset_backend=dataset_backend,
             dataset_backend_configuration=self.dataset_backend_configuration(),
+            purpose=self['purpose'],
         )
 
     def _libcloud_runner(self, package_source, dataset_backend,
@@ -945,7 +976,7 @@ class RunOptions(CommonOptions):
         :param FilePath top_level: The top-level of the flocker repository.
         """
         super(RunOptions, self).__init__(top_level)
-        self['purpose'] = 'acceptance-testing'
+        self['purpose'] = u'acceptance-testing'
 
     def parseArgs(self, *trial_args):
         self['trial-args'] = trial_args
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -8,7 +8,6 @@ import os
 import yaml
 import json
 import re
-import string
 from pipes import quote as shell_quote
 from tempfile import mkdtemp
 
@@ -17,7 +16,7 @@ from characteristic import attributes
 from eliot import (
     add_destination, write_failure, FileDestination
 )
-from pyrsistent import pvector
+from pyrsistent import PClass, field, pvector
 from bitmath import GiB
 
 from twisted.internet.error import ProcessTerminated
@@ -29,6 +28,8 @@ from twisted.python.reflect import prefixedMethodNames
 
 from effect import parallel
 from txeffect import perform
+
+from uuid import UUID
 
 from admin.vagrant import vagrant_version
 from flocker.common import RACKSPACE_MINIMUM_VOLUME_SIZE, gather_deferreds
@@ -138,6 +139,29 @@ def run_tests(reactor, cluster, trial_args):
             )
 
 
+class ClusterIdentity(PClass):
+    """
+    The information that is used to identify a cluster.
+
+    :ivar str purpose: The intended purpose of the cluster.
+    :ivar str name: The name of the cluster.
+    :ivar UUID id: The UUID of the cluster, ``None`` means a random ID.
+    """
+    purpose = field(mandatory=True, type=unicode)
+    name = field(mandatory=True, type=bytes)
+    id = field(mandatory=False, type=(UUID, type(None)), initial=None)
+
+    @property
+    def metadata(self):
+        """
+        The commonly used metadata describing the cluster.
+
+        :return: The metadata.
+        :rtype: dict of str:str
+        """
+        return {u'purpose': self.purpose}
+
+
 class IClusterRunner(Interface):
     """
     Interface for starting and stopping a cluster for acceptance testing.
@@ -205,10 +229,10 @@ class ManagedRunner(object):
         use when they are "started".
     :ivar dict dataset_backend_configuration: The backend-specific
         configuration the nodes will be given for their dataset backend.
-    :ivar str purpose: The purpose of the cluster.
+    :ivar ClusterIdentity identity: The identity information of the cluster.
     """
     def __init__(self, node_addresses, package_source, distribution,
-                 dataset_backend, dataset_backend_configuration, purpose):
+                 dataset_backend, dataset_backend_configuration, identity):
         """
         :param list: A ``list`` of public IP addresses or
             ``[private_address, public_address]`` lists.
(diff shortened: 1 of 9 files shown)


admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
jongiddy: Use
```
# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
```

admin/setup-cluster
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env python
+# Copyright Hybrid Logic Ltd.  See LICENSE file for details.
```
jongiddy: Ditto for the copyright notice.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+         'The dataset backend to test against. '
+         'One of {}'.format(', '.join(backend.name for backend
+                                      in DatasetBackend.iterconstants()))],
+        ['config-file', None, None,
+         'Configuration for compute-resource providers and dataset backends.'],
+        ['branch', None, None, 'Branch to grab packages from'],
+        ['flocker-version', None, None, 'Version of flocker to install'],
+        ['build-server', None, 'http://build.clusterhq.com/',
+         'Base URL of build server for package downloads'],
+        ['node-count', None, 2, 'Number of nodes to create (where applicable)',
```
jongiddy: What does "(where applicable)" mean?  While there are some acceptance tests that only require 1 node, I would expect that `start-test-cluster` always starts `node-count` nodes.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+         int],
+        ['app-template', None, None,
+         'Configuration to use for each application container'],
+        ['apps-per-node', None, 1, 'Number of application containers per node',
```
jongiddy: Make the default value 0. It would be reasonable to want a test cluster with no containers.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+            dataset_backend_configuration=self.dataset_backend_configuration(),
+        )
+
+    def _libcloud_runner(self, package_source, dataset_backend,
+                         provider, provider_config):
+        """
+        Run some nodes using ``libcloud``.
+
+        By default, two nodes are run.  This can be overridden by setting
+        ``FLOCKER_ACCEPTANCE_NUM_NODES`` in the environment.
```
jongiddy: This part of the docstring can be removed, as it no longer uses the environment variable.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+                        name=environment_variable,
+                        value=shell_quote(
+                            environment_variables[environment_variable]),
+                    ))
+                print("Be sure to preserve the required files.")
+
+    raise SystemExit(result)
+
+
+def _build_config(cluster, application_template,  per_node):
```
jongiddy: Remove extra space before `per_node`.

jongiddy: Add a docstring for function.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+        nodes[node.private_address] = []
+        for i in range(per_node):
+            name = "app_%s_%d" % (node.private_address, i)
+            nodes[node.private_address].append(name)
+
+    return {"applications": application_root,
+            "deployment": deployment_root}
+
+
+def _configure(reactor, cluster, configuration):
```
jongiddy: Add docstring.

admin/cluster_setup.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def got_all_nodes():
+        d = treq_client.get(
+            base_url + b"/state/nodes",
+            persistent=False
+        )
+        d.addCallback(check_and_decode_json, OK)
+        d.addCallbacks(
+            lambda nodes: len(nodes) >= len(cluster.agent_nodes),
+            lambda _: False
```
jongiddy: Because this drops the Failure passed to the errBack, it should log the error.  You can use `twisted.python.log.err` as the error callback (without a lambda), since it accepts a Failure, and returns `None` (which is equivalent to `False` in `loop_until`.

admin/cluster_setup.py
avg-I: done

admin/setup-cluster
avg-I: done

admin/cluster_setup.py
avg-I: Well, I kept support for using `managed` nodes, so in that case the node count comes from the configuration.  I am not sure if we actually need to support that use case.

admin/cluster_setup.py
avg-I: okay

admin/cluster_setup.py
avg-I: okay

admin/cluster_setup.py
avg-I: 2x okay

admin/cluster_setup.py
avg-I: okay

admin/cluster_setup.py
avg-I: Thank you for the suggestion and for teaching me about `twisted.python.log.err`.
Does it work nicely with eliot?

avg-I:
Updated the code.

admin/cluster_setup.py
```diff
@@ -0,0 +1,497 @@
...
+            config_file = FilePath(self['config-file'])
+            self['config'] = yaml.safe_load(config_file.getContent())
+        else:
+            self['config'] = {}
+
+        if self['app-template'] is not None:
+            template_file = FilePath(self['app-template'])
+            self['template'] = yaml.safe_load(template_file.getContent())
+        else:
+            self['template'] = {}
```
jongiddy: If `app-template` is not set, but `apps-per-node > 0` the code gives the error:
```
Application 'app_172.31.37.42_0' has a config error. Missing 'image' key.
```

Raise a UsageError if this combination of options is provided.

admin/cluster_setup.py
```diff
@@ -0,0 +1,497 @@
...
+    """
+    base_url = b"https://{}:{}/v1".format(
+        cluster.control_node.address, REST_API_PORT
+    )
+    certificates_path = cluster.certificates_path
+    cluster_cert = certificates_path.child(b"cluster.crt")
+    user_cert = certificates_path.child(b"user.crt")
+    user_key = certificates_path.child(b"user.key")
+    body = dumps(configuration)
+    treq_client = treq_with_authentication(
```
jongiddy: Have you considered using apiclient.FlockerClient here?  I guess the only problem is the lack of a method to perform `/configuration/_compose`?

jongiddy:
It would be good to add a page to the documentation, describing the use of the new script.  I'd suggest that you copy the `docs/gettinginvolved/acceptance-testing.rst`, and modify it to add/remove supported options. Remove or rename any link targets (`..a-name:`) so they are unique. Add your new page to the `docs/gettinginvolved/index.rst` file.

jongiddy:
The code is a little too repetitive. One of the advantages of Twisted's Options class is that it allows options to be subclassed.  I think that would be useful here. Create a common file imported into both `cluster_setup.py` and `acceptance.py` which defines the common subset of Options, and then subclass it in each file to get it back to the current state. This common file can also contain some other bits, like the supported `DISTRIBUTIONS`.

Let me know if you need clarification of this.

jongiddy:
Thanks Andriy. Please address the comments above, and re-submit.

admin/cluster_setup.py
jongiddy: Jean-Paul suggests using `eliot.write_failure`, which has the same property of accepting a Failure and returning `None`.

admin/cluster_setup.py
avg-I: Yes, `FlockerClient` does not expose a method like that.
FWIW, I wrote an alternative implementation using `FlockerClient`. I think that it looks much nicer but likely its performance is worse (especially for large setups):
https://github.com/ClusterHQ/flocker/commit/7c56245937f6555d0a193a562a6c20b12e5153d3

admin/acceptance.py
```diff
@@ -585,8 +623,8 @@ def start_cluster(self, reactor):
         print "Assigning random tag:", random_tag
 
         for index in range(self.num_nodes):
-            name = "acceptance-test-%s-%s-%d" % (
-                self.creator, random_tag, index,
+            name = "%s-%s-%s-%d" % (
+                self.purpose, self.creator, random_tag, index,
```
jongiddy: With the addition of the unique identifier, I would rather see the more distinguishing info at the start.  Move the purpose to the final part of the label (e.g. `jongiddy-lk3hEs-1-acceptance-testing`).

admin/acceptance.py
```diff
@@ -344,11 +350,36 @@ def generate_certificates(cluster_id, nodes):
...
+
+    :param str purpose: The purpose of the cluster.
+                        "acceptance-testing" is a special value.
+    :param DatasetBackend dataset_backend: The dataset backend used by
+                        the cluster.
+
+    :rtype: tuple of bytes, UUID
+    :returns: The generated name and UUID of the cluster as a tuple.
+    """
+    if purpose == u'acceptance-testing':
```
jongiddy: This special value should be represented as a constant string defined in one place, and imported everywhere it is used.

admin/acceptance.py
```diff
@@ -562,6 +593,13 @@ def __init__(self):
             )
         self.creator = creator
 
+        if any(x not in string.ascii_letters + string.digits + '-'
+               for x in self.purpose):
+            raise UsageError(
+                "Purpose may have only alphanumeric symbols and dash. " +
+                "Found {!r}".format(self.purpose)
+            )
```
jongiddy: Validation of the `purpose` value should be done in `postOptions`.

admin/cluster_setup.py
```diff
@@ -0,0 +1,258 @@
...
+)
+
+from flocker.acceptance.testtools import check_and_decode_json
+from flocker.ca import treq_with_authentication
+from flocker.common import gather_deferreds, loop_until
+from flocker.control.httpapi import REST_API_PORT
+
+
+class RunOptions(CommonOptions):
+    description = "Set up a flocker cluster."
```
jongiddy: Use capital-F for this and all mentions of Flocker.

admin/cluster_setup.py
```diff
@@ -0,0 +1,258 @@
+# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
+"""
+Set up a flocker cluster.
```
jongiddy: Capital-F for Flocker

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -0,0 +1,226 @@
...
+    --distribution centos-7 \
+    --provider rackspace \
+    --config-file $PWD/cluster.yml \
+    --number-of-nodes 2 \
+    --app-template $PWD/application.yml \
+    --apps-per-node 5
+
+Note that all application instances will have exactly the same configuration.
+
+Configuration File
```
jongiddy: Replace this section with a link to the acceptance test docs.

jongiddy:
The `purpose` is currently fairly awkward, creating `if` statements to try to handle legacy acceptance test formats, while adding the new formats.

I'd suggest that you create a small class that has methods `get_cluster_name`, `get_cluster_id`, `get_metadata`, and `make_node_name`. The `make_node_name` takes an index and returns a name with the index appended.

An instance is created in the acceptance test (returning the original values), and a different instance is created in `setup_cluster` returning values based on the purpose flag, or possibly additional flags.

Pass the instance to the Runners to give them the names and metadata.

admin/acceptance.py
avg-I: okay

admin/acceptance.py
avg-I: okay

admin/acceptance.py
avg-I: I followed the example of validating `creator` here, but indeed `creator` is not a command-line parameter.

admin/cluster_setup.py
avg-I: okay

docs/gettinginvolved/cluster-setup.rst
avg-I: okay!

avg-I:
> I'd suggest that you create a small class that has methods get_cluster_name, get_cluster_id, get_metadata, and make_node_name.

Will do. Working on this now.

admin/acceptance.py
avg-I: Not doing this change as per the conversation with JP.

admin/acceptance.py
avg-I: I think that this comment has become obsolete as after implementing your other suggestion the string is going to be used in only one place.

avg-I:
I've ended up implementing this suggestion in a slightly reduced form.
As node names are generated only for the Libcloud backend I decided not to include that functionality into the new class. Maybe it still would make sense to extract the code into its own method or function, but in my opinion it would not belong in the new class (also `creator` seems to be used only with Libcloud as well).

admin/acceptance.py
```diff
@@ -794,6 +811,20 @@ def postOptions(self):
...
+    def _make_cluster_identity(self, dataset_backend):
+        """
+        Build a cluster identity based on the parameters.
+        """
+        cluster_id = make_cluster_id(
+            TestTypes.ACCEPTANCE,
+            _provider_for_cluster_id(dataset_backend),
+        )
+        return ClusterIdentity(
+            purpose=u'acceptance-testing',
```
jongiddy: The current script uses `acceptance-test-` as the prefix for the nodes, but `acceptance-testing` for the metadata purpose. Add a `prefix` field to the ClusterIdentity and set it to `acceptance-test` here.  Use `prefix` when generating the node name.

In `cluster_setup.py`, `prefix` can be set to the same value as `purpose`.

avg-I: That makes sense (and it's easy to do :smile:).

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -0,0 +1,200 @@
...
+
+  admin/setup-cluster \
+    --distribution centos-7 \
+    --provider rackspace \
+    --config-file $PWD/cluster.yml \
+    --number-of-nodes 2 \
+    --app-template $PWD/application.yml \
+    --apps-per-node 5
+
+Note that all application instances will have exactly the same configuration.
```
jongiddy: Add "In particular, multiple containers may fail to start if they use a common host resource (e.g. host ports)."

admin/cluster_setup.py
```diff
@@ -0,0 +1,278 @@
...
+            lambda nodes: len(nodes) >= len(cluster.agent_nodes)
+        )
+        d.addErrback(write_failure, logger=None)
+        return d
+
+    got_nodes = loop_until(reactor, got_all_nodes, repeat(1, 300))
+
+    def do_configure(success):
+        if not success:
+            return 1
```
jongiddy: Rather than passing a non-zero return value through the `do_configure` function, raise exceptions as necessary.

Unconditionally set `result = 0` inside the `try...except` in `main`, and rely on the `except` clause to set result to 1.

jongiddy:
Thanks, Andriy. Please address 3 new comments, and merge when happy with tests.

docs/gettinginvolved/cluster-setup.rst
avg-I: That's a good warning to have here, thanks!

admin/cluster_setup.py
avg-I: okay!

jongiddy:
Remove the `sys.stderr.write`'s from `got_response`. Either add eliot logging (perhaps `with start_action...`), or just drop them all together - the error should get displayed from the raised exception.

avg-I:
okay

admin/cluster_setup.py
```diff
@@ -0,0 +1,286 @@
...
+        gather_deferreds(results)
+
+        if options['apps-per-node'] > 0:
+            config = _build_config(cluster, options['template'],
+                                   options['apps-per-node'])
+            yield _configure(reactor, cluster, config)
+
+        result = 0
+
+    except BaseException:
```
jongiddy: Is there a specific reason for changing to `BaseException`? Generally, you should use `Exception` as the highest-level exception to catch. The `BaseException` classes that are not `Exception`s are usually non-errors (`SystemExit`, `StopIteration`).

avg-I: Yes, it's explained a little bit here bb131ed424a5c7d. In `acceptance.py` we have a naked `except:` in that place.
Specifically, I've seen a situation where `GeneratorExit` was injected into `main`.  And we want to set `result` in those cases, so that we could clean up the cluster.

PR #2312, opened 2015-12-09
[FLOC-3597] Increase OS X timeout up to 60 minutes

jml:
We can spend an awful lot of time importing base box for the OS X image. Although we can probably do some things to make this faster, a good first step is to increase the timeout, thus avoiding spurious failures.


Azulinho:
thanks dude,

just that comment regarding the timeout change,
address, and merge when green.

PR #2313, opened 2015-12-09
[FLOC 3664] Remove the Vagrant Dev Box instructions from the docs

davidoliverSP2:
Fixes 3664 (a follow on from FLOC 3568)

The Vagrant dev box is broken, and there is no current plan to fix it, so this PR removes mentions of the Vagrant box from the Getting Involved sections of the docs.


jml:
LGTM. Thanks.

PR #2316, opened 2015-12-09
[FLOC-3666] Use https for dependency link.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3666

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,9 +112,9 @@ setup(
         # Use our fork of testtools until #165, #171, and #172 are merged and
         # released. See FLOC-3498.
         #
-        # "git+git" weirdness is due to setuptools expecting:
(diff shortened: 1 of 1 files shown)


sarum90:
Thanks, looks great.

We might not need to call "git+https" weirdness, it certainly doesn't look as odd as "git+git", but it's also fine as is.

I verified by spinning up a new vagrant box and running:

```
sudo apt-get install git
git clone https://github.com/ClusterHQ/testtools
```

This works even on a new VM, and I think it represents approximately what python does to install with git+https.

Thanks for making us more secure :+1:

sarum90:
As far as I'm concerned you can merge whenever. With this change it seems to me if you get any greens you probably did it correctly. I'll let you merge at your discretion.

PR #2317, opened 2015-12-09
[FLOC-3670] Define acceptance testing jobs more broadly

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3670

Maybe also puts less load on cloud providers to run our acceptance tests since it cuts the number of jobs from 9 to 3.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -789,15 +789,9 @@ run_trial_cli_as_root: &run_trial_cli_as_root [
 # run_acceptance_modules contains the list of Flocker modules to be executed
 # during the acceptance tests
 run_acceptance_modules: &run_acceptance_modules
-  - flocker.acceptance.endtoend.test_dataset
-  - flocker.acceptance.endtoend.test_diagnostics
-  - flocker.acceptance.endtoend.test_dockerplugin
-  - flocker.acceptance.endtoend.test_leases
-  - flocker.acceptance.integration.test_mongodb
-  - flocker.acceptance.integration.test_platform
-  - flocker.acceptance.integration.test_postgres
-  - flocker.acceptance.obsolete.test_cli
-  - flocker.acceptance.obsolete.test_containers
+  - flocker.acceptance.endtoend
+  - flocker.acceptance.integration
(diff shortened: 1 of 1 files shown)


sarum90:
Code looks great, looking at the jenkins-multijob it looks like there are fewer acceptance tests running now.

For reviewing this further I decided to think the advantages and disadvantages of doing this change:

Good:
- Less load on cloud providers (de-sharding reduces number of clusters brought up), which _might_ decrease flakiness.
  - We did this on Buildbot and it helped with RequestLimitExceeded errors.
  - We have seen RequestLimitExceeded errors on Jenkins. (Similar things we think on Rackspace)
- 1 Less thing that you have to remember if you are trying to add an acceptance test suite.

Bad:
* Decreasing sharding makes the unit of retry more course.
  * As in, if we had 3 tests each which would fail 1/2 times you probably could get them all green with just a few presses of retry. Now we have 1 test that fails 7/8 of the time, might be more painful.
  *  So long as our failure rates are low this is less of an issue.
- Decreases sharding which _might_ increase full test time.
  - Sorting by last duration, we see runs of endtoend taking 29min, 27min, 26min, and 24min. These are the tests that take the longest in the suite, so this change has potential to increase total test time.
  - Going to master, clicking on all of the current endtoend tests and then clicking on trends and then sorting by duration we can approximate how long these tests are currently taking. For passing test runs it looks like in general the average and median runtimes are much closer to 20-24 (just eyeballing like 10 different pages). Lets say this could add 5 minutes to test times if we had perfect parallelism.

I don't think we have perfect parallelism, in fact I think we might have more than 5 minutes of scheduling delay right now because we have so many jobs running. I might re-visit the trends of these tests after we submit to verify we aren't making things take too long.

Even if it did take 5 minutes longer, if this reduces overall flakiness it probably still is a benefit.

All told I think the good probably outweighs the bad. Certainly enough to try it.

Given that you're only failing test was run_lint, and this doesn't change any python I think I'm just going to merge this for you.

Thanks.

PR #2318, opened 2015-12-09
[FLOC-3669] Loopback acceptance tests in Jenkins

itamarst:
Run acceptance tests against loopback backend... hopefully. Theory is new Jenkins builds will appear.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -677,6 +677,7 @@ class RunOptions(Options):
         ['flocker-version', None, None, 'Version of flocker to install'],
         ['build-server', None, 'http://build.clusterhq.com/',
          'Base URL of build server for package downloads'],
+        ['number-of-nodes', None, 2, 'Number of nodes to launch.', int],
     ]
 
     optFlags = [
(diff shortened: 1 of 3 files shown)


build.yaml
```diff
@@ -1031,6 +1048,48 @@ job_type:
       clean_repo: true
       archive_artifacts: *acceptance_tests_artifacts_ubuntu_special_case
       timeout: 45
+    run_acceptance_loopback_on_AWS_CentOS_7_for:
+      on_nodes_with_labels: 'aws-centos-7-SELinux-T2Medium'
+      with_modules:
+        - flocker.acceptance
```
sarum90: :grey_question: Slight preference for extracting this into some variable "run_acceptance_modules_lumped" and/or adding a comment. Probably something to the effect of:

"For setups that run quickly enough we run all of the acceptance tests together rather than sharding them out across projects in Jenkins."

If that's _not_ the reason you lumped them together here then definitely add a comment.

sarum90:
This is excellent.

I love the idea of being able to more quickly run acceptance tests. Quickly running endtoend tests should drastically increase my speed of development.

New tests appear to exist and be passing on Jenkins. Please add a comment about why we are lumping all acceptance tests together for the loopback runs and then merge.

Thanks!

admin/acceptance.py
```diff
@@ -878,7 +879,7 @@ def _libcloud_runner(self, package_source, dataset_backend,
             dataset_backend=dataset_backend,
             dataset_backend_configuration=self.dataset_backend_configuration(),
             variants=self['variants'],
-            num_nodes=int(os.environ.get("FLOCKER_ACCEPTANCE_NUM_NODES", "2")),
+            num_nodes=self['number-of-nodes'],
```
exarkun: Argh.  While this is clearly an improvement, this means now we have to make sure there are no users of the environment variable that are going to break because of this change.

itamarst: OK, I added fallback and warning for people who still use it.

PR #2320, opened 2015-12-10
[FLOC-3672] Run agent tests as root.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3672

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -179,27 +179,7 @@ common_cli:
     # https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/20
     # So we will place the virtualenv in /tmp/v instead
     #
-    if is_osx; then
-      # OSX doesn't allow us to sudo (no tty)
-      # we can't be certain that cloudbees will always gives us a new OSX slave
-      # so we make sure we remove any old files
-      rm -f results.xml
-      rm -f trial.log
-      rm -rf _trial_temp/
-      rm -rf \${venv}
-      rm -f "Flocker*.rb"
-      rm -f "dist/Flocker-*.tar.gz"
-    fi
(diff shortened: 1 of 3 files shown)


build.yaml
```diff
@@ -179,26 +179,7 @@ common_cli:
...
-    if is_centos || is_ubuntu; then
-      # some of our tests (which run as root), leave some files behind when
-      # they abort or fail to complete.
-      # We use 'sudo' so that we can remove those files before a next job run.
-      sudo rm -rf \${venv}
-      sudo rm -f results.xml
-      sudo rm -f trial.log
-      sudo rm -rf _trial_temp/
-    fi
+    rm -rf \${venv}
```
tomprince: This always occurs after running `git clean -fxd` so there will never be extraneous files in the repository. Thus we only need to clean files that are outside of it.

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -34,6 +34,7 @@ def setUp(self):
         """
         Establish testing infrastructure for test cases.
         """
+        super(BlockDeviceManager, self)
```
tomprince: These tests never ran, so this failure was never noticed.

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -34,6 +34,7 @@ def setUp(self):
         """
         Establish testing infrastructure for test cases.
         """
+        super(BlockDeviceManagerTests, self).setUp()
```
tomprince: These tests never ran, so the fact that this didn't upcall was never noticed.

jml:
:+1: 

Merge when OK w/ test results.

PR #2321, opened 2015-12-10
[FLOC 3650] Update the intro to the API User Cert doc

davidoliverSP2:
Fixes 3650

The current wording allowed the user to believe that generating API User Cert was optional. This PR rewords the intro to make it clear all users must do this step.


jml:
All good. Thanks!

PR #2322, opened 2015-12-10
[FLOC 3632] Update the Full Install information 

davidoliverSP2:
Fixes 3632

The current wording allowed the user to believe that the installation of the plugin was a separate page - this is no longer true, as the Plugin installation is now an optional step in the node services instructions.


jml:
Lovely, thanks! Merge when passes doc tests.

PR #2323, opened 2015-12-10
[FLOC 3652] Rename the API User to API Client

davidoliverSP2:
Fixes 3652

"API User" was identified as implying a human user, whereas the API can actually be invoked by either a real user, or a machine. Renamed our use of the API user as the API client. 

In a few situations I specified "the API client, or end user"

### DIFF
--- a/docs/config/generate-api-certificates.rst
+++ b/docs/config/generate-api-certificates.rst
@@ -1,16 +1,16 @@
 .. _generate-api:
 
-==================================
-Generating an API User Certificate
-==================================
+====================================
+Generating an API Client Certificate
+====================================
 
-To send instructions to the control service, whether it is via the API directly, or the CLI, or by any other method, you will need to follow the instructions below to generate an API user certificate:
+To send instructions to the control service, whether it is via the API directly, or the CLI, or by any other method, you will need to follow the instructions below to generate an API client certificate:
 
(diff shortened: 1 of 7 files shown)


docs/config/generate-api-certificates.rst
```diff
@@ -1,16 +1,16 @@
...
-To send instructions to the control service, whether it is via the API directly, or the CLI, or by any other method, you will need to follow the instructions below to generate an API user certificate:
+To send instructions to the control service, whether it is via the API directly, or the CLI, or by any other method, you will need to follow the instructions below to generate an API client certificate:
 
-#. Generate an API user certificate:
+#. Generate an API client certificate:
 
    Run the following command from the directory which contains the certificate authority files generated when you first installed the cluster. For more information, see :ref:`authentication`.
 
-   Replace ``<username>`` with a unique username for an API user.
+   Replace ``<username>`` with a unique username for an API client.
```
jml: "unique identifier" maybe, suggesting username as an example?

Also, probably should change `username` to `client_name` in  the examples.

jml:
Thanks, that's much clearer, and (at least to me) seems to be more standard terminology.

I've made a couple of suggestions. Please apply them at your discretion and then merge when it passes doc tests.

PR #2324, opened 2015-12-10
[FLOC-3663] Restore Twisted log to _trial_temp/test.log

jml:
Re-enables logging in acceptance tests to go to _trial_temp/test.log. 

Uses the latest version of our fork of testtools (which applies testing-cabal/testtools#172) to disable the automatic suppression & capture of Twisted logs, and then manually invokes the fixture to capture the logs.

This means that the logs will appear as the `twisted-log` detail, and _also_ in `test.log`.

I couldn't think of a non-awful automated test for the latter, but I did a manual test and it appeared to work.

I've also added a couple of exceptions which I hope will do something to mitigate the awfulness of using the testtools fork.

### DIFF
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -11,7 +11,9 @@ import tempfile
 import testtools
 from testtools.content import text_content
 from testtools.deferredruntest import (
-    AsynchronousDeferredRunTestForBrokenTwisted)
+    AsynchronousDeferredRunTestForBrokenTwisted,
+    CaptureTwistedLogs,
+)
 
 from twisted.python.filepath import FilePath
 from twisted.trial import unittest
@@ -70,6 +72,10 @@ class AsyncTestCase(testtools.TestCase):
         super(AsyncTestCase, self).__init__(*args, **kwargs)
         # XXX: Work around testing-cabal/unittest-ext#60
         self.exception_handlers.insert(-1, (unittest.SkipTest, _test_skipped))
+
+    def setUp(self):
+        super(AsyncTestCase, self).setUp()
+        self.useFixture(CaptureTwistedLogs())
 
     def mktemp(self):
         """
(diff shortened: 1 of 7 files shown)


flocker/testtools/_base.py
```diff
@@ -11,7 +11,20 @@
...
+
+try:
+    from testtools.deferredruntest import CaptureTwistedLogs
+except ImportError:
+    # We are using a fork of testtools, which unfortunately means that we need
+    # to do special things to make sure we're using the latest version. Raise
+    # an error message that will help people figure out what they need to do.
+    raise Exception(
+        'Cannot import CaptureTwistedLogs. Maybe upgrade your version of '
+        'testtools: pip install --upgrade --process-dependency-links .[dev]'
```
sarum90: :grey_question: Optional for this PR, but commenting anyway because you added this Exception.

doing git grep 'pip install' reveals a lot of docs that should be updated to have "--process-dependency-links" and things that should be changed from Flocker[dev] to .[dev]  (I think).

Specifically:
- tox.ini
- CONTRIBUTING.rst
- docs/gettinginvolved/*

Might make more sense as a follow up. Right now I jut went to Slack to look up that line, and I realized it'd be nicer if it was in our official docs.

sarum90:
Thanks so much, logs are great, even for passing tests :smile: 

VerifiedThat:
* `pip install --upgrade --process-dependency-links .[dev]` can install latest version from github.
* running flocker.acceptance.endtoend.test_dataset.DatasetAPITests now has logs in _trial_temp/test.log

Thanks so much!

Tests are green except for Rackspace, address the one optional comment if you want and then merge whenever.

flocker/testtools/_base.py
jml: Done. I'll wait for a successful build of docs.

PR #2326, opened 2015-12-10
[FLOC-3291] Benchmark read request parametization

jongiddy:
This PR allows the user to select a method to be called during a benchmark read request operation.  The allowed methods are the zero-parameter methods of the interface `IFlockerAPIV1Client`.

The default method is the relatively simple `version` method, but other methods can be set to observe if they behave differently during the benchmarks.

### DIFF
--- a/benchmark/operations/read_request.py
+++ b/benchmark/operations/read_request.py
@@ -2,7 +2,36 @@ from zope.interface import implementer
 
 from twisted.internet.defer import succeed
 
+from flocker.apiclient import IFlockerAPIV1Client
+
 from .._interfaces import IProbe, IOperation
+
+
+class InvalidMethod(Exception):
+    """
+    Method not suitable for use with ReadRequest operation.
+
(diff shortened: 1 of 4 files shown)


benchmark/operations/read_request.py
```diff
@@ -2,21 +2,49 @@
...
+    :raise InvalidMethod: if name is not valid or requires parameters.
+    """
+    for name, method in interface.namesAndDescriptions():
+        if name == method_name:
+            if len(method.getSignatureInfo()['required']) > 0:
+                raise InvalidMethod('Require no-arg method')
+            return
+    raise InvalidMethod(
+        'Method not found in interface {}'.format(interface.__name__)
+    )
```
jml: Please include `method_name` in the exception message.

Optionally, change `InvalidMethod` to take the interface & method as parameters and create the message there.

jml:
Nice, thanks. Please address comments & merge when tests are OK.

PR #2328, opened 2015-12-10
[FLOC-3678] Fix tox to work with forked testtools.

itamarst:
You can test by running `tox -e py27`, `tox -e sphinx`, `tox -e admin` - previously they'd blow up because they couldn't find forked testtools.


jml:
Great, thanks.

PR #2329, opened 2015-12-10
[FLOC-3647] List live nodes

itamarst:
This is part 1 of a bigger design to fix issue where if a node is shutdown attached volumes are never detached . This adds ability to list live nodes in the cluster. Once we know that we can discover if a dataset is attached to a dead node (part 2) and then detach it remotely by another node if it's moved (part 3).

Note that I haven't actually verified in any way that the code works, this PR is first live run :grinning: 

Builds on #2287 .


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5001,3 +5004,40 @@ def test_get_device_path_until_detach(self):
...
+    """
+    :param blockdevice_api_factory: A factory which will be called
+        with the generated ``TestCase`` during the ``setUp`` for each
+        test and which should return a provider of both ``IBlockDeviceAPI``
+        and ``ICloudAPI`` to be tested.
+
+    :returns: A ``TestCase`` with tests that will be performed on the
+       supplied ``IBlockDeviceAPI``/``ICloudAPI`` provider.
+    """
+    class Tests(SynchronousTestCase):
```
jml: Shouldn't these be _asynchronous_ tests? They are running threads and rely on what looks like the real reactor.

flocker/node/agents/blockdevice.py
```diff
@@ -1102,6 +1102,39 @@ def get_device_path(blockdevice_id):
...
+    In particular the presumption is that nodes are also managed by a
+    centralized infrastructure.
+    """
+    def list_live_nodes():
+        """
+        Return the compute IDs of all nodes that are currently up and running.
+
+        :returns: A collection of compute instance IDs, compatible with
+            those returned by ``IBlockDeviceAPI.compute_instance_id``.
+        """
```
jml: :question: Maybe add a fake implementation?

setup.cfg
```diff
@@ -11,3 +11,7 @@ versionfile_build = flocker/_version.py
 tag_prefix =
 parentdir_prefix = Flocker-
 
+[flake8]
+# Ignore the following annoying pep8 errors:
+# E402: module level import not at top of file
+ignore = E402
```
jml: I think this triggers a break in the lint builder: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/list-live-nodes-3467/job/run_lint/2/console

jml:
Looks like a good move. Please address comments and merge after fixing the lint tests, unless you'd particularly welcome another round of review.

flocker/node/agents/blockdevice.py
itamarst: Going to have one in next stacked branch, IIRC.

PR #2331, opened 2015-12-10
[FLOC-3386] Calculate desired datasets floc 3386

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3386

The code changes are probably easiest reviewed looking directly at the [code](https://github.com/ClusterHQ/flocker/blob/calculate-desired-datasets-FLOC-3386/flocker/node/agents/blockdevice.py#L1635-L1738) rather than the diff since github does poorly at generating a sensible diff.

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -691,6 +691,7 @@ class AttachVolume(PClass):
 
 
 @implementer(IStateChange)
+@provider(IDatasetStateChangeFactory)
 class ActionNeeded(PClass):
     """
     We need to take some action on a dataset but lack the necessary
@@ -708,6 +709,12 @@ class ActionNeeded(PClass):
     # Nominal interface compliance; we don't expect this to be ever run,
     # it's just a marker object basically.
     eliot_action = None
+
+    @classmethod
+    def from_state_and_config(cls, discovered_dataset, desired_dataset):
+        return cls(
+            dataset_id=discovered_dataset.dataset_id,
+        )
 
     def run(self, deployer):
         """
@@ -1350,8 +1357,9 @@ Desired = Discovered = DatasetStates
 DATASET_TRANSITIONS = {
     Desired.MOUNTED: {
         Discovered.NON_EXISTENT: CreateBlockDeviceDataset,
-        # Other node will need to deatch first
-        Discovered.ATTACHED_ELSEWHERE: DoNothing,
+        # Other node will need to deatch first, but we we need to
+        # wake up to notice that it has detached.
+        Discovered.ATTACHED_ELSEWHERE: ActionNeeded,
         Discovered.ATTACHED_NO_FILESYSTEM: CreateFilesystem,
         Discovered.NON_MANIFEST: AttachVolume,
         DatasetStates.ATTACHED: MountBlockDevice,
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -1276,8 +1276,6 @@ class BlockDeviceDeployerLocalState(PClass):
     hostname = field(type=unicode, mandatory=True)
     node_uuid = field(type=UUID, mandatory=True)
     datasets = pmap_field(UUID, DiscoveredDataset)
-    # XXX This should go away in FLOC-3386
-    volumes = pvector_field(BlockDeviceVolume)
 
     def shared_state_changes(self):
         """
@@ -1619,8 +1617,6 @@ class BlockDeviceDeployer(PClass):
             node_uuid=self.node_uuid,
             hostname=self.hostname,
             datasets=datasets,
-            # XXX This should go away in FLOC-3386
-            volumes=raw_state.volumes,
         )
 
         return succeed(local_state)
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -119,6 +119,7 @@ class DesiredDataset(PClass):
     state = field(
         invariant=lambda state: (state in DatasetStates.iterconstants(),
                                  "Not a valid state"),
+        mandatory=True,
     )
     dataset_id = field(type=UUID, mandatory=True)
     maximum_size = field(type=int)
@@ -129,6 +130,15 @@ class DesiredDataset(PClass):
     mount_point = field(FilePath)
     filesystem = field(unicode, initial=u"ext4", mandatory=True,
                        invariant=lambda v: (v == "ext4", "Must be 'ext4'."))
+
+    __invariant__ = TaggedUnionInvariant(
+        tag_attribute='state',
+        attributes_for_tag={
+            DatasetStates.NON_MANIFEST: {"maximum_size"},
+            DatasetStates.MOUNTED: {"maximum_size", "mount_point"},
+            DatasetStates.DELETED: set(),
+        },
+    )
 
 
 class IDatasetStateChangeFactory(Interface):
@@ -1439,6 +1449,8 @@ class BlockDeviceDeployer(PClass):
         ``None`` in real-world use.
     :ivar block_device_manager: An ``IBlockDeviceManager`` implementation used
         to interact with the system regarding block devices.
+    :ivar ICalculator calculator: The object to use to calculate dataset
+        changes.
     """
     hostname = field(type=unicode, mandatory=True)
     node_uuid = field(type=UUID, mandatory=True)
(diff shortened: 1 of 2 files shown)


itamarst:
Possibly from previous pieces, but I noticed that `from_state_and_config` is not documented as taking `None`, but will end up doing so. And it works, presumably, but the difference between interface, actual usage, and expectations of implementations might lead to bugs in the future.

flocker/node/agents/blockdevice.py
```diff
@@ -1639,311 +1647,95 @@ def _calculate_desired_for_manifestation(self, manifestation):
...
-                # It's already attached here.
-                continue
-            if manifestation.dataset.deleted is True:
-                # Don't attach a dataset to be deleted.
-                continue
-            if manifestation.dataset_id in nonmanifest:
-                volume = _blockdevice_volume_from_datasetid(volumes,
-                                                            dataset_id)
-                if volume is None:
-                    # We know something is non-manifest, but lack any info
```
itamarst: Is this edge case still handled?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1043,6 +1032,18 @@ class UnusableAPI(object):
     """
 
 
+@implementer(ICalculator)
+class RecordingCalculator(object):
```
itamarst: Docstring.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1094,6 +1095,14 @@ class BlockDeviceCalculatorInterfaceTests(
     """
 
 
+class RecordingCalculatorInterfaceTests(
+    make_icalculator_tests(lambda: RecordingCalculator(NoOp()))
+):
+    """
+    Tests for ``BlockDeviceCalculator``'s implementation of ``ICalculator``.
```
itamarst: This docstring is wrong.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+    deployer,
+    desired_manifestations=(),
+    local_datasets=(),
+    local_applications=(),
+    additional_node_config=set(),
+    expected_datasets=(),
+    leases=Leases(),
+):
+    """
+    Assert that ``calculate_changes`` calculates the given desired datasetsg
```
itamarst: Typo at end of line.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+    invoked with the given state and configuration.
+
+    :param TestCase test_case: The ``TestCase`` which is being run.
+    :param BlockDeviceDeployer deployer: The deployer that will be asked to
+        calculate the desired datasets.
+    :param desired_manifeations: Manifestations to include in the local nodes
```
itamarst: Typo in parameter name.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+        If there is a manifesation on this node, then the
+        corresponding dataset has a desired state of ``MOUNTED``.
+        """
+        assert_desired_datasets(
+            self, self.deployer,
+            desired_manifestations=[ScenarioMixin.MANIFESTATION.transform(
+                ['dataset', 'metadata'], ScenarioMixin.METADATA,
+            )],
+            expected_datasets=[
+                DesiredDataset(
```
itamarst: This chunk seems like it's repeated a lot, perhaps we can have a default expected `DesiredDataset` which gets reused in the test bodies?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+                    maximum_size=int(REALISTIC_BLOCKDEVICE_SIZE.bytes),
+                    mount_point=self.deployer.mountroot.child(
+                        unicode(ScenarioMixin.DATASET_ID)
+                    ),
+                ),
+            ],
+        )
+
+    def test_manifestation_default_size(self):
+        """
```
itamarst: Docstirng should mention this is testing for size.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+        assert_desired_datasets(
+            self, self.deployer,
+            desired_manifestations=[
+                ScenarioMixin.MANIFESTATION.transform(
+                    ["dataset", "deleted"], True,
+                    ['dataset', 'metadata'], ScenarioMixin.METADATA,
+                ),
+            ],
+            # XXX FLOC-1772 We only need to do this to handle
```
itamarst: FLOC-1772 has been fixed, I believe, so can just delete this comment? Or reduce number of tests?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+            ],
+        )
+
+    def test_deleted_attached_manifest_dataset(self):
+        """
+        If there is a non-manifest dataset that is deleted on this
+        node, the corresponding dataset has a desired state of
+        ``DELETED``.
+
+        This is only necessary until we can delete non-manifest datasets
```
itamarst: 1772 is fixed.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+                    metadata=ScenarioMixin.METADATA,
+                ),
+            ],
+        )
+
+    def test_leased_mounted_manifestaion(self):
+        """
+        If there is a lease for a mounted dataset present on node,
+        there is a corresponding dataset that has a desired state of
+        ``MOUNTED``.
```
itamarst: add something like "even if the configuration does not mention the dataset, since the lease keeps it from being moved."

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+                ),
+            ],
+            leases=Leases().acquire(
+                now=datetime.now(tz=UTC),
+                dataset_id=ScenarioMixin.DATASET_ID,
+                node_id=self.deployer.node_uuid,
+            )
+        )
+
+    def test_leased_attached_manifestaion(self):
```
itamarst: Typo.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+        """
+        If there is a lease for an attached dataset present on node,
+        there is not a corresponding desired dataset.
```
itamarst: Add explanation that this is in context of not mentioning dataset in configuration.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+                now=datetime.now(tz=UTC),
+                dataset_id=ScenarioMixin.DATASET_ID,
+                node_id=self.deployer.node_uuid,
+            )
+        )
+
+    def test_lease_elsewhere(self):
+        """
+        If there is a lease for a dataset on another node,
+        there isn't a corresponding dataset that has a desired state
```
itamarst: This says "isn't", test body *does* have it though....

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1273,483 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+                dataset_id=ScenarioMixin.DATASET_ID,
+                node_id=self.deployer.node_uuid,
+            )
+        )
+
+    def test_deleted_leased_manifestation(self):
+        """
+        If there is a manfestation on this node that is deleted and
+        there is a lease on the volume for this node, the
+        corresponding dataset has a desired state of ``DELETED``.
```
itamarst: Docstring doesn't match test body.

itamarst:
So close to being done! and so much better than when you started.

A bunch of fairly minor comments above to address, in particular:

1. I worry about the edge case fixed in https://clusterhq.atlassian.net/browse/FLOC-3487 may have been reintroduced (ActionNeeded thing in attach code.) There are in fact `DatasetAlreadyExists` errors in one of the AWS acceptance logs, though I haven't looked much further into it.
2. The tests for leases interaction are somewhat confusing in their descriptions, and sometimes don't match the code. The code may well be correct, but it'd be easier to assess if those were clarified. Also if they were made a little less repetitive by re-using a `DesiredDataset` object.

Please address and resubmit.

flocker/node/agents/blockdevice.py
tomprince: This edge case occurred because we used the data on non-manifest datasets from the control node, but the list of block-device volumes from discovery. Those could get out of sync leading to this edge case. With this change, we only ever look at state discovered locally to make decisions, so the edge case no longer exists.

flocker/node/agents/test/test_blockdevice.py
tomprince: I got tripped up by this too. It has a discovered dataset, but not a desired dataset. I've adjusted the docstring slightly, but I'm open to improvements to clarify this.

tomprince:
> There are in fact DatasetAlreadyExists errors in one of the AWS acceptance logs, though I haven't looked much further into it.

Those are client `DatasetAlreadyExists` errors from the docker plugin, so unrelated. They also occur in master.

flocker/node/agents/blockdevice.py
itamarst: Ah, in that case we have a different issue :grinning: 

When the convergence loop gets an updated global cluster state it runs a calculate_changes with cached local state, to see if this change might lead to some actions. If it does the convergence loop wakes up and does real local discovery and calculation and actually runs local actions.

But... if we never look at global cluster state for non-manifest datasets in this speculative should-I-wakeup calculation then we won't notice we need to wake up when remote node detaches something we want to attach locally.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1278,326 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+            leases=Leases().acquire(
+                now=datetime.now(tz=UTC),
+                dataset_id=ScenarioMixin.DATASET_ID,
+                node_id=uuid4(),
+            )
+        )
+
+    def test_application_mounted_manifestation(self):
+        """
+        If there is an application with attached volume, there is a
```
itamarst: :question: This doesn't quite match existing behavior, which merely prevents detaching. But... lease implies you want to use it? Not sure this is particularly a problem.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1264,6 +1278,326 @@ def test_all_transitions(self, desired_state, discovered_state):
...
+            leases=Leases().acquire(
+                now=datetime.now(tz=UTC),
+                dataset_id=ScenarioMixin.DATASET_ID,
+                node_id=self.deployer.node_uuid,
+            )
+        )
+
+    def test_deleted_leased_manifestation(self):
+        """
+        If there is a manfestation on this node that is deleted and there is a
```
itamarst: This however seems like it'll result in dataset being recreated, which I don't think we want. So this seems wrong.

itamarst:
Assuming green build, one major issue left: making sure we wake up convergence loop if we want to move volume locally and have just been told it's no longer attached. Address that, any minor comments above, then merge.

flocker/node/agents/test/test_blockdevice.py
tomprince: I don't *think* this is a change in behavior. This logic triggers if
- there is a lease on this node or a running (not configured) application using it
*and*
- we discovered the dataset is already mounted.

PR #2333, opened 2015-12-11
[FLOC-3676] Separate Eliot logs from Twisted logs in acceptance test failures

jml:
testtools tests can attach content that shows up when tests fail. This PR attaches the Twisted log _but_ it processes the Twisted log, separating out the `ELIOT` lines from the rest of the log and pretty prints the Eliot logs.

### DIFF
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -94,6 +94,11 @@ class AsyncTestCase(testtools.TestCase):
     def setUp(self):
         super(AsyncTestCase, self).setUp()
         # Need this to run _after_ the clean-up in CaptureTwistedLogs.
+        #
+        # Would ideally like to move post-processing to its own fixture that
+        # wraps up CaptureTwistedLogs, but there doesn't seem to be a way to
+        # do that. https://github.com/testing-cabal/fixtures/pull/20 for
+        # details.
         self.addCleanup(self._post_process_twisted_logs)
         self.useFixture(CaptureTwistedLogs())
 
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -196,7 +196,6 @@ def _prettyformat_lines(lines):
 
 
 _ELIOT_MARKER = ' [-] ELIOT: '
-_ELIOT_MARKER_LENGTH = len(_ELIOT_MARKER)
 
 
 def _get_eliot_data(twisted_log_line):
@@ -210,10 +209,9 @@ def _get_eliot_data(twisted_log_line):
         ``None``.
     :rtype: unicode or ``NoneType``.
     """
-    index = twisted_log_line.find(_ELIOT_MARKER)
-    if index < 0:
-        return None
-    return twisted_log_line[index + _ELIOT_MARKER_LENGTH:].strip()
+    _, _, eliot_data = twisted_log_line.partition(_ELIOT_MARKER)
+    if eliot_data:
+        return eliot_data.strip()
 
 
 def _iter_content_lines(content):
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -7,9 +7,11 @@ Base classes for unit tests.
 from datetime import timedelta
 from functools import partial
 from itertools import tee
+import json
 import sys
 import tempfile
 
+from eliot.prettyprint import pretty_format
 import testtools
 from testtools.content import Content, text_content
 from testtools.content_type import UTF8_TEXT
@@ -134,7 +136,10 @@ def _fix_twisted_logs(log_content):
     twisted_lines, eliot_lines = tee(_iter_content_lines(log_content))
     return (
         Content(UTF8_TEXT, partial(_pure_twisted_logs, twisted_lines)),
-        Content(UTF8_TEXT, partial(_iter_eliot_logs, eliot_lines)),
+        Content(
+            UTF8_TEXT,
+            lambda: _prettyformat_lines(_iter_eliot_logs(eliot_lines)),
+        ),
     )
 
 
@@ -144,6 +149,15 @@ def _pure_twisted_logs(log_lines):
     are not Eliot data.
     """
     return (line for line in log_lines if _get_eliot_data(line) is None)
+
+
+def _prettyformat_lines(lines):
+    """
+    Pretty format lines of Eliot logs.
+    """
+    for line in lines:
+        data = json.loads(line)
+        yield pretty_format(data)
 
 
 def _iter_eliot_logs(log_lines):
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -178,7 +178,7 @@ def _prettyformat_lines(lines):
     """
     for line in lines:
         data = json.loads(line)
-        yield pretty_format(data)
+        yield pretty_format(data) + '\n'
 
 
 _ELIOT_MARKER = ' [-] ELIOT: '
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -224,12 +224,12 @@ def _iter_content_lines(content):
     return _iter_lines(content.iter_bytes(), '\n')
 
 
-def _iter_lines(byte_iter, separator='\n'):
+def _iter_lines(byte_iter, line_separator):
     """
     Iterate over the lines that make up ``content``.
 
     :param iter(bytes) byte_iter: An iterable of bytes.
-    :param bytes separator: A single byte that marks the end of a line.
+    :param bytes line_separator: The bytes that mark the end of a line.
     :yield: Separator-terminated bytestrings.
     """
     # XXX: Someone must have written this before.
@@ -237,13 +237,12 @@ def _iter_lines(byte_iter, separator='\n'):
     chunks = []
     for data in byte_iter:
         while data:
-            index = data.find(separator)
-            if index < 0:
-                chunks.append(data)
+            head, sep, data = data.partition(line_separator)
+            if not sep:
+                chunks.append(head)
                 break
 
-            head, data = data[:index + 1], data[index + 1:]
-            chunks.append(head)
+            chunks.append(head + sep)
             yield ''.join(chunks)
             chunks = []
 
--- a/flocker/testtools/_base.py
+++ b/flocker/testtools/_base.py
@@ -142,22 +142,34 @@ def _fix_twisted_logs(log_content):
         messages and the second containing line-separated Eliot JSON messages.
     :rtype: (Content, Content)
     """
-    twisted_lines, eliot_lines = tee(_iter_content_lines(log_content))
+    twisted_lines, eliot_lines = _split_map_maybe(
+        _get_eliot_data, _iter_content_lines(log_content))
     return (
-        Content(UTF8_TEXT, partial(_pure_twisted_logs, twisted_lines)),
-        Content(
-            UTF8_TEXT,
-            lambda: _prettyformat_lines(_iter_eliot_logs(eliot_lines)),
-        ),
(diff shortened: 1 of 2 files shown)


flocker/testtools/_base.py
```diff
@@ -77,6 +82,8 @@ class AsyncTestCase(testtools.TestCase):
     Base class for asynchronous test cases.
```
tomprince: This functionality will definitely also be useful for synchronous test cases.

flocker/testtools/_base.py
```diff
@@ -86,8 +93,25 @@ def __init__(self, *args, **kwargs):
 
     def setUp(self):
         super(AsyncTestCase, self).setUp()
+        # Need this to run _after_ the clean-up in CaptureTwistedLogs.
```
tomprince: I'm guessing that this comment is explaining why this cleanup is before the `useFixture`, since I'm assuming cleanups are run in reverse order to when they are added. If that is the case, I think you should make that explanation explicit.

flocker/testtools/_base.py
```diff
@@ -86,8 +93,25 @@ def __init__(self, *args, **kwargs):
...
+        #
+        # Would ideally like to move post-processing to its own fixture that
+        # wraps up CaptureTwistedLogs, but there doesn't seem to be a way to
+        # do that. https://github.com/testing-cabal/fixtures/pull/20 for
+        # details.
```
tomprince: Even if this can't be a fixture wrapping another fixture, would it make sense to still abstract it as a fixutre that needs to be added before `CaptureTwistedLogs`?

flocker/testtools/_base.py
```diff
@@ -104,6 +128,112 @@ def mktemp(self):
...
+        ),
+    )
+
+
+def _pure_twisted_logs(log_lines):
+    """
+    Given an iterable of lines of bytes of Twisted logs, yield the lines that
+    are not Eliot data.
+    """
+    return (line for line in log_lines if _get_eliot_data(line) is None)
```
tomprince: :question: It probably doesn't matter, since this code isn't performance sensitive, but it looks like `_get_eliot_data` is being called twice on every line.

You could perhaps do something like
```
lines_with_eliot_data = (line, get_eliot_data(line)) for line in _iter_content_lines(log_content)
t, e = tee(lines_with_eliot_data)
twisted_lines = line for (line, eliot_data) in t if eliot_data is None
eliot_lines = eliot_datat for (line, eliot_data in e if eliot_data is not None
```
but I'm not sure if the additional complication is worth it.

flocker/testtools/_base.py
```diff
@@ -104,6 +128,112 @@ def mktemp(self):
...
+
+
+def _prettyformat_lines(lines):
+    """
+    Pretty format lines of Eliot logs.
+    """
+    for line in lines:
+        data = json.loads(line)
+        yield pretty_format(data)
```
tomprince: `eliot-prettyprint` puts a blank line between stanzas. Is it worth doing that here?

flocker/testtools/_base.py
tomprince: Actually, as I commented [here](https://github.com/testing-cabal/fixtures/pull/20#issuecomment-164371596) I think it is actually possible to post-process details. You just can't use a cleanup to do it.

tomprince:
Given that I think it is possible to make a fixture that does this, please change the code to do so, and re-submit.

A couple of other points to consider as follow ups:
- I wonder if eliot-tree style logs would be more useful.
- It would be nice if the conversion to junitxml preserved this information in some usable way, so that this collected information where available somewhere other than in the console logs.

tomprince:
One thing just occurred to me: we go through [some trouble](https://github.com/ClusterHQ/flocker/blob/master/flocker/__init__.py#L51-L65) to get the eliot log output into the twisted logs for tests and then here we are turning around and trying to extract that information. It seems like it would be simpler to just capture the eliot logs here directly. There are a couple of problems with this thought be directly actionable though:
- we want to capture eliot logs for all tests, not just failing ones
- this code doesn't preserve the raw format of the eliot logs, so the output can't be used by our log-processing tools.

flocker/testtools/_base.py
jml: Absolutely. Likewise the `mktemp` mixin. The [plan](https://clusterhq.atlassian.net/browse/FLOC-3281) is to move synchronous test case over to testtools once we've got a regression-free `AsyncTestCase`. We are very, very close to this.

Added a comment here along these lines.

flocker/testtools/_base.py
jml: That's right. Comment added.

flocker/testtools/_base.py
jml: I did this, putting the logic in a separate function. Maybe a bit abstronauty.

flocker/testtools/_base.py
jml: Why not? Done.

flocker/testtools/_base.py
jml: > Even if this can't be a fixture wrapping another fixture, would it make sense to still abstract it as a fixutre that needs to be added before CaptureTwistedLogs?

I don't think so, actually. Doing so ends up with two very closely intertwined fixtures that appear independent.

> Actually, as I commented here I think it is actually possible to post-process details. You just can't use a cleanup to do it.

Noted. See below.

jml:
> Given that I think it is possible to make a fixture that does this, please change the code to do so, and re-submit.

Done.

FWIW, I think the hoops that I had to jump through to make this work are not actually worth it, and prefer the earlier non-Fixture code use.

> A couple of other points to consider as follow ups:
> 
> * I wonder if eliot-tree style logs would be more useful.

I also wonder this. I considered using eliot-tree in this patch, but opted for pretty-print since it has far fewer failure conditions to deal with. I'm keen to pursue this later.

> * It would be nice if the conversion to junitxml preserved this information in some usable way, so that this collected information where available somewhere other than in the console logs.

Yes, it would. There's a whole bunch of work in that pipeline that would be valuable to pursue. (e.g. it would also be nice to emit subunit2 output from the reporter).

> One thing just occurred to me: we go through some trouble to get the eliot log output into the twisted logs for tests and then here we are turning around and trying to extract that information. It seems like it would be simpler to just capture the eliot logs here directly. There are a couple of problems with this thought be directly actionable though:
>
> * we want to capture eliot logs for all tests, not just failing ones
> * this code doesn't preserve the raw format of the eliot logs, so the output can't be used by our log-processing tools.

Yes, the thought occurred to me also, although my reason for not pursuing it was different: I think a change to `_redirect_eliot_logs_for_trial` would be much more likely to disrupt other people's development than the approach taken in this patch. I do think it would be an interesting thing to revisit later.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,166 @@ def mktemp(self):
...
+            installed. The fixture will mutate the ``getDetails()`` dict
+            of this test case.
+        """
+        super(_SplitEliotLogs, self).__init__()
+        self._case = case
+
+    def _setUp(self):
+        # Need the cleanups in this to run *after* the cleanup in
+        # CaptureTwistedLogs, so add it first, because cleanups are run in
+        # reverse order.
```
tomprince: This comment is superfluous now.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,166 @@ def mktemp(self):
...
+
+        # The trick here is that we can't call detailed.getDetails() directly.
+        # We can only call it inside the iter_bytes of the Content objects
+        # that we add. This is because the only time that we *know* the
+        # details are populated is when the details are evaluated. If we call
+        # it in _setUp(), the logs are empty. If we call it in cleanup, the
+        # detail is gone.
+
+        # XXX: Use a temporary name for the log detail that we'll fix up in
+        # cleanup. If we try to use `detail_name`, we get infinite recursion.
```
tomprince: As are both of these.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,153 @@ def mktemp(self):
         return make_temporary_directory(self).child('temp').path
 
 
+class _SplitEliotLogs(Fixture):
+    """
+    Split the Eliot logs out of Twisted logs.
+
+    Assumes that Twisted logs contain Eliot logs as per
+    ``flocker._redirect_eliot_logs_for_trial``, and that these logs have been
+    attached to a test case as a detail named with the value of
```
tomprince: It doesn't assume that they have been so attached, it ensure that they are itself.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,153 @@ def mktemp(self):
...
+    If there is no Eliot message on that line, return ``None``.
+
+    :return: A logged eliot message without Twisted logging preamble, or
+        ``None``.
+    :rtype: unicode or ``NoneType``.
+    """
+    index = twisted_log_line.find(_ELIOT_MARKER)
+    if index < 0:
+        return None
+    return twisted_log_line[index + _ELIOT_MARKER_LENGTH:].strip()
```
tomprince: :question: This could be also written as
```python
_, _, eliot_data = twisted_log_line.partition(_ELIOT_MARKGER)
if eliot_data:
    return eliot_data.strip()
else:
    return None
```
I'm not sure which is clearer.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,153 @@ def mktemp(self):
...
+    """
+    Iterate over the lines that make up ``content``.
+
+    :param Content content: Arbitrary newline-separated content.
+    :yield: Newline-terminated bytestrings that make up the content.
+    """
+    return _iter_lines(content.iter_bytes(), '\n')
+
+
+def _iter_lines(byte_iter, separator='\n'):
```
tomprince: :question: I'm curious about having both a default value for `separator` and passing it explicitly.

flocker/testtools/_base.py
```diff
@@ -104,6 +111,153 @@ def mktemp(self):
...
+    Assumes that Twisted logs contain Eliot logs as per
+    ``flocker._redirect_eliot_logs_for_trial``, and that these logs have been
+    attached to a test case as a detail named with the value of
+    ``CaptureTwistedLogs.LOG_DETAIL_NAME``.
+
+    Takes the Eliot logs that are in the Trial logs and splits them into a
+    separate detail that contains only the pretty printed Eliot logs.
+    """
+
+    _ELIOT_LOG_DETAIL_NAME = 'twisted-eliot-log'
```
tomprince: Is there anything particularly "twisted" about the log?

flocker/testtools/_base.py
jml: Ah yes, old iteration. Amended.

flocker/testtools/_base.py
jml: Removed.

flocker/testtools/_base.py
jml: Thanks! I thought there was something like this on `str`, but couldn't remember what it was called.

flocker/testtools/_base.py
jml: The provenance. The attached eliot logs come from the Twisted logging system, as opposed to somewhere else. 

Merging as-is, since it's easy to change later.

flocker/testtools/_base.py
jml: I've dropped the default and renamed to `line_separator` so I don't have to go to the hassle of thinking of a more appropriate name than `_iter_lines`. Have also updated the implementation to use `partition` which makes it clearer, allows `line_separator` to be multiple bytes and maybe even makes it faster.

PR #2336, opened 2015-12-11
[FLOC-3691] Add an invariant for `providedBy`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3691

### DIFF
--- a/flocker/common/_interface.py
+++ b/flocker/common/_interface.py
@@ -75,6 +75,6 @@ def provides(interface):
             return (False, "{value!r} doesn't provide {interface}".format(
                 value=value, interface=interface_name,
             ))
-    invariant.__name__ = "{interface}_invariant".format(interface_name)
+    invariant.__name__ = "provides_{}_invariant".format(interface_name)
 
     return invariant
--- a/flocker/common/_interface.py
+++ b/flocker/common/_interface.py
@@ -54,3 +54,27 @@ def interface_decorator(decorator_name, interface, method_decorator,
             setattr(cls, name, method_decorator(name, *args, **kwargs))
         return cls
     return class_decorator
+
+
+def provides(interface):
+    """
+    Create an invariant that asserts that the given value provides the given
+    interface.
+
+    :param InterfaceClass interface: The interface to check for.
+
+    :return: A function that takes an object and returns a tuple of `bool` and
+        a error message.
+    """
+    interface_name = interface.__name__
+
+    def invariant(value):
+        if interface.providedBy(value):
+            return (True, "")
+        else:
+            return (False, "{value!r} doesn't provide {interface}".format(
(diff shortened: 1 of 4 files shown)


flocker/common/_interface.py
```diff
@@ -54,3 +54,27 @@ def class_decorator(cls):
...
+
+    :param InterfaceClass interface: The interface to check for.
+
+    :return: A function that takes an object and returns a tuple of `bool` and
+        a error message.
+    """
+    interface_name = interface.__name__
+
+    def invariant(value):
+        if interface.providedBy(value):
```
sarum90: :grey_question: Is there a reason not to just do:

```
return (interface.providedBy(value), "{value!r} doesn't provide {interface}".format(
            value=value, interface=interface_name,
        ))
```

The API seems to suggest that's the more standard way to do it:
http://pyrsistent.readthedocs.org/en/latest/intro.html#invariants

flocker/common/test/test_interface.py
```diff
@@ -132,3 +133,54 @@ def test_raise(self, logger):
...
+        """
+        self.assertRaises(
+            InvariantException,
+            InterfaceHolder, value=object(),
+        )
+
+    def test_invariant_message(self):
+        """
+        When an object not providing the given interface is provided,
+        an ``InvariantException`` is raised.
```
sarum90: :grey_question: docstring doesn't seem to be updated for this test method.

flocker/common/test/test_interface.py
```diff
@@ -132,3 +133,54 @@ def test_raise(self, logger):
...
+        exception = self.assertRaises(
+            InvariantException,
+            InterfaceHolder, value=True,
+        )
+        self.assertEqual(
+            exception.invariant_errors,
+            ("True doesn't provide IDummy",),
+        )
+
+    def test_invariant_name(self):
```
sarum90: :grey_question: Do we require a docstring on all test methods? it looks like we do, but I'm not sure what it adds in this case. Name is rather descriptive, test is simple... Fix if you desire to and just missed this one, I'm indifferent.

sarum90:
That's a nice little utility :smile: 

I've got a couple optional docs-related queries for you to glance over. Address them or not at your discretion. Otherwise, merge when you believe the tests to be green-enough.

flocker/common/_interface.py
tomprince: I don't want to pay the cost of generating the representation if it isn't needed.

tomprince: https://github.com/tobgu/pyrsistent/issues/75

flocker/common/test/test_interface.py
tomprince: We do, I just missed adding it. Thanks for pointing it out. :+1:

PR #2319, opened 2015-12-09
[FLOC-3671] ICalculater 

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3671

This merges #2320 to run the new tests.

PR #2339, opened 2015-12-12
[FLOC-3692] Failing test for protecting the dataset paths.

sarum90:


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+
+@attributes(["_proxy_object", "_callback"], apply_with_init=False)
+class CallbackProxy(object):
+    """
+    A proxy to another object that calls a callback before and after each call
+    to the underlying object. This enables tests where you want to inject code
+    before or after any calls to an underlying API.
+
+    Note that this proxy assumes that the underlying object only has public
+    methods and not public attributes.
```
tomprince: My inclination would be to require an interface here and only proxy the methods of that interface. ``flocker.common.interface_decorator`` would be useful here.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+        object and returns a wrapped function that calls the callback before
+        and after its execution.
+
+        :param name: The name of the method on the proxy object that should be
+            called.
+        """
+        method = getattr(self._proxy_object, name)
+
+        def proxied_function(*args, **kwargs):
+            self._callback('Before %s' % name)
```
tomprince: It would be better to pass structured data here. Probably a constant indicating when the callback is being called and the actual method being called.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+        :param cluster_state_callback: A callable that will be called with a
+            single argument, the cluster state, after every discover_state
+            cycle.
+
+        :return: ``Deferred`` that fires when the necessary changes are done.
+        """
+        # XXX: Note to the reviewer - I wish I could use something closer to
+        # the actaul production code here, but I didn't see how to do that. Do
+        # you? The one exception is that I want there to be no state persisted
+        # from one run of converge to the next.
```
tomprince: This looks very much like the code in https://github.com/ClusterHQ/flocker/pull/2319/files#diff-85cea39da1659d56c9a2c44357b4120dR1191

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+            """
+            An invariant that should always hold. Specifically, after the path
+            to a dataset is known by the cluster then writes to that path
+            should either succeed and be persisted to the underlying block
+            device, or the write should fail.
+
+            Returns False if none of the known paths of the dataset can be
+            written to.
+            """
+            write_success = []
```
tomprince: Does this need to be a list? It seems like you could just use a boolean that starts off false.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+            for path in dataset_info_from_cluster['paths']:
+                device = dataset_info_from_cluster['dev']
+                filename = str(uuid4())
+                try:
+                    path.child(filename).touch()
+                except OSError:
+                    # If we failed to write, that is okay in many scenarios. At
+                    # least the failing write won't lead to the end user
+                    # thinking that they have an actually persisted data. Note
+                    # the failure, but don't immediately fail the test.
```
tomprince: Nothing appears to be done with the failure here?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+                else:
+                    return None
+
+            result.addCallback(repeat_if_necessary)
+            return result
+        return loop_until_converged()
+
+    def test_protect_against_docker(self):
+        """
+        Verify that in a standard lifecycle of a dataset there is never a time
```
tomprince: `Verify that` is redundant (see https://jml.io/pages/test-docstrings.html)

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5249,3 +5257,231 @@ def test_get_device_path_until_detach(self):
...
+            actual_blockdevice_manager.mount(blockdevice, test_mountpoint)
+            result = test_mountpoint.child(filename).exists()
+            actual_blockdevice_manager.unmount(blockdevice)
+            test_mountpoint.remove()
+            return result
+
+        # This is information about the dataset that is intentionally only
+        # populated once the path of the dataset is known by the cluster and
+        # thus could be known by an API client.
+        dataset_info_from_cluster = {
```
tomprince: This is sort-of used like a global, in that the places where it used are far from the places it gets updated. I think the test would be clearer if there wasn't this action at a distance.

flocker/node/agents/test/test_blockdevice.py
tomprince: In particular, I think we shouldn't add to places in the test that have different versions of this logic. I also think that the test would be clearer if it was expressed in terms of the newly introduced explicit states instead.

flocker/node/agents/test/test_blockdevice.py
tomprince: In particular, I think I would be inclined to just discover the state directly when you need this information.

tomprince:
This looks like a sensible test to have. :+1:  However, I do think the logic of the test could be made clearer, and have made a few suggestions on how that might done. Since I think that the code will probably significantly change in detail, if not in outline, please re-submit after considering the suggestions.

flocker/node/agents/test/test_blockdevice.py
sarum90: So I'll probably do this cause that code looks interesting, but I'm not 100% it adds that much, and it comes at some complexity.

@tomprince 
What's the value add for this? Another layer of double checking the consumer of this interface is only using the interface and not cheating by using non-interface methods and variables?

Feel free to respond out of band as I do plan on using the interface_decorator for this.

sarum90: Done.

flocker/node/agents/test/test_blockdevice.py
sarum90: I'll look into de-duping, but I'm unsure what explicit states you are referring to. If you mean explicit states of datasets, I'm sort of trying to avoid having to depend on those. Instead I'm trying to just change configuration and then run the loop until it converges. If you mean explicit states of the machinist loop that normally drives that, I would believe that. Although... I haven't dug deep enough into that code to know of hand.

For now I'll look at the link you provided and see how I can leverage it.

tomprince: > If you mean explicit states of datasets, I'm sort of trying to avoid having to depend on those.

I do mean that, and I don't think it makes sense to try avoid depending on them. I think it much easier to reason about them that it is to reason about cluster configurations (and in particular, easier to see that you have full coverage of possible states and state transitions.

flocker/node/agents/test/test_blockdevice.py
sarum90: :+1: Also I realized that stack traces work better than random strings for debugging so I've removed for now. If I'm tempted to add back I'll be sure to use structured data.

flocker/node/agents/test/test_blockdevice.py
sarum90: Ok switched to using that infrastructure.  I sort of like only using cluster state rather than DesiredDatasets for driving the change since that is complete black box testing, but this seems 1:1 with that anyway.

flocker/node/agents/test/test_blockdevice.py
sarum90: So it's a list in case we ever have 2 different paths for a dataset at 2 different times. Current implementation does not do that, but if you imagine an implementation where you mount datasets at random paths every time then for a single dataset after a manifest->non-manifest->manifest transition you could imagine there would be two paths that an external agent might try to write to for the dataset.

flocker/node/agents/test/test_blockdevice.py
sarum90: An OSError just means that we failed to write. That's an okay outcome (and in fact desirable if it happens after a umount). We are only trying to prevent the case where a write is successful, but it is not persisted to the backend.

flocker/node/agents/test/test_blockdevice.py
sarum90: Thanks! I'm always bad about that... even having read that before...

flocker/node/agents/test/test_blockdevice.py
sarum90: Yea... the thing is I want to try to write to paths that are no longer being reported by flocker as valid paths for the dataset (Once flocker reports a path as a location that a dataset is located at, I want to keep trying to write there even if flocker is no longer reporting that as a valid path for the dataset) AFAIK this is similar to protecting against docker trying to use a path after flocker has removed it from a node for whatever reason (stale caches / configurations / etc).

I refactored this into an external agent class which hopefully more clearly shows that this is information that is known by some external agent (i.e. potential views of the world that docker may have, that are intentionally arbitrarily stale).

sarum90:
This PR is gross after merging things in ... I think I'm going to make a new branch and rebase off of master so it's more clear. I'll post back here with new PR number if I find success.

sarum90:
Deleting this in favor of #2349, which has a cleaner history without force rebasing this PR.

flocker/node/agents/test/test_blockdevice.py
tomprince: You say "Not the failure, but don't immediately fail the test". You don't appear to actually be noting the failure.

sarum90: Ah thanks, yea Note... I'll go fix that in the other branch now...

sarum90: Oh I just meant, "Note" as in programatically note. I'll try to make this more clear in the other branch...

PR #2343, opened 2015-12-14
[FLOC-3648] [FLOC-3649] Discover and detach from dead nodes

itamarst:
If we discover a dataset is attached to a dead node, try to detach it in the destination node.

I've not actually seen a passing acceptance test, so no idea if it actually works, but maybe I actually found last bug? :see_no_evil:


flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,3 +109,69 @@ def not_exists():
...
+    @require_moving_backend
+    @run_test_with(async_runner(timeout=timedelta(minutes=6)))
+    @require_cluster(2)
+    def test_dataset_move_from_dead_node(self, cluster):
+        """
+        A dataset can be moved from one node to another.
+
+        All attributes, including the maximum size, are preserved.
+        """
+        api = get_backend_api(self, cluster.cluster_uuid)
```
sarum90: The current implementation of get_backend_api only constructs aws backends (sorry, I wasn't sure if this api would be in general useful in acceptance tests so I didn't put the work in upfront). If we are claiming OpenStack in the release notes we should probably test on rackspace as well.

Shouldn't be too much work - something like making _DEFAULT_BACKENDS public from flocker/node/script.py:459 and using it in get_backend_api to construct backends of any type.

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,3 +109,69 @@ def not_exists():
...
+                    cluster.control_node.public_address)[0]
+        other_node = list(other_node for other_node in cluster.nodes
+                          if other_node != node)[0]
+        waiting_for_create = create_dataset(self, cluster, node=node)
+
+        def startup_node(node_id):
+            api.start_node(node_id)
+            # Give node some minimal amount of time to boot so next test
+            # is happier:
+            sleep(20)
```
sarum90: Heh... sleeps are always fun. I guess 20 seconds on a 25+ minute test suite isn't that much, but how sure are we that 20 seconds will always win the race?

I think we should probably try to see this pass like 4-10 times on Jenkins before merging, just to double check we are not actively making work for the green-ci team.

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,3 +109,69 @@ def not_exists():
...
+            # Wait for shutdown to be far enough long that node is down:
+            d.addCallback(
+                lambda _:
+                loop_until(reactor, lambda:
+                           set(api.list_live_nodes()) != live_node_ids))
+            # Schedule node start up:
+            d.addCallback(
+                lambda _: self.addCleanup(
+                    startup_node,
+                    (live_node_ids - set(api.list_live_nodes())).pop()))
```
sarum90: :grey_question: It's sorta goofy that you have to do this to find the cloud node_id. Just reading the code it'd be cool if this could just be node.cloud_id or something, but it looks like we never propagate compute_instance_id() outside of the dataset agent so that's probably impossible.

No action required, but if you have some clever idea of how to make this code more obvious, I was just noting that it seemed slightly convoluted.

flocker/node/agents/blockdevice.py
```diff
@@ -1299,6 +1318,7 @@ class RawState(PClass):
     The raw state of a node.
 
     :param unicode compute_instance_id: The identifier for this node.
+    :param unicode live_instances: The identifiers of live nodes.
```
sarum90: nit:
:type live_instances: ``pset`` of unicode node identifiers, rather than just `unicode` type.

flocker/node/agents/blockdevice.py
```diff
@@ -1598,8 +1621,15 @@ def is_existing_block_device(dataset_id, path):
                     # in the filesystem yet.
                     pass
 
+        if ICloudAPI.providedBy(api):
+            live_instances = api.list_live_nodes()
+        else:
+            # Assume everything is alive:
+            live_instances = [volume.attached_to for volume in volumes
+                              if volume.attached_to is not None]
```
sarum90: This seems brittle to me.  If someone uses live_instances for anything other than checking volume.attached_to they might be in for a surprise.

In fact if this node doesn't have any volumes attached then `self.compute_instance_id() in live_instances` is False o.O

Perhaps instead have `live_instances == None` indicate that we do not know which instances are live, and set it to None here? Then add a method to RawState:

```
def is_instance_live(self, instance_id):
   if self.live_instances is None:
     return True
   return instance_id in self.live_instances
```
and use that below.

sarum90: Or maybe even:
```
def is_known_dead_instance(self, instance_id):
  if self.live_instances is None:
     return False
  return instance_id not in self.live_instances
```

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,3 +109,69 @@ def not_exists():
             return deleted
         created.addCallback(delete_dataset)
         return created
+
+    @require_moving_backend
+    @run_test_with(async_runner(timeout=timedelta(minutes=6)))
+    @require_cluster(2)
+    def test_dataset_move_from_dead_node(self, cluster):
+        """
+        A dataset can be moved from one node to another.
```
sarum90: s/one node to another/a downed node to a live node/g

flocker/node/agents/blockdevice.py
```diff
@@ -1408,7 +1430,8 @@ def from_state_and_config(discovered_dataset, desired_dataset):
         Discovered.ATTACHED_ELSEWHERE: ActionNeeded,
         Discovered.ATTACHED_NO_FILESYSTEM: CreateFilesystem,
         Discovered.NON_MANIFEST: AttachVolume,
-        DatasetStates.ATTACHED: MountBlockDevice,
+        Discovered.ATTACHED: MountBlockDevice,
+        Discovered.ATTACHED_TO_DEAD_NODE: DetachVolume,
```
sarum90: No action required: This looks too easy, can we go back to a bunch of convoluted code across multiple functions for this sort of change? :stuck_out_tongue: Thanks @tomprince for the cleanup work!

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -428,37 +428,44 @@ def test_non_manifest_dataset(self):
             expected_changes,
         )
 
-    def attached_dataset_test(self, state):
+    def attached_dataset_test(self, state, pass_device_path=True):
         """
         When there is a a dataset in the given attached state,
```
sarum90: :grey_question: nit: s/a a/a/

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -428,37 +428,44 @@ def test_non_manifest_dataset(self):
             expected_changes,
         )
 
-    def attached_dataset_test(self, state):
+    def attached_dataset_test(self, state, pass_device_path=True):
```
sarum90: I think with the new argument, test_non_manifest_dataset can use this helper as well and set it to False.

Probably should rename it to non_manifest_dataset_test and move it above the preceding test in that case.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1612,10 +1689,10 @@ def assert_calculated_changes(
 
     :param discovered_datasets: Collection of ``DiscoveredDataset`` to
         expose as local state.
+    :param api: ``IBlockDeviceAPI`` provider to use; by default a
```
sarum90: Strongly request that this not be made a parameter.

I think it's vital that calculate_changes is a pure function of it's arguments. We arguably could have done a different design to force that, but in lieu of a complete re-design this test infrastructure is the best thing we have to ensure we aren't making API calls during calculate_changes.

I don't think you actually use the parameter so it shouldn't be an issue?

sarum90:
Looks good - really nice and clean now that we've refactored blockdevice.py. I've got some comments throughout, but no major refactors requested, so I'll be accepting the PR. Address comments, get some green runs of the test on Jenkins and merge.

I am a little concerned about the acceptance test with the 20 seconds wait. I think I'd prefer to see a couple green runs of the test before we merge, just to double check we aren't just merging in a flaky test and generating more work for the green-ci team.

flocker/acceptance/endtoend/test_dataset.py
itamarst: I could loop on list_live_nodes instead, I guess? That might be more reliable.

itamarst: Ah, wait for SSH to come up, which is what I did for reboot test, and that worked well.

flocker/node/agents/test/test_blockdevice.py
itamarst: I think I assumed that assertion did discovery too, but it doesn't. calculate_changes should be functional, yeah.

itamarst:
I've asked Swisscom to try this out, won't merge until we get feedback.

itamarst:
Since the last review I identified and fixed a few bugs:
1. `ICloudAPI` was never detected in practice since the check was on the `ProcessLifetimeCache` that wraps the underlying `IBlockDeviceAPI`, and doesn't proxy `ICloudAPI`.
2. Cinder backend didn't allow detach on remote nodes.
3. Cinder backend listed all nodes as being live.

itamarst:
Swisscom has run a successful test of this branch.

itamarst:
There is no changelog entry, BTW, since we don't have acceptance tests so I don't want to publicly promise this will work.

flocker/node/agents/cinder.py
```diff
@@ -539,15 +539,16 @@ def attach_volume(self, blockdevice_id, attach_to):
         return attached_volume
 
     def detach_volume(self, blockdevice_id):
-        our_id = self.compute_instance_id()
         try:
             cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
         except CinderNotFound:
             raise UnknownVolume(blockdevice_id)
+        server_id = _blockdevicevolume_from_cinder_volume(
```
sarum90: :grey_question: I'm surprised this change didn't break any cinder functional tests.

Do you think there would be any value at testing this at the functional level? Perhaps it's not worth setting up the two instances to run the functional test... but it'd be nice to have some coverage here.

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,3 +108,72 @@ def not_exists():
             return deleted
         created.addCallback(delete_dataset)
         return created
+
+    @skipIf(True,
+            "Shutting down a node invalidates a public IP, which breaks all "
```
sarum90: :grey_question: point of curiosity, have you gotten this to run/pass? Just with static IPs or something?

Just curious because this should have caught bugs such as not using _underlying_blockdevice_api, so I think it does have some value.

sarum90:
New changes to actual code look great, cool to see us be able to solve this issue for people running OpenStack

Some concern regarding:

> There is no changelog entry, BTW, since we don't have acceptance tests so I don't want to publicly promise this will work.

1) Given there is no acceptance test, do we expect people who depend on this functionality to verify it in some sort of staging environment every time before they roll out a new version of flocker?

2) Is it worth the code complexity in `blockdevice.py` for functionality that is only needed in certain environments? or is there some way we could hide this complexity off in `cinder.py` or something?

I think this is probably fine as is, and am fine if it is merged. Although we should be sure to set long term expectations on the testing burden re (1). Without the acceptance test, it seems very possible to me that this functionality might be lost in the future in a refactor.

itamarst:
We'll discuss 1st point yeah. Re 2nd point, this functionality is actually a very general one; in the future you can imagine replacing `ICloudAPI` with a `INodeLiveness` API with e.g. a Kubernetes provider that just relies on k8s' notion of being alive.

sarum90:
I'm suspect that this might be increasing our number of API calls by a fairly significant amount. I'm getting RequestLimitExceeded errors when I run two sets of the acceptance at a time at 4pm Pacific time, which hadn't happened in the past.

I might be wrong though, just a lingering suspicion.

PR #2344, opened 2015-12-14
[FLOC-3705] Add default volume size, if available, to benchmark cluster

jongiddy:
In order to introduce a "create dataset" operation, it is useful to have a size for the dataset. Since valid sizes differ by clouds platform, it is great if the user gives us this information.

If we use the acceptance test environment variables (currently our preferred solution), an appropriate value is available in one of the environment variables.

If we use the UFT Flocker Installer, there is no value provided. However, the API allows `None` to be provided as this parameter.  We can live with this, as benchmarking is an internal process, and if it causes problems, we may find them before someone else.


zubron:
Looks good, thanks. The benchmark tests are green. Merge when you're satisfied with the others :)

PR #2346, opened 2015-12-14
[FLOC-3711] Fix count of iterations in failure message.

sarum90:
Fixes FLOC-3711 by passing number of iterations through in the Exception and using that count in the error message.

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -26,6 +26,8 @@ from pyrsistent import (
     PClass, field, discard, pmap, pvector,
 )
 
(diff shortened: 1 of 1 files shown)


zubron:
Great, thanks @sarum90 :)

sarum90:
Thanks for the review!

A note on tests. run_sphinx is failing the same way that it is failing on master, so I don't think that should block submission here. The *CentOS_7_for_flocker.acceptance.obsolete tests that also failed I'm not too concerned about as they should be unaffected by this code change. I'll dive into them and be sure they are covered by JIRA issues, but I'm just going to merge this as is rather than wait on them to be green before merging.

PR #2347, opened 2015-12-15
[FLOC-3713] Enable test_simple_transitions

sarum90:
This test was being skipped because when it failed it would not clean itself up nicely, and would then cause failures of other tests afterwards.

Changing the unmount cleanup code to unmount by mount point rather than by device has alleviated that issue. There were circumstances where loop devices got in a state where you could no longer unmount them by the loop device name, but could still unmount by the mount point.

Once that was fixed the test was still failing regularly because the tagged union invariants were being broken by either the initial desired dataset state, or by the next desired dataset with a different tag. Rather than just changing the state of the desired dataset, I added a new capability to merge in all the attributes from one tagged_union into another tagged_union. This brings in all attributes that are common to both tags.  Using this new utility in the test enables a way to create two different DesiredDataset states for the same dataset regardless of which states they are without a lot of special case attribute manipulation.

Note: This test case takes about a minute to execute, but I think it is pretty powerful, and ensuring it does not regress is probably more important than the test time lost.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1210,7 +1210,7 @@ def run_convergence_step(self, desired_datasets):
         note("Running changes: {changes}".format(changes=changes))
         self.successResultOf(run_state_change(changes, self.deployer))
 
-    def run_to_convergence(self, desired_datasets, max_iterations=4):
+    def run_to_convergence(self, desired_datasets, max_iterations=6):
```
itamarst: Why 6? If that's what it takes exactly, it might take more in future... and then we'll get mysterious failures. On presumption that iterations are cheap once convergence is achieved, I suggest picking a number that is definitely big enough, e.g. 20, and explaining why.

flocker/common/test/test_algebraic.py
```diff
@@ -185,3 +187,46 @@ def test_invalid_states(self, state, extra_value):
...
+
+
+class MergeTaggedUnionsTests(SynchronousTestCase):
+    """
+    Tests for merge_tagged_unions.
+    """
+
+    def test_simple_merge(self):
+        """
+        Merging takes all attributes form the src.
```
itamarst: "from the source that are not already present in the target" perhaps?

flocker/common/test/test_algebraic.py
```diff
@@ -185,3 +187,46 @@ def test_invalid_states(self, state, extra_value):
...
+                               one=False)
+
+        src = AlgebraicType(state=States.WITH_TWO_ATTRIBUTES,
+                            extra=True,
+                            one=True,
+                            two=True)
+
+        self.assertEqual(merge_tagged_unions(AlgebraicType, src, target),
+                         AlgebraicType(state=States.WITH_ATTRIBUTE,
+                                       extra=True,
```
itamarst: So... sometimes it overwrites and sometimes it doesn't? Why? When?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1232,21 +1232,31 @@ def run_to_convergence(self, desired_datasets, max_iterations=4):
 
     @given(
```
itamarst: In an ideal world the fact the next dataset is related to the initial one would be visible to Hypothesis, as it was before... This change reduces the search space as a result, since we get more duplicate iterations; limited number of cases we have in each run are wasted on generating effectively duplicate random variants of next dataset.

Possibly https://hypothesis.readthedocs.org/en/latest/data.html#chaining-strategies-together might help? Not sure.

flocker/common/test/test_algebraic.py
itamarst: "from the source, keeping any additional ones in the target"?

flocker/node/agents/test/test_blockdevice.py
itamarst: You could probably do something with either `flatmap` or more plausibly `composite`, which is documented later in that page: sample an initial dataset. then sample a next state, and for each additional field in the new type sample values from relevant strategy.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -4119,13 +4125,14 @@ class FakeProfiledLoopbackBlockDeviceIProfiledBlockDeviceTests(
     """
 
 
-def umount(device_file):
+def umount(unmount_target):
     """
     Unmount a filesystem.
 
-    :param FilePath device_file: The device file that is mounted.
+    :param FilePath unmount_target: The device file that is mounted or
```
itamarst: Is this ever used with the device now? Or is it always the mountpoint?

itamarst:
I'm really glad you're getting this going live, it's a great improvement over current tests.

I'd prefer a Hypothesis-based implementation, as noted above, which will improve coverage. That being said, this is way better than before, so your choice: if you feel like doing it, which is probably educational in terms of Hypothesis usage then please do that and resubmit. If you want to proceed with this implementation strategy, a few minor comments and then merge.

Thanks!

flocker/common/test/test_algebraic.py
sarum90: Done.

flocker/node/agents/test/test_blockdevice.py
sarum90: Agreed, done and done.

flocker/node/agents/test/test_blockdevice.py
sarum90: Decided to do this right. I'll probably send back out for review if I can get a quick glance over it.

flocker/node/agents/test/test_blockdevice.py
sarum90: Looks like it's always the mountpoint now, but in it was both before this PR.

I'm sort of inclined to keep this comment as is with this PR, because we have used this code for either in the past, and we don't need to change the body of this function if we change the way we use it. This is also the interface that is exposed by the unmount command line utility and system call so it should be a fairly recognizable API.

sarum90:
Resubmitting with new strategy.

flocker/node/agents/test/test_blockdevice.py
sarum90: Given that we're doing a second round of review though, let me know if you strongly feel like we should only document the mountpoint API and I can change it.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -148,19 +148,35 @@
     blockdevice_id=_create_blockdevice_id_for_test(dataset.dataset_id),
 ))
 
+DESIRED_DATASET_ATTRIBUTE_STRATEGIES = {
+    'dataset_id': uuids(),
+    'maximum_size': integers(min_value=0).map(
+        lambda n:
+        LOOPBACK_MINIMUM_ALLOCATABLE_SIZE
+        + n*LOOPBACK_ALLOCATION_UNIT),
+    'metadata': dictionaries(keys=text(), values=text()),
```
itamarst: I recommend making these shorter, e.g. https://github.com/ClusterHQ/eliot/blob/master/eliot/tests/strategies.py#L43

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -148,19 +148,35 @@
...
+# This strategy creates two `DESIRED_DATASET_STRATEGY`s with as much in common
+# as possible. This is supposed to approximate two potentially different
+# `DesiredDataset` states for the same dataset.
+TWO_DESIRED_DATASET_STRATEGY = DESIRED_DATASET_STRATEGY.flatmap(
+    lambda x: tuples(just(x), tagged_union_strategy(
+        DesiredDataset,
+        dict(DESIRED_DATASET_ATTRIBUTE_STRATEGIES, **{
+            attribute: just(getattr(x, attribute))
+            for attribute in DESIRED_DATASET_ATTRIBUTE_STRATEGIES
+            if hasattr(x, attribute)
```
itamarst: `hasattr` is kinda broken... maybe we should have a `safe_hasattr` based on `getattr`.

itamarst:
1. Address comments above as you see fit.
2. There's some relevant red in Jenkins, fix that too.

Merge when green.

flocker/node/agents/test/test_blockdevice.py
sarum90: Added inline getattr implementation.

flocker/node/agents/test/test_blockdevice.py
sarum90: Unfortunately it's still 60 seconds of runtime, but I'm sure it'll be nicer in logs.

sarum90:
Regarding tests: This is only changing flocker/node/agents/test/test_blockdevice.py, and I've got green results on Jenkins for run_as_root flocker.node.agents on both CentOS 7 and Ubuntu on the latest commit. I've checked the logs and verified successful runs of the particular test in question as well.

Given that I'm going to merge as is. I'll be sure to follow up with any Jenkins red to verify that there are JIRA issues for any flakes that I see.

jml:
This caused regressions, alas: https://clusterhq.atlassian.net/browse/FLOC-3738

PR #2348, opened 2015-12-15
[FLOC-3695] Fix broken run_sphinx job by fixing broken docs links

jml:
Docker probably changed their SSL configuration, which broke our links that were redirecting from http://docs.docker.com to https://docs.docker.com. Updating the links to correctly point to https://docs.docker.com removes these errors from the build. Unfortunately, it does so because we do not check https links. Fixing that problem is out of scope for this patch, since the priority is to unbreak master.

In addition, since master started breaking due to https links, we added extra failures by deleting some buildbot jobs. I've taken the liberty of updating the documentation that used those examples.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -271,14 +271,14 @@ common_cli:
       --no-passthrough --output-to=results.xml
 
   run_sphinx: &run_sphinx |
-    parse_logs python setup.py --version
+    parse_logs \${venv}/bin/python setup.py --version
     cd docs
     # check spelling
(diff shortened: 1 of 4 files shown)


exarkun:
I'm satisfied with these changes.

davidoliverSP2:
links are working - thanks @jml

exarkun:
Not much luck getting a lot of green on jenkins.  However, the failures all seem to be jenkins problems.  Manually verified this change fixes the problems on my system (where I was able to reproduce the problems).  And pretty sure there's no mechanism for these changes to be causing any of the other red.  So, merging.

PR #2349, opened 2015-12-16
[FLOC-3692] Failing test for protecting the dataset paths.

sarum90:
Cleaned up version of #2339

Note that this branches off of make-test-simple-transitions-work-FLOC-3713 because it uses the test infrastructure that was fixed in that branch.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1195,29 +1203,156 @@ class DidNotConverge(Exception):
...
+    def invariant(self, case, current_cluster_state):
+        """
+        An invariant that should always hold. Specifically, after the path
+        to a dataset is known by the cluster then writes to that path
+        should either succeed and be persisted to the underlying block
+        device, or the write should fail.
+
+        :param case: A ``TestCase`` to use for making assertions.
+        :param current_cluster_state: A callable that returns the current state
+            of the cluster.
```
mpgomez: return?

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1195,29 +1203,156 @@ class DidNotConverge(Exception):
...
+        """
+        thing = False
```
mpgomez: It is a big difficult to understand what "thing" is

sarum90: Thanks, I didn't mean to have that variable anymore.

flocker/node/agents/test/test_blockdevice.py
sarum90: Supposed to return nothing :) -- the return was left over from some debugging code.

sarum90:
I sort of want this to be merged to non-master. Closing and re-openning.

PR #2350, opened 2015-12-16
[FLOC-3293] Benchmark dataset creation

jongiddy:
Add a benchmarking operation to create a dataset, and then wait for the cluster to converge.  This operation:
1. selects one of the nodes in the cluster, 
2. creates a dataset on it (sets the configuration), 
3. waits for the dataset to be created (inspects the actual state), and
4. removes the dataset.

Steps 2 and 3 are in the `run` method, and hence are the actions measured for benchmarking.

Also added `Node` to the list of objects exported by the `flocker.apiclient` module, to avoid importing it from the evidently private `_client` submodule.


benchmark/operations/create_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+class EmptyClusterError(Exception):
+    """
+    Exception indicating that the cluster contains no nodes.
+    """
+
+
+@implementer(IProbe)
+class CreateDatasetConvergenceProbe(PClass):
+    """
+    Probe to create a dataset and wait for cluster to converge.
```
itamarst: It doesn't really wait for the cluster to converge in general, so maybe "create a dataset in the configuration and wait for the dataset to be created and mounted"?

benchmark/operations/create_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+        :param control_service: Benchmark control service.
+        :param UUID dataset_id: UUID for created dataset.
+        :param int volume_size: Size of created volume, in bytes.
+        :return: Deferred firing with a new probe.
+        """
+        d = control_service.list_nodes()
+
+        # Select an arbitrary node to be the primary for the dataset.
+        def pick_primary(nodes):
+            for node in nodes:
```
itamarst: `random.choice()` would be better, it's possible the order of nodes is determistic.

benchmark/operations/create_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+            )
+        d.addCallback(create_probe)
+
+        return d
+
+    def _converged(self, expected):
+        """
+        Check whether a dataset has been created.
+
+        :param expected: A dataset dictionary to match against the results of
```
itamarst: Not a dictionary, it's a ... `flocker.apiclient.Dataset` or something.

benchmark/operations/create_dataset.py
```diff
@@ -0,0 +1,131 @@
...
+    volume_size = field(mandatory=True)
+
+    @classmethod
+    def setup(cls, reactor, control_service, dataset_id, volume_size):
+        """
+        Create a probe.
+
+        :param reactor: Twisted Reactor.
+        :param control_service: Benchmark control service.
+        :param UUID dataset_id: UUID for created dataset.
```
itamarst: I would suggest relying on Flocker to generate the `dataset_id`, rather than passing it in; in the long term I'd like to deprecate the ability to do the latter.

benchmark/operations/test/test_create_dataset.py
```diff
@@ -0,0 +1,96 @@
...
+class CreateDatasetTests(SynchronousTestCase):
+    """
+    CreateDataset operation tests.
+    """
+
+    def test_implements_IOperation(self):
+        """
+        CreateDatasetConvergence provides the IOperation interface.
+        """
+        verifyClass(IOperation, CreateDatasetConvergence)
```
itamarst: Not necessarily relevant in this case, but it's more useful to do `verifyObject` on an instance, cause sometimes the class can't do everything and so `verifyObject` on an instance does more validation. I believe e.g. `Attribute` is only verified by `verifyObject`.

benchmark/operations/create_dataset.py
itamarst: Which I guess mean `run()` needs to store the dataset ID and `clean()` needs to be OK with it being missing. And could admittedly lead to leaked datasets... so maybe not worth it.

docs/gettinginvolved/benchmarking.rst
```diff
@@ -115,6 +115,14 @@ Scenario Types
 Operation Types
 ~~~~~~~~~~~~~~~
 
+.. option:: create-dataset
+
+   Create a dataset and wait for cluster to converge.
```
itamarst: See comment above about replacing "wait for cluster to converge", probably applies here too.

itamarst:
Looking good, address and merge.

PR #2351, opened 2015-12-16
[FLOC-3435] upload and consume clusterhq vagrant boxes from S3

Azulinho:
* unsets 'set -x' so that we won't leak secrets to the logs
* removes a chunk of dead code (parse_logs)
* adds a retry function, to retry the most common failures with the vagrant build jobs
* adds a few helper functions for the vagrant operations
* fixes some issues where the vagrant boxes would run out of mem
* updates the vagrant box to use centos 7.1 and not spend too much time updating the OS
* OSX client tests will now pull the latest OSX box from S3
* OSX and tutorial will now upload the nightly built box to S3

This PR will make sure our vagrant builds will upload the resulting arfifacts to a bucket in S3.
This will allows us to grown our mesos cluster and remove a dependency where the OSX client instalation tests would fail if the OSX vagrant basebox hadn't been built yet. This so far worked as our mesos cluster is a single box, as we move to multiple boxes the client tests would fail as the OSX basebox wasn't available locally on the host.
The client tests now will pull the latest box from S3 through a 'vagrant box update'

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -88,6 +88,9 @@ common_cli:
   # TODO: https://clusterhq.atlassian.net/browse/FLOC-2986
   add_shell_functions: &add_shell_functions |
 
+    # set a default AWS endpoint for S3 operations
+    export S3_ENDPOINT=us-west-2
+
     # The long directory names where we build our code cause pip to fail.
     # https://gitlab.com/gitlab-org/gitlab-ci-multi-runner/issues/20
     # http://stackoverflow.com/questions/10813538/shebang-line-limit-in-bash-and-linux-kernel
@@ -199,7 +202,7 @@ common_cli:
     function s3_upload() {
       echo "uploading \${2} to S3 bucket ${1} ..."
       retry_n_times_with_timeout 3 900 \
-        aws --region us-west-2 s3 cp "\${2}" s3://\${1}/
+        aws --region \${S3_ENDPOINT} s3 cp "\${2}" s3://\${1}/
     }
 
     # creates a metadata file to be used with vagrant up; vagrant box update
@@ -688,7 +691,7 @@ common_cli:
        config.vm.synced_folder '.', '/vagrant', disabled: true
        config.vm.provider 'virtualbox' do |vb|
           vb.memory = '1024'
-          # VirtualBox 5 now supports para-virtualization, lets use it.
+          # VirtualBox 5 now supports para-virtualization, let's use it.
           vb.customize("pre-boot", ["modifyvm", :id, "--paravirtprovider", "KVM"])
        end
     end
@@ -814,21 +817,21 @@ common_cli:
   get_s3_creds_for_bucket_vagrant_jenkins_boxes: &get_s3_creds_for_bucket_vagrant_jenkins_boxes |
     # collects the S3 credentials that allows us to upload files to the S3
     # vagrant-jenkins_boxes for our vagrant boxes
-    # the file with the credentials are deployed by the jenkins master to the
+    # the file with the credentials is deployed by the jenkins master to the
     # /tmp of the slave when the slave is instantiated.
     . /tmp/s3_vagrant_jenkins_boxes_clusterhq.com
 
   get_s3_creds_for_clusterhq_dev_archive: &get_s3_creds_for_clusterhq_dev_archive |
     # collects the S3 credentials that allows us to upload files to the S3
     # clusterhq-dev-archive for our vagrant boxes
-    # the file with the credentials are deployed by the jenkins master to the
+    # the file with the credentials is deployed by the jenkins master to the
     # /tmp of the slave when the slave is instantiated.
     . /tmp/s3_dev_archive_clusterhq.com
 
   upload_new_docs_html_to_s3_staging_docs: &upload_new_docs_html_to_s3_staging_docs |
     # uploads the new generated html to a folder on the S3 staging-docs bucket
     cd _build/html
-    aws --region us-west-2 s3 sync . s3://clusterhq-staging-docs/\$TRIGGERED_BRANCH
+    aws --region \${S3_ENDPOINT} s3 sync . s3://clusterhq-staging-docs/\$TRIGGERED_BRANCH
     echo "This branch is available at :"
     echo "http://clusterhq-staging-docs.s3.amazonaws.com/\$TRIGGERED_BRANCH/index.html"
 
@@ -852,22 +855,15 @@ common_cli:
     VAGRANT_BOX_VERSION="\$( echo \${FLOCKER_VERSION} | cut -f 1,2,3 -d "." )"
 
     cd \$BOX_PATH
+
     # generates the metadata.json, it used by vagrant box update operations.
-    cat <<EOF_METADATA > metadata.json
-    {
-      "name": "clusterhq/flocker-tutorial",
-      "description": "This box contains the flocker-tutorial VM.",
-      "versions": [{
-        "version": "\${VAGRANT_BOX_VERSION}",
-        "providers": [{
-          "name": "virtualbox",
-          "url": "\${HTTP_BOX_URL}",
-          "checksum_type": "sha1",
-          "checksum": "\${SHA1}"
-        }]
-      }]
-    }
-    EOF_METADATA
+    create_vagrant_metadata_file \
+      metadata.json \
+      clusterhq/flocker-tutorial \
+      'This box contains the flocker-tutorial VM.' \
+      \${VAGRANT_BOX_VERSION} \
+      \${HTTP_BOX_URL} \
+      \${SHA1}
 
     cp metadata.json \${BOX_NAME}.json
 
@@ -925,8 +921,7 @@ common_cli:
       s3_upload \${S3_BUCKET} "metadata.json"
     ) || abort_build
 
-    echo "This vagrant box is available is available at :"
-    echo "\${S3_BOX_JSON_LATEST}"
+    echo "This vagrant box is available at : \${S3_BOX_JSON_LATEST}"
 
 
   run_lint: &run_lint |
@@ -1468,8 +1463,8 @@ job_type:
                    *install_pkgs_on_vanilla_osx_yosemite_box,
                    *end_build_sh_EOF,
 
-                  # starts and provisions an OSX vagrant machine
-                  # using the build.sh script created above
+                   # starts and provisions an OSX vagrant machine
+                   # using the build.sh script created above
                    *start_vanilla_osx_yosemite_vagrant_box,
                    *run_build_sh_script_on_vagrant_box,
 
--- a/build.yaml
+++ b/build.yaml
@@ -696,7 +696,7 @@ common_cli:
 
 
   source_git_commit: &source_git_commit |
-    # we use a script 'git-commit.sh' to pass any variables from the host
+    # we use a script 'git-commit.sh' to pass any the git commit sha1
     # to our vagrant machine. This file is copied and sourced as part of our
     # build.sh run inside the virtual machine.
     chmod 755 git-commit.sh
--- a/build.yaml
+++ b/build.yaml
@@ -359,14 +359,14 @@ common_cli:
       --no-passthrough --output-to=results.xml
 
   run_sphinx: &run_sphinx |
-    parse_logs \${venv}/bin/python setup.py --version
+    \${venv}/bin/python setup.py --version
     cd docs
     # check spelling
-    parse_logs \${venv}/bin/sphinx-build -d _build/doctree -b spelling . _build/spelling
+    \${venv}/bin/sphinx-build -d _build/doctree -b spelling . _build/spelling
     # check links
-    parse_logs \${venv}/bin/sphinx-build -d _build/doctree -b linkcheck . _build/linkcheck
+    \${venv}/bin/sphinx-build -d _build/doctree -b linkcheck . _build/linkcheck
     # build html pages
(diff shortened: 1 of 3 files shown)


build.yaml
```diff
@@ -169,6 +162,105 @@ common_cli:
...
+    # Usage:
+    # retry_n_times_with_timeout <max retries> <timeout in seconds> command
+    #
+    function retry_n_times_with_timeout {
+      set +e
+      max_tries=\${1}
+      seconds=\${2}
+      shift 2
+
+      let delay=15
```
jml: Please add a comment saying what unit this is.

build.yaml
```diff
@@ -586,22 +681,26 @@ common_cli:
...
-    cat <<'EOF' >Vagrantfile
+    cat <<EOF_VAGRANTFILE >Vagrantfile
     Vagrant.configure(2) do |config|
-       config.vm.box = 'jenkins-osx-yosemite'
+       config.vm.box = 'clusterhq/jenkins-osx-yosemite'
+       config.vm.box_url = "s3://\${ENV['S3_BUCKET']}/metadata.json"
        config.vm.synced_folder '.', '/vagrant', disabled: true
        config.vm.provider 'virtualbox' do |vb|
           vb.memory = '1024'
+          # VirtualBox 5 now supports para-virtualization, lets use it.
```
jml: Nit: "let's"

build.yaml
```diff
@@ -612,35 +711,37 @@ common_cli:
...
     vagrant ssh-config > ssh-config
     rsync -ave 'ssh -F ssh-config' . default:.
-    vagrant scp /tmp/pip.sh /tmp/pip.sh
-    vagrant scp build.sh build.sh
-    vagrant ssh -c 'chmod 755 /tmp/pip.sh'
-    vagrant ssh -c 'chmod 755 build.sh'
 
-  run_build_sh_script_on_vagrant_box: &run_build_sh_script_on_vagrant_box |
+    # this sets some variables used by the tests, since the tests run inside
+    # virtualbox we store them to a file and copy them to the vagrant box.
```
jml: Thanks. Comments like these really help follow what's going on.

build.yaml
```diff
@@ -612,35 +711,37 @@ common_cli:
...
+    done
+
     # make sure we don't abort on the first error
     set +e
     # run the build.sh script inside our vagrant box
-    vagrant ssh -c 'bash build.sh'
-    JOB_EXIT_STATUS="\$( updateExitStatus \$? )"
+    vagrant ssh -c 'bash build.sh' || abort_build
+
+    # this removes the secrets files we used during provisioning
```
jml: This step won't happen if the build gets aborted, if I understand correctly. Is this problem? If so, please fix.

build.yaml
```diff
@@ -694,13 +811,124 @@ common_cli:
     # /tmp of the slave when the slave is instantiated.
     . /tmp/s3_staging_docs_clusterhq_com.sh
 
+  get_s3_creds_for_bucket_vagrant_jenkins_boxes: &get_s3_creds_for_bucket_vagrant_jenkins_boxes |
+    # collects the S3 credentials that allows us to upload files to the S3
+    # vagrant-jenkins_boxes for our vagrant boxes
+    # the file with the credentials are deployed by the jenkins master to the
```
jml: Nit: Either "the file with the credentials is" or "the files with the credentials are". Also in the next clause.

build.yaml
```diff
@@ -694,13 +811,124 @@ common_cli:
...
+    # /tmp of the slave when the slave is instantiated.
```
jml: Does this happen in this file? Either way can you please add a "see also" pointing to where this happens (esp. if it's in another file or another git repo).

build.yaml
```diff
@@ -169,6 +162,105 @@ common_cli:
...
+      done
+    }
+
+    # uploads a file to S3, retrying on failure.
+    # usage:
+    # s3_upload <bucket> <filename>
+    function s3_upload() {
+      echo "uploading \${2} to S3 bucket ${1} ..."
+      retry_n_times_with_timeout 3 900 \
+        aws --region us-west-2 s3 cp "\${2}" s3://\${1}/
```
jml: Why are we using this specific region? 

Could we please make this a constant that gets referred to, with some documentation on region choice (even if it's just "arbitrary region to use for s3 uploads")?

build.yaml
```diff
@@ -565,10 +651,19 @@ common_cli:
...
+          --box tutorial \
+          --branch \${TRIGGERED_BRANCH} \
+          --build-server http://\${eth0_ip}
+    )|| abort_build
+
+    (
+      retry_n_times_with_timeout 2 1200 \
+        vagrant box add --force "clusterhq/flocker-tutorial" \
+          vagrant/tutorial/flocker-tutorial-\${FLOCKER_VERSION}.box
+    ) || abort_build
```
jml: These timeouts are fairly large. How much lee-way is there if these take the maximum amount of time vs when the other timeouts (_main_multijob, tutorial)

build.yaml
```diff
@@ -694,13 +811,124 @@ common_cli:
...
+    # clusterhq-dev-archive for our vagrant boxes
+    # the file with the credentials are deployed by the jenkins master to the
+    # /tmp of the slave when the slave is instantiated.
+    . /tmp/s3_dev_archive_clusterhq.com
+
   upload_new_docs_html_to_s3_staging_docs: &upload_new_docs_html_to_s3_staging_docs |
     # uploads the new generated html to a folder on the S3 staging-docs bucket
     cd _build/html
-    aws s3 sync . s3://clusterhq-staging-docs/\$TRIGGERED_BRANCH
+    aws --region us-west-2 s3 sync . s3://clusterhq-staging-docs/\$TRIGGERED_BRANCH
```
jml: Please also update this region to use a constant.

build.yaml
```diff
@@ -694,13 +811,124 @@ common_cli:
...
+    # uploads the vagrant box and metadata.json files
+    # We won't overwrite the json metadata files if the upload of the box
+    # failed
+    (
+      s3_upload \${S3_BUCKET} "\${BOX_NAME}" &&  \
+      s3_upload \${S3_BUCKET} "\${BOX_NAME}.json" &&  \
+      s3_upload \${S3_BUCKET} "metadata.json"
+    ) || abort_build
+
+    echo "This vagrant box is available is available at :"
```
jml: Nit: only one "is available" needed

vagrant/tutorial/Vagrantfile
```diff
@@ -11,7 +11,14 @@ unless Vagrant.has_plugin?('vagrant-reload')
 end
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
-  config.vm.box = 'box-cutter/centos70'
+
+  config.vm.provider :virtualbox do |vb|
+    # VirtualBox 5 now supports paravirtualization
+    vb.customize("pre-boot", ["modifyvm", :id, "--paravirtprovider", "KVM"])
```
jml: :question: Is there any way to factor out the paravirtualization stuff so that it's not repeated everywhere?

build.yaml
```diff
@@ -169,6 +162,105 @@ common_cli:
...
+      max_tries=\${1}
+      seconds=\${2}
+      shift 2
+
+      let delay=15
+      let tries=0
+      while [ \${tries} -lt \${max_tries} ]; do
+        echo "executing...  \${*}"
+        if (timeout --foreground --signal=SIGKILL \${seconds} bash -c "\${*}"); then
+            break
```
jml: What error message will we see if this timeout gets exceeded? Is it something that will help someone who isn't familiar with this code to figure out what's going on? 

If not, please add an error message.

build.yaml
```diff
@@ -513,9 +599,9 @@ common_cli:
   # By doing this we speed up the bootstrapping of our client/acceptance tests.
   #
   build_dockerfile_centos7: &build_dockerfile_centos7 |
-    # Download the latest pip requirements file from master branch of flocker
+    # Download the latest pip requirements file from flocker
     wget -c \
-    https://raw.githubusercontent.com/ClusterHQ/flocker/master/requirements.txt
+    https://raw.githubusercontent.com/ClusterHQ/flocker/\${TRIGGERED_BRANCH}/requirements.txt
```
jml: No change required, but just curious if this fixes a bug or something? It looks like a really good idea.

build.yaml
```diff
@@ -663,29 +764,45 @@ common_cli:
...
-    JOB_EXIT_STATUS="\$( updateExitStatus \$? )"
-    vagrant box add --force 'jenkins-osx-yosemite'   jenkins-osx-yosemite.box
-    JOB_EXIT_STATUS="\$( updateExitStatus \$? )"
-    rm -f jenkins-osx-yosemite.box
-    vagrant destroy -f
-    JOB_EXIT_STATUS="\$( updateExitStatus \$? )"
-    exit \${JOB_EXIT_STATUS}
+    # let's stop it and wait a bit, as sometimes the box is not fully down
+    # when the vagrant halt command finishes.
+    vagrant halt ; sleep 60
```
jml: :question: Sleeping is a bit of a compromise. Either it's going to be too long a sleep and create unnecessary delays (perhaps irrelevant given how slow everything else is), or it's going to be too short and we'll get bugs. Is there a way to poll until the box is fully down?

build.yaml
```diff
@@ -694,13 +811,124 @@ common_cli:
...
+    # We won't overwrite the json metadata files if the upload of the box
+    # failed
+    (
+      s3_upload \${S3_BUCKET} "\${BOX_NAME}" &&  \
+      s3_upload \${S3_BUCKET} "\${BOX_NAME}.json" &&  \
+      s3_upload \${S3_BUCKET} "metadata.json"
+    ) || abort_build
+
+    echo "This vagrant box is available is available at :"
+    echo "\${S3_BOX_JSON_LATEST}"
```
jml: :question: These two stanzas (`upload_vagrant_basebox_osx_yosemite_to_s3` and `upload_vagrant_tutorial_basebox_to_s3`) are really similar. Any way to share more code?

build.yaml
```diff
@@ -1210,26 +1439,59 @@ job_type:
...
+                   # it installs XCode command line tools and other required
+                   # packages.
                    *begin_build_sh_EOF,
+                   *add_shell_functions,
                    *install_pkgs_on_vanilla_osx_yosemite_box,
                    *end_build_sh_EOF,
-                   *copy_files_to_osx_yosemite_box,
+
+                  # starts and provisions an OSX vagrant machine
+                  # using the build.sh script created above
```
jml: Nit: please align comments with the rest of the list.

jml:
Thanks!

Most of the comments are pretty small. Please address them and merge when ready.

build.yaml
Azulinho: do you mean ?
retry_n_times_with_timeout <max retries> <timeout in seconds>

build.yaml
Azulinho: sorry, which step ?
github excerpts can be confusing, are you refering to L737 ?

if the build.sh fails, then abort_build is executed
'||' in bash means 'OR' and evaluated left to righ

```
\>ls /non-existing-file || echo "that's was borked"
ls: cannot access /non-existing-file: No such file or directory
that's was borked
```

build.yaml
Azulinho: no one talks like that in south-london

Azulinho: I always 'we'

build.yaml
Azulinho: it is consumed in a couple of points
```
\>grep get_s3_creds_for_bucket_vagrant_jenkins_boxes build.yaml 
  get_s3_creds_for_bucket_vagrant_jenkins_boxes: &get_s3_creds_for_bucket_vagrant_jenkins_boxes |
                   *get_s3_creds_for_bucket_vagrant_jenkins_boxes,
                   *get_s3_creds_for_bucket_vagrant_jenkins_boxes,
```

build.yaml
Azulinho: they are on the fat side, however these jobs take a long time to run and only run overnight.
They're not part of the multijob

vagrant/tutorial/Vagrantfile
Azulinho: possibly, but not sure where else these files are used. Likely to be breaking more stuff than currently fixing

build.yaml
Azulinho: not that I know off, 
10 seconds were probably enough, 60s is mostly cause they won't make any difference on a 60min long job

build.yaml
Azulinho: allows me to test changes to that job based on my current branch.

build.yaml
Azulinho: There is a message that shows the command failed, just a few lines below:
```
 echo Command failed, waiting \${delay}...  
```
timeout won't produce any usefull output AFAICSFTMP

Azulinho:
thanks @jml 
ALL GREEN
and with that I now close this Epic

build.yaml
jml: I mean that:

 1. `delay` specifically should have a comment saying that it's in seconds
 2. Maybe `delay` should be a parameter, e.g. 

```
retry_n_times_with_timeout <max retries> <timeout in seconds>  <initial delay in seconds>
```

build.yaml
jml: Fair enough re Github. I mean the step for removing secrets files won't get executed if `abort_build` is executed. I haven't checked the code for `abort_build` to see if it also removes secrets.

PR #2353, opened 2015-12-16
[FLOC-3615] Use correct keys for NO_NEW_DEVICE_IN_OS Eliot message

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3615

 * I fixed the Eliot keys for this message, `EXPECTED_SIZE` rather than `SIZE`.
 * added a test to demonstrate that the message is logged when a device does not appear.

Note that there is another bug in this code which is described in 
 * https://clusterhq.atlassian.net/browse/FLOC-3578

We shouldn't be raising `AttachedUnexpectedDevice` when we hit the `NO_NEW_DEVICE_IN_OS` situation. It doesn't make sense and it prevents the retry loop in the parent method from retrying the attachment.


itamarst:
Merge when green.

wallrj:
The new test revealed that `lsblk /dev/loopX` command exits with return code `1` when you supply a loop device that isn't attached to a backing file. (i.e. it doesn't show up in `losetup --all`).

I considered catching `CalledProcessError` and returning `0` size in that case.
But the return code, in the Ubuntu 14.04 version of lsblk, is `1` and there's no `stderr` output to distinguish this case from other unexpected failures.

I also tried `blockdev --getsize64 /dev/loopX` which returns `0` for unbacked loop devices, but requires root access to issue ioctl calls on the block device.

Finally I decided up on the simpler method of reading `/sys/block/<disk>/size` and multiplying by 512. (the  sector size in bytes).

My only worry is  whether the sector size  / block size is always 512.

flocker/node/agents/test/test_ebs.py
```diff
@@ -189,6 +192,23 @@ def test_unexpected_device_discovered(self, device):
...
+            _attach_volume_and_wait_for_device,
+            volume=self.volume,
+            attach_to=self.compute_id,
+            attach_volume=lambda *a, **kw: None,
+            detach_volume=lambda *a, **kw: None,
+            device=b'/dev/sdh',
+            blockdevices=[],
+            time_limit=0,
+        )
+
```
exarkun: This test relies on the code being able to read some relevant information from `/sys/block`.  So it's subject to interference from other activity on the system.  Since it's only a test for the logging behavior, that's a bit unfortunate.  This is basically a symptom of the factoring of the ebs driver, though, which smears low-level system interactions throughout - so it is hard to avoid having such an interaction in any particular code path.

I'm torn between suggesting we fix the factoring so this doesn't have to be the case and letting this pass for now as a consequence of the existing state of the system.

However, it also occurs to me that if the test exercised `_wait_for_device` directly and `_wait_for_device` parameterized `FilePath("/sys/block")` and `FilePath("/dev")` then this interaction could be avoided...  Probably?

wallrj: Yep, testing `_wait_for_device` directly is much simpler and doesn't require any changes to the time_limit arguments. I've moved the test to to a new `WaitForDeviceTests` case.

wallrj:
The run_client_installation_on_OSX builder failed again, apparently due to some virtualbox locking problem.
That builder is going to be changed to a nightly build anyway so I'll merge this now.

PR #2357, opened 2015-12-17
[FLOC-3733] Serialize non-inspectable callables for loop_until logging

jongiddy:
Passing a `functools.partial` to `loop_until` fails to serialize correctly.

Add an extra exception clause to the `function_serializer` function to handle callables that are rejected by the `inspect` module functions.  In particular, handle `partial` usefully by adding the details of the wrapped function.

Although it would be possible, I have not included the partial's arguments in the logging, as they may cause further serialization problems or expose sensitive information.

### DIFF
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -43,6 +43,16 @@ def function_serializer(function):
         return {
             "function": str(function),
         }
+    except TypeError:
+        # Callable not supported by inspect module
+        if isinstance(function, partial):
+            return {
+                'partial': function_serializer(function.func)
+            }
+        else:
+            return {
+                "function": str(function),
+            }
 
 
 class LoopExceeded(Exception):
--- a/flocker/common/_retry.py
+++ b/flocker/common/_retry.py
@@ -44,7 +44,7 @@ def function_serializer(function):
             "function": str(function),
         }
     except TypeError:
-        # Callable not supported by inspect module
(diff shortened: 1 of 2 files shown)


flocker/common/_retry.py
```diff
@@ -43,6 +43,16 @@ def function_serializer(function):
         return {
             "function": str(function),
         }
+    except TypeError:
```
zubron: Which operation within the `try` clause will cause a `TypeError` to be raised?

flocker/common/test/test_retry.py
```diff
@@ -193,6 +194,32 @@ def predicate():
...
+            return result
+        predicate = partial(check)
+
+        clock = Clock()
+        d = loop_until(clock, predicate)
+        self.assertEqual(
+            self.successResultOf(d),
+            result)
+
+        action = LoggedAction.of_type(logger.messages, LOOP_UNTIL_ACTION)[0]
```
zubron: :question: You could use list unpacking here instead to ensure that there is only one `LOOP_UNTIL_ACTION`.

zubron:
Thanks, @jongiddy. I've made a request for a comment to be added and one other optional change. Please address and merge when you're satisfied with the test results.

flocker/common/_retry.py
jongiddy: Updated comment.

PR #2358, opened 2015-12-17
[FLOC-3712] do not overwrite cluster.crt on the control node

avg-I:
flocker-control may be already starting up when cluster.crt gets
written again along with the node certificate and key files.

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -730,11 +730,13 @@ def task_install_control_certificates(ca_cert, control_cert, control_key):
         ])
 
 
-def task_install_node_certificates(ca_cert, node_cert, node_key):
+def task_install_node_certificates(ca_cert, install_ca_cert,
+                                   node_cert, node_key):
     """
     Install certificates and private key required by a node.
 
     :param FilePath ca_cert: Path to CA certificate on local machine.
+    :param bool install_ca_cert: Install the CA certificate if ``True``.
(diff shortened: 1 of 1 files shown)


jml:
Thanks Andriy.

This doesn't fix the ssl handshake error problem (see http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/FLOC-3712-do-not-overwrite-cluster-crt/job/run_acceptance_on_AWS_CentOS_7_for_flocker.acceptance.endtoend/1/artifact/remote_logs.log), but it's still probably a good fix. 

Please update FLOC-3712, and merge when you're satisfied with the test results.

avg-I:
Please note that the error is different from the error in FLOC-3712. The original error had `SSL3_READ_BYTES` in it while the new error has `SSL23_READ`. That hints at a different protocol version and it is quite possible that the error is a transient error during protocol negotiation, but I am not sure about that.
Also, judging from http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/FLOC-3712-do-not-overwrite-cluster-crt/job/run_acceptance_on_AWS_CentOS_7_for_flocker.acceptance.endtoend/1/consoleText there were some successful tests. With the original problem it was impossible to communicate with `flocker-control` at all.

avg-I:
Seems that the root cause of the problem is different what I thought it was.
Just reproduced the problem with this change...

PR #2362, opened 2015-12-17
[FLOC-3295] Implement read request load scenario for benchmarking tool

zubron:
Fixes [FLOC-3295](https://clusterhq.atlassian.net/browse/FLOC-3295)

Paired with: @mpgomez 

This change adds support for a read request load scenario which can be used to create a specified load from read requests when performing operations.

The scenario can be used as follows:

```
./benchmark --cluster cluster --scenario read-request-5 --metric cputime --operation no-op
```

This will start a read request load during the benchmarking run of 5 read requests per second.

When a scenario is being used during benchmarking, the scenario will not be in a 'started' state until the requested load rate has been achieved. If the load rate drops during the benchmarking run, it will throw an exception and terminate the run. It's likely that we will want to add tolerance to the check for the request rate in the future but it hasn't been addressed in this PR. When the scenario is stopped, it will wait for any outstanding requests to be completed within a provided timeout period.


benchmark/scenarios/read_request_load.py
```diff
@@ -0,0 +1,250 @@
...
+from .._interfaces import IScenario
+
+DEFAULT_SAMPLE_SIZE = 5
+
+
+class RateMeasurer(object):
+    """
+    Measures the rate of requests in requests per second.
+
+    :ivar sample_size: The number of samples to collect.
```
dwgebler: Document the other ivars.

benchmark/scenarios/read_request_load.py
```diff
@@ -0,0 +1,250 @@
...
+        self._rate = 0
+        self.sample_size = sample_size
+
+    def request_sent(self):
+        """
+        Increase the number of sent requests.
+        """
+        self.sent += 1
+
+    def response_received(self, result):
```
dwgebler: What does `result` do here?

benchmark/scenarios/read_request_load.py
```diff
@@ -0,0 +1,250 @@
...
+        """
+        Increase the number of received requests.
+        """
+        self.received += 1
+
+    def request_failed(self, result):
```
dwgebler: Same question.

benchmark/scenarios/__init__.py
```diff
@@ -1 +1,5 @@
 from .no_load import NoLoadScenario
+from .read_request_load import (
+    ReadRequestLoadScenario, RateMeasurer, RequestRateTooLow,
+    RequestRateNotReached, RequestOverload
+)
```
jongiddy: Only export truly public objects in `__init__.py`. The `RateMeasurer` is an internal object, and it's only expected use is in the test module.  The test module can import directly from the module being tested.

benchmark/scenarios/read_request_load.py
jongiddy: It appears that Twisted code documents internal attributes, but that the documentation processing strips out attributes that start with an underscore.  Hence, please add the private attributes to the docstring.

benchmark/scenarios/read_request_load.py
```diff
@@ -0,0 +1,261 @@
...
+        )
+        self._counts.append(self._received)
+
+    def outstanding(self):
+        """
+        Return the number of outstanding requests.
+        """
+        return self._sent - self._received - self._errors
+
+    def rate(self):
```
jongiddy: Add a simple docstring, echoing what the rate is.

jongiddy:
A few minor comments. Please merge when addressed. Thanks.

PR #2363, opened 2015-12-17
[FLOC-3483] Make sure all @flaky tests are retried

jml:
We had the `@flaky` decorator on a bunch of tests that weren't being retried, because that requires a base `TestCase` that does the retries and we wanted to fix some regression (esp. re logging) before we spread the new base TestCases around the flocker code base.

Now that the regressions are fixed, we want to start retrying everything decorated with `@flaky`. This is it.

The guts of the change are:

* Change `flocker.testtools.TestCase` to inherit from [testtools](https://testtools.rtfd.org) `TestCase`
* Share compatibility code written to work with `AsyncTestCase` to also work with `TestCase`, and update tests
* Update tests for `@flaky` retries to operate on standard `testtools.TestCase`.
* Fix ordering issue in blockdevice tests that arose out of differences in implementation in `assertItemsEqual`
* Add `assertFailure` compatibility helper w/ tests
* Removed the temporary logging code and replace it with Eliot logs (printing to stdout caused some failures in Jenkins due to subunit parsing issues)
* Update tests to use the new base test cases

### DIFF
--- a/flocker/testtools/test/test_base.py
+++ b/flocker/testtools/test/test_base.py
@@ -13,8 +13,11 @@ import unittest
 from eliot import MessageType, fields
 from hypothesis import assume, given
 from hypothesis.strategies import binary, integers, lists, text
-from testtools import PlaceHolder, TestResult
+
+# Use testtools' TestCase for most of these tests so that bugs in our base test
+# case classes don't invalidate the tests for those classes.
 from testtools import TestCase as TesttoolsTestCase
+from testtools import PlaceHolder, TestResult
 from testtools.matchers import (
     AllMatch,
     AfterPreprocessing,
--- a/flocker/testtools/test/test_base.py
+++ b/flocker/testtools/test/test_base.py
@@ -10,7 +10,7 @@ import unittest
 
 from eliot import MessageType, fields
 from hypothesis import assume, given
-from hypothesis.strategies import binary, integers, lists, sampled_from, text
+from hypothesis.strategies import binary, integers, lists, text
 from testtools import PlaceHolder, TestResult
 from testtools import TestCase as TesttoolsTestCase
 from testtools.matchers import (
@@ -37,8 +37,6 @@ from testtools.matchers import (
 from twisted.python.filepath import FilePath
 
 from .._base import (
-    AsyncTestCase,
-    TestCase,
     make_temporary_directory,
     _SplitEliotLogs,
     _get_eliot_data,
@@ -46,13 +44,11 @@ from .._base import (
     _path_for_test_id,
 )
 from .._testhelpers import (
+    base_test_cases,
     has_results,
     only_skips,
     run_test,
 )
-
-
-base_test_cases = sampled_from([AsyncTestCase, TestCase])
 
 
 class BaseTestCaseTests(TesttoolsTestCase):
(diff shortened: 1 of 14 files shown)


flocker/testtools/test/test_base.py
```diff
@@ -11,7 +11,8 @@
 from eliot import MessageType, fields
 from hypothesis import assume, given
 from hypothesis.strategies import binary, integers, lists, text
-from testtools import PlaceHolder, TestCase, TestResult
+from testtools import PlaceHolder, TestResult
+from testtools import TestCase as TesttoolsTestCase
```
tomprince: The reason why we use this might warrant a comment. It took staring at the code to realize that we are
- using this to test that our testcase subclasses behave sanely
- don't want to use the subclasses themselves, in case they are buggy

flocker/testtools/test/test_flaky.py
```diff
@@ -134,9 +130,9 @@ def test_successful_flaky_test(self):
 
         # We use 'unittest' here to avoid accidentally depending on Twisted
```
tomprince: We don't appear to be using stdlib unittest here anymore.

flocker/testtools/test/test_flaky.py
```diff
@@ -322,3 +297,68 @@ def test_delayed(self):
...
+            @flaky(jira_keys, max_runs, min_passes)
+            def test_error(self):
+                1/0
+
+        test = ErroringTest('test_error')
+        self.assertThat(
+            run_test(test),
+            has_results(
+                tests_run=Equals(1),
+                errors=MatchesListwise([
```
tomprince: I'm not sure that I understand what this is supposed to be matching, or how this is related to the test description. Please add a comment explaining what a sample expected results looks like, and why that indicates that the test has been retried.

flocker/testtools/test/test_flaky.py
```diff
@@ -322,3 +297,68 @@ def test_delayed(self):
...
+        class SkippingTest(base_test_case):
+            run_tests_with = retry_flaky()
+
+            @flaky(jira_keys, max_runs, min_passes)
+            def test_skip(self):
+                observed_reasons.append(reasons.next())
+                raise unittest.SkipTest(observed_reasons[-1])
+
+        test = SkippingTest('test_skip')
+        self.assertThat(run_test(test), only_skips(1, observed_reasons))
```
tomprince: I'm not sure that I understand the reason for the `1` here.

tomprince:
`flocker.node.functional.test_docker.GenericDockerClientTests.test_pull_image_if_necessary` and `test_restart_policy_always` are on a twisted `TestCase` but have the flaky decorator applied to them. That should be changed. I wonder if (as a followup?) it should be made an error to apply `@flaky` to something that isn't using the `retry_flaky` run-test.

tomprince:
Please address the above comments and then merge.

tomprince:
(once jenkins is happy enough too)

flocker/testtools/test/test_flaky.py
exarkun: It's just asserting that one test was run.

exarkun:
Comments addressed.  Build is pretty good.  Red looks like known issues.  Merging.

PR #2364, opened 2015-12-17
[FLOC-3713] More carefully enable test_simple_transitions.

sarum90:
Resubmission of #2347 that uses less loop devices at a time.

This was merged yesterday, but then caused errors on master (https://clusterhq.atlassian.net/browse/FLOC-3738) and was reverted.

I was easily able to reproduce those errors on Jenkins:

Builds #1 and #2 at:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/more-carefully-enable-test-simple-transitions-FLOC-3713/job/run_trial_on_AWS_Ubuntu_Trusty_as_root_flocker.node.agents/

I believe the issue is that Ubuntu cannot handle creating ~30 loop devices, which is approximately what this test does if you don't properly clean up the loop devices after each hypothesis example.

We were destroying the loop devices, but we weren't unmounting all of the locations they were mounted to. Loop devices aren't actually destroyed until they are unmounted, so this resulted in us not actually destroying the loop devices between runs. The added unmount results in only 1 loop device being used by this test rather than ~30.

I plan on running run_trial_on_AWS_Ubuntu_Trusty_as_root_flocker.node.agents 3 or 4 more times to verify the fix before merging.

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -8,7 +8,6 @@ from errno import ENOTDIR
 from functools import partial
 from os import getuid
 import time
-from time import sleep
 from uuid import UUID, uuid4
 from subprocess import STDOUT, PIPE, Popen, check_output, check_call
 from stat import S_IRWXU
@@ -41,7 +40,6 @@ from twisted.python.components import proxyForInterface
 from twisted.python.runtime import platform
 from twisted.python.filepath import FilePath
 from twisted.trial.unittest import SynchronousTestCase, SkipTest, TestCase
-from twisted.internet.task import Clock
 
 from eliot import start_action, write_traceback, Message, Logger
 from eliot.testing import (
@@ -114,9 +112,7 @@ from ....control._model import Leases
 
 # Move these somewhere else, write tests for them. FLOC-1774
 from ....common.test.test_thread import NonThreadPool, NonReactor
-from ....common import (
-    retry_failure, gather_deferreds, RACKSPACE_MINIMUM_VOLUME_SIZE
-)
+from ....common import RACKSPACE_MINIMUM_VOLUME_SIZE
 
 CLEANUP_RETRY_LIMIT = 10
 LOOPBACK_ALLOCATION_UNIT = int(MiB(1).to_Byte().value)
@@ -1208,7 +1204,7 @@ class BlockDeviceCalculatorTests(SynchronousTestCase):
         """
         Cleanup after running a hypothesis example.
         """
-        umount_all(self.deployer.mountroot, self)
+        umount_all(self.deployer.mountroot)
         detach_destroy_volumes(self.deployer.block_device_api)
 
     def current_datasets(self):
@@ -4159,13 +4155,11 @@ def umount(unmount_target):
     check_output(['umount', unmount_target.path])
 
 
-def umount_all(root_path, test_case):
+def umount_all(root_path):
     """
     Unmount all devices with mount points contained in ``root_path``.
 
     :param FilePath root_path: A directory in which to search for mount points.
-    :param TestCase test_case: A Twisted ``TestCase`` with a
-        ``successResultOf`` method.
     """
     def is_under_root(path):
         try:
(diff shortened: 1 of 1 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1182,6 +1206,7 @@ def teardown_example(self, token):
         """
         Cleanup after running a hypothesis example.
         """
+        umount_all(self.deployer.mountroot)
```
tomprince: Unfortunately, `umount_all` returns a deferred, so
- this won't complete immediately
- it may end up cleaning up things from the next iteration.

The simplest solution(:frowning:) would be to change `umount_all` to not be asynchronous. A better long term solution would be to teach hypothesis how to run deferred running functions (by having an executor that does that[1]) but that is outside the scope of this issue.

[1] Maybe something that does `@hypothesis.executors.executor.extend(AyncTestCase)`.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -148,19 +148,43 @@
...
+    DESIRED_DATASET_ATTRIBUTE_STRATEGIES
+)
+
+_NoSuchThing = object()
+
+# This strategy creates two `DESIRED_DATASET_STRATEGY`s with as much in common
+# as possible. This is supposed to approximate two potentially different
+# `DesiredDataset` states for the same dataset.
+TWO_DESIRED_DATASET_STRATEGY = DESIRED_DATASET_STRATEGY.flatmap(
+    lambda x: tuples(just(x), tagged_union_strategy(
```
tomprince: This seems like something that could be abstracted.

tomprince:
Unfortunately, I don't think we can get away with calling the current `umount_all` synchronously. Please resubmit after addressing that issue.

flocker/node/agents/test/test_blockdevice.py
sarum90: Nice catch! That's actually a pre-existing bug... umount_all is already used in contexts where it is expected to be synchronous (people call addCleanup on SyncronousTestCasts with umount_all... I'll look into how to fix that.

flocker/node/agents/test/test_blockdevice.py
sarum90: Yea I tried to write it in a way that could be re-used. My general rule is if you only want to use something once, do it in-line when you are tempted to copy it make it a function.

I'd like to see a second use case to see where the right abstraction layer is before making this a proper utility in common.algebraic or something. If something strikes me in the next hour I'll code it up, otherwise I'm thinking about leaving this as is for now.

flocker/node/agents/test/test_blockdevice.py
sarum90: So I realized that the reason umount_all had that retry_failure nonsence was to work around situations where we were umounting by blockdevice and the blockdevices were mounted to the same mountpoint. Now that we are umounting by mountpoint we don't need the retry logic. This made it trivial to make the function synchronous.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1231,22 +1257,26 @@ def run_to_convergence(self, desired_datasets, max_iterations=4):
...
-        initial_dataset = initial_dataset.set(
-            mount_point=self.deployer._mountpath_for_dataset_id(
-                unicode(dataset_id)),
-        )
+
+        # Set the mountpoint to a real mountpoint in desired dataset states
+        # that have a mount point attribute.
+        mount_point = self.deployer._mountpath_for_dataset_id(
+            unicode(dataset_id))
+        if hasattr(initial_dataset, 'mount_point'):
```
zubron: :question: After the recent discussions about whether to use `hasattr` or `getattr`, is that something that we need to consider here?

zubron:
Thanks, @sarum90! This looks good. Just one question, but it won't block this being merged. I like your changes to `umount_all`, much more concise now :+1: 

The test is now passing on Jenkins and I've successfully run it locally. I was able to reproduce the issue of being unable to create the extra loop devices with the previous PR. I don't know if @tomprince wants to look at this again before merging but I'm happy for you to just address the one comment and then merge when you're satisfied with the rest of the test results.

flocker/node/agents/test/test_blockdevice.py
sarum90: Thanks, fixed.

jml:
Meta-review all good by me, thanks @zubron.

sarum90:
Waiting to see some green with the latest commit, then merging this.

I'll be sure it doesn't break master as well, although I'm like 99% sure I've fixed the problem that originally caused that.

sarum90:
Actually, the most recent commit only changes test_simple_transitions, and I have green runs of that test for commit 0f9a23a on both CentOS and Ubuntu (http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/more-carefully-enable-test-simple-transitions-FLOC-3713/job/run_trial_on_AWS_CentOS_7_as_root_flocker.node.agents/lastSuccessfulBuild/consoleFull and http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/more-carefully-enable-test-simple-transitions-FLOC-3713/job/run_trial_on_AWS_Ubuntu_Trusty_as_root_flocker.node.agents/lastSuccessfulBuild/consoleFull) So I'm going to go ahead and merge this now.

I'll follow up with bugs for Green CI from failing runs in the CI tests.

PR #2365, opened 2015-12-17
[FLOC-3712] Make flocker-control serve the right service on the right port

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3712

This removes flocker-control socket activation.  The systemd documentation suggests that the way Twisted's systemd socket inheritance feature works makes it unreliable to use that feature to heterogeneous service configuration.

(discovered and fixed while pairing with @jml)

### DIFF
--- a/admin/package-files/systemd/flocker-control-agent.socket
+++ b/admin/package-files/systemd/flocker-control-agent.socket
@@ -1,9 +1,0 @@
-[Unit]
-Description=Flocker Control Agent Socket
-
-[Socket]
-ListenStream=4524
-Service=flocker-control.service
-
-[Install]
-WantedBy=sockets.target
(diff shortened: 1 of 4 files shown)


tomprince:
This makes me sad, but the code changes look obviously correct, given the diagnosis. Please merge once jenkins is happy. Please also file tickets against twisted for
- documenting this limitation
- providing support for addressing passed description by name.

jml:
* docs: https://twistedmatrix.com/trac/ticket/8146#ticket
* support for names:  https://tm.tl/#8147

exarkun:
Acceptance test runs pointed out that we were still using a private network namespace, preventing anyone from connecting to flocker-control network services.  Removed those, hopefully next builds will be good.  If so, I'll merge.

exarkun:
Builds had some problems on Jenkins.  Problems all look unrelated.  Local test runs look good, merging.

PR #2366, opened 2015-12-18
[FLOC-3665] run-acceptance-tests: make cluster clean up more robust

avg-I:
Because main() is decorated with @inlineCallbacks there is no guarantee
that the control flow always returns to it after a yield statement.
In particular, that's the case if the reactor is stopped.
And that could happen, for instance, if a user presses Ctrl-C because
twisted installs a SIGINT handler which stops the reactor.

Thus, to improve the chances of the cluster clean-up being executed
the clean-up code is moved to a before-shutdown system event trigger.

Also, we now destroy the cluster if we couldn't finish its setup
even if --keep option is set.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -1176,16 +1176,15 @@ def main(reactor, args, base_path, top_level):
 
     setup_succeeded = False
     reached_finally = False
-    stop_cluster = True
 
     def cluster_cleanup():
         if not reached_finally:
             print "interrupted..."
-        if stop_cluster:
-            print "stopping cluster"
-            return runner.stop_cluster(reactor)
+        print "stopping cluster"
+        return runner.stop_cluster(reactor)
 
-    reactor.addSystemEventTrigger('before', 'shutdown', cluster_cleanup)
+    cleanup_trigger_id = reactor.addSystemEventTrigger('before', 'shutdown',
+                                                       cluster_cleanup)
 
     try:
         yield runner.ensure_keys(reactor)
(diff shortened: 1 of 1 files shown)


admin/acceptance.py
```diff
@@ -1212,25 +1227,27 @@ def main(reactor, args, base_path, top_level):
...
+        elif not options['keep']:
+            print "not keeping cluster"
         else:
             print "--keep specified, not destroying nodes."
-            if cluster is None:
-                print ("Didn't finish creating the cluster.")
-            else:
-                print ("To run acceptance tests against these nodes, "
-                       "set the following environment variables: ")
+            assert cluster is not None
```
exarkun: This should be an `if` with some well-defined failure behavior.  `assert` might get "optimized" away if Python is run in a certain way and, in general, we just don't use `assert` anywhere in Flocker.

admin/acceptance.py
```diff
@@ -1212,25 +1227,27 @@ def main(reactor, args, base_path, top_level):
...
-                        value=shell_quote(
-                            environment_variables[environment_variable]),
-                    )
+            for environment_variable in environment_variables:
+                print "export {name}={value};".format(
+                    name=environment_variable,
+                    value=shell_quote(
+                        environment_variables[environment_variable]),
+                )
+            stop_cluster = False
```
exarkun: If you want, instead of this `stop_cluster` flag, you could use `reactor.removeSystemEventTrigger`.  It takes as an argument the object returned by `addSystemEventTrigger`.

exarkun:
Thanks.  Apart from minor comments inline, looks good.  Please merge when you're happy with this.

PR #2367, opened 2015-12-18
[FLOC-3706]  place a 30 second timeout on a secureConnection() attempt 

avg-I:
I've observed an ssh connection getting indefinitely stuck here.

Plus a couple of small improvements in the nearby code.

### DIFF
--- a/flocker/provision/_ssh/_conch.py
+++ b/flocker/provision/_ssh/_conch.py
@@ -68,7 +68,6 @@ class CommandProtocol(LineOnlyReceiver, object):
     delimiter = b'\n'
 
     def connectionMade(self):
-        from functools import partial
         self.transport.disconnecting = False
         # SSHCommandClientEndpoint doesn't support capturing stderr.
         # We patch the SSHChannel to interleave it.
--- a/flocker/provision/_ssh/_conch.py
+++ b/flocker/provision/_ssh/_conch.py
@@ -30,7 +30,7 @@ from twisted.protocols.basic import LineOnlyReceiver
 from twisted.python.filepath import FilePath
 import os
 
-from ...common import loop_until
+from ...common import loop_until, timeout
 from ._model import (
     Run, Sudo, Put, Comment, RunRemotely, perform_comment, perform_put,
     perform_sudo)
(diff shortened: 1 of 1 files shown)


flocker/provision/_ssh/_conch.py
```diff
@@ -154,7 +153,8 @@ def perform_run_remotely(reactor, base_dispatcher, intent):
 
     def connect():
         connection = connection_helper.secureConnection()
-        connection.addErrback(lambda _: False)
+        connection.addErrback(write_failure)
```
sarum90: :grey_question: No action required, this relies on write_failure returning a Falsey value for retries (it returns None as of now so it's probably okay). It be sort of bad if someone changed eliot so that write_failure returned something else and suddenly we lost retries. That also seems pretty low risk to me though, so fine as is.

sarum90:
I think this looks like an obvious improvement to the current code. Any attempt to connect should conclude in 30 seconds and if it hasn't we should just retry. Thanks!

Once the tests are sufficiently green go ahead and merge.

flocker/provision/_ssh/_conch.py
avg-I: Yes, and I see a couple places in the existing code that rely on the same assumption.

avg-I:
Only two Jenkins projects are failing consistently for the same reason:
* http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/FLOC-3706-ssh-connect-timeout/job/run_trial_on_AWS_Ubuntu_Trusty_flocker.common/lastFailedBuild/consoleText
* http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/FLOC-3706-ssh-connect-timeout/job/run_trial_on_AWS_CentOS_7_flocker.common/lastFailedBuild/console

```
test: flocker.common.test.test_retry.TimeoutTests.test_advancing_after_failure
error: flocker.common.test.test_retry.TimeoutTests.test_advancing_after_failure [
testtools.testresult.real._StringException: Traceback (most recent call last):
ValueError: In File: /tmp/workspace/ClusterHQ-flocker/FLOC-3706-ssh-connect-timeout/run_trial_on_AWS_Ubuntu_Trusty_flocker.common/flocker/common/test/test_retry.py
TestCase.setUp was not called. Have you upcalled all the way up the hierarchy from your setUp? e.g. Call super(TimeoutTests, self).setUp() from your setUp().

]
...
```
The message comes from `testtools.TestCase`.
It seems that the problem was exposed by f278f204e6fe5a40a313f7ffce385a4e60428668

PR #2370, opened 2015-12-18
[FLOC-3743] Fix a syntax error in the tail command used to capture upstart logs

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3743

Ubuntu acceptance test jobs on master have empty remote logs.  Hopefully equivalent jobs for this branch will have non-empty remote logs.

Also, hopefully, the change is obviously correct.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -990,7 +990,7 @@ def capture_upstart(reactor, host, output_file):
         username='root',
         command=[
             b'tail',
-            b'-F'
+            b'-F',
             b'/var/log/flocker/flocker-control.log',
             b'/var/log/flocker/flocker-dataset-agent.log',
             b'/var/log/flocker/flocker-container-agent.log',
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -970,6 +970,7 @@ def eliot_output(message):
     sys.stdout.write(format % message)
     sys.stdout.flush()
 
+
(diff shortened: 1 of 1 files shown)


sarum90:
Thanks!  I was noticing this earlier today myself, always nice to see that someone was already on top of fixing it.

I agree that looks like a bug, I agree this looks like the right fix.

http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/commas-FLOC-3743/job/run_acceptance_loopback_on_AWS_Ubuntu_Trusty_for_flocker.acceptance/lastSuccessfulBuild/artifact/remote_logs.log  Agrees with that

On master that file is 0 bytes so it seems this PR is the one that fixes it.

I glanced at the other failing tests, and none of them seem related to this change, so I'm going to go ahead and merge as is.  I'll glance over the failures and make some more JIRA issues if need be after I merge.

sarum90:
Actually... it seems a little odd that all of the acceptance tests except for the loop ones failed. I think I'll just accept for now and kick the tests for you. I'd like to see at least 1 non-loop-device passing run of acceptance tests on ubuntu to double check this isn't in some weird way affecting those tests.

sarum90:
Okay looks like everything I kicked passed. I'm going to merge this, because I think it's safe and logs are better than no logs.

Thanks JP for fixing this!

PR #2371, opened 2015-12-19
[FLOC-3692] Failing test for protecting the dataset paths using new dataset states.

sarum90:
Cleaned up version of #2339 against the right branch.


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,146 @@ class DidNotConverge(Exception):
...
+        Returns whether the blockdevice at self._device_path has a file with
+        the given name.
+        """
+        test_mountpoint = self._testing_mountroot.child(str(uuid4()))
+        test_mountpoint.makedirs()
+        try:
+            self._blockdevice_manager.mount(self._device_path, test_mountpoint)
+            result = test_mountpoint.child(filename).exists()
+            self._blockdevice_manager.unmount(self._device_path)
+            test_mountpoint.remove()
```
itamarst: You probably want the `remove()` in a `finally:`  clause, so it gets removed even if mount failed. Otherwise next time around `makedirs()` might fail.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,146 @@ class DidNotConverge(Exception):
...
+    Representation of an external client of the Flocker API. This client can be
+    used to try to write to paths on the system and verify that the writes are
+    persisted to the underlying blockdevice.
+
+    :ivar _known_mountpoints: Known mountpoints of the dataset. Once flocker
+        has returned a path to an external agent, there is no promising how
+        long that agent might cache that information before it tries to write
+        to that location. This agent keeps track of all mountpoints it has ever
+        seen and tries to write to all of them.
+    :ivar _device_path: The device path of the blockdevice that backs the
```
itamarst: This shouldn't be part of the state, it should just be an argument to appropriate utility functions.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,146 @@ class DidNotConverge(Exception):
...
+        devices = node.devices or {}
+        device = devices.get(self._dataset_id)
+        if device:
+            self._device_path = device
+
+        for path in self._known_mountpoints:
+            device = self._device_path
+            filename = str(uuid4())
+            try:
+                path.child(filename).touch()
```
itamarst: So if I get this straight, you do:

1. Touch a path on a mounted filesystem.
2. If that succeeds: makedirs, mount the filesystem again, check the file is there. Not sure yet if in same place or elsewhere.

Suggestions:
1. I'm not sure it's possible to mount the same block device twice. Is it? Maybe better logic is "unmount (if that fails we screwed up somehow), remount, check file is there".
2. Don't just touch the file. Too easy for bug somewhere (in test, elsewhere) make something pass by mistake. Instead, generate random string, write it, make sure that comes back when you read.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1270,37 +1414,116 @@ def test_simple_transitions(self, two_dataset_states):
...
+    @given(
+        two_dataset_states=TWO_DESIRED_DATASET_STRATEGY
+    )
+    def test_docker_protection(self, two_dataset_states):
+        """
+        Given an initial empty state, ``BlockDeviceCalculator`` will converge
+        to any ``DesiredDataset``, followed by any other state of the same
+        dataset. During that time, it will never be the case that a
+        successfully write to a mount path that has been sent to the control
+        agent will not be persisted to the backing volume.
```
itamarst: Not quite what we're asserting. Rather "will only be persisted to the backing volume; it might however fail to write if the dataset is not mounted."

Possibly a more clever test would actually distinguish between "Failed to write" and "successfully wrote" based on current dataset state? Cause if it's *always* fail-to-write this test will pass as is, but that's obviously not ideal.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5333,6 +5556,49 @@ def test_current_machine_is_live(self):
...
+        callback()  # After method call.
+        return result
+
+    return proxied_function
+
+
+def create_callback_blockdevice_manager_proxy(proxy_object, callback):
+    """
+    Creates a provider of ``IBlockDeviceManager`` that proxies to another
+    ``IBlockDeviceManager`` privider and calls a callback before and after
```
itamarst: `privider` is typo

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1270,37 +1414,116 @@ def test_simple_transitions(self, two_dataset_states):
...
-        # Converge to the initial state.
-        try:
-            self.run_to_convergence([initial_dataset])
-        except DidNotConverge as e:
-            self.fail(
-                "Did not converge to initial state after %d iterations." %
-                e.iteration_count)
+        evaluate_function = None
+
+        def nullable_callback():
```
itamarst: :cry:

I suggest a class with `__call__`. As it is I've no confidence `evaluate_function` will ever be called, scoping rules are too tricksy...

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1270,37 +1414,116 @@ def test_simple_transitions(self, two_dataset_states):
...
+            external_agent = _WriteVerifyingExternalClient(
+                dataset_id,
+                test_objects.deployer.node_uuid,
+                mountroot_for_test(self),
+                actual_blockdevice_manager)
+
+            evaluate_function = partial(external_agent.invariant, self,
+                                        test_objects.current_cluster_state)
+
+            # Set the mountpoint to a real mountpoint in desired dataset states
```
itamarst: This bit is repeated code that code be merged with previous test?

itamarst:
Thanks! Address (mostly about making the test more robust and less likely to pass by mistake) and resubmit; hopefully next pass will be much faster.

flocker/node/agents/test/test_blockdevice.py
sarum90: Done, thanks.

flocker/node/agents/test/test_blockdevice.py
sarum90: (2) Done. Although in the near future I intend on changing this to "run docker with a volume mount" rather than "use twisted FilePaths to change file contents" as that is actually what we want to protect against.

(1) I'm still thinking on that. It never occurred to me that mounting the same blockdevice at 2 different locations would be an odd use of mounts. I'd sort of like the method that is now named `_has_file_with_content` to work no matter what state the blockdevice is in (well, perhaps with the exception of non_manifest). Specifically I want it to work regardless of whether it's mounted or not, regardless of whether the current mount is in use by another program or not (something that might cause an unmount of the existing mount to fail.)

flocker/node/agents/test/test_blockdevice.py
sarum90: Updated the docstring (originally I just wanted sed 's/successfully/successful/') and added the cleverness based on whether or not the node reports the dataset as "manifest" in the NodeState that it sends the control node. (I had that before I hypothesized this test, but found a way to add it to the invariant only using already used interface)

flocker/node/agents/test/test_blockdevice.py
sarum90: Thanks, fixed. It was even highlighted as a typo in my editor...

flocker/node/agents/test/test_blockdevice.py
sarum90: Thanks, yea I was bitten by scoping rules a couple times on this. My confidence in evaluate_function being called comes from watching it fail with TDD :)

flocker/node/agents/test/test_blockdevice.py
sarum90: So I'll add a comment here, I think this actually has to be part of state. Otherwise after the dataset gets to ATTACHED I think the device path is removed from NodeState, so this test harness would have no hook to determine the write was persisted to a blockdevice or not without some knowledge of the fact that there is no way to get a write to a backing blockdevice if flocker isn't reporting a device in `NodeState`. This happens to be true right now, but is the sort of white-box assumption-based testing I try to avoid.

flocker/node/agents/test/test_blockdevice.py
sarum90: (1) Realized it was pretty simple to inspect the OS for a list of all mounts and re-use existing mounts of the backing blockdevice, and then fall back to re-mounting the blockdevice only if we had to. Done.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1261,6 +1485,55 @@ def run_to_convergence(self, desired_datasets, max_iterations=20):
...
+        try:
+            self.run_to_convergence([self.next_dataset])
+        except DidNotConverge as e:
+            self.test_case.fail(
+                "Did not converge to next state after %d iterations." %
+                e.iteration_count)
+
+
+@attributes([Attribute("callback", default_value=None)])
+class _NullableCallback(object):
```
tomprince: :question: `Mutable` might be more accurate than `Nullable`

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,231 @@ class DidNotConverge(Exception):
...
         """
-        Cleanup after running a hypothesis example.
+        Return the current ``BlockDeviceDeployerLocalState`` from the deployer.
+        """
+        local_state = self.test_case.successResultOf(
+            self.deployer.discover_state(_empty_node_state(self.deployer))
+        )
+
+        for change in local_state.shared_state_changes():
+            self.cluster_state = change.update_cluster_state(
```
tomprince: I *very much* dislike the fact that querying `current_local_state` mutates `cluster_state`.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,231 @@ class DidNotConverge(Exception):
...
+        :param case: A ``TestCase`` to use for making assertions.
+        :param current_cluster_state: A callable that returns the current state
+            of the cluster.
+        """
+        dataset = unicode(self._dataset_id)
+
+        # First poll the cluster state to see if there is a new mountpoint that
+        # we should try to use to attempt to write to the dataset.
+        [node] = list(n for n in current_cluster_state().nodes
+                      if n.uuid == self._node_uuid)
```
tomprince: Following up on the discussion [here](https://github.com/ClusterHQ/flocker/pull/2339#discussion_r47719027), it occurs to me that the state you are inspecting is *still* internal state. If you want the externally visible state, you'd need to look at the result of [this](https://github.com/ClusterHQ/flocker/blob/master/flocker/control/httpapi.py#L466). I don't think it is worth it to reach that far out, so please use the locally discovered state rather than the cluster state. Part of the reason is that I hope the later will change significantly in the near future, and don't want to have to have another place that references it.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,231 @@ class DidNotConverge(Exception):
...
+        :param unicode content: The expected content of the file.
+
+        :returns: Whether the blockdevice that backs ``self._dataset_id`` has a
+            file with the given name that contains the given content.
+        """
+        test_mountpoint = self._testing_mountroot.child(str(uuid4()))
+        test_mountpoint.makedirs()
+
+        # If the device is already mounted, read the file from the existing
+        # mountpoint.
```
tomprince: As discussed on slack, I think it is fine to remount an already mounted block device, and am concerned about relying on interacting with more of the system under test than is necessary. Please remove this check and just always use the fallback to mounting the device on a fresh mount.

flocker/node/agents/test/test_blockdevice.py
tomprince: As discussed above, I don't think we actually need to inspect the `cluster_state` anyway, so this can probably just be removed.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1198,16 +1205,231 @@ class DidNotConverge(Exception):
...
+        uuid=deployer.node_uuid,
+        hostname=deployer.hostname,
+        applications=[],
+    )
+
+
+@attributes(["test_case", "deployer", "cluster_state"], apply_with_init=False)
+class BlockDeviceCalculatorTestObjects(object):
+    """
+    Object to house all test objects for the tests in
```
tomprince: :question: This multiplication of objects makes me slightly unhappy, but I understand the reason for it, and I don't have any better suggestions.

tomprince:
This looks good. Make the above changes and merge once jenkins is happy.

flocker/node/agents/test/test_blockdevice.py
sarum90: Done.

flocker/node/agents/test/test_blockdevice.py
sarum90: Hrm... I wish that wasn't so far of a reach... I'll probably do a follow up actually using that.

For now using local state.

flocker/node/agents/test/test_blockdevice.py
sarum90: Agreed, Done.

flocker/node/agents/test/test_blockdevice.py
sarum90: Hrm, IMHO, setUp and tearDown are broken for the usual reasons that also back, "prefer composition over inheritance". Especially when I just want to pass a per-test parameter into setUp. In the past I've found these helper objects to be a nice compromise for code re-use.

Obviously this one is a bit more stateful than desired, sorry about that.

PR #2375, opened 2015-12-21
[FLOC-3741] Remove Jinja2

itamarst:
This rewrites the groovy file to not use Jinja2 at all; the file name is the same until the CI platform code is updated, which I may not get to in this round.

1. Viewing the diff of `jobs.groovy.j2` is probably useless. I suggest looking at old file in one editor tab and new file in another; groovy-mode or your editor's equivalent will help a lot.
2. You can compare the Jenkins XML config format *output* for various jobs (see pinned Google Doc in jenkins slack channel). I've done a first pass, but would highly recommend doing another pass.
3. You can look at Jenkins build for latest PR and see that its output and UI look right. You should also see the finished build result show up in GitHub.

Special-cased `master` and cron jobs would probably require spinning up a new Jenkins server, which I haven't gotten around to...


tomprince:
Note: The `build.yaml` changes is exactly `sed -i 's/\\\$/$/g' build.yaml`.

jobs.groovy.j2
```diff
@@ -1,22 +1,24 @@
...
+  The file is consumed by the Jenkins job setup_ClusterHQ_Flocker. That job
+  contains two build steps, the first executes a python file 'render.py' which
+  reads the build.yaml file and expands this jobs.groovy.j2 template using
+  the contents of that yaml dictionary. The resulting jobs.groovy file is then
+  read by a second build step of type 'Process Job DSL' which produces all
+  the jenkins objects.
 
-   A typical jenkins job contains the following sections, this is just an
-   example of possible actions within a particular section:
+  jinja2 is only used to populate the config, which is loaded from build.yaml.
```
tomprince: This comment is out of date.

jobs.groovy.j2
```diff
@@ -80,688 +83,645 @@
...
+           the git repository so that we can track when changes are pushed upstream.
+           We add a SCM block pointing to our flocker code base.
+        */
+        scm build_scm(git_url, branchName)
+        /* By adding a trigger of type 'github' we enable automatic builds when
+           changes are pushed to the repository.
+           This however only happens for the master branch, no other branches are
+           automatically built when new commits are pushed to those branches.
+           The logic that manages which branch is built automaticaly, is performed
+           in the triggers jinja2 macro.
```
tomprince: This is no longer a jinja2 macro.

jobs.groovy.j2
```diff
@@ -1,22 +1,24 @@
 // vim: ai ts=2 sts=2 et sw=2 ft=jinja fdm=indent et foldlevel=0
-{# jobs.groovy.j2
+/* jobs.groovy.j2
 
   This jinja2 template is used to produce a jobs.groovy file containing the
```
tomprince: A comment about how this isn't really using jinja2 now?

tomprince:
Having compared the new and old `jobs.groovy.j2`, they appear to be equivalent.

jobs.groovy.j2
```diff
@@ -80,688 +83,645 @@
...
+            job("${dashProject}/${dashBranchName}/${_job_name}") {
+                parameters {
+                    // we pass the 'MODULE' parameter as the flocker module to test with trial
+                    textParam("MODULE", _module, "Module to test" )
+                    textParam("TRIGGERED_BRANCH", branchName,
+                              "Branch that triggered this job" )
+                }
+                // limit execution to jenkins slaves with a particular label
+                label(job_values.on_nodes_with_labels)
+                directories_to_delete = ["${WORKSPACE}/_trial_temp",
```
tomprince: These are doing interpolation, which they shouldn't be. There are a number of occurrences.

tomprince:
It appears that this should be merged forward to get the changes to the acceptance tests running. It appears that the changes don't conflict, but it makes the generated configurations difficult to compare.

tomprince:
> It appears that this should be merged forward to get the changes to the acceptance tests running. It appears that the changes don't conflict, but it makes the generated configurations difficult to compare.

It turns out this is just due to old jobs sticking around.

jobs.groovy.j2
```diff
@@ -80,688 +83,645 @@
...
+                for (job_type_entry in GLOBAL_CONFIG.job_type) {
+                    job_type = job_type_entry.key
+                    job_type_values = job_type_entry.value
+
+                    /* add the 'run_trial' style jobs, where there is a job per module */
+                    if (job_type in ['run_trial', 'run_trial_for_storage_driver',
+                                     'run_acceptance']) {
+                        for (job_entry in job_type_values) {
+                            for (_module in job_entry.value.with_modules) {
+                                _job_name = job_entry.key + '_' + _module
```
tomprince: This is missing `.replace('/','_')`

tomprince:
> Special-cased master and cron jobs would probably require spinning up a new Jenkins server, which I haven't gotten around to...

Doing this could be handled by creating a test branch that special-cases itself.

tomprince:
I checked the xml diff by logging into the jenkins server then doing
```
cd /var/lib/jenkins/jobs/ClusterHQ-flocker/jobs/
find master remove-jinja-FLOC-3741/ -name config.xml | cpio -ov > /tmp/config.cpio
```
to extract all the config files.

Other than as mentioned above, the differences are:
- ordering in the main and multijob configuration
- newlines in shell-scripts aren't escaped (groovy escapes newlines preceded by `\` which jinga2 splatted in)
- triggers tied to `master`
None of those are significant.

Please resubmit after addressing the above points.

itamarst:
I don't really understand what you mean by "special-cased branch" - add another branch in addition to "master"?

tomprince:
This looks good. I tested against a branch that had the triggers based on that branch instead of "master" and the only differences are just ordering and trivial improvements from less-insane levels of quoting.

PR #2376, opened 2015-12-21
[FLOC-3740] Add call to obtain container state

jongiddy:
The `flocker.apiclient.FlockerClient` does not contain an operation to call the `/state/containers` API.

Adding it to support the benchmarking requirement to measure creation of a container and waiting for it to start.

The code is almost a combination of the call to obtain Dataset state and obtain Container configuration.  

Code for both the fake and real containers need additional information to be transferred from configuration to state during synchronization.  The `FakeFlockerClient` copies configuration to state during synchronization.  The tests for the real `FlockerClient` copy the applications data from the persistence service to the cluster state service.

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -122,12 +122,12 @@ class ContainerState(PClass):
     :attr UUID node_uuid: The UUID of a node in the cluster where the container
         will run.
     :attr unicode name: The unique name of the container.
-    :attr unicode image: The name of the Docker image.
+    :attr DockerImage image: The name of the Docker image.
     :attr bool running: Whether the container is running.
     """
     node_uuid = field(type=UUID, mandatory=True)
     name = field(type=unicode, mandatory=True)
-    image = field(type=unicode, mandatory=True)
+    image = field(type=DockerImage, mandatory=True)
     running = field(type=bool, mandatory=True)
 
 
@@ -438,7 +438,7 @@ class FakeFlockerClient(object):
             ContainerState(
                 node_uuid=container.node_uuid,
                 name=container.name,
-                image=unicode(container.image.full_name),
+                image=container.image,
                 running=True,
             ) for container in self._configured_containers.values()
         ]
@@ -786,7 +786,7 @@ class FlockerClient(object):
             return ContainerState(
                 node_uuid=UUID(container[u'node_uuid']),
                 name=container[u'name'],
-                image=container[u'image'],
+                image=DockerImage.from_string(container[u'image']),
                 running=container[u'running'],
             )
         d.addCallback(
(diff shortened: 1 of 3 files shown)


flocker/apiclient/_client.py
```diff
@@ -744,6 +779,21 @@ def list_containers_configuration(self):
         )
         return d
 
+    def list_containers_state(self):
+        d = self._request(b"GET", b"/state/containers", None, {OK})
+
+        def parse(container):
+            return ContainerState(
+                node_uuid=UUID(container['node_uuid']),
```
sarum90: :grey_question: In _parse_configuration_container we explicitly use unicode keys. For consistency, prefer using explicitly unicode keys here as well [u'node_uuid']

Also I take some issue with the client library raising and "AttributeError" if the server is misbehaving (returning a dictionary without a given key). But... I think that is a battle that is already lost. Certainly your code meets the existing standards, so please don't take any action on that grievance.  Mostly I'm just commenting so that you can ease my mind if there actually is some validation of the input that I'm missing (I don't see any input validation in _request_headers, but I know it took me awhile to find it in our server code).

flocker/apiclient/test/test_client.py
```diff
@@ -496,6 +496,29 @@ def got_result(container):
...
+            d.addCallback(lambda _ignored: self.client.list_containers_state())
+
+            d.addCallback(
+                lambda containers: [(c.node_uuid, c.name) for c in containers]
+            )
+
+            d.addCallback(
+                lambda containers: self.assertIn(
+                    (expected.node_uuid, expected.name),
+                    containers
```
sarum90: :grey_question: No action required: Looks like we don't technically have test coverage of the image and running attributes, but risk seems pretty low (we should get attribute errors in the test below if the server doesn't actually return a JSON blob with those keys I think?). Seems like an okay state of the world to me, just wanted to call it out in case you had a clever idea of how to add test coverage and just forgot to come back and add it or something.

sarum90:
Looks great!

A little surprised to be doing work adding support for container agent and container status, but if it makes benchmarking work easier, I'm all for it.

One note about using unicode keys that you should probably consider, otherwise merge when tests are sufficiently green. Thanks!

flocker/apiclient/test/test_client.py
jongiddy: I made some simple changes to make these attributes testable. So, the test now checks everything.

flocker/apiclient/_client.py
jongiddy: I have added a more specific error if the server returns a malformed response.  It may only be one call, but we can fix it in other places as we modify them.

PR #2383, opened 2015-12-22
[FLOC-3750] Change flocker.node to use new Flocker test case base classes

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3750

This also makes a change to `flocker.common.gather_deferreds`.  That function now logs failures using Eliot rather than Twisted.  This supports porting of some `flocker.node` tests which were previously using `flushLoggedErrors` (a feature of Twisted not available from testtools).

### DIFF
--- a/flocker/common/_defer.py
+++ b/flocker/common/_defer.py
@@ -6,7 +6,8 @@ Various helpers for dealing with Deferred APIs in flocker.
 """
 
 from twisted.internet.defer import gatherResults
-from twisted.python import log
+
+from eliot import write_failure
 
 
 def gather_deferreds(deferreds):
@@ -41,7 +42,7 @@ def gather_deferreds(deferreds):
 
         :param Failure failure: The ``Failure`` to be logged.
         """
-        log.err(failure)
+        write_failure(failure)
 
     for deferred in deferreds:
         deferred.addErrback(log_and_discard)
--- a/flocker/node/test/test_change.py
+++ b/flocker/node/test/test_change.py
@@ -10,7 +10,6 @@ from zope.interface import implementer
 
 from pyrsistent import PClass, field
 
-from twisted.trial.unittest import SynchronousTestCase
 from twisted.internet.defer import FirstError, Deferred, succeed, fail
 from twisted.python.components import proxyForInterface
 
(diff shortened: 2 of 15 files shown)


flocker/common/_defer.py
```diff
@@ -41,7 +42,7 @@ def log_and_discard(failure):
 
         :param Failure failure: The ``Failure`` to be logged.
         """
-        log.err(failure)
+        write_failure(failure)
 
     for deferred in deferreds:
         deferred.addErrback(log_and_discard)
```
sarum90: :grey_question: no action required, nit: I think you can just use write_failure directly here, we use write_failure to discard failures elsewhere in the code - I understand if you want it explicit, but it's fairly explicitly that way in the API (https://github.com/ClusterHQ/eliot/blob/88698c7c4603065413430edfd19018cbfa0a2ff2/eliot/_traceback.py#L99)

flocker/node/test/test_change.py
```diff
@@ -413,7 +414,7 @@ def test_changes_run_after_exception(self):
         change = in_parallel(changes=subchanges)
         result = run_state_change(change, DEPLOYER)
         self.failureResultOf(result, FirstError)
-        self.flushLoggedErrors(CustomException)
+        # self.flushLoggedErrors(CustomException)
```
sarum90: I'm guessing you meant to delete this line.

flocker/node/test/test_change.py
```diff
@@ -431,9 +432,10 @@ def test_failure_result(self):
         result = run_state_change(change, DEPLOYER)
         failure = self.failureResultOf(result, FirstError)
         self.assertEqual(failure.value.subFailure.type, RuntimeError)
-        self.flushLoggedErrors(RuntimeError)
+        # self.flushLoggedErrors(RuntimeError)
```
sarum90: And this one as well?

sarum90:
A couple missed commented out lines to remove, otherwise looks good.

Thanks!

jml:
FWIW, testtools has `testtools.deferredruntest.flush_logged_errors`. Sorry that it's not in the public docs—the Twisted docs have suffered generally, and I'd really like to find a non-global way of doing what `flush_logged_errors` does.

jml:
A Jenkins issue delayed the build of @exarkun's last commit. I've resolved the issue and kicked off a new build. Many initial builds failed due to `NullPointerException` and IO issues in Jenkins, undoubtedly unrelated to this branch. I've retried all of them.

PR #2384, opened 2015-12-23
[FLOC-3764] Use flocker test cases in flocker.node

jml:
`git grep -l 'from twisted.trial.unittest import .*TestCase'` revealed that there were still a few references to Twisted test cases after #2383. This polishes those off.

### DIFF
--- a/flocker/node/test/istatechange.py
+++ b/flocker/node/test/istatechange.py
@@ -17,9 +17,9 @@ from eliot import Logger, start_action
 from pyrsistent import PClass, field
 from characteristic import attributes
 
-from twisted.trial.unittest import SynchronousTestCase
 from twisted.internet.defer import succeed
 
+from flocker.testtools import TestCase
 from .. import IStateChange
 
 
@@ -43,7 +43,7 @@ def make_comparison_tests(klass, kwargs1, kwargs2):
             return klass(**kwargs)
         return klass(**kwargs())
 
-    class Tests(SynchronousTestCase):
+    class Tests(TestCase):
         def test_equality(self):
             """
             Instances with the same arguments are equal.
--- a/flocker/node/test/istatechange.py
+++ b/flocker/node/test/istatechange.py
@@ -19,7 +19,7 @@ from characteristic import attributes
 
 from twisted.internet.defer import succeed
 
-from flocker.testtools import TestCase
+from ...testtools import TestCase
 from .. import IStateChange
 
 
--- a/flocker/node/testtools.py
+++ b/flocker/node/testtools.py
@@ -18,7 +18,7 @@ from zope.interface import implementer
 
 from characteristic import attributes
 
-from twisted.trial.unittest import TestCase, SkipTest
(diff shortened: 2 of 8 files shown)


flocker/node/testtools.py
```diff
@@ -18,7 +18,7 @@
 
 from characteristic import attributes
 
-from twisted.trial.unittest import TestCase, SkipTest
+from twisted.trial.unittest import SkipTest
```
itamarst: Probably could be `from unittest import SkipTest`?

flocker/node/test/istatechange.py
```diff
@@ -17,9 +17,9 @@
 from pyrsistent import PClass, field
 from characteristic import attributes
 
-from twisted.trial.unittest import SynchronousTestCase
 from twisted.internet.defer import succeed
 
+from flocker.testtools import TestCase
```
itamarst: :question: Relative imports, for when we finally decide to pick a non-silly name :stuck_out_tongue_winking_eye:

itamarst:
Address minor comments if you care to (or not), merge when appropriately green.

flocker/node/testtools.py
itamarst: or actually one is supposed to use `self.skipTest` now?

flocker/node/testtools.py
```diff
@@ -248,7 +248,7 @@ def ideployer_tests_factory(fixture):
 
     :return: ``TestCase`` subclass that will test the given fixture.
     """
-    class IDeployerTests(TestCase):
+    class IDeployerTests(AsyncTestCase):
```
itamarst: You may need to update the fixture implementation to use `test.reactor` instead of global reactor.

flocker/node/testtools.py
jml: Any or either would work. I vaguely prefer `self.skipTest`. In this case, it has to be the exception, since it's raised in a decorator without access to the `TestCase`. We could refactor it to use `@skipUnless` but later.

flocker/node/test/istatechange.py
jml: Changed, but for the record, I think relative imports of testtools modules are more confusing.

flocker/node/testtools.py
jml: As discussed on Slack, `test.reactor` isn't anything special, it's just a guaranteed way of the tests getting the running reactor, so there's no _need_, but it's probably good practice. Relying on globals sucks.

I went to fix this but discovered:
* `test_blockdevice` tests use `NonReactor`
* `test_testtools` tests don't use a reactor at all
* `test_deploy` tests mostly duplicate `test_testtools` tests

Thus, I removed the duplication, and updated the test subclasses to use `SynchronousDeferredRunTest`, which handles tests that return Deferreds but don't use the reactor. 

It's not a thing we should commonly use—tests that want that should use `successResultOf` and friends—but it's a helpful check here to see whether these tests have any hidden dependencies on the reactor.

PR #2385, opened 2015-12-23
[FLOC-3765] Use flocker base test cases in acceptance tests

jml:
Previously, almost but not all of the acceptance tests used the new `flocker.testtools.TestCase` and `AsyncTestCase` base classes. This patch make it all.

Things to watch out for:

* Update trial `SkipTest` references to `self.skipTest`
* Use the `reactor` instance variable that's available on `AsyncTestCase`

This is part of a general migration to use the new base test cases. See [FLOC-3281](https://clusterhq.atlassian.net/browse/FLOC-3281).


itamarst:
Merge when appropriately green.

PR #2387, opened 2015-12-24
[FLOC-3768] Stop using Trial SkipTest

jml:
Makes it easier to grep for things that are still using Trial base cases.

In particular, I want to grep with:

```
$ grep -l 'from twisted.trial.unittest import'
```

### DIFF
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -10,7 +10,6 @@ from distutils.version import LooseVersion
 from testtools import run_test_with
 
 from twisted.internet import reactor
-from twisted.trial.unittest import SkipTest
 
 from hypothesis.strategies import integers
 
@@ -47,7 +46,7 @@ class DockerPluginTests(AsyncTestCase):
         client_version = LooseVersion(client.version()['Version'])
         minimum_version = LooseVersion(required_version)
         if client_version < minimum_version:
-            raise SkipTest(
+            self.skipTest(
                 'This test requires at least Docker {}. '
                 'Actual version: {}'.format(minimum_version, client_version)
             )
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -24,7 +24,6 @@ from keystoneclient.openstack.common.apiclient.exceptions import Unauthorized
 
 from twisted.python.filepath import FilePath
 from twisted.python.procutils import which
-from twisted.trial.unittest import SkipTest
 
 from flocker.ca import (
     RootCredential, AUTHORITY_CERTIFICATE_FILENAME, NodeCredential
@@ -119,7 +118,7 @@ class CinderBlockDeviceAPIInterfaceTests(
         try:
             config = get_blockdevice_config(ProviderType.openstack)
         except InvalidConfig as e:
-            raise SkipTest(str(e))
+            self.skipTest(str(e))
         session = get_keystone_session(**config)
         region = get_openstack_region_for_test()
         return get_cinder_v1_client(session, region)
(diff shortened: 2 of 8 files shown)


jongiddy:
The code changes appear reasonable.

However, when I run `trial flocker` for this branch on my Ubuntu 14.04 laptop, I get these errors, which do not appear in `master`:
```
[FAIL]
Failure: testtools.testresult.real._StringException: Empty attachments:
  twisted-eliot-log
  twisted-log

  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/eliot/testing.py", line 298, in <lambda>
  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/eliot/testing.py", line 409, in assertHasAction
    testCase.assertTrue(actions, "No actions of type %s" % (actionType,))
  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/unittest2/case.py", line 702, in assertTrue
AssertionError: [] is not true : No actions of type <eliot._validation.ActionType object at 0x7fa2fa30ec50>


flocker.node.agents.test.test_blockdevice.AttachVolumeTests.test_run
[FAIL]
Failure: testtools.testresult.real._StringException: Empty attachments:
  twisted-eliot-log
  twisted-log

  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/eliot/testing.py", line 298, in <lambda>
  File "/home/jon/projects/clusterhq/flocker/flocker/node/agents/test/test_blockdevice.py", line 4609, in <lambda>
    logger.messages, UNMOUNT_BLOCK_DEVICE
  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/testtools/testcase.py", line 354, in assertEqual
    self.assertThat(observed, matcher, message)
  File "/home/jon/.virtualenvs/flocker/local/lib/python2.7/site-packages/testtools/testcase.py", line 439, in assertThat
    raise mismatch_error
testtools.matchers._impl.MismatchError: 1 != 0


flocker.node.agents.test.test_blockdevice.UnmountBlockDeviceTests.test_run
```

Notably, Jenkins only appears to run these tests as root, hence failing to see any problems in a non-root context.

jongiddy:
Can you please check the above observation, and re-submit with any findings?

Also, should Jenkins also run the blockdevice tests as non-root? There are tests that are not skipped as non-root, although I guess they are also run as root. Maybe running as non-root only tests the skip functionality?

jml:
> Can you please check the above observation, and re-submit with any findings?

Done. Good catch.

Basically it's because of eliot hard-coding a check for `SkipTest` (see https://github.com/ClusterHQ/eliot/blob/master/eliot/testing.py#L301), and testtools using a different exception (see https://bugs.launchpad.net/testtools/+bug/1529143 and https://github.com/testing-cabal/testtools/blob/master/testtools/testcase.py#L194). Hopefully the comment makes it all clear.

I've changed the base test cases to use `SkipTest` and added comments. I tried using a mixin but it didn't work, and after spending 5 minutes trying to understand multiple inheritance I realized I was violating [Kernighan's debugging dictum](https://en.wikiquote.org/wiki/Brian_Kernighan) and left it as is.

While poking around, I realized that I could probably fix the skipping issue in the constructor easily, so I put together https://github.com/testing-cabal/testtools/pull/189.

> Also, should Jenkins also run the blockdevice tests as non-root? There are tests that are not skipped as non-root, although I guess they are also run as root. Maybe running as non-root only tests the skip functionality?

Running as non-root at least tests the skip functionality, but the skip functionality isn't something our users care about, it's only there to help us. By definition, skipping means we don't care about the functionality of the tests under these circumstances.

It _does_ matter for our own understanding of the code—skips are essentially documentation—so _maybe_ it's worth running the tests for that. I'd personally only be willing to consider working on that once we've got solid green on the tests & builds we already have.

As for the non-skipped tests, I don't know. Do they represent something useful for end-users that they want to run as non-root and we want to proactively support?

jongiddy:
It looks like any package introducing a new skip-test exception would integrate better if it subclassed `unittest.SkipTest`.

Anyhoo, the code looks good to me, and passes relevant tests locally and on Jenkins. Please merge when you're happy with CI.

jml:
> It looks like any package introducing a new skip-test exception would integrate better if it subclassed `unittest.SkipTest`.

Yep, that's https://bugs.launchpad.net/testtools/+bug/1529143. My hunch is that `TestSkipped` pre-dates `SkipTest` in stdlib.

PR #2388, opened 2015-12-24
[FLOC-3767] Container volume support in API client

jongiddy:
For benchmarking, we still require the container API. However, the `flocker.apiclient.Client` does not support volumes attached to containers, which is needed for benchmarking.

This PR adds support for stateful containers to the API client, both when creating containers, and when receiving information about container configuration and state.

Note the use of PEP 484 type annotations in the docstrings. Currently we have a variety of patterns for non-primitive docstring types (including `|`, top-level type only (`dict`), verbose descriptions, and leaving them out completely).  Despite there being no implementation support for them in Python 2, the PEP 484 type annotations work particularly well for a function like `_parse_volumes`, where they quickly make apparent the transformation taking place.

### DIFF
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -155,7 +155,7 @@ class ContainerState(PClass):
     name = field(type=unicode, mandatory=True)
     image = field(type=DockerImage, mandatory=True)
     running = field(type=bool, mandatory=True)
-    volumes = field(initial=None)
+    volumes = field(initial=None, mandatory=True)
 
 
 class Node(PClass):
--- a/flocker/apiclient/_client.py
+++ b/flocker/apiclient/_client.py
@@ -111,6 +111,17 @@ class Lease(PClass):
     expires = field(type=(float, int, NoneType), mandatory=True)
 
 
+class MountedDataset(PClass):
+    """
+    A mounted dataset.
+
+    :attr UUID dataset_id: The UUID of the dataset.
+    :attr unicode mountpoint: The filesystem location of the dataset.
+    """
+    dataset_id = field(type=UUID, mandatory=True)
+    mountpoint = field(type=unicode, mandatory=True)
+
+
 class Container(PClass):
     """
     A container in the configuration.
@@ -119,10 +130,13 @@ class Container(PClass):
         will run.
     :attr unicode name: The unique name of the container.
     :attr DockerImage image: The Docker image the container will run.
+    :attr Sequence[MountedDataset] volumes: Flocker volumes mounted in
+        container.
     """
     node_uuid = field(type=UUID, mandatory=True)
     name = field(type=unicode, mandatory=True)
     image = field(type=DockerImage, mandatory=True)
+    volumes = field(initial=None)
 
 
 class ContainerState(PClass):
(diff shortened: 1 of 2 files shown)


flocker/apiclient/_client.py
```diff
@@ -760,15 +801,23 @@ def _parse_configuration_container(self, container_dict):
         :return: ``Container`` instance.
         """
         return Container(
-            node_uuid=UUID(hex=container_dict[u"node_uuid"], version=4),
+            node_uuid=UUID(hex=container_dict[u"node_uuid"]),
```
jongiddy: This change is not really related to this PR. But setting the version of a UUID only makes sense if we are minting a new UUID from some data. Overriding the version when creating a UUID instance from existing UUID data is asking for trouble.

flocker/apiclient/_client.py
```diff
@@ -111,34 +111,51 @@ class Lease(PClass):
...
     :attr DockerImage image: The name of the Docker image.
     :attr bool running: Whether the container is running.
+    :attr Optional[Sequence[MountedDataset]] volumes: Flocker volumes
+        mounted in container.
     """
     node_uuid = field(type=UUID, mandatory=True)
     name = field(type=unicode, mandatory=True)
     image = field(type=DockerImage, mandatory=True)
     running = field(type=bool, mandatory=True)
+    volumes = field(initial=None)
```
jml: Missing `mandatory=True` means that some instances of this might not have the `volumes` attribute set at all (i.e. `state.volumes` will raise an `AttributeError`). Is this deliberate?

flocker/apiclient/_client.py
```diff
@@ -751,6 +772,26 @@ def version(self):
             b"GET", b"/version", None, {OK}
         )
 
+    @staticmethod
+    def _parse_volumes(volume_list):
```
jml: :question: I'd make this a function, since there's no reason for inheritance.

flocker/apiclient/_client.py
jml: Thanks.

flocker/apiclient/_client.py
```diff
@@ -751,6 +772,26 @@ def version(self):
...
+        :return Optional[Sequence[MountedDataset]]: List of mounted datasets,
+            or None if no volumes.
+        """
+        if volume_list:
+            return [
+                MountedDataset(
+                    dataset_id=UUID(volume[u'dataset_id']),
+                    mountpoint=volume[u'mountpoint'],
+                ) for volume in volume_list
+            ]
```
jml: :question: Maybe make this an generator expression?

jml:
I'm happy w/ the PEP 484 things in docstrings.

Thanks!

flocker/apiclient/_client.py
jongiddy: @jml I guess I'm confused by the relationship between `initial` and `mandatory`. It seems to me that `initial` implies `mandatory`. Is it possible to create a PClass with a missing field if the field is defined with an `initial` value?

Even removing an attribute as an evolution seems to return it to the initial value:
```python
class X(PClass):
    a = field(initial=1)

x = X()
x = x.set(a=2)
x = x.remove('a')
assert x.a == 1
```

jml: OK, it looks like I was confused. Sorry.

jongiddy: No worries. PRecord works that way, and it does seem reasonable that removing an attribute should either completely remove the attribute or fail if the attribute is mandatory.  To the point that the existing behaviour of PClass might be a bug.  However, fixing the "bug" would be a major backwards-incompatible change.

jongiddy: I have filed https://github.com/tobgu/pyrsistent/issues/76

jongiddy: I've added the `mandatory` property, in case Pyrsistent (which is pre-1.0) changes in this regard.

flocker/apiclient/_client.py
jongiddy: Generator expressions are iterators, but the return value must be a restartable iterable.

PR #2389, opened 2015-12-24
[FLOC-3755] Fix benchmark no load scenario test

jongiddy:
Fixes a non-executed test, that also contained a bug. The test passed a callback result to the `stop` method as an extraneous parameter. This was not detected because the test method did not start with the `test` prefix.

Also, changed interface test to avoid use of `globals()`, following a comment on Slack. https://clusterhqteam.slack.com/archives/flocker-benchmarking/p1450723154000373


jml:
Great, thanks.

PR #2390, opened 2015-12-24
[FLOC-3769] Suppress log information in benchmark scenario tests

jongiddy:
Mostly, this is just continuing a pattern of suppressing the Eliot logs from the test output, without actually checking their contents.

However, the `test_scenario_stops_only_when_no_outstanding_requests` tests are slightly interesting. They manage to log an error to Eliot that is unflushed. Not really sure what this means, but following the suggestion at https://github.com/ClusterHQ/eliot/blob/9d61e6a93c07bbe20e777d51b028b432b3399a89/eliot/testing.py#L243-L245 fixes it.

Hopefully the cause of this problem is the other branch of the "either" clause in that comment.  Maybe @itamarst can shed further light?


benchmark/scenarios/test/test_write_request_load.py
```diff
@@ -472,7 +488,11 @@ def test_scenario_stops_only_when_no_outstanding_requests(self):
         c.advance(delay)
         self.successResultOf(d)
 
-    def test_scenario_timeouts_if_requests_not_completed(self):
+        # Suppress Eliot error for unflushed exceptions
+        logger.flushTracebacks(FakeNetworkError)
```
jml: I *think* this should be moved to after the `fail_requests = False` statement, since AFAICT, after then, everything ought to be working, and any logged exceptions after then will be genuine errors.

However, the whole thing smells a bit like a bug to me. Since we're creating the exceptions, we should be able to add errbacks that suppress the logging of tracebacks. Would welcome @itamarst 's opinion.

jml:
Thanks Jon!

I think you should probably wait for Itamar's thoughts, but I'll leave it up to your judgement.

Please do move the flushTracebacks call before landing.

PR #2394, opened 2015-12-30
[FLOC 3773] Fix broken doc links

davidoliverSP2:
Fixes 3773, which replaces 2 doc-dev links with the new master build from Jenkins.


jongiddy:
I tried to fix this problem by making the `doc-dev` URL map to the S3 hosted site, but it proves to be more difficult than I think it should be.

So, given this page is mainly for internal ClusterHQ use, the change of URLs is acceptable.

The new links work fine, so please merge.

PR #2395, opened 2015-12-30
[FLOC-3726] Increase the Jenkins timeouts for running the acceptance test suite.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3726

Give CentOS, slower probably because of bad Docker configuration, two hours to run the acceptance tests.
Give Ubuntu, which almost always finishes in less than 45 minutes, 90 minutes to be sure we don't hit the timeout.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1426,6 +1426,11 @@ job_type:
           }
       clean_repo: true
       archive_artifacts: *acceptance_tests_artifacts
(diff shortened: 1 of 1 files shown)


jml:
Looks good. Maybe make comments on the timeouts so that people have a clue as to why they're set the way they are?

PR #2396, opened 2015-12-30
[FLOC-3680] Increase the Jenkins timeouts for running the Cinder functional test suite

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3680

Give them twice as long to run on the premise that they pass most of the time now so
they should have a really, really good chance of passing all the time with twice as
long.


jml:
Looks good to me. Maybe add a comment on the reason for the timeout?

PR #2397, opened 2015-12-30
[FLOC-3756] Benchmark scenario refactoring 

mpgomez:
Refactoring of the benchmark load scenarios to move all the common code to a single place, and make it a bit more modular and structured.

Documentation (docstrings) may need to be reviewed and reworked a bit, and I may need some extra testing (or at least, I now could have some more tests), but I want to have some feedback about it before doing any more work on it. Thanks!


benchmark/_interfaces.py
```diff
@@ -87,3 +87,25 @@ def measure(f, *a, **kw):
         :return: Deferred firing when measurement has been taken, with the
             value of the measurement.
         """
+
+
```
mpgomez: Two new interfaces, the common stuff an scenario needs: a setup, and a request generator that will make a request.

benchmark/scenarios/rate_measurer.py
```diff
@@ -0,0 +1,70 @@
+# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
```
mpgomez: New class with all the common functionality that was duplicated in the read and write scenarios

benchmark/scenarios/read_request_load.py
```diff
@@ -2,105 +2,41 @@
 """
```
mpgomez: Common code taken out. The read scenario is now quite simple, as most of the code was common and it doesn't need a setup

benchmark/scenarios/request_load.py
```diff
@@ -0,0 +1,220 @@
+# Copyright 2015 ClusterHQ Inc.  See LICENSE file for details.
```
mpgomez: Class containing all the scenarios common code

benchmark/scenarios/test/test_rate_measurer.py
```diff
@@ -0,0 +1,147 @@
+from flocker.testtools import TestCase
```
mpgomez: The tests for the rate measurer were duplicated. Now they are run only once and have been moved to its own file

benchmark/scenarios/write_request_load.py
```diff
@@ -2,112 +2,22 @@
 """
```
mpgomez: Common code take out and scenario simplified.

benchmark/scenarios/request_load.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class NoNodesFound(Exception):
+    """
+    No nodes were provided by the control service.
+    """
+
+
+@implementer(IScenario)
+class RequestLoadScenario(object):
```
jongiddy: Make this a `PClass`, since there is no modification of the instance fields.  List the docstring `ivar`s in the same order as the PClass fields.

benchmark/scenarios/request_load.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    A scenario that places load on the cluster by performing read
+    requests at a specified rate.
```
jongiddy: Remove reference to "read requests" - just "requests" now.

benchmark/scenarios/request_load.py
```diff
@@ -0,0 +1,220 @@
...
+        self._maintained.errback(exception)
+
+    def check_rate(self):
+        """
+        Verify that the rate hasn't decreased and that the scenario is
+        not overloaded. A scenario is overloaded if there are too many
+        outstanding requests.
+
+        :raise: `RequestRateTooLow` if the rate has dropped.
+        :raise: `RequestOverload` if the scenario is overloaded.
```
jongiddy: Add the exception before the colon, with no quotes:
```
:raise RequestOverload: if the scenario...
```

benchmark/scenarios/read_request_load.py
```diff
@@ -2,105 +2,41 @@
...
-from collections import deque
-from itertools import repeat
 
 from zope.interface import implementer
-from eliot import start_action, write_failure, Message
-from eliot.twisted import DeferredContext
 
-from twisted.internet.defer import CancelledError, Deferred
-from twisted.internet.task import LoopingCall
+from twisted.internet.defer import Deferred
```
jongiddy: Deferred isn't used.

benchmark/scenarios/write_request_load.py
```diff
@@ -208,151 +77,90 @@ def _get_dataset_node(self, nodes):
...
-            failure.trap(CancelledError)
-            raise WRequestRateNotReached
 
-        # Loop until we reach the expected rate, or we timeout
-        waiting_for_target_rate = loop_until(self.reactor,
-                                             reached_target_rate,
-                                             repeat(1))
-        timeout(self.reactor, waiting_for_target_rate, self.timeout)
-        waiting_for_target_rate.addErrback(handle_timeout)
+@implementer(IScenario, IScenarioSetup)
```
jongiddy: Class does not implement `IScenarioSetup`

benchmark/scenarios/write_request_load.py
```diff
@@ -208,151 +77,90 @@ def _get_dataset_node(self, nodes):
...
+        the rate.
+    :ivar timeout: Maximum time in seconds to wait for the requested
+        rate to be reached.
+    """
 
-        waiting_for_target_rate.addCallback(monitor_scenario_status)
+    def __init__(self, reactor, cluster, request_rate=10,
+                 sample_size=DEFAULT_SAMPLE_SIZE, timeout=45):
+        self.setup = WriteRequest(reactor, cluster)
+        self.read_request = self.setup
```
jongiddy: There doesn't appear to be a good reason to set these as instance fields. :question:

jongiddy:
Regarding the comment "Not sure about handling errors and timeout in the same errback" in `write_request_load.py`, note that a timeout is a form of error. The errback uses `failure.trap`, which you can see as an `except CancelledError:` clause - it is a specific catcher for the named failure, and other failures pass through.

Essentially, this is the correct Twisted way to handle specific failures, while passing through non-handled failures.

Please keep the code as is, but remove the comment.

benchmark/scenarios/write_request_load.py
```diff
@@ -208,151 +77,90 @@ def _get_dataset_node(self, nodes):
...
-        :raise: `WNoNodesFound` if the given list of nodes was empty.
+        :raise: `NoNodesFound` if the given list of nodes was empty.
         """
         if not nodes:
-            raise WNoNodesFound()
+            raise NoNodesFound()
         return random.choice(nodes)
 
-    def _request_and_measure(self, count):
+    def _set_dataset_node(self, nodes):
```
jongiddy: Unused function.

benchmark/scenarios/read_request_load.py
```diff
@@ -2,105 +2,41 @@
...
+    def make_request(self):
+        return self.control_service.list_nodes()
 
-class RequestRateNotReached(Exception):
-    """
-    The request rate did not reach the target level.
-    """
 
+@implementer(IRequestGenerator)
+class ReadRequest(object):
```
jongiddy: Remove duplicated `ReadRequest` class.

benchmark/scenarios/read_request_load.py
```diff
@@ -117,153 +53,26 @@ class ReadRequestLoadScenario(object):
...
-        :raise: `RequestRateTooLow` if the rate has dropped.
-        :raise: `RequestOverload` if the scenario is overloaded.
-        """
-        rate = self.rate_measurer.rate()
-        if rate < self.request_rate:
-            self._fail(RequestRateTooLow(rate))
-
-        if self.rate_measurer.outstanding() > self.max_outstanding:
-            self._fail(RequestOverload())
+    def __init__(self, reactor, cluster, request_rate=10,
```
jongiddy: Note, you can replace the class `ReadRequestLoadScenario` with a function of the same name, that returns a `RequestLoadScenario`. This avoids having to implement the simple wrapper methods.

benchmark/scenarios/rate_measurer.py
jongiddy: Rename this file with an underscore prefix, as it should not be exported from this module.

benchmark/scenarios/read_request_load.py
jongiddy: Add an underscore prefix to the filename.

benchmark/scenarios/request_load.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        rate = self.rate_measurer.rate()
+        if rate < self.request_rate:
+            self._fail(RequestRateTooLow(rate))
+
+        if self.rate_measurer.outstanding() > self.max_outstanding:
+            self._fail(RequestOverload())
+
+    def start(self):
+        if self.setup is None:
```
jongiddy: Rather than using an `if` here, have the caller provide a no-op setup function when nothing needs to be done:
```
def setup():
    return succeed()
```

benchmark/_interfaces.py
```diff
@@ -87,3 +87,25 @@ def measure(f, *a, **kw):
...
+    A request generator that provides a method to generate
+    a request of certain type (like read or write)
+    """
+    def make_request():
+        """
+        Interface for request generator
+        """
+
+
+class IScenarioSetup(Interface):
```
jongiddy: Given each of these classes contains one method, I'd say either:
1. abandon the use of interfaces, and just pass the two functions separately.
2. combine these into a single interface with `request_setup` and `make_request` methods.  I note that your only non-trivial implementation combines both interfaces into a single implementing class.

jongiddy:
Good first attempt. I think the `RequestLoadScenario` class is good, but would like to see a tidier structure for the functions passed to it.  Several other comments to be addressed.

benchmark/scenarios/write_request_load.py
```diff
@@ -208,151 +77,90 @@ def _get_dataset_node(self, nodes):
...
-            raise WRequestRateNotReached
 
-        # Loop until we reach the expected rate, or we timeout
-        waiting_for_target_rate = loop_until(self.reactor,
-                                             reached_target_rate,
-                                             repeat(1))
-        timeout(self.reactor, waiting_for_target_rate, self.timeout)
-        waiting_for_target_rate.addErrback(handle_timeout)
+@implementer(IScenario, IScenarioSetup)
+class WriteRequestLoadScenario(object):
```
jongiddy: Same comment as for ReadRequest - make this a factory function returning a `RequestLoadScenario`.

benchmark/_interfaces.py
```diff
@@ -87,3 +87,20 @@ def measure(f, *a, **kw):
...
+    make requests of a certain type
+    """
+    def make_request():
+        """
+        Interface for request generator
+        """
+    def run_setup(self):
+        """
+        Interface for the scenario load setup
+        """
```
exarkun: I don't think I could implement this interface based on the documentation given here.  Can you elaborate on the lifecycle of this object in the class docstring and add information about what each of the methods is supposed to do and return to the method docstrings?

Also, `run_setup` has `self` as an argument and I suspect it shouldn't.

benchmark/scenarios/__init__.py
```diff
@@ -1,9 +1,15 @@
...
+)
+from ._rate_measurer import (
+    RateMeasurer, DEFAULT_SAMPLE_SIZE
+)
+from ._request_load import (
+    RequestRateTooLow, RequestRateNotReached,
+    RequestOverload, NoNodesFound, RequestLoadScenario
 )
+
+
```
exarkun: Pyflakes has a lot to say about this module.  The general form of the error is pre-existing but you may as well fix it in this branch, I think.

The idiomatic way to declare that these names are all being imported for re-export is to put them into an `__all__` definition.  See `flocker/common/__init__.py` for an example.

Also this file is missing a module(/package) docstring and copyright header.

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+
+
+@implementer(IScenario)
+class RequestLoadScenario(object):
+    """
+    A scenario that places load on the cluster by performing
+    requests at a specified rate.
+
+    :ivar reactor: Reactor to use.
```
exarkun: Where instance attributes are one-to-one with `__init__` parameters, it's prefered to document them on `__init__` as parameters.

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+    :ivar scenario_setup: instance of an implementation of the interface
+        `IRequestScenarioSetup`.
```
exarkun: Instead of "instance of an implementation of theinterface" say "provider of"

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+    :ivar request_rate: The target number of requests per second.
+    :ivar sample_size: The number of samples to collect when measuring
+        the rate.
+    :ivar timeout: Maximum time in seconds to wait for the requested
```
exarkun: New code should prefer `datetime.timedelta` for representing time intervals.

benchmark/scenarios/_request_load.py
exarkun: Also, use double backtick for references like this, eg \`\`IRequestScenarioSetup\`\`

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+
+        :raise RequestRateTooLow: if the rate has dropped.
+        :raise RequestOverload: if the scenario is overloaded.
+        """
+        rate = self.rate_measurer.rate()
+        if rate < self.request_rate:
+            self._fail(RequestRateTooLow(rate))
+
+        if self.rate_measurer.outstanding() > self.max_outstanding:
+            self._fail(RequestOverload())
```
exarkun: Both of the `_fail` calls in this method could run on a single call to `check_rate`.  This would double call  `self._maintained.errback` - which is an error and will raise `AlreadyCalledError`.

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def start(self):
+        """
+        Runs the setup and starts the scenario
```
exarkun: Since running the scenario uses instance state, it's invalid to call `start` twice (at least concurrently).  Might be worth stating or even explicitly checking so as to be able to present a coherent exception.

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+            reached.
+        """
+        self.loop.start(interval=1)
+
+        def reached_target_rate():
+            return self.rate_measurer.rate() >= self.request_rate
+
+        def handle_timeout(failure):
+            failure.trap(CancelledError)
+            raise RequestRateNotReached
```
exarkun: Raise instances of exceptions, not exception classes.  `raise RequestRateNotReached()`

exarkun: I'm not sure this errback is correct.  Any code that calls `cancel` on `waiting_for_target_rate` will cause `handle_timeout` to be called with a `CancelledError` failure - that includes the `timeout` call below as well as any other timeouts or any other code that decides to use cancellation.

I think the solution is to pass a custom exception to`timeout` and handle that exception in `handle_timeout` instead.

benchmark/scenarios/_request_load.py
```diff
@@ -0,0 +1,234 @@
...
+        ):
+            def handle_timeout(failure):
+                failure.trap(CancelledError)
+                msg = (
+                    "Force stopping the scenario. "
+                    "There are {num_requests} outstanding requests"
+                ).format(
+                    num_requests=outstanding_requests
+                )
+                Message.log(key='force_stop_request', value=msg)
```
exarkun: I think `message_type` is more idiomatic than `key`.

exarkun: Also... Textual messages are a bit contrary to the design principles of Eliot.

benchmark/scenarios/test/test_rate_measurer.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def receive_requests(self, rate_measurer, num_requests, num_samples):
+        """
+        Helper function that will receive the desired number of requests.
+
+        :param rate_measurer: The `RateMeasurer` we are testing.
+        :param num_requests: The number of request we want to receive.
+        :param num_samples: The number of samples to collect.
+        """
+        ignored = ""
```
exarkun: Bytes or unicode?  Please make this explicit.  Either use `u""`/`b""` or use the `unicode_literals` future import in this module (and fix any other strings).

benchmark/scenarios/test/test_rate_measurer.py
```diff
@@ -0,0 +1,147 @@
...
+        :param rate_measurer: The `RateMeasurer` we are testing.
+        :param num_requests: The number of request we want to make.
+        :param num_samples: The number of samples to collect.
+        """
+        self.send_requests(rate_measurer, num_requests, num_samples)
+        self.receive_requests(rate_measurer, num_requests, num_samples)
+
+    def test_rate_is_zero_when_no_samples(self):
+        """
+        When no samples have been collected, the rate should be 0.
```
exarkun: Write test method docstrings without "should" - eg "When no samples have been collected, the rate is 0."

benchmark/scenarios/test/test_rate_measurer.py
```diff
@@ -0,0 +1,147 @@
...
+        target_rate = 5
+        num_samples = r.sample_size - 1
+
+        self.increase_rate(r, target_rate, num_samples)
+
+        self.assertTrue(r.rate() < target_rate)
+
+    def test_rate_is_correct_when_enough_samples(self):
+        """
+        A RateMeasurer should correctly report the rate when enough
```
exarkun: Write test method docstrings without the word "correctly" - describe what it is that is considered correct.

exarkun:
Thanks.  A few more comments left inline.  Mostly stylistic and many of which I think are actually about code that's only moving, not being added or changed in by this PR.  Feel free to disregard those for the purposes of this PR but it would probably be nice to fix at least some of them as follow-up tasks.

Please merge once points are addressed and Jenkins is happy.

mpgomez:
Green build:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/benchmark-scenario-refactoring-FLOC-3756/

Merging pull request :)

PR #2398, opened 2016-01-01
[FLOC-3780] Switch to shell syntax and rely on the shell to do this interpolation

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3780

Can't say I understand how the old version worked or why it stopped working.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -687,7 +687,7 @@ common_cli:
     cat <<EOF_VAGRANTFILE >Vagrantfile
     Vagrant.configure(2) do |config|
        config.vm.box = 'clusterhq/jenkins-osx-yosemite'
-       config.vm.box_url = "s3://${ENV['S3_BUCKET']}/metadata.json"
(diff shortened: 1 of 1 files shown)


jml:
I likewise don't understand the change, but the code here seems clear, and this PR passes the relevant builder.

Please merge when you're satisfied with the tests.

exarkun:
Elsewhere, Tom described a likely explanation.  As part of the Jinja2 removal, a backslash was removed from the `$` in this expression.  With the backslash, bash handed the expression through to the Vagrantfile where *Ruby* interpreted it.  Without the backslash, bash tries to evaluate it and comes up with nothing.

Putting the backslash back would probably also fix things but I'm just as happy to see less logic on the Ruby side of things, so I'll let this version of the fix stand.

exarkun:
Failing jobs on Jenkins look like they're all Jenkins problems and none of them have to do with the modified code path - so, merging.

PR #2400, opened 2016-01-03
[FLOC-3585] Update doc-dev to latest docs from master

jongiddy:
Keep the docs at http://doc-dev.clusterhq.com/ in sync with the latest master branch. When a branch's docs are uploaded to S3, run a sync operation between `cluster-staging-docs/master` and `doc-dev.clusterhq.com`. 

The sync only copies files that have changed, so will normally only perform actions when the branch being updated is `master`.  We could use a shell conditional to restrict the sync to only when the updated branch is `master`, but I didn't want to push my Jenkins non-knowledge too far.

It appears that S3 syncs between buckets are sensitive to selecting the correct region, so code makes sure that we use the correct region.  It also sets the public read permissions, since we are setting the to-level of the bucket, so there is no parent to inherit permissions from.

I have run the command line manually, but I'm not sure how to test new Jenkins configuration properly.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -839,7 +839,7 @@ common_cli:
 
   sync_master_docs_to_doc_dev: &sync_master_docs_to_doc_dev |
     # ensure master docs at doc-dev are up-to-date
-    aws --region ${S3_DOCS_REGION} s3 sync --acl public-read \
+    [ "$TRIGGERED_BRANCH" != master ] || aws --region ${S3_DOCS_REGION} s3 sync --acl public-read \
       s3://clusterhq-staging-docs/master s3://doc-dev.clusterhq.com
 
   upload_vagrant_tutorial_basebox_to_s3: &upload_vagrant_tutorial_basebox_to_s3 |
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -831,10 +833,15 @@ common_cli:
...
     # uploads the new generated html to a folder on the S3 staging-docs bucket
     cd _build/html
-    aws --region ${S3_ENDPOINT} s3 sync . s3://clusterhq-staging-docs/$TRIGGERED_BRANCH
+    aws --region ${S3_DOCS_REGION} s3 sync . s3://clusterhq-staging-docs/$TRIGGERED_BRANCH
     echo "This branch is available at :"
     echo "http://clusterhq-staging-docs.s3.amazonaws.com/$TRIGGERED_BRANCH/index.html"
 
+  sync_master_docs_to_doc_dev: &sync_master_docs_to_doc_dev |
+    # ensure master docs at doc-dev are up-to-date
+    aws --region ${S3_DOCS_REGION} s3 sync --acl public-read \
```
zubron: My preference would be for this to only run on commits to master rather than having this in the job configuration for every branch. The `run_sphinx` job is parameterised with the parameter `TRIGGERED_BRANCH`. We could use this to make this step conditional like you previously mentioned.

zubron:
Thanks @jongiddy! I've checked the latest docs build and the permissions for the sync operation are correct. Please address the comment above and feel free to merge after a docs build that shows that non-master branches don't perform the sync operation.

PR #2401, opened 2016-01-04
[FLOC-3777] Migrate apiclient tests to new base cases

jml:
Part of the overall march of migration. 

Had to tweak one test slightly, because `Lease` object isn't ordered.

### DIFF
--- a/flocker/apiclient/test/test_client.py
+++ b/flocker/apiclient/test/test_client.py
@@ -19,7 +19,6 @@ from pyrsistent import pmap
 from eliot import ActionType
 from eliot.testing import capture_logging, assertHasAction, LoggedAction
 
-from twisted.trial.unittest import SynchronousTestCase
 from twisted.python.filepath import FilePath
 from twisted.internet.task import Clock
 from twisted.internet import reactor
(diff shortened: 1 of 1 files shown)


jml:
Have discovered at least one genuine failure due to ordering issues, and also noted a problem in the eliot log parser: https://clusterhq.atlassian.net/browse/FLOC-3786

flocker/apiclient/test/test_client.py
```diff
@@ -403,16 +405,17 @@ def test_list_leases(self):
                 ])
             d.addCallback(lambda _: self.client.release_lease(d2))
             d.addCallback(lambda _: self.client.list_leases())
+            d.addCallback(frozenset)
             d.addCallback(
-                self.assertItemsEqual,
-                [
+                self.assertEqual,
```
sarum90: :grey_question: Curious why not to just use assertItemsEqal. Implementation details like performance aside, it seems to express intent better (to me).

jml: Newer Python implementations of `assertItemsEqual` rely on ordering being implemented consistently. `PClass` subclasses don't do that.

sarum90: Never mind figured it out. I did a tiny repo and I found out that 1 field PClasses are orderable, whereas 2+ field PClasses are not orderable... which makes sense, but was more magic than I was expecting.

sarum90:
Looks good, given that you got a green run of the tests I'm going to merge this for you.

PR #2402, opened 2016-01-04
[FLOC-3778] Migrate flocker.ca tests to new base test cases

jml:
Almost all mechanical. None of the tests in `test_script` looked async, so I switched them to use the synchronous base case (`flocker.testtools.TestCase`)


dwgebler:
Thanks, LGTM when build passes.

PR #2403, opened 2016-01-04
[FLOC-3779] Migrate flocker.cli tests to new base test cases

jml:
Mostly mechanical. Changed a couple of test cases to be synchronous, since they weren't returning Deferreds.


flocker/cli/test/test_cli_script.py
```diff
@@ -12,7 +13,7 @@ class FlockerCLITests(FlockerScriptTestsMixin, TestCase):
     command_name = u'flocker'
 
 
-class CLIOptionsTests(StandardOptionsTestsMixin, SynchronousTestCase):
```
sarum90: :grey_question:  No action required, but this seems non-obvious so I wanted to be sure I got it right. Feel free to add a comment or clean up if you agree that it's cleaner some other way.

Maybe I'm crazy, but shouldn't the TestCase be subclassed at the point of implementation of the test? I guess that would require a class factory so that the bare mixin wasn't run by trial :cry:. It seemed odd to me that for this and the above test I had to look at the implementation of the Mixin in order to determine if it was okay to use TestCase rather than AsyncTestCase. This also seems to indicate that these "Mixins" are not really what I would call Mixins... they seem to actually be parametrized testcase implementations. I guess that's known badness and we have ideas of how to do it in the future using testtools. Probably good enough for now.

One compromise that might not be the worst idea in the world:

in the definition of the mixin:
```
class StandardOptionsTestsMixin(object):
   BaseTestCase = testtools.TestCase
    ...
```

Then here:
```
class CLIOptionsTests(StandardOptionsTestsMixin, StandardOptionsTestsMixin.BaseTestCase):
  ...
```

sarum90:
Thanks for continuing the migration -- Looks good. Although you got a funky failure in your presubmits. Looks totally unrelated but probably worth tracking down in JIRA to double check it's covered.

One comment for you to consider, but it's probably overkill -- feel free to merge as is.

flocker/cli/test/test_cli_script.py
jml: Good catch!

Probably the "right" way to do it is the same way we do our interface tests for other things (e.g. `make_iblockdeviceapi_tests`). I don't quite have the energy for that right now, but I've filed https://clusterhq.atlassian.net/browse/FLOC-3794 and put it on the board. If no one picks it up before tomorrow's British morning, I'll have a go. In the mean time, I'll add a comment.

sarum90: Yeah I'm sorta 50/50 on if it's worth doing. Definitely an improvement to the code, but arguably not the most effective use of engineer time? On the other hand if it appears in the review column I'll gladly review it.

jml:
All failures are spurious & known. Merging.

PR #2405, opened 2016-01-04
[FLOC-3686] EBS check attachment state

dwgebler:
Fixes FLOC-3686. Not sure if we need a functional test for this? it should be possible to do one by calling out to a `mount --rbind` command on the node after creating and mounting a volume, then deleting the volume and seeing the `DetachFailed` exception in the log.

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -45,7 +45,7 @@ from eliot import Message, register_exception_extractor
 from .blockdevice import (
     IBlockDeviceAPI, IProfiledBlockDeviceAPI, BlockDeviceVolume, UnknownVolume,
     AlreadyAttachedVolume, UnattachedVolume, UnknownInstanceID,
-    MandatoryProfiles, ICloudAPI,
+    MandatoryProfiles
 )
 
 from ..script import StorageInitializationError
@@ -75,6 +75,8 @@ IOPS_MIN_SIZE = 4
 # for error details:
 NOT_FOUND = u'InvalidVolume.NotFound'
 INVALID_PARAMETER_VALUE = u'InvalidParameterValue'
+
+VOLUME_ATTACHMENT_BUSY = u'busy'
 
 
 # Register Eliot field extractor for ClientError responses.
@@ -276,6 +278,14 @@ class AttachFailed(Exception):
     """
 
 
+class DetachFailed(Exception):
+    """
+    A volume could not be detached, due to being in busy state.
+    """
+    def __init__(self, volume):
+        Exception.__init__(volume.id, volume.attachments)
+
+
 class InvalidRegionError(Exception):
     """
     The supplied region is not a valid AWS endpoint.
@@ -389,6 +399,27 @@ def _enable_boto_logging():
     logger.addHandler(EliotLogHandler())
 
 _enable_boto_logging()
+
+
+class AttachUnexpectedInstance(Exception):
+    """
+    An attempt was made to attach a volume to an instance other than the
+    one this ``IBlockDeviceAPI`` implementation is running on.
+    """
+    def __init__(self, blockdevice_id, instance_id, local_instance_id):
+        """
+        :param unicode blockdevice_id: The identity of the block device we
+            are attempting to attach.
+        :param unicode instance_id: The identity of the instance we are
+            attempting to attach the block device to.
+        :param unicode local_instance_id: The identity of the instance this
+            API is running on.
+        """
+        Exception.__init__(
+            self, blockdevice_id, instance_id, local_instance_id)
+        self.blockdevice_id = blockdevice_id
+        self.instance_id = instance_id
+        self.local_instance_id = local_instance_id
 
 
 @with_cmp(["requested", "discovered"])
@@ -524,9 +555,10 @@ def _get_volume_tag(volume, name):
 
     :return: A ``str`` representing the value of the tag.
     """
-    for tag in volume.tags:
-        if tag['Key'] == name:
-            return tag['Value']
+    if volume.tags:
+        for tag in volume.tags:
+            if tag['Key'] == name:
+                return tag['Value']
     raise TagNotFound(volume.id, name, volume.tags)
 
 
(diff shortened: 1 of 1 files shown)


flocker/node/agents/ebs.py
```diff
@@ -1309,13 +1323,6 @@ def get_device_path(self, blockdevice_id):
 
         return _expected_device(ebs_volume.attachments[0]['Device'])
 
-    # ICloudAPI:
-    @boto3_log
-    def list_live_nodes(self):
-        instances = self.connection.instances.filter(
-            Filters=[{'Name': 'instance-state-name', 'Values': ['running']}])
-        return list(instance.id for instance in instances)
-
```
jml: Are the `ICloudAPI` removals part of the fix? From here, they look like a bad merge.

flocker/node/agents/ebs.py
```diff
@@ -1245,6 +1254,11 @@ def detach_volume(self, blockdevice_id):
         if ebs_volume.state != VolumeStates.IN_USE.value:
             raise UnattachedVolume(blockdevice_id)
 
+        if ebs_volume.attachments:
+            attachment_state = ebs_volume.attachments[0]['State']
```
jml: Why are we only looking at the first attachment? What happens if the first attachment isn't busy but one of the others are?

flocker/node/agents/ebs.py
```diff
@@ -276,6 +278,14 @@ class AttachFailed(Exception):
     """
 
 
+class DetachFailed(Exception):
+    """
+    A volume could not be detached, due to being in busy state.
+    """
```
jml: Given the description, please rename this exception to `VolumeBusy` to reflect the cause, and to be consistent with the other exceptions raised by detach.

flocker/node/agents/ebs.py
```diff
@@ -1245,6 +1254,11 @@ def detach_volume(self, blockdevice_id):
         if ebs_volume.state != VolumeStates.IN_USE.value:
             raise UnattachedVolume(blockdevice_id)
 
+        if ebs_volume.attachments:
+            attachment_state = ebs_volume.attachments[0]['State']
+            if attachment_state == VOLUME_ATTACHMENT_BUSY:
+                raise DetachFailed(ebs_volume)
```
jml: Please document the new exception in the docstring.

jml:
Thanks for looking into this. I think we should add a functional test for this, both out of general principle and because I'm not 100% sure of the problem we're solving. 

The bug description says we should raise an error with a list of mount references that are preventing the volume from being detached—this PR doesn't yet do that. The bug description doesn't described the observed behaviour of the current system under these conditions (instead it has only the desired behaviour of the system). I think it would be very interesting to add a functional test and run it _without_ the changes in this PR to be clear as to exactly what we're fixing.

flocker/node/agents/ebs.py
jml: This overall approach is "look before you leap", which is generally speaking less good than trying-and-failing. In this particular case, it's possible that although the volume _was_ busy on line 1258, it's no longer busy by line 1262, or that it's busy, but the symptoms are different. A more reliable way would be to catch whatever error is raised `_detach_ebs_volume`.

jml:
Please address comments & resubmit for review.

flocker/node/agents/ebs.py
dwgebler: Grr, yeah something's gone wrong here. This isn't supposed to be part of the PR.

flocker/node/agents/ebs.py
dwgebler: For Flocker created volumes, there will only ever be at most one attachment entry, unless something's gone horribly wrong somewhere. That we have to check `volume.attachments[0]` is a quirk of using Boto3; we didn't have a list prior to that upgrade.

dwgebler:
> The bug description says we should raise an error with a list of mount references that are preventing the volume from being detached—this PR doesn't yet do that.

I don't know how we can do that - at least not purely in the EBS driver - how can we obtain a list of such mount references? In the EBS driver, we don't even know if existing mounts are the reason the volume attachment is in a busy state, though that *might* be a reliable assumption.

dwgebler:
> I think it would be very interesting to add a functional test and run it without the changes in this PR to be clear as to exactly what we're fixing.

A functional test is a good idea I think, yes. Just for clarity - sorry this wasn't on the ticket - the buggy behaviour is that the volume doesn't detach and a `TimeoutException` will eventually be raised from `flocker.node.agents.ebs` and the dataset agent gets stuck in a loop with:

```
Jan 04 15:25:23 ip-172-31-10-30 flocker-dataset-agent[7512]: {"timestamp": 1451921123.984151, "task_uuid": "a4d5117b-9484-4cc5-b1d3-2481bf0b1104", "message_type": "flocker:node:agents:blockdevice:aws:boto_logs", "message": "Calling ec2:describe_volumes with {u'VolumeIds': [u'vol-b4674752']}", "task_level": [2, 4, 4, 2, 237]}
```

where `vol-b4674752` is whatever volume ID is stuck in the `busy` attachment state.

flocker/node/agents/ebs.py
dwgebler: The EC2 API won't return an error if attachment state is busy when you attempt to detach, rather it will report detaching state:

```
{'ResponseMetadata': {'HTTPStatusCode': 200, 'RequestId': 'a52cd151-b8a9-4ee1-b91f-4d355c65acb1'}, u'AttachTime': datetime.datetime(2016, 1, 5, 10, 2, 7, tzinfo=tzutc()), u'InstanceId': 'i-fcc27026', u'VolumeId': 'vol-986c4e7e', u'State': 'detaching', u'Device': '/dev/sdf'}
```

However, the detach will not happen and you will continue to see actual attachment state of e.g.

```
[{u'AttachTime': datetime.datetime(2016, 1, 5, 10, 2, 7, tzinfo=tzutc()), u'InstanceId': 'i-fcc27026', u'VolumeId': 'vol-986c4e7e', u'State': 'busy', u'DeleteOnTermination': False, u'Device': '/dev/sdf'}]
```

dwgebler:
Closing this PR because there was somehow a bad merge that's messed up the branch diff against master. I have a separate, clean branch for this issue now and will open a new PR against that when ready.

PR #2408, opened 2016-01-04
[FLOC-3790] Move flocker.provision tests to new base cases

jml:
Again, straightforward migration. No tests changed synchronicity. I updated some tests that were importing the reactor to use the one provided by `AsyncTestCase` (actually just the global reactor under the hood, so no real difference).


exarkun:
Great!  And CI agrees.  Please merge.

PR #2409, opened 2016-01-04
[FLOC-3795] Use symlinks for the shared paths.

sarum90:
Note that this is not up for merge with master (there needs to be better upgrade testing before it is merged). I'd still like to have the code reviewed before I land it in the branch that I am using to stage and test these changes before looking to pull them all to master.

Changes the dataset agent to create mountpoints in /var/flocker/mounts, and then atomically create a symlink in /flocker/v2 once they are ready for use. This enables atomic creation and deletion of the paths that flocker hands out to its client. This aids in protecting those agents from being able to successfully write to those paths, but have the writes not persist to the backing volume.

Adds 2 new IStateChanges that facilitate the symlinking and the unlink of the symlink.

Combined with the bind mounting done in a previous PR (if shadow mounting is set up), this makes it so that docker cannot even start a container with a volume in /flocker/v2 unless flocker has set up a dataset there.


itamarst:
Thanks for all the time you've been putting in to this. As a first pass, could you explain why this doesn't match original design of bind mounts? In particular does this still solve the problem of Docker auto-creating paths in `/flocker` if a dataset is unmounted?

flocker/node/agents/blockdevice.py
```diff
@@ -74,10 +74,17 @@ class DatasetStates(Names):
     ATTACHED = NamedConstant()
     # Mounted on this node
     MOUNTED = NamedConstant()
+    # Symlinked to the mountpoint created, ready to be shared
+    SYMLINKED = NamedConstant()
     # Deleted from the driver
     DELETED = NamedConstant()
 
 
+# The DatasetState that corresponds to a dataset that is manifest and ready for
```
itamarst: Maybe just call it "MANIFEST" to begin with?

itamarst:
I'm of two minds... on the one hand "yay explicit state machines", on the other hand that's a lot of code for a very minor change. Still thinking...

itamarst:
An alternative I can imagine is:

1. If the symlink is deleted but the block device is mounted, the state is `ATTACHED` not `MOUNTED`.
2. `MountBlockDevice` mounts if necessary, then creates symlink.
3. `UnmountBlockDevice` deletes the symlink if necessary, then unmounts.
4. Since this mean mounts left dangling if Flocker crashes half way through unmount, on startup mounts without symlinsk are unmounted.

I'm not sure I like that though.

flocker/node/agents/blockdevice.py
```diff
@@ -543,6 +569,79 @@ def _valid_size(size):
 
 @implementer(IStateChange)
 @provider(IDatasetStateChangeFactory)
+class CreateMountSymlink(PClass):
```
itamarst: This is starting to feel very Java-y :crying_cat_face: Not worth fixing here though.

itamarst: (That is, huge amounts of repetitive boiler plate).

itamarst:
I just skimmed it, so definitely will need a code review. Adding an additional state does seem like the right thing to do, so proceed unless you have a clever idea.

flocker/node/agents/blockdevice.py
sarum90: I realized in making this PR that prior to this change "MANIFEST" meant "MOUNTED" and after this change "MANIFEST" meant "SYMLINKED".  But I had to go through the code and manually evaluate every use of "MOUNTED" to determine if it meant "The dataset is ready for use" or "the dataset has been mounted". It was sort of painful, and separating the uses makes it significantly less for future maintainers.

sarum90:
Bind mounts also still in progress (https://github.com/ClusterHQ/flocker/pull/2446) this just removes 1 more window where things could go wrong.

sarum90:
Latest tests all passed except for sphinx: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/use-symlinks-FLOC-3795/job/__main_multijob/7/

Sphinx passed on rerun: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/use-symlinks-FLOC-3795/job/run_sphinx/8/

flocker/node/agents/blockdevice.py
```diff
@@ -551,6 +577,83 @@ def _valid_size(size):
...
+        return CREATE_MOUNT_SYMLINK(_logger, dataset_id=self.dataset_id)
+
+    def run(self, deployer):
+        """
+        Create a symlink from the mounted blockdevice to a path that will given
+        to flocker clients. This layer of indirection is added so that we can
+        atomically create a mounted directory path.
+        """
+        parent = self.link_path.parent()
+        if not parent.exists():
```
itamarst: That seems problematic. There's a bunch of  permissios stuff you probably want to get right, for exampe. It rather seems like it should  just assume it's bee pre-created.

flocker/node/agents/blockdevice.py
```diff
@@ -1533,8 +1644,10 @@ class BlockDeviceDeployer(PClass):
     block_device_api = field(mandatory=True)
     _profiled_blockdevice_api = field(mandatory=True, initial=None)
     _async_block_device_api = field(mandatory=True, initial=None)
-    mountroot = field(type=FilePath, initial=FilePath(b"/flocker"))
-    shared_root = field(type=FilePath, initial=FilePath(b"/flocker"))
+    mountroot = field(type=FilePath,
+                      initial=FilePath(b"/var/flocker/mounts"))
+    link_root = field(type=FilePath, initial=FilePath(b"/flocker/v2"))
```
itamarst: I asume these will no longer be the same in a future iteration?

flocker/node/agents/blockdevice.py
```diff
@@ -1619,6 +1732,15 @@ def is_existing_block_device(dataset_id, path):
                     # in the filesystem yet.
                     pass
 
+        symlinks = {}
+        if self.link_root.exists() and self.link_root.isdir():
+            symlinks = {
+                link: link.realpath()
+                for link in iter(self.link_root.child(f)
```
itamarst: You can do `for link in self.link_root.children()`.

itamarst:
Out of brain cells. Will continue later.

itamarst:
OK, address and merge when happy.

flocker/node/agents/blockdevice.py
sarum90: At this stage it's needed to make the acceptance tests pass.

For now moved to initiation code, and added a note to add validation code there once it is removed (code that verifies everything is set up correctly.)

flocker/node/agents/blockdevice.py
sarum90: Correct. I want there to be code that verifies that everything is set up correctly before I change these to be different values.

Even in this PR there are tests that set these to different values.

flocker/node/agents/blockdevice.py
sarum90: That's much nicer.

sarum90:
With passing tests, I'm going to merge.

PR #2410, opened 2016-01-05
[FLOC-3794] replace FlockerScriptTestsMixin should follow the make_foo_tests pattern

justnoise:
As requested in the Jira ticket, replaces inheritence based classes with a class generator.

### DIFF
--- a/flocker/cli/test/test_deploy_script.py
+++ b/flocker/cli/test/test_deploy_script.py
@@ -6,19 +6,18 @@ Unit tests for the implementation of ``flocker-deploy``.
 
 from twisted.python.filepath import FilePath
 from twisted.python.usage import UsageError
-from twisted.trial.unittest import TestCase, SynchronousTestCase
+from twisted.trial.unittest import SynchronousTestCase
 
 from ...testtools import (
-    FlockerScriptTestsMixin, StandardOptionsTestsMixin)
+    make_flocker_script_test, StandardOptionsTestsMixin)
 from ..script import DeployScript, DeployOptions
 from ...control.httpapi import REST_API_PORT
 
 
-class FlockerDeployTests(FlockerScriptTestsMixin, TestCase):
+class FlockerDeployTests(
+        make_flocker_script_test(DeployScript, DeployOptions, u'flocker-deploy')
+):
     """Tests for ``flocker-deploy``."""
-    script = DeployScript
-    options = DeployOptions
-    command_name = u'flocker-deploy'
 
 
 CONTROL_HOST = u"192.168.1.1"
--- a/flocker/cli/test/test_deploy_script.py
+++ b/flocker/cli/test/test_deploy_script.py
@@ -14,9 +14,9 @@ from ..script import DeployScript, DeployOptions
 from ...control.httpapi import REST_API_PORT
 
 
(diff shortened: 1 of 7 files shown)


flocker/testtools/__init__.py
```diff
@@ -281,171 +281,180 @@ def __init__(self, argv=None):
...
 
-class StandardOptionsTestsMixin(object):
-    """Tests for classes decorated with ``flocker_standard_options``.
 
-    Tests for the standard options that should be available on every flocker
-    command.
+def make_standard_options_test(options):
+    """
+    Return a ``FlockerScriptTestCase`` which tests that the script
+    class provides ICommandLineScript
```
james-w: I'm not sure which "script class" this is referring to, it's not one of the
arguments to this function.

flocker/testtools/__init__.py
```diff
@@ -281,171 +281,180 @@ def __init__(self, argv=None):
...
-    options = None
 
-    def test_sys_module_default(self):
+    :returns: a ``TestCase``
+    """
+    class StandardOptionsTestCase(TestCase):
         """
-        ``flocker_standard_options`` adds a ``_sys_module`` attribute which is
-        ``sys`` by default.
+        Test for classes that implement ICommandLineScript
```
james-w: I think this should perhaps be the docstring from the original Mixin?

i.e.

"Tests for the standard options that should be available on every flocker command."

flocker/cli/test/test_deploy_script.py
```diff
@@ -6,33 +6,30 @@
...
-    script = DeployScript
-    options = DeployOptions
-    command_name = u'flocker-deploy'
 
 
 CONTROL_HOST = u"192.168.1.1"
 
 
-class DeployOptionsTests(StandardOptionsTestsMixin, SynchronousTestCase):
+class DeployOptionsTests(make_standard_options_test(DeployOptions)):
```
james-w: :question: The factory returns a class for the standard interface tests. Perhaps the extra tests for the
behaviour of these specific options would be good as a separate class?

james-w:
Hi,

Thanks for the branch, it looks great and is much cleaner.

There's a couple of docstrings that I think could be improved, and one
comment on the structure that you may want to use or to ignore.

I've asked jml to take a look at the branch as well as this is my first review.

Thanks,

James

flocker/testtools/__init__.py
jml: If they also implement `ICommandLineScript`, then the docstring should also mention that (in addition to the original docstring).

Also, fwiw, use double backquotes to mark something as code in docstrings, e.g.

    ``ICommandLineScript``

jml:
Thanks Brendan! It's great to see this code getting cleaned up.

There are merge conflicts (probably my fault, sorry). Please resolve them and make sure all tests pass before landing.

Thanks for the review @james-w, only a minor tweak suggested.

justnoise:
Thanks for your comments James, I spaced a bit on the docstrings.  I'll get into resolving those merge conflicts.

sarum90:
Updates look good. Just wait for greens on Jenkins and we can go ahead and merge.

PR #2411, opened 2016-01-05
[FLOC-3735] Correctly skip --journald test when libsystemd is available.

james-w:
Eliot uses the presence of libsystemd to decided whether journald
is available, so have the test use the same check for skipping,
rather than assume that journalctl will be present if the user
can write to journald.

https://clusterhq.atlassian.net/browse/FLOC-3735

### DIFF
--- a/flocker/common/functional/test_script.py
+++ b/flocker/common/functional/test_script.py
@@ -8,7 +8,7 @@ import os
 import sys
 from json import loads
 from signal import SIGINT
-from unittest import skipUnless
+from unittest import skipUnless, skipIf
 from subprocess import check_output, CalledProcessError, STDOUT, Popen
 
 from bitmath import MiB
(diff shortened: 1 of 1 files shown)


flocker/common/functional/test_script.py
```diff
@@ -18,6 +18,12 @@
 from eliot import Message
 from eliot.testing import assertContainsFields
 
+try:
+    from eliot.journald import JournaldDestination
+except ImportError as e:
+    # This platform doesn't have journald.
+    JournaldDestination = None
+
```
jml: :question: Since `JournaldDestination` is only used to determine whether to skip, perhaps it would be better to save a boolean. I would suggest `journald_available`, but that will be easily confused with `_journald_available()`.

Perhaps the best thing to do is to _also_ rename `_journald_available` so that the distinction between "can query journald" and "can log to journald" is more clear.

jml:
Thanks for fixing this! Please fix the lint and at your discretion the optional suggestion, and merge when the tests pass.

PR #2412, opened 2016-01-05
[FLOC-3798] Update Docker container name constraints.

jongiddy:
Docker accepts container names according to the pattern at https://github.com/docker/docker/blob/09ad1bbc5897e34c6e6d5612b0e1ba709ed28a7d/utils/names.go

`flocker-deploy` does not perform any validation of the name, but Flocker's HTTP API constrains JSON requests and responses to a more restricted pattern. This can cause status 500 errors if containers are created using `flocker-deploy` and then interrogated using the HTTP API. Update the HTTP API to match the Docker pattern.

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -177,7 +177,7 @@ definitions:
     title: "Container name"
     description: "The name of the container."
     type: string
-    pattern: "^[a-zA-Z0-9_-]+$"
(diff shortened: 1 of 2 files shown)


jml:
Sorry review took so long—multi-tasking is the mind-killer.

jml:
LGTM!

PR #2419, opened 2016-01-05
[FLOC-3685] Expand blockdevice manager for upcoming work protecting our dataset mounts.

sarum90:
To do some of the fancy bind mounts, remounts, and tmpfs mounts that are required for FLOC-3254, we'll have to expand the interface with the kernel.

Putting this behind an interface because we've noticed that it behaves slightly different on CentOS and Ubuntu, so I'll probably be writing verified fakes for each of those, and I need some interface in order to be able to inject the different verified fakes.

### DIFF
--- a/flocker/node/agents/blockdevice_manager.py
+++ b/flocker/node/agents/blockdevice_manager.py
@@ -74,6 +74,19 @@ class RemountError(Exception):
         return self.__repr__()
 
 
+@attributes(["mountpoint", "source_message"])
+class MakeTmpfsMountError(Exception):
+    """
+    Raised from errors while making a tmpfs mount.
+
+    :ivar FilePath mountpoint: The mountpoint for the new tmpfs mount.
+    :ivar unicode source_message: The error message describing the error.
+    """
+
+    def __str__(self):
+        return self.__repr__()
+
+
 @attributes(["blockdevice", "source_message"])
 class MakeFilesystemError(Exception):
     """Raised from errors while making a filesystem on a blockdevice.
@@ -160,7 +173,10 @@ class IBlockDeviceManager(Interface):
 
     def get_mounts():
         """
-        Returns all known mounts on the system.
+        Returns all known disk device mounts on the system.
+
+        This only includes mounted block devices and not tmpfs mounts or bind
+        mounts.
 
         :returns: An iterable of ``MountInfo``s of all known mounts.
         """
@@ -186,8 +202,19 @@ class IBlockDeviceManager(Interface):
             mountpoint with.
 
         :raises: ``RemountError`` on any failure from the system. This includes
-            user kill signals, so this may even be raised on successful bind
-            mounts.
+            user kill signals, so this may even be raised on successful
+            remounts.
+        """
+
+    def make_tmpfs_mount(mountpoint):
+        """
+        Creates a tmpfs mount at the given mountpoint.
+
+        :param FilePath mountpoint: The target path for the mount.
+
+        :raises: ``MakeTmpfsMountError`` on any failure from the system. This
+            includes user kill signals, so this may even be raised on
+            successful mounts.
         """
 
 
@@ -276,3 +303,12 @@ class BlockDeviceManager(PClass):
             raise RemountError(mountpoint=mountpoint,
                                permissions=permissions,
                                source_message="\n".join([str(e), e.output]))
+
+    def make_tmpfs_mount(self, mountpoint):
+        try:
+            check_output([b"mount", "-t", "tmpfs", "tmpfs", mountpoint.path],
+                         stderr=STDOUT)
+        except CalledProcessError as e:
+            raise MakeTmpfsMountError(mountpoint=mountpoint,
+                                      source_message="\n".join(
+                                          [str(e), e.output]))
--- a/flocker/node/agents/blockdevice_manager.py
+++ b/flocker/node/agents/blockdevice_manager.py
@@ -218,6 +218,38 @@ class IBlockDeviceManager(Interface):
         """
 
 
+class _CommandResult(PClass):
+    """
+    Helper object to represent the result of a command.
+
+    :ivar bool succeeded: True if the command was successful otherwise false.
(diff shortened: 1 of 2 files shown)


flocker/node/agents/blockdevice_manager.py
```diff
@@ -190,3 +283,32 @@ def get_mounts(self):
...
+
+    def bind_mount(self, source_path, mountpoint):
+        try:
+            check_output([b"mount", "--bind", source_path.path,
+                          mountpoint.path],
+                         stderr=STDOUT)
+        except CalledProcessError as e:
+            raise BindMountError(source_path=source_path,
+                                 mountpoint=mountpoint,
+                                 source_message="\n".join([str(e), e.output]))
```
james-w: :question: Would it make sense to have some sort of prettify helper for this repeated operation?

james-w:
Hi,

Thanks for the branch. This looks good to me. It's all very clear to read.

I've left one suggestion for a small helper inline.

I'm also interested in why the tests check the behaviour of the different
types of mounts, rather than e.g. checking the kernel thinks the
mount exists by checking the output of mount. That's just curiosity
though.

I'll ask jml for a mentor review in the morning, but if you find someone
else beforehand then that's fine too.

Thanks,

James

james-w:
Oh, and jenkins found a bit of lint:

flocker/node/agents/test/test_blockdevice_manager.py:9:1: F401 'check_output' imported but unused

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -157,3 +167,102 @@ def test_formatted_bad_type(self):
...
+        src_directory = self._get_directory_for_mount()
+        target_directory = self._get_directory_for_mount()
+        self.manager_under_test.bind_mount(src_directory, target_directory)
+        for create, view in [(target_directory, src_directory),
+                             (src_directory, target_directory)]:
+            filename = str(uuid4())
+            new_file = create.child(filename)
+            new_file.touch()
+            self.expectThat(view.child(filename).exists(), Equals(True),
+                            'Created file not visible through bind mount.')
```
jml: :question: You might want to use the `FileExists` matcher (see [the docs](http://testtools.readthedocs.org/en/latest/for-test-authors.html#file-and-path-related-matchers)).

e.g.

```python
self.expectThat(view.child(filename).path, FileExists(), 'Created file not ...')
```

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -157,3 +167,102 @@ def test_formatted_bad_type(self):
...
+
+        with ExpectedException(OSError):
+            second_file.touch()
+        self.expectThat(second_file.exists(), Equals(False))
+
+        self.manager_under_test.remount(mountpoint, Permissions.READ_WRITE)
+        self.expectThat(first_file.exists(), Equals(True))
+        self.expectThat(second_file.exists(), Equals(False))
+        third_file.touch()
+        self.expectThat(third_file.exists(), Equals(True))
```
jml: Likewise `FileExists` here.

jml: :question: Maybe at some point we should start using a library like [treeshape](https://github.com/jml/treeshape)

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -157,3 +167,102 @@ def test_formatted_bad_type(self):
...
+        to. Once the mount is unmounted all files should be gone.
+        """
+        mountpoint = self._get_directory_for_mount()
+        test_filename = unicode(uuid4())
+
+        # Twisted caches FilePath.exists, so if run path.exists() before and
+        # after you unmount a mountpoint, the FilePath will lie to you and
+        # claim that it still exists after the tmpfs mount was unmounted.
+        # Instead, create a new FilePath object every time to avoid bugs in the
+        # state of FilePath.
```
jml: I _think_ calling the `restat` method (see [docs](http://twistedmatrix.com/documents/current/api/twisted.python.filepath.FilePath.html#restat)) will do what you want here.

flocker/node/agents/blockdevice_manager.py
sarum90: Added - Thanks.

flocker/node/agents/test/test_blockdevice_manager.py
sarum90: Oh I tried that, you get an OSError because the file doesn't exist :facepalm:

flocker/node/agents/test/test_blockdevice_manager.py
sarum90: I thought about that, couldn't decide if ".exists(), Equals(True)" or ".path, FileExists()" was better.

I'll switch to your way though because twisted's ".exists()" is subtly broken, and I can remove a massive comment if I don't use it :)

jml: Also, you get a better error message if you use `FileExists`.

flocker/node/agents/test/test_blockdevice_manager.py
sarum90: Interesting, I don't quite see what treeshape gets us here (just simpler than the manual .touch() commands?). Although there is definitely something we could do in this space.... not sure what it is.

sarum90:
Thanks for the reviews! Just pushed updates, I'll wait for tests to finish.

Re: 
> checking the kernel thinks the mount exists by checking the output of mount

I could have done that - I guess I'm always a little scared about trying to parse output from things like `mount` and assuming it won't change over time. Although, thinking about that now the output of mount is probably stable enough to trust.

On top of this layer I also will be adding tests that actually need to verify the behavior of RO/RW/bind mounts that we know to be different on different kernel versions, but I guess I could have waited til the next layer for that sort of test.

sarum90:
Got a bunch of Jenkins NullPtrExceptions on the first run of the tests. Retrying seems to have made those go away. Merging.

PR #2420, opened 2016-01-06
[FLOC-3802] Move volume tests to new base cases

jml:
Also add successResultOf, failureResultOf, and assertNoResult to
AsyncTestCase so that the volume tests can run.


exarkun:
Thanks.  This all looks right to me.  Bit unfortunate about having to add `successResultOf` et al to the mixin ... but seems like the right move for the time being (if it's really a problem, it's pre-existing and we can fix it separately, and maybe it's not really a problem).

Merge when builds are green.

PR #2421, opened 2016-01-06
[FLOC-3806] More reliable way to find always-present process name

jongiddy:
The current code for testing the cputime benchmark requires knowledge of an always present process name, and assumes that Ubuntu uses `init` and CentOS uses `systemd`. However, Ubuntu 15.10 uses `systemd`.  Rather than update the current distribution detection, make it more robust by getting the name of process 1, and using that.

This has been tested on Ubuntu 14.04, 15.10, and OS X (where it skips the tests that use `_standard_process`).


james-w:
This looks good to me. I'll ask jml for a mentor review.

Thanks,

James

jml:
@jongiddy What a good idea! I wish I'd thought of it.

@james-w Review all good.

PR #2423, opened 2016-01-06
[FLOC-3294] Benchmark container creation

jongiddy:
Add a benchmark operation to create a stateful container and wait for it to start.

This performs the following actions:
1. create a container and then delete it, to ensure the image will not need to be pulled later, during the measured part of the benchmark;
2. create a dataset;
3. create a container attached to the dataset;
4. wait for the container to be running;
5. delete the container and dataset.

Steps 3 and 4 are in the `run` method of the probe, and hence are the operations that are measured by the benchmark.

If you have a Flocker cluster handy, and the acceptance test environment variables set, change to the `benchmark` directory and run:
```
./benchmark --operation create-container
```
which creates output like:
```
{
  "scenario": {
    "type": "no-load", 
    "name": "default"
  }, 
  "timestamp": "2016-01-06T12:07:09.795871", 
  "metric": {
    "type": "wallclock", 
    "name": "default"
  }, 
  "client": {
    "username": "jon", 
    "platform": "Linux-3.13.0-74-generic-x86_64-with-Ubuntu-14.04-trusty", 
    "working_directory": "/home/jon/projects/clusterhq/flocker/benchmark", 
    "nodename": "ecuador", 
    "flocker_version": "1.9.0.dev1+877.g40b4a15"
  }, 
  "samples": [
    {
      "value": 1.7430620193481445, 
      "success": true
    }, 
    {
      "value": 1.7483181953430176, 
      "success": true
    }, 
    {
      "value": 16.281208992004395, 
      "success": true
    }
  ], 
  "control_service": {
    "flocker_version": "1.9.0.dev1+945.ga00891e", 
    "host": "52.32.171.65"
  }, 
  "operation": {
    "type": "create-container", 
    "name": "create-container"
  }
}
```

In addition, a problem was fixed where if the `get_probe` operation failed, the benchmarking hung. The handling of errors has been moved around, so that any error in the sampling creates an unsuccessful sample.

### DIFF
--- a/benchmark/operations/create_container.py
+++ b/benchmark/operations/create_container.py
@@ -4,7 +4,6 @@ Operation to create a container.
 """
 
 from functools import partial
-import random
 from uuid import UUID, uuid4
 
 from pyrsistent import PClass, field
@@ -14,13 +13,8 @@ from flocker.apiclient import MountedDataset
 from flocker.common import gather_deferreds, loop_until
 from flocker.control import DockerImage
 
-from .._interfaces import IProbe, IOperation
-
-
-class EmptyClusterError(Exception):
-    """
-    Exception indicating that the cluster contains no nodes.
-    """
+from benchmark._interfaces import IProbe, IOperation
+from benchmark.operations._common import select_node
 
 
 def loop_until_state_found(reactor, get_states, state_matches):
@@ -201,12 +195,6 @@ class CreateContainerProbe(PClass):
         d = control_service.list_nodes()
 
         # Select an arbitrary node on which to create the container.
-        def select_node(nodes):
-            if nodes:
-                return random.choice(nodes)
-            else:
-                # Cannot proceed if there are no nodes in the cluster!
-                raise EmptyClusterError("Cluster contains no nodes.")
         d.addCallback(select_node)
 
         def parallel_setup(node):
--- a/benchmark/operations/create_container.py
+++ b/benchmark/operations/create_container.py
@@ -111,7 +111,7 @@ def create_container(
             return (
                 expected.name == inspecting.name and
                 expected.node_uuid == inspecting.node_uuid and
-                inspecting.running is True
+                inspecting.running
             )
 
         def find_match(existing_state):
@@ -160,7 +160,7 @@ def delete_container(reactor, control_service, container):
             return (
                 expected.name == inspecting.name and
                 expected.node_uuid == inspecting.node_uuid and
-                inspecting.running is True
+                inspecting.running
             )
 
         def no_running_match(existing_state):
--- a/benchmark/operations/create_container.py
+++ b/benchmark/operations/create_container.py
@@ -296,7 +296,7 @@ class CreateContainerProbe(PClass):
 class CreateContainer(object):
 
     def __init__(
-        self, reactor, cluster, image=u'debian', volume_size=None,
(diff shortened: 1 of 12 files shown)


benchmark/operations/create_container.py
```diff
@@ -0,0 +1,319 @@
...
+        service.
+    :param UUID node_uuid: Node on which to start the container.
+    :param unicode name: Name of the container.
+    :param DockerImage image: Docker image for the container.
+    :param Optional[Sequence[MountedDataset]] volumes: Volumes to attach
+        to the container.
+    :return Deferred[ContainerState]: The state of the created container.
+    """
+
+    def container_started(expected):
```
james-w: :question: This is very similar to the dataset_mounted function above. Could at least
the machinery to do the loop over the states using a callable parameter to do the
matching be extracted?

In fact would it be useful to have a helper to do the loop_until part? It seems
like it is fairly common in the benchmarking code. I don't know if it would be useful
to have one specialised to take a state matcher and an operation that returns a list
of states and to loop until the callable return value is not None.

benchmark/operations/create_container.py
```diff
@@ -0,0 +1,319 @@
...
+from flocker.common import gather_deferreds, loop_until
+from flocker.control import DockerImage
+
+from .._interfaces import IProbe, IOperation
+
+
+class EmptyClusterError(Exception):
+    """
+    Exception indicating that the cluster contains no nodes.
+    """
```
james-w: This hasn't come up in any previous operation? Seems like a fairly generic exception that
could affect a lot of operations. Is there somewhere more central to put it to be reused later?

james-w:
Hi,

Thanks for the branch, great to see such an important operation being added.

I had a couple of questions about possible code reuse, because I don't know
what your plans for the followup branches are.

I'll ask jml for the mentor review.

Thanks,

James

benchmark/operations/create_container.py
```diff
@@ -0,0 +1,319 @@
...
+            state, or None if the expected container is not found or is not
+            running.
+        """
+        d = control_service.list_containers_state()
+
+        def container_matches(inspecting, expected):
+            return (
+                expected.name == inspecting.name and
+                expected.node_uuid == inspecting.node_uuid and
+                inspecting.running is True
```
jml: Please just use `inspecting.running`. See https://www.python.org/dev/peps/pep-0008/#programming-recommendations (grep for "is True")

benchmark/operations/create_container.py
```diff
@@ -0,0 +1,319 @@
...
+from .._interfaces import IProbe, IOperation
+
+
+class EmptyClusterError(Exception):
+    """
+    Exception indicating that the cluster contains no nodes.
+    """
+
+
+def create_dataset(
```
jml: :question: I'd consider either renaming this to `_create_dataset`, or adding some tests that exercise it directly. Likewise for some of the other top-level functions.

jml:
Thanks Jon! It's good to see these benchmarks getting rounded out.

@james-w I've noted one or two things that you've missed (the PEP 8 one needs to be translated into the handbook), but otherwise good stuff.

benchmark/operations/create_container.py
jongiddy: I'm going to leave it as is for now, since these top-level functions will probably get moved to a separate module to help implement the scenario in https://clusterhq.atlassian.net/browse/FLOC-3807

PR #2425, opened 2016-01-06
[FLOC-3686] EBS check attachment state

dwgebler:
Fixes FLOC-3686

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -284,9 +284,19 @@ class VolumeBusy(Exception):
     """
     def __init__(self, volume):
         Exception.__init__(self, volume.id, volume.attachments)
+        logged_attachments = []
+        for attachment in volume.attachments:
+            logged_attachments.append(
+                dict(
+                    instance_id=attachment['InstanceId'],
+                    volume_id=attachment['VolumeId'],
+                    device=attachment['Device'],
+                    state=attachment['State'],
+                )
+            )
         VOLUME_BUSY_MESSAGE(
             volume_id=volume.id,
-            attachments=volume.attachments
+            attachments=logged_attachments
         ).write()
 
 
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -284,8 +284,7 @@ class VolumeBusy(Exception):
     """
     def __init__(self, volume):
         Exception.__init__(self, volume.id, volume.attachments)
-        Message.new(
-            message_type=VOLUME_BUSY_MESSAGE,
+        VOLUME_BUSY_MESSAGE(
             volume_id=volume.id,
             attachments=volume.attachments
         ).write()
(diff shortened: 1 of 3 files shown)


flocker/node/agents/ebs.py
```diff
@@ -276,6 +278,28 @@ class AttachFailed(Exception):
...
+                    instance_id=attachment['InstanceId'],
+                    volume_id=attachment['VolumeId'],
+                    device=attachment['Device'],
+                    state=attachment['State'],
+                )
+            )
+        VOLUME_BUSY_MESSAGE(
+            volume_id=volume.id,
+            attachments=logged_attachments
+        ).write()
```
jml: Please split the logging part of this out of the exception, and instead explicitly log before raising it. 

It's generally better to avoid doing IO in constructors, as it prevents the object in question from being used as a value. (Granted, this is a relatively theoretical consideration with this exception).

flocker/node/agents/functional/test_ebs.py
```diff
@@ -66,6 +69,61 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+    def test_volume_busy_error_on_busy_state(self, logger):
+        """
+        A ``VolumeBusy`` is raised if a volume's attachment state is "busy"
+        when attempting to detach the volume. This can occur if an attempt
+        is made to detach a volume that has not been unmounted.
+        """
+        try:
+            config = get_blockdevice_config(ProviderType.aws)
+        except InvalidConfig as e:
+            self.skipTest(str(e))
```
jml: :question: This snippet is repeated quite a lot. Consider extracting a helper function that does this check, e.g. `require_aws_blockdevice_config`.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -66,6 +69,61 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+
+        # Try to detach and get the VolumeBusy exception.
+        self.assertRaises(
+            VolumeBusy, self.api.detach_volume, flocker_volume.blockdevice_id)
+
+        expected_keys = set(["volume_id", "attachments"])
+        for message in LoggedMessage.of_type(
+            logger.messages, VOLUME_BUSY_MESSAGE
+        ):
+            keys = set(message.message.keys())
```
jml: Don't need to call `.keys()` here. `set` does the right thing.

flocker/node/agents/functional/test_ebs.py
```diff
@@ -66,6 +69,61 @@ class EBSBlockDeviceAPIInterfaceTests(
...
+            self.assertTrue(expected_keys.issubset(keys))
```
jml: You can replace the loop and the assertion with the following matcher:

```python
messages = [message.message for message in LoggedMessage.of_type(logger.messages, VOLUME_BUSY_MESSAGE]
self.assertThat(messages, AllMatch(ContainsAll(expected_keys)))
```

The advantage is that you'll get much more information when the assertion fails. The matchers can be imported from `testtools.matchers`.

jml:
Thanks for looking into this again! A few comments & suggestions. Please merge after addressing and after your tests pass.

flocker/node/agents/functional/test_ebs.py
dwgebler: Nifty! I didn't know about these matchers.

PR #2426, opened 2016-01-07
[FLOC-3812] Allow specification of AWS instance type for acceptance/benchmark tests

jongiddy:
AWS acceptance and benchmark tests always use an `m3.large` instance.  This patch allows the `aws` configuration stanza to contain an `instance_type` entry to override this.

The field is optional. If missing, it default to the current behavior.

To test, add `instance_type: m3.xlarge` (or any other AWS instance type) to your acceptance testing configuration file and start a cluster using `admin/run-acceptance-tests` or `admin/setup-cluster`. Using the AWS console, you can see the instance types for the cluster nodes.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -961,6 +961,7 @@ class CommonOptions(Options):
                  secret_access_token: <aws secret access token>
                  keyname: <ssh-key-name>
                  security_groups: ["<permissive security group>"]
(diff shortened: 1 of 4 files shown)


jml:
Great stuff. I love incremental improvements like these. Only failure is a known flake, so merge when ready.

PR #2427, opened 2016-01-07
[FLOC-3760] Benchmark userdata option

jongiddy:
Provide the ability to supply additional data for the JSON output of the benchmark script.  This is useful to provide data about the benchmark run that cannot easily be determined automatically by the script.

The particular use-case is to supply the Git branch for the benchmark cluster. While the Git SHA is easily obtained through the version number, it is difficult (and, in general, impossible) to determine the intended branch name.  This PR allows that information to be provided by the caller of the benchmark script, who has more information about the cluster, including which branch they deployed.

### DIFF
--- a/benchmark/script.py
+++ b/benchmark/script.py
@@ -206,12 +206,12 @@ def get_cluster(options, env):
     return cluster
 
 
-def add_userdata(options, result):
+def parse_userdata(options):
     """
     Parse the userdata option and add to result.
 
     :param BenchmarkOptions options: Script options.
-    :param dict result: Result dictionary.
+    :return: Parsed user data.
     """
     userdata = options['userdata']
     if userdata:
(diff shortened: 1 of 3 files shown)


benchmark/script.py
```diff
@@ -205,6 +206,31 @@ def get_cluster(options, env):
...
+                        result['userdata'] = json.load(f)
+                except IOError as e:
+                    usage(
+                        options,
+                        'Invalid user data file: {}'.format(e.strerror)
+                    )
+            else:
+                result['userdata'] = json.loads(userdata)
+        except ValueError as e:
+            usage(options, 'Invalid user data: {}'.format(e.args[0]))
```
jml: :question: I would strongly consider changing this function so that it _doesn't_ take `result` and instead just returns the userdata extracted from `BenchmarkOptions` (or even just take `options['userdata']` as a parameter). Not only will it simplify the function, but it will also simplify the tests.

benchmark/test/test_script.py
```diff
@@ -424,3 +427,101 @@ def test_default_metric(self):
...
+        add_userdata(options, result)
+        self.assertEqual(result, {"userdata": {"branch": "master"}})
+
+    def test_invalid_json(self):
+        """
+        Invalid JSON string handled.
+        """
+        options = BenchmarkOptions()
+        options.parseOptions(['--userdata', '"branch": "master"'])
+        with capture_stderr() as captured_stderr:
```
jml: :question: It's introduced in this branch, so optional, but I think `capture_stderr` needs to have a `try` / `finally` around the yield. Probably there should also be tests to check that it restores stderr.

jml:
Thanks! A couple of suggestions, but no need for a follow-up review unless you particularly want one. Please merge after addressing comments to your satisfaction and after the tests pass.

PR #2428, opened 2016-01-07
[FLOC-3823] Parametrize number of benchmark samples to take

jongiddy:
Allow the number of benchmarking samples to be provided as a command line parameter, rather than fixed at 3.

A particular use case is to help with generating benchmarks that are specified as meeting a limit in X% of cases. Running 100 samples can generate results to easily check such benchmarks.

### DIFF
--- a/benchmark/_driver.py
+++ b/benchmark/_driver.py
@@ -62,7 +62,7 @@ def sample(operation, metric, name):
         return sampling.addActionFinish()
 
 
-def benchmark(scenario, operation, metric, num_samples=3):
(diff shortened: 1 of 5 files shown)


james-w:
:question: The code all looks fine, but there aren't any tests that the parameter is threaded
through correctly. Would it be possible to test that at some level, even if it's not right
from the command line?

@jml: mentor review please.

Thanks,

James

jml:
:+1: on mentor review.

jml:
Also, thanks @jongiddy for so consistently updating the prose docs.

PR #2429, opened 2016-01-07
[FLOC-3757] Benchmark add tolerance to load scenarios 

mpgomez:


benchmark/scenarios/_request_load.py
```diff
@@ -75,7 +75,8 @@ class RequestLoadScenario(object):
 
     def __init__(
         self, reactor, scenario_setup_instance, request_rate=10,
-        sample_size=DEFAULT_SAMPLE_SIZE, timeout=45
+        sample_size=DEFAULT_SAMPLE_SIZE, timeout=45,
+        tolerance_percentage=0.2
```
justnoise: Probably want to update docstrings with the new parameter

benchmark/scenarios/read_request_load.py
```diff
@@ -65,4 +66,5 @@ def read_request_load_scenario(reactor, cluster, request_rate=10,
         request_rate=request_rate,
         sample_size=sample_size,
         timeout=timeout,
+        tolerance_percentage=tolerance_percentage,
```
justnoise: The tolerance_percentage needs to be added to the docstring.

benchmark/scenarios/test/test_read_request_load.py
```diff
@@ -183,6 +183,7 @@ def test_scenario_throws_exception_when_already_started(self, _logger):
 
         self.assertRaises(RequestScenarioAlreadyStarted, s.start)
 
+
```
sarum90: :grey_question: I thought we did just 1 empty line between member functions, but lint appears to be happy so I'm happy. For consistency, it'd probably be better with only 1 newline. But that's better enforced by linters than reviewers, so no action required.

benchmark/scenarios/test/test_read_request_load.py
```diff
@@ -216,6 +218,37 @@ def test_scenario_throws_exception_when_rate_drops(self, _logger):
...
+        cluster = self.make_cluster(RequestDroppingFakeFlockerClient)
+        sample_size = 5
+        s = read_request_load_scenario(c, cluster, sample_size=sample_size,
+                                       tolerance_percentage=0.6)
+        cluster.get_control_service(c).drop_requests = True
+        d = s.start()
+
+        c.pump(repeat(1, sample_size))
+
+        cluster.get_control_service(c).drop_requests = True
```
sarum90: I don't see the purpose of lines 241-243. They look like they might be a copy/paste mistake? Or maybe they are needed, but certainly line 241 would merit a further explanation coment. Looking at RequestDroppingFakeFlockerClient, I don't see how drop_requests would become unset, but I'm pretty new to this code.

benchmark/scenarios/test/test_read_request_load.py
```diff
@@ -216,6 +218,37 @@ def test_scenario_throws_exception_when_rate_drops(self, _logger):
...
+
+        c.pump(repeat(1, sample_size))
+        s.maintained().addBoth(lambda x: self.fail())
+        d.addCallback(lambda ignored: s.stop())
+        # Generate enough samples to finish the scenario
+        c.pump(repeat(1, sample_size*s.request_rate))
```
sarum90: :grey_question: I'm curious, is there any reason not to only pump the clock at the end? Since the code is all async, I was expecting something more along the lines of:
```
 # Set up deferreds:
...
 s.maintained().addBoth(lambda x: self.fail())
....
 # Pump the clock until the scenario is done:
  c.pump(repeat(1, sample_size * s.request_rate))
  self.successResultOf(d)
```
without any of the earlier pumps. Unless you meant for a block of successes before you started dropping requests?

sarum90:
Looks like a nice approach to the problem, Thanks.

Looks great to me - just a couple tips to make the test a bit cleaner. Thanks @justnoise for your review as well, it'd be nice to keep the docstrings up to date.

mpgomez:
Benchmark green. Jenkins missbehaving. Merging with a partially green build.

PR #2430, opened 2016-01-07
[FLOC-3814] First pass implementation of GCE Driver.

sarum90:
Just enough of a GCE driver to pass make_iblockdeviceapi_tests.

This leaves lots of TODOs which will be sharded into JIRA issues and completed before the GCE driver is released.


flocker/node/agents/pd.py
```diff
@@ -0,0 +1,371 @@
...
+    Design implications:
+        - The GCE disk names meet the length and format requirements, and are
+            as unique as blockdevice_ids.
+        - We can perform operations on a disk only knowing its blockdevice_id
+            without making additional API calls or looking things up in a
+            table.
+        - dataset_id is a pure function of blockdevice_id and vice versa.
+        - You can have multiple clusters within the same project.
+        - Multiple clusters within the same project cannot have datasets with
+            the same UUID. **
```
sarum90: Design Reviewer: Your thoughts on this please. If you think this breaks the implementation we could do something like:

- picking a new UUID for the blockdevice_id (setting name to that), storing the dataset_id and cluster_id in the description

This would require that we would always have to list(or get) the volume every time we needed the dataset_id from the blockdevice_id. Maybe that's not the end of the world, but it might make the user experience worse as the GCE console only shows name in the menu, you have to click on a volume to get the description.

Also then we couldn't filter on cluster_id using the built in filter, we would have to do it manually. Again not the end of the world, but something to consider.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,371 @@
...
+
+    Design:
+        - For a given dataset_id (06e07bcc-fea1-4810-8c6d-7487196998b6) prefix
+            the dataset_id with "flocker-v1-" to get the unicode blockdevice-id
+            (flocker-v1-06e07bcc-fea1-4810-8c6d-7487196998b6).
+        - Set the name of the GCE disk resource to the blockdevice_id. This
+            lets us perform operations on the disk only knowing its
+            blockdevice_id.
+        - Set the description of the disk resource to:
+            "flocker-v1-cluster: <cluster-uuid>".
```
sarum90: This isn't implemented yet.

I was thinking of sending this up for code review once this is implemented, and then have new PRs for the TODOs and adding more tests after that. Let me know if you think that is a bad plan.

flocker/node/agents/test/blockdevicefactory.py
```diff
@@ -209,10 +213,24 @@ def _aws(cluster_id, config):
...
+        necessary to authenticate a PD session.
+    :return: A PDBlockDeviceAPI instance.
+    """
+    return PDBlockDeviceAPI(**config)
+
 # Map provider labels to IBlockDeviceAPI factory.
 _BLOCKDEVICE_TYPES = {
     Providers.OPENSTACK: _openstack,
     Providers.AWS: _aws,
+    Providers.GCE: _gce,
```
sarum90: Release planning question. I don't intend on putting this in release notes until GCE support is fully tested (in our Jenkins, and all the TODOs taken care of).

However (I'll have to double check this) but I think you can technically attempt to use GCE as soon as it is in the code. Should we be putting it behind some sort of release flag (or other technical barrier) until it is "ready"?

sarum90:
My three comments above are specific questions for the design reviewer. If you don't know the answer lmk and I'll escalate to slack to get feedback.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,371 @@
...
+    users = disk.get('users', [])
+    if not users:
+        return None
+    return unicode(users[0].split('/')[-1])
+
+
+@implementer(IBlockDeviceAPI)
+class PDBlockDeviceAPI(object):
+    """
+    A PD implementation of ``IBlockDeviceAPI`` which creates block devices in a
```
sarum90: Here is the meat of the design. The only think I am expecting in design review is for you to read through this and glance at my other comments on the PR. If you want to dive into the other implementation details you can, but that will already happen in code review.

flocker/node/agents/pd.py
itamarst: UUIDs are supposed to be universal, and we assume that in lots of places, so that seems fine.

Except... users can pick dataset IDs! We ripped out logic from Docker plugin that requires that as of next release, so there's just `flocker-deploy` (which is a toy, and will be deprecated soon I hope) and the fact our API allows it. But really if you're picking duplicate IDs you're doing it wrong, so I think it's fine to have this restriction so long as it's documented.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,383 @@
...
+            equals and not equals.
+        - GCE lets you set a token in the SCSI info pages on your blockdevices
+            every time you attach them to a VM so that symlinks are created at
+            '/dev/disk/by-id/google-<token>'.
+            - This technically is brittle to implementation, configuration, and
+                version of udev and the SCSI driver within the guest OS.
+        - GCE resource names and descriptions are immutable after creation.
+
+    Design:
+        - For a given dataset_id (06e07bcc-fea1-4810-8c6d-7487196998b6) prefix
```
itamarst: Can you set these atomically with creation? If so that's way better than AWS.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,383 @@
+# -*- test-case-name: flocker.node.agents.functional.test_pd -*-
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+A PD implementation of the ``IBlockDeviceAPI``.
```
itamarst: 1. Please define "PD".
2. Link to relevant documentation (API and tutorial) to ease life for reviewers and later maintainers.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,383 @@
...
+            as unique as blockdevice_ids.
+        - We can perform operations on a disk only knowing its blockdevice_id
+            without making additional API calls or looking things up in a
+            table.
+        - dataset_id is a pure function of blockdevice_id and vice versa.
+        - You can have multiple clusters within the same project.
+        - Multiple clusters within the same project cannot have datasets with
+            the same UUID. **
+        - We could add filtering by cluster by filtering on description.
+        - The path of the volume (or at least the path to a symlink to a path
```
itamarst: Presumably this means the "path to the device". This sounds like a great way to reduce errors, yeah. However, you indicated it was brittle... is it sufficiently not brittle?

flocker/node/agents/test/blockdevicefactory.py
itamarst: This is test code, doesn't seem like an issue.

flocker/node/agents/pd.py
itamarst: API docs say yes.

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,383 @@
...
+                return latest_operation
+            return None
+
+        return poll_until(finished_operation_result, [1]*35)
+
+    def allocation_unit(self):
+        """
+        Can only allocate PDs in GiB units.
+
+        Documentation claims `GB` but experimentally this was determined to
```
itamarst: :crying_cat_face: :gun:

flocker/node/agents/pd.py
```diff
@@ -0,0 +1,383 @@
...
+                # way to get a 400.
+                raise UnknownVolume(blockdevice_id)
+            else:
+                raise e
+        errors = result.get('error', {}).get('errors', [])
+        for e in errors:
+            if e.get('code') == u"RESOURCE_IN_USE_BY_ANOTHER_RESOURCE":
+                raise AlreadyAttachedVolume(blockdevice_id)
+
+        # TODO(mewert): This is sort of bad to just add an extra API call here
```
itamarst: Attaches don't happen very often, doesn't seem worth caching.

flocker/node/agents/pd.py
itamarst: Possibly this module should be named "gce"? I know that's inconsistent but maybe it's the other modules that have bad names :grinning:

flocker/node/agents/pd.py
itamarst: I think that's reasonable. I would suggest having the first code review PR actually have a Jenkins builder setup... it's much more useful to actually have passing tests.

itamarst:
Design sounds reasonable at first glance; we already have some backends that hard code dataset ID in block device ID.

However... we plan on having *Flocker* record mapping between dataset and block device IDs and be the canonical source of information. So if the creation API call succeeds but Flocker never gets a response, Flocker might retry... and the create will fail with "already created". Which I guess is a case we'd have to handle anyway. So just something to keep in mind I guess, doesn't impact design.

itamarst:
Please proceed!

flocker/node/agents/pd.py
sarum90: I think for a given distro and release it is sufficiently not brittle (i.e. if it works for CentOS 7 and ubuntu 14.04 then it should always work for CentOS 7 and ubuntu 14.04).

Also the brittle bar is pretty low in comparison to the brittleness of trying to manage our own guessing as to which /dev/sd* blockdevice we just attached.

I'll probably rephrase this to something along the lines of "This might be brittle if you are running flocker on an unsupported OS". just as a breadcrumb for those who try to do that.

flocker/node/agents/pd.py
justnoise: We've added a comment on FLOC-3848 to be sure to document this assumption.

flocker/node/agents/pd.py
justnoise: To avoid a gigantic PR, we'll likely add a new PR that addresses the lack of jenkins tests.  Those are important.

flocker/node/agents/pd.py
justnoise: Sounds good.  We've worked to convince ourselves that we'll use "the most recognizable and reasonable product name" as the module name.

flocker/node/agents/functional/test_gce.py
```diff
@@ -0,0 +1,97 @@
...
+``FLOCKER_FUNCTIONAL_TEST=TRUE
+FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE=$HOME/acceptance.yml
+FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER=gce``
+
+The configuration stanza for the GCE backend should resemble:
+``
+gce:
+    zone: <gce-region>
+    project: <gce-project-name>
+``
```
jml: :question: It doesn't matter because we don't render docs, but block code syntax is:

```rst

blah blah the following environment variables::

    FLOCKER_THING=aoeusa
    FLOCKER_Ouoa

```

(two colons, blank line, four space indent, blank line)

flocker/node/agents/functional/test_gce.py
```diff
@@ -0,0 +1,97 @@
...
+from ..test.test_blockdevice import (
+    make_iblockdeviceapi_tests
+)
+
+from ..test.blockdevicefactory import (
+    ProviderType, get_blockdeviceapi_with_cleanup,
+    get_minimum_allocatable_size, get_device_allocation_unit
+)
+
+from ....testtools import TestCase
```
jml: :crying_cat_face:

flocker/node/agents/functional/test_gce.py
```diff
@@ -0,0 +1,97 @@
...
+    :class:`IBlockDeviceAPI` Interface adherence Tests for
+    :class:`GCEBlockDeviceAPI`.
+    """
+
+    def test_attach_elsewhere_attached_volume(self):
+        # This test in make_iblockdevice_api is a terrible hack:
+        # https://clusterhq.atlassian.net/browse/FLOC-1839
+        # Rather than add racy code that checks for if a volume is attached
+        # before attempting the attach, just skip this test for this driver.
+        # TODO(mewert): replace with a good test.
```
jml: :question: Maybe file a bug / subtask for this and reference here?

flocker/node/agents/gce.py
```diff
@@ -0,0 +1,385 @@
+# -*- test-case-name: flocker.node.agents.functional.test_gce -*-
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+A GCE Persistent Disk (PD) implementation of the ``IBlockDeviceAPI``.
+
+The following resources are helpful to referrence while maintaining this
```
jml: nit: "reference" (but why not "refer to"?)

flocker/node/agents/gce.py
```diff
@@ -0,0 +1,385 @@
...
+            latest_operation = self._compute.zoneOperations().get(
+                project=self._project,
+                zone=self._zone,
+                operation=operation_name).execute()
+            # TODO Logging
+            if latest_operation['status'] == 'DONE':
+                return latest_operation
+            return None
+
+        return poll_until(finished_operation_result, [1]*35)
```
jml: Maybe add a comment saying why this timeout?

flocker/node/agents/gce.py
```diff
@@ -0,0 +1,385 @@
...
+        `function` returns an object that has an `execute()` method that
+        returns a GCE operation resource dict.
+
+        This function will then poll the operation until it reaches state
+        'DONE' or times out, and then returns the final operation resource
+        dict.
+
+        :param function: Callable that takes keyword arguments project and
+            zone, and returns an executable that results in a GCE operation
+            resource dict as described above.
```
jml: :question: Is it worth defining a local interface for `IExecutable`? I don't think so.

flocker/node/agents/gce.py
```diff
@@ -0,0 +1,385 @@
...
+            if (disk['name'].startswith(_PREFIX) and
+                disk['description'] == self._disk_resource_description())
+        )
+
+    def compute_instance_id(self):
+        """
+        GCE does operations based on the `name` of resources, and also assigns
+        the name
+
+        """
```
jml: Nit: remove blank line.

jml:
Thanks Marcus! Great to see this happen. 

Is there a builder set up that will run these functional tests in our CI? Is there a ticket for setting up such a builder?

Otherwise, only nits & super optional comments. Please merge after addressing and the tests pass (ideally with the new builder in place, but OK without it if we actually add it asap).

flocker/node/agents/gce.py
```diff
@@ -0,0 +1,385 @@
...
+
+"""
+A GCE Persistent Disk (PD) implementation of the ``IBlockDeviceAPI``.
+
+The following resources are helpful to referrence while maintaining this
+driver:
+- Rest API: https://cloud.google.com/compute/docs/reference/latest/
+- Python Client: https://cloud.google.com/compute/docs/tutorials/python-guide
+- Python API: https://google-api-client-libraries.appspot.com/documentation/compute/v1/python/latest/
+- Python Oauth: https://developers.google.com/identity/protocols/OAuth2ServiceAccount#authorizingrequests
```
jml: These trigger lint errors. Not 100% sure how best to fix.

flocker/node/agents/functional/test_gce.py
sarum90: Clarified offline. No action required, jml is lamenting the use of relative imports. I have sympathy for the feeling that relative imports are hard to maintain, but they seem to be standard in the flocker codebase so probably going to leave as is.

flocker/node/agents/gce.py
justnoise: I'm hoping adding a `# noqa` to the end of the line should do it.  If that doesn't work, I'll look into something more exciting.

flocker/node/agents/gce.py
justnoise: We plan on gathering sampling latency for various operations inside GCE and use that to create better informed timeouts.

flocker/node/agents/gce.py
justnoise: We've had an offline discussion about this and it's probably not worth it.

sarum90:
FLOC-3824 is up for adding these tests to jenkins. Unfortunately these have to be run on GCE, which means we have to have GCE slaves in jenkins. Brendan and I worked all week on that without any immediate success, so leaving that as future work. That is definitely on the top of the stack as the next thing to do though.

PR #2433, opened 2016-01-08
[FLOC-3833] Allow the method used by read request scenario to be set as a parameter.

jongiddy:
Allow the operation used in the benchmarking read load scenario to be set as a parameter.

FLOC-3399 specifies a particular operation to be used for benchmarking, which was different to the existing operation.  The read request /operation/ already allowed the method to be specified as a parameter.

This PR adds the same ability to the read load scenario. Major changes include the extraction of common code for validating the method name into a separate file, and modifying the fake benchmark clients for testing to handle any method, not just the original one.

Also, there was no documentation for the write load scenario, so this was added while updating the read load scenario docs.

### DIFF
--- a/benchmark/_method.py
+++ b/benchmark/_method.py
@@ -1,0 +1,31 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Functions for checking a method name provided by the user.
+"""
+
+
+class InvalidMethod(Exception):
+    """
+    Method not meeting requested criteria.
+    """
+
+
+def validate_no_arg_method(interface, method_name):
+    """
+    Check that method name exists in interface and requires no parameters.
+
+    :param zope.interface.Interface interface: Interface to validate against.
+    :param str method_name: Method name to validate.
+    :raise InvalidMethod: if name is not valid or requires parameters.
+    """
+    for name, method in interface.namesAndDescriptions():
+        if name == method_name:
+            if len(method.getSignatureInfo()['required']) > 0:
+                raise InvalidMethod(
+                    'Method {!r} requires parameters'.format(method_name)
+                )
+            return
+    raise InvalidMethod(
+        'Method {!r} not found in interface {}'.format(
+            method_name, interface.__name__)
+    )
--- a/benchmark/test/test_method.py
+++ b/benchmark/test/test_method.py
@@ -1,0 +1,53 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Method tests for the control service benchmarks.
+"""
+
+from zope.interface import Interface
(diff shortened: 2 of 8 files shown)


benchmark/_method.py
```diff
@@ -0,0 +1,31 @@
...
+def validate_no_arg_method(interface, method_name):
+    """
+    Check that method name exists in interface and requires no parameters.
+
+    :param zope.interface.Interface interface: Interface to validate against.
+    :param str method_name: Method name to validate.
+    :raise InvalidMethod: if name is not valid or requires parameters.
+    """
+    for name, method in interface.namesAndDescriptions():
+        if name == method_name:
```
jml: :question:

```python
method = dict(interface.namesAndDescriptions()).get(method_name)
```

(I think what you've got is fine, and am not sure mine is better).

benchmark/_method.py
```diff
@@ -0,0 +1,31 @@
...
+            if len(method.getSignatureInfo()['required']) > 0:
+                raise InvalidMethod(
+                    'Method {!r} requires parameters'.format(method_name)
+                )
+            return
+    raise InvalidMethod(
+        'Method {!r} not found in interface {}'.format(
+            method_name, interface.__name__)
+    )
```
jml: :question: It occurs to me that a different way of handling this is to have a named interface for a no-arg function, and just use Zope's builtin verification. e.g.

```python

class INoArgMethod(Interface):
    def __call__(self):
        """
        Meaningful documentation.
        """
```

Not 100% convinced it's better though.

benchmark/test/test_method.py
```diff
@@ -0,0 +1,53 @@
...
+        self.assertIn('requires parameters', str(exception))
+
+    def not_present_fails_noargs_method(self):
+        """
+        A non-present method fails to validate as no-arg method.
+        """
+        exception = self.assertRaises(
+            InvalidMethod, validate_no_arg_method, ITest, 'notpresent'
+        )
+        self.assertIn('not found in interface', str(exception))
```
jml: None of these tests will be executed (missing `test_` prefix). Is this deliberate?

jml:
Thanks Jon! 

I think after all the faffing about I've only got one comment. Please address & merge after tests pass.

benchmark/test/test_method.py
jongiddy: I need a linter that complains about methods in a TestCase that do not start with `test` and are not called from any other method.

jml: You can probably configure pylint to do that actually. A thing I'd like to do (in my copious free time) is add a pylint check to our suite and gradually customize it to implement our style.

PR #2434, opened 2016-01-08
[FLOC-3832] Add flaky decorator for affected test

dwgebler:
The build failures we're seeing for this test are most likely AWS' eventual consistency biting us - the logs for the test indicate Flocker is attempting to do the right things it's supposed to be doing, but not getting the expected volume states from EC2. So this PR just adds the flaky decorator to allow us to accept at least 1 out of 3 runs passing.


jml:
:+1: Thanks!

PR #2435, opened 2016-01-08
[FLOC-3799] Collect artifacts (logs) from cron-triggered acceptance test jobs

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3799

This removes the separate `copyArtifact`-calling loop over some of the jobs (all except the cron jobs) and replaces it with a `copyArtifact` call in the `build_steps` method that is called for every job.

This has two consequences:
  * The `CopyArtifact` configuration is now part of each job rather than all of the `CopyArtifact` configuration being part of the main multijob (incidental but, I think, an improvement).
  * Cron-style jobs now have `CopyArtifact` configuration if they define artifacts that should be copied (the actual point of this PR).

Visual inspection of the generated configuration xml suggest these are actually the consequences achieved by this PR.

### DIFF
--- a/jobs.groovy.j2
+++ b/jobs.groovy.j2
@@ -6,6 +6,24 @@
   required views, folders, and jobs for a particular project in the Jenkins
   interface.
 
+  The easiest way to debug or experiment with changes to this is:
+    * check them in to a branch
+    * trigger a build of that branch on the setup_ClusterHQ-flocker job
+    * ssh into the master (same hostname as the web ui)
+    * compare the master configuration xml with the branch configuration xml:
+
+        BASE=/var/lib/jenkins/jobs/ClusterHQ-flocker/jobs/
+        A=master
+        B=copy-artifacts-for-all-jobs-FLOC-3799
+        for config in $(find ${BASE}/${A} -name config.xml | cut -d '/' -f 10- | sort); do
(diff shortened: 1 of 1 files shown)


jobs.groovy.j2
```diff
@@ -289,14 +307,36 @@ def build_publishers(v) {
...
             }
         }
+
+        /* not every job produces an artifact, so make sure we
+           don't try to fetch artifacts for jobs that don't
+           produce them */
+        if (job.archive_artifacts) {
+            for (artifact in job.archive_artifacts) {
+                copyArtifacts(
+                    "${dash_project}/${dash_branch_name}/${job_name}") {
```
james-w: :question: This looks like it would be a common pattern, could it be a function?

james-w:
Hi, thanks for the branch, it looks like a better organisation anyway. Special
thanks for better documenting how to test changes.

:+1: Please land as you like, pending mentor review.

@jml mentor review please.

jml:
Mentor review approval. No feedback, meta or otherwise.

jobs.groovy.j2
exarkun: Yea, I agree that would make sense.  I started down this road while working on this and encountered some unexpected difficulties that made me set that aside.  I may make another attempt if I get a chance, or we should at least suggest this to whoever does the next significant bit of work on this file. :)

exarkun:
Thanks!  Merging now.

PR #2436, opened 2016-01-08
[FLOC-3836] Run all of test_docker

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3836

Does what it says on the tin.


itamarst:
Merge when green.

PR #2437, opened 2016-01-11
[FLOC-3820] Benchmark latency count

jongiddy:
Add benchmark to provide statistics for request time for the read and write request load scenarios.  This is the final benchmark required for January/March benchmark tests.

The statistics for these benchmarks were already being generated in the `RateMeasurer` object used for the scenarios that generate requests, so this PR focusses on extracting the information and providing them to the user.

Normally a benchmark measures a specific operation, with a scenario running outside the measurement to provide the required load. Since the metric, which measures the operation, does not even know about the scenario, the PR adds a `scenario_metrics` output to the usual measured samples.  This keeps the scenario and metric independent of each other.

The output for the scenario now contains additional values, which can be used to calculate the percentage of requests completing within a time limit.  The `calltimes` dictionary indicates the number of times the request took the specified seconds (i.e. the dictionary key is the seconds required for the request, the dictionary value is the number of occurrences of a request taking that many seconds).  

Seconds are integers, and are rounded down from actual value (e.g. 1.9 seconds increments the "1" key).  Since we're looking for calls that take longer than 30 seconds, this rounding is unlikely to significantly affect the results.

```
$  ./benchmark --scenario read-request-5 --operation wait-10
{
  "scenario": {
    "metrics": {
      "ok_count": 185, 
      "calltimes": {
        "0": 1, 
        "1": 169, 
        "2": 15
      }, 
      "err_count": 0, 
      "errors": {}
    }, 
    "type": "read-request-load", 
    "request_rate": 5, 
    "name": "read-request-5"
  }, 
  "samples": [
    {
      "value": 10.000745058059692, 
      "success": true
    }, 
    {
      "value": 10.000056982040405, 
      "success": true
    }, 
    {
      "value": 10.00004506111145, 
      "success": true
    }
  ], 
...
}
```

### DIFF
--- a/benchmark/_driver.py
+++ b/benchmark/_driver.py
@@ -70,12 +70,11 @@ def benchmark(scenario, operation, metric, num_samples=3):
     :param IOperation operation: An operation to perform.
     :param IMetric metric: A quantity to measure.
     :param int num_samples: Number of samples to take.
-    XXX - modify docstring
-    :return: Deferred firing with a list of samples. Each sample is a
-        dictionary containing a ``success`` boolean. If ``success is True``,
-        the dictionary also contains a ``value`` for the sample measurement.
-        If ``success is False``, the dictionary also contains a ``reason`` for
-        failure.
+    :return: Deferred firing with a tuple containing one list of
+        benchmark samples and one scenario metrics result. See the
+        ``sample`` function for the structure of the samples.  The
+        scenario metrics are a dictionary containing information about
+        the scenario.
     """
     scenario_established = scenario.start()
 
@@ -110,7 +109,10 @@ def benchmark(scenario, operation, metric, num_samples=3):
         d.addCallback(combine_results)
 
         return d
-    benchmarking.addBoth(stop_scenario)
+    benchmarking.addCallbacks(
+        stop_scenario,
+        bypass, errbackArgs=[scenario.stop]
+    )
 
     return benchmarking
 
--- a/benchmark/_driver.py
+++ b/benchmark/_driver.py
@@ -70,6 +70,7 @@ def benchmark(scenario, operation, metric, num_samples=3):
     :param IOperation operation: An operation to perform.
     :param IMetric metric: A quantity to measure.
     :param int num_samples: Number of samples to take.
+    XXX - modify docstring
     :return: Deferred firing with a list of samples. Each sample is a
         dictionary containing a ``success`` boolean. If ``success is True``,
         the dictionary also contains a ``value`` for the sample measurement.
@@ -101,7 +102,15 @@ def benchmark(scenario, operation, metric, num_samples=3):
 
     benchmarking = scenario_established.addCallback(collect_samples)
 
-    benchmarking.addBoth(bypass, scenario.stop)
+    def stop_scenario(samples):
+        d = scenario.stop()
+
+        def combine_results(scenario_metrics):
+            return (samples, scenario_metrics)
+        d.addCallback(combine_results)
+
+        return d
+    benchmarking.addBoth(stop_scenario)
 
     return benchmarking
 
@@ -141,8 +150,11 @@ def driver(
 
     d.addCallback(run_benchmark)
 
-    def add_samples(samples, result):
+    def add_samples(outputs, result):
+        samples, scenario_metrics = outputs
         result['samples'] = samples
+        if scenario_metrics:
+            result['scenario']['metrics'] = scenario_metrics
         return result
 
     d.addCallback(add_samples, result)
--- a/benchmark/scenarios/_rate_measurer.py
+++ b/benchmark/scenarios/_rate_measurer.py
@@ -23,6 +23,7 @@ class RateMeasurer(object):
         self._received = 0
         self._errors = 0
         self._rate = 0
+        self._count = {}  # XXX - count of calltimes
 
     def request_sent(self):
         """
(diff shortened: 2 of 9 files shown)


benchmark/_driver.py
```diff
@@ -70,11 +70,11 @@ def benchmark(scenario, operation, metric, num_samples):
...
-    :return: Deferred firing with a list of samples. Each sample is a
-        dictionary containing a ``success`` boolean. If ``success is True``,
-        the dictionary also contains a ``value`` for the sample measurement.
-        If ``success is False``, the dictionary also contains a ``reason`` for
-        failure.
+    :return: Deferred firing with a tuple containing one list of
+        benchmark samples and one scenario metrics result. See the
+        ``sample`` function for the structure of the samples.  The
+        scenario metrics are a dictionary containing information about
+        the scenario.
```
jml: :question: Alternatively, define a `PRecord` for samples and express this type using PEP-484 syntax. (I guess `Deferred[([[Sample], ScenarioMetricsResult)]`?)

benchmark/scenarios/_rate_measurer.py
```diff
@@ -30,23 +36,25 @@ def request_sent(self):
         """
         self._sent += 1
 
-    def response_received(self, ignored):
+    def response_received(self, calltime):
```
jml: I think this should be `call_time`. `call_duration` would be even more clear.

benchmark/scenarios/_rate_measurer.py
```diff
@@ -30,23 +36,25 @@ def request_sent(self):
...
         """
         Increase the number of received requests.
 
-        :param ignored: The result of a callback. This parameter is
-            not used.
+        :param float calltime: Time to perform call
```
jml: Please say "Time taken to perform call" to disambiguate from the other meaning of "When to perform the call"

benchmark/scenarios/_rate_measurer.py
```diff
@@ -30,23 +36,25 @@ def request_sent(self):
...
         """
         self._received += 1
+        calltime = int(calltime)
```
jml: :question: I saw your comment in the cover letter. Could we perhaps store these as milliseconds for increased granularity? It'd be nice to aim high.

benchmark/scenarios/_rate_measurer.py
```diff
@@ -61,10 +69,21 @@ def outstanding(self):
...
         Return the current rate.
         """
         return self._rate
+
+    def get_metrics(self):
+        """
+        Return the collected metrics.
+        """
+        return {
+            'calltimes': self._calltimes,
```
jml: `call_times` or `call_durations` please.

benchmark/scenarios/_request_load.py
```diff
@@ -128,10 +128,19 @@ def handle_request_error(result):
...
+                return self.reactor.seconds() - t0
+            d.addCallback(get_time)
+
             self.rate_measurer.request_sent()
-            d.addCallbacks(self.rate_measurer.response_received,
-                           errback=handle_request_error)
+            d.addCallbacks(
+                self.rate_measurer.response_received,
+                handle_request_error
+            )
```
jml: I realize this isn't introduced in this branch, but, what's up with the Deferred not being returned?

benchmark/scenarios/_request_load.py
```diff
@@ -248,6 +257,15 @@ def stop(self):
             action_type=u'flocker:benchmark:scenario:stop',
             scenario='request_load'
         ):
+            def no_outstanding_requests():
+                return self.rate_measurer.outstanding() == 0
+
+            scenario_stopped = loop_until(self.reactor,
+                                          no_outstanding_requests,
+                                          repeat(1))
```
jml: :question: Move to a constant?

jml:
Thanks! Please merge after tests pass and addressing comments.

benchmark/scenarios/_request_load.py
jongiddy: Initially I thought it was a bug that escaped because the result gets wrapped in `maybeDeferred`, so it doesn't cause any detectable errors, except that it might mean that the function is re-entered while it is still running.  But even that isn't actually a problem.  It just makes it less likely that a call will be skipped, and have to be caught up using the `count` parameter.

In fact, gathering the Deferreds would possibly break things because we're trying to launch `request_rate` operations per second, and waiting for a laggard would drop the rate too far.

Essentially, the code is correct to return immediately since we are trying to start a certain number of requests per second, and we don't want this function to wait for anything.  The Deferreds that are created inside the function are handled correctly by reporting to the Rate Measurer. And we have a separate check for overload in `check_rate`.

I do see 2 things that need to be fixed:
1. We should return `succeed(None)` and add some comments to indicate that we know what we're doing.
2. There is a bug in that if the loop is delayed, the rate measurer catches up (it gets called `count` times), but we only start one batch of requests, not `count` batches.  However, I will fix these in a separate PR.

benchmark/scenarios/_rate_measurer.py
jongiddy: The nice thing about integers is that it gives a shape to the raw data that can be quickly scanned and basic stats can be performed in your head. As a compromise, I'm changing it to 1 decimal place.

jml: FWIW, I meant milliseconds as integers.

PR #2439, opened 2016-01-11
[FLOC-3839] AWS session token support for acceptance and benchmark tests

jongiddy:
The AWS benchmarking account uses roles to allow main account users access.  Roles use temporary credentials, which are like normal credentials, but add a session token.

This PR allows the session token to be provided as an additional property in the YAML configuration file for acceptance and benchmark tests.

The token is optional. If not supplied in the YAML file, it defaults to `None`, which indicates that no session token is being provided.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -959,6 +959,7 @@ class CommonOptions(Options):
                  zone: <aws zone, e.g. "us-west-2a">
                  access_key: <aws access key>
                  secret_access_token: <aws secret access token>
+                 session_token: <optional session token>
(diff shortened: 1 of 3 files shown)


james-w:
:+1: Thanks for making it work for the rest of us.

I'm guessing there will be some more in-depth docs appearing somewhere
at some point so we know when we need this token and where to get it from.

@jml: mentor review please

@jongiddy: land as you wish assuming @jml approves.

Thanks,

James

jml:
:+1: Thanks.

PR #2440, opened 2016-01-11
[FLOC-3840] Add `--process-dependency-links` to the release `pip install` step

zubron:
To install our forked version of testtools, we need to specify `--process-dependency-links` when performing `pip install`. This flag was missing from the script which sets up the release environment.


jml:
LGTM. Thanks!

PR #2441, opened 2016-01-11
[FLOC-3781] Replace broken ActionNeeded wakeup placeholder with polling

itamarst:
*How we got here:*

When the convergence loop is sleeping and gets an update of configuration or cluster state it runs calculate_changes on the deployer with cached local state to see if it should wake up. If a remote node has a volume we want attached, we would return ActionNeeded to indicate we wanted to wake up. Originally ActionNeeded was never run.

After Tom refactored the block device deployer, ActionNeeded ended up actually being run, which resulted in an exception in the logs. Moreover, the ability to sleep until the remote node detached volume we're supposed to move locally went away (temporarily, it'll return eventually). So the choices were either "sleep for 60 seconds, then check if remote detached" or "poll every 0.1 seconds" which hits the cloud API pretty hard. The latter was what was happening in practice.

Thus we were getting exceptions in logs and polling the cloud API 10x a second while waiting for remote volumes to detach.

*My solution:*

1. Refactor convergence loop/IDeployer interaction so the sleep time is determined by a returned `NoOp`, rather than being a binary "sleep for fixed amount / wake up real soon". This is a generally useful change since it allows deployers more flexibility.
2. For BlockDevice deployer specifically, return a longer sleep for the general "nothing to do", and a shorter sleep for the situation where we want to poll more frequently to see if volume is now attachable (replacing the need for ActionNeeded). This bit is an expedient fix until we split up the block device agent into a local agent and a cloud-interaction agent.

### DIFF
--- a/flocker/node/_change.py
+++ b/flocker/node/_change.py
@@ -14,6 +14,8 @@ providers.
 changes.
 """
 
+from datetime import timedelta
+
 from zope.interface import Interface, Attribute, implementer
 
 from pyrsistent import PVector, pvector, field, PClass
@@ -108,7 +110,7 @@ class _InParallel(PClass):
         ))
 
 
-def in_parallel(changes):
+def in_parallel(changes, sleep_when_empty=timedelta(seconds=60)):
     """
     Run a series of changes in parallel.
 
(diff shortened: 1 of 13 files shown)


flocker/node/_change.py
```diff
@@ -117,14 +119,19 @@ def in_parallel(changes):
...
-        ``NoOp``.
+        ``NoOp``. In former case sleep will be ``sleep_when_empty``, in
+        latter the minimum sleep of the ``NoOp`` instances.
     """
-    if all(c == NoOp() for c in changes):
-        return NoOp()
+    if all(isinstance(c, NoOp) for c in changes):
+        sleep = (min(c.sleep for c in changes) if changes
+                 else sleep_when_empty)
+        return NoOp(sleep=sleep)
```
jml: :question: alternatively, could add NoOp(sleep=sleep_when_empty) to the list of changes and spare the `if`.

jml:
LGTM. Thanks.

sarum90:
Just passing by, wanted to say the change looks good, I believe this will fix the issue we were having.

(If you wanted to manually test, starting flocker with a loopback driver and then re-installing flocker is a way to force this state.)

Don't bother blocking this PR on the following, but something to think about / what are your thoughts on the following design:

One comment about the use of "isinstance(NoOp)". Since NoOp implements `IStateChange` already, I probably would have just extended the `IStateChange` interface to have a sleep_duration() method, and have that return 0 for most `IStateChange`s. Then most of the blocks that are presently: `if isinstance(change, NoOp):` would turn into `if change.sleep_duration() > 0:`.

I _think_ that's a little easier to reason about because you no longer have to think about "Does the `IStateChange` that I have actually implement NoOp", and instead you just have to reason about "How much time does the `IStateChange` provider that I have think I have to wait for the next change to be ready".  It might just be me, but it doesn't seem intuitive to me that `PollUntilAttached` creates a `NoOp`, but it does seem intuitive that it would create something that had some `sleep_duration`.

Anyway, just food for thought, although it would be a rather annoying think to implement. (Running around to all of the IStateChanges and adding a new method).

PR #2442, opened 2016-01-11
[FLOC-3842] Remove unnecessary git pull from initialize-release script

zubron:
Fixes [FLOC-3842](https://clusterhq.atlassian.net/browse/FLOC-3842).

After initialising the release flocker repository, the remote refs are fetched. The first available ref is pulled and the local `master` branch is updated to this point. If the first available ref is not `master`, this can result in a merge conflict when `origin master` is pulled later. The first pull is not necessary. We can simply fetch, checkout `master` and then pull `origin master`.


dwgebler:
:+1: LGTM

zubron:
Amazingly fast review - thanks @dwgebler! :smile:

PR #2443, opened 2016-01-12
[FLOC-3850] cluster cleanup script

avg-I:
A script to clean up all containers and datasets in a cluster.

### DIFF
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -53,6 +53,14 @@ def main(reactor, args, base_path, top_level):
 
 @inlineCallbacks
 def cleanup_cluster(client, timeout):
+    """
+    Delete all containers and datasets in the given cluster.
+
+    :param FlockerClient client: The API client instance for the cluster.
+    :param int timeout: A timeout in seconds for waiting until the deletions
+        take effect.
+    :returns: Deferred that fires when the clean up is complete.
+    """
     containers = yield client.list_containers_configuration()
     results = []
     for container in containers:
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -27,7 +27,6 @@ class ScriptOptions(Options):
             raise UsageError("Certificates directory must be provided.")
 
 
-@inlineCallbacks
 def main(reactor, args, base_path, top_level):
     try:
         options = ScriptOptions()
@@ -46,7 +45,11 @@ def main(reactor, args, base_path, top_level):
     user_key = certificates_path.child(b"user.key")
     client = FlockerClient(reactor, options['control-node'], REST_API_PORT,
                            cluster_cert, user_cert, user_key)
+    return cleanup_cluster(client)
 
+
+@inlineCallbacks
+def cleanup_cluster(client):
     containers = yield client.list_containers_configuration()
     results = []
     for container in containers:
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -15,9 +15,9 @@ from flocker.control.httpapi import REST_API_PORT
 class ScriptOptions(Options):
     optParameters = [
         ['control-node', None, None,
-         "The address of the cluster's control node"],
+         "The address of the cluster's control node."],
         ['cert-directory', None, None,
-         "The directory containing the cluster certificates"],
+         "The directory containing the cluster certificates."],
         ['timeout', None, 300,
          "The timeout in seconds for waiting until the operation is complete",
          int],
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -1,3 +1,4 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
 
 import sys
 
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -69,7 +69,7 @@ def cleanup_cluster(client, timeout):
     yield gather_deferreds(results)
 
     def containers_deleted():
-        d = client.list_containers_configuration()
+        d = client.list_containers_state()
         d.addCallback(lambda containers: not containers)
         return d
 
@@ -83,8 +83,8 @@ def cleanup_cluster(client, timeout):
     yield gather_deferreds(results)
 
     def datasets_deleted():
-        d = client.list_datasets_configuration()
-        d.addCallback(lambda datasets: not datasets.datasets)
+        d = client.list_datasets_state()
+        d.addCallback(lambda datasets: not datasets)
         return d
 
     yield loop_until(client._reactor, datasets_deleted, repeat(1, timeout))
--- a/admin/cluster_cleanup.py
+++ b/admin/cluster_cleanup.py
@@ -1,0 +1,76 @@
+
+import sys
+
+from itertools import repeat
+
+from twisted.internet.defer import inlineCallbacks
+from twisted.python.filepath import FilePath
+from twisted.python.usage import Options, UsageError
+
+from flocker.apiclient import FlockerClient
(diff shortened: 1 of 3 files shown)


admin/cluster_cleanup.py
```diff
@@ -0,0 +1,76 @@
...
+        sys.stderr.write(options.getUsage())
+        raise SystemExit(1)
+
+    certificates_path = FilePath(options['cert-directory'])
+    cluster_cert = certificates_path.child(b"cluster.crt")
+    user_cert = certificates_path.child(b"user.crt")
+    user_key = certificates_path.child(b"user.key")
+    client = FlockerClient(reactor, options['control-node'], REST_API_PORT,
+                           cluster_cert, user_cert, user_key)
+
```
jongiddy: Put the code below this point into a separate function, as it is independently useful outside of the script.

admin/cluster_cleanup.py
```diff
@@ -0,0 +1,76 @@
...
+        print "deleting container", container.name
+        results.append(client.delete_container(container.name))
+    yield gather_deferreds(results)
+
+    def containers_deleted():
+        d = client.list_containers_configuration()
+        d.addCallback(lambda containers: not containers)
+        return d
+
+    yield loop_until(client._reactor, containers_deleted, repeat(1, 300))
```
jongiddy: Put the timeout into a constant (or possibly a script option?).

jongiddy:
Please add some docs to `gettinginvolved/cluster-setup.rst` on cleaning up a cluster. Include a summary of the options, and the information that this script clears the cluster configuration but does not wait until the agents have actually removed the containers and datasets.

jongiddy:
:question: Possibly add a `--wait` option that does wait for the containers and datasets to be removed.

admin/cluster_cleanup.py
avg-I: If I make the timeout a parameter of the suggested new function, should I still create a constant for it or can I just do something like `cleanup_cluster(client, 300)` ?

admin/cluster_cleanup.py
avg-I: Okay.

avg-I:
Regarding `--wait`: I think that we have to wait for the containers to be removed before attempting to remove the datasets otherwise the latter could fail. So the option would control only whether we wait for the dataset removal to complete...

avg-I:
And right now the script actually does wait until all the deletions are done,

avg-I:
Actually I could be wrong about that. Considering `list_datasets_configuration` vs `list_datasets_state` as an example, does the former report a projection of the desired state while the latter does that for the actual (converged) state? If yes, should we use `list_datasets_state` instead?

admin/cluster_cleanup.py
jongiddy: Any numbers that are fixed in the source code, but could sensibly change to another value, should be constants.

jongiddy:
Yes, the code currently just confirms that the configuration is changed. If it should wait until everything is actually deleted, you need to use `list_containers_state` and `list_datasets_state` instead. You should check both because there may be containers with no attached datasets, or datasets that are not attached to containers.

avg-I:
Okay!

jongiddy:
I haven't tested it fully on large clusters, but I'll put it back in Coding so you can address the above issues.

avg-I:
I've submitted the updated code, but now a thought has occurred to me: perhaps with `--wait` option the script should wait indefinitely?

wallrj:
Sorry to chip in but there is already the `flocker.acceptance.testtools.Cluster` class which has a `clean_nodes` method.
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/testtools.py#L694

Re-use that?

avg-I:
Not at all! Thank you for pointing to that code. It's impossible to know about all the treasures in our code base, although I should have known that the tests should have some clean-up code.
I see that that code uses "docker client", I am not sure what it is, so it it okay?
We used `FlockerClient` interface here, not sure if there's much difference.

jml:
Chiming in, but @jongiddy is the one doing the review, so please feel free to disregard if this only makes things harder, not easier.

`Cluster` uses docker client to remove containers that are not managed by Flocker. At a guess, this is to make sure we are _really_ clean, even if our code is buggy.

I'd encourage you to re-use `Cluster`. If there's no difference, then it should be easy enough to switch to it. If there is an important difference, then it would good for that difference to be represented in code.

avg-I:
@jml, thank you for the explanation!
It's a little bit confusing that `Cluster` uses all of docker client, `FlockerClient` and direct HTTP requests.
Also, creating a `Cluster` object is not that easy comparing to a `FlockerClient` object.  For example, the `Cluster` requires a list of nodes to be provided to it (why?). I am looking at `_get_test_cluster` and it's about the same size as the whole `cluster_cleanup.py`.

jml:
Doing a quick grep for `self.nodes` in `Cluster`, it looks like it's only used for by the code that runs the docker client clean up as described earlier.

Interestingly, it is passed as an empty list in `_get_test_cluster`—maybe all of that code is now unused?

Likewise, treq is only used by `current_nodes`, which isn't used by `clean_nodes`.

So, maybe something like this?

```python
    cluster = Cluster(
        control_node=ControlService(public_address=control_node),
        nodes=[],
        treq=None,
        client=FlockerClient(reactor, control_node, REST_API_PORT,
                             cluster_cert, user_cert, user_key),
        certificates_path=certificates_path,
        cluster_uuid=user_credential.cluster_uuid,
    )
```

Possibly even extract a helper function/classmethod convenience for creating `Cluster` with some sensible defaults (e.g. it takes the control_node address, the keys, and constructs the client & the `ControlService` before passing into `Cluster`)?

I don't want to block you with endless bikeshedding, so reiterating that it's OK for you to ignore me here :)

avg-I:
Well, I do not see this as bikeshedding but rather me learning about the code that we already have, so I do not mind :)
It seems that `treq` is actually used by  `clean_nodes`, but that's not a problem, it's easy to set it.

I see two related issues with `clean_nodes`:
- there is no way to say that I do not care to wait for the flocker cluster to converge after the configuration changes
- there are a few timeouts hardcoded in `clean_nodes` and they are rather short for the kind of configurations with we use for benchmarks

Both problems are easily solvable, but I am not sure if that's something that should be done now.

admin/cluster_cleanup.py
```diff
@@ -0,0 +1,102 @@
+
```
jongiddy: Add copyright header and a module docstring.

avg-I: Can't believe I missed that

jongiddy:
Regarding the acceptance test cluster code, that code is specific to the structure of an acceptance testing cluster, which may not reflect the structure of a benchmark cluster.  In particular, the required firewall settings and time-outs are potential issues. However, the code does bring up the question of whether leases should be included in the clean-up. We don't currently use them in the benchmarks, so maybe that can wait until we do.

jongiddy:
:question: Another thought is that since this script is meant to be used for benchmark clusters, it could live in the `benchmark` directory and take advantage of the `BenchmarkCluster` class, sharing some options with the `benchmark` script to specify the cluster.

jongiddy:
The code does the job. It may be useful to find some commonality with the acceptance testing code or with the benchmark code, but that would make this a much bigger PR. I think it is OK as is, and once we have some more experience, we might do the extra work.

Merge when you are happy. If you make any major changes, feel free to put it up for further review.

avg-I:
Okay, thank you, everyone, for your reviews, suggestions and ideas. I am merging this now, but I sure want to refactor the code later.

PR #2444, opened 2016-01-12
[FLOC-3854] AWS session token support for backend

jongiddy:
In FLOC-3839, support for AWS session tokens was added to the `aws` stanza used to start cluster nodes.

However, the nodes themselves use AWS for managing volumes, so the AWS backend also needs to use session tokens where appropriate.

Session tokens typically only last for an hour, but Flocker agents are likely to last longer than that. So, under the circumstances where they are needed, session tokens will need to be generated by the backend.  FLOC-3857 has been created to solve this.

In the meantime, this PR uses the same solution from FLOC-3839 for short-lived tests.

### DIFF
--- a/docs/config/aws-configuration.rst
+++ b/docs/config/aws-configuration.rst
@@ -20,6 +20,20 @@ The configuration item to use AWS should look like:
 Make sure that the ``region`` and ``zone`` match each other and that both match the region and zone where the Flocker agent nodes run.
 AWS must be able to attach volumes created in that availability zone to your Flocker nodes.
 
+In addition to the mandatory properties shown, there are some optional properties:
+
+.. option:: session_token
+
+   An AWS session token.
+   This allows cross-account access.
+   It is mainly useful for testing since session tokens only last for a short time.
+
+.. option:: validate_region
+
+   Boolean indicating whether to validate the supplied region.
+   This defaults to True.
+   It is set to False for testing purposes.
+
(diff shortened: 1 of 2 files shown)


docs/config/aws-configuration.rst
```diff
@@ -20,6 +20,20 @@ The configuration item to use AWS should look like:
 Make sure that the ``region`` and ``zone`` match each other and that both match the region and zone where the Flocker agent nodes run.
 AWS must be able to attach volumes created in that availability zone to your Flocker nodes.
 
+In addition to the mandatory properties shown, there are some optional properties:
+
+.. option:: session_token
+
+   An AWS session token.
+   This allows cross-account access.
+   It is mainly useful for testing since session tokens only last for a short time.
```
james-w: Is it possible to say how you will know you will need this, or where to get one if you do?

james-w:
:+1: Looks good, thanks for finding a small solution that will allow work to proceed
without baking in a lot of things to unpick later. Merge at will assuming that
jml agrees.

@jml mentor review please.

docs/config/aws-configuration.rst
```diff
@@ -20,6 +20,20 @@ The configuration item to use AWS should look like:
...
+
+.. option:: validate_region
+
+   Boolean indicating whether to validate the supplied region.
+   This defaults to True.
+   It is set to False for testing purposes.
```
jml: :question: Maybe elaborate on what those purposes are?

jml:
Agreed. Thanks @jongiddy & @james-w

PR #2445, opened 2016-01-12
[FLOC-3684] Get rid of deperecated `ClusterStateService.apply_changes`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3864

### DIFF
--- a/flocker/apiclient/test/test_client.py
+++ b/flocker/apiclient/test/test_client.py
@@ -770,7 +770,8 @@ class FlockerClientTests(make_clientv1_tests()):
                                         in node.manifestations.values()},
                                  devices={})
                        for node in deployment.nodes]
-        self.cluster_state_service.apply_changes(node_states)
+        self.cluster_state_service.apply_changes_from_source(
+            ChangeSource(), node_states)
 
     def get_configuration_tag(self):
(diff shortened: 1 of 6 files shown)


wallrj:
Thanks @tomprince 

Looks good to me but:

 * There was a single failed build. I restarted it.
 * :question: With the state era endpoint do we still need the extra complication of ChangeSource's?
   * ChangeSource was introduced to record the timestamp of each state update and to allow garbage collection of stale state reported by now-offline nodes.
   * but now we have `era`s we have a better way, right?
   * Mightn't we want to re-instate the deprecated but simplified API?
 * :question: I thought we said that Flocker isn't a library. In which case, why did we bother carefully deprecating that API?

Anyway, merge if you're sure this is the right direction to go and when the builds are all green.

tomprince:
>  I thought we said that Flocker isn't a library. In which case, why did we bother carefully deprecating that API?

I suspect to avoid having to change test code at the time.

> With the state era endpoint do we still need the extra complication of ChangeSource's? 

I think we still want to get rid of stale data. The era takes care of invalidating data from a node that has restarted, but doesn't handle the case when an agent has died.

PR #2446, opened 2016-01-12
[FLOC-3822] Implement the shadow mount directories and verify they solve the underlying issue.

sarum90:
N.B. This change is not against master. This just verifies that the approach does fix the underlying problem. I'd like to have it reviewed before I run extensive upgrade tests on it.

The approach here is to separate the "mount directory" from the "shared path" within flocker. This enables flocker to mount datasets in one directory, but use a different directory when handing paths out to flocker clients.

The branch also adds a new concept of a shadow_tmpfs_mount, which is a:

* tmpfs mount mounted read-only and made into a shared directory.
* This mount is bound to a different location while it is read only.
* The original mount is then remount read-write.

This sequence results in a directory that can be modified (some secret path that only flocker knows about), and a second read-only directory that is backed by the same tmpfs mount. This enables flocker to add mounts to the secret path, and have them reflected in the read only path, thus enabling only flocker to add mounts to a directory by virtue of only flocker knowing the secret path.

This branch then uses that new concept to verify that this protects against docker creating bind mounts to paths that flocker is not actively backing with a volume.


flocker/node/agents/blockdevice.py
```diff
@@ -1575,7 +1586,7 @@ def _discover_raw_state(self):
         compute_instance_id = api.compute_instance_id()
         volumes = api.list_volumes()
         system_mounts = {
-            mount.blockdevice: mount.mountpoint
+            mount.mountpoint: mount.blockdevice
```
sarum90: I had to flip this because on some systems (like the vagrant dev box) bind mount actually causes the blockdevice to be mounted in two directories:

```
$ cd /tmp
$ mkdir a b
$ mount --bind a b
$ mkdir a/c
$ mount -t tmpfs tmpfs a/c
$ mount
....
tmpfs on /tmp/a/c type tmpfs (rw,relatime)
tmpfs on /tmp/b/c type tmpfs (rw,relatime)
```

flocker/node/agents/test/test_shadow_bind_mounter.py
```diff
@@ -0,0 +1,97 @@
...
+from twisted.python.filepath import FilePath
+
+from ....testtools import TestCase, run_process, _CalledProcessError
+from ...testtools import if_docker_configured
+from ..testtools import blockdevice_manager_for_test
+
+from ..shadow_bind_mounter import create_tmpfs_shadow_mount
+
+
+def write_as_docker(directory, filename, content):
```
sarum90: Move to common location and use in both tests that want this functionality.

flocker/node/agents/testtools.py
```diff
@@ -58,3 +96,156 @@ def setUp(self):
...
+        Perform all cleanup operations.
+        """
+        cleanup_errors = []
+        for operation in reversed(self._cleanup_operations):
+            try:
+                operation.execute(self.original)
+            except CleanupError as e:
+                cleanup_errors.append(e)
+        if cleanup_errors:
+            raise cleanup_errors[0].error
```
jml: :question: This generic logic is also in https://pypi.python.org/pypi/fixtures, on which we already depend. Maybe use that if you feel like it would help.

flocker/node/agents/blockdevice.py
```diff
@@ -90,6 +90,8 @@ class DiscoveredDataset(PClass):
         the node where the dataset is attached.
     :ivar FilePath mount_point: The absolute path to the location on the node
         where the dataset will be mounted.
+    :ivar FilePath share_path: The absolute path to the location on the node
+        where the dataset is available for use by an external client.
```
jml: I think it would be helpful to give an example, or perhaps to say something about how `share_path` relates to `mount_point`.

flocker/node/agents/blockdevice.py
```diff
@@ -1505,6 +1511,10 @@ class BlockDeviceDeployer(PClass):
         profiles.
     :ivar FilePath mountroot: The directory where block devices will be
         mounted.
+    :ivar FilePath sharedroot: The parent directory for all dataset paths that
+        will be shared externally. Defaults to being the same as mountroot.
+        This is designed to either be the mountroot or be a directory that is
+        bind mounted read only to the mountroot.
```
jml: :question: Not introduced in this patch, but I wish there were an underscore before `root`.

flocker/node/agents/blockdevice.py
```diff
@@ -1631,9 +1642,12 @@ def discover_state(self, node_state):
...
                     unicode(dataset_id)
                 )
+                share_path = self._sharepath_for_dataset_id(
+                    unicode(dataset_id)
+                )
                 if (
-                    device_path in raw_state.system_mounts and
-                    raw_state.system_mounts[device_path] == mount_point
+                    mount_point in raw_state.system_mounts and
+                    raw_state.system_mounts[mount_point] == device_path
```
jml: I'd personally appreciate a comment saying what this condition means.

flocker/node/agents/shadow_bind_mounter.py
```diff
@@ -0,0 +1,31 @@
...
+    Creates a "shadow mount directory". This is a tmpfs mount that is bind
+    mounted read-only to another location. The intent is to have the following
+    characteristics:
+
+        - The read_only_directory is cleared on reboots.
+        - The read_only_directory can be changed only by agents that know the
+            path of the backing directory.
+
+    :param FilePath backing_directory: The directory that will be created if
+        need be to back the shadow mount. This will be read write.
```
jml: Nit: I think split the first sentence in two: "The directory that backs the shadow mount. Will be created if it doesn't exist. This will be read write."

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -1308,7 +1310,7 @@ def _has_file_with_content(self, filename, content):
             test_mountpoint.remove()
         return result
 
-    def invariant(self, case, current_local_state):
+    def invariant(self, case, test_objects):
         """
         An invariant that should always hold. Specifically, after the path to a
```
jml: :question: I think this first sentence is redundant. Maybe delete it. Maybe rename from `invariant` (which doesn't help), to `writes_persist_or_fail`.

OTOH, this patch is already rather large, so maybe don't bother.

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -257,3 +261,135 @@ def test_make_tmpfs_mount_failure(self):
...
+
+    def test_mount_cleaned_up(self):
+        """
+        The cleanup implementation cleans up mounts.
+        """
+        blockdevice = self._get_free_blockdevice()
+        self.manager_under_test.make_filesystem(blockdevice, 'ext4')
+        mountdir = self._get_directory_for_mount()
+        self.manager_under_test.mount(blockdevice, mountdir)
+        self.assertEqual(len(self.manager_under_test._cleanup_operations), 1)
```
jml: :question: 

```python
self.assertThat(self.manager_under_test._cleanup_operations, HasLength(1))
```

Will show you the list if the assertion fails.

flocker/node/agents/test/test_shadow_bind_mounter.py
```diff
@@ -0,0 +1,75 @@
...
+        ro_dir.makedirs()
+        content = unicode(uuid4())
+        filename = unicode(uuid4())
+        write_file = backing_directory.child(filename)
+        read_file = ro_dir.child(filename)
+        create_tmpfs_shadow_mount(backing_directory,
+                                  ro_dir,
+                                  self._blockdevice_manager)
+        write_file.setContent(content)
+        self.assertEquals(read_file.getContent(), content)
```
jml: No action required.

I think I'm going to add some matchers.

flocker/node/agents/testtools.py
```diff
@@ -58,3 +96,156 @@ def setUp(self):
...
+    Creates a blockdevice_manager that cleans itself up during test cleanup.
+
+    Cleanup is defined as unmounting all bind mounts, tmpfs mounts, and
+    blockdevice mounts.
+
+    :param test_case: The :class:`TestCase` to use to add cleanup callbacks.
+    """
+    manager = CleanupBlockDeviceManager(BlockDeviceManager())
+    test_case.addCleanup(manager.cleanup)
+    return manager
```
jml: :question: If you made `CleanupBlockDeviceManager` subclass `Fixture`, then this function wouldn't be necessary, as tests could just do:

```python
manager = self.useFixture(CleanupBlockDeviceManager(BlockDeviceManager())
```

But, tbh, not convinced it's worth it.

flocker/node/agents/shadow_bind_mounter.py
```diff
@@ -0,0 +1,31 @@
...
+def create_tmpfs_shadow_mount(backing_directory, read_only_directory,
+                              blockdevice_manager):
+    """
+    Creates a "shadow mount directory". This is a tmpfs mount that is bind
+    mounted read-only to another location. The intent is to have the following
+    characteristics:
+
+        - The read_only_directory is cleared on reboots.
+        - The read_only_directory can be changed only by agents that know the
+            path of the backing directory.
```
jml: It's worth adding a 'so that' for each of these characteristics, explaining why each is desirable. (I can guess the second, but not the first)

flocker/node/agents/shadow_bind_mounter.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def create_tmpfs_shadow_mount(backing_directory, read_only_directory,
+                              blockdevice_manager):
+    """
+    Creates a "shadow mount directory". This is a tmpfs mount that is bind
+    mounted read-only to another location. The intent is to have the following
+    characteristics:
+
+        - The read_only_directory is cleared on reboots.
```
jml: It doesn't look like we test this. Should we? (Or have I missed it).

jml:
Thanks! This looks like a really interesting approach.

After reviewing and understanding the patch, I started to wonder whether this is two separate changes:
  1. shared_mount
  2. shadow mounting

Is this right? (Mostly just want to confirm my understanding).

I think the shared mount change is a good one.

I'm less qualified to judge re the shadow mounting, mostly because I'm not super familiar with the problem being addressed, and I have only a surface understanding of tmpfs. I guess the big risks are:

  1. we accidentally try to store data on the tmpfs
  2. we behave in undefined ways on reboot

Are these valid? Do we try to reduce them in the patch?

As far as I'm concerned, I'm happy for you to go ahead with these changes, so am marking Approved. I'd completely understand if you want someone with deeper domain knowledge to take a look.

flocker/node/agents/testtools.py
sarum90: That's pretty cool, but unfortunately I want to be able to remove cleanup operations, and it doesn't look like fixtures help with that very much? Let me know if I'm reading that wrong

https://github.com/testing-cabal/fixtures/blob/master/fixtures/fixture.py#L77 seems to suggest they don't.

I'll fix up `blockdevice_manager_for_test` to return a fixture rather than take a test case though. I think that makes sense.

flocker/node/agents/blockdevice.py
sarum90: Added brief note.

flocker/node/agents/blockdevice.py
sarum90: done. luckily sharedroot was only used for this, so sed could fix that pretty easily.

flocker/node/agents/blockdevice.py
sarum90: Done

flocker/node/agents/shadow_bind_mounter.py
sarum90: Fixed.

flocker/node/agents/test/test_blockdevice_manager.py
sarum90: Perfect, thanks.

flocker/node/agents/test/test_shadow_bind_mounter.py
sarum90: :) please do.

flocker/node/agents/testtools.py
sarum90: I actually did this using the MethodFixture. I like the idea of using fixtures rather than passing in test_cases just to add cleanups.

flocker/node/agents/shadow_bind_mounter.py
sarum90: Added.

flocker/node/agents/shadow_bind_mounter.py
sarum90: Added a test that verifies that it is cleared on unmount. not worth rebooting to verify :)

sarum90:
Correct about this actually being two changes. Sorry that they are intertwined. The shared_path part was added.

As far as accidentally storing data on tmpfs. I have high confidence that will not happen because of a test I have recently added (the test_docker_protection test). After _every_ system operation (method on blockdevice_manager) This attempts to write to a path that flocker would expose, and guarantees that the writes either fail or are persisted.

as far as "we behave in undefined ways on reboot". I'm still working on the system that will set up that shadow mount. In those PRs I'll be putting something together that validates the shadow mount is set up correctly, and errors out if it is not. It should also more concretely define the ways we behave on reboot and upgrade.

PR #2447, opened 2016-01-12
[FLOC-3853] Upgrade to awscli-1.9.17

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3853

Required for:
 * https://clusterhq.atlassian.net/browse/FLOC-2748 and
 * https://clusterhq.atlassian.net/browse/FLOC-3849

There were a bunch of missing transient dependencies in the dev-requirements.txt file so I added them. It may be that some of the dependencies had been deliberately omitted because they also appear in `requirements.txt` but there aren't any tools to manage these files  and the package versions in the two files were already out of sync so I just added them and updated the versions where they were different.


dev-requirements.txt
```diff
@@ -1,41 +1,60 @@
 alabaster==0.7.6
 apache-libcloud==0.17.0
-awscli==1.7.35
+awscli==1.9.17
+Babel==2.2.0
 backports.ssl-match-hostname==3.4.0.2
```
sarum90: I think the name is actually:

`backports.ssl_match_hostname`

sarum90:
So it is not clear to me when something should be in dev-requirements.txt _and_ in requirements.txt.

Do we expect anything to work if someone only installs dev-requirements.txt?

Alternatively, it's not clear when something should be in requirements.txt but not in dev-requirements.txt. I would think you would need all of the requirements.txt in order to run the tests.

Suggest adding a comment at the top of dev-requirements.txt so it is a bit more clear.

Potentially just remove packages that are in both from dev-requirements.txt rather than trying to keep them in sync. Something like #2450

sarum90:
Discussed offline, I think we'll want to do something like #2450

wallrj:
Thanks @sarum90 

> Do we expect anything to work if someone only installs dev-requirements.txt?

I've changed my mind. Yes, I was expecting to be able to `pip install -r dev-requirements.txt` in a fresh virtualenv and run the acceptance tests.

In practice, that's not possible because:
 *  dev-requirements.txt doesn't contain Twisted etc
 * `pip install -r dev-requirements.txt` doesn't support the `--process-dependency-links` option because the dependency link is in setup.py.
 * Probably other reasons....

> Suggest adding a comment at the top of dev-requirements.txt so it is a bit more clear.

It's a good idea, but it's still not clear to me whether dev-requirments.txt should contain transient dependencies and whether those transient dependencies should be duplicated if they're already present in requirements.txt.

I found the script that @tomprince originally wrote to manage these files.
 * https://github.com/ClusterHQ/flocker/blob/49efb6d85ce3e5cebae886a55b3c8310746949e1/admin/requirements.py

And an issue to finish that script:
 * https://clusterhq.atlassian.net/browse/FLOC-2528

Anyway, this is all a can of worms.
The branch now has the minimum necessary change which allows me to run the cloudformation test in #2449

sarum90:
Looks good. I still might go in and clean these up sometime, cause it'd be better if it was more obvious what should be in here, but for now please proceed.
Thanks!

wallrj:
Thanks @sarum90 

I re-ran the lint and run_trial_on_AWS_Ubuntu_Trusty_as_root_flocker.node.agents builders. All  are now green so I'll merge this.

PR #2449, opened 2016-01-12
[FLOC-3849] Test cloudformation installer

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3849

So far we've written a single end-to-end test which boots a cloudformation stack based on configuration published at:
 * https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json

Then runs the demo compose files on the client machine to create and then move a PostgreSQL database container and then verifies that data was moved.

On Slack / Engineering meeting it was suggested that this was probably too fragile and that we should be testing smaller parts of the cloudformation installation process.
We need to discuss that further.


flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+        lambda process_result: (process_result, docker_compose_output)
+    )
+    return d
+
+
+def remote_postgres(client_ip, host, command):
+    postgres_output = []
+    d = deferLater(
+        reactor,
+        60,
```
jml: Why 60? Maybe add a comment or define a constant or both. 

If it's because something else needs to happen first, then instead of waiting a fixed length of time, we should only try after that condition is fixed (otherwise we're just adding more flaky tests). `flocker.common.retry` has a bunch of helpers for this.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+        else:
+            Message.new(
+                message='Waiting for stack target status',
+                stack_id=stack_id,
+                target_status=target_status,
+                current_status=current_status,
+                time_running=time_running,
+            ).write()
+
+            time.sleep(10)
```
jml: Please use on of the helpers in `flocker.common.retry`. Feel free to tweak them if they aren't directly suitable.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+        self.docker_host = self.control_node_ip + ':2376'
+        local_certs_path = self.mktemp()
+        check_call(
+            ['scp', '-o', 'StrictHostKeyChecking no', '-r',
+             'ubuntu@{}:/etc/flocker'.format(self.client_ip),
+             local_certs_path]
+        )
+        self.addCleanup(cleanup, self, local_certs_path)
+        d = maybeDeferred(super(DockerComposeTests, self).setUp)
+        return d
```
jml: Just returning the result of upcalling `setUp` is enough. No reason to wrap in `maybeDeferred`.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+         '--command={}'.format(command)),
+        handle_stdout=postgres_output.append
+    )
+    d.addCallback(
+        lambda process_result: (process_result, postgres_output)
+    )
+    return d
+
+
+def cleanup(test_case, local_certs_path):
```
jml: This is pretty intimately tied up with the test case. I'd strongly consider either:

* making it a method
* making individual `addCleanup` calls as appropriate (e.g. after a call to `docker up`, add a cleanup to bring it back down)
* passing the attributes as parameters
* passing the attributes of `test_case`

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+        nodes=[],
+        treq=treq_with_authentication(
+            reactor, cluster_cert, user_cert, user_key),
+        client=FlockerClient(
+            reactor, test_case.control_node_ip.encode("ascii"),
+            REST_API_PORT, cluster_cert, user_cert, user_key
+        ),
+        certificates_path=certificates_path,
+        cluster_uuid=user_credential.cluster_uuid,
+    )
```
jml: :question: from this branch and #2443 it looks like `Cluster` could really use a convenience method/function for constructing.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+    check_call(
+        ['aws', 's3', 'rb', 's3://{}'.format(s3_bucket_name), '--force']
+    )
+
+    check_call(
+        ['aws', '--region', REGION, 'cloudformation', 'delete-stack',
+         '--stack-name', stack_id]
+    )
+
+    return wait_for_stack_status(stack_id, 'DELETE_COMPLETE')
```
jml: :question: These could be combined into a `Fixture` object which could then be used with `TestCase.useFixture`, which would automatically add a cleanup. I've found the abstraction useful in the past, but am not sold enough on it to recommend it without reservation.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+        self.addCleanup(cleanup, self, local_certs_path)
+        d = maybeDeferred(super(DockerComposeTests, self).setUp)
+        return d
+
+    def test_docker_compose_up_postgres(self):
+        """
+        """
+        d = remote_docker_compose(
+            self.client_ip, self.docker_host, COMPOSE_NODE0, 'up', '-d'
+        )
```
jml: :question: Maybe add a helper method `remote_docker_compose` that is equivalent `partial(remote_docker_compose, self.client_ip, self.docker_host)`? Maybe even it should take an ignored first parameter to work better with callbacks?

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+    },
+    {
+        'ParameterKey': 'AccessKeyID',
+        'ParameterValue': os.environ['ACCESS_KEY_ID']
+    },
+    {
+        'ParameterKey': 'SecretAccessKey',
+        'ParameterValue': os.environ['SECRET_ACCESS_KEY']
+    }
+]
```
jml: Is this the normal way we use environment parameters in tests? We don't fail nicely if they are not set or document that they are necessary?

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,319 @@
...
+
+COMPOSE_NODE0 = '/home/ubuntu/postgres/docker-compose-node0.yml'
+COMPOSE_NODE1 = '/home/ubuntu/postgres/docker-compose-node1.yml'
```
jml: :question: Maybe worth adding a comment that this installer only works on Ubuntu for now.

jml:
Thanks for testing this. I also would like to see more, smaller tests, but don't see why that excludes a stable, unfragile acceptance test. I would like to see evidence of its non-fragility before it lands in master please.

I've made a few comments about various things, and would like a chance to look again before this lands.

wallrj:
Ready for another review.

Followup work:

1. Build the cloudformation template and run the test automatically from Jenkins
2. Modify the test to reflect the documented moby-docker application.

flocker/acceptance/endtoend/test_installer.py
```diff
@@ -0,0 +1,507 @@
...
+
+        def set_stack_variables(stack_report):
+            outputs = stack_report['Outputs']
+            self.stack_id = stack_report['StackId']
+            for variable_name, stack_output_name in STACK_VARIABLES.items():
+                setattr(
+                    self, variable_name, get_output(outputs, stack_output_name)
+                )
+            if 'KEEP_STACK' not in os.environ:
+                self.addCleanup(delete_cloudformation_stack, self.stack_id)
```
jml: No need for `stack_id` to be an instance variable, since it's not used.

jml:
Thanks!

It's hard to tell exactly, but it looks like you're getting a sphinx failure that might be genuine. Could you please run the sphinx tests locally on your branch, verifying that they pass before merging. Please also address above comments.

wallrj:
Since review, I:
* Updated to accomodate changes to the output names
* Changed the way the compose templates are uploaded so that they don't clash on successive runs.
* Added support for the new Volume Hub Token parameter
* reran some failed builders until they passed. 

So I'll merge this now.

PR #2450, opened 2016-01-12
Remove dev-requirements.txt in requirements.txt

sarum90:
Just a thought for #2447


wallrj:
Thanks @sarum90 

I updated the parent branch with a much simpler change.

PR #2453, opened 2016-01-13
[FLOC-3869] Fix lint problems

jongiddy:
These lint problems are mostly:
1. imports appearing after other code
2. assigned lambdas
3. operators appearing after newline rather than before.

Most of the fixes are to shuffle code around a bit, and convert lambdas to functions.

In `flocker/node`, there was a circular import which worked previously because one of the imports was after the definition of an Exception object.  The exception was moved to its own file to break the circular import.

### DIFF
--- a/flocker/testtools/_flaky.py
+++ b/flocker/testtools/_flaky.py
@@ -77,7 +77,7 @@ def _get_flaky_annotation(case):
     return getattr(method, _FLAKY_ATTRIBUTE, None)
 
 
-def _get_invariants(x):
+def _flaky_invariants(x):
     return (
         (x.max_runs >= x.min_passes, "Can't pass more than we run"),
         (len(x.jira_keys) > 0, "Must provide a jira key"),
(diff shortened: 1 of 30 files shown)


flocker/testtools/_flaky.py
```diff
@@ -77,6 +77,13 @@ def _get_flaky_annotation(case):
     return getattr(method, _FLAKY_ATTRIBUTE, None)
 
 
+def _get_invariants(x):
```
jml: Please rename to `_flaky_invariants`.

jml:
Thanks! It's regrettable that you've lost your morning to this, and that @wallrj and @justnoise have lost a day of momentum. 

FWIW, I personally disagree with all of the "new" rules, but think (at least for these) that we should just lump it.

Please merge when the tests pass.

jongiddy:
@jml Now that the lint test is green, should we add an issue to disable some checks? The restriction against lambda assignment, in particular, seems obnoxious, especially for class attributes.

jml:
Yeah, I think so. lambda assignment for sure and maybe also the `__all__` / `import` thing.

PR #2454, opened 2016-01-13
Rename jobs.groovy.j2 to jobs.groovy now it doesn't use jinja.

james-w:
In preparation for dropping the jinja translation step
we rename jobs.groovy.j2 to jobs.groovy so that it would be
found by the setup_ClusterHQ-flocker job config.

A symlink with the old name is kept around so that there
doesn't have to be a co-ordinated change across all
branches with Jenkins.

Once jenkins has changed to not using jobs.groovy.j2 then
the symlink can be removed.


jobs.groovy
```diff
@@ -0,0 +1,732 @@
...
+      cron-style jobs are only created for the master branch.
+
+  The full Jenkins job DSL reference can be found on the URL below:
+  https://github.com/jenkinsci/job-dsl-plugin/wiki/Job-reference
+
+  The file is consumed by the Jenkins job setup_ClusterHQ_Flocker. This file
+  is read by a build step of type 'Process Job DSL' which produces all the
+  jenkins objects.
+
+  A typical jenkins job contains the following sections, this is just an
```
james-w: The display of the diff is terrible. This file only changed above this point.

jobs.groovy.j2
```diff
@@ -1,738 +0,0 @@
-// vim: ai ts=2 sts=2 et sw=2 ft=jinja fdm=indent et foldlevel=0
```
james-w: This is a symlink, not sure why it's displayed like this.

jml:
LGTM

PR #2457, opened 2016-01-13
[FLOC-3880] Don't expect foo as end of string

dwgebler:
Fixes FLOC-3880

### DIFF
--- a/flocker/testtools/test/test_base.py
+++ b/flocker/testtools/test/test_base.py
@@ -159,7 +159,7 @@ class BaseTestCaseTests(TesttoolsTestCase):
             test.getDetails(),
             ContainsDict({
                 'twisted-log': match_text_content(MatchesRegex(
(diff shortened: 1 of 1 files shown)


flocker/testtools/test/test_base.py
```diff
@@ -159,7 +159,7 @@ def test_something(self):
             test.getDetails(),
             ContainsDict({
                 'twisted-log': match_text_content(MatchesRegex(
-                    r'^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}[+-]\d{4} \[-\] foo$'
+                    r'^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}[+-]\d{4} \[-\] foo'
```
jml: :question: Adding, `, flags=re.MULTILINE` would better preserve the intent of the test, but this is OK.

jml:
LGTM.

PR #2459, opened 2016-01-13
[FLOC-3852] add --cert-directory option to setup-cluster

avg-I:


admin/acceptance.py
```diff
@@ -357,22 +363,24 @@ def _provider_for_cluster_id(dataset_backend):
     return Providers.UNSPECIFIED
 
 
-def generate_certificates(cluster_name, cluster_id, nodes):
+def generate_certificates(cluster_name, cluster_id, nodes, cert_path):
     """
     Generate a new set of certificates for the given nodes.
 
+    :param bytes name: The name of the cluster.
```
zubron: The parameter name should be `cluster_name`.

admin/cluster_setup.py
```diff
@@ -79,6 +82,24 @@ def postOptions(self):
...
+                if not cert_path.isdir():
+                    raise UsageError(
+                        "{} is not a directory".format(self['cert-directory'])
+                    )
+                if cert_path.listdir():
+                    raise UsageError(
+                        "{} is not empty".format(self['cert-directory'])
+                    )
+            else:
+                cert_path.makedirs()
```
zubron: If `makedirs()` fails, then we should report this to the user. For example, if they don't have permission to create that directory path.

zubron:
This looks good, thanks! Just a couple of small points to address. Also, some of the admin tests are failing as they need to be updated to include the new parameters. I'm happy for this to be merged when comments are addressed and the tests are passing. This change will also need a mentor review from @jml.

admin/cluster_setup.py
jml: :question: I'd also consider extracting the body if the conditional to a separate function / method (perhaps `_ensure_empty_dir`?) that takes an arbitrary `FilePath` as a parameter.

jml:
Looks good @zubron, thanks for doing the review.

PR #2451, opened 2016-01-13
[FLOC-3710] Templates for generating Flocker AMI images with Packer.

wallrj:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3710

In this first version of the process I've documented the steps for generating the AMIs with Packer.

Next I need to figure out how to integrate this into the build / release process.

Perhaps I should aim to automatically build new AMIs on each master merge and when a release tag is generated.

TODO:
 * Figure out how to add Flocker version information to the image name or metadata.

PR #2463, opened 2016-01-14
[FLOC-3709] AWS CloudFormation installer for {Flocker, Docker, Swarm} stack

myechuri:
This branch adds scripts needed to produce a JSON template for AWS CloudFormation (https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json). 

Stack architecture designed by the set of scripts: https://github.com/ClusterHQ/flocker/blob/reorg-integrations-FLOC-3718/docs/images/cloudformation.png + Docker TLS

Supporting branches for this PR:

1) Test: https://github.com/ClusterHQ/flocker/pull/2449 .
2) Base AMI generation: https://github.com/ClusterHQ/flocker/pull/2449
3) Documentation (work in progress): https://github.com/ClusterHQ/flocker/blob/reorg-integrations-FLOC-3718/docs/docker-integration/cloudformation.rst


README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
 
 Flocker is also tested using `continuous integration`_.
 
+Cloudformation Deployment
```
sarum90: nit: Cloudformation here, but CloudFormation in all of the other files. CloudFormation looks better to me, and seems to be what Amazon uses (https://aws.amazon.com/cloudformation/)

README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
...
+-------------------------
+
+* Generate a Cloudformation template:
+
+```
+$ python ./admin/installer/cloudformation.py > flocker-cloudformation-template.json
```
sarum90: :grey_question: with these instructions, I think we should either put `flocker-cloudformation-template.json` into our .gitignore, or instead place `flocker-cloudformation-template.json` off in `/tmp/`

README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
...
+```
+
+* Validate the template:
+
+```
+$ aws cloudformation validate-template --template-body file://flocker-cloudformation-template.json
```
sarum90: when I run this, I get:
```
  File "/Users/mewert/.virtualenvs/flocker-review/bin/aws", line 27, in <module>
    sys.exit(main())
  File "/Users/mewert/.virtualenvs/flocker-review/bin/aws", line 23, in main
    return awscli.clidriver.main()
  File "/Users/mewert/.virtualenvs/flocker-review/lib/python2.7/site-packages/awscli/clidriver.py", line 48, in main
    return driver.main()
  File "/Users/mewert/.virtualenvs/flocker-review/lib/python2.7/site-packages/awscli/clidriver.py", line 173, in main
    parser = self._create_parser()
  File "/Users/mewert/.virtualenvs/flocker-review/lib/python2.7/site-packages/awscli/clidriver.py", line 154, in _create_parser
    command_table['help'] = self.create_help_command()
  File "/Users/mewert/.virtualenvs/flocker-review/lib/python2.7/site-packages/awscli/clidriver.py", line 149, in create_help_command
    cli_data.get('help_usage', None))
  File "/Users/mewert/.virtualenvs/flocker-review/lib/python2.7/site-packages/awscli/help.py", line 259, in __init__
    HelpCommand.__init__(self, session, session.provider,
AttributeError: 'Session' object has no attribute 'provider'
```

sarum90: No action required. Thanks for merging in master! I was getting _all_ the aws/botocore incompatibility errors earlier.

sarum90: I think this may still need work. (The template is too long now for verification :cry: )
```
flocker-review-3 ♦ ~/dev/flocker-review (flocker-cloudformation-FLOC-3709)$ aws cloudformation validate-template --template-body file://flocker-cloudformation-template.json 2>&1| tail
                        "IpProtocol": "udp",
                        "ToPort": "65535"
                    }
                ]
            },
            "Type": "AWS::EC2::SecurityGroup"
        }
    }
}
' at 'templateBody' failed to satisfy constraint: Member must have length less than or equal to 51200
flocker-review-3 ♦ ~/dev/flocker-review (flocker-cloudformation-FLOC-3709)$
```

README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
...
+```
+
+* Boot the cluster:
+
+```
+$ cfn \
```
sarum90: Validation error seems to block this as well:
```
flocker-review-3 ♦ ~/dev/flocker-review (flocker-cloudformation-FLOC-3709)$ cfn --region us-east-1 --create flocker-cloudformation-template.json --parameter KeyName=$AWS_KEY_PAIR --parameter AccessKeyID=$AWS_ACCESS_KEY_ID --parameter SecretAccessKey=$AWS_SECRET_ACCESS_KEY flocker-stack-$RANDOM
  File "/Users/mewert/.virtualenvs/flocker-review-3/bin/cfn", line 112, in <module>
    create_stack(conn, values.stack, template, None, values.params)
  File "/Users/mewert/.virtualenvs/flocker-review-3/bin/cfn", line 40, in create_stack
    (literal_eval(e.error_message)['Error']['Message'],))
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 49, in literal_eval
    node_or_string = parse(node_or_string, mode='eval')
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/ast.py", line 37, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    1 validation error detected: Value '{
               ^
SyntaxError: invalid syntax
```

README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
...
+    --region us-east-1 \
+    --create flocker-cloudformation-template.json \
+    --parameter KeyName=<aws_ssh_keyname> \
+    --parameter AccessKeyID=<aws_key_id> \
+    --parameter SecretAccessKey=<aws_secret_access_key> \
+    flocker-stack-$RANDOM
```
sarum90: :grey_question: nit: is it useful for the user to know this for later? Might be better served by doing:

```
TMP_ID=$RANDOM
echo $TMP_ID
....
flocker-stack-$TMP_ID
```

Not needed if cfn prints this out again, couldn't verify during review.

admin/installer/cloudformation.py
```diff
@@ -0,0 +1,293 @@
...
+1 Control Node with Flocker Control Service, (TLS-enabled) Swarm Manager,
+                    (TLS-enabled) Docker, Ubuntu 14.04
+2 Agent Nodes with Flocker Dataset Agent, Swarm Agent, (TLS-enabled) Docker,
+                   Ubuntu 14.04
+1 Client Node with Flockerctl, Docker, Docker-compose, Ubuntu 14.04
+
+To manifest the blueprint, please input the JSON template at AWS CloudFormation
+Create Stack console (after replacing ``us-east-1`` with your Region):
+https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/new
+
```
sarum90: :grey_question:  ...or use cfn (as described above).

admin/installer/cloudformation.py
```diff
@@ -0,0 +1,293 @@
...
+# Base AMIs.
+template.add_mapping(
+    'RegionMap', {
+        'us-east-1':      {"FlockerAMI": "ami-5f401635",
+                           "ClientAMI": "ami-86590fec"},
+        'us-west-1':      {"FlockerAMI": "ami-e83a5088",
+                           "ClientAMI": "ami-0a254f6a"},
+        'us-west-2':      {"FlockerAMI": "ami-bd2b35dc",
+                           "ClientAMI": "ami-fa2b359b"},
```
sarum90: Do we need this copied everywhere? or should we document that we only support these three regions somewhere?

admin/installer/mysql/docker-compose-node1.yml
```diff
@@ -0,0 +1,10 @@
+database10:
```
sarum90: I think this file is unused? It be nice to either use it or at least mention it in some documentation so people know what it could be used for.

admin/installer/postgres/docker-compose-node1.yml
```diff
@@ -0,0 +1,12 @@
+postgres:
+  image: postgres
+  ports:
+     -  "5432:5432"
+  environment:
+     - "constraint:flocker-node==2"
+     - "POSTGRES_USER=flocker"
+     - "POSTGRES_PASSWORD=flocker"
```
sarum90: :grey_question: Setting these here is a bit concerning to me (although not quite as concerning as the MySQL one). It seems conceivable to me that people scan the internet for postgress and mysql databases running on the standard ports, and they might poke around for insecure passwords and usernames. Consider doing something to force the user to provide/enter these. (If you don't set them to anything, what happens?)

README.rst
```diff
@@ -49,6 +49,88 @@ You can run all of the tox environments using the command ``tox``.
...
+Cloudformation Deployment
+-------------------------
+
+* Generate a Cloudformation template:
+
+```
+$ python ./admin/installer/cloudformation.py > flocker-cloudformation-template.json
+```
+
+* Validate the template:
```
sarum90: As discussed offline, probably remove these directions as they don't currently work.

admin/installer/cloudformation.py
sarum90: Looks like we do:
Error
Template validation error: Template error: Unable to get mapping for RegionMap::ap-northeast-1::FlockerAMI

Please either document that you can only use this template in these regions or port the images around.

admin/installer/setup_client.sh
```diff
@@ -0,0 +1,61 @@
...
+# Get Postgres image.
+apt-get update
+sudo apt-get install -y postgresql-client
+
+# Get docker-compose.
+curl -L https://github.com/docker/compose/releases/download/1.5.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+chmod +x /usr/local/bin/docker-compose
+
+mkdir -p /home/ubuntu/postgres
+curl https://raw.githubusercontent.com/ClusterHQ/flocker/flocker-cloudformation-FLOC-3709/admin/installer/postgres/docker-compose-node0.yml > /home/ubuntu/postgres/docker-compose-node0.yml
```
sarum90: Before merge, probably change these to go from master.

admin/installer/cloudformation.py
```diff
@@ -0,0 +1,293 @@
...
+                    (TLS-enabled) Docker, Ubuntu 14.04
+2 Agent Nodes with Flocker Dataset Agent, Swarm Agent, (TLS-enabled) Docker,
+                   Ubuntu 14.04
+1 Client Node with Flockerctl, Docker, Docker-compose, Ubuntu 14.04
+
+To manifest the blueprint, please input the JSON template at AWS CloudFormation
+Create Stack console (after replacing ``us-east-1`` with your Region):
+https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/new
+
+"""
```
sarum90: For the sake of future maintainers, could I ask for a blob with the architecture here.

Specifically:
- Brief overview of the fact that you use  User Data [http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html#instancedata-add-user-data] To inject startup scripts to the instances.
- The user data scripts configure and install software, and then makes a curl request to a <cloudformation wait condition? (is that what it's called)> one the node is ready to use.
- If any of the wait conditions are not met after 600 seconds the script times out, and cloud formation rolls back the whole thing.

admin/installer/cloudformation.py
```diff
@@ -0,0 +1,293 @@
...
+}
+pattern = r'%([A-Z_]+)%'
+parts = re.split(pattern, parameter_string)
+iparts = iter(parts)
+new_parts = [base_url]
+for part in iparts:
+    new_parts.append(urllib.quote_plus(part))
+    key = next(iparts, None)
+    if key is not None:
+        new_parts.append(variables[key])
```
sarum90: From parameters to here is kind of hard to follow and a bit subtle. Looks like you need to generate an array  of (strings|refs) that can be concattenated into a url-encoded json blob after refs are resolved. I guess your guaranteed that AWS::StackName won't have any invalid URL characters, so this approach works. The hand-rolling of your own template language is a little non-obvious. Any chance you want to just put:

```
pattern = r'%([A-Z_]+)%'
 parts = re.split(pattern, parameter_string)
 iparts = iter(parts)
 new_parts = [base_url]
 for part in iparts:
     new_parts.append(urllib.quote_plus(part))
     key = next(iparts, None)
     if key is not None:
         new_parts.append(variables[key])
```

Into a helper function? Might make it a bit cleaner.

sarum90: actually.. can you do the urllib.quote_plus before hand? I guess you'd have to not use '%' in your pseudo template language then...

admin/installer/docker-swarm-ca-setup.sh
```diff
@@ -0,0 +1,77 @@
...
+send -- "${PASSPHRASE}\r"
+expect eof
+EOF
+chmod +x ${DOCKER_CERT_HOME}/createca1.exp
+${DOCKER_CERT_HOME}/createca1.exp
+
+cat > ${DOCKER_CERT_HOME}/createca2.exp << EOF
+#!/usr/bin/expect -f
+set timeout -1
+spawn openssl req -new -x509 -days 365 -key ${DOCKER_CERT_HOME}/ca-key.pem -sha256 -out ${DOCKER_CERT_HOME}/ca.pem
```
sarum90: :grey_question: @wallrj Seemed to think there might be a config file for openssl that we could use instead of using expect. Contemplate using that instead. I feel like that might be less brittle.

sarum90:
It appears that deletes of these clusters fail to delete the S3 buckets, have you had a chance to look into that?

https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks?stackId=arn:aws:cloudformation:us-east-1:121466501720:stack%2Fbcox-flockerformation%2F03215230-b65c-11e5-bfce-500c219ab02a&tab=events&filter=active

admin/installer/cloudformation.py
sarum90: Also, Add a tip for debugging:

- Turn off rollback on failure.
- See the log (stdout) from the  user data script by going to "get system log" in the AWS console. (or getting the serial output from the command line).

sarum90:
That's a really slick way to spin up a cluster!

I'm not crazy about trying to debug/maintain this big blob of bash in the future, but for kicks I added a random "&& false" to the middle of one one of the bash files and I didn't find it that hard to figure out what went wrong, so I'll call that good enough.

A bunch of nits here and there, but no major refactors requested. Merge once you've addressed issues.

README.rst
myechuri: Thanks! Fixed.

README.rst
myechuri: Thanks for pointing out. Output to ``/tmp``.

myechuri: Upon discussing with @wallrj , we decided to move the README edits to a followup branch if needed. Since docs reorg branch output (http://cube.lukemarsden.net:8080/docker-integration/cloudformation.html) covers most of this content, we will evaluate any missing steps once that shapes up. Thanks!

admin/installer/cloudformation.py
myechuri: @wallrj is working on adding Packer AMIs for all non-US regions in his Packer branch (https://github.com/ClusterHQ/flocker/pull/2451). Added a comment about the work in progress. Thanks!

admin/installer/mysql/docker-compose-node1.yml
myechuri: Thanks for pointing out! We used mysql as our endtoend test workload before moving on to postgres. Removed.

admin/installer/postgres/docker-compose-node1.yml
myechuri: Thanks for raising the concern!
If we do not set POSTGRES_USER, it defaults to ``postgres`` as superuser account. This maps to a default ``postgres`` OS user, whose account is locked.

If we want to move away from hardcoding ``POSTGRES_USER`` and ``POSTGRES_PASSWORD``, here are the options:

1) Have tests use Docker instead of docker-compose, since Docker allows ``-e`` flag:

``docker run --name db -d -e POSTGRES_PASSWORD=flocker -e POSTGRES_USER=flocker postgres``

Docker-compose does not seem to have a way to pass in env variable from command line.

2) Move docker-compose yaml files into test branch and auto-generate them with secret password.

3) Explicitly call out changing ``POSTGRES_USER`` and ``POSTGRES_PASSWORD`` in installer documentation, as part of post-installation steps.

Will discuss the options with @wallrj and update the thread. Thanks!

admin/installer/setup_client.sh
myechuri: Will do, thanks for the reminder!

admin/installer/cloudformation.py
myechuri: Thanks for the suggestion! Drafted https://github.com/ClusterHQ/flocker/blob/flocker-cloudformation-FLOC-3709/admin/installer/README.rst . Will get it reviewed by @wallrj .

admin/installer/cloudformation.py
myechuri: Removed this block because @wallrj and i thought it looked ugly in CloudFormation Output. Thanks!

admin/installer/docker-swarm-ca-setup.sh
myechuri: Thanks! Will explore it as a follow up. Please let me know if you think it is a blocker for this PR.

myechuri:
> It appears that deletes of these clusters fail to delete the S3 buckets, have you had a chance to look into that?

We wanted to leave S3 buckets as-is in case user wants to reuse {Flocker, Docker, Swarm} certs and config. Will explore the best way to preserve config while not blocking stack deletion. @wallrj and i were aware of this issue, but decided to keep S3 as is. We will devote cycles to it in a follow up PR. Thanks a lot for noticing!

myechuri:
@sarum90 : thanks for the timely and detailed review comments. @wallrj and i are grateful for your time and thoughts!

admin/installer/cloudformation.py
wallrj: Done. All except ap-northeast-2 which isn't yet supported by packer.

admin/installer/postgres/docker-compose-node1.yml
wallrj: Let's create a followup issue to improve the security of the stack.
 * The nodes can all be created inside a VPC
 * With only private IP addresses
 * Flocker agents and control service can communicate internally using private IP addresses.
 * Moby docker app can access Postgres using private IP address
 * User can access Mobydocker web port using a floating / NAT public IP address.
 * User can access the client machine using a floating / NAT IP address
 * Might also make the stack cheaper by cutting down the need for precious IPv4 Addresses.

admin/installer/setup_client.sh
wallrj: I think we should probably get rid of these.
The current documentation asks the user to download compose templates from the docs so there's no point trying to make them available this way.
I'll have to update the tests though.

admin/installer/README.rst
```diff
@@ -0,0 +1,74 @@
+FAQ for Maintainers
```
wallrj: I think we should move the documentation to `docs/gettingstarted/infrastructure`. That's where I've documented the installer tests and the packer AMI builder.

wallrj: And the current style is to put each sentence on a separate line.

admin/installer/setup_client.sh
myechuri: Removed the two yaml files from the branch.

admin/installer/README.rst
```diff
@@ -0,0 +1,74 @@
...
+
+How are user-specific inputs (like AWS AccessKeyID) sourced?
+------------------------------------------------------------
+
+User's AWS ``AccessKeyID``, ``SecretAccessKey``, and ``KeyName`` are sourced as `InputParameters`_ in CloudFormation template.
+
+Why is there a wrapper around S3 commands?
+------------------------------------------
+
+Startup dependency for Agent Node(s) and Client Node on Control Node only ensures that these nodes' EC2 instances are started after Control Node EC2 instance is booted up. This dependecy does not account for post-boot time taken by Control Agent to populate S3 bucket. Hence, the wait and retry loop around S3 commands to allow Agent Node(s) and Client Node wait for S3 bucket to be populated with data by Control Node.
```
wallrj: Perhaps we don't need this since we've got the WaitCondition resources.

admin/installer/cloudformation.py
```diff
@@ -0,0 +1,277 @@
+"""
+Troposphere script to generate an AWS CloudFormation JSON template.
+
+Sample usage:
+python cloudformation.py > /tmp/flocker-cluster.cloudformation.json
```
wallrj: We should move this to the documentation section.
And we should probably make a proper stub script for it `./admin/make_cloudformation` which imports functions from `./admin/cloudformation`.

admin/installer/docker-swarm-ca-setup.sh
sarum90: Not a blocker at all--just a thought.

admin/installer/setup_client.sh
sarum90: Probably also meant to remove these curl lines? Luckily curl doesn't fail on not founds, but we're just adding random files with the content `Not Found` to the image now:

```
 ♦ ~$ curl https://raw.githubusercontent.com/ClusterHQ/flocker/flocker-cloudformation-FLOC-3709/admin/installer/postgres/docker-compose-node0.yml
Not Found ♦ ~$ echo $?
0
 ♦ ~$ 
```

PR #2465, opened 2016-01-14
[FLOC-3843] Fix docker test private registry image

dwgebler:
This *may* be a fix for FLOC-3843. My current working theory is that the timeout on this test occurs when there is still another container running at the point this test is run, causing us to loop indefinitely as we wait for a response on the wrong port. (this scenario actually happened on my dev laptop, so it's definitely a potential cause)


flocker/node/functional/test_docker.py
```diff
@@ -627,6 +627,7 @@ def run_registry(self):
         :return: A ``Registry`` describing the registry which was started.
         """
         registry_name = random_name(self)
+        external_port = find_free_port()[1]
```
jml: This introduces a different race condition: between the time we find a free port and when we start the container, something else might come along and use that port. The way to avoid this is to bind to port 0 and somehow discover the port, as the original code was doing.

flocker/node/functional/test_docker.py
```diff
@@ -635,32 +636,20 @@ def run_registry(self):
...
         )
 
-        def extract_listening_port(client):
-            listing = client.list()
-            listing.addCallback(
-                lambda applications: list(
-                    next(iter(application.ports)).external_port
-                    for application
-                    in applications
-                )[0]
```
jml: If your hypothesis is correct (and it certainly seems plausible), then maybe a good first step is to raise an error here if there is more than one application. Although it's not a final solution, it will be an incremental improvement over a timeout (faster failure; more informative error message; better logs).

jml:
Thanks Dave! Please address comments and resubmit.

jml:
Thanks! LGTM

PR #2466, opened 2016-01-14
[FLOC-3883] FilePath matchers and temporary helpers

jml:
Adds new helper methods to our base test cases:

* `make_temporary_path` – like `mktemp`, but returns a `FilePath`
* `make_temporary_directory` – returns a temporary directory as a `FilePath`
* `make_temporary_file` – returns a temporary file, with optional content

Adds a new `flocker.testtools.matchers` module that provides `FilePath`-related matchers:

* `path_exists` – matches if a `FilePath` exists on disk
* `dir_exists` – matches if a `FilePath` exists and is a directory
* `file_exists` – matches if a `FilePath` exists and is a file
* `file_contents` – matches on the contents of a file if it exists

Note that `file_contents` is analogous to `testtools.matchers.FileContains` but different in that it _always_ takes a matcher and has a name that reflects this.

In order to make writing the above matchers easier, I added:

* `after` – a thing like `AfterPreprocessing` that operates on Matcher factories
* `on_mismatch` – like `after`, but applies to mismatches
* `OnMismatch` – life `AfterPreprocessing`, but applies to mismatches
* A new `_Mismatch` type that also stores the mismatched object, to allow for better composition

None of these have direct tests, as they came out of refactoring efforts.

In order to make _testing_ these changes easier, I added a `flocker.testtools.strategies` module that collates existing strategies and adds a couple for file paths. I also added a helper for creating simple test cases.

There are also unrelated lint fixes to `flocker/testtools/__init__.py`.

Apologies for the size of this PR. I was originally going to focus on the matchers, but found that doing the temp helpers was a natural pair.

### DIFF
--- a/flocker/testtools/_testhelpers.py
+++ b/flocker/testtools/_testhelpers.py
@@ -78,3 +78,19 @@ def run_test(case):
     result = unittest.TestResult()
     case.run(result)
     return result
+
+
+def make_test_case(base_case, behavior=lambda case: None):
+    """
+    Make a single test subclasses ``base_case`` and does ``behavior``.
+
+    :param type base_case: A ``TestCase`` class.
+    :param (TestCase -> A) behavior: The contents of the test. If not
+        specified, defaults to a passing test.
+
+    :rtype: ``base_case``
+    """
+    class FooTests(base_case):
+        def test_something(self):
+            pass
+    return FooTests('test_something')
--- a/flocker/testtools/_testhelpers.py
+++ b/flocker/testtools/_testhelpers.py
@@ -80,13 +80,11 @@ def run_test(case):
     return result
 
 
-def make_test_case(base_case, behavior=lambda case: None):
+def make_test_case(base_case):
     """
-    Make a single test subclasses ``base_case`` and does ``behavior``.
+    Make a single test that subclasses ``base_case`` and passes.
 
     :param type base_case: A ``TestCase`` class.
-    :param (TestCase -> A) behavior: The contents of the test. If not
-        specified, defaults to a passing test.
 
     :rtype: ``base_case``
     """
--- a/flocker/testtools/matchers.py
+++ b/flocker/testtools/matchers.py
@@ -172,11 +172,13 @@ Match if a file exists on disk.
 
 def file_contents(matcher):
     """
-    Match if a files contents match ``matcher``.
+    Match if a file's contents match ``matcher``.
 
     For example::
 
-        self.assertThat('/foo/bar/baz', file_contents(Equals('hello world')))
+        self.assertThat(
+            FilePath('/foo/bar/baz'),
+            file_contents(Equals('hello world')))
 
     Will match if there is a file ``/foo/bar/baz`` and if its contents are
     exactly ``hello world``.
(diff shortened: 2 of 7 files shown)


flocker/testtools/_testhelpers.py
```diff
@@ -78,3 +78,19 @@ def run_test(case):
     result = unittest.TestResult()
     case.run(result)
     return result
+
+
+def make_test_case(base_case, behavior=lambda case: None):
+    """
+    Make a single test subclasses ``base_case`` and does ``behavior``.
```
zubron: Should this be: "Make a single test *that* subclasses..."?

flocker/testtools/strategies.py
```diff
@@ -0,0 +1,49 @@
...
+e.g. 'foo', 'rc.local'.
+"""
+
+
+paths = lists(path_segments).map(lambda ps: FilePath('/'.join(ps)))
+"""
+Paths
+
+e.g. '/usr/local', 'foo/bar/bar'.
+"""
```
james-w: The docstring isn't quite right, as it won't generate absolute paths using join.

The strategy probably wants to generate absolute paths though, so changing
that rather than the docstring would be preferred.

flocker/testtools/matchers.py
```diff
@@ -0,0 +1,203 @@
...
+"""
+
+
+def file_contents(matcher):
+    """
+    Match if a files contents match ``matcher``.
+
+    For example::
+
+        self.assertThat('/foo/bar/baz', file_contents(Equals('hello world')))
```
james-w: The example is wrong, as the matcher works on FilePaths.

flocker/testtools/matchers.py
```diff
@@ -0,0 +1,203 @@
...
+
+file_exists = _on_filepath(FileExists)
+"""
+Match if a file exists on disk.
+"""
+
+
+def file_contents(matcher):
+    """
+    Match if a files contents match ``matcher``.
```
james-w: nit: file's contents

zubron:
Thanks for this change, @jml and thanks @james-w for helping with the review!

Apart from the few small comments, I think the change looks good. The Jenkins job wasn't triggered for this PR so I started a build. I'm happy for this to be merged once the comments are addressed and the tests are passing but feel free to get another review since you are my mentor for reviews. :smile:

zubron:
I'm going to move this issue to "Accepted" but I can change it back if you would like another review.

flocker/testtools/_testhelpers.py
jml: Yes. Fixed.

Also, noticed `behavior` was unused, so deleted.

flocker/testtools/strategies.py
jml: `FilePath` always abspaths what it gets, so it's probably OK. I've updated the examples to include `FilePath`.

flocker/testtools/matchers.py
jml: Fixed.

flocker/testtools/matchers.py
jml: We have met the enemy, and he is us. Fixed.

jml:
Thanks for the reviews! I'll merge this after I get a new test run.

PR #2467, opened 2016-01-14
[FLOC-3873] Generate a second set of jobs for release builds that don't merge master.

james-w:
Release builds shouldn't merge in to master before running. Firstly
this will sometimes fail for good reason, but more importantly
the merge isn't the code we want to test.

Any branch that named release/* will get a second set of jobs
generated that don't do the merge.

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -714,7 +714,8 @@ branches.each {
     println("iterating over branch... ${it}")
     branchName = it
     dashBranchName = branchName.replace("/","-")
-    isReleaseBuild = branchName.startsWith("target-branch-")
+    // our convention for release branches is release/flocker-<version>
+    isReleaseBuild = branchName.startsWith("release/*")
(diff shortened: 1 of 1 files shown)


jobs.groovy
```diff
@@ -692,12 +696,31 @@ branches.each {
...
+/* --------------------
+   MAIN ACTION:
+   --------------------
+*/
+// Iterate over every branch, and create folders, jobs
+branches.each {
+    println("iterating over branch... ${it}")
+    branchName = it
+    dashBranchName = branchName.replace("/","-")
+    isReleaseBuild = branchName.startsWith("target-branch-")
```
jml: Where does the "target-branch-" magic string come from? Please add a comment explaining why.

jml:
LGTM. Please merge after verifying that there are no regressions to the regular developer build process.

PR #2469, opened 2016-01-14
[FLOC-3888] correctly determine ip addresses to use for deployment via _compose

avg-I:
For some platforms the address is node.private_address while
for others private_address is not set and the right address
is node.address.


avg-I:
@mpgomez This change is going to create a conflict with your WIP.

jml:
LGTM.

PR #2470, opened 2016-01-14
[FLOC-3851]Create containers script

mpgomez:
Removed container creation from cluster setup script and created a new one so it can be done independently, reusing an existent cluster.

The script will wait until all the containers have been created, as that will make automation easier.


admin/cluster_setup.py
```diff
@@ -143,11 +143,6 @@ def main(reactor, args, base_path, top_level):
                                )
         gather_deferreds(results)
 
-        if options['apps-per-node'] > 0:
```
jongiddy: Remove the `apps-per-node` and `app-template` options from `RunOptions` and all the code after `main`, which is no longer used.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+
+from flocker.common import loop_until
+from flocker.control.httpapi import REST_API_PORT
+from flocker.apiclient import FlockerClient
+from flocker.ca import treq_with_authentication
+
+
+class ContainerOptions(usage.Options):
+    """
+    Parses the options pased as an argument to the create container script.
```
jongiddy: "pased" -> "passed"

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+    """
+    # XXX mandatory options
+    # XXX validations
```
jongiddy: What do these comments mean?  Handle any actions and delete them.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+         "No waiting is done by default."],
+        ['cluster', None, None,
+         'Configuration of the cluster'],
+    ]
+
+    synopsis = ('Usage: setup-cluster-containers --app-per-node <containers'
+                'per node> --app-template <name of the file> '
+                '--control-node <IPAddress>'
+                '--cert-directory <path where all the certificates are>'
+                '[--cluster <file> (UNUSED)')
```
jongiddy: Don't have an unused option in new code. If this is for future work, add it when it does something.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+import yaml
+import os
+import sys
+
+from copy import deepcopy
+from json import dumps
+from itertools import repeat
+from treq import json_content
```
jongiddy: Mix these standard lib imports into the first group, except `treq` which is a third-party library.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+
+        if self['wait'] is not None:
+            try:
+                self['wait'] = int(self['wait'])
+            except ValueError:
+                raise usage.UsageError("The wait timeout must be an integer.")
+
+
+def main(reactor, argv):
+    environ = os.environ
```
jongiddy: Pass `environ` in as a parameter. It will make this function more testable.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+        ``ClusterContainerDeployment`` constructor.
+
+        :param reactor: ``Reactor`` we are using.
+        :param env: ``environ`` with the current environment.
+            NOTE: alternative of making it work with env variables pending
+            implementation
+        :param options: ``ContainerOptions`` with the options passed to the
+            the script.
+        """
+        # XXX add the capability to use env variables if the options are not
```
jongiddy: :question: A fairly major change that would help here is to put this code in the `benchmark` directory (the only place it is likely to be used) and use the `BenchmarkCluster` class, which accepts either the env vars or the UFT installer configuration, and provides the control server object.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def deploy(self):
+        """
+        Deploy the new configuration: create the requested containers
+        and dataset in the cluster nodes.
+
+        :return Deferred: that will fire once the request to create all
+            the containers and datasets has been sent.
+        """
+        print "Listing current nodes"
```
jongiddy: You should replace explicit `print` statements with either eliot logging or with a class attribute (provided to `__init__`) that can be called with a string.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+        d.addCallback(self._set_nodes)
+        print "Building config"
+        d.addCallback(self._build_config)
+        print "Deploying new config"
+        d.addCallback(self._configure)
+        return d
+
+    def is_datasets_deployment_complete(self):
+        """
+        Check if all the dataset have been created.
```
jongiddy: "dataset" -> "datasets"

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,324 @@
...
+
+        :return `Deferred`: that will fire once the list datasets call
+            has been completed, and which result will bee True if all the
+            dataset have been created, or false otherwise.
```
jongiddy: Remove backquotes from `Deferred`. Inside the `:return...:` clause, types do not need backquotes.  

Change the description to state simply that the Deferred "fires with True if all datasets have been created, or False otherwise".  You don't need to mention that the call has completed.  That is an implementation detail - if the code changed to use magic to determine whether all datasets have been created, then the simpler description would still be true - the current longer description would be wrong.

admin/setup-cluster-containers
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Set up a cluster.
```
jongiddy: "Setup a number of stateful containers on a cluster."

jongiddy:
Please address comments and re-submit. Thanks.

admin/cluster_containers_setup.py
mpgomez: They are there because there is no much validation anywhere, and after speaking with Jono yesterday, it looks like there is no standard way of doing it, or standard on what to validate and what not :/ 
I added the validation that seems reasonable to me. It may be too much, or too less, or there might be better ways of doing it, but I didn't like the no validation approach.

admin/cluster_containers_setup.py
mpgomez: Ok. I can consider doing it with the next PR. Thanks!

mpgomez:
The eliot logging is not working
I have spent the whole morning trying to figure out what is wrong with it and it is getting silly. The logs are not being printed anywhere, and I am just trying to get eliot to print to the standard output
so I have sent it for review in case someone can give me a hint of how to use eliot.
I have used the code under benchmark/scenarios as an example of how to simply log with eliot. 
When that didn't work, I tried to use the acceptance.py eliot_output. It was being called, but the messages it was receiving were empty.
The I following the example in the zfs code, making my own custom messages and trying to use write, thinking that if I had my own logger instance I could set it up to write to stdout. Logger() doesn't get any arguments. And I already tried with add_destination in the previous approach. 
Help is more than welcome.

mpgomez:
Pushed working version of eliot logging. I think it should be ok now

itamarst:
Ah, Jon already did first review pass so I'll leave it for him.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,369 @@
...
+    ]
+
+    synopsis = ('Usage: setup-cluster-containers --app-per-node <containers'
+                'per node> --app-template <name of the file> '
+                '--control-node <IPAddress>'
+                '--cert-directory <path where all the certificates are>'
+                '[--wait <seconds to wait>]')
+
+    def postOptions(self):
+        # Mnadatory parameters
```
jongiddy: Mnadatory => Mandatory

jongiddy:
Setting up large number of virtually unusable containers has no real use outside of benchmarking. I suggest you move the two new files from the `admin` directory to the `benchmark` directory, to avoid cluttering the `admin` directory.  Remember to change the import in the script.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,369 @@
...
+                        raise ResponseError(response.code, error)
+
+                    d.addCallback(got_error)
+                    return d
+
+            posted.addCallback(got_response)
+            return posted
+
+        return do_configure()
+
```
jongiddy: Remove the extra blank line at the end to avoid a lint error. You should have one blank line at the end of the file.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,369 @@
...
+            try:
+                self['wait'] = int(self['wait'])
+            except ValueError:
+                raise usage.UsageError("The wait timeout must be an integer.")
+
+
+def main(reactor, argv):
+    environ = os.environ
+    add_destination(eliot_output)
+    # to_file(sys.stdout)
```
jongiddy: Remove this or uncomment it. 

:question: Send logs to stderr instead of stdout.

admin/cluster_setup.py
```diff
@@ -6,20 +6,14 @@
 import stat
 import string
 import sys
-import yaml
```
jongiddy: Add this line back. The yaml module is still used on line 188.

admin/cluster_containers_setup.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+class ContainerOptions(usage.Options):
+    """
+    Parses the options passed as an argument to the create container script.
+    """
+    description = "Set up containers in a Flocker cluster."
+
+    optParameters = [
+        ['apps-per-node', None, 0, 'Number of application containers per node',
```
jongiddy: A default of 0 made sense when this code was part of `setup-cluster`. Now that starting containers is the only purpose of this script, make the default 1.

jongiddy:
Please address the comments and re-submit. Also, there appear to be some comments from the previously addressed review that have not been addressed.  If there is a reason to not make the change suggested in a comment, please add a note to provide reasons for not making the change.

If all the above issues are addressed, a final review should be very quick.

admin/cluster_setup.py
mpgomez: I did. And the merged with master -.- Sorry.

admin/cluster_containers_setup.py
mpgomez: Those logs are not errors, so I don't think we should really send them to stderr. We could send them to a log file eventually, but now it is using eliot, it is just matter of changing the function that is currently printing to stdoout

benchmark/cleanup-cluster
```diff
@@ -4,11 +4,11 @@
 Clean up all containers and datasets in a cluster.
 """
 
-from _preamble import TOPLEVEL, BASEPATH
+from admin._preamble import TOPLEVEL, BASEPATH
```
jongiddy: These values are not used in `cluster_cleanup.main`. Remove these two parameters from `main`.

benchmark/cluster_containers_setup.py
```diff
@@ -0,0 +1,363 @@
...
+        ['control-node', None, None,
+         'Public IP address of the control node'],
+        ['cert-directory', None, None,
+         'Location of the user and control certificates and user key'],
+        ['wait', None, None,
+         "The timeout in seconds for waiting until the operation is complete. "
+         "No waiting is done by default."],
+    ]
+
+    synopsis = ('Usage: setup-cluster-containers --app-per-node <containers'
```
jongiddy: Add a space to end of this string.

benchmark/cluster_containers_setup.py
```diff
@@ -0,0 +1,363 @@
...
+                'per node> --app-template <name of the file> '
+                '--control-node <IPAddress>'
```
jongiddy: and to end of this string

benchmark/cluster_containers_setup.py
```diff
@@ -0,0 +1,363 @@
...
+                '--cert-directory <path where all the certificates are>'
```
jongiddy: and to end of this string.

benchmark/setup-cluster-containers
```diff
@@ -0,0 +1,13 @@
...
+"""
+Set up a cluster.
+"""
+
+import sys
+
+if __name__ == '__main__':
+    from twisted.internet.task import react
+    from benchmark.cluster_containers_setup import main
+    import os
```
jongiddy: Move all the imports to the top-level.

benchmark/cleanup-cluster
```diff
@@ -4,11 +4,11 @@
...
 
 import sys
 
 if __name__ == '__main__':
     from twisted.internet.task import react
-    from admin.cluster_cleanup import main
+    from benchmark.cluster_cleanup import main
```
jongiddy: Move all the imports to the top-level.

jongiddy:
Thanks for working on this. The two scripts make the benchmark setup and cleanup much easier.

A few minor comments to address, and then please merge when green (enough).

benchmark/cleanup-cluster
mpgomez: I'll do it, but that is not even my code :(

mpgomez:
Green build. Merging pull request

PR #2471, opened 2016-01-14
[FLOC-3893] Add a cleanup of the bind mount.

sarum90:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3893

Unfortunately I had could only reproduce the issue on an AWS Ubuntu instance. See the following gist for the logs of my run of the test that convinced me the issue is fixed:

https://gist.github.com/sarum90/db5e48b613f07678d887

### DIFF
--- a/flocker/node/agents/test/test_blockdevice_manager.py
+++ b/flocker/node/agents/test/test_blockdevice_manager.py
@@ -174,6 +174,7 @@ class BlockDeviceManagerTests(TestCase):
         src_directory = self._get_directory_for_mount()
         target_directory = self._get_directory_for_mount()
         self.manager_under_test.bind_mount(src_directory, target_directory)
+        self.addCleanup(self.manager_under_test.unmount, target_directory)
         for create, view in [(target_directory, src_directory),
                              (src_directory, target_directory)]:
             filename = str(uuid4())


james-w:
:+1: thanks for fixing this, the approach looks perfectly reasonable. Please
merge when tests pass, assuming a mentor agrees.

@jml mentor review please.

jml:
:+1:

flocker/node/agents/test/test_blockdevice_manager.py
```diff
@@ -174,6 +174,7 @@ def test_bind_mount(self):
         src_directory = self._get_directory_for_mount()
         target_directory = self._get_directory_for_mount()
         self.manager_under_test.bind_mount(src_directory, target_directory)
+        self.addCleanup(self.manager_under_test.unmount, target_directory)
```
jml: :question: No action required here, but I wonder if we could have an API that avoids the need to remember to `addCleanup`

sarum90: Yea +1. parse the output of `mount` or something. The quick solution seemed right for fast track, but I agree that'd be a more robust solution.

sarum90:
Thanks to whomever restarted the test that failed. Saved me a context switch

PR #2472, opened 2016-01-14
[FLOC-3894] Lower the clone timeout and raise the checkout retry count.

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3894

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -440,6 +440,10 @@ def define_job(dashBranchName, branchName, job_type, job_name, job_values, isRel
                     textParam("TRIGGERED_BRANCH", branchName,
                               "Branch that triggered this job" )
                 }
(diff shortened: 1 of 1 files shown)


exarkun:
Some bugs fixed in the GitSCM Jenkins plugin, in 2.4.0, might also be relevant to this change.  So it might be productive to look into upgrading our deployment to use that version of the plugin as well.

james-w:
:+1: Looks perfectly reasonable, thanks for improving it. Down with all failure conditions!

@jml mentor review please

james-w:
btw you got a red from jenkins, but all the actual tests passed, I'm looking
in to what is causing that today.

jml:
:+1:

PR #2473, opened 2016-01-14
[FLOC-3895] Count timeout for effect retry from the first effect failure

exarkun:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3895


jml:
LGTM. Merge when tests pass.

exarkun:
Thanks.  Only main multijob failed, and failed because of a known problem tracked elsewhere.  Merging.

PR #2475, opened 2016-01-15
[FLOC-3841] Add timeout to benchmark container and dataset creation.

jongiddy:
The benchmark script waits for datasets and containers to be created. Occasionally, this may fail, either due to incorrect configuration or Docker getting into a bad state.  Rather than have the benchmark script hang indefinitely, have it timeout.

The timeout is set to 10 minutes, which is a reasonable first guess for a liveable value.  It is not configurable at this stage.

### DIFF
--- a/benchmark/operations/create_container.py
+++ b/benchmark/operations/create_container.py
@@ -4,6 +4,7 @@ Operation to create a container.
 """
 
 from functools import partial
+from datetime import timedelta
 from uuid import UUID, uuid4
 
 from pyrsistent import PClass, field
@@ -17,7 +18,7 @@ from benchmark._interfaces import IProbe, IOperation
 from benchmark.operations._common import select_node
 
 
-DEFAULT_TIMEOUT = 600
+DEFAULT_TIMEOUT = timedelta(minutes=10)
 
 
 def loop_until_state_found(reactor, get_states, state_matches, timeout):
@@ -28,7 +29,7 @@ def loop_until_state_found(reactor, get_states, state_matches, timeout):
         of states.
     :param state_matches: Callable that accepts a state parameter, and
         returns a boolean indicating whether the state matches.
-    :param int timeout: Maximum seconds to wait for state to be found.
+    :param timedelta timeout: Maximum time to wait for state to be found.
     :return Deferred[Any]: The matching state.
     """
     def state_reached():
@@ -44,7 +45,7 @@ def loop_until_state_found(reactor, get_states, state_matches, timeout):
         return d
 
     d = loop_until(reactor, state_reached)
-    _timeout(reactor, d, timeout)
+    _timeout(reactor, d, timeout.total_seconds())
     return d
 
 
@@ -61,7 +62,8 @@ def create_dataset(
     :param UUID node_uuid: Node on which to create dataset.
     :param UUID dataset_id: ID for created dataset.
     :param int volume_size: Size of volume in bytes.
-    :param int timeout: Maximum seconds to wait for dataset to be mounted.
+    :param timedelta timeout: Maximum time to wait for dataset to be
+        mounted.
     :return Deferred[DatasetState]: The state of the created dataset.
     """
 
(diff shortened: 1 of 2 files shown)


benchmark/operations/create_container.py
```diff
@@ -10,14 +10,17 @@
...
-from flocker.common import gather_deferreds, loop_until
+from flocker.common import gather_deferreds, loop_until, timeout as _timeout
 from flocker.control import DockerImage
 
 from benchmark._interfaces import IProbe, IOperation
 from benchmark.operations._common import select_node
 
 
-def loop_until_state_found(reactor, get_states, state_matches):
+DEFAULT_TIMEOUT = 600
```
jml: :question: Consider a `timedelta`, which has unambiguous units. (e.g. `timedelta(minutes=10)`)

benchmark/operations/test/test_create_container.py
```diff
@@ -84,6 +84,43 @@ def cleanup(result):
...
+
+        def run_probe(probe):
+            def cleanup(result):
+                cleaned_up = probe.cleanup()
+                cleaned_up.addCallback(lambda _ignored: result)
+                return cleaned_up
+            d = probe.run()
+            d.addCallback(cleanup)
+            return d
+        d.addCallback(run_probe)
```
jml: :question: Since everything is synchronous, I don't think we need to have the cleanup in a callback. We could instead have:

```python
d = operation.get_probe()
def run_probe(probe):
    self.addCleanup(self.successResultOf, probe.cleanup)
    return probe.run()
d.addCallback(run_probe)
```

This has the advantage of clarifying exactly where we expect the failure. I'll make a couple of other comments that relate to this optional suggestion.

benchmark/operations/test/test_create_container.py
```diff
@@ -84,6 +84,43 @@ def cleanup(result):
...
+        operation = CreateContainer(clock, cluster)
+        d = operation.get_probe()
+
+        def run_probe(probe):
+            def cleanup(result):
+                cleaned_up = probe.cleanup()
+                cleaned_up.addCallback(lambda _ignored: result)
+                return cleaned_up
+            d = probe.run()
+            d.addCallback(cleanup)
```
jml: We only want to clean up if successful. Is that right?

benchmark/operations/test/test_create_container.py
```diff
@@ -84,6 +84,43 @@ def cleanup(result):
...
+            return d
+        d.addCallback(run_probe)
+
+        clock.advance(DEFAULT_TIMEOUT)
+
+        # No control_service.synchronize_state() call, so cluster state
+        # never shows container is created.
+
+        # The Deferred fails if container not created within 10 minutes.
+        self.failureResultOf(d)
```
jml: Which thing are we expecting to fail? Running the probe, the cleanup, or getting the probe? This assertion catches all of them. Can we make it a narrower assertion?

benchmark/operations/create_container.py
```diff
@@ -39,11 +43,14 @@ def find_match(states):
 
         return d
 
-    return loop_until(reactor, state_reached)
+    d = loop_until(reactor, state_reached)
+    _timeout(reactor, d, timeout)
```
jml: :question: Another way of implementing a timeout is to pass a finite iterator to `loop_until`. e.g. 

```python
d = loop_until(reactor, state_reached, [0.1] * DEFAULT_TIMEOUT / 0.1)
```

Obviously this isn't _quite_ the same thing, since it doesn't account for the probe time.

jml:
Thanks Jon! Timeouts & retries on all the things! 

I'd appreciate having a quick glance over this before it goes in.

benchmark/operations/create_container.py
jongiddy: I wanted this to be based on time, since the main cause of delay is the Docker image pull, which I guess should usually take < 10 minutes.  And if experience says not, I'd like to be able to ensure we allow more time, not just more retries.

benchmark/operations/test/test_create_container.py
```diff
@@ -84,6 +84,72 @@ def cleanup(result):
...
+        clock.advance(1)
+        control_service.synchronize_state()  # deletion of pull container
+        clock.advance(1)
+
+        # get_probe has completed successfully
+        self.successResultOf(d)
+
+        def run_probe(probe):
+            return probe.run()
+        d.addCallback(run_probe)
```
jml: Alternatively:

```python
probe = self.successResultOf(d)

d = probe.run()
```

(I think)

jml:
Thanks @jongiddy. LGTM. One suggestion. Please merge after addressing & after the tests pass.

PR #2476, opened 2016-01-15
[FLOC-3885] generate and save managed config for a newly created cluster

avg-I:
This is useful for the easy re-use of the cluster nodes for a new cluster.
Along with `managed.yaml` we also now save a `environment.env` file that contains all the
environment variables needed for running acceptance tests on the cluster
(and potentially interacting with it in other useful ways).

### DIFF
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -171,10 +171,15 @@ def main(reactor, args, base_path, top_level):
     try:
         yield runner.ensure_keys(reactor)
         cluster = yield runner.start_cluster(reactor)
-        managed_config = options['cert-directory'].child("managed.yaml")
-        managed_config.setContent(
-            yaml.safe_dump(generate_managed_config(cluster))
+
+        managed_config_file = options['cert-directory'].child("managed.yaml")
+        managed_config = dict()
+        managed_config.update(options['config'])
+        managed_config.update(generate_managed_config(cluster))
+        managed_config_file.setContent(
+            yaml.safe_dump(managed_config, default_flow_style=False)
         )
+
         if options['distribution'] in ('centos-7',):
             remote_logs_file = open("remote_logs.log", "a")
             for node in cluster.all_nodes:
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -124,6 +124,16 @@ def _ensure_empty_directory(path):
         )
 
 
+def generate_managed_config(cluster):
+    addresses = list()
+    for node in cluster.agent_nodes:
+        if node.private_address is not None:
+            addresses.append([node.private_address, node.address])
+        else:
+            addresses.append(node.address)
+    return {"managed": {"addresses": addresses}}
+
+
 @inlineCallbacks
 def main(reactor, args, base_path, top_level):
     """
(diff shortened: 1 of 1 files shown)


admin/cluster_setup.py
```diff
@@ -124,6 +126,19 @@ def _ensure_empty_directory(path):
         )
 
 
+def generate_managed_config(cluster):
```
jml: Docstring please.

admin/cluster_setup.py
```diff
@@ -159,6 +174,15 @@ def main(reactor, args, base_path, top_level):
...
         yield runner.ensure_keys(reactor)
         cluster = yield runner.start_cluster(reactor)
+
+        managed_config_file = options['cert-directory'].child("managed.yaml")
+        managed_config = dict()
+        managed_config.update(options['config'])
+        managed_config.update(generate_managed_config(cluster))
+        managed_config_file.setContent(
+            yaml.safe_dump(managed_config, default_flow_style=False)
+        )
```
jml: :question: Maybe make this a separate function that takes `managed_config_file`, `options['config']`, and `cluster` as parameters?

admin/cluster_setup.py
```diff
@@ -193,14 +217,25 @@ def main(reactor, args, base_path, top_level):
...
+                        "export {name}={value};\n".format(
+                            name=environment_variable,
+                            value=shell_quote(
+                                environment_variables[environment_variable]
+                            ),
+                        )
+                    )
+                environment = ''.join(environment_strings)
+                print("The following variables describe the cluster:")
+                print(environment)
```
jml: Since we're already mixing IO with computation, there's no reason to build `environment_strings`, join it, then print it. Instead, we might as well print the "export ..." string as we figure it out.

admin/cluster_setup.py
```diff
@@ -124,6 +126,19 @@ def _ensure_empty_directory(path):
...
+    addresses = list()
+    for node in cluster.agent_nodes:
+        if node.private_address is not None:
+            addresses.append([node.private_address, node.address])
+        else:
+            addresses.append(node.address)
+    config = dict()
+    config["addresses"] = addresses
+    config["upgrade"] = True
+    return {"managed": config}
```
jml: Best to construct this all at once:

```python
return {
    'managed': {
        'addresses': addresses,
        'upgrade': True,
    },
}
```

jml:
LGTM. Please merge after addressing comments and making sure tests pass.

admin/cluster_setup.py
avg-I: The code before my change did exactly that. With my change the variables are not only printed, but are also saved  to a file (see the next 5 lines).  So, I thought that it made sense to build the full content and then to print it and to save it rather than printing it and appending it to a file piecemeal.

admin/cluster_setup.py
avg-I: Done.

admin/cluster_setup.py
avg-I: Done this slightly differently.

admin/cluster_setup.py

PR #2477, opened 2016-01-15
[FLOC-3896] Move the copyArtifacts tasks back the multijob.

james-w:
This reverts 31c881ccda449a3961896148f80093f853709c89 which moved the
copyArtifacts steps to the invidual jobs, rather than the multijob.

This means that after a build each job tries to copy the artifacts from the
previous run of the job to the current one. The intention is to copy the
results from the current build to the multijob so that they can be examined
in aggregate with the other jobs.

In order to do that the copy needs to be run in the multijob, as it's
"copy <files> to here" operation.

The original intent of the change was to have the artifacts published for
the cron jobs. That is in fact done by defining publishers, and was done in
a followup commit b245c5f7387285db60545f33d29336d2af78f3a2.

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -648,35 +648,33 @@ def build_multijob(dashBranchName, branchName, isReleaseBuild) {
             } /* ends parallel phase */
 
             /* we've added the jobs to the multijob, we now need to fetch and
-               archive all the artifacts produced by the different jobs */
-            for (job_type_entry in GLOBAL_CONFIG.job_type) {
+               archive all the artifacts produced by the different jobs
+               Skip the cron jobs as they don't run from this multijob,
+               so the copied artifacts could be from a build of a previous
+               revision, which may confuse things. */
+            for (job_type_entry in GLOBAL_CONFIG.job_type.findAll { it.key != 'cronly_jobs' } ) {
                 job_type = job_type_entry.key
-                /* Skip the cron jobs as they don't run from this multijob,
-                   so the copied artifacts could be from a build of a previous
-                   revision, which may confuse things. */
-                if (job_type != 'cronly_jobs') {
-                    for (job_entry in job_type_entry.value) {
-                        job_name = job_entry.key
-                        job_values = job_entry.value
-                        for (_module in job_values.with_modules) {
-                            _job_name = job_name + '_' + escape_name(_module)
-                            /* no every job produces an artifact, so make sure wew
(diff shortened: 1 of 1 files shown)


james-w:
To see the results of this you can go to http://jenkins:8080/job/ClusterHQ-flocker/job/fix-jenkins-copy-arfifacts-FLOC-3896/job/__main_multijob/5/ and see that it published JUnit report and coverage report, and in http://jenkins:8080/job/ClusterHQ-flocker/job/fix-jenkins-copy-arfifacts-FLOC-3896/job/__main_multijob/5/console you can see it running the job, then copying the artifacts from it (the other jobs were disabled by hand so there is nothing to copy from them)

james-w:
It wasn't green, but you can see a JUnit result on http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/fix-jenkins-copy-arfifacts-FLOC-3896/job/__main_multijob/1/ which wouldn't have been there before.

james-w:
Now it is green, showing that it fixes the problem with __main_multijob failing even if all
the sub-builds fail.

jobs.groovy
```diff
@@ -581,6 +559,17 @@ def define_job(dashBranchName, branchName, job_type, job_name, job_values, isRel
...
 
+def escape_name(name) {
+    /*
+        Escape a name to make it suitable for use in a path.
+
+        :param unicode name: the name to escape.
+        :return unicode: the escaped name.
+    */
+    return name.replace('/', '-')
+}
```
jml: :+1:

jobs.groovy
```diff
@@ -657,6 +646,43 @@ def build_multijob(dashBranchName, branchName, isReleaseBuild) {
...
+                                    fingerprintArtifacts(true)
+                                    buildSelector {
+                                        workspace()
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
```
jml: :question: This is pretty dense and pretty nested. It's not going to be easy for someone else (at least, not me) to come back later and grok this.

Could we try:

* moving to a separate function
* using `next` or `continue` or something 
* turning into multiple functions

But I see that this is a revert, so never mind.

jml:
LGTM, thanks.

jobs.groovy
james-w: Yeah, wanted to leave it as close to a revert as possible. I have another branch in 
the works to clean up this file a bit.

PR #2478, opened 2016-01-15
[FLOC-3901] Refactor some of the groovy code to reduce duplication.

james-w:
Rather than relying on constructing the same or similar strings in several
places use functions to encapsulate the details to reduce the chance of a
mismatch.

Unify some of the job definition as indicated by XXX comments.

Extract the code to trigger a sub job from a multijob without failing the
multijob in to a function.

Migrate from using "job" to trigger a sub job from a multijob as that is
deprecated in favour of using phaseJob. This is currently showing the
deprecation warning on the live instance.

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -573,6 +573,34 @@ def trigger_sub_job(context, name) {
 }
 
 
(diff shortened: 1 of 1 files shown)


jml:
Looks good, thanks. Much clearer. (I wish we had tests).

Issue #1903, opened 2015-08-31
Flocker cannot attach openstack volume if instance use floating ip


chernogorsky:
Hi,
I have this configuration
Flocker 1.2
Node: Centos 7
Control: Ubuntu 15.04 + unofficial cli
Openstack: Kilo

Flocker-Instance (FI) 
Private IP
Floating IP

Inside FI there is only Private IP (and services IPs)

When I have been trying to create dataset (using flocker-volumes ) I have received that log:
Aug 31 06:12:23 docker-2-hoster.novalocal flocker-dataset-agent[8956]: {"api_ips": "{IPv4Address('XXX.XXX.XXX.114'): u'137e95d5-8c94-4c4f-aeb1-c2c4b9d5d933', IPv4Address('10.0.0.214'): u'243edf5d-1324-46a1-89ff-d55c106ff930', IPv4Address('10.0.0.216'): u'115fc519-c0f5-4c5f-a79c-a6625396a84b', IPv4Address('10.0.0.165'): u'b77a6529-2649-4ec7-bb71-38cb0805ab89', IPv4Address('XXX.XXX.XXX.121'): u'115fc519-c0f5-4c5f-a79c-a6625396a84b', IPv4Address('XXX.XXX.XXX.116'): u'b77a6529-2649-4ec7-bb71-38cb0805ab89', IPv4Address('10.0.0.215'): u'137e95d5-8c94-4c4f-aeb1-c2c4b9d5d933', IPv4Address('10.0.0.213'): u'9bfa678f-d6db-4473-9940-6315c917cf17'}", "task_uuid": "65debb45-2616-4d9c-b5d0-a34fecb269f1", "local_ips": "set([IPv4Address('127.0.0.1'), IPv6Address('fe80::f816:3eff:fe23:3b60'), IPv6Address('::1'), IPv4Address('172.17.42.1'), IPv4Address('10.0.0.216')])", "timestamp": 1441001543.45506, "message_type": "flocker:node:agents:blockdevice:openstack:compute_instance_id:not_found", "task_level": [2, 3]}

After some investigation, I have found that "problem" in this string (file flocker/node/agents/cinder.py)
if api_addresses and api_addresses.issubset(local_ips):

I suppose, because, it tries to find all addresses (including floating ip) in the list of local ips
After changing to this
if api_addresses and len(api_addresses.intersection(local_ips)):

It start to work

itamarst:
[...]

PR #2479, opened 2016-01-15
[FLOC-2966] Floating IPs no longer break node identification in Cinder backend

itamarst:
https://clusterhq.atlassian.net/browse/FLOC-2966


wallrj:
Thanks @itamarst 

1. There are some test failures which look related: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/floating-IPs-FLOC-2966/job/run_trial_for_cinder_storage_driver_on_CentOS_7_flocker-node-agents-cinder.py/lastFailedBuild/console
2. It seems risky to match on *any* intersection of local and API reported IP addresses. But I can't remember which local / private / loopback addresses are reported by the nova API. 
 * What if multiple nodes share a private network, maybe in separate virtual networks? 
 * What if an administrator has (for some reason) assigned one of the floating IPs directly to a machine.
   * Is there some way to identify floating IPs so as to exclude them when we do the match?
  * I used the ``nova show`` command to look at some of our acceptance test nodes and the docker network IPs etc weren't reported there, so I guess it's not such a problem.
3. It looks like you can only set up floating IP in rackspace private cloud so I haven't been able to test this fully.
4. I think it might be worth investigating (again) whether there's some unambiguous way to identify *this* node. I'm pretty sure that other OpenStack deployments do make the instance_id available in /sys in the same way that other openstack deployments / hypervisors make the cinder block device ID available in the disk serial number inside the OS. And since the OP with floating IPs can't have been using Rackspace, perhaps we should be finding out what type of OpenStack they're on and adding a compute_id algorithm for that platform.
5. The tests in this branch probably address this issue: https://clusterhq.atlassian.net/browse/FLOC-2281


Ok that's all I can think of.
Please answer or address the points above, get green build results and merge unless you think we need to investigate a different algorithm.

wallrj:
For example, it looks like the `curl http://169.254.169.254/openstack/latest/meta_data.json` method might work on Redhat Openstack, so perhaps that should be attempted in preference to the IP address matching algorithm.
 * https://ask.openstack.org/en/question/10140/wget-http1692541692542009-04-04meta-datainstance-id-error-404/?answer=26638#post-id-26638

itamarst:
I've seen some discussion on that in general, so will try that approach.

itamarst:
Apparently on Rackspace VMs you have to use yet another, different mechanism: https://community.rackspace.com/products/f/25/t/4316

itamarst:
I can't even tell if the 169 thing works in all-but-Rackspace or just randomly depending on your vendor. I am all out of caring for the day so will revisit tomorrow. Or maybe in 2026, the Year of Working Openstack.

wallrj:
> I can't even tell if the 169 thing works in all-but-Rackspace or just randomly depending on your vendor. I am all out of caring for the day so will revisit tomorrow

The 169 thing doesn't work with Rackspace. Or at least it didn't when we were originally writing the Cinder driver.

I just thought it might work with Redhat OpenStack and could be tried before falling back to the IP matching method.

itamarst:
On Rackspace the 169 thing just... sits there. So you have to have a timeout on every call? Or something.

wallrj:
@wallnerryan I've put this up for review as an interim fix for Flocker 1.15.0

There's a discussion in https://github.com/ClusterHQ/flocker/issues/1903 about first checking various metadata sources before falling back to IP address matching.
 * https://github.com/kubernetes/kubernetes/pull/17041
   (although that PR was never merged)

But since the release is tomorrow I think this is worth merging as-is.

Later we can copy the strategy used by Kubernetes in:
 * https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/rackspace/rackspace.go
 * https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/openstack/openstack.go

wallrj:
There are still some `UnknownInstanceID` failures in the `flocker-dataset-agent` logs, but these occur during the Flocker upgrade tests which will downgrade Flocker to the last released version....a version which still has the bug.

Compare the error time stamps with the timestamp of the failed test.

```
2016-09-12 08:43:34.591816Z
  action_type: 'flocker:agent:discovery'
  action_status: 'failed'
  exception: 'flocker.node.agents.blockdevice.UnknownInstanceID'
--
09b0a26c-5b22-441f-a9ee-b23af39f6408 -> /2/2/2/2/2/3
2016-09-12 08:43:34.592373Z
  message_type: 'eliot:traceback'
  exception: 'flocker.node.agents.blockdevice.UnknownInstanceID'
  reason: 'Could not find valid instance ID for <flocker.node.agents.cinder.CinderBlockDeviceAPI object at 0x7f39ec981e10>'
  traceback: "Traceback: <class 'flocker.node.agents.blockdevice.UnknownInstanceID'>: Could not find valid instance ID for <flocker.node.agents.cinder.CinderBlockDeviceAPI object at 0x7f39ec981e10>

...


2016-09-12 08:48:51.081978Z
  action_type: 'flocker:agent:discovery'
  action_status: 'failed'
  exception: 'flocker.node.agents.blockdevice.UnknownInstanceID'
--
4b89e845-88ab-42f4-8e96-6803e8d22a41 -> /4/2/2/2/3/2/3
2016-09-12 08:48:51.082494Z
  message_type: 'eliot:traceback'
  exception: 'flocker.node.agents.blockdevice.UnknownInstanceID'
  reason: 'Could not find valid instance ID for <flocker.node.agents.cinder.CinderBlockDeviceAPI object at 0x7f39ec981e10>'
  traceback: "Traceback: <class 'flocker.node.agents.blockdevice.UnknownInstanceID'>: Could not find valid instance ID for <flocker.node.agents.cinder.CinderBlockDeviceAPI object at 0x7f39ec981e10>
```


The errors stop once that test has timed out.

flocker/node/agents/cinder.py
```diff
@@ -380,6 +380,31 @@ def _extract_nova_server_addresses(addresses):
...
 
+def _get_compute_id(local_ips, id_to_node_ips):
+    """
+    Compute the instance ID of the local machine.
+
+    Expectation is that our local IPs intersect with one (only) of the
+    remote nodes' sets of IPs.
+
+    :param set local_ips: The local machine's IPs.
+    :param id_to_node_ips: Mapping from instance IDs to sets of IPs, as
```
wallnerryan: Does this include ipv6

flocker/node/agents/cinder.py
```diff
@@ -380,6 +380,31 @@ def _extract_nova_server_addresses(addresses):
...
+        reported by OpenStack.
+
+    :return: Instance ID of local machine.
+    """
+    matching_instances = []
+    for server_id, api_addresses in id_to_node_ips.items():
```
wallnerryan: given the conversation on slack, its unclear to me how vital the need to do the metadata validations first would be. The only case I can think of is multiple virtual networks in the same openstack cluster that may report overlapping layer 2 addresses.

wallnerryan:
looks like `flocker.acceptance.endtoend.test_dataset.DatasetAPITests.test_upgrade` is failing on trusty again but likely/assuming this is the issue with FLOC-4495

otherwise I think it looks OK, apart of the considerations of the overlapping local addresses in openstack. WDYT? Is it work merging as is then for 1.16 we _should_ implement the proper metadata checks first? Or possibly add a check if there are 2 id's returned because of an overlap, have a secondary way to figure out the right one?

flocker/node/agents/cinder.py
wallrj: Yep, it includes IPv6. And I think that it might be Floating IPv6 addresses that are causing the problem on Ubuntu 14.04.
On other Ubuntu's the IPv6 addresses are assigned to interfaces in the OS.

wallrj:
> looks like flocker.acceptance.endtoend.test_dataset.DatasetAPITests.test_upgrade is failing on trusty > again but likely/assuming this is the issue with FLOC-4495

Nope, see above. It's because that test first downgrades...to a version where Floating IPs prevent instance identification so the volumes are never manifested.

wallrj:
Closing in favour of #2901

wallrj:
I've got configdrive and metadataservice branches ready for review, but it turns out that neither of those will work on our Rackspace build slaves because they are booted without the config drive option and Rackspace doesn't provide a metadata service.
So I'll re-open and merge this branch and then combine it with the other mechanisms.

wallrj:
The Ubuntu Trusty  + Rackspace acceptance tests are still not going to get beyond the upgrade test because once that fails with a 300s timeout, Flocker never gets re-installed with a working version.
I'm going to merge this anyway and fix that problem in a separate branch.

PR #2481, opened 2016-01-15
[FLOC-3904] Match Docker plugin API docs

itamarst:
`Err` in success cases is supposed to be empty string, not null/None.

https://github.com/docker/docker/commit/cc62a88524b623c8f86075d9ed220196cf5e92da

### DIFF
--- a/flocker/dockerplugin/test/test_schemas.py
+++ b/flocker/dockerplugin/test/test_schemas.py
@@ -32,10 +32,10 @@ def build_simple_test(command_name):
             # Wrong Err types:
             {"Err": 1}, {"Err": {}},
             # Extra field:
-            {"Err": None, "Extra": ""},
+            {"Err": "", "Extra": ""},
         ],
         passing_instances=[
-            {"Err": None},
+            {"Err": ""},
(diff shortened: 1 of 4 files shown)


flocker/dockerplugin/test/test_schemas.py
```diff
@@ -32,10 +32,10 @@ def build_simple_test(command_name):
             # Wrong Err types:
             {"Err": 1}, {"Err": {}},
```
sarum90: :grey_question: Should we make sure {"Err": None} fails the schema now?

sarum90:
Thanks! Nice to be ready for docker 1.10 before its released :)

One optional comment. I'm not sure it adds much though. I trust the JSON schema to work as described, added tests seem more like a maintenance headache.

Change looks like it was largely mechanical and done correctly. Am I correct in claiming that the currently passing presubmits means that this does not break support for docker 1.9?  If so, merge whenever you are satisfied with the jenkins tests.

itamarst:
If acceptance tests passed that means current Docker works, yeah.

PR #2484, opened 2016-01-18
[FLOC-3900] tool for submitting benchmark results to a server

avg-I:
The tool expects the result document to be provided via the standard
input.

### DIFF
--- a/benchmark/submit.py
+++ b/benchmark/submit.py
@@ -13,6 +13,7 @@ from twisted.internet.defer import succeed
 from twisted.internet.task import react
 from twisted.python import usage
 from twisted.web import client, http
+from twisted.web.http_headers import Headers
 from twisted.web.iweb import IBodyProducer
 
 from zope.interface import implementer
@@ -44,7 +45,12 @@ def submit(agent, result):
     :return: Deferred that fires when a reponse from the server is received.
     """
     body = StringProducer(json.dumps(result))
-    req = agent.request("POST", "/v1/benchmark-results", bodyProducer=body)
+    req = agent.request(
+        "POST",
+        "/v1/benchmark-results",
+        headers=Headers({'Content-Type': ['application/json']}),
+        bodyProducer=body
+    )
 
     def get_response_body(response):
         d = client.readBody(response)
--- a/benchmark/submit.py
+++ b/benchmark/submit.py
@@ -1,0 +1,123 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Submit benchmark results to a server.
+"""
+
+from __future__ import print_function
+
+import json
+import sys
+
+from twisted.internet import endpoints
+from twisted.internet.defer import succeed
+from twisted.internet.task import react
+from twisted.python import usage
+from twisted.web import client, http
+from twisted.web.iweb import IBodyProducer
+
+from zope.interface import implementer
+
+
+@implementer(IBodyProducer)
+class StringProducer(object):
+    def __init__(self, body):
+        self.body = body
+        self.length = len(body)
+
+    def startProducing(self, consumer):
+        consumer.write(self.body)
+        return succeed(None)
+
+    def pauseProducing(self):
+        pass
+
+    def stopProducing(self):
+        pass
+
+
+def submit(agent, result):
+    """
+    Post the given result using the given agent.
+
+    :param Agent agent: The HTTP client agent.
+    :param dict result: The JSON-compatible result.
+    :return: Deferred that fires when a reponse from the server is received.
+    """
+    body = StringProducer(json.dumps(result))
+    req = agent.request("POST", "/v1/benchmark-results", bodyProducer=body)
+
+    def get_response_body(response):
+        d = client.readBody(response)
+        d.addCallback(lambda body: (response, body))
+        return d
+
+    req.addCallback(get_response_body)
+
+    def process_response(response_and_body):
+        (response, body) = response_and_body
+        if response.code != http.CREATED:
+            print("Failed to submit the request", file=sys.stderr)
+            print("{} {}".format(response.code, response.phrase),
+                  file=sys.stderr)
+        else:
+            print("Successfully submitted the result")
+
+        json_body = json.loads(body)
+
+        if response.code != http.CREATED:
+            print("Message: {}".format(json_body['message']), file=sys.stderr)
+        else:
+            print("Assigned result ID: {}".format(json_body['id']))
+
+    req.addCallback(process_response)
+    return req
+
+
+class Options(usage.Options):
+    optParameters = [
+        ['address', None, None, "The address of the benchmark server."],
+        ['port', None, 8888, "The port of the benchmark server.", int],
+    ]
+
+    def postOptions(self):
+        if self.get('address') is None:
+            raise usage.UsageError("The address must be specified.")
+
+
+def main(reactor, args):
+    try:
+        options = Options()
+        options.parseOptions(args)
+    except usage.UsageError as e:
+        sys.stderr.write(e.args[0])
+        sys.stderr.write('\n\n')
+        sys.stderr.write(options.getSynopsis())
+        sys.stderr.write('\n')
+        sys.stderr.write(options.getUsage())
+        raise SystemExit(1)
+
+    agent = client.ProxyAgent(
+        endpoints.TCP4ClientEndpoint(
+            reactor,
+            options['address'],
+            options['port'],
+        ),
+        reactor,
+    )
+
+    try:
+        result_input = sys.stdin.read()
+        result = json.loads(result_input)
+    except ValueError as e:
+        print(result_input, file=sys.stderr)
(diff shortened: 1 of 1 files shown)


benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+def main(reactor, args):
+    try:
+        options = Options()
+        options.parseOptions(args)
+    except usage.UsageError as e:
+        sys.stderr.write(e.args[0])
+        sys.stderr.write('\n\n')
+        sys.stderr.write(options.getSynopsis())
```
sarum90: :grey_question: No action required. Some experimentation and looking at http://twistedmatrix.com/trac/browser/tags/releases/twisted-15.5.0/twisted/python/usage.py#L194 suggests that the next three lines could be replaced with:

```
sys.stderr.write(str(options))
```

Rather than manually printing both the synopsis and the usage.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+from twisted.python import usage
+from twisted.web import client, http
+from twisted.web.http_headers import Headers
+from twisted.web.iweb import IBodyProducer
+
+from zope.interface import implementer
+
+
+@implementer(IBodyProducer)
+class StringProducer(object):
```
sarum90: Want to just add a tiny class comment describing what this does?

As an aside, I'm surprised twisted doesn't provide this.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Submit benchmark results to a server.
```
sarum90: Would you mind extending this to talk about intended use of this script? It looks like this is grabbed for the option text, and it would be great if it at least mentioned that it expects a json blob on the standard in that will be submitted to a  <reference  to benchmark server code>. Right now just running it it's not clear what will be submitted.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+class StringProducer(object):
+    def __init__(self, body):
+        self.body = body
+        self.length = len(body)
+
+    def startProducing(self, consumer):
+        consumer.write(self.body)
+        return succeed(None)
+
+    def pauseProducing(self):
```
sarum90: :grey_question: nit: resumeProducing is also required in order to implement IBodyProducer.

This suggests you might want to make a unit tests and just run something from zope.interface.verify on this class.  Might be overkill for this utility though, I don't know.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+        options.parseOptions(args)
+    except usage.UsageError as e:
+        sys.stderr.write(e.args[0])
+        sys.stderr.write('\n\n')
+        sys.stderr.write(options.getSynopsis())
+        sys.stderr.write('\n')
+        sys.stderr.write(options.getUsage())
+        raise SystemExit(1)
+
+    agent = client.ProxyAgent(
```
sarum90: :grey_question:  Consider using https://github.com/twisted/treq rather than twisted directly. I think you might not have to roll your own string producer then. On the other hand, this is already written, so it might not be worth going back and redoing.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+        return succeed(None)
+
+    def pauseProducing(self):
+        pass
+
+    def stopProducing(self):
+        pass
+
+
+def submit(agent, result):
```
sarum90: If you intend to make this public and use it elsewhere, perhaps remove the `print` from this function (have that handled by main instead, and errback or callback the deferred depending on if it a success).

On the other hand, if you intend to keep this private, request naming it `_submit` instead of `submit`

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+        if response.code != http.CREATED:
+            print("Message: {}".format(json_body['message']), file=sys.stderr)
+        else:
+            print("Assigned result ID: {}".format(json_body['id']))
+
+    req.addCallback(process_response)
+    return req
+
+
+class Options(usage.Options):
```
sarum90: :grey_question: nit: Looks like we normally have class docstrings on usage.Options subclasses looking around our codebase.

benchmark/submit.py
```diff
@@ -0,0 +1,129 @@
...
+            options['port'],
+        ),
+        reactor,
+    )
+
+    try:
+        result_input = sys.stdin.read()
+        result = json.loads(result_input)
+    except ValueError as e:
+        print(result_input, file=sys.stderr)
```
sarum90: :grey_question:  Any reason for mixing sys.stderr.write and print(..., file=sys.stderr)? Slight preference for consistency within the file.

avg-I:
All the great suggestions, I'll do them all. Thank you!
I expected that the code would look quite a bit different in the end, so another review pass might be a good idea.

benchmark/submit.py
avg-I: Done.

benchmark/submit.py
avg-I: Yeah, I see this class used in an [example](https://twistedmatrix.com/documents/current/web/howto/client.html) in the Twisted documentation (and many more of its copies on the internet), but I couldn't find anything like it in the library.

benchmark/submit.py
avg-I: I kept this doc-string short, but I've now added `longdesc` field to the `Options` class.

benchmark/submit.py
avg-I: treq is more convenient to use, indeed.

benchmark/submit.py
avg-I: I think that we might want to use this functional in a master script that does everything to run a benchmark, from setting up the environment to posting the results. So, I am keeping the public name and I am changing the function to return its results in a more usable fashion.

benchmark/submit.py
avg-I: Yeah, and in most cases the dcostrings seem to be very obvious and tautological like
```
Arguments for ``initialize-release`` script
```
I think that that does not add much of useful information and is prone to getting out-of-sync with reality (e.g. because of renaming or copying).

benchmark/submit.py
avg-I: Fixed.

benchmark/submit.py
sarum90: Perfect, thanks.

benchmark/submit.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        raise SystemExit(1)
+
+    d = submit(server_url, result)
+
+    def succeeded(id):
+        print("Assigned result ID: {}".format(id))
+
+    def failed(failure):
+        print("Failed to submit the result: {}".format(failure.value),
```
sarum90: :grey_question: nit: I don't know if this was intentional or not, but you might consider exiting non-zero in these conditions.
```
flocker-review ♦ ~/dev/flocker-review (FLOC-3900-tool-for-submitting-benchmark-results)$ echo {} | python benchmark/submit.py --address 127.0.0.1
Failed to submit the result: No JSON object could be decoded
flocker-review ♦ ~/dev/flocker-review (FLOC-3900-tool-for-submitting-benchmark-results)$ echo $?
0
``

File: benchmark/submit.py [hunk 10]
```diff
@@ -0,0 +1,117 @@
...
+              file=sys.stderr)
```
sarum90: :grey_question: Just a little nit, definitely no action required for submission.

using netcat I decided to see how this would react with different responses from the server, some annotated results:

```
# Non 201 response, great logging:
Failed to submit the result: [200 OK] dog
# Server returns a quoted string instead of JSON object, sort of odd:
Failed to submit the result: string indices must be integers
# Server returns non 201, but resulting JSON object does not have "message" key, sort of odd:
Failed to submit the result: 'message'
# Server returns 201, but the resulting JSON object does not have "id" key, sort of odd:
Failed to submit the result: 'id'
```

Perhaps use something like https://twistedmatrix.com/documents/current/api/twisted.python.failure.Failure.html#getErrorMessage or https://docs.python.org/2/library/traceback.html to fix a large number of these without too much work.

sarum90:
Thanks, Looks great. It's a lot simpler with `treq` rather than using Twisted directly.

A couple comments about some error path handling, otherwise looks good.

benchmark/submit.py
avg-I: Good catch! Fixed.

benchmark/submit.py
avg-I: Interesting tests! I decided to do the nice reporting only for errors returned by the REST API, that is, for `SubmitFailure` and used `Failure.trap()` for that. I treat all other exceptions as unexpected conditions and let them propagate all the way up, they are reported in a more legible fashion now. I guess tat I could have the nice reporting for Twisted network errors as well, but it seems that there are quite a few of them and they do not seem to have a distinct common base.

PR #2485, opened 2016-01-18
[FLOC-3906] Remove Ubuntu 15.04 build target Dockerfile

zubron:
When releasing flocker-1.9.0, the `admin/publish-artifacts` step failed as it uses the presence of Dockerfiles within `admin/build_targets` to determine whether packages for a particular distribution need to be published. We no longer support Ubuntu 15.04 so this build target should be removed.


zubron:
See https://github.com/ClusterHQ/flocker/blob/remove-ubuntu-15-04-build-target-FLOC-3906/admin/packaging.py#L1098-L1115 for the relevant function which uses this information.

james-w:
Looks good, reducing the number of supported platforms is great.
Merge when mentor review has happened.

@jml mentor review please.

jml:
LGTM

PR #2486, opened 2016-01-18
[FLOC-3911] Add verbose logging to vagrant commands.

james-w:
We are hoping that more verbose vagrant logging may point to
why we are seeing failures.


jml:
LGTM.

PR #2488, opened 2016-01-19
[FLOC-3914] Integrate the CloudFormation installer with Volume Hub

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3914

We've generated a new CloudFormation template and uploaded it to https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json 

You can test this branch by logging in to Volume Hub and then starting the Cloud Formation installer supplying the VolumeHubToken when you get to the parameters page.

Or by running the automated tests from #2449 with a command like:

```
KEEP_STACK=1 \
CLOUDFORMATION_TEMPLATE_URL=https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json \
AWS_DEFAULT_REGION=us-west-1 \
KEY_PAIR="..." \
ACCESS_KEY_ID="..." \
SECRET_ACCESS_KEY="..." \
VOLUMEHUB_TOKEN="..." \
trial flocker.acceptance.endtoend.test_installer
```


admin/installer/setup_volumehub.sh
```diff
@@ -0,0 +1,38 @@
...
+
+control_service=false
+agent_node=false
+agent_number="0"
+
+if test "${node_number}" -eq "0"; then
+    control_service=true
+else
+    agent_node=true
+    # This only works because we start numbering CloudFormation nodes at 0.
```
mpgomez: Is that reliable, or is it likely to change? I don't know if there is a nice and reasonably straight-forward alternative.

admin/installer/setup_volumehub.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Set up Volume Hub if a token has been supplied.
+set -ex
+
+: ${volumehub_token:?}
+: ${node_number:?}
+
+if test -z "${volumehub_token}"; then
+    # No Volume Hub token supplied.
```
mpgomez: Is that logged somewhere? Should we get some kind of feedback or help saying how to properly call the script?

admin/installer/setup_volumehub.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+# Set up Volume Hub if a token has been supplied.
```
mpgomez: Do we need copyright on shell scripts too?

mpgomez:
It looks fine to me . I've made some comments, but I don't know if you need to take any action on those. That is all I could see.
I haven't tested it as I am not familiar with CloudFormation and volume hub.I guess today's demo in the engineering meeting will fix that :) I will try sometime this week though, as I want to learn how it works

admin/installer/setup_volumehub.sh
wallrj: Probably, but all these scripts get appended to one long cloud-init script so we'd get mulitple copyright lines....which isn't the end of the world.
But also, we may end up replacing this "appended shell script" mechanism with a single Python module that contains all the necessary helper function and a single copyright header.
I'll leave it for now.

admin/installer/setup_volumehub.sh
wallrj: That part was broken actually.
If you omitted the volumehub_token then the entire (combined) script would exit and the final wait condition callback didn't happen.

I got rid of that block.

admin/installer/setup_volumehub.sh
wallrj: We've mixed up 0 index and 1 index.
I'll export a new flocker_agent_number variable so that the script doesn't have to work it out its self.

wallrj:
Thanks @mpgomez 

All good points which I've addressed or answered.
Also merged in the (now merged) automated test from master and modified it to accommodate the volumehub_token parameter.
That revealed a bug in the setup script which I've also fixed.
The test isn't yet run by Jenkins but I verified that it passed locally with and without a volumehub_token.
So I'll merge this now.

PR #2489, opened 2016-01-19
[FLOC-3920] Increase timeout on node.functional.test_docker tests

james-w:
Unfortunately this means increasing the timeout for
all trial tests, or defining new stanzas for just the
test_docker module.

test_docker is currently timing out fairly frequently
because the tests are slow to run. Increasing the
timeout will allow more green runs to happen (if slower)
while we look in to whether we can speed up the tests,
or parallelise them to get the complete suite faster.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1062,7 +1062,10 @@ job_type:
       publish_test_results: true
       coverage_report: true
       clean_repo: true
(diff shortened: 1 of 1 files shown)


jml:
LGTM.

PR #2490, opened 2016-01-19
[FLOC-3804] Move the OSX installation test to run nightly.

james-w:
The test is more of an acceptance test, and is also unlikely
to be broken by any particular change, so running it nightly
is reasonable.

In addition it is currently failing often, perhaps related
to parallel versions of the job, so running once a day will
avoid us having to diagnose the root cause of that.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1277,45 +1277,6 @@ job_type:
           }
       clean_repo: true
       timeout: 30
-    run_client_installation_on_OSX:
-      on_nodes_with_labels: 'mesos-4GB'
-      with_steps:
-        - { type: 'shell',
-            cli: [ *hashbang,
-                   *add_shell_functions,
-
-                   *begin_build_sh_EOF,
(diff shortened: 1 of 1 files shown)


build.yaml
```diff
@@ -1435,6 +1396,46 @@ job_type:
             ]
           }
       timeout: 120
+    run_client_installation_on_OSX:
+      at: '0 6 * * *'
```
james-w: This is the only added line, the rest is the context, it has moved from
one block to the cronly_jobs block, which means that jobs.groovy
will run it nightly on master.

Azulinho: I would suggest here to run that job at 7:30, on the basis that the refresh OSX yosemite job runs at 5:00 and its timeout is 120minutes.
This job uses that VM

Azulinho:
just a recommendation on the schedule,
merge when happy

PR #2492, opened 2016-01-19
Report to github the status of individual jobs rather than the multijob.

james-w:
Report to github the status of individual jobs rather than the multijob.
    
This means that the information on the github page will be more useful,
as you can see which tests failed without clicking through and
searching. In addition, when a job is retried it will update its status
again, so a passing run will show all green on github even if it
required a job to be retried.
   
This commit changes the "pending" notifications to be for each test
we intend to run, and then each job does the successful/failing
notifications itself.
    
The github_status.sh script with the auth details isn't available
on the slaves, so this now embeds it in the script step.

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -122,18 +122,17 @@ GLOBAL_CONFIG = jsonSlurper.parseText(
 // --------------------------------------------------------------------------
 
 def updateGitHubStatus(status, description, branch, dashProject,
-                       dashBranchName) {
+                       dashBranchName, jobName) {
     return """#!/bin/bash
 # Jenkins jobs are usually run using 'set -x' however we use 'set +x' here
 # to avoid leaking credentials into the build logs.
 set +x
-# Read the GitHub credentials
-. /etc/slave_config/github_status.sh
-
+# Set the GitHub credentials
+""" + new File('/etc/slave_config/github_status.sh').text + """
 # Send a request to Github to update the commit with the build status
 REMOTE_GIT_COMMIT=`git rev-list --max-count=1 upstream/${branch}`
-BUILD_URL="\$JENKINS_URL/job/${dashProject}/job/${dashBranchName}/"
-PAYLOAD="{\\"state\\": \\"${status}\\", \\"target_url\\": \\"\$BUILD_URL\\", \\"description\\": \\"${description}\\", \\"context\\": \\"jenkins-multijob\\" }"
+BUILD_URL="\$JENKINS_URL/job/${dashProject}/job/${dashBranchName}/job/${jobName}/"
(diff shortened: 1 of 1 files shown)


jobs.groovy
```diff
@@ -122,18 +122,17 @@ GLOBAL_CONFIG = jsonSlurper.parseText(
...
+                       dashBranchName, jobName) {
     return """#!/bin/bash
 # Jenkins jobs are usually run using 'set -x' however we use 'set +x' here
 # to avoid leaking credentials into the build logs.
 set +x
-# Read the GitHub credentials
-. /etc/slave_config/github_status.sh
-
+# Set the GitHub credentials
+""" + new File('/etc/slave_config/github_status.sh').text + """
```
james-w: We embed this script because it's not available on the slaves, where the success/failure notification will now be sent from.

jml: Maybe this should be a comment in the groovy file?

jobs.groovy
```diff
@@ -379,6 +378,35 @@ def build_publishers(v) {
...
+                condition {
+                    status('FAILURE', 'FAILURE')
+                }
+                step {
+                    shell(updateGitHubStatus(
+                          'failure', 'Build failed',
+                          branchName, dashProject, dashBranchName, job_name))
+                }
+            }
+        }
```
jml: :question: Separate function?

jobs.groovy
```diff
@@ -600,14 +678,17 @@ def copy_artifacts_from(context, full_job_name, short_job_name, archive_artifact
     }
 }
 
+PULL_REQUEST_JOB_TYPES = ['run_trial', 'run_trial_for_storage_driver',
+                          'run_acceptance', 'run_sphinx', 'omnibus',
+                          'run_lint', 'run_client']
```
jml: :eyes: No action required. I guess that we're getting to the stage where "job type" should be an object with properties like "appear in pull requests".

jobs.groovy
```diff
@@ -675,26 +740,14 @@ def build_multijob(dashBranchName, branchName, isReleaseBuild) {
...
-                           produce them */
-                        if (job_values.archive_artifacts) {
-                            _job_name = job_entry.key + '_' + escape_name(_module)
-                            copy_artifacts_from(
-                                delegate, full_job_name(dashProject, dashBranchName, _job_name),
-                                 _job_name, job_values.archive_artifacts
-                            )
-                        }
-                    }
+            list_jobs(dashProject, dashBranchName).findAll { it.job_type != 'cronly_jobs' }.each {
```
jml: Shouldn't this be `it.type` rather than `it.job_type`? More worryingly, shouldn't this cause something to fail / explode?

jml:
Thanks. I think this is nicer (although I long for the day when an aggregate thing is enough!).

One potential bug that should be fixed plus some minor comments. Merge when happy.

PR #2493, opened 2016-01-19
[FLOC-3926] Remove duplicate documentation.

sarum90:
The AWS and the Rackspace drivers had duplicate documentation with the actual docs at http://doc-dev.clusterhq.com/. They also had links to that documentation. Removing the duplicate in favor of the reference to the official documentation.


jongiddy:
A good idea.

Can you please add `dict` as the type for `provider_config`, and describe the parameter as a dictionary parsed from the (platform) section of the acceptance testing configuration file.

sarum90:
Thanks for the comment, fixed. I'm going to wait til I get a green tests for lint before I submit.

sarum90:
Lint passed and this only changes docstrings. I'm going to submit as is.

PR #2494, opened 2016-01-19
[FLOC-3813] Run more tests that require root.

itamarst:


build.yaml
```diff
@@ -1002,13 +1002,16 @@ run_trial_modules: &run_trial_modules
...
 run_trial_as_root_modules: &run_trial_as_root_modules
+  # iptables requires root
+  - flocker.route.functional
+  # journald access requires root
+  - flocker.common.functional
   # flocker.volume needs to run as root due to ZFS calls
   # so we run it as part of the run_trial_on_<Cloud>_<OS>_as_root jobs
   - flocker.volume
   # flocker.node.agent needs to run as root due to mounting filesystems
   - flocker.node.agents
```
Azulinho: if these need to run as root,
shouldn't you instead move flocker.route, and flocker.common out of the non-root run_trial block ?
https://github.com/ClusterHQ/flocker/blob/master/build.yaml#L982-L1003

and move those into this block instead ?

Or is there any advantage in running both, one as root and another as non-root ?

itamarst: There's e.g. `flocker.common.test` too, which doesn't need root. So some minor duplication but skips are fast, typically, so not a big deal.

Azulinho:
ok, 
I was thinking of costs, as its an additional builder.
merge it when happy

Azulinho:
ok, 
I was thinking of costs, as its an additional builder.
merge it when happy

Azulinho:
ok, 
I was thinking of costs, as its an additional builder.
merge it when happy

Azulinho:
ok, 
I was thinking of costs, as its an additional builder.
merge it when happy

PR #2495, opened 2016-01-20
[FLOC-3929] Adding docs for setup container script

mpgomez:


docs/gettinginvolved/cluster-setup.rst
```diff
@@ -199,6 +171,79 @@ Or you can run, for example, the acceptance tests against the created cluster:
      --flocker-version='' \
      flocker.acceptance.obsolete.test_containers.ContainerAPITests.test_create_container_with_ports
 
+==============================
+Adding containers and datasets
+==============================
+To make it easier to reuse a test cluster, and test under different configurations, Flocker provides a tool to create a certain number of datasets and containers per node.
```
jml: "re-use". No need for comma after "test cluster"

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -199,6 +171,79 @@ Or you can run, for example, the acceptance tests against the created cluster:
...
+It is called like this:
```
jml: @davidoliverSP2 Do we have a more usual way of saying this?

jml:
LGTM, but assigning to @davidoliverSP2 for another pass.

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -199,6 +171,79 @@ Or you can run, for example, the acceptance tests against the created cluster:
      --flocker-version='' \
      flocker.acceptance.obsolete.test_containers.ContainerAPITests.test_create_container_with_ports
 
+==============================
```
davidoliverSP2: This is a sub-section, so doesn't require the ===== above the title (just below it)
We also use title case on our headings, so the title should be "Adding Containers and Datasets"

docs/gettinginvolved/cluster-setup.rst
davidoliverSP2: We would normally say "Run the following command to xxxx", but this is probably ok - I don't think the user would be confused by the wording here. I'll leave this one up to you @mpgomez

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -199,6 +171,79 @@ Or you can run, for example, the acceptance tests against the created cluster:
...
+   or, in other words, for the cluster to converge.
+   If this parameter is not set, then the program will wait up to two hours.
+
+If :option:`--wait` is used the script waits for the deletions to take effect.
+After the script successfully finishes the cluster should be in a converged state
+with the requested containers and datasets.
+If :option:`--wait` is not specified, then the script will wait for up to two hours.
+
+Application Template
+====================
```
davidoliverSP2: If you want "Application Template" to be a sub-heading of "Adding Containers and Datasets", then you can use ----- instead of =====. As it stands, Application Template is a peer of Adding Containers etc.

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -199,6 +171,79 @@ Or you can run, for example, the acceptance tests against the created cluster:
...
+The configuration file given for the ``--app-template`` parameter describes a single application.
+At the very least it should specify a name of a Docker image to use for an application container.
+
+.. code-block:: yaml
+
+  image: "clusterhq/mongodb"
+  volume:
+    mountpoint: "/data/db"
+
+See :doc:`../control/cli/application-config` for more details.
```
davidoliverSP2: The recommended way of linking to another topic is via a ref tag. 

I noticed that control/cli/application-config doesn't have a ref tag, so I would suggest adding one:

On the 1st line of control/cli/application-config, add the following:
.. _application-configuration:

Then in this document replace :doc:`../control/cli/application-config` with :ref:`application-configuration`

If you need any help, please let me know!

docs/gettinginvolved/cluster-setup.rst
davidoliverSP2: Or is it a sub-section? I've just noticed that some headings in this topic are classed as top level headers, so maybe the syntax here is correct? Your call @mpgomez

davidoliverSP2:
a few recommended changes, ignore my comments about headings if you think its labelled correctly, but I'm happy once the ref tag has been added for link to Application Configuration.

docs/gettinginvolved/cluster-setup.rst
mpgomez: Happy to reword it @davidoliverSP2 . I am not great getting the best way of explaining stuff in English - all I can learn on that is more than welcome!

docs/gettinginvolved/cluster-setup.rst
mpgomez: Ops, no coma before and. @jml question about re-use. In word reference appears as reuse, and Miriam Webster doesn't find "re-use" and suggests "reuse" instead.
http://www.merriam-webster.com/dictionary/suggestions/re-use
Where is the place to look at that stuff?

docs/gettinginvolved/cluster-setup.rst
mpgomez: Cleaning up the cluster configuration was a new section as well, so I thought that setting up the cluster configuration would have to be one as well? I tried to put them as just subsections, but they have overlapping parameters, so sphinx would complain. I am leaving it like that for now. @davidoliverSP2 let me know if you want me to change it to something else, I will need to update the docs again as part of the PR I'm currently working on :)

docs/gettinginvolved/cluster-setup.rst
mpgomez: Changed! The application template is going away in the next PR anyway

PR #2496, opened 2016-01-20
[FLOC-3930] Flaky audit script

jml:
New script to dump all of the flaky tests in Flocker.

```
find-flaky-tests --report=tree flocker
```

Will show the flaky tests as a trial output tree. --bugs and --tests will show only bugs or only tests.

### DIFF
--- a/admin/flaky.py
+++ b/admin/flaky.py
@@ -66,9 +66,9 @@ def report_flaky_tests(output, flaky_tests):
         output.write('{}\n'.format(test))
 
 
-def report_bugs(output, flaky_tests):
+def report_issues(output, flaky_tests):
     """
-    Print all bugs for flaky tests.
+    Print all issues for flaky tests.
     """
     jira_keys = frozenset().union(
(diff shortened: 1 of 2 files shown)


jongiddy:
Print a useful message if no suites are specified. Possibly the usage message.

The usage message should have `suites...` in it. Currently, it only mentions the `--report` option, but not the positional arguments.

admin/flaky.py
```diff
@@ -0,0 +1,156 @@
...
+    for (test, flaky) in flaky_tests:
+        new_test = clone_test_with_new_id(test, '{}({})'.format(test.id(), ', '.join(flaky.jira_keys)))
+        reporter.startTest(new_test)
+        reporter.addSuccess(new_test)
+        reporter.stopTest(new_test)
+    reporter.done()
+
+
+_REPORTS = {
+    'bugs': report_bugs,
```
jongiddy: :question: "bugs" -> "issues"

jongiddy:
Seems to work well. Please consider the above comments, but OK to merge.

PR #2497, opened 2016-01-20
[FLOC-3935] Give Docker client test an hour to run

jml:
Justification in the comments in the code.

### DIFF
--- a/flocker/node/functional/test_docker.py
+++ b/flocker/node/functional/test_docker.py
@@ -6,6 +6,7 @@ Functional tests for :module:`flocker.node._docker`.
 
 from __future__ import absolute_import
 
+from datetime import timedelta
 from functools import partial
 import time
 import socket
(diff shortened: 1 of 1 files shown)


exarkun:
Thanks!

Can you describe what the next step in improving this part of the test suite might be?  In particular, what do you think we should do if either:

(a) this change causes this test module to stop having test timeouts completely
(b) this change causes individual tests to hang for the full timeout (ie, very much longer than their current timeouts allow) before the job is killed?

Code change itself looks like it accomplishes the intent.  So go ahead and merge when you're ready and CI shows sufficiently good results.

jml:
(a) later, as a lower priority task, we make these tests faster (as @itamarst often says, we want to delete this code)
(b) we instrument the crap out of the tests, and probably lower the timeout back (depending on what improves developer experience)

But, also, I want to reproduce the `KeyboardInterrupt` problem, as I suspect it's an issue with testtools. That will require running the tests with the current timeouts in a branch.

PR #2498, opened 2016-01-20
[FLOC-3927] Configure dependency logging using agent.yml

jongiddy:
For debugging and benchmarking, it is useful to configure the logging provided by dependencies, which use the standard Python logging rather than Eliot.

This PR adds a `logging` stanza to the `agent.yml` file, which follows the PEP 391 format for specifying logging configuration.

For benchmarking AWS API calls, for example, we can add the following stanza to the YAML configuration file (`acceptance.yml` or equivalent):
```yaml
logging:
   version: 1
   formatters:
      timestamped:
         format: "%(asctime)s %(message)s"
         datefmt: "%Y-%m-%d %H:%M:%S"
   handlers:
      boto2:
         class: logging.handlers.WatchedFileHandler
         level: DEBUG
         filename: /tmp/boto2.log
         encoding: utf-8
         formatter: timestamped
      boto3:
         class: logging.handlers.WatchedFileHandler
         level: DEBUG
         filename: /tmp/boto3.log
         encoding: utf-8
         formatter: timestamped
   loggers:
      boto:
         handlers: ["boto2"]
         level: DEBUG
      boto3:
         handlers: ["boto3"]
         level: DEBUG
```

For ease of review, the changes can be divided into three parts:
1. changes in `flocker.node` which allow the agents to parse the `agent.yml` and setup logging;
2. changes in `flocker.provision` and `admin` which provision nodes with an `agent.yml` generated from a YAML config file;
3. changes in `docs`, which describe the additions in the YAML file.

There is also a fix in `flocker.provision` to the Put effect provider function, to escape backslashes and per-cent symbols correctly.

### DIFF
--- a/flocker/node/test/test_script.py
+++ b/flocker/node/test/test_script.py
@@ -66,8 +66,7 @@ def setup_config(test, control_address=u"10.0.0.1", control_port=1234,
     if name is None:
         name = random_name(test)
     ca_set = get_credential_sets()[0]
-    scratch_directory = FilePath(test.mktemp())
-    scratch_directory.makedirs()
+    scratch_directory = test.make_temporary_directory()
     contents = {
         u"control-service": {
             u"hostname": control_address,
@@ -269,7 +268,7 @@ class AgentServiceFromConfigurationTests(TestCase):
         host = b"192.0.2.13"
         port = 2314
         name = u"from_config-test"
-        logfile = self.make_temporary_directory().child('logfile')
+        logfile = self.make_temporary_path()
         log_config = {
             'version': 1,
             'handlers': {
--- a/flocker/node/test/test_script.py
+++ b/flocker/node/test/test_script.py
@@ -4,6 +4,7 @@
 Tests for :module:`flocker.node.script`.
 """
 
+import logging
 import socket
 from unittest import skipUnless
 
@@ -45,7 +46,7 @@ from .dummybackend import DUMMY_API
 
 
 def setup_config(test, control_address=u"10.0.0.1", control_port=1234,
-                 name=None):
+                 name=None, log_config=None):
     """
     Create a configuration file and certificates for a dataset agent in a
     temporary directory.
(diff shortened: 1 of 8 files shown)


flocker/node/test/test_script.py
```diff
@@ -229,6 +234,58 @@ def test_initialized(self):
             ),
         )
 
+    def test_logging(self):
+        """
+        Logging is setup by a logging stanza.
```
jml: Nit: "set up" ("setup" is a noun)

flocker/node/test/test_script.py
```diff
@@ -229,6 +234,58 @@ def test_initialized(self):
...
+            for handler in logger.handlers[:]:
+                logger.removeHandler(handler)
+            for handler in handlers:
+                logger.addHandler(handler)
+
+        logger = logging.getLogger()
+        self.addCleanup(
+            restore_logger, logger, logger.getEffectiveLevel(),
+            logger.handlers[:]
+        )
```
jml: :question: This might be a better done as a Fixture.

flocker/node/test/test_script.py
```diff
@@ -229,6 +234,58 @@ def test_initialized(self):
...
+
+        # Setup an AgentService with a logging stanza
+        host = b"192.0.2.13"
+        port = 2314
+        name = u"from_config-test"
+        logfile = self.make_temporary_directory().child('logfile')
```
jml: `make_temporary_file` would work too, right? :question:

flocker/node/test/test_script.py
```diff
@@ -229,6 +234,58 @@ def test_initialized(self):
...
+        )
+        options = DatasetAgentOptions()
+        options.parseOptions([b"--agent-config", self.config.path])
+        config = get_configuration(options)
+        AgentService.from_configuration(config)
+
+        # Root logger now logs to file
+        log_message = 'My LoG tEsT.'
+        logger.info(log_message)
+        self.assertIn(log_message, logfile.getContent())
```
jml: :question: Alternatively:

```python
self.assertThat(logfile, file_contents(Contains(log_message)))
```

flocker/provision/test/test_ssh_conch.py
```diff
@@ -56,12 +56,12 @@ def test_put(self):
...
         The ``Put`` intent puts the provided contents in the specified file.
         """
-
+        content = '\\hell%o'
         command = run_remotely(
             username="root",
             address=str(self.server.ip),
             port=self.server.port,
-            commands=put(content="hello", path="file"),
+            commands=put(content=content, path="file"),
```
jml: This would be a good use case for hypothesis. :question:

flocker/provision/test/test_ssh_conch.py
```diff
@@ -71,7 +71,7 @@ def test_put(self):
 
         def check(_):
             self.assertEqual(self.server.home.child('file').getContent(),
-                             "hello")
+                             content)
```
jml: :question: Alternatively:

```python
self.assertThat(self.server.home.child('file'), file_contents(Equals(content)))
```

Better error message at least. *shrug*

flocker/node/test/test_script.py
```diff
@@ -58,29 +59,33 @@ def setup_config(test, control_address=u"10.0.0.1", control_port=1234,
...
     :param unicode name: The ZFS pool name.  If ``None``, a random one will be
         generated that identifies the given test.
+    :param dict log_config: A logging configuration dictionary. If ``None``,
+        no logging stanza will be added to the configuration file.
     """
     if name is None:
         name = random_name(test)
     ca_set = get_credential_sets()[0]
     scratch_directory = FilePath(test.mktemp())
     scratch_directory.makedirs()
```
jml: :question: Can now be test.make_temporary_directory()

flocker/node/test/test_script.py
```diff
@@ -58,29 +59,33 @@ def setup_config(test, control_address=u"10.0.0.1", control_port=1234,
...
-                u"volume_config_path": scratch_directory.child(
-                    b"volume_config.json"
-                ).path,
-            },
-            u"version": 1,
-        }))
+    test.config.setContent(yaml.safe_dump(contents))
     ca_set.copy_to(scratch_directory, node=True)
     test.ca_set = ca_set
     test.non_existent_file = scratch_directory.child('missing-config.yml')
```
jml: :question: Probably shouldn't be done in this branch, but not a huge fan of setting attributes on the test object. When looking at an individual test, it makes it much harder to see where the fixtures are coming from (http://xunitpatterns.com/Obscure%20Test.html#Mystery Guest).

jml:
Thanks! Lots of comments, but they're all pretty minor, and most are matters of opinion. Merge after addressing & when tests pass.

flocker/node/test/test_script.py
jongiddy: Yeah, I tried that initially. But because this test then goes on to call `setup_config` which uses `mktemp()`, it seemed like I was creating a file and a directory with the same name. Since it actually seemed to work, maybe I was mistaken, but I went with a safer version.

flocker/provision/test/test_ssh_conch.py
jongiddy: I'll tell you after 9th February. :smile:

flocker/node/test/test_script.py
jongiddy: OK, looked at the testtools code again, and can see that each call to `make_*` creates a new directory for the contents.

flocker/provision/test/test_ssh_conch.py
jongiddy: I'm leaving this for the moment because a similar test occurs elsewhere in the file, and consistency is good.

PR #2499, opened 2016-01-20
[FLOC-3937] Remove stale flaky decorators

jml:
All of these are now closed. Future errors are real errors.


exarkun:
Awesome, thanks.  Merge when these pass on CI.

PR #2501, opened 2016-01-21
[FLOC 3948] simplify setup cluster

avg-I:
Because main() is decorated with @inlineCallbacks there is no guarantee
that the control flow always returns to it after a yield statement.
In particular, that's the case if the reactor is stopped.
And that could happen, for instance, if a user presses Ctrl-C because
twisted installs a SIGINT handler which stops the reactor.

Thus, to improve the chances of the cluster clean-up being executed
the clean-up code is moved to a before-shutdown system event trigger.

Also, there is no reason now to have except and finally clauses as
(a) there is no guarantee that they are executed, (b) all the except
block did was to re-raise an exception, so no reason to capture it
at all, (c) the finally block would contain only actions that should
be executed on success after the clean-up code was moved to the
shutdown handler.

### DIFF
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -208,6 +208,9 @@ def main(reactor, args, base_path, top_level):
                                            node.address,
                                            remote_logs_file)
                            )
+    # gather_deferreds() below does more than just creating a DeferredList.
+    # So, while we are not using its result in any way, it is still useful
+    # because it logs any failures coming from the gathered deferreds.
     gather_deferreds(results)
 
     if options['no-keep']:
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -177,66 +177,61 @@ def main(reactor, args, base_path, top_level):
     reactor.addSystemEventTrigger(
         'before', 'shutdown', log_writer.stopService)
 
-    cluster = None
+    def cluster_cleanup():
+        print("stopping cluster")
+        return runner.stop_cluster(reactor)
+
+    cleanup_trigger_id = reactor.addSystemEventTrigger('before', 'shutdown',
+                                                       cluster_cleanup)
+
     results = []
-    try:
-        yield runner.ensure_keys(reactor)
-        cluster = yield runner.start_cluster(reactor)
+    yield runner.ensure_keys(reactor)
+    cluster = yield runner.start_cluster(reactor)
 
-        managed_config_file = options['cert-directory'].child("managed.yaml")
-        managed_config = create_managed_config(options['config'], cluster)
-        managed_config_file.setContent(
-            yaml.safe_dump(managed_config, default_flow_style=False)
-        )
+    managed_config_file = options['cert-directory'].child("managed.yaml")
+    managed_config = create_managed_config(options['config'], cluster)
+    managed_config_file.setContent(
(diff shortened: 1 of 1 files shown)


admin/cluster_setup.py
```diff
@@ -177,66 +177,64 @@ def main(reactor, args, base_path, top_level):
...
+        print("stopping cluster")
+        return runner.stop_cluster(reactor)
 
-        managed_config_file = options['cert-directory'].child("managed.yaml")
-        managed_config = create_managed_config(options['config'], cluster)
-        managed_config_file.setContent(
-            yaml.safe_dump(managed_config, default_flow_style=False)
-        )
+    cleanup_trigger_id = reactor.addSystemEventTrigger('before', 'shutdown',
+                                                       cluster_cleanup)
```
exarkun: Triggers run in the order they're added.  I bet it would be good if we did the cluster cleanup before we stopped the log writer rather than after.

admin/cluster_setup.py
```diff
@@ -177,66 +177,64 @@ def main(reactor, args, base_path, top_level):
...
+        remote_logs_file = open("remote_logs.log", "a")
+        for node in cluster.all_nodes:
+            results.append(capture_upstart(reactor,
+                                           node.address,
+                                           remote_logs_file)
+                           )
+    # gather_deferreds() below does more than just creating a DeferredList.
+    # So, while we are not using its result in any way, it is still useful
+    # because it logs any failures coming from the gathered deferreds.
+    gather_deferreds(results)
```
exarkun: :question: The idiomatic (and just less confusing) way to do this would be to change the above calls to `capture_upstart` and `capture_journal` to be like `capture_xxx(...).addErrback(write_failure)`.  Then `results` isn't even needed anymore, afaik.  Since this is how it worked before and not really related to this refactoring, optional.

admin/cluster_setup.py
```diff
@@ -177,66 +177,64 @@ def main(reactor, args, base_path, top_level):
...
+
+    if options['no-keep']:
+        print("not keeping cluster")
```
exarkun: I'm a bit confused by this functionality.  If you pass `--no-keep` then as soon as the cluster is spun up it is immediately destroyed?  I guess that's the behavior this code provided before so the refactoring here isn't a problem... But I don't see the point of doing this.  Maybe better documentation about the `--no-keep` option would help.

exarkun:
Thanks.  Only those minor comments inline.  I tried this and it started and provisioned a two node cluster and then cleaned it up.  Address to your satisfaction.

exarkun:
Then merge.

admin/cluster_setup.py
avg-I: Indeed.

admin/cluster_setup.py
avg-I: Being explicit is so much nicer that I went ahead and implemented your suggestion. Thanks!

admin/cluster_setup.py
avg-I: The option is described as "Do not keep VMs around (when testing)". My only use for it was to test that the script successfully runs (I didn't actually need a cluster).
Not sure if it's worthwhile keeping the option.

PR #2502, opened 2016-01-21
[FLOC-3949] wait until all provisioned nodes join the newly created flocker cluster

avg-I:
To be sure that the cluster is completely ready to use after setup-cluster finishes.

### DIFF
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -207,7 +207,7 @@ def main(reactor, args, base_path, top_level):
                             remote_logs_file).addErrback(write_failure)
 
     flocker_client = _make_client(reactor, cluster)
-    yield _wait_for_nodes(flocker_client, len(cluster.agent_nodes))
+    yield _wait_for_nodes(reactor, flocker_client, len(cluster.agent_nodes))
 
     if options['no-keep']:
         print("not keeping cluster")
@@ -254,10 +254,11 @@ def _make_client(reactor, cluster):
                          cluster_cert, user_cert, user_key)
 
 
-def _wait_for_nodes(client, count):
+def _wait_for_nodes(reactor, client, count):
     """
     Wait until nodes join the cluster.
 
+    :param reactor: The reactor.
     :param flocker.apiclient.FlockerClient client: The client connected to
         the cluster (its control node).
     :param int count: The expected number of nodes in the cluster.
@@ -276,4 +277,4 @@ def _wait_for_nodes(client, count):
         d.addCallback(check_node_count)
         return d
 
-    return loop_until(client._reactor, got_all_nodes, repeat(1, 120))
+    return loop_until(reactor, got_all_nodes, repeat(1, 120))
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -266,7 +266,7 @@ def _wait_for_nodes(client, count):
     """
     def got_all_nodes():
         d = client.list_nodes()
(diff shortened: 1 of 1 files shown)


admin/cluster_setup.py
```diff
@@ -240,3 +246,46 @@ def main(reactor, args, base_path, top_level):
...
+
+        def check_node_count(nodes):
+            print("Waiting for nodes, "
+                  "got {} out of {}".format(len(nodes), count))
+            return len(nodes) == count
+
+        d.addCallback(check_node_count)
+        return d
+
+    return loop_until(client._reactor, got_all_nodes, repeat(1, 120))
```
exarkun: `_reactor` is a private attribute of the client so `_wait_for_nodes` really shouldn't use it.  `_`-prefix is just a convention but ignoring it defeats the purpose.  If the reactor from the client absolutely must be used (and maybe it does, I'm not sure yet, just starting to read this PR) then a strictly better solution is to rename `_reactor` to `reactor` to reflect the fact that it is actually part of the client's public interface.

admin/cluster_setup.py
```diff
@@ -205,6 +208,9 @@ def main(reactor, args, base_path, top_level):
                                )
         gather_deferreds(results)
 
+        flocker_client = _make_client(reactor, cluster)
+        yield _wait_for_nodes(flocker_client, len(cluster.agent_nodes))
```
exarkun: But since you already have the `reactor` local here, perhaps better than renaming `flocker_client._reactor` to `flocker_client.reactor` would be to have `_wait_for_nodes` accept the reactor as a separate parameter and pass it in explicitly at this call site.

admin/cluster_setup.py
exarkun: I'm starting to cringe at `loop_until` uses because `loop_until` fails somewhat opaquely.  This isn't really a problem with this PR, more a problem with the timeout behavior of `loop_until`.  Maybe I should go file an issue about that ...

So I guess this comment isn't actionable.

admin/cluster_setup.py
```diff
@@ -240,3 +246,46 @@ def main(reactor, args, base_path, top_level):
...
+
+    :param flocker.apiclient.FlockerClient client: The client connected to
+        the cluster (its control node).
+    :param int count: The expected number of nodes in the cluster.
+    :return: ``Deferred`` firing when the number of nodes in the cluster
+        reaches the target.
+    """
+    def got_all_nodes():
+        d = client.list_nodes()
+        d.addErrback(write_failure, logger=None)
```
exarkun: `logger` defaults to `None` nowadays so you don't actually have to pass it in explicitly anymore.

exarkun:
I guess this is going to conflict with the other outstanding PR for cluster_setup.py changes.  Usually I'll try doing "stacked" branches (ie, branch master to fix X, branch *that* branch to fix Y) in this case to avoid (sometimes) having to resolve the conflicts.

Just a couple interesting comments inline, I think.  Address those, resolve conflict (unless I'm wrong about that), and then merge.  Thanks.

admin/cluster_setup.py
avg-I: Yes, I am doing this.

admin/cluster_setup.py

admin/cluster_setup.py

avg-I:
Yes, there is a simple, mechanical conflict because of the changed indentation.

PR #2503, opened 2016-01-21
[FLOC-3042] Canonical docs

dwgebler:
This is the proposed fix for https://clusterhq.atlassian.net/browse/FLOC-3042
Also fixes [FLOC-3915] https://clusterhq.atlassian.net/browse/FLOC-3915

You can read about the strategic intent of this work in respect of `<link rel="canonical" ...>` on the FLOC-3042 issue description and comments. The idea here is that when we publish a version - for the sake of this example, 1.10.0, the docs are:

- published to `/en/1.10.0` with a canonical link pointing to `/en/latest/`
- also published on `/en/latest/` with a canonical link self-referencing `/en/latest/`

Then, when we release 1.11.0:

- the 1.10.0 docs remain at `/en/1.10.0` and the canonical link is updated to `/en/1.10.0`
- the 1.11.0 docs are published at `/en/1.11.0` with the canonical link pointing to `/en/latest/`
- the 1.11.0 docs are also published to `/en/latest/` with the canonical link pointing to `/en/latest/`

To make the process of releasing Flocker and publishing docs easier, more maintainable and less error-prone for us (i.e. to not have to separately build and publish two different versions of the docs for each release), I have employed a methodology in this branch of using JavaScript to insert an `/en/latest/` canonical link (including the full requested path) in to all published docs pages, which is then dynamically changed to `/en/x.y.z` in the event that our existing code which checks the docs version being viewed against the latest version (to produce the out-dated docs admonition notice) determines that `x.y.z` is no longer the latest version. It's important to note here: conventional web wisdom will tell you that we should be using statically generated `<link>` elements in the HTML, however while this was once unarguably true, modern Googlebot can and will execute the JavaScript correctly and internally render the page to itself with the desired dynamic `<link>` elements. That is, Google can "see" the `<link>` elements we dynamically insert and treats them the same way as if they were part of the static HTML delivered in the HTTP response body. This is something I have experimentally verified via Google's own tools prior to going ahead with the JS method.

Because the broad strategy for this issue is to actually publish the docs to both `/en/latest/` and `/en/x.y.z/` locations in our S3 buckets (i.e. copy the files twice), the script for publishing the docs has also been modified to reflect those changes - we will no longer 302 redirect `/en/latest` to `/en/x.y.z`.

The necessary changes to the publishing script also therefore fix FLOC-3915, as previous changes to our docs hosting on S3 had meant that script was already broken.

I have updated the unit tests for the publishing script in accordance with the code changes, however it's important to note there has been no live testing of this new process - something we can't really do without performing another release and watching the process closely.

I'm not 100% clear on the effect of these changes in respect of dev/non-production publishing of docs - I'm not sure it matters, in that we don't seem to use `docs.clusterhq.com/en/devel/` any more. In fact, that should just redirect to `doc-dev.clusterhq.com`, which is the real canonical source for in-development version documentation and published automatically by Jenkins after each build of `master`.


jml:
Thanks for looking into this.

Code changes all look good, but I have some questions about the behaviour:

* If we have a document `foo` that links to `bar` in 1.10, but in 1.11 we delete the `bar` docs entirely, what happens when someone clicks the 'bar' link in `/en/1.10/foo.html`?
* If we have a document `foo` that links to `bar` in 1.10, but in 1.11 we rename `bar` to `baz`, what happens when someone clicks the 'baz' link in `/en/1.10/foo.html`?

As discussed, it'd be a good idea if @msh100 took a look at this before we go live.

dwgebler:
@jml good questions

> If we have a document foo that links to bar in 1.10, but in 1.11 we delete the bar docs entirely, what happens when someone clicks the 'bar' link in /en/1.10/foo.html?

When we publish 1.11.0, the canonical link in `https://docs.clusterhq.com/en/1.10.0/foo.html` will point to itself, i.e. `<link rel="canonical" href="https://docs.clusterhq.com/en/1.10.0/foo.html"/>`, because the JS will detect that 1.10.0 is no longer the `en/latest/` documentation version. The changes in this branch don't affect what would happen when a user clicks a link (or a bot follows a link) - these would still point to the existing `/en/1.10.0/bar.html`.

> If we have a document foo that links to bar in 1.10, but in 1.11 we rename bar to baz, what happens when someone clicks the 'baz' link in /en/1.10/foo.html?

Likewise for this question, the links will remain valid and point to whatever published version of the docs you're looking at (so looking at `/en/latest/foo.html`, a clickable link will point to `/en/latest/bar.html` and looking at `/en/1.10.0/foo.html`, a clickable link will point to `/en/1.10.0/bar.html`. The only difference this branch makes is that when we publish 1.10.0 for example, for the time 1.10.0 is identical to the content copied to `en/latest/`, when Googlebot crawls `en/1.10.0/foo.html`, it will see the canonical location of the page is `en/latest/foo.html`.

The result should be that Google always updates its indexes for `en/latest/` when we publish a release and favours newer `en/x.y.z` versions over older ones (because for the time they were up-to-date, the `x.y.z` versions pointed their canonical location to `latest`)

docs/_themes/clusterhq/layout_docs.html
```diff
@@ -256,10 +261,12 @@
 
             // Is version equal to this documentation release?
             if (currentversion !== DOCUMENTATION_OPTIONS.VERSION) {
+                $('link[rel="canonical"]').attr('href', 'https://docs.clusterhq.com/en/' + DOCUMENTATION_OPTIONS.VERSION + '/' + canonicalLocation);
```
msh100: My only concern here is if the documentation is non-latest, the canonical link is going to be defined as latest and then changed shortly after. I know you (@dwgebler) said that Google respects canonical links defined by JS however do you know how it would react to this?

I'd recommend removing the contents of line 255, and moving to an else statement as to avoid it changing during execution.

dwgebler: > Google respects canonical links defined by JS however do you know how it would react to this?

The experimental result from Google's tools is that it is the updated canonical link here that gets rendered by Googlebot. So, what you're suggesting is interesting because it's something I also considered - the thinking behind this way was that if in the (rare) event the AJAX call to the version indicator fails, we still have a meaningful canonical link for Google to promote `latest`, which is better than none.

However, a better approach _might_ be to assume we should canonicalize the versioned location rather than `latest` in the event that we fail to call out to `version.html`, which we could do via a failure callback on the AJAX request, as well as in the event that it succeeds and we find `x.y.z` is no longer `latest`. Not because there's any reason to think updating the canonical via JS won't be respected by Google (certainly Googlebot is aware of the update), but just to err on the side of caution and ensure we don't by failure canonicalise latest for resources that may not exist in latest or be different content.

msh100:
:+1: I am happy with the JS here - lets hope this helps :smile:

jml:
LGTM. Thanks.

dwgebler:
Note the functional Docker test build failure is unrelated and a known timing-out green CI issue.

PR #2504, opened 2016-01-21
[FLOC-3950] Fix attribute error in ManagedRunner

jongiddy:
FLOC-3927 #2498 unfortunately assumed that `ManagedRunner` followed the structure of `LibcloudRunner`, and used `self.config`, which is not a valid attribute for `ManagedRunner`. @avg-I spotted that this caused an `AttributeError`

This PR changes the `ManagedRunner` class to receive the `logging_config` dictionary directly.  It's use has now been independently tested by both @avg-I and me.


exarkun:
Thanks.  Go ahead and merge.

PR #2507, opened 2016-01-22
[FLOC-3958] Remove --testmodule from storage driver tests.

james-w:
The test definitions for those tests now specify the test module
directly, so don't try and use --testmodule to find it.


jml:
No issues w/ the code. Please make sure it passes before landing.

PR #2509, opened 2016-01-22
[FLOC-3889] Benchmark - remove use of compose 

mpgomez:
Compose was used in the container setup script to create datasets and containers in a cluster. As it is deprecated, I have refactored the code to use the flocker client instead. That means that:
* app-template yaml file is no longer used: we only needed to know the Docker image, mountpoint, and we could specify the maximum size of the dataset. Anything else in the yaml file would be ignored, plus it didn't make sense to use a format that doesn't match with what we need to call create_dataset and create_container.
* The options that we do need have been added as command line arguments. That will, in addition, make it easier to integrate this script with the other ones, as we are aiming to have one single script that will do the benchmarking, including creating the cluster and setting up the containers. The parameters added are the docker image and the mountpoint (and an optional volume size)
* An extra improvement has been made to the way the scripts wait until the containers are created. In order to avoid overloading the control service with list calls while looping, instead of listing every second, now we list every four seconds. That waiting time can also be changed passing the new value as a command line argument

### DIFF
--- a/benchmark/cluster_containers_setup.py
+++ b/benchmark/cluster_containers_setup.py
@@ -48,9 +48,9 @@ class ContainerOptions(usage.Options):
          'Location of the user and control certificates and user key'],
         ['max-size', None, 1,
          'Size of the volume, in gigabytes. One GB by default'],
-        ['wait', None, 72000,
+        ['wait', None, 7200,
          "The timeout in seconds for waiting until the operation is complete. "
-         "No waiting is done by default."],
+         "Waits two hours by default."],
         ['wait-interval', None, 4,
          "How often are we going to check if the creation of containers and "
          "datasets has finished, in second"],
@@ -150,14 +150,14 @@ class ClusterContainerDeployment(object):
 
     :ivar options: ``ContainerOptions`` with the options passed to the script
     :ivar image: ``DockerImage`` for the containers.
-    :ivar max_size: maximum volume (dataset) size.
+    :ivar max_size: maximum volume (dataset) size in bytes.
     :ivar mountpoint: unicode string containing the absolute path of the
         mountpoint.
     :ivar timeout: total time to wait for the containers and datasets
         to be created.
     :ivar wait_interval: how much to wait between list calls when waiting for
         the containers and datasets to be created.
-    :ivar _num_loops: numer of times to repeat the looping call based on the
+    :ivar _num_loops: number of times to repeat the looping call based on the
         ``tiemeout`` and the ``wait_interval``.
     :ivar per_node: number of containers and dataset per node.
     :ivar control_node_address: public ip address of the control node.
@@ -191,7 +191,7 @@ class ClusterContainerDeployment(object):
             if self.timeout is None:
                 # Wait two hours by default
                 self.timeout = 72000
-            self._num_loops = 10
+            self._num_loops = 1
             if self.wait_interval < self.timeout:
                 self._num_loops = self.timeout / self.wait_interval
 
@@ -381,7 +381,6 @@ class ClusterContainerDeployment(object):
         """
         deferred_list = []
         for node in self.nodes:
-            # Todo create a partial with the node info
             create_container_in_node = partial(self.create_container,
                                                node=node)
             for i in range(self.per_node):
--- a/benchmark/cluster_containers_setup.py
+++ b/benchmark/cluster_containers_setup.py
@@ -257,6 +257,17 @@ class ClusterContainerDeployment(object):
         datasets = []
         containers = []
 
+        # Listing datasets and containers to know the initial number of
+        # datasets and containers, so we know the total number of them
+        # we are expecting to have.
+        # XXX please note that, if some of those initial datasets or containers
+        # are being deleted, the output of this script won't be as expected,
+        # and it will end up timing out waiting for all the containers and
+        # datasets to be ready. This is a possible future improvement (bear
+        # it in mind if reusing the code), but it is not needed by benchmarking
+        # as it is not an scenario we will have, and even if we had it, we
+        # still can re-run this script to create the extra datasets and
+        # containers we may need, or even cleanup the cluster and start again.
         def list_datasets(ignored):
             return client.list_datasets_state()
 
--- a/benchmark/cluster_containers_setup.py
+++ b/benchmark/cluster_containers_setup.py
@@ -7,7 +7,7 @@ from uuid import uuid4
 from bitmath import GiB
 
 from eliot import add_destination, Message
-from twisted.internet.defer import inlineCallbacks
+from twisted.internet.defer import inlineCallbacks, gatherResults
 from twisted.python.filepath import FilePath
 from twisted.python import usage
 
@@ -253,10 +253,6 @@ class ClusterContainerDeployment(object):
             user_key
         )
 
-        nodes = []
-        datasets = []
-        containers = []
-
         # Listing datasets and containers to know the initial number of
         # datasets and containers, so we know the total number of them
         # we are expecting to have.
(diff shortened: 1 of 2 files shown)


benchmark/cluster_containers_setup.py
```diff
@@ -40,33 +38,48 @@ class ContainerOptions(usage.Options):
     optParameters = [
         ['apps-per-node', None, 1, 'Number of application containers per node',
          int],
```
mpgomez: Remove the parameters we no longer need, add the new ones, and validate them

benchmark/cluster_containers_setup.py
```diff
@@ -114,23 +143,22 @@ def main(reactor, argv, environ):
     return container_deployment.deploy_and_wait_for_creation()
 
 
-class ResponseError(Exception):
```
mpgomez: We no longer need that, as we don't do any REST calls from here.

benchmark/cluster_containers_setup.py
```diff
@@ -173,6 +208,8 @@ def __init__(self, reactor, env, options):
         self.cluster_cert = certificates_path.child(b"cluster.crt")
         self.user_cert = certificates_path.child(b"user.crt")
         self.user_key = certificates_path.child(b"user.key")
```
mpgomez: This is not par of the refactoring, but it was a bug. We didn't get the initial number of containers and datasets, so we could exit the loop where we wait until all the containers and dataset were created before it really happened, if we already had some datasets and containers in the cluster already.

benchmark/cluster_containers_setup.py
```diff
@@ -277,87 +366,63 @@ def deploy_and_wait_for_creation(self):
...
         yield loop_until(self.reactor,
                          self.is_datasets_deployment_complete,
-                         repeat(1, self.timeout))
+                         repeat(self.wait_interval, self._num_loops))
         yield loop_until(self.reactor,
                          self.is_container_deployment_complete,
-                         repeat(1, self.timeout))
+                         repeat(self.wait_interval, self._num_loops))
 
-    def _build_config(self, ignored):
```
mpgomez: Really ugly diff. The build_config function has been deleted.
The create_datasets_and_containers function is a new one. They don't really have code in common.

benchmark/cluster_containers_setup.py
```diff
@@ -277,87 +366,63 @@ def deploy_and_wait_for_creation(self):
...
+                    node_uuid=node.uuid,
+                )
+                Message.log(action=msg)
 
-        return {"applications": application_root,
-                "deployment": deployment_root}
+                d = self.client.create_dataset(node.uuid,
+                                               maximum_size=self.max_size)
+                d.addCallback(create_container_in_node)
+                deferred_list.append(d)
```
mpgomez: Ugly ugly diff. _configure has been deleted as well. create_container is new.

docs/gettinginvolved/cluster-setup.rst
```diff
@@ -185,20 +185,23 @@ The :program:`setup-cluster-containers` script has the following command line op
```
mpgomez: We no longer have the --app-template parameter.
Now we have some more options. The mandatory ones are:
--image
-- mountpoint
The optional ones are:
--max-size
--wait-interval

benchmark/cluster_containers_setup.py
```diff
@@ -40,33 +38,48 @@ class ContainerOptions(usage.Options):
...
         ['control-node', None, None,
          'Public IP address of the control node'],
         ['cert-directory', None, None,
          'Location of the user and control certificates and user key'],
-        ['wait', None, None,
+        ['max-size', None, 1,
+         'Size of the volume, in gigabytes. One GB by default'],
+        ['wait', None, 72000,
          "The timeout in seconds for waiting until the operation is complete. "
          "No waiting is done by default."],
```
itamarst: That's not quite true anymore. Although 72000 is a long time :grinning:

benchmark/cluster_containers_setup.py
```diff
@@ -40,33 +38,48 @@ class ContainerOptions(usage.Options):
...
+        ['wait-interval', None, 4,
+         "How often are we going to check if the creation of containers and "
+         "datasets has finished, in second"],
     ]
 
     synopsis = ('Usage: setup-cluster-containers --app-per-node <containers '
```
itamarst: :question: In theory `Options` is supposed to auto-generate this sort of thing, I thought...

benchmark/cluster_containers_setup.py
```diff
@@ -114,23 +143,22 @@ def main(reactor, argv, environ):
...
-
 class ClusterContainerDeployment(object):
     """
     Class that contains all the methods needed to deploy a new config in a
     cluster.
 
     :ivar options: ``ContainerOptions`` with the options passed to the script
-    :ivar application_template: template of the containers to deploy.
+    :ivar image: ``DockerImage`` for the containers.
+    :ivar max_size: maximum volume (dataset) size.
```
itamarst: "in bytes"

benchmark/cluster_containers_setup.py
```diff
@@ -114,23 +143,22 @@ def main(reactor, argv, environ):
...
+    :ivar mountpoint: unicode string containing the absolute path of the
+        mountpoint.
+    :ivar timeout: total time to wait for the containers and datasets
+        to be created.
+    :ivar wait_interval: how much to wait between list calls when waiting for
+        the containers and datasets to be created.
+    :ivar _num_loops: numer of times to repeat the looping call based on the
+        ``tiemeout`` and the ``wait_interval``.
```
itamarst: Typo.

benchmark/cluster_containers_setup.py
```diff
@@ -153,13 +181,20 @@ def __init__(self, reactor, env, options):
...
+            self.mountpoint = unicode(self.options['mountpoint'])
             self.per_node = self.options['apps-per-node']
             self.control_node_address = self.options['control-node']
             self.timeout = self.options['wait']
+            self.wait_interval = self.options['wait-interval']
             if self.timeout is None:
                 # Wait two hours by default
                 self.timeout = 72000
+            self._num_loops = 10
+            if self.wait_interval < self.timeout:
```
itamarst: By default they'll be equal so by default there will be ten loops? Maybe default should be 1, not 10.

benchmark/cluster_containers_setup.py
```diff
@@ -173,6 +208,8 @@ def __init__(self, reactor, env, options):
         self.cluster_cert = certificates_path.child(b"cluster.crt")
         self.user_cert = certificates_path.child(b"user.crt")
         self.user_key = certificates_path.child(b"user.key")
+        self._initial_num_datasets = 0
+        self._initial_num_containers = 0
```
itamarst: :question: Not particularly important in this case, but in general it seems better to avoid extra state. These two pieces of information seem like they could be looked up and then passed on to `create_datasets_and_containers` as arguments, rather than having to look them as attributes.

benchmark/cluster_containers_setup.py
```diff
@@ -277,87 +366,63 @@ def deploy_and_wait_for_creation(self):
...
+            containers are made.
         """
-        Message.log(action="Building config")
-        application_root = {}
-        applications = {}
-        application_root["version"] = 1
-        application_root["applications"] = applications
+        deferred_list = []
         for node in self.nodes:
+            # Todo create a partial with the node info
```
itamarst: Is this comment still relevant?

itamarst:
Address and merge - thank you!

mpgomez:
I finally did a bit of refactoring with the guidance of Jon, to avoid internal states in the class. Now there is a factory function that will return a deferred that will fire once the object has been constructed.

benchmark/cluster_containers_setup.py
```diff
@@ -169,31 +238,72 @@ def __init__(self, reactor, env, options):
...
             REST_API_PORT,
-            self.cluster_cert,
-            self.user_cert,
-            self.user_key
+            cluster_cert,
+            user_cert,
+            user_key
         )
 
+        nodes = []
```
itamarst: :question: A slightly nicer idiom:

```
from twisted.internet.defer import gatherResults

d = gatherResults([client.list_datasets_state(), client.list_containers_state(), client.list_nodes()])

def got_results(result):
    datasets, containers, nodes = result
    # ...
d.addCallback(got_results)
```

benchmark/cluster_containers_setup.py
```diff
@@ -208,10 +318,36 @@ def deploy(self):
...
+        before requesting the creation of any containers or datasets so
+        we can know the total number of datasets and containers to expect.
+
+        :return Deferred: that will fire once the ``_initial_num_datasets``
+            and ``_initial_num_containers`` are populated.
+        """
+        d1 = self.client.list_containers_state()
+
+        def set_initial_num_containers(containers):
+            self._initial_num_containers = len(containers)
```
itamarst: :question: These are still unnecessary state, but again not worth doing more work on.

itamarst:
Mostly just skimmed this time. Didn't see anything actionable, so if this runs correctly for you merge when green.

PR #2510, opened 2016-01-22
[FLOC-3903] Give test_docker on CentOS a larger slave and timeout.

james-w:
docker on CentOS runs slow in the default configuration because
it uses loop-devmapper. It's quite hard for us to change that,
but the tests seem more reliable on a slightly larger instance
with a slightly longer timeout. Give it more resources so it
more consistently passes while we find another way to improve
the situation.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -405,6 +405,13 @@ common_cli:
     - remote_logs.log
     - _trial_temp/test.log
 
+  run_trial_directories_to_delete: &run_trial_directories_to_delete
+    - ${WORKSPACE}/_trial_temp
+    - ${WORKSPACE}/.hypothesis
+
+  run_acceptance_directories_to_delete: &run_acceptance_directories_to_delete
+    - ${WORKSPACE}/repo
+
   run_trial_with_coverage: &run_trial_with_coverage |
     # The jobs.groovy.j2 file produces jobs that contain a parameterized job
     # type. These type of jobs always require a parameter to be passed on in
@@ -1002,7 +1009,6 @@ run_trial_modules: &run_trial_modules
   - flocker.docs
   - flocker.dockerplugin
   - flocker.node.test
-  - flocker.node.functional.test_docker
   - flocker.node.functional.test_script
   - flocker.node.functional.test_deploy
   - flocker.provision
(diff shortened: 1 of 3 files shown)


jobs.groovy
```diff
@@ -473,7 +473,13 @@ def list_jobs(dashProject, dashBranchName) {
                                  module: module, values: job_values,
                                  full_name: full_name])
                 }
-            } else {
+            } else if (job_values.module) {
+                full_name = full_job_name(
+                    dashProject, dashBranchName, job_name)
+                results.add([type: job_type, name: job_name,
+                             module: job_values.module, values: job_values,
+                             full_name: full_name])
```
james-w: Allow a new type of definition that uses "module" rather than a list of
"with_modules". The difference is that this doesn't do the name mangling
that with_modules does, but still puts $MODULE in the environment.

This allows us to define the stanzas for test_docker, while keeping the same
names as they have now, and using the same body for the jobs, as they
rely on running trial ${MODULE}.

jobs.groovy
```diff
@@ -544,93 +550,29 @@ def build_tabs(dashBranchName) {
 def define_job(dashProject, dashBranchName, branchName, job_type, job_name,
                full_name, job_values, module, isReleaseBuild) {
 
-    if (job_type in ['run_trial', 'run_trial_for_storage_driver', 'run_acceptance']) {
-        assert module != null
-        job(full_name) {
-            parameters {
-                // we pass the 'MODULE' parameter as the flocker module to test with trial
+    job(full_name) {
```
james-w: The only remaining difference between the job types was modules_to_delete, this reads
that from build.yaml, so we now just loop over all the jobs and use the values, largely
ignoring the type.

The type is still used for cron jobs, but that could likely be unified as well. Keeping the
job type in the build.yaml may be useful for human understanding though.

build.yaml
```diff
@@ -1113,7 +1116,38 @@ job_type:
...
+      module: flocker.node.functional.test_docker
+      with_steps:
+        - { type: 'shell', cli: *run_trial_cli }
+      archive_artifacts: *flocker_artifacts
+      publish_test_results: true
+      coverage_report: true
+      clean_repo: true
+      # Increase the timeout due to FLOC-3903
+      timeout: 45
+      directories_to_delete: *run_trial_directories_to_delete
```
james-w: This stanza is the point of the branch. The rest is supporting this
change. This is largely copied from the existing run_trial stanza.

james-w:
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/medium-for-test-docker-FLOC-3903/job/run_trial_on_AWS_CentOS_7_flocker.node.functional.test_docker/ shows 10 passes in a row for this branch.

jobs.groovy
jml: Not a huge fan of duplicating this type information across languages, but can't see an alternative.

jml: Also, I should note that it's not introduced in this branch.

jml:
LGTM. Thanks for fixing this and for testing it so thoroughly. Merge after addressing comments (up to you).

PR #2511, opened 2016-01-22
[FLOC-3825] Run acceptance tests on GCE instances.

sarum90:
This PR sets up nightly runs of GCE acceptance tests using the loopback driver. (GCE driver still under construction, not scheduled to be enabled in acceptance tests until after we have the functional tests available in Jenkins). Once it is enabled with the GCE driver the loopback tests will probably be turned off and be replaced with the gce driver tests.

Along the way I also did some opportunistic cleanups:

- Enhanced logging if your configuration yaml does not match what the provisioner is expecting. This required implementing something that inspects the provisioner factory to compare expected arguments with the configuration blob.
- Removed some unused arguments throughout the provisioning code.
- Reused some of the retry logic from the gce driver.
- Reused some of the provisioning logic from the aws driver.

There was certainly room for more cleanups (and especially more room for adding better logging for some of the things I got tripped up on along the way), but this seemed to be a reasonable incremental change.

Note this depends on some jenkins PRs being deployed before I can prove this works on master:
- deployment of credentials.
- rendering the credentials correct via ansible: https://github.com/ClusterHQ/ci-platform/pull/10

But I can show them working on my personal jenkins master:
http://52.26.214.217:8080/job/ClusterHQ-flocker/job/run-acceptance-test-gce-FLOC-3825/
(Login credentials are the same as the usual server)


sarum90:
Running all acceptance tests because I'm messing with the provisioning code.

build.yaml
```diff
@@ -1554,6 +1571,52 @@ job_type:
       timeout: 90
       directories_to_delete: []
 
+    run_acceptance_on_GCE_CentOS_7_with_LOOPBACK:
+      at: '0 5 * * *'
+      on_nodes_with_labels: 'aws-centos-7-SELinux-T2Medium'
```
james-w: Presumably this needs to be updated to point to a GCE node?

build.yaml
```diff
@@ -497,6 +497,23 @@ common_cli:
...
+    # We gather the return code but make sure we come out of these tests with 0
+    # we store that code and pass it to the end of the job execution,
+    # as part of the JOB_EXIT_STATUS variable.
+    #
+    # The admin/run-acceptance-tests will provision a flocker cluster with a
+    # single node since the loopback backend doesn't support moving data across
+    # hosts.
+    ${venv}/bin/python admin/run-acceptance-tests \
+    --distribution ${DISTRIBUTION_NAME} --number-of-nodes 1 \
+    --provider gce --dataset-backend loopback --branch ${TRIGGERED_BRANCH} \
```
james-w: :question: If the difference here is the provider string, you could consider keeping
these the same, and using the same trick as for ${ACCEPTANCE_TEST_MODULE}

flocker/common/_interface.py
```diff
@@ -78,3 +80,50 @@ def invariant(value):
...
+    See :func:`validate_signature_against_kwargs` raises clause.
+    """
+    def __str__(self):
+        return repr(self)
+
+
+def validate_signature_against_kwargs(function, keyword_arguments):
+    """
+    Validates that ``function`` can be called with keyword arguments with
+    exactly the passed in ``keyword_arguments_keys``. In this case validation
```
james-w: :question: "the passed in" is rather clunky, and looks wrong to me until half way through. "with exactly the names in ``keyword_argument_keys`` or something sounds better to me, or "specified" instead of "passed in"

flocker/common/_interface.py
```diff
@@ -78,3 +80,50 @@ def invariant(value):
...
+    is verifying that the function's signature allows it to be called with
+    exactly the given keyword arguments.
+
+    :param function: The function of which to verify the signature.
+    :param set keyword_arguments_keys: A set of keyword argument names to
+        validate against the function signature.
+
+    :returns: ``None`` if the validation succeeds.
+
+    :raises InvalidSignature: If validation fails.
```
james-w: :question: "None or raises" is a rather awkward function signature, though
it does make some sense for this. Did you consider returning the value
and perhaps making it not inherit Exception. It looks like the only use
catches the exception and unpacks it, raising something else.

Also the similarity with a matcher is interesting, I note you define a matcher
based on raising this later.

flocker/common/_interface.py
```diff
@@ -78,3 +80,50 @@ def invariant(value):
...
+    """
+    arg_spec = getargspec(function)
+    accepted_arguments = frozenset(arg_spec.args)
+    optional_arguments = frozenset()
+    if arg_spec.defaults is not None:
+        optional_arguments = frozenset(arg_spec.args[-len(arg_spec.defaults):])
+
+    unexpected_arguments = frozenset(keyword_arguments - accepted_arguments)
```
james-w: This isn't going to do the expected thing in the case of *args, **kwargs, so maybe raise
an exception if they are in the function signature?

flocker/node/agents/gce.py
```diff
@@ -35,6 +35,68 @@
...
+            )
+        update = get_global_operation
+
+    def finished_operation_result():
+        latest_operation = update().execute()
+        if latest_operation['status'] == 'DONE':
+            return latest_operation
+        return None
+
+    return poll_until(finished_operation_result, [1]*timeout)
```
james-w: :question: is a static 1s delay reasonable here? We've just been looking at
the AWS docs, and they recommend the first delay is until the operation is
likely done (however you are supposed to determine that) and then exponential
backoff. This is largely because of their rate limiting. If GCE don't have that
then this seems ok.

flocker/node/agents/gce.py
```diff
@@ -35,6 +35,68 @@
...
+        # For zone operations, we need to extract the zone and the project.
+        # An example of a 'zone' value of a zone operation dict is:
+        #
+        # 'https://content.googleapis.com/compute/v1/'
+        # 'projects/<project-name>/zones/<zone-name>'
+        #
+        # Note that everything before 'projects' is sometimes not included.
+        zone_url_parts = operation['zone'].split('/')
+        project = zone_url_parts[-3]
+        zone = zone_url_parts[-1]
```
james-w: :question: unpacking these from a URL isn't very pleasant. From a quick look it
seems most callers know these values, could they be passed in, perhaps with
two functions: one for global, one for zone operations?

If 'status' == 'DONE' is standard across all GCE operations, then something
that loops, checks responses for that, and handles the timeouts, delays etc.
seems good.

If feels like it could be composed with functions to create these update checks,
and perhaps one to check the output for success, and avoid having these URL
operations.

I'd be ok with what we have here though, it seems GCE is consistent enough
that we won't have a problem with it.

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+                zone=self.zone,
+                instance=self.name
+            ).execute()
+            wait_for_operation(self.compute, operation, timeout=60)
+
+        But, as that is untested, it will remain unimplemented.
+        """
+        raise NotImplementedError(
+            "GCE does not have reboot implemented because it has "
+            "experimentally been determined to not be needed.")
```
james-w: :question: seems a little off to have this just skipped in GCE, but I can
understand you not wanting to inflate this effort even further.

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+
+
+@implementer(IProvisioner)
+class GCEProvisioner(PClass):
+    """
+    A provisioner that can create instances on GCE.
+
+    :ivar unicode zone: The zone in which instances will be provisioned in.
```
james-w: two "in"s

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+    :ivar unicode project: The project under which instances will be
+        provisioned in.
```
james-w: extra "in"

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+            project=self.project,
+            zone=self.zone,
+            body=config
+        ).execute()
+
+        operation_result = wait_for_operation(
+            self.compute, operation, timeout=60)
+
+        if not operation_result:
+            raise ValueError("Timed out waiting for VM creation")
```
james-w: :question: it would be really nice if this gave a little more information about what went wrong.
Getting the current status would be a little tricky, but could it at least give the instance
id or something to make finding which one failed easier?

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+
+        operation_result["targetLink"].split("/")[-1]
```
james-w: Is this needed?

flocker/provision/_gce.py
```diff
@@ -0,0 +1,494 @@
...
+    address = field(type=bytes)
+    private_address = field(type=bytes)
+    distribution = field(type=bytes)
+    project = field(type=unicode)
+    zone = field(type=unicode)
+    name = field(type=unicode)
+    compute = field()
+
+    def get_default_username(self):
+        return bytes(_GCE_ACCEPTANCE_USERNAME)
```
james-w: This is assumed to exist, because it's assumed that GCEProvisioner used it
when creating the instance. Could it be an attribute that is set by GCEProvisioner.
Indeed perhaps gce_provisioner should pick it and pass it to GCEProvisioner,
which uses it then passes it to here.

james-w:
Hi,

Thanks for the branch, it was very clear and the code was clearly documented.

I think it's pretty close. There were a few things that I think should be changed,
and some that I think could be changed. Given the former I'll ask for a
re-review before landing this.

Thanks,

James

build.yaml
sarum90: Ooo I hadn't thought of that, nice.

Nothing has ever gone wrong with pretending global variables are arguments to functions before, right?

build.yaml
sarum90: Interestingly it doesn't (and in fact can't). This just specifies which slaves the run_acceptance_test code will be run. Presently jenkins has no GCE slaves because teaching jenkins how to spin up GCE instances turns out to be non-trivial. So, this will be running code on AWS slaves that will spin up a flocker cluster on GCE  and run tests against them.

It is sort of subtle, but it is also what we do with rackspace (The assumption is we prefer to use AWS slaves for all jobs unless there is a reason not to).

flocker/common/_interface.py
sarum90: Done

flocker/common/_interface.py
sarum90: I'll think on this. To me this seems more pythonic. Although maybe if I return a new object "signature differences" or something... I'll get back to this before I'm done.

flocker/common/_interface.py
sarum90: Decided to add tests for these situations.

Seemed to me that *args can just be ignored (updated comment to be clear that we don't care about positional arguments), and **kwargs being in the signature should just force unexpected_arguments to an empty set.

In second review please feel free to comment on the added tests if you think their specification is unexpected.

flocker/node/agents/gce.py
sarum90: Callers do know... but they just gave that information to a different call, and the information is available within the operation structure. So usably, it's not nice to have to keep specifying project and zone, and model wise it'd be a shame if we ever got a project/zone that was different than the project/zone that we could infer from the operation's structure.

Point taken about how unpleasant this is though, I think the unpleasantness comes from the brittleness of the parsing, and the terrible errors that it will produce if it guesses wrong.

I'll try to harden the parsing code and wrap key errors and index errors in a nicer error so that this can be considered to be more pleasant.

flocker/node/agents/gce.py
sarum90: I haven't hit rate limiting, but that's not to say it doesn't exist. punting this up a layer.

For provisioning - we do it so infrequently I think static is probably fine. For actual GCE driver code, we've got a TODO to do something more sophisticated. Punting up a layer should enable us to do something better from that layer.

flocker/provision/_gce.py
sarum90: Done. Although layers are kind of tricky here. "get_default_username" doesn't really make sense in a GCE world, where it will create any user you want it to create. This is probably the best compromise though, where the username is passed from the thing that injects the public key into the VM to the object that can answer, "who should I try to log in as?".

flocker/provision/_gce.py
sarum90: +1

maybe I should tear it out of everywhere. I think I looked at that and that would mean tearing out all zfs work or something. Either way despite the oddity I think raising an error is preferred to trying to check in code that has never been run.

flocker/provision/_gce.py
sarum90: It appears as if I'm trying a little too hard to end my sentences in a preposition in.

flocker/provision/_gce.py
sarum90: It appears as if I'm trying a little too hard to end my sentences in a preposition in.

flocker/provision/_gce.py
sarum90: Certainly doesn't look like it, nice catch!

flocker/provision/_gce.py
sarum90: instance name added, good idea. I think status would actually be boring (simple state machine of PENDING->DONE).

sarum90:
Thanks for the great review! Going to run the tests again, if they pass I'll put back up for review.

james-w:
Hi,

I think your changes look good, thanks for making them.

I would still lean towards passing in the project/zone details
explicitly, but I'm happy to disagree on that one.

The changes to the kwargs checking are good I think. The only
thing I don't like about the tests are the magic names of the arguments
("one" etc.) because they exist so far away from their definition
in the helper functions. That's partly because creating functions
in Python is harder than ideal. You could consider:

  * Some sort of constants for them, rather than magic strings in the tests
  * More descriptive names "required", "extra", "missing" etc. so the intent of the test reads more naturally
  * Inlining the function definitions to each test, at the cost of repetition.

However, I don't think you need to do any of them.

Overall I'm happy with this as it stands. :heavy_check_mark: Please merge when green and happy.

sarum90:
Thanks for the reviews.

Re explicitly passing zone/project. Yea I can almost go either way on that too, but "don't store duplicate state" trumps "don't randomly attempt to parse urls from a convoluted API when you don't have to" for me :). I guess I could probably write a wrapper around the whole google api and get the best of both worlds.

Also I _think_ your test improvements would probably be good for those unit tests, but I'm not sure that the value of me going back and making them just a little bit better exceeds the cost, especially since that utility is pretty much just meant for use in that one spot in the code. If it gets more uses I might look at improving it's signature (rather than None or raises) and improving the unit tests. Fixed up some of them.

Also discovered a little rackspace error that I've ironed out now.

I have green tests, so I'm going to merge.

james-w:
Yeah, I think ideally the API would be such that you could just ask an operation for
the URL to check to see if it had completed, then we wouldn't have to worry about
it at all.

I agree the benefit of changing the tests would be low.

Thanks for considering my comments.

PR #2512, opened 2016-01-25
[FLOC-3965] Ensure log message is written if EBS volume is not ready

jongiddy:
Line to log a message is missing the `write()` method.

This is amongst the reasons it was difficult to see what is generating the AWS RequestLimitExceeded errors.

### DIFF
--- a/flocker/node/agents/_logging.py
+++ b/flocker/node/agents/_logging.py
@@ -88,12 +88,9 @@ TARGET_STATUS = Field.for_types(
 WAIT_TIME = Field.for_types(
     u"wait_time", [int, float],
     u"Time, in seconds, system waited for the volume to reach target status.")
-NEEDS_ATTACH_DATA = Field.for_types(
-    u"needs_attach_data", [bool],
-    u"True if target volume status leads to non-empty EBS Volume attach data.")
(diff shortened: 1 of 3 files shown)


dwgebler:
Thanks, LGTM when tests pass.

jongiddy:
Resubmitting this PR for review, since the simple change to add a `write` method to the log message turns out to do lots of validation, which failed due to missing fields in the log message.

In order to add the missing fields, I moved the log message into the nested function, where the information was more readily available.  This has the effect of displaying the message one extra time. This could be considered a good thing, as it now displays a log message each time it does the check. If everything runs quickly, it now prints one log message instead of zero.

Once this change was done, I simplified the calling function by replacing with `poll_until` and passing the elapsed time instead of start time as a parameter. This latter change allows a sleep in the functional tests to be removed, saving 5 seconds off the test run time.

dwgebler:
That's some quite nice refactoring :)
Interesting that even if the write() had worked in the previous code, an immediate success from `_should_finish` would prevent us getting the log message, so yes this is definitely an improvement.

(my fault for not checking it more thoroughly first time it went up for review, sorry! also wrongly assumed validation occurred before writing the message, so didn't occur to me errors would occur for that reason at the writing stage)

This is passing tests and looks sound, so I think it's good to merge.

PR #2514, opened 2016-01-25
[FLOC-3832] EBS flaky test

dwgebler:
The problem reported on this issue seems to be a result of AWS eventual consistency sometimes biting us. Not sure what else we can realistically do except give the tests more time to run and therefore hopefully encounter the volume state they expect to be able to proceed. There is (I think) some planned work around solving eventual consistency problems more broadly in Flocker by keeping track of state locally. See FLOC-3219

### DIFF
--- a/flocker/node/agents/functional/test_ebs.py
+++ b/flocker/node/agents/functional/test_ebs.py
@@ -29,7 +29,7 @@ from ..ebs import (
     EBSMandatoryProfileAttributes, _get_volume_tag,
     AttachUnexpectedInstance, VolumeBusy,
 )
-from ....testtools import AsyncTestCase, flaky, async_runner
+from ....testtools import AsyncTestCase, async_runner
 
 from .._logging import (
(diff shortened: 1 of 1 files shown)


flocker/node/agents/functional/test_ebs.py
```diff
@@ -68,6 +70,8 @@ class EBSBlockDeviceAPIInterfaceTests(
         )
 ):
 
+    run_tests_with = async_runner(timeout=timedelta(hours=1))
```
jml: This gives each & every test in the suite an hour to run, effectively disabling the timeout for all tests (since the Jenkins timeout is <1hr). Is that what you meant to do? 

If so, please add a comment. If not, please pick a different timeout (also adding a comment). The comment should link to FLOC-3219 and say more or less what you say in the PR description.

jml:
Thanks! Please address comment & merge.

dwgebler:
Ignoring Cinder build failure as it's known flaky and unrelated to this branch.

PR #2515, opened 2016-01-25
[FLOC-3970] Remove verbose Vagrant logging.

james-w:
It's really rather verbose, so obscures any errors, and
will be rarely useful.


jml:
LGTM. Thanks.

PR #2516, opened 2016-01-25
[FLOC-3933] Use testtools 1.9.0 

jml:
Increments the version of testtools that we use.

Practical benefits:

* `expectThat` will actually fail async tests
* new matchers for Deferreds (see http://testtools.readthedocs.org/en/latest/twisted-support.html)
* `testtools.twistedsupport` new namespace for Twisted stuff
* less weird interaction with tests in case of timeout
* slightly less boilerplate in our base cases

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -42,7 +42,7 @@ sphinx-prompt==1.0.0
 sphinx-rtd-theme==0.1.8
 sphinxcontrib-httpdomain==1.3.0
 sphinxcontrib-spelling==2.1.2
-testtools==1.8.2chq2
+testtools==1.9.0chq3
 tox==2.1.1
 troposphere==1.3.0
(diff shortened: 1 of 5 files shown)


james-w:
Looks good to me, thanks for keeping us up to date.

Not sure who should mentor review this since you are my mentor.

sarum90:
Looks good, glad to see those `XXX` comments die. And +1 for the review (I think no mentor was probably fine for this one).

But.... Jenkins isn't happy:

```
Exception: Cannot import CaptureTwistedLogs. Maybe upgrade your version of testtools: pip install --upgrade --process-dependency-links .[dev]
```

If making jenkins happy is easy, go ahead and merge once fixed without re-review. If it's hard, I'm sure james or I will be willing to do a quick glance over the fix.

PR #2519, opened 2016-01-26
[FLOC-3966]cleaning-after-ourselves-is-a-good-practice

mpgomez:
Adding cleanup functionality to the scenarios.
A dataset was being left behind when running the write load scenario.

I have added a new function, run_cleanup, alongside with the ones that were already part of the setup interface (run_setup and make_request).

In the read_scenario, both, run_setup and run_cleanup will actually do nothing, but they will for the write scenario. We will expect them to do something as well in future scenarios.


benchmark/scenarios/test/test_write_request_load.py
```diff
@@ -167,6 +168,39 @@ def assert_created(returned_datasets):
         d2.addCallback(assert_created)
         s.stop()
 
+    @capture_logging(None)
+    def test_cleanup_deletes_dataset(self, _logger):
+        """
+        ``write_request_load_scenario`` deletes the dataset created by
+        the setup when the scenario is stopped.
+        """
```
brendan-donegan: Some more descriptive names would be good here - e.g. clock rather than c, scenario rather than s.

benchmark/_interfaces.py
```diff
@@ -114,6 +114,13 @@ def make_request():
         that has been setup and/or created in ``run_setup``, and has the
         pre-requesite that ``run_setup`` has successfully finished.
 
-        :return: ``Deferred`` that fires when the REST request has been
+        :return Deferred: that fires when the REST request has been
             completed, and returns the results of the request.
         """
+    def run_cleanup():
```
jongiddy: Add a blank line between function definitions.

benchmark/scenarios/test/test_write_request_load.py
```diff
@@ -167,6 +168,39 @@ def assert_created(returned_datasets):
...
+        c = Clock()
+        cluster = self.make_cluster(self.get_fake_flocker_client_instance())
```
jongiddy: Extract the control service from the cluster here, to avoid using `s.scenario_setup.control_service` later.  You should not be relying on the fact that the scenario setup object also extracts the control service.

:question: This should be applied to all the tests in this file, and `WriteRequest` should be modified to accept the control service instead of the cluster (since it doesn't need the cluster object).  However, this changes a lot of lines that are not part of this PR.

benchmark/scenarios/test/test_write_request_load.py
```diff
@@ -167,6 +168,39 @@ def assert_created(returned_datasets):
...
+        s = write_request_load_scenario(c, cluster, 5, sample_size=3)
+
+        def assert_created(returned_datasets):
+            self.assertNotEqual(returned_datasets, [])
+
+        # Create a datasest and verify we get a success
+        d = s.scenario_setup._create_dataset(self.node1)
+        d.addCallback(s.scenario_setup._set_dataset_id)
```
jongiddy: This test is too low-level, essentially copying the implementation out of the tested code.

It should ideally run `scenario.start` followed by `scenario.stop`. However, I would also accept `scenario_setup.run_setup` followed by `scenario_setup.run_cleanup`.

jongiddy:
Please address the above comments and re-submit.  Thanks.

benchmark/scenarios/test/test_write_request_load.py
mpgomez: I could change the naming in this test, but refactoring the whole file may be out of scope for this issue. Please, feel free to raise a issue if you think it needs a bit of refactoring - it may not be a bad idea, as they could be simpler and better structured

benchmark/scenarios/test/test_write_request_load.py
mpgomez: Yes, it will not only affect Write Request, but all the scenarios. I have opened a PR: 
https://clusterhq.atlassian.net/browse/FLOC-3991

james-w:
:heavy_check_mark: Looks good to me, please merge when green. I retried the
2 failing tests.

benchmark/scenarios/test/test_write_request_load.py
jongiddy: Yes, I think you're right Pilar. Descriptive names are good, but consistency is also good.  If the other tests use the same abbreviated names, then they should all be changed, as a separate issue, since the read_request_tests probably use the same scheme.

benchmark/scenarios/test/test_write_request_load.py
```diff
@@ -453,4 +478,4 @@ def test_scenario_timeouts_if_requests_not_completed(self, _logger):
         self.assertNoResult(d)
         c.advance(s.timeout + 1)
         self.assertTrue(s.rate_measurer.outstanding() > 0)
-        self.successResultOf(d)
+        self.successResultOf(d)
```
jongiddy: Add a newline after last line. This may (should) cause the lint test to fail.

PR #2521, opened 2016-01-26
[FLOC-3977] Fixup labs admonition image and padding.

lukemarsden:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3977

While I was there, fixed bottom padding for these admonitions on low width displays.


jml:
LGTM

lukemarsden:
Thanks!

PR #2522, opened 2016-01-26
[FLOC-3500] Disable timeout for Mongo tests

dwgebler:
Should fix https://clusterhq.atlassian.net/browse/FLOC-3500

The code for the integration tests is about as broadly reliable as it's going to get; failures are typically due to these tests having only 2 minutes to run.

### DIFF
--- a/flocker/acceptance/integration/test_mongodb.py
+++ b/flocker/acceptance/integration/test_mongodb.py
@@ -4,10 +4,14 @@
 Tests for interoperability between MongoDB and Flocker datasets.
 """
 
+from datetime import timedelta
+
 from twisted.python.filepath import FilePath
 
 from ..testtools import get_mongo_client, require_mongo
 from .testtools import make_dataset_integration_testcase
+
+from ...testtools import async_runner
 
 
 def insert_data(test_case, host, port):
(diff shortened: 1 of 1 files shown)


flocker/acceptance/integration/test_mongodb.py
```diff
@@ -4,11 +4,15 @@
 Tests for interoperability between MongoDB and Flocker datasets.
 """
 
+from datetime import timedelta
+
 from twisted.python.filepath import FilePath
 
 from ..testtools import get_mongo_client, require_mongo
 from .testtools import make_dataset_integration_testcase
```
mpgomez: All the imports from testtools could be grouped together?

flocker/acceptance/integration/test_mongodb.py
```diff
@@ -57,6 +61,10 @@ class MongoIntegrationTests(make_dataset_integration_testcase(
     """
     Integration tests for MongoDB.
     """
+
+    # Disable timeout for MongoDB integration tests.
```
mpgomez: I have seen a decorator that looks pretty similar to it, @run_test_with. Is there any difference between using one or the other?

flocker/acceptance/integration/test_mongodb.py
dwgebler: We confusingly have several different modules called `testtools`, all in different places; this test case imports from `flocker.acceptance.integration.testtools`, `flocker.acceptance.testtools` and `flocker.testtools`.

mpgomez:
Looks good to me. Please, take a look at the comments and address them if they need to be addressed :)
Thanks!

flocker/acceptance/integration/test_mongodb.py
dwgebler: `run_test_with` allows this kind of behaviour to be applied to a single test, whereas `run_tests_with` does the same thing for all the tests in a class.

PR #2523, opened 2016-01-26
Fix indentation in AWS section of acceptance-testing docs

brendan-donegan:
Really minor doc fix, one of the yaml lines had too many spaces so sorted that out. Tested running make html


exarkun:
Thanks.  Ideally, we'd have automated test coverage to check the examples in the documentation for validity.  The values for some parts of the example, like the secret API key, would be tricky to validate.  It would be straightforward to validate the structure, though (ie, do the keys present match up with the allowed and required arguments to the Python code that consumes this stuff).

Some of our documentation has some automatic test coverage - by virtue of being derived from actual code that is executed by our test suite.  That kind of coverage probably doesn't apply in this case since the example is intentionally unusable (at least by not using a real AWS API key).

It's tempting to wonder whether some kind of doctest could help with this case.  Off hand, though, I don't quite know if you can write a doctest for a yaml blob without inserting user-facing contortions into the documentation - so perhaps that's not suitable.

I could imagine a sort of unit test that loads this particular file, pulls out the various examples, and throws them at the Python consumers to make sure they're happy.  I could also imagine codifying the exact required structure with something like jsonschema and then validating using that instead.  For this structure I think we're quite happy to make incompatible changes from release to release.  That suggests to me that a jsonschema-style testing approach may get in the way more than it helps.  Just throwing the structure at the relevant Python code would ensure we don't make simple mistakes like this one and that when we make changes to the implementation we get a reminder to fix the documentation instead.

Finally, I could also imagine generating this example from the implementation so that there isn't this duplication of content to maintain at all.  We already do some custom Sphinx generation (`flocker/docs/`) so this might not be particularly far out of reach.

This all implies more work than the simple (apparently correct, fwiw!) change in this PR.  Considering the number of bugs we've had in these examples, I think it's probably actually worth it.  However, automated test coverage for the documentation is *not* a baseline requirement for Flocker documentation development so I won't say that introducing such test coverage is required to land this fix.  I will suggest that if you feel like spending some time on it, it would probably be useful. :)

Regardless of that, it looks like the problem was introduced as a result of @davidoliverSP2 's editor configuration - which allowed him to insert a hard tab into the documentation.  It's probably worth drawing his attention to that (which perhaps I am succeeding in doing here) so he can fix that mis-configuration and avoid repeating this particularly problem again.

Go ahead and merge if you don't build some kind of wider-scoped fix like one of the things discussed here.  Otherwise, please re-submit for review with those improvements.

exarkun:
Ah, last thing.  It may also be worth auditing the changeset which introduced this bug to see if any other instances of it were introduced.

brendan-donegan:
Certainly agree with the sentiment and had thought about how to address it,
but it seems like it might be involved enough to deserve a task or story of
its own. As you mentioned the main issue is how to extract the examples and
validate them correctly. The main thing that sprang to my mind was how to
tell the difference between e.g. yaml code sections and shell code sections.

In summary I think this should land (modulo the audit of the changeset that
introduced it) and I will make a point of creating a story for
documentation validation.

brendan-donegan:
Actually I can see now that the sections are marked up with e.g. code-block: yaml or prompt: bash, so we can take advantage of that. At least for YAML it should be very easy to extract the yaml and just parse it to check it's valid.

brendan-donegan:
Re: the audit, the docs folder seems to have quite a lot of hard tabs. Should we be having a blanket ban across the repository?

PR #2524, opened 2016-01-26
[FLOC-3975] Add counts of the most commonly called AWS API calls.

jongiddy:
We often get `RequestLimitExceeded` errors on AWS.  This PR adds a count field to the logs of functions performing `boto` calls to show how often that operation has been called.  In addition, calls to the higher-level `list_volumes` function are counted, to isolate where they are coming from.

Sample log output:
```json
{"timestamp": 1453822992.761821, "task_uuid": "2e129379-1515-411c-9007-798c8aa27fe0", "action_type": "flocker:agent:discovery", "action_status": "started", "task_level": [2, 2, 1]}
{"count": 406, "task_uuid": "2e129379-1515-411c-9007-798c8aa27fe0", "timestamp": 1453822992.762302, "operation": ["_discover_raw_state", [], {}], "message_type": "flocker:node:agents:blockdevice:list_volumes", "task_level": [2, 2, 2]}
{"count": 406, "task_uuid": "2e129379-1515-411c-9007-798c8aa27fe0", "task_level": [2, 2, 3, 1], "action_type": "flocker:node:agents:blockdevice:aws", "timestamp": 1453822992.763307, "operation": ["_list_ebs_volumes", [], {}], "action_status": "started"}
```

### DIFF
--- a/flocker/node/agents/blockdevice.py
+++ b/flocker/node/agents/blockdevice.py
@@ -420,6 +420,15 @@ DISCOVERED_RAW_STATE = MessageType(
     [Field(u"raw_state", safe_repr)],
     u"The discovered raw state of the node's block device volumes.")
 
+FUNCTION_NAME = Field.for_types(
+    "function", [bytes, unicode],
+    u"The name of the function.")
+
+CALL_LIST_VOLUMES = MessageType(
+    u"flocker:node:agents:blockdevice:list_volumes",
+    [FUNCTION_NAME, COUNT],
+    u"list_volumes called.",)
+
 
 def _volume_field():
     """
@@ -1228,12 +1237,6 @@ class _SyncToThreadedAsyncAPIAdapter(PClass):
     _threadpool = field()
 
 
-CALL_LIST_VOLUMES = MessageType(
-    u"flocker:node:agents:blockdevice:list_volumes",
-    [OPERATION, COUNT],
-    u"list_volumes called.",)
-
-
 def log_list_volumes(function):
     """
     Decorator to count calls to list_volumes.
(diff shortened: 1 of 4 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -1232,6 +1228,36 @@ class _SyncToThreadedAsyncAPIAdapter(PClass):
     _threadpool = field()
 
 
+CALL_LIST_VOLUMES = MessageType(
+    u"flocker:node:agents:blockdevice:list_volumes",
+    [OPERATION, COUNT],
+    u"list_volumes called.",)
+
+
+def log_list_volumes(function):
```
sarum90: :grey_question: maybe unit test this? We have tools to unit test eliot logs, so hopefully it won't be that hard?

Perhaps this isn't that important until we have something consuming the logs, and is only as important as that consumer is.

sarum90:
Awesome!

That's a cute little trick to get those stats. In the past I've just had a jq / python blob to do this for me.

One comment about maybe adding a test. 

Couple questions on further improvements to this:

1) Do we have any intention to code something to consume these logs for long-term reporting?
2) Are we sure that boto3 only makes 1 API call per list_volumes call? I think it might walk the pages for us. Is there another appoach to count API calls, maybe from AWS's side? (GCE has this capability, so maybe AWS does as well)
3) ICloudAPI.list_live_nodes() also makes an API call, it'd be a nice-to-have as well.

Consider all of the above points as optional. This looks great.

jongiddy:
I'm not sure about long-term reporting.  Top priority is to reduce the number of API calls now. If these logs serve their purpose, we might remove them.

As well as this logging, we also get `boto3` to log its calls (and count them), which might help log retries or paging (or not). But if `boto` doesn't help us, the only other data comes from CloudTrail. The next step might be to insert a proxy server between our services and AWS, but only if we completely cannot solve the API request problems.

`ICloudAPI.list_live_nodes()` generates logs for the AWS EBS driver. If there are a lot of calls, we will add higher-level logs as well.

PR #2528, opened 2016-01-27
[FLOC-3916] [FLOC-3917] Kill vagrant.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3916
https://clusterhq.atlassian.net/browse/FLOC-3917


CONTRIBUTING.rst
```diff
@@ -54,38 +54,18 @@ Have questions or need help?
...
-
-   vagrant up
-   vagrant ssh
-
-The ``flocker`` directory created above will be shared in the virtual machine at ``/vagrant``.
 Install Flocker's development dependencies in a ``virtualenv`` by running the following commands:
 
 .. prompt:: bash $
 
-   cd /vagrant
```
sarum90: Drive by comment: This is part of my workflow developing on an OSX box when I need to run the loopback blockdevice APIs.  But... I suppose it wouldn't have to be, I could also do that development work on my pet AWS instance.

Ultimately this change will make my dev process mildly more annoying, but that might not be worth the maintenance burden.

Might want to check with the other OSXers though to double check this isn't putting a thorn in their back.

tomprince: Well, the dev-box hasn't changed for quite some time, and I'm not proposing that we delete it. Also, now that we have started to use packer elsewhere, it might make more sense to use that to build a vagrant box for development, instead?

tomprince: A couple of more thoughts:
- nothing is currently exercising any of the dev-box related code.
- We could just move all the installation stuff into the `Vagrantfile` at the top-level, instead of building a box and then using that box as base.

jml: +1 to using the packer set up for development. The tests haven't passed on the dev box for some time, and we have deliberately ruled out making them pass (for the moment).

flocker/testtools/__init__.py
```diff
@@ -684,35 +682,6 @@ def remove_image():
         return d.addCallback(lambda ignored: tag)
 
 
-def skip_on_broken_permissions(test_method):
```
jml: I'm -0 about this change. If we set up something with packer for doing development on OS X (and we should do that), then we'll probably still need this or something like it. 

Or, another way, removing this seems like a separate thing to removing the tutorial box set-up code.

jml:
Thanks Tom! Always nice to see code go away.

I think the `skip_on_broken_permissions` removal should be done separately, or not at all. Approved (dependent on tests) if you do so. If not, then would appreciate an opportunity to be convinced before you land this.

PR #2529, opened 2016-01-27
[FLOC-3763] Remove the jobs.groovy.j2 symlink.

james-w:
Jenkins no longer uses the file, so we can remove the compatibility
symlink.


jml:
LGTM

PR #2532, opened 2016-01-27
[FLOC-3983] Acceptance tests prep for node shutdown

itamarst:
When a node is shutdown and later started its public IP may have changed. We want to have an acceptance test that shuts down a node, to test moving datasets off dead nodes. We can therefore not rely on a fixed public IP for all tests.

Did so by trusting Flocker's reported IPs for nodes by default. However, if the backend provides `ICloudAPI` we use that to figure out public IP for each node.

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -18,6 +18,8 @@ import ssl
 
 from docker.tls import TLSConfig
 
+from ipaddress import ip_address
+
 from twisted.web.http import OK, CREATED
 from twisted.python.filepath import FilePath
 from twisted.python.constants import Names, NamedConstant
@@ -45,6 +47,7 @@ from ..testtools import random_name
 from ..apiclient import FlockerClient, DatasetState
 from ..node.script import get_backend, get_api
 from ..node import dockerpy_client
+from ..node.agents.blockdevice import ICloudAPI
 
 from .node_scripts import SCRIPTS as NODE_SCRIPTS
 
@@ -826,7 +829,7 @@ class Cluster(PClass):
 
 def connected_cluster(
         reactor, control_node, certificates_path, num_agent_nodes,
-        hostname_to_public_address, username='user',
+        username='user',
 ):
     cluster_cert = certificates_path.child(b"cluster.crt")
     user_cert = certificates_path.child(
@@ -869,24 +872,56 @@ def connected_cluster(
                        failed_query)
         return d
     agents_connected = loop_until(reactor, nodes_available)
+    agents_connected.addCallback(_add_nodes)
+    return agents_connected
 
-    # Extract node hostnames from API that lists nodes. Currently we
-    # happen know these in advance, but in FLOC-1631 node identification
-    # will switch to UUIDs instead.
-    agents_connected.addCallback(lambda _: cluster.current_nodes())
+
+def _add_nodes(cluster):
+    """
+    Configure the ``Node`` objects for a newly created ``Cluster`` whose
+    nodes are known to be alive.
+
+    :param Cluster cluster: Cluster that still needs nodes set.
+    :return: ``cluster`` updated with appropriate ``nodes`` set.
+    """
+    # By default we just trust address returned by Flocker
+    def default_get_public_ip(address):
+        return address
+
+    try:
+        backend = get_backend_api(None, cluster.cluster_uuid)
+    except SkipTest:
+        # Can't load backend, will have to trust Flocker's reported IPs.
+        get_public_ip = default_get_public_ip
+    else:
+        if ICloudAPI.providedBy(backend):
+            node_ips = list(set(ip_address(i) for i in ips)
+                            for ips in backend.list_live_nodes().values())
+
+            def get_public_ip(address):
+                for ips in node_ips:
+                    if ip_address(address) in ips:
+                        return [unicode(ip) for ip in ips
+                                if not any(ip.is_private, ip.is_link_local,
+                                           ip.is_loopback)][0]
+                raise ValueError(
+                    "Couldn't find address in cloud API reported IPs")
+        else:
+            get_public_ip = default_get_public_ip
 
     def node_from_dict(node):
         reported_hostname = node["host"]
-        public_address = hostname_to_public_address.get(
-            reported_hostname, reported_hostname)
+        public_address = get_public_ip(reported_hostname)
         return Node(
             uuid=node[u"uuid"],
             public_address=public_address.encode("ascii"),
             reported_hostname=reported_hostname.encode("ascii"),
         )
-    agents_connected.addCallback(lambda nodes: cluster.set(
-        "nodes", map(node_from_dict, nodes)))
-    return agents_connected
+
+    d = cluster.current_nodes()
+    d.addCallback(
+        lambda nodes: cluster.set("nodes", map(node_from_dict, nodes)))
+    return d
 
 
 def _get_test_cluster(reactor):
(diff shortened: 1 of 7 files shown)


flocker/node/agents/blockdevice.py
```diff
@@ -1121,8 +1121,10 @@ def list_live_nodes():
         This is used to figure out which nodes are dead, so that other
         nodes can do the detach.
 
-        :returns: A collection of ``unicode`` compute instance IDs, compatible
-            with those returned by ``IBlockDeviceAPI.compute_instance_id``.
+        :returns: A mapping of ``unicode`` compute instance IDs
+            (compatible with those returned by
+            ``IBlockDeviceAPI.compute_instance_id``) to IPs of those
+            nodes, also as unicode..
```
sarum90: nit: extra `.`

flocker/node/agents/blockdevice.py
```diff
@@ -1647,7 +1649,8 @@ def is_existing_block_device(dataset_id, path):
                     pass
 
         if ICloudAPI.providedBy(self._underlying_blockdevice_api):
-            live_instances = self._underlying_blockdevice_api.list_live_nodes()
+            live_instances = list(
+                self._underlying_blockdevice_api.list_live_nodes())
```
sarum90: :grey_question: I personally prefer `dict(a=1, b=2).keys()` over `list(dict(a=1, b=2))` everything else being equal. Seems it's only 1 more character, but seems to more explicitly express intent.

TBH, I always think that turning a dictionary into a list should have the behavior of `dict().iteritems()`, and the mental burden of remembering that I'm wrong is tiresome :stuck_out_tongue:. Optional as this is obviously personal preference.

flocker/node/agents/blockdevice.py
```diff
@@ -1121,8 +1121,10 @@ def list_live_nodes():
         This is used to figure out which nodes are dead, so that other
         nodes can do the detach.
 
-        :returns: A collection of ``unicode`` compute instance IDs, compatible
-            with those returned by ``IBlockDeviceAPI.compute_instance_id``.
+        :returns: A mapping of ``unicode`` compute instance IDs
+            (compatible with those returned by
+            ``IBlockDeviceAPI.compute_instance_id``) to IPs of those
```
sarum90: I misunderstood this until I read the code further down, thinking it was one IP per compute instance. perhaps `as lists of unicode values` instead of `also as unicode`?

flocker/node/agents/blockdevice.py
```diff
@@ -1121,8 +1121,10 @@ def list_live_nodes():
         This is used to figure out which nodes are dead, so that other
```
sarum90: :grey_question: Potentially awkward interface note. This is kind of like a `Get details on the nodes, and filter results so that I only see the live nodes`. For instance, if this was directly an interface on GCE I would probably be a bit upset.

I think more natural would be 'list_nodes' with an optional filter for only live nodes.

flocker/node/agents/blockdevice.py
```diff
@@ -1121,8 +1121,10 @@ def list_live_nodes():
...
         nodes can do the detach.
 
-        :returns: A collection of ``unicode`` compute instance IDs, compatible
-            with those returned by ``IBlockDeviceAPI.compute_instance_id``.
+        :returns: A mapping of ``unicode`` compute instance IDs
```
sarum90: Thought about this more... I think we might want to introduce a new type here rather than returning a raw mapping.

Something like returning a list of `CloudComputeInstance`'s:

```
class CloudComputeInstance(PClass):
   id = field(type=unicode)
   ip_addresses = pvector_field(type=unicode)
```

Rational:

I don't know what the following does without looking at the source:
```
"ssh root@{}".format(api.list_live_nodes()['i-3ab8f212'][0])
```

I can guess what the following does without going to the source:
```
"ssh root@{}"format(next(x for x in api.list_live_nodes() if x.compute_id = 'i-3ab8f212').ip_addresses[0])
```

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -754,7 +754,12 @@ def __init__(self, block_api, live_nodes=()):
         self.live_nodes = live_nodes
 
     def list_live_nodes(self):
-        return [self.compute_instance_id()] + list(self.live_nodes)
+        result = {self.compute_instance_id():
+                  set(unicode(i) for i in get_all_ips()
+                      if i != b"127.0.0.1")}
+        result.update({node: [u"10.1.1.{}".format(i)]
+                       for i, node in enumerate(self.live_nodes)})
```
sarum90: :grey_question: would you mind documenting the type of `self.live_nodes` on line 751? Slowed my review down not having it there.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5501,11 +5506,25 @@ def test_current_machine_is_live(self):
...
+                        local_addresses)))
+            return d
+
         def test_list_live_nodes(self):
             """
             ``list_live_nodes`` returns an iterable of unicode values.
             """
-            live_nodes = self.api.list_live_nodes()
+            live_nodes = list(self.api.list_live_nodes())
             self.assertThat(live_nodes, AllMatch(IsInstance(unicode)))
```
sarum90: Considering how brittle getting the wrong type may be, strongly prefer to verify the types of everything returned.

```
self.assertThat(live_nodes.itieritems(), AllMatch(
    MatchesListwise(
        (IsInstance(unicode), AllMatch(IsInstance(unicode)))
    )
))
```

flocker/acceptance/testtools.py
```diff
@@ -869,24 +872,56 @@ def failed_query(failure):
...
+        get_public_ip = default_get_public_ip
+    else:
+        if ICloudAPI.providedBy(backend):
+            node_ips = list(set(ip_address(i) for i in ips)
+                            for ips in backend.list_live_nodes().values())
+
+            def get_public_ip(address):
+                for ips in node_ips:
+                    if ip_address(address) in ips:
+                        return [unicode(ip) for ip in ips
```
sarum90: prefer `list(unicode(ip) for ...)` rather than `[unicode(ip) ...]` so that we don't pollute the `ip` variable into the outer namespace.

flocker/acceptance/testtools.py
```diff
@@ -869,24 +872,56 @@ def failed_query(failure):
...
+                raise ValueError(
+                    "Couldn't find address in cloud API reported IPs")
+        else:
+            get_public_ip = default_get_public_ip
 
     def node_from_dict(node):
         reported_hostname = node["host"]
-        public_address = hostname_to_public_address.get(
-            reported_hostname, reported_hostname)
+        public_address = get_public_ip(reported_hostname)
```
sarum90: :grey_question: This is kinda gross. This assumes that `reported_hostname` is a unicode IP address, and will blow up if it is not. Which is bonus fun because it's named "hostname" and not "ip_address".

Perhaps okay since this is test code, and it blows up... but still.

flocker/acceptance/testtools.py
```diff
@@ -869,24 +872,56 @@ def failed_query(failure):
...
         return Node(
             uuid=node[u"uuid"],
             public_address=public_address.encode("ascii"),
             reported_hostname=reported_hostname.encode("ascii"),
```
sarum90: :grey_question: ew at this. But it looks like it's nearly unused and pending removal, so probably not worth fixing up.

sarum90:
Good stuff, looks like a nice way to work around the problem.

Although, it seems that the loopback tests are currently failing with this change :(

I'd really like to see a slightly better interface for listing the live nodes (specifically returning something that is not a map from compute_instance_id to list of ip addresses).  I would be much more happy about maintaining something that looked like that long term.

For now I think I'm going to ask for a second review after things are patched up. Should be a quick second review though, just to verify the fix for the presently failing loopback acceptance tests and look at the new interface.

flocker/node/agents/blockdevice.py
itamarst: I do think we need to rewrite this before it's public. I'll add a comment to that effect.

itamarst:
The tests are failing because loopback backend doesn't support `ICloudAPI`. I think the way to solve this is to restore the env variables for public IP, and use them if `ICloudAPI` is not available on the backend. Since `ICloudAPI` isn't available the shutdown test won't run, so the public IP won't be invalidated and using the env variables will therefore be OK.

PR #2536, opened 2016-01-27
[FLOC-3990] Enable pylint linting: a start

itamarst:
1. Added pylint to build.          
2. Configured pylint to only use meaningful/non-spammy warnings and errors. We skipped all the random "your code is terrible" messages.
3. Fixed real bug in flocker.dockerplugin._api.                             
4. Fixed a bunch more minor warnings/bugs.

There are many more warnings we could enable, but we'll do that in follow-up branches. This is already a good improvement.

You can see the functional Docker plugin test failing without the fix here (IN THEORY, in practice it is still running): http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/test-docker-plugin-bug/view/acceptance/job/run_acceptance_loopback_on_AWS_Ubuntu_Trusty_for_flocker.acceptance/1/console


itamarst:
For the before and after Docker plugin tests need custom VMs with Docker 1.10... so that'll take some manual setup.

.pylintrc
```diff
@@ -0,0 +1,586 @@
...
+disable=all
+
+enable=import-error,
+       import-self,
+       reimported,
+       wildcard-import,
+       misplaced-future,
+       import-error,
+       import-self,
+       reimported,
```
james-w: These 3 are duplicated at the start.

.pylintrc
```diff
@@ -0,0 +1,586 @@
...
+       relative-import,
+       deprecated-module,
+       wildcard-import,
+       misplaced-future,
```
james-w: These two are duplicated earlier.

.pylintrc
```diff
@@ -0,0 +1,586 @@
...
+notes=FIXME,XXX,TODO
+
+
+[BASIC]
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,input
+
+# Good variable names which should always be accepted, separated by a comma
+good-names=i,j,k,ex,Run,_
```
james-w: :question: do we want "d" in this list for deferreds.

.pylintrc
```diff
@@ -0,0 +1,586 @@
...
+[MISCELLANEOUS]
+
+# List of note tags to take in consideration, separated by a comma.
+notes=FIXME,XXX,TODO
+
+
+[BASIC]
+
+# List of builtins function names that should not be used, separated by a comma
+bad-functions=map,filter,input
```
james-w: :question: map and filter seem like perfectly good functions to me.

itamarst: It's just some random garbage pylint sets by default. The actual messages warning about using those are disabled.

itamarst:
Note this is not yet ready for review pending validating that the Docker plugin fix actually works.

.pylintrc
itamarst: We don't use this at all, currently.

itamarst:
@brendan-donegan In the future could you commit on the PR itself, rather than commits? ("Files Changed" tab). Makes it easier to see comments in context of code. Thanks!

brendan-donegan:
@itamarst Yes I can see now it's not ideal to comment on commits - noted

itamarst:
Re disabling and then re-enabling, that's because if you put it on same line the lines gets too long, and flake8 complains.

itamarst:
Re random stuff in pylintrc - I suggest ignoring most of the configuration, it's auto-generated. In particular, we're deliberately not disabling most of the pylint warnings because they're useless garbage or too noisy, so most of the options aren't even used. This branch just enables a first set of warnings and errors that are legitimate "this finds a real problem". Later branches will enable more of those.

Eventually we might add coding standard style warnings, but that's a ways off. For now this is just focusing on finding actual or potential bugs (which it has!).

itamarst:
With Docker 1.10rc2:
1. The acceptance test fails with old code.
2. Passes with new code.

I have separately validated the bug exists via `docker` CLI: with original code one can create two volumes, but `docker volume ls` doesn't show their respective names, it shows same name for both. After fix it shows their differing names.

itamarst:
OK, addressed comments above. Ready for final review.

james-w:
:heavy_check_mark: looks good. Thanks for adding to the quality checks,
while simultaneously fixing a bug and proving the value of static analysis.

Merge when you are happy with it.

PR #2537, opened 2016-01-27
[FLOC-3988] Logs for timeouts

jml:
Makes it so that we get logs for tests that time out. 

Relies on switching the branch we use for testtools. Rather than clusterhq-fork, which was becoming a bit of an unnavigable nightmare, we're using optional-twisted-logging, which is a clone of the branch that's up for review at testing-cabal/testtools#172.

The hack is a bit gross. I'll try to get something nicer upstream.


james-w:
:heavy_check_mark: Looks good to me, thanks for caring for failure cases. Merge when green.

james-w:
:heavy_check_mark: still looks good.

That's rather an obscure bug, and a shame that it's a bit of a "more magic" line of
code for us now.

PR #2539, opened 2016-01-27
[FLOC-3995] Allow running cron jobs on release branches without merging forward.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3995


james-w:
Hi,

This looks good, thanks for fixing it.

I think there is one unintended change in here, it will mean that the
cronly_jobs now notify github of their status.

I think that's probably ok, though we might want to stop that
happening, I'm not entirely sure.

Similarly it makes me realise that the release jobs that don't merge
master will be conflicting with the results for merging the release
branch back to master.

Please go ahead and merge when you are happy with it.

tomprince:
> I think there is one unintended change in here, it will mean that the
cronly_jobs now notify github of their status

I don't think that is a problem, and definitely beneficial for release builds. If it is for master, we can address that directly.

PR #2540, opened 2016-01-28
[FLOC-3947] Tool for adding new nodes to an existing Flocker cluster

avg-I:
Also, `setup-cluster` now uses alternative implementation of `LibcloudRunner.start_cluster`.
The alternative implementation should be faster as it creates cloud nodes
in parallel.  Also it should be more robust as it does not fail as long
as it is able to create at least a control node for the cluster.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -458,7 +458,7 @@ def make_managed_nodes(addresses, distribution):
 
 def get_default_volume_size(dataset_backend_configuration):
     """
-    Calcuate the default volume size for the given dataset backend
+    Calculate the default volume size for the given dataset backend
     configuration.
 
     :param dataset_backend_configuration: The backend configuration.
@@ -649,7 +649,7 @@ class LibcloudRunner(object):
                     self._destroy_node(node)
                     return failure
 
-                # Make sure too return the node.
+                # Make sure to return the node.
                 d.addCallbacks(lambda _: node, errback=provisioning_failed)
                 return d
 
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -644,14 +644,13 @@ class LibcloudRunner(object):
             def provision(node):
                 d = self._provision_node(reactor, node)
 
-                def error(failure):
+                def provisioning_failed(failure):
                     # Destroy a node if we failed to provision it.
                     self._destroy_node(node)
                     return failure
 
-                d.addErrback(error)
                 # Make sure too return the node.
-                d.addCallback(lambda _: node)
+                d.addCallbacks(lambda _: node, errback=provisioning_failed)
                 return d
 
             d.addCallback(provision)
@@ -720,7 +719,7 @@ class LibcloudRunner(object):
             cluster.agent_nodes.append(node)
             return node
 
-        def error(failure):
+        def configure_failed(failure):
             print "Failed to configure node {}".format(node.name)
             write_failure(failure)
             if node is not cluster.control_node:
@@ -730,8 +729,7 @@ class LibcloudRunner(object):
             # the partial success.
             return failure
 
-        d.addErrback(error)
-        d.addCallback(add_node)
+        d.addCallbacks(add_node, errback=configure_failed)
         return d
 
     def extend_cluster(self, reactor, cluster, count, tag, starting_index):
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -36,7 +36,6 @@ from flocker.common import (
     RACKSPACE_MINIMUM_VOLUME_SIZE,
     gather_deferreds,
     loop_until,
-    poll_until,
 )
 from flocker.provision import PackageSource, Variants, CLOUD_PROVIDERS
 from flocker.provision._ssh import (
@@ -603,22 +602,6 @@ class LibcloudRunner(object):
             write_failure(Failure())
             raise
 
-    def _create_node_with_retry(self, name, retries=10):
-        """
-        Create a cloud node with the given name retrying on failures.
-
-        :param str name: The name.
-        :param int retries: Maximum number of retries.
-        :return: Node instance.
-        """
-        def create():
-            try:
-                return self._create_node(name)
-            except Exception:
-                return None
-
-        return poll_until(create, repeat(0, retries))
-
     def _provision_node(self, reactor, node):
         """
         Make the given node ready for Flocker installation by provisioning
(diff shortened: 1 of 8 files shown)


admin/acceptance.py
```diff
@@ -195,6 +203,23 @@ def ensure_keys(reactor):
             found.  Otherwise, fails with ``AgentNotFound`` or ``KeyNotFound``.
         """
 
+    def extend_cluster(reactor, cluster, count, tag, starting_index):
+        """
+        Create new nodes and add them to the given cluster.
+
+        :param reactor: Reactor to use.
```
brendan-donegan: To be honest, the practice of specifying types in docstrings is not something I've seen before but we should at least be consistent in how it's done. Should then be 'Reactor reactor'. Really Reactor to use is not very descriptive - what is the reactor used for?

admin/acceptance.py
```diff
@@ -195,6 +203,23 @@ def ensure_keys(reactor):
...
+        :param int count: How many new nodes to create.
+        :param str tag: Tag used for naming the new nodes.
+        :param int starting_index: A number used for naming the new nodes.
+        :return: List of ``Deferred``s that fire when a new node is configured
+            or fails.
+
+        .. note::
+            The cluster object is modified by this call and, thus, it should be
+            mutable
+        """
```
brendan-donegan: How would it not be mutable, it's an object?

admin/acceptance.py
```diff
@@ -411,6 +425,68 @@ def _save_backend_configuration(dataset_backend_name,
...
+        # A list of strings.
+        return [
+            ManagedNode(address=address, distribution=distribution)
+            for address in addresses
+        ]
+
+
+def get_default_volume_size(dataset_backend_configuration):
+    """
+    Calcuate the default volume size for the given dataset backend
```
brendan-donegan: Typo, s/calcuate/calculate/

admin/acceptance.py
```diff
@@ -528,19 +577,196 @@ def start_cluster(self, reactor):
...
+        Configure the given node as an agent node for the given cluster.
+
+        :param reactor: The reactor.
+        :param reactor: The cluster.
+        :param Node node: The node.
+        :param int index: The index of the ndoe in the cluster.
+        :return: Deferred that fires with the :param:`node` when it is
+            configured.
+        """
+        node_certnkey = cluster.certificates.add_node(index)
```
brendan-donegan: What does 'certnkey' mean? Is it an abbreviation of something? Should I read it as 'cert_n_key', 'certn_k

admin/acceptance.py
```diff
@@ -528,19 +577,196 @@ def start_cluster(self, reactor):
...
+            write_failure(failure)
+            if node is not cluster.control_node:
+                self._destroy_node(node)
+            # Failure to add a node to a cluster is not fatal.
+            # As long as we have the control node we can claim
+            # the partial success.
+            return failure
+
+        d.addErrback(error)
+        d.addCallback(add_node)
```
brendan-donegan: Could this instead be 'd.addCallbacks(add_node, errback=error)

Also maybe 'add_node_failed' is a better name for the errback

admin/acceptance.py
```diff
@@ -528,19 +577,196 @@ def start_cluster(self, reactor):
...
+
+        def add_node(node, index):
+            return self._add_node_to_cluster(reactor, cluster, node, index)
+
+        for i, d in enumerate(results):
+            d.addCallback(add_node, starting_index + i)
+
+        return results
+
+    @inlineCallbacks
```
brendan-donegan: Our engineering handbook says not to use inlineCallbacks. It's not even clear how this function is using them anyway, although my knowledge of them is little so I'm probably missing something. http://twistedmatrix.com/documents/10.2.0/api/twisted.internet.defer.inlineCallbacks.html suggests it works by yielding from functions that return Deferreds but I don't see any yields in here.

admin/cluster_add_nodes.py
```diff
@@ -0,0 +1,203 @@
...
+    def _check_cert_directory(self):
+        cert_path = FilePath(self['cert-directory'])
+        self['cert-directory'] = cert_path
+        if not cert_path.exists():
+            raise UsageError("{} does not exist".format(cert_path.path))
+        if not cert_path.isdir():
+            raise UsageError("{} is not a directory".format(cert_path.path))
+
+
+@inlineCallbacks
```
brendan-donegan: I do see a yield in here, but the comment about inlineCallbacks being 'banned' applies still

admin/cluster_setup.py
```diff
@@ -11,23 +11,159 @@
...
-from twisted.python.usage import UsageError
+from twisted.internet.defer import DeferredList, inlineCallbacks
 from twisted.python.filepath import FilePath
+from twisted.python.usage import UsageError
 
 from .acceptance import (
+    CLOUD_PROVIDERS,
     ClusterIdentity,
     CommonOptions,
+    LibcloudRunner as OldLibcloudRunner,
```
brendan-donegan: Importing with an alias then subclassing using that alias is a slightly odd pattern. Why not just give a descriptive subclass name?

admin/cluster_setup.py
```diff
@@ -11,23 +11,159 @@
...
+    """
+    An alternative approach to setting up a cluster using
+    a libcloud-compatible provisioner.
+    """
+    def _setup_control_node(self, reactor, node, index):
+        print "Selecting node {} for control service".format(node.name)
+        certificates = Certificates.generate(
+            self.cert_path,
+            node.address,
+            0,
```
brendan-donegan: I think self.cert_path, node.address and particularly '0' deserve the <argument>= in front of them. Someone reading this would have no idea what 0 is, specifically

admin/cluster_setup.py
```diff
@@ -149,6 +319,24 @@ def create_managed_config(base_config, cluster):
...
+    :param Cluster cluser: The cluster.
+    """
+    managed_config_file = directory.child("managed.yaml")
+    managed_config = create_managed_config(base_config, cluster)
+    managed_config_file.setContent(
+        yaml.safe_dump(managed_config, default_flow_style=False)
+    )
+
+
 @inlineCallbacks
```
brendan-donegan: More @inlineCallbacks

admin/acceptance.py
jongiddy: The reactor is the core Twisted event loop runner. Hence why it is pervasive through the code, and generally not documented everywhere it appears.  If it is typed, it is usually as `twisted.internet.interfaces.IReactorTime` which is the interface shared by Twisted reactors and the `twisted.internet.task.Clock` used for testing.

admin/acceptance.py
jongiddy: Although I am not answering the question - I will leave that to Andriy - note that using `addCallbacks` is semantically different.  As the code is now, if the `error` function returns any object other than a `Failure`, it will then call `add_node`.

Changing to `addCallbacks` would make it call one or the other, never both.

Given that the `error` function appears to always return a `Failure`, they are probably effectively the same.  So, yes, it should be changed unless `error` could potentially offer a path to redemption (from error state back to normal flow).

admin/acceptance.py
jongiddy: I can see `yield`s, but not in the added code, later in the function.

inlineCallbacks do get used occasionally, always outside the core Flocker code, and more relevantly, in the acceptance testing code from which this code was derived. You can consider it the "grandfathering" loophole of the engineering guide.

admin/acceptance.py
avg-I: We are not super-consistent with documenting types. I always try to document them. But the reactor, as Jon explains above, is assumed to be self-evident.

admin/acceptance.py
avg-I: We use `pyrsistent` to have immutable objects. In particular, `Cluster` is a `PClass` and its node lists are typically `pvector`s. In this case we require that they are regular lists.

admin/acceptance.py
avg-I: Fixed.

admin/cluster_add_nodes.py
```diff
@@ -0,0 +1,203 @@
...
+    :param list args: The arguments passed to the script.
+    :param FilePath base_path: The executable being run.
+    :param FilePath top_level: The top-level of the Flocker repository.
+    """
+    add_destination(eliot_output)
+    options = RunOptions(top_level=top_level)
+    try:
+        options.parseOptions(args)
+    except UsageError as e:
+        sys.stderr.write("%s: %s\n" % (base_path.basename(), e))
```
jongiddy: Add `sys.stderr.write(options.getUsage())` so these errors display the full usage message.

admin/acceptance.py
avg-I: Yes :-) It's "certificate and key", the abbreviation is picked up from the existing code in `flocker/provision/_install.py`.

admin/acceptance.py
avg-I: Makes sense. I am making the change. And I'll see if there are other similar cases.

admin/cluster_setup.py
avg-I: Well, the "new" `LibcloudRunner` is designed as a drop-in replacement for the old `LibcloudRunner` and the replacement is supposed to happen "very soon". Hence, the pattern.  But I am open to changing the name.

admin/cluster_setup.py

admin/acceptance.py
avg-I: Regarding
> Also maybe 'add_node_failed' is a better name for the errback

I will make the errback's name more descriptive, just a note that it's chained before `add_node`.
Thanks!

admin/cluster_add_nodes.py
avg-I: Done.

avg-I:
There is a merge conflict with master. I am merging master into this branch.

admin/acceptance.py
brendan-donegan: I think it's a bad name - I would at least not use it here, but maybe also change it elsewhere it is used. I don't see any reason not to call it cert_and_key

avg-I: Indeed. Fixing.

PR #2542, opened 2016-01-28
[FLOC-3998] Flocker dataset config

dwgebler:
Fixes https://clusterhq.atlassian.net/browse/FLOC-3998

Ensures that the loaded config contains at a minimum the parameters required by the chosen backend's API factory, preventing failure with an ugly `TypeError` missing parameters traceback when the config is missing data.


jml:
LGTM. Thanks! Please merge when tests pass (I don't see any test results for this)

PR #2543, opened 2016-01-28
using only control service in write setup

mpgomez:
I feel awfully bad for creating an issue for this.

I misunderstood what I needed to refactor. Bellow the one line of code changes.

Thank you gvernor for your time. Whoever you are :)


james-w:
:heavy_check_mark: This looks good, thanks for the improvement.

I think that if you don't make a JIRA ticket then most people will miss the PR,
because they look at JIRA to see what needs reviewing, so I suggest you
either make the ticket, or mention the PR on slack to get a reviewer. I only
found it because I noticed the email and wondered why I hadn't seen it on
JIRA.

PR #2545, opened 2016-01-28
[FLOC-4002] Script to demangle Eliot logs

jml:
```
./admin/filter-eliot [PATH]
```

Takes test.log as input and produces valid Eliot as output.

### DIFF
--- a/admin/eliotlogging.py
+++ b/admin/eliotlogging.py
@@ -1,0 +1,51 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Filter Eliot logs out of Twisted test.log.
+"""
+
+import sys
+
+from flocker.testtools._base import _get_eliot_data
+
+
+def chain_files(paths):
+    """
+    Iterate through many files, line by line.
+    """
+    for path in paths:
+        with open(path, 'r') as f:
+            for line in f:
+                yield line
+
+
+def process_line(line):
+    eliot_line = _get_eliot_data(line)
+    if eliot_line:
+        yield eliot_line
+        yield '\n'
+
+
+def filter_eliot_main(args, base_path, top_level, stdin=None, stdout=None):
+    """
+    Filter logs.
+
+    :param args: Iterable of file names to read logs from. If empty, reads
+        from stdin.
+    :param FilePath base_path: The path to the executable.
+    :param FilePath top_level: The path to the directory where the flocker
+        package is.
+    :param file stdin: File-like object to read from if ``args`` is empty.
+        If ``None``, defaults to STDIN.
+    :param file stdout: File-like object to write filtered logs to. If
+        ``None``, defaults to STDOUT.
+    """
+    stdin = sys.stdin if stdin is None else stdin
+    stdout = sys.stdout if stdout is None else stdout
+    if args:
+        lines = chain_files(args)
+    else:
+        lines = stdin
+    for line in lines:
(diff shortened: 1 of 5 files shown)


admin/eliotlogging.py
```diff
@@ -0,0 +1,51 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Filter Eliot logs out of Twisted test.log.
+"""
+
+import sys
+
+from flocker.testtools._base import _get_eliot_data
```
sarum90: We should probably not be reaching into a private module to grab a private function. 

Care to make these public before you use them?

(Also probably rename `_get_eliot_data` to something more precise, and document its parameter)

admin/eliotlogging.py
```diff
@@ -0,0 +1,51 @@
...
+    Iterate through many files, line by line.
+    """
+    for path in paths:
+        with open(path, 'r') as f:
+            for line in f:
+                yield line
+
+
+def process_line(line):
+    eliot_line = _get_eliot_data(line)
```
sarum90: :grey_question: I think _get_eliot_data can raise ValueErrors if the line does not have the ELIOT marker in it. It'd be nice if a single bad line in the logs didn't cause the whole utility to crash (My biggest grievance about eliot-tree and the like).

The tool I want the most would be one that would print to STDERR the line numbers that had errors and what the errors were (perhaps aggregating similar errors), and would print out all lines that didn't have errors on stdout for consumption by eliot-tree or whatnot.

sarum90:
Cool! beats the sed script I always use.

Couple other nice-to-haves that I'll do in a follow up if you don't do here.

- Only output lines that are valid JSON (perhaps with task_uuid).
- Pass through lines that are already JSON (no `ELIOT:`), enables use cases like:
  `admin/filter-eliot /tmp/debug2/archive/{*.log,_trial_temp/test.log} | eliot-tree -l 0 `

admin/eliotlogging.py
sarum90: Also looking at its implementation, and looking at a `test.log`... what would happen to a line with:

```
... [HTTP11ClientProtocol (TLSMemoryBIOProtocol),client] ELIOT: {...
```

I think it would just be dropped since it didn't have `[-]`? Seems like that method could be improved. Maybe we shouldn't use it? Another approach would be to strip off everything before the first '{' and everything after the final '}', then print the line so long as it can be parsed as JSON.

jml:
I _think_ your suggested implementation of `_get_eliot_data` (now `extract_eliot_from_twisted_log`) achieves the two future work things you suggest.

admin/eliotlogging.py
jml: Done both. I wish Eliot had terms that were better defined, because I don't really know what to call a JSON-encoded Eliot message/action.

jml:
Thanks for the review! I'm waiting for tests to pass, but please feel free to merge before then.

sarum90:
I think you should probably use `rfind` rather than `find` for the closing bracket...

and.... I think that strongly enough that I'm going to write the commit and push it, then re-kick the tests.

sarum90:
Leaving to you to review my commit and merge when you are happy.

jml:
> `rfind` I think would be better here.

Yes. I used `rfind` in my repl experimentation and somehow didn't make it all the way to my editor.

jml:
Great, thanks so much!

PR #2546, opened 2016-01-28
[FLOC-3739] Failover acceptance test enabled

itamarst:
1. Enable the test on EBS, for now.
2. Fix bug in EBS code that was caught by this. (The API for starting nodes is only used by this test, so it's a bit of a hack...).
3. Fix some bugs in the test itself.
4. News entry since acceptance test means the implementation is hopefully going to *continue* to work, so we can tell people about it.

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -16,6 +16,7 @@ Next Release
 * The new :ref:`CloudFormation installer <cloudformation>` has been made available, to provide a far simpler installation experience for users on AWS.
 * The :ref:`Flocker plugin for Docker <plugin>` should support the direct volume listing and inspection functionality added to Docker 1.10.
 * Fixed a regression that caused block device agents to poll backend APIs like EBS too frequently in some circumstances.
+* Datasets can now be moved off of shutdown EC2 instances. Previously they could only be moved off of terminated instances.
 
 This Release
 ============
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -97,7 +97,7 @@ from ..loopback import (
     _backing_file_name,
 )
 from ....common.algebraic import tagged_union_strategy
-
+from ....common import get_all_ips
 
 from ... import run_state_change, in_parallel, ILocalState, IStateChange, NoOp
 from ...testtools import (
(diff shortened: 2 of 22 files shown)


james-w:
The two tests keep failing, and then get timed out by Jenkins before they can
print the errors. I've increased the timeout on those two jobs on Jenkins to see
what the failures are.

Note that with these changes if it now takes more than 45 minutes for
run_acceptance_loopback_on_AWS_CentOS_7_for_flocker.acceptance
and run_acceptance_loopback_on_AWS_Ubuntu_Trusty_for_flocker.acceptance
to pass then build.yaml will have to be updated for them.

Pushing to this branch will overwrite my cowboy, so you might want to
change the timeouts in build.yaml if you have to do more work on this
branch, even if they are reverted before merging to master because
a passing run takes less than 45 minutes.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -5501,11 +5506,25 @@ def test_current_machine_is_live(self):
...
+        def test_current_machine_has_appropriate_ip(self):
+            """
+            The machine's known IP is set for the current node.
+            """
+            local_addresses = set(unicode(i) for i in get_all_ips()
+                                  if i != b"127.0.0.1")
+            d = self.async_cloud_api.list_live_nodes()
+            d.addCallback(
+                lambda live:
+                self.assertTrue(
```
james-w: Does assertIn work here?

docs/releasenotes/index.rst
```diff
@@ -16,6 +16,7 @@ Next Release
 * The new :ref:`CloudFormation installer <cloudformation>` has been made available, to provide a far simpler installation experience for users on AWS.
 * The :ref:`Flocker plugin for Docker <plugin>` should support the direct volume listing and inspection functionality added to Docker 1.10.
 * Fixed a regression that caused block device agents to poll backend APIs like EBS too frequently in some circumstances.
+* Datasets can now be moved off of shutdown EC2 instances. Previously they could only be moved off of terminated instances.
```
james-w: Is there anything that prevents someone trying this on Rackspace?

james-w:
:heavy_check_mark: I'm happy with this, thanks for the improvements.

Merge when you are happy and there are green tests, but I suspect there
may be work to do to get passing tests. A re-review of any changes needed
for that is at your discretion.

docs/releasenotes/index.rst
itamarst: I've tried it on Rackspace manually. It doesn't work due to Rackspace bugs. It *does* work on OpenStack elsewhere (at least on Swisscom's cluster) but I don't want to claim that publicly until we can enable the acceptance tests.

james-w: Right, but is it a matter of hoping that no one tries it? If so are we
happy with that?

PR #2547, opened 2016-01-28
[FLOC-3525] Roll forward flocker upgrade acceptance test 

sarum90:
This branch is mostly a re-submission of a branch that caused a regression in Centos7 on rackspace.

Thus http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/reapply-flocker-acceptance-test-FLOC-3525/job/_run_acceptance_on_Rackspace_CentOS_7_with_Cinder/ is probably the most interesting test to review.

To assist in your review I will annotate those results below:

1-18: Me trying things. Instead I probably should have used buildbot to build the package and iterated on run_acceptance_tests in my local checkout

19-25: Flaky results because I was restarting the control agent in parallel with the dataset agent. Because of firewalld rule ordering and restarting docker, it is actually important to restart the dataset agent (and docker) _after_ the control agent is restarted.

26+: runs on the final code.

27: A flake in node creation, presumed to be rackspace slowness.

Summary of changes since last submit:

- Retry version getting, something about the firewall/networking setup in Rackspace meant that sometimes when the trial process was trying to get the version of flocker from the control node before the control agent had restarted it would fail. Simple retry until the control agent is restarted seems to have improved that.
- Re-opening the firewall after the control agent and docker were restarted. This seems to be required only on Rackspace,CentOS7 and not on other distros/providers.
- Figured out the order in which docker, the control agents, and firewalld needed to be restarted in order for flocker to function.
- Some review notes on getting environment variables.


flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,112 @@ def test_dataset_creation(self, cluster):
...
+        :param unicode default_version: The version of flocker to use
+            if not specified in environment variables.
+
+        :return: A ``PackageSource`` that can be used to install the version of
+            flocker under test.
+        """
+        env_vars = ['FLOCKER_ACCEPTANCE_PACKAGE_BRANCH',
+                    'FLOCKER_ACCEPTANCE_PACKAGE_VERSION',
+                    'FLOCKER_ACCEPTANCE_PACKAGE_BUILD_SERVER']
+        missing_vars = list(x for x in env_vars if x not in os.environ)
```
brendan-donegan: Why not plain old list comprehension syntax here?

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,112 @@ def test_dataset_creation(self, cluster):
...
+        if missing_vars:
+            raise SkipTest(
+                'Missing environment variables for upgrade test: %s.' %
+                ', '.join(missing_vars))
+        version = (os.environ.get('FLOCKER_ACCEPTANCE_PACKAGE_VERSION') or
+                   default_version)
+        return PackageSource(
+            version=version,
+            branch=os.environ.get('FLOCKER_ACCEPTANCE_PACKAGE_BRANCH'),
+            build_server=os.environ['FLOCKER_ACCEPTANCE_PACKAGE_BUILD_SERVER'])
```
brendan-donegan: Is there a specific reason to 'get' PACKAGE_BRANCH but key for PACKAGE_BUILD_SERVER? Seems like get could and should be used in both cases

flocker/acceptance/endtoend/test_dataset.py
sarum90: If I understand your question, then the answer is namespace pollution:
```
>>> [x for x in range(6) if x < 3]
[0, 1, 2]
>>> x
5
>>> list(y for y in range(6) if y < 3)
[0, 1, 2]
>>> y
NameError: name 'y' is not defined
>>>
```

I'd rather x not be defined after this line is executed.

flocker/acceptance/endtoend/test_dataset.py
sarum90: Hrm... actually I've got a guard above verifying that they both exist. Agreed that they should both be the same though.

Given that I've got the check above, I think I want the world to blow up (`KeyError` will suffice) if they actually don't exist, as that would indicate that my above checking is not working.

sarum90:
This branch is mostly a re-submission of a branch that caused a regression in Centos7 on rackspace.

Thus http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/reapply-flocker-acceptance-test-FLOC-3525/job/_run_acceptance_on_Rackspace_CentOS_7_with_Cinder/ is probably the most interesting test to review.

To assist in your review I will annotate those results below:

1-18: Me trying things. Instead I probably should have used buildbot to build the package and iterated on run_acceptance_tests in my local checkout

19-25: Flaky results because I was restarting the control agent in parallel with the dataset agent. Because of firewalld rule ordering and restarting docker, it is actually important to restart the dataset agent (and docker) _after_ the control agent is restarted.

26+: runs on the final code.

27: A flake in node creation, presumed to be rackspace slowness.

Summary of changes since last submit:

- Retry version getting, something about the firewall/networking setup in Rackspace meant that sometimes when the trial process was trying to get the version of flocker from the control node before the control agent had restarted it would fail. Simple retry until the control agent is restarted seems to have improved that.
- Re-opening the firewall after the control agent and docker were restarted. This seems to be required only on Rackspace,CentOS7 and not on other distros/providers.
- Figured out the order in which docker, the control agents, and firewalld needed to be restarted in order for flocker to function.
- Some review notes on getting environment variables.

admin/acceptance.py
```diff
@@ -112,17 +115,23 @@ def get_trial_environment(cluster):
             cluster.default_volume_size
         ),
         'FLOCKER_ACCEPTANCE_TEST_VOLUME_BACKEND_CONFIG':
-            cluster.dataset_backend_config_file.path
+            cluster.dataset_backend_config_file.path,
+        'FLOCKER_ACCEPTANCE_DISTRIBUTION': cluster.control_node.distribution,
+        'FLOCKER_ACCEPTANCE_PACKAGE_BRANCH': package_source.branch or '',
+        'FLOCKER_ACCEPTANCE_PACKAGE_VERSION': package_source.version or '',
+        'FLOCKER_ACCEPTANCE_PACKAGE_BUILD_SERVER': package_source.build_server,
```
wallrj: :question: It'd be nice if some day we had ``run-acceptance-tests --repository`` and ``PackageSource.repository`` instead of ``--build-server``.

admin/acceptance.py
```diff
@@ -927,6 +936,14 @@ def dataset_backend(self):
                 )
             )
 
+    def package_source(self):
+        """Getter for the configured package source."""
```
wallrj: :question: Use the house style for docstrings.

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,112 @@ def test_dataset_creation(self, cluster):
...
+        env_vars = ['FLOCKER_ACCEPTANCE_PACKAGE_BRANCH',
+                    'FLOCKER_ACCEPTANCE_PACKAGE_VERSION',
+                    'FLOCKER_ACCEPTANCE_PACKAGE_BUILD_SERVER']
+        missing_vars = list(x for x in env_vars if x not in os.environ)
+        if missing_vars:
+            raise SkipTest(
+                'Missing environment variables for upgrade test: %s.' %
+                ', '.join(missing_vars))
+        version = (os.environ['FLOCKER_ACCEPTANCE_PACKAGE_VERSION'] or
+                   default_version)
```
wallrj: :question: alternatively ``os.environ.get('FLOCKER_ACCEPTANCE_PACKAGE_VERSION', default_version)``

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,112 @@ def test_dataset_creation(self, cluster):
...
+    def test_upgrade(self, cluster):
+        """
+        Given a dataset created and used with the previously installable
+        version of flocker, uninstalling the previous version of flocker and
+        installing HEAD does not destroy the data on the dataset.
+        """
+        node = cluster.nodes[0]
+        SAMPLE_STR = '123456' * 100
+
+        upgrade_from_version = get_installable_version(version)
```
wallrj: :question: `version` confused me for a while. This is the module imported `version` rather than the local `version` used in nearby functions. Might have been clearer to ``from flocker import __version__ as INSTALLED_FLOCKER_VERSION`` .

flocker/common/version.py
```diff
@@ -137,7 +137,8 @@ def get_installable_version(version):
     installed (CLI and node).
     """
     parsed_version = _parse_version(version)
-    return parsed_version.installable_release
+    returned_version = parsed_version.installable_release
+    return returned_version
```
wallrj: :question: Looks like an unnecessary change. Maybe there was some debug code in here during development?

flocker/acceptance/endtoend/test_dataset.py
brendan-donegan: Probably not significant, but these aren't quite the same:

>>> os.environ['MINE']
''
>>> os.environ.get('MINE')
''
>>> os.environ.get('MINE') or 'default'
'default'

flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -35,6 +40,112 @@ def test_dataset_creation(self, cluster):
...
+
+        # Get the initial flocker version and setup a cleanup call to restore
+        # flocker to that version when the test is done.
+        d = cluster.client.version()
+        original_package_source = [None]
+
+        def setup_restore_original_flocker(version):
+            original_package_source[0] = (
+                self._get_package_source(
+                    default_version=str(version.get('flocker')) or None)
```
wallrj: :question: ``version`` is like ``dict(flocker="1.9.0.dev.asdsadsa")`` right? So why not just ``version.get('flocker', None)`` ?

flocker/acceptance/testtools.py
```diff
@@ -547,13 +572,14 @@ def created():
             def got_results(results):
                 # State has unpredictable path, so we don't bother
                 # checking for its contents:
-                actual_dataset_states = [d.set(path=None) for d in results]
-                return expected_dataset_state in actual_dataset_states
+                actual_dataset_states = list(
+                    d for d in results
+                    if d.set('path', None) == expected_dataset_state)
+                return (actual_dataset_states or [None])[0]
```
wallrj: :question: I may just be old and slow but I find these expressions difficult to read. Personally I'd rather see:
```
if matching_dataset_states:
    [matching_dataset_state] = matching_dataset_states
    return matching_dataset_state
else:
    return None
```

(or something similar)

flocker/acceptance/testtools.py
```diff
@@ -703,6 +729,59 @@ def current_nodes(self):
...
 
     @log_method
+    def install_flocker_version(self, package_source):
+        """
+        Change the version of flocker installed on all of the nodes to the
+        version indicated by `package_source`.
+        """
+        control_node_address = self.control_node.public_address
+        all_cluster_nodes = set([x.public_address for x in self.nodes] +
+                                [control_node_address])
```
wallrj: :question: This'll leak an `x` too, right?

flocker/acceptance/testtools.py
```diff
@@ -703,6 +729,59 @@ def current_nodes(self):
...
+            # Retry getting the flocker version for 10 seconds. Flocker might
+            # not be running yet on the control node when this is called.
+            d = timeout(
+                self.reactor,
+                retry_failure(
+                    self.reactor,
+                    self.client.version
+                ),
+                10
+            )
```
wallrj: :question: Why not just use ``retry_failure`` with a sequence of steps? I haven't seen this pattern before.

flocker/acceptance/testtools.py
```diff
@@ -703,6 +729,59 @@ def current_nodes(self):
...
+
+        d = get_flocker_version()
+
+        # If we fail to get the current version, assume we must reinstall
+        # flocker.
+        d.addErrback(write_failure)
+
+        def reinstall_if_needed(v):
+            if v and v == package_source.version:
+                return v
```
wallrj: :question: So if we failed to get a version from the control service `v` might be the `None` returned by `write_failure` ? I wonder if that could be made a little clearer?

flocker/acceptance/testtools.py
```diff
@@ -703,6 +729,59 @@ def current_nodes(self):
...
+            return reinstall_flocker_at_version(
+                reactor, all_cluster_nodes, control_node_address,
+                package_source, distribution)
```
wallrj: :question: I looked at the name of this function and expected it to require a ``version`` argument.

flocker/provision/_effect.py
```diff
@@ -80,9 +85,48 @@ def reducer(next_effect, effect):
     return reduce(reducer, reversed(effects), results)
 
 
+class HTTPGet(PClass):
+    """
+    Intent for HTTP GET requests.
+
+    :ivar bytes url: The URL to make a GET request to.
+    """
+    url = field(type=bytes)
```
wallrj: :question: ``mandatory`` ?

flocker/provision/_effect.py
```diff
@@ -80,9 +85,48 @@ def reducer(next_effect, effect):
...
+    return Effect(HTTPGet(url=url))
+
+
+@deferred_performer
+def treq_get(dispatcher, intent):
+    """
+    Performer to execute an HTTP GET.
+
+    :param dispatcher: The dispatcher used to dispatch this performance.
+    :param HTTPGet intent: The intent to be performed.
```
wallrj: :question: documenting these gets really tedious.  I think we should agree add this to the eng handbook as an example of something that doesn't need documenting.

flocker/provision/_install.py
```diff
@@ -1532,6 +1581,90 @@ def configure_cluster(
     ])
 
 
+def reinstall_flocker_at_version(
+    reactor, nodes, control_node, package_source, distribution
+):
```
wallrj: :question: Could this be shared with the ManagedRunner which also upgrades Flocker?

flocker/acceptance/endtoend/test_dataset.py
wallrj: Yeah that's my point. The environment variable should either be set or unset. It should never be an empty string.

wallrj:
Thanks @sarum90 

This looks really good. Thanks for persevering!

I made some comments which you can choose to answer, address or defer as you see fit.

I ran ``run-acceptance-tests`` locally with centos-7, Rackspace, + master branch and they all passed.

```
Ran 43 tests in 2354.128s

PASSED (skips=6, successes=37)
--keep specified, not destroying nodes.
To run acceptance tests against these nodes, set the following environment variables: 
export FLOCKER_ACCEPTANCE_HOSTNAME_TO_PUBLIC_ADDRESS='{}';
export FLOCKER_ACCEPTANCE_DISTRIBUTION=centos-7;
export FLOCKER_ACCEPTANCE_DEFAULT_VOLUME_SIZE=80530636800;
export FLOCKER_ACCEPTANCE_CONTROL_NODE=119.9.92.212;
export FLOCKER_ACCEPTANCE_PACKAGE_VERSION='';
export FLOCKER_ACCEPTANCE_VOLUME_BACKEND=openstack;
export FLOCKER_ACCEPTANCE_PACKAGE_BUILD_SERVER=http://build.clusterhq.com/;
export FLOCKER_ACCEPTANCE_TEST_VOLUME_BACKEND_CONFIG=/tmp/tmpcHnCOp/dataset-backend.yml;
export FLOCKER_ACCEPTANCE_NUM_AGENT_NODES=2;
export FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH=/tmp/tmps7Z4_T;
export FLOCKER_ACCEPTANCE_PACKAGE_BRANCH=master;

real	47m39.104s
user	2m2.077s
sys	0m9.357s

```

I might have a go at re-running the tests with the managed runner and try the new setup-cluster script.

But no need to wait for that. 
We can fix those later if they're broken.
Please merge when you've got green builds.

flocker/acceptance/endtoend/test_dataset.py
brendan-donegan: Yuck.. Surprising that list comprehensions are broken like that. FWIW, Python3 behaviour is different (correct)

brendan-donegan: One a side note, using x as a name even in a list comprehension, isn't really good form. You could use 'var' or somesuch here instead of x

wallrj:
* Re-running the tests with `trial` + the printed environment variables worked fine.
* `./admin/run-acceptance-tests --keep --distribution=centos-7 --provider=managed --dataset-backend=rackspace --config-file=$PWD/acceptance.yml --branch=master` worked too.

admin/acceptance.py
sarum90: +1 ... probably should be done in a way that doesn't break people's bash history though. It does sort of assume buildbot was used though, which could definitely be a source of confusion.

admin/acceptance.py
sarum90: Thanks, done.

flocker/acceptance/endtoend/test_dataset.py
sarum90: Changed to 'var', although I'm not sure that I agree in the general case. In small inner loops I'm usually okay with small non-descriptive. It's like the 1 thing I agree with on the linux coding standard. var is short enough though that it doesn't seem worth messing with.

flocker/acceptance/endtoend/test_dataset.py
sarum90: Hrm so I've taken the stance that all of these env vars must be defined in order for the test not to be skipped, and that `None` values from https://github.com/ClusterHQ/flocker/blob/741f880d4171a892c6d559270bb00ce781833150/flocker/provision/_common.py#L16 get turned into empty strings. (see https://github.com/ClusterHQ/flocker/pull/2547/files#diff-d7d14368f6d81514989386a295faa983R120).

Some of this was motivated by https://github.com/ClusterHQ/flocker/blob/741f880d4171a892c6d559270bb00ce781833150/flocker/provision/_install.py#L324 which suggests that empty strings have the same semantics as `None`. So I'd have to write more intelligent code when setting the environment variables to remove them if they are empty strings.

The other part was trying to protect against typos if people were specifying the environment variable by hand. It would be sort of lousy for:

```
export FLOCKER_ACCEPTANCE_PACKAGE_VERSOIN=1.8.0-1-322
```

to result in testing against the default version, when that's probably not the intent. Instead I'd rather error out letting me know that I forgot to specify `FLOCKER_ACCEPTANCE_PACKAGE_VERSION`.

So I'd sort of like to keep those semantics (all three must be set, and empty strings mean use the defaults), but I'll add additional information to the docstrings and enhance the error reporting to make that more obvious.

flocker/acceptance/endtoend/test_dataset.py
sarum90: used HEAD_FLOCKER_VERSION, as this is gotten from git in the current checkout, rather than reading from the version installed on the nodes.

flocker/acceptance/endtoend/test_dataset.py
sarum90: Unfortunately `version` is parsed JSON, so it's more like `dict(flocker=u"1.9.0.dev.asdsadsa")`

And PackageSource requires version to be of type bytes. Switched to explicitly 'encoding' the unicode, and corrected a bug here.

flocker/acceptance/testtools.py
sarum90: Fixed.

Also how would you find:

```
next(iter(actual_dataset_states), None)
```

Which is the new syntax for something like this I recently came across.

flocker/acceptance/testtools.py
sarum90: Yup, thanks for catching, fixed.

flocker/acceptance/testtools.py
sarum90: If the connection hangs indefinitely that will hang indefinitely. If we have a chance to break an indefinitely hanging connection, deferred cancellation is our best chance.

I've seen a few of the benchmarking folks start to use this pattern: for instance, the in-flight: https://github.com/ClusterHQ/flocker/pull/2581/files#diff-5c624e764d7d12f04e516a85cae8ff77R39

Room for disagreement about timing out each retry over timing out the whole set of retries. Not sure I have strong opinions either way. This will for sure finish in 10 seconds, that would for sure finish in timeout*steps seconds. Some data collection about what the failure modes are like would probably be required to make an educated decision.

flocker/acceptance/testtools.py
sarum90: Added comment, stopped using single letter variables.

flocker/acceptance/testtools.py
sarum90: renamed to `reinstall_flocker_from_package_source`

flocker/common/version.py
sarum90: Thanks, yea that certainly was a removed `print()` call.

flocker/provision/_effect.py
sarum90: Thank you, I always forget you have to specify that for some reason.

flocker/provision/_effect.py
sarum90: +1 on it not adding any value... unless it lets us use an automated pylint-style checker to verify all new parameters are documented.

flocker/provision/_install.py
sarum90: So I copied the approach from there, and they use common code for uninstalling and re-installing flocker.

The one difference is the ManagedRunner generates new certificates every time, whereas I just want to restart the flocker agents.

It might make sense for the managed runner to start with a fresh cluster_id so that there aren't old volumes polluting the test, whereas this specifically wants to use the same cluster_id. I think we're hitting a reasonable amount of codeshare, although there definitely would be room for improvement.

sarum90:
Sorry about having to put this back up for review. Discovered 2 things on a recent Jenkins test:

1) Cascading failures were back. The cancellation error that the timeout was causing was being eaten and forgotten about by retry_failure. Composed these two helpers in the other manner so that once again we can continue even if we cannot get the version of flocker before a reinstall.

2) We only support upgrade of persisted control agent state, not downgrade. Thus, at the beginning of the upgrade test where we downgrade to the latest released version of flocker, we must resolve the issue that the current_configuration.json file has state that we can no longer interpret. Worked around the issue by cleaning the cluster right before we downgrade, and simply deleting (or at least hiding) the current_configuration.json when we downgrade.

In order to test these I made a fake configuration change and ran the acceptance tests. Not that the last commit in this change reverts that change so that this does not introduce any actual configuration change.

wallrj:
Thanks @sarum90 

Those latest changes look great.

PS Try the new `./admin/merge_pr https://github.com/ClusterHQ/flocker/pull/2547` tool.

sarum90:
Trying the tool (even though I did wait until all the tests were green anyway :p)

PR #2548, opened 2016-01-29
[FLOC-4004] Add forgotten GCE unit tests.

sarum90:
Found these sitting in one of my git trees. Thought they would have more use if they were checked in.


james-w:
:heavy_check_mark: Thanks, these looks good.

Merge when green.

sarum90:
Yea, these were actually how I was addressing the 'URL parsing is terrible comment'. I was sort of embarrassed when I realized I didn't even check in the proof that it was better now.

PR #2550, opened 2016-01-29
[FLOC-4003] Start CloudFormation Client Node after {Agent, Control} Nodes, and signal successful stack creation

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4003

Start CloudFormation Client Node after cluster (Agent + Control nodes) is up and running {Flocker, Docker, Swarm} components. This enables signaling successful stack creation at the end of Client Node setup.

The PR also fixes a bug in control flow when user opts out of VolumeHub during CloudFormation workflow.


admin/installer/setup_client.sh
```diff
@@ -53,3 +54,6 @@ ${DOCKER_CERT_HOME}/createclient.exp
 
 # Copy Docker Certificates directory to Ubuntu's home directory.
 cp -r ${DOCKER_CERT_HOME} ${UBUNTU_HOME}
+
+# Signal successful stack creation.
+curl ${SUCCESS_SIGNAL_URL}
```
sarum90: No action required: Just documenting review process here.

Richard downloaded all of the logs from `s3://check.clusterhq.com/logs/` and grepped through all of the logs for `/cloudformation-complete.txt` and found some hits. They seemed to be around yesterday midday which seems to be approximately when Richard and Madhuri were adding this code.

This confirms that we have logging enabled on the bucket and we believe it is logging requests that came from runs of the cloudformation template.

This bucket was not being tracked by qloudstat though... I re-configured qloudstat to also consume the logs from this bucket. I expect that to work, and it's more important that we get this checked in so that we log the installs.

sarum90:
Cool! I didn't dive into the details of the volumehub bug, but the fix looks like the sort of fix I would expect (undefined variables and the like).

Commented above about why I believe your new tracking to be working. Looks great! merge at will.

myechuri:
@sarum90 : thanks a lot for the review! Alice found a bug with this change, so, i included the fix in the latest commit. Can i please bother you for a go over of the latest diff? 

The bug is as follows: 75% of installs fail because Client Node fails to come up within 600 seconds. This is a fallout from current change since we now create Client Node after {Control, Agent} Nodes come up, instead of parallel starts. 600 seconds is no longer enough for end to end. A few of Alice and other testers' installs failed with Client Node WaitCondition timeout as a result.

The only diff from previously reviewed code to latest edition is an increase in Client Node WaitCondition timeout from 600 seconds to 900 seconds.

CC @wallrj 

Thanks!

exarkun:
What % of startups complete within 900 seconds?

myechuri:
@exarkun : i was able to reproduce the bug 100% of installs today (2 attempts). Yesterday, 1 out of 1 attempt succeeded (with 600 seconds).  With bump up to 900 seconds, 2 out of 2 installs succeeded today.

sarum90:
Looks fine, just had a thought, we probably don't want to fail the build if curl returns non-zero.

Perhaps we should add ` || echo "Failed to report success to Logs"` to the curl command?

Either way looks good to me as is, feel free to merge once you both (or one of you) has a successful run of the test.

wallrj:
Tests passed:

```
$ AWS_DEFAULT_REGION=us-west-1 VOLUMEHUB_TOKEN='...' KEY_PAIR='...' SECRET_ACCESS_KEY='...' ACCESS_KEY_ID='...' CLOUDFORMATION_TEMPLATE_URL='https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation-richardw.json' trial flocker.acceptance.endtoend.test_installer
  DockerComposeTests
    test_docker_compose_up_postgres ... Warning: Permanently added '52.53.209.247' (ECDSA) to the list of known hosts.
                                   [OK]

Ran 1 tests in 780.779s

PASSED (successes=1)

```

myechuri:
@sarum90 : thanks for the suggestion! Applied and tested.

PR #2552, opened 2016-01-29
[FLOC-2901] Increase maximum attached EBS volumes

jongiddy:
Flocker currently allows only 11 volumes to be attached on a single AWS node.  This is due to Flocker code only allowing the use of the 11 device names `/dev/sdf` to `/dev/sdp`.

AWS documentation indicates that this subset should always be available on every AMI.

However, we need to support more than 11 attached volumes.

This PR adds another 10 device names (`/dev/sdq` to `/dev/sdz`), which should be available on all but "some custom kernels".

The core change is the string that had the letters `f-p` now has `f-z`.  Everything else is refactoring, mostly to make the tests easier to create.

### DIFF
--- a/flocker/node/agents/ebs.py
+++ b/flocker/node/agents/ebs.py
@@ -911,9 +911,7 @@ def _next_device():
     """
     Get the next available EBS device name for this EC2 instance.
 
-    :return Optional[unicode] file_name: available device name for
-        attaching EBS volume, or None if suitable EBS device names on this
-        EC2 instance are currently occupied
+    :return unicode file_name: available device name for attaching EBS volume.
     """
 
     return _select_free_device(_find_allocated_devices())
@@ -965,9 +963,7 @@ def _select_free_device(existing):
 
     :param Sequence[bytes]: List of allocated device basenames
         (e.g. ``[b'sda']``).
-    :return Optional[unicode] file_name: available device name for
-        attaching EBS volume, or None if suitable EBS device names on this
-        EC2 instance are currently occupied
+    :return unicode file_name: available device name for attaching EBS volume.
     """
 
     local_devices = frozenset(existing)
@@ -1266,6 +1262,7 @@ class EBSBlockDeviceAPI(object):
             associate the volume with the expected OS device file.  This
             indicates use on an unsupported OS, a misunderstanding of the EBS
             device assignment rules, or some other bug in this implementation.
+        :raises NoAvailableDevice: If there are no available device names.
         """
         ebs_volume = self._get_ebs_volume(blockdevice_id)
         volume = _blockdevicevolume_from_ebs_volume(ebs_volume)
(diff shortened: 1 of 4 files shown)


flocker/node/agents/ebs.py
```diff
@@ -900,6 +900,91 @@ def _get_blockdevices():
...
+        ]
+        if not any(
+            list(device in local_devices for device in possible_devices)
+        ):
+            return file_name
+
+    # Could not find any suitable device that is available
+    # for attachment. Log to Eliot before giving up.
+    NO_AVAILABLE_DEVICE(devices=sorted_devices).write()
+    return None
```
itamarst: On the principle of "return successfully or not at all", I suggest that this throws an exception rather than returning `None`.

flocker/node/agents/ebs.py
```diff
@@ -900,6 +900,91 @@ def _get_blockdevices():
...
+    """
+
+    local_devices = pset(existing)
+    sorted_devices = sorted(existing)
+    IN_USE_DEVICES(devices=sorted_devices).write()
+
+    for suffix in b"fghijklmonpqrstuvwxyz":
+        next_local_device = b'xvd' + suffix
+        next_local_sd_device = b'sd' + suffix
+        file_name = u'/dev/sd' + unicode(suffix)
```
itamarst: The input is a also a list of bytes, so this should continue to be bytes.

flocker/node/agents/ebs.py
```diff
@@ -900,6 +900,91 @@ def _get_blockdevices():
...
+    the AMI uses paravirtualization or HVM.
+
+    :param Sequence[bytes]: List of allocated device basenames
+        (e.g. ``[b'sda']``).
+    :return Optional[unicode] file_name: available device name for
+        attaching EBS volume, or None if suitable EBS device names on this
+        EC2 instance are currently occupied
+    """
+
+    local_devices = pset(existing)
```
itamarst: :question: The logic below would be simplified by just using `local_devices` and doing `if not possible_devices.intersection(local_devices)`.

flocker/node/agents/test/test_ebs.py
```diff
@@ -246,3 +249,34 @@ def test_no_new_device_logged(self, logger):
...
+class FindAllocatedDeviceTests(TestCase):
+    """
+    Tests for finding allocated devices.
+    """
+
+    def test_returns_device_name_list(self):
+        """
+        Returns at least one device (the root device).
+        """
+        self.assertGreater(len(_find_allocated_devices()), 0)
```
itamarst: And maybe also that it starts with /dev/? Or that they exist and are block devices?

itamarst:
1. All minor comments on pre-existing issues, so address to the extent you wish to.
2. Add news entry since this is user-impacting enhancement.
3. Merge when green.

Thanks!

PR #2553, opened 2016-01-29
[FLOC-3999] Documentation redirects.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3999

### DIFF
--- a/admin/test/test_release.py
+++ b/admin/test/test_release.py
@@ -17,7 +17,12 @@ from unittest import skipUnless, skipIf
 from effect import sync_perform, ComposedDispatcher, base_dispatcher
 from git import Repo
 
+from hypothesis import given
+from hypothesis.strategies import text, sampled_from
+
 from requests.exceptions import HTTPError
+
+from boto.s3.website import RoutingRules, RoutingRule
 
 from twisted.python.filepath import FilePath
 from twisted.python.procutils import which
@@ -27,7 +32,7 @@ from .. import release
 
 from ..release import (
     upload_python_packages, upload_packages, update_repo,
-    publish_docs, Environments,
+    parse_routing_rules, publish_docs, Environments,
     DocumentationRelease, DOCUMENTATION_CONFIGURATIONS, NotTagged, NotARelease,
     calculate_base_branch, create_release_branch,
     CreateReleaseBranchOptions, BranchExists, TagExists,
@@ -42,6 +47,8 @@ from ..aws import FakeAWS, CreateCloudFrontInvalidation
 from ..yum import FakeYum, yum_dispatcher
 
 from flocker.testtools import TestCase
+
+from testtools.matchers import AfterPreprocessing, Equals
 
 FLOCKER_PATH = FilePath(__file__).parent().parent().parent()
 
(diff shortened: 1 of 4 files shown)


admin/test/test_release.py
```diff
@@ -63,13 +70,123 @@ def hard_linking_possible():
...
 
 
+def MatchesRoutingRules(rules):
+    """
+    Matches against routing rules.
+
+    :param rules: The routing rules to match against.
+    :type rules: ``list`` of ``RoutingRule``
+    """
+    return AfterPreprocessing(RoutingRules.to_xml,
```
itamarst: This looks like it'll always be true?

itamarst: Ah, no, I see.

docs/redirects.yaml
```diff
@@ -0,0 +1,35 @@
+"en/latest/":
+    # Redirects in /latest/ introduced for 1.10.0 doc reorganization
+    "introduction/": {replace_key: ""}
+    "introduction/flocker-supports.html": {replace_key: "supported/"}
```
itamarst: So this means `docs.clusterhq.com/en/latest/introduction/flocker-supports.html` gets redirected to `docs.clusterhq.com/en/latest/supported/`? And I see `http://doc-dev.clusterhq.com/supported/` exists and that makes sense. OK.

docs/redirects.yaml
```diff
@@ -0,0 +1,35 @@
...
+    "concepts/": {replace_key: "flocker-features/"}
+    "labs/installer.html": {replace_key: "flocker-standalone/installer.html"}
+    "labs/volumes-cli.html": {replace_key: "flocker-features/flockerctl.html"}
+    "labs/volumes-gui.html": {hostname: "clusterhq.com", replace_key: "volumehub/"}
+    "labs/swarm.html": {replace_key: "docker-integration/"}
+    "labs/compose.html": {replace_key: "docker-integration/"}
+    "labs/mesosphere.html": {replace_key: "mesos-integration/"}
+    "labs/kubernetes.html": {replace_key: "kubernetes-integration/"}
+    "labs/flockercontainers.html": {replace_key: "supported/flockercontainers.html"}
+    "labs/docker-plugin.html": {replace_key: "docker-integration/"}
```
itamarst: docs.clusterhq.com/en/latest/labs/docker-plugin.html is a 404.  Maybe "introduction/flocker-plugin.html" though?

docs/redirects.yaml
```diff
@@ -0,0 +1,35 @@
+"en/latest/":
```
itamarst: So I wonder about this... Currently `docs.clusterhq.com/en/latest/` immediately redirects to `docs.clusterhq.com/en/1.9.0`. That means external people are not going to be linking to it, mostly, but rather to whatever version of docs they randomly found. And new visitors will get warning and that's OK I guess. So is this mostly for internal links from our marketing material/website?

itamarst:
So skimming this it looks vaguely correct, but the proof of the pudding is in the eating. I guess live test as part of release will verify this? So address the one minor actionable comment and merge when you're happy.

docs/redirects.yaml
tomprince: Once 1.10.0 is released, the 1.9.0 will have a link to the corresponding page in latest, which doesn't currently exist, but this introduces a redirect for it.

docs/redirects.yaml
lukemarsden: Yes, but it wasn't previously. We have had error reports from users landing on this page -- the link must be out there somewhere. We should add `introduction/flocker-plugin.html` as a link to there as well though.

docs/redirects.yaml
lukemarsden: @dwgebler's recent change in https://github.com/ClusterHQ/flocker/pull/2503 makes `/en/latest` stop being a redirect, and instead be the canonical location for the docs, fwiw.

docs/redirects.yaml
tomprince: For reference, this existed in 1.4.0.

PR #2554, opened 2016-01-29
[FLOC-3876] Notify slack if a daily jenkins job fails.

james-w:
Jenkins is now configured to be able to notify slack. This adjusts
the job config of the daily jobs to send a notification to

### DIFF
--- a/jobs.groovy
+++ b/jobs.groovy
@@ -262,6 +262,17 @@ def build_wrappers(v) {
 }
 
 
+/*
+    Is the specified branch the master branch?
+
+    :param unicode branchName: the name of the branch.
+    :return bool: whether the branch is the master branch.
+*/
+def is_master(branchName) {
+    return branchName == "master"
+}
+
+
 /* adds a list of triggers to the build job
 
    param  _type: type of job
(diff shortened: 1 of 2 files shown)


jobs.groovy
```diff
@@ -427,6 +438,33 @@ def build_steps(job) {
...
+/*
+    Closure that configures the current project to notify slack on failure.
+    The specified channel will be notified.
+*/
+Closure notify_slack(String channel) {
+    return {
+        // I'm not sure how to put these magic strings in a variable
+        // or avoid them altogether.
+        it / 'properties' << 'jenkins.plugins.slack.SlackNotifier_-SlackJobProperty'(plugin: 'slack@1.8.1') {
+            room(channel)
```
james-w: The fix was here, it needed to use the parameter. It was working before
because it was accidentally referencing a global. When I merged with
changes in master the global was no longer present and it failed.

james-w:
I figured out how to test this as if it were master on my own instance, so
you can see on http://52.35.61.118:8080/job/ClusterHQ-flocker/job/master/job/_run_acceptance_on_Rackspace_CentOS_7_with_Cinder/configure (same creds
as main instance) that it has a "slack notification" post-build action,
and slack config near the top.

I also triggered a build failure to see the notification in #engineering.

jml:
:heavy_check_mark:  thanks.

PR #2555, opened 2016-01-29
[FLOC-4007] Correctly skip the conch tests.

james-w:
When we moved to the testtools base cases the module-level
skip started being ignored, but we didn't notice because the
tests weren't failing until today.


itamarst:
Merge when green. Thanks!

PR #2559, opened 2016-01-29
Enable the pylint redefined-outer-name check.

james-w:


.pylintrc
```diff
@@ -77,13 +77,13 @@ enable=import-error,
        # Has common issues with our style due to
        # https://github.com/PyCQA/pylint/issues/210
        unused-variable,
-       boolean-datetime
+       boolean-datetime,
+       redefined-outer-name,
```
itamarst: Extra comma *might* break pylint.

.pylintrc
```diff
@@ -91,7 +91,6 @@ enable=import-error,
 # 2. See if there's spurious ones; if so disable.
 # 3. Record above.
 # 4. Remove from this list.
-       # deprecated-method,
```
itamarst: Why'd you delete it without enabling it? Local testing suggests it's not too noisy so should be useful.

james-w: Sorry, an extra change because I was playing with other ones, I'll reinstate it.

PR #2560, opened 2016-01-29
[FLOC-4019] Fix a regression in rackspace acceptance tests.

sarum90:
The comma turns this value into a tuple, which makes rackspace fail to be able to set metadata.

We should probably not have such a fragile API with our provisioners.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -572,7 +572,7 @@ class LibcloudRunner(object):
         self.creator = creator
 
         self.metadata.update(self.identity.metadata)
-        self.metadata['distribution'] = self.distribution,
+        self.metadata['distribution'] = self.distribution
 
         # Try to make names unique even if the same creator is starting
         # multiple clusters at the same time.  This lets other code use the


sarum90:
Rackspace can't build an instance right now... so, not sure how to prove this works...

wallrj:
> The comma turns this value into a tuple, which makes rackspace fail to be able to set metadata.

Looks like this was introduced recently in: https://github.com/ClusterHQ/flocker/pull/2540

> We should probably not have such a fragile API with our provisioners.

I guess we could turn the `metadata` into a `pyrsistent.CheckedPMap`

I'll try running `run-acceptance-tests` on rackspace...

wallrj:
Change of plan. @sarum90 is running the acceptance tests on  Rackspace via Jenkins now.

The change LGTM. 

Please merge if the Rackspace tests pass.

sarum90:
re:tests I think rackspace is just having a bad day.

Ultimately my change should not affect the rackspace functional tests, and both the amazon and the rackspace acceptance tests are passing, so I'm going to merge.

PR #2562, opened 2016-02-01
[FLOC-4014] Fixing scenario cleanup so we don't lose the metrics output

mpgomez:
Cleanup deferred that returned nothing, added after the metrics results, was causing the write load scenario not to generate any output. 
The only change needed was to add the metrics generation as the last step.


james-w:
Hi,

The change looks good, but can we get a test to verify
that the results are as we expect?

Thanks,

James

james-w:
:heavy_check_mark: looks good, thanks for adding the test. I think that's all the validation
needed for this change. Validating the result further might be useful for other reasons.

Merge when green.

PR #2564, opened 2016-02-01
[FLOC-3846] Benchmark scenario catch up

jongiddy:
The benchmarking load scenarios work by generating `n` requests per second in a LoopingCall.  LoopingCall can skip iterations, if a single iteration takes longer than the interval between calls.  If that happens, a `count` > 1 is passed to the next iteration.

Current code uses this count to increase the number of seconds passed in the rate measurer, but failed to increase the number of new requests generated - it should be the rate (per second) times the number of seconds passed.

This PR fixes that problem so that the number of requests generated catches up all missed intervals.

Also rename `IRequestScenarioSetup` to the simpler name `IRequest`, since it is mainly encapsulating an arbitrary request.

### DIFF
--- a/benchmark/scenarios/_request_load.py
+++ b/benchmark/scenarios/_request_load.py
@@ -55,7 +55,7 @@ class RequestLoadScenario(object):
 
     :ivar reactor: Reactor to use.
     :ivar scenario_setup: provider of the interface
-        ``IRequestScenarioSetup``.
+        ``IRequest``.
     :ivar request_rate: The target number of requests per second.
     :ivar sample_size: The number of samples to collect when measuring
         the rate.
@@ -85,7 +85,7 @@ class RequestLoadScenario(object):
 
         :param reactor: Reactor to use.
         :param scenario_setup_instance: provider of the
-            ``IRequestScenarioSetup`` interface.
+            ``IRequest`` interface.
         :param request_rate: target number of request per second.
         :param sample_size: number of samples to collect when measuring
             the rate.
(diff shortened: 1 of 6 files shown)


benchmark/scenarios/_request_load.py
```diff
@@ -54,8 +54,7 @@ class RequestLoadScenario(object):
     requests at a specified rate.
 
     :ivar reactor: Reactor to use.
-    :ivar scenario_setup: provider of the interface
-        ``IRequestScenarioSetup``.
+    :ivar scenario_setup: provider of the interface ``IRequest``.
```
zubron: I think we should rename `scenario_setup` and `scenario_setup_instance` to something that implies that they are for requests. In the test code, `request` is used, can we use a name like that here?

benchmark/scenarios/_request_load.py
```diff
@@ -84,8 +83,7 @@ def __init__(
         ``RequestLoadScenario`` constructor.
 
         :param reactor: Reactor to use.
-        :param scenario_setup_instance: provider of the
-            ``IRequestScenarioSetup`` interface.
+        :param scenario_setup_instance: provider of the ``IRequest`` interface.
```
zubron: Other variable name to change.

zubron:
This looks good, thanks. I'm happy for this to be merged once the comments have been addressed. It will also need a review from @jml.

jml:
All good. Thanks @zubron!

PR #2565, opened 2016-02-01
[FLOC-4033] Fix scenarios for big clusters

mpgomez:
Changing the rate measurer to count the current rate based on the sent requests instead of the received responses. That will be more reliable, as the first approach demonstrated to end up in spikes, and we do not really care about that.
Now we just send X requests/second, and keep track of the outstanding requests (sent but not received), and fail if the scenario overloads - that is, fails if we have too many outstanding requests.


benchmark/scenarios/test/test_rate_measurer.py
```diff
@@ -20,10 +20,12 @@ def send_requests(self, rate_measurer, num_requests, num_samples):
         :param num_requests: The number of request we want to send.
         :param num_samples: The number of samples to collect.
         """
-        for i in range(num_samples * num_requests):
-            rate_measurer.request_sent()
+        call_duration = 4.567
+        for i in range(num_samples):
+            for i in range(num_requests):
+                rate_measurer.response_received(call_duration)
```
james-w: This could use the single loop approach the other version did.
(for i in range(num_samples * num_requests))

james-w:
:heavy_check_mark: Looks good. Not sure that my suggestion improves
the code.

Merge when happy.

benchmark/scenarios/_rate_measurer.py
```diff
@@ -61,9 +61,9 @@ def update_rate(self):
         Update the current rate and record a new sample.
         """
         self._rate = (
-            (self._received - self._samples[0]) / float(self.sample_size)
+            (self._sent - self._samples[0]) / float(self.sample_size)
         )
-        self._samples.append(self._received)
+        self._samples.append(self._sent)
```
jongiddy: The code looks good, but it has no way to handle a large number of error responses.  Errors are different to outstanding calls. There is a fixed limit of outstanding calls we should tolerate. Errors on the other hand can be tolerated relative to how many calls we have sent.

I'd suggest changing as follows:
```
    def update_rate(self):
        """
        Update the current rate and record a new sample.
        """
        # Number of calls sent minus number of errors received is our best
        # indicator of "good calls".  Number of calls received has shown to be
        # too lumpy to generate a smooth rate under heavy load.  While errors
        # received might also be lumpy, they are either a small proportion of
        # total calls, or if a large proportion, we should fail anyway.
        good_count = self._sent - self._error_count
        self._rate = (
            (good_count - self._samples[0]) / float(self.sample_size)
        )
        self._samples.append(good_count)
```

This might make the test `test_scenario_throws_exception_when_rate_drops` relevant again :question:

jongiddy:
Looks good. One suggestion for a further change, and then this can be merged.

PR #2566, opened 2016-02-01
[FLOC-3492] Include the underlying error in CloudKeyNotFound.

james-w:
The underlying reason why the key can't be found by flocker
might be important.

In particular our tests sometimes fail with this error and we
don't know why so it will help us if that ever happens again.


jml:
LGTM.

james-w:
Passed all test, including runs of a couple of the acceptance tests (though they didn't
hit this codepath), so I'm going to merge.

PR #2567, opened 2016-02-01
[FLOC-4016] Add a place to hang persistent non-discoverable state.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4016


flocker/control/_model.py
```diff
@@ -620,6 +624,7 @@ class Deployment(PClass):
     """
     nodes = pset_field(Node)
     leases = field(type=Leases, mandatory=True, initial=Leases())
+    persistent_state = field(type=PersistentState, initial=PersistentState())
```
sarum90: Document new ivar.

flocker/control/_model.py
```diff
@@ -609,6 +609,10 @@ def expire(self, now):
         return updated
 
 
+class PersistentState(PClass):
```
sarum90: Document purpose of new class.

flocker/control/_persistence.py
```diff
@@ -143,6 +143,21 @@ def upgrade_from_v2(cls, config):
...
+    def upgrade_from_v3(cls, config):
+        """
+        Migrate a v3 JSON configuration to v4.
+
+        :param bytes config: The v3 JSON data.
+        :return bytes: The v4 JSON data.
+        """
+        decoded_config = loads(config)
+        decoded_config[u"version"] = 4
+        decoded_config[u"deployment"][u"persistent_state"] = {
```
sarum90: :grey_question: confusing to have something named u"persistent_state" in a file named "_persistence" suggests that everything in this file could be under this field.

For instance, why don't leases belong within this structure?

flocker/control/test/configurations/configuration_v4.py
```diff
@@ -0,0 +1,11 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+# Generate a v4 configuration.
+# Hash to recreate: e99b89c0a7c036e9d50f8705871ded0c829d1a83
```
sarum90: :grey_question: Reminder to update this and the above file if you change the name of persistent state.

Almost seems like we should check these tests in after we have merged a commit that adds the new version. That way if we iterate on what "v4" looks like we don't run the risk of forgetting to update the json file.

flocker/control/test/persisted_model.json
```diff
@@ -149,6 +152,10 @@
                     "flocker.control._model.Node"
                 ]
             },
+            "flocker.control._model.PersistentState": {
+                "category": "record",
+                "fields": {}
```
sarum90: :grey_question: Trying to review this without looking too far ahead in your PR chain... does this mean that we should expect fields to be added here, and that for every field that is added we rev the version of the persistence state?

I'm still not sure when I as a developer should add a field here without additional documentation.

flocker/node/_deploy.py
```diff
@@ -62,7 +62,7 @@ class IDeployer(Interface):
     node_uuid = Attribute("")
     hostname = Attribute("")
 
-    def discover_state(cluster_state):
+    def discover_state(cluster_state, persistent_state):
```
sarum90: document new parameter

flocker/node/functional/test_deploy.py
```diff
@@ -51,13 +53,16 @@ def __init__(self, hostname, volume_service, docker_client=None,
...
+    def discover_state(self, cluster_state, persistent_state):
+        d = self.manifestations_deployer.discover_state(
+            cluster_state, persistent_state=persistent_state)
 
         def got_manifestations_state(manifestations_local_state):
             manifestations_state = manifestations_local_state.node_state
             app_discovery = self.applications_deployer.discover_state(
-                DeploymentState(nodes={manifestations_state}))
+                DeploymentState(nodes={manifestations_state}),
+                persistent_state=PersistentState(),
```
sarum90: :grey_question: Should this really be an empty persistent state, or should we instead use a constant in this file? Concerned because if we are going to add fields to PersistentState, I would rather minimize the number of times I had to go around and add a default value for those fields throughout the tests.

flocker/node/test/test_p2p.py
```diff
@@ -171,7 +172,9 @@ def test_uuid(self):
         """
         deployer = self._setup_datasets()
         node_state = self.successResultOf(deployer.discover_state(
-            DeploymentState(nodes={self.EMPTY_NODESTATE}))).node_state
+            DeploymentState(nodes={self.EMPTY_NODESTATE}),
+            persistent_state=PersistentState(),
```
sarum90: :grey_question: Ditto the question about perhaps using a constant in this file for future maintainability.

sarum90:
Seems reasonable and I'm glad we're tackling this bothersome AWS annoyance.

One concern though, this PR adds complexity in some of our core logic, and it's hard to evaluate in isolation. Any chance you could add some comments for the new classes and arguments to show how they will be used? I'd like to review it after those comments are in-place.

flocker/control/_persistence.py
tomprince: `Leases` are configuration, as opposed to state. I left some notes in `_model` as opposed to here.

flocker/node/_deploy.py
tomprince: Perhaps the first argument should be `discovererable_state`? It wouldn't be unreasonable to have it as part of the first argument, except:
- their lifecycle is different: a `DeploymentState` is managed by `ClusterStateService` and updated by `IClusterStateChange` returned from discovery and is purely-ephemeral; `PersistentState` is managed by `ConfigurationPersistenceService` (perhaps now misnamed) and is persisted to disk.
- I want to say something about both objects being immutable and changing at different times, so there isn't an obvious place to create a composite object (other than perhaps here), but if that is the case, two arguments is pretty much equivalent.

flocker/control/_model.py
```diff
@@ -617,9 +628,15 @@ class Deployment(PClass):
     :ivar PSet nodes: A set containing ``Node`` instances
         describing the configuration of each cooperating node.
     :ivar Leases leases: A map of ``Lease`` instances by dataset id.
+    :ivar PersistentState persistent_state: The non-discoverable persistent
+        state of the cluster. (Note: XXX This should idealy be a sibling to the
+        configuration of the cluster, instead of child; but the required
+        refactoring doesn't seem worth it currently, and can be done later if
+        ever).
```
tomprince: Putting this here rather than as a sibling in `Configuration` or in a separate file isn't ideal, but required less changes (i.e. none) to `ConfigPersistenceService` to handle. Making it a sibling would have required adding new `.get` and `.save` methods (and probably renaming the existing ones for consistency). Adding a separate file would have in addition required refactoring the upgrade logic to dealt with two files (and there is the question of whether we might at some point need cross-file upgrades).

Putting it here is a bit of expediency, but I think the only cost is that it is slightly confusing (that is, I don't think it will cause us any development headaches).

flocker/node/functional/test_deploy.py
tomprince: I *think* this should be an empty persistent state. I can't think of reason that we should have a `PersistentState` that isn't default-constructable with a reasonable value. If it turns out to be the case, I think it makes sense to make this change then.

flocker/control/_model.py
tomprince: I've documented the general purpose. The specific purpose for this series of PRs is to have place to record a mapping from `dataset_id`s to `blockdevice_id`s. This is going to supersede the metadata we are currently using to discover this, since the eventual consistency of our backends means we might create duplicates and then randomly pick different volumes depending on the precise results from the backend.

This won't actually prevent duplicates from being created, but it will prevent them from being used.

tomprince:
Thanks review. I've been looking at this code long enough that having a fresh pair of eyes looking at it is useful to see where things aren't clear.

The big distinction that I've tried to clarify is that there is "configuration", "discoverable state", and "non-discovreable (or persistent) state". We previously only had the first two. The third is for when flocker makes a decisions (for example) and needs a record of it, because the underlying system can't consistently report it.

Looking ahead: when we create a blockdevice volume, we record the association of it with the dataset it is for. This is basically a choice the deployer is making: this volume I created is going to be associated with this dataset; there is also some metadata on the volume, but there isn't a way of preventing multiple volumes from being created with that metadata. This way, if the backend doesn't immediately report that volume, we don't continue creating new volumes for that dataset.

flocker/control/_model.py
sarum90: :grey_question: Ah cool, this is making a lot more sense to me, thanks for the new comments, they are great :+1: 

A couple questions here:

- You mention that ideally this would be sibling in configuration, what about if you instead shove `nodes` and `leases` down a layer of the hierarchy (under some new `configured_state` field)? Curious if you think that is more ugly or less ugly than current architecture, and how much work that is. More just exploring the design space here than making a real recommendation.

- Would you mind  adding a note that `leases` is configuration to the ivar docstring? I should have remembered that, but noting it here I think makes the "configure state that needs to be persisted" vs "non-configured persistent state" distinction more clear.

flocker/control/test/persisted_model.json
sarum90: No action required: Much more clear with the comments in `_model`. If I ever make an arbitrary decision that I need to be persisted (and is not configuration), then it goes in this bucket.

sarum90:
Thanks for the comments, makes it much more easy for me to understand the motivation. I probably could have gotten there myself if I had thought long enough, but it's much nicer to just read the documentation since you already thought it through :).

My 1 concern is that we have a fair amount of code that assumes that a `Deployment` is 1:1 with cluster configuration (`git grep Deployment | grep configuration` shows things like:

`flocker/node/functional/test_deploy.py:        desired_configuration = Deployment(nodes=[`

and

`flocker/control/_protocol.py:        :param Deployment configuration: The desired configuration for the`

It seems to me we might be taking on a bit of tech debt with this change. At least at some level it seems like it might be a bit harder to understand how flocker works by grepping through our code. You might see one of the above descriptions and assume that Deployment only consists of flocker-user-supplied-configuration.

However, most of the references are in tests, I haven't done a full audit, but I don't think this breaks any uses of Deployment, and the cleanup is sort of under the realm of the XXX that you added to the code. So... I don't think that should block this PR. At your discretion, open a FLOC-XXX ticket to go along with your XXX so we have a place to add comments and discuss what cleanup if any should occur to offset any confusion between "configuration" and "persistent state" in the documentation around the flocker codebase.

Looking forward to seeing the interface for changing the persistent state from the dataset agents. :smile: Thanks for entertaining my review comments.

sarum90:
Also, merge when test are green.

tomprince:
> It seems to me we might be taking on a bit of tech debt with this change.

Yes. But I think the cost is worth it.

> Thanks for entertaining my review comments.

Thanks for making them, I'm much happier with the documentation than before.

> My 1 concern is that we have a fair amount of code that assumes that a Deployment is 1:1 with cluster configuration [...]

I suspect that
- most of these places also assume that there isn't any non-discoverable state.
- most of these places don't care about any state (I think most state is likely at least somewhat deployer specific, and so anything that cares about state will need to be adjusted to deal with it anyway.

> I probably could have gotten there myself if I had thought long enough, but it's much nicer to just read the documentation since you already thought it through :).

I'm glad you didn't since it gave me the chance to improve the code in response to your confusion.

PR #2568, opened 2016-02-01
[FLOC-4047] Add a blockdevice mapping to the persistent state.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4047

This depends on https://github.com/ClusterHQ/flocker/pull/2567 and https://github.com/ClusterHQ/flocker/pull/2569


flocker/control/_model.py
```diff
@@ -609,6 +609,47 @@ def expire(self, now):
...
+class DatasetAlreadyOwned(Exception):
+    """
+    There is already a blockdevice owned by the given dataset, when trying to
+    record ownership of a differnt blockdevice.
+    """
+
+
+class BlockDeviceOwnership(CheckedPMap):
+    """
+    Persistent mapping of blockdevices to datasets.
```
james-w: I think it maps datasets to blockdevices?

flocker/control/_model.py
```diff
@@ -609,6 +609,47 @@ def expire(self, now):
...
+        auto-generated from name, e.g. flocker-deploy or Docker
+        plugin. That is pre-existing issue, though.
+        """
+        current_blockdevice_id = self.get(blockdevice_id)
+        if current_blockdevice_id not in (None, blockdevice_id):
+            raise DatasetAlreadyOwned()
+        return self.set(dataset_id, blockdevice_id)
+
+
+class PersistentState(PClass):
```
james-w: :eyes: It's not particularly Persistent to store these things in memory. Is there a plan
to persist further? Maybe it's part of a bigger project to make flocker more robust?

flocker/control/_registry.py
```diff
@@ -0,0 +1,60 @@
...
+        Record that blockdevice_id is the relevant one for given dataset_id.
+
+        Once a record is made no other entry can overwrite the existing
+        one; the relationship is hardcoded and permanent. XXX this may
+        interact badly with deletion of dataset where dataset_id is
+        auto-generated from name, e.g. flocker-deploy or Docker
+        plugin. That is pre-existing issue, though.
+
+        XXX having IBlockDeviceAPI specific method is kinda bogus. Some
+        sort of generic method for storing data moving forward?
```
james-w: :question: Is the same XXX being in two places useful?

flocker/control/_registry.py
```diff
@@ -0,0 +1,60 @@
...
+
+from ._model import PersistentState, BlockDeviceOwnership
+
+
+class IStatePersister(Interface):
+    """
+    Interface for updating ``PersistentState``.
+    """
+
+    def record_ownership(dataset_id, blockdevice_id):
```
james-w: Should the interface document the types of the args and return?

flocker/control/_model.py
james-w: Now I've found the reference to ConfigurationPersistenceService, so it
seems like we frequently serialize the configuration to this service,
and that will include this new mapping.

Presumably there's a window of potential data loss with this arrangement,
but that's the design we have chosen.

james-w:
Hi,

:heavy_check_mark: thanks for this, it will be very useful to have these
extra checks.

It all looks good to me, aside from a couple of minor comments and
me not being familiar with how all this works yet. Please go ahead
and merge when you are happy.

flocker/control/_model.py
tomprince: There is a window where the data isn't saved, but there isn't a window where the data is used, but not been saved.

PR #2569, opened 2016-02-01
[FLOC-4017] Add a interface for manipulating persistent state.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4017

~~This depends on https://github.com/ClusterHQ/flocker/pull/2567~~


flocker/control/testtools.py
```diff
@@ -0,0 +1,91 @@
...
+    return IStatePersisterTests
+
+
+def make_loopback_control_client(test_case, clock):
+    """
+    Create a control service and a client connected to it.
+
+    :return: A tuple of a ``ControlAMPService`` and a
+        ``LoopbackAMPClient`` connected to it.
+    """
```
james-w: :question: we don't have this helper somewhere already?

flocker/control/testtools.py
```diff
@@ -0,0 +1,91 @@
...
+    from twisted.python.filepath import FilePath
+    path = FilePath(test_case.mktemp())
```
james-w: This is test_case.make_temporary_path() in our base testcase.

There's also make_temporary_directory() to get a directory that
will already exist.

flocker/control/testtools.py
```diff
@@ -0,0 +1,91 @@
...
+    path.createDirectory()
+    persistence_service = ConfigurationPersistenceService(
+        reactor=clock,
+        path=path,  # path=test_case.make_temporary_directory()
```
james-w: This comment seems redundant. It also suggests you tried make_temporary_directory()
and it didn't work for some reason?

james-w:
Hi,

:heavy_check_mark: Mostly very boring, thanks :smile: Hopefully
this makes the substantial changes in later branches much clearer.

I have a couple of comments/questions about the tests, but overall
please land this when you are happy.

flocker/control/testtools.py
tomprince: This was originally based on a version of the code that didn't have that function. I had planned to switch to it once I merged forward, but forgot. Thanks for catching this.

flocker/control/testtools.py
tomprince: Some but not all of it. I factored out the bit we had in #2594 to reuse here.

PR #2572, opened 2016-02-01
[FLOC-4050] A parser for extracting regional AMI IDs from Packer machine readable output

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4050

The big picture is:
 *  we want to run Packer to build AMI images
 *  record the AMI IDs in JSON format in  an S3 bucket.
 *  run that script from jenkins.
 * feed those AMI IDs into the CloudFormation template builder.

Something like: https://github.com/ClusterHQ/flocker/compare/publish-amis-FLOC-3940

This parser is extracted from that because the branch grew too large.


admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Tests for ``admin.installer``.
```
sarum90: :grey_question: Other packages in flocker I think test internal modules in their own module (suggesting these tests should be moved to 'test/test_images.py').

I think there is some value of keeping your internal tests separate from your external tests. That way someone who uses your API can just look a test_installer.py (your external tests) to see how to use your public APIs, and your internal tests are only seen by people working within the package.

No change required, but something to consider.

admin/installer/__init__.py
```diff
@@ -0,0 +1,5 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Helper utilities for CloudFormation Installer's Packer images.
```
sarum90: :grey_question: Given that this module only has 1 submodule in it right now, it sort of makes sense that the docstring for the module is the same as the submodule. That said, it'd be sort of nice if this was slightly more generic so as to suggest what else might be put in this folder.

No action required, but it might make this more maintainable.

admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
...
+1450420216,amazon-ebs,artifact,0,files-count,0
+1450420216,amazon-ebs,artifact,0,end
+1450420216,,ui,say,--> amazon-ebs: AMIs were created:\n\n\
+us-west-1: ami-e098f380
+""",
+    output={u"us-west-1": u"ami-e098f380"}
+)
+
+PACKER_OUTPUT_US_ALL = ParserData(
+    input=u"""\
```
sarum90: :grey_question: I don't follow the use of backslashes here. Are they needed and the end of some lines, but not others?

admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
...
+1450420216,,ui,say,\n==> Builds finished. \
+The artifacts of successful builds are:
+1450420216,amazon-ebs,artifact-count,1
+1450420216,amazon-ebs,artifact,0,builder-id,mitchellh.amazonebs
+1450420216,amazon-ebs,artifact,0,id,\
+us-east-1:ami-dc4410b6%!(PACKER_COMMA)\
+us-west-1:ami-e098f380%!(PACKER_COMMA)\
+us-west-2:ami-8c8f90ed
+1450420216,amazon-ebs,artifact,0,string,AMIs were created:\n\n\
+us-east-1: ami-dc4410b6\n\
```
sarum90: `\n\` is a really strange pattern. "Add a newline, and then ignore the next newline" perhaps just remove `\n\` if it occurs at the end of a line?

```
>>> c = u"""
... cat\n\
... """
>>> d = u"""
... cat
... """
>>> c == d
True
>>> c
u'\ncat\n'
>>> d
u'\ncat\n'
>>>
```

admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
...
+    _packer_amis, _PackerOutputParser,
+)
+
+
+class ParserData(PClass):
+    input = field(type=unicode, mandatory=True)
+    output = pmap_field(key_type=unicode, value_type=unicode, optional=False)
+
+
+PACKER_OUTPUT_US_WEST_1 = ParserData(
```
sarum90: Would you mind adding a tiny blurb about where this text came from?

Looks like it's just output copied from a run of packer, but it'd be nice to make it explicit.

admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
...
+from flocker.testtools import (
+    TestCase
+)
+
+from ..installer._images import (
+    _packer_amis, _PackerOutputParser,
+)
+
+
+class ParserData(PClass):
```
sarum90: :grey_question: I don't think it adds much, but some people would want a docstring here that describes ParserData and its instance variables.

In this case that seems like more of a maintenance burden.

I'm only commenting just in case you meant to add a docstring and just forgot. Otherwise ignore this comment and move on.

admin/test/test_installer.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_one_ami(self):
+        """
+        If there is a single AMI artifact, the return value is a single item
+        dictionary.
+        """
+        self.assert_packer_amis(PACKER_OUTPUT_US_WEST_1)
+
+    def test_multiple_ami(self):
+        """
```
sarum90: nit: +`If there are multiple AMI artifacts, the return value is a multiple item dictionary`

admin/installer/_images.py
```diff
@@ -0,0 +1,83 @@
...
+        Parse line parts containing information about an artifact.
+
+        :param list parts: The parts of resulting from splitting a comma
+            separated packer output line.
+        """
+        artifact_type = parts[1]
+        if parts[4] == 'end':
+            self._current_artifact['type'] = artifact_type
+            self.artifacts.append(self._current_artifact)
+            self._current_artifact = {}
```
sarum90: nit: also return here, or put the rest in an else block?

Otherwise you are adding an {'end' : []} key to the end of the next _current_artifact.

admin/installer/_images.py
```diff
@@ -0,0 +1,83 @@
...
+    :return: A ``dict`` of the keys and values found.
+    """
+    packer_dict = {}
+    for item in serialized_packer_dict.split("%!(PACKER_COMMA)"):
+        key, value = item.split(":")
+        packer_dict[key] = value
+    return freeze(packer_dict)
+
+
+def _packer_amis(parser):
```
sarum90: Any reason for this not to just be a public method of _PackerOutputParser?

admin/installer/_images.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    :return: A ``dict`` of ``{aws_region: ami_id}`` found in the
+        ``packer_output``.
```
sarum90: not sure what ``packer_output`` is

admin/installer/_images.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    for artifact in parser.artifacts:
```
sarum90: :grey_question: perhaps freeze parser.artifacts to verify that you don't accidentally modify the argument.

alt: make parser.artifacts a PVector of PMaps.

sarum90:
Looks good!

Nice, simple, well-tested PR. Seems like a good building block from something a bit more complicated.

Some considerations throughout. Most are small cleanup things. One concern about your canned inputs. Specifically, I've never run packer, but I can't tell if the \n are supposed to be newlines or literal \n. I think in the current situation python is interpreting them as newlines though, which might not be what you want.

admin/test/test_installer.py
wallrj: The packer output contains long lines containing escaped newlines.
To make pep8 line length checker happy I just inserted end of line escapes and broke the lines up without much thought.
I've now moved the sample output into separate files which is more realistic and easier to maintain.

PR #2574, opened 2016-02-02
[FLOC-4051] A script to build Flocker AMIs and publish their IDs to S3

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4051

This is the script that I propose calling from Jenkins.

By calling it twice in succession, you can create both the Docker base AMI and a Flocker AMI on top of that.
The Docker AMI can be updated as and when new versions of Docker are released (or new versions of the Ubuntu base AMI become available)

The published AMI IDs will then be fed into the cloudformation template builder.

I install the newest Flocker found built by buildbot for a branch, the branch is configured by way of a FLOCKER_BRANCH environment variable when you run ``publish-installer-images``.

I want to be able to store the Flocker version with each AMI ID in the S3 bucket so that the Cloudformation template can be given a Flocker version too which will allow the released documentation to point to a specific CloudFormation template.

This branch is already quite large so I propose dealing with that in a followup issue.


wallrj:
Need to update the documentation too: http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/maintaining-cloudformation.html#building-cloudformation-machine-images

wallrj:
I ran the integration tests manually and they passed:
```
(4051)(publish-ami-ids-FLOC-4051 ✕? =)[~/.../HybridLogic/flocker]$ trial admin.test.test_images.PublishInstallerImagesIntegrationTests
admin.test.test_images
  PublishInstallerImagesIntegrationTests
    test_build_both ...                                                    [OK]
    test_help ...                                                          [OK]

Ran 2 tests in 2413.502s

PASSED (successes=2)
```

admin/installer/_images.py
```diff
@@ -3,7 +3,60 @@
...
+
+from flocker.common.runner import run
+
+
+PACKER_TEMPLATE_DIR = FilePath(__file__).sibling('packer')
+
+# The AWS regions supported by Packer.
+# XXX ap-northeast-2 is not supported by the current packer release:
+# https://github.com/mitchellh/packer/issues/3058
+AWS_REGIONS = (
```
sarum90: :grey_question: It'd be nice if there was a packer command you could run to get all of the regions that it thinks are valid. Looks like it's just a private array in a method though: https://github.com/chulkilee/packer/commit/1aa24035fcd287f7ad00f0366f4617b9230b3ddd#diff-743b70f36793a4f1edd68908b6048ef1L6

So I guess this is really just #moaning . Thanks for leaving a note to the github issue.

admin/installer/_images.py
```diff
@@ -3,7 +3,60 @@
...
+    u"sa-east-1",
+    u"us-east-1",
+    u"us-west-1",
+    u"us-west-2",
+)
+# Hard coded Ubuntu 14.04 base AMI.
+# XXX These AMIs are constantly being updated.
+# This should be looked up dynamically using Canonical's AWS owner ID
+# 099720109477. See:
+# http://askubuntu.com/a/53568
```
sarum90: This sent me to something about ext on windows. I think you may have meant http://askubuntu.com/questions/53582

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
         packer_dict[key] = value
     return freeze(packer_dict)
+
+
+class PackerConfigure(PClass):
+    """
+    The attributes necessary to create a custom packer configuration file from
+    the prototype files in ``admin/installer/packer``.
+
+    :ivar build_region: The AWS region to build images in.
```
sarum90: :question: Could be a `ValueConstant` (and then the above AWS_REGION tuple could be just a ValueConstant declaration). No action required, but might be nice to catch errors earlier.

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    :ivar publish_regions: The AWS regions to publish the build images to.
```
sarum90: :question: ditto ValueConstant

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    :ivar configuration_directory: The directory containing prototype
+        configuration templates.
+    :ivar source_ami: The AMI ID to use as the base image.
+    :ivar working_directory: The directory in which templates will be copied
+        and modified.
+    """
+    build_region = field(type=unicode, mandatory=True)
+    publish_regions = pvector_field(item_type=unicode)
+    template = field(type=unicode, mandatory=True)
+    distribution = field(type=unicode, mandatory=True)
```
sarum90: :question: Looking at it again. All of the above _could_ be ValueConstants rather than unicode ... not sure if it is worth the extra code or not.

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    configuration_directory = field(type=FilePath, initial=PACKER_TEMPLATE_DIR)
+    source_ami = field(type=unicode, mandatory=True)
+    working_directory = field(type=FilePath, mandatory=True)
+
+
+@sync_performer
+def perform_packer_configure(dispatcher, intent):
+    """
+    Copy the prototype configuration files and provisioning scripts to a
+    temporary location and modify one of the configurations witht the values
```
sarum90: s/witht/with/g

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    )
+    template_path = temporary_configuration_directory.child(
+        template_name
+    )
+
+    with template_path.open('r') as infile:
+        configuration = json.load(infile)
+
+    # If a region was specified, just build the image there and don't do
+    # any copying.
```
sarum90: I'm not 100% sure what this comment refers to.

Specifically, it does not look like it is possible not to specify a region?

This comment suggests that there should be some sort of more complicated logic around the ['ami_regions'] code?

I like the codes, I think this comment could probably go away though.

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    # XXX temporarySibling sets alwaysCreate = True for some reason.
+    output_template_path.alwaysCreate = False
+    return output_template_path
+
+
+class PackerBuild(PClass):
+    """
+    The attributes necessary to run ``packer build``.
+
+    # XXX This attribute should be called configuration_path
```
sarum90: Is there a reason not to just call it configuration_path before you merge?

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+    :ivar template: The path to a packer build configuration file.
+    :ivar reactor: The Twisted reactor object used to run the ``packer build``
+        subprocess.
+    :ivar sys_module: A ``sys`` like object with ``stdout`` and ``stderr``
```
sarum90: :question: Interesting thing to think about:

I'm not sure which is better, but my inclination would actually be to put sys_module and reactor into the performer rather than have them come in from the intent.

```
class PackerBuildPerformer(PClass):
   reactor = field(...)
   sys_module = field(...)
  
   def operator __call__(self, dispatcher, intent):
       self.reactor.run(intent.template)....
```

So on and so forth.

I'm nowhere near familiar enough with `Effect` to even evaluate the plusses and minuses of both implementations.

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+        attributes. The ``stderr`` of ``packer build`` will be written to
+        ``sys_module.stderr``.
+    """
+    template = field(type=FilePath)
+    reactor = field(mandatory=True)
+    sys_module = field(initial=sys)
```
sarum90: :question: I think I'd rather have this just be `stderr` rather than sys_module. Then I could pass a StringIO in for testing purposes.

admin/installer/_images.py
```diff
@@ -67,3 +120,269 @@ def _unserialize_packer_dict(serialized_packer_dict):
...
+@deferred_performer
+def perform_packer_build(dispatcher, intent):
+    """
+    Run ``packer build`` using the configuration in the supplied ``intent`` and
+    parse its output.
+
+    :returns: A ``Deferred`` which fires with a dict mapping the ID of the AMI
+        published to each AWS region.
+    """
+    command = ['/opt/packer/packer', 'build',
```
sarum90: :question: I think this clip from my experience on MAC OS X might change how this should look:

```
♦ ~/dev/flocker$ brew install packer
==> Downloading https://homebrew.bintray.com/bottles/packer-0.8.6.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring packer-0.8.6.yosemite.bottle.tar.gz
🍺  /usr/local/Cellar/packer/0.8.6: 40 files, 446M
♦ ~/dev/flocker$ which packer
/usr/local/bin/packer
♦ ~/dev/flocker$
```

Alternatively, have we updated some docs somewhere to inform people they need to have packer installed to run this script?

admin/test/test_images.py
```diff
@@ -116,3 +167,359 @@ def test_multiple_ami(self):
...
+    """
+    Tests for ``PackerConfigure``.
+    """
+    def setUp(self):
+        """
+        Create a temporary working directory.
+        """
+        super(PackerConfigureTests, self).setUp()
+        self.working_directory = FilePath(self.mktemp())
+        self.working_directory.makedirs()
```
sarum90: :question: in `flocker/testtools/_base _MktempMixin` we have some helpers for this, any chance they would work for you?

admin/test/test_images.py
```diff
@@ -116,3 +167,359 @@ def test_multiple_ami(self):
...
+        """
+        template = FilePath(self.mktemp())
+        template.setContent("")
+
+        d = self.perform_packer_build(template)
+
+        d = self.assertFailure(d, ProcessTerminated)
+
+        def check_error(exception):
+            self.assertEqual(1, exception.exitCode)
```
sarum90: :question: perhaps just verify non-zero

admin/test/test_images.py
```diff
@@ -116,3 +167,359 @@ def test_multiple_ami(self):
...
+                        'CalledProcessError {!r}'.format(args),
+                        text_content(str(e))
+                    )
+
+                    if expect_error:
+                        return_code = e.returncode
+                    else:
+                        raise
+
+        return (return_code, stdout_path, stderr_path,)
```
sarum90: :question: we almost have a nice wrapper for this in `flocker/testtools.__init__py`, but it doesn't look like it quite fits your needs.

admin/installer/_images.py
sarum90: I guess it's not a concern if you only ever want to call it from jenkins. Can you just rely on $PATH to have packer? or is there a reason not to do that?

sarum90:
Looks good!

A couple things to consider in the implementation, but the tests look pretty great.

I did not try to verify the skip message was actionable, but it looks like it probably is pretty good at first glance.

admin/installer/_images.py
wallrj: Actually I meant to link directly to the relevant answer: http://askubuntu.com/a/53586

admin/installer/_images.py
wallrj: I originally had it so that the default was to publish to all regions but if you specified a region it would build in that region without any copying. I removed the comment.

admin/installer/_images.py
wallrj: Done. I just forgot to do it.

admin/installer/_images.py
wallrj: Done.

admin/installer/_images.py
wallrj: It helped a lot with the up front validation of build_region, 
I ditched `publish_regions` since I don't think I'll need that yet.
I looked for some existing "supported distributions" constants...there some stuff in admin/packaging  which I can reuse / refactor in a followup.

admin/installer/_images.py
wallrj: A really good suggestion! You're right. I did something ziffect style and put all my "real" performers that need the reactor and sys_module and working_directory in a class.
I tried adding extra arguments directly to the performer function and adding them to the dispatcher with `partial` but that didn't work out for the ``async_performer`` decorator.
See what you think.
It simplified the code a lot.

admin/test/test_images.py
wallrj: Cool. I didn't know about those. Done.

admin/test/test_images.py
wallrj: Yeah, I looked at that but it Eliot logs the stdout and stderr and mixes the two together.

PR #2576, opened 2016-02-02
[FLOC-3967] Add link to appendix from acceptance test doc

davidoliverSP2:
Fixes FLOC 3967

Adds a note in the configuration file section of the Acceptance Test doc, which links to the example config.yml file in the appendix.


tomprince:
A couple of thoughts on this:
- is there any reason that the example is in a separate file, rather than just a section in this file?
- there is already a link to that example in the first paragraph in the changed section (I guess maybe it isn't obvious enough?)

I wonder if the example should also have some words saying that you don't need to include all the sections, just the ones that are relevant to the configuration you want to test?

davidoliverSP2:
Good point. 

Especially as that is the only content on the appendix

davidoliverSP2:
closing, and creating a new branch for this PR

PR #2581, opened 2016-02-02
[FLOC-4010] Slower container creation

jongiddy:
For large clusters, creating and destroying large numbers of datasets causes problems, due to errors like AWS's RequestLimitExceeded.  This is a problem for benchmarking, where we need to quickly create and delete large numbers of datasets.

This PR changes the benchmarking script that creates stateful containers to slow it down.  Rather than sending up a massive cluster change, we start one stateful container per node in sequence, waiting for them to be running, until all are created.

I decided not to fix the cleanup-containers script because it does work - it just behaves badly while cleaning up - e.g. the AWS console will fail to report on status.  Slowing down cleanup means we must hang around staggering calls, when we may not want to.  Also, users are more likely to try this at home (cleaning up lots of datasets quickly) so it's not bad to remind ourselves about this problem.

This problem actually needs to be dealt with at a lower level. The cluster should accept any size of configuration changes, and throttle them as required.  Created FLOC-4059 for that.

The change is actually a fairly hefty rewrite of the `cluster-containers-setup.py` file, so it may be easier to review as a new file rather than as a diff to the existing file.  The other major cause of line changes was extracting some functions to start datasets and containers out of `benchmark/operations/create_container.py` and putting them into a separate file.

The code is also more careful to delete datasets and containers if an error occurs, as orphaned datasets can be a real problem when we're operating near the limits. (Benchmarking requires 20 datasets/node, our AWS code supports a maximum of 21 datasets/node).

An example run:
```
$ ./admin/setup-cluster --distribution ubuntu-14.04 --provider aws --dataset-backend aws --branch master --config-file benchmark.yml --purpose benchmark --number-of-nodes 5
(set environment variables)

$ ./benchmark/setup-cluster-containers --apps-per-node 20 --control-node ${FLOCKER_ACCEPTANCE_CONTROL_NODE} --cert-directory ${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} --wait 180 --image clusterhq/mongodb --mountpoint /data/db
Starting 20 containers per node on 5 nodes...
0 / 100
0 / 100
5 / 100
6 / 100
10 / 100
10 / 100
13 / 100
15 / 100
18 / 100
...
90 / 100
92 / 100
92 / 100
95 / 100
97 / 100
97 / 100
98 / 100
99 / 100
99 / 100
99 / 100
Started 100 containers with 0 failures.

$ ./benchmark/cleanup-cluster --control-node ${FLOCKER_ACCEPTANCE_CONTROL_NODE} --cert-directory ${FLOCKER_ACCEPTANCE_API_CERTIFICATES_PATH} 

```

### DIFF
--- a/benchmark/_flocker.py
+++ b/benchmark/_flocker.py
@@ -1,0 +1,166 @@
+# Copyright 2016 ClusterHQ Inc.  See LICENSE file for details.
+"""
+Utilities to perform Flocker operations.
+"""
+
+from functools import partial
+from datetime import timedelta
+
+from flocker.common import loop_until, timeout as _timeout
+
+DEFAULT_TIMEOUT = timedelta(minutes=10)
+
+
+def loop_until_state_found(reactor, get_states, state_matches, timeout):
+    """
+    Loop until a state has been reached.
+
+    :param get_states: Callable returning a Deferred firing with a list
+        of states.
+    :param state_matches: Callable that accepts a state parameter, and
+        returns a boolean indicating whether the state matches.
+    :param timedelta timeout: Maximum time to wait for state to be found.
+    :return Deferred[Any]: The matching state.
+    """
+    def state_reached():
+        d = get_states()
+
+        def find_match(states):
+            for state in states:
+                if state_matches(state):
+                    return state
+            return None
+        d.addCallback(find_match)
+
+        return d
+
+    d = loop_until(reactor, state_reached)
+    _timeout(reactor, d, timeout.total_seconds())
+    return d
+
+
+def create_dataset(
+    reactor, control_service, node_uuid, dataset_id, volume_size,
+    timeout=DEFAULT_TIMEOUT
+):
+    """
+    Create a dataset, then wait for it to be mounted.
+
+    :param IReactorTime reactor: Twisted Reactor.
+    :param IFlockerAPIV1Client control_service: Benchmark control
+        service.
+    :param UUID node_uuid: Node on which to create dataset.
+    :param UUID dataset_id: ID for created dataset.
+    :param int volume_size: Size of volume in bytes.
+    :param timedelta timeout: Maximum time to wait for dataset to be
+        mounted.
+    :return Deferred[DatasetState]: The state of the created dataset.
+    """
+
+    d = control_service.create_dataset(
+        primary=node_uuid,
+        maximum_size=volume_size,
+        dataset_id=dataset_id,
+    )
+
+    def dataset_matches(dataset, state):
+        return (
+            state.dataset_id == dataset.dataset_id and
+            state.primary == dataset.primary and
+            state.path is not None
+        )
+
+    d.addCallback(
+        lambda dataset: loop_until_state_found(
+            reactor, control_service.list_datasets_state,
+            partial(dataset_matches, dataset), timeout
+        )
+    )
+
+    return d
+
+
+def create_container(
+    reactor, control_service, node_uuid, name, image, volumes=None,
+    timeout=DEFAULT_TIMEOUT
+):
+    """
+    Create a container, then wait for it to be running.
+
+    :param IReactorTime reactor: Twisted Reactor.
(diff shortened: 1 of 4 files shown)


benchmark/_flocker.py
```diff
@@ -0,0 +1,175 @@
...
+            for state in states:
+                if state_matches(state):
+                    return state
+            return None
+        d.addCallback(find_match)
+
+        return d
+
+    d = loop_until(reactor, state_reached)
+    _timeout(reactor, d, timeout.total_seconds())
```
sarum90: :grey_question: No action required... Now I wish I had remembered to use a timedelta when I originally wrote that timeout function :cry: 

As far as you are concerned :+1: on using the timedelta for this interface.

benchmark/_flocker.py
```diff
@@ -0,0 +1,175 @@
...
+from flocker.common import loop_until, timeout as _timeout
+
+DEFAULT_TIMEOUT = timedelta(minutes=10)
+
+
+def loop_until_state_found(reactor, get_states, state_matches, timeout):
+    """
+    Loop until a state has been reached.
+
+    :param get_states: Callable returning a Deferred firing with a list
```
sarum90: document reactor parameter

benchmark/_flocker.py
jongiddy: Thanks. Also would have been useful to call it `deferred_timeout` or similar to avoid the common name clash with `timeout` for the value.

Maybe we could add a timedelta-based version with a longer name.

benchmark/_flocker.py
```diff
@@ -0,0 +1,176 @@
...
+                    return state
+            return None
+        d.addCallback(find_match)
+
+        return d
+
+    d = loop_until(reactor, state_reached)
+    _timeout(reactor, d, timeout.total_seconds())
+    return d
+
```
mpgomez: Should we have two blank lines between functions or just one?

benchmark/cluster_containers_setup.py
```diff
@@ -299,95 +243,61 @@ def _dataset_to_volume(self, dataset):
...
+                    control_service=self.client,
+                    node_uuid=node.uuid,
+                    name=unicode(uuid4()),
+                    image=self.image,
+                    volumes=[volume],
+                    timeout=self.timeout)
 
-        d.addCallback(do_we_have_enough_datasets)
-        return d
+                # If container creation fails, delete dataset as well
```
mpgomez: Nice one!

benchmark/cluster_containers_setup.py
```diff
@@ -395,49 +305,51 @@ def create_datasets_and_containers(self):
...
-
-                ).format(
-                    num_dataset=i+1,
-                    node_uuid=node.uuid,
-                )
-                Message.log(action=msg)
+        d = self.client.list_nodes()
+
+        def start_containers(nodes):
+            print('Starting {} containers per node on {} nodes...'.format(
```
mpgomez: We have now a mix of eliot and prints. Shouldn't we use eliot everywhere?

benchmark/cluster_containers_setup.py
```diff
@@ -48,12 +52,9 @@ class ContainerOptions(usage.Options):
          'Location of the user and control certificates and user key'],
         ['max-size', None, 1,
          'Size of the volume, in gigabytes. One GB by default'],
-        ['wait', None, 7200,
+        ['wait', None, 3600,
```
mpgomez: It may be useful to have those values in a variable instead of hardcoded?

mpgomez:
LGTM! I added comments for the few thoughts I had while reviewing it.
It looks way nicer now. Thanks Jon!

benchmark/_flocker.py
jongiddy: @mpgomez The definitive answers for blank lines are in https://www.python.org/dev/peps/pep-0008/#blank-lines

PR #2582, opened 2016-02-02
[FLOC-4061] Rename titles on Integration homepage

davidoliverSP2:
Fixes 4061

The current titles for Tutorials and Blog Articles did not test well in User Testing - this issue reverts the titles to something simple, adding an internal note that they will need to be changed when further tutorials are added.


tomprince:
This seems clear enough.

PR #2583, opened 2016-02-02
[FLOC-4066] Move leases documentation to Flocker Features section

davidoliverSP2:
Fixes FLOC 4066

Leases documentation was appearing under reference section - this didn't test well in User Testing, where most users expected it to be found under Flocker Features.


tomprince:
I'm curious about
- the ordering of this in the features section (if it belongs here, I think it should be later)
- whether we expect end-users to be using leases.

As I understand it, leases are designed to be used be integrations (such as the docker plugin) to prevent volumes being moved under them. This suggests to me that it should be under a section aimed at integration authors, rather than a section aimed at end-users, which the rest of the features section seems targeted at.

tomprince:
I'm vaguely curious as to what kind of user-testing indicated this should be moved? If it was asking people to find information on leases, given a link to the documentation, then I'm not sure how relevant that is, since it isn't clear that somebody is going to go looking for that information out-of-context. That is, leases, in and of themselves, don't seem to be a feature that users would be directly interested in, but rather, a tool that enables the user to do other things.

davidoliverSP2:
Thanks @tomprince 

I'm going to move this feedback into the Jira issue, and close this PR. I think there are several pieces of small restructuring required to the docs, and we would benefit from a re-design to make sure we're not regressing the docs.

PR #2584, opened 2016-02-02
Revert "[FLOC-3998] Flocker dataset config"

dwgebler:
Reverts ClusterHQ/flocker#2542


exarkun:
Thanks.  Looks like the right thing.  Please merge when jenkins builds are green.

PR #2585, opened 2016-02-02
[FLOC 4069] Move the flockerctl doc to the Reference section

davidoliverSP2:
Fixes 4069

User Testing results indicated the flockerctl documentation is better located alongside the API docs in the reference section.

This issue moves the document


tomprince:
Having just reviewed #2583 which moves some stuff from reference to features and now seeing this moving something the other way, I'm not sure I understand at all what "Features" is supposed to cover. I guess I don't know enough to properly review either of these branches.

wallrj:
I think this is probably an improvement -- "features" definitely seems like the wrong place for the CLI reference docs. 

"Reference" seems like a better place, but not what I'd call obvious.

Merge if you want this change urgently.

Otherwise, consider the following hierarchy:

 * Flocker Administration
   * Managing cluster nodes and certificates
   * Managing volumes
     * ...examples of using ``flockerctl`` to view the status of volumes and to delete volumes.
   * Upgrading to a new version of Flocker
   * Common problems / Diagnostics and debugging
 * REST API
   * API Reference
   * Leases 
 * Command Line Interface (CLI)
   * ``flockerctl`` 
   * ``flocker-ca``
   * ``flocker-diagnostics``

And finally:
 * I find this paragraph  in "Install & Configure" really confusing. 

```
flockerctl is packaged along with the Labs Installer. You don't have to use the labs installer in order to use flockerctl, though. You can use flockerctl with Flocker installed via any installation mechanism.
```

We shouldn't distract users with talk of the "Labs installer" here.

* I also find this paragraph confusing:

```
flockerctl is a CLI for controlling the Flocker Control Service. It is specific to datasets, and should not be confused with flocker-deploy (which is deprecated) or flocker-ca (which can be used as part of the manual initial configuration of a cluster's certificates).
```

If this is really a source of confusion, we should change the name of the tool to `flocker-volume-cli` 
There's no need to even mention `flocker-deploy` anymore.
And ``flocker-ca`` should be documented separately.

That's all for now.

wallrj:
I notice that Docker have a "Command and API Reference" section.
 * https://docs.docker.com/

davidoliverSP2:
Thanks @wallrj 

I'm going to close this unmerged, and I've opened some new Jira issues to tackle this a little differently.

PR #2587, opened 2016-02-02
[FLOC-3980] Release Flocker 1.10

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-3980


itamarst:
I'm happy with this, only concern is that this will presumably need to be 1.10.1 since tag already exists for 1.10.0.

PR #2588, opened 2016-02-02
[FLOC-3810] Acceptance jobs junit

brendan-donegan:
Updates Jenkins jobs to add test results for acceptance jobs. This will allow us to see which tests fail rather than being restricted to a binary output


mpgomez:
LGTM! Thanks!

PR #2591, opened 2016-02-03
[FLOC-3998] Flocker dataset config

dwgebler:
Fixes FLOC-3998, including a default value for `required_config` to prevent existing plugins breaking.


james-w:
:heavy_check_mark: great, thanks for improving this. Merge when ready.

PR #2595, opened 2016-02-03
[FLOC-4085] Tighten constraints on CloudFormation user Inputs

myechuri:
CloudFormation user can leave {AmazonAccessKeyID, AmazonSecretAccessKey, EC2KeyPair} Inputs empty, leading to failure in stack creation, and subsequent trouble in triaging cause of failure (due to stack rollback being default action).

This PR attempts to impose stricter constraints around user Inputs. The constraints are gathered from Amazon documentation.

Master copy of CloudFormation template has been updated with the changes in this PR in order to facilitate testing.

Testing:

* Manual tests passed

* Endtoend test passed:

```
ubuntu@ip-172-31-25-23:~/flocker/admin/installer$ KEEP_STACK=1 CLOUDFORMATION_TEMPLATE_URL=https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json KEY_PAIR=$KEY ACCESS_KEY_ID=$ACCESS_KEY_ID SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY VOLUMEHUB_TOKEN='' trial flocker.acceptance.endtoend.test_installer
  DockerComposeTests
    test_docker_compose_up_postgres ... Warning: Permanently added '54.153.118.31' (ECDSA) to the list of known hosts.
                                   [OK]

Ran 1 tests in 751.492s

PASSED (successes=1)
ubuntu@ip-172-31-25-23:~/flocker/admin/installer$ 
```

### DIFF
--- a/admin/installer/cloudformation.py
+++ b/admin/installer/cloudformation.py
@@ -70,22 +70,29 @@ template = Template()
 # Keys corresponding to CloudFormation user Inputs.
 access_key_id_param = template.add_parameter(Parameter(
     "AmazonAccessKeyID",
-    Description="Your Amazon AWS access key ID",
+    Description="Your Amazon AWS access key ID (mandatory)",
     Type="String",
     NoEcho=True,
     AllowedPattern="[\w]+",
+    MinLength="16",
+    MaxLength="32",
 ))
 secret_access_key_param = template.add_parameter(Parameter(
     "AmazonSecretAccessKey",
-    Description="Your Amazon AWS secret access key.",
+    Description="Your Amazon AWS secret access key (mandatory)",
     Type="String",
     NoEcho=True,
+    MinLength="1",
 ))
 keyname_param = template.add_parameter(Parameter(
     "EC2KeyPair",
     Description="Name of an existing EC2 KeyPair to enable SSH "
-                "access to the instance",
+                "access to the instance (mandatory)",
     Type="String",
+    MinLength="1",
+    AllowedPattern="[\x20-\x7E]*",
+    MaxLength="255",
+    ConstraintDescription="can contain only ASCII characters.",
 ))
 volumehub_token = template.add_parameter(Parameter(
     "VolumeHubToken",
--- a/admin/installer/cloudformation.py
+++ b/admin/installer/cloudformation.py
@@ -72,11 +72,14 @@ access_key_id_param = template.add_parameter(Parameter(
     "AmazonAccessKeyID",
     Description="Your Amazon AWS access key ID",
     Type="String",
+    NoEcho=True,
+    AllowedPattern="[\w]+",
 ))
 secret_access_key_param = template.add_parameter(Parameter(
     "AmazonSecretAccessKey",
     Description="Your Amazon AWS secret access key.",
     Type="String",
+    NoEcho=True,
 ))
 keyname_param = template.add_parameter(Parameter(
     "EC2KeyPair",


admin/installer/cloudformation.py
```diff
@@ -70,19 +70,29 @@ def _sibling_lines(filename):
...
     Type="String",
+    NoEcho=True,
+    MinLength="1",
 ))
 keyname_param = template.add_parameter(Parameter(
     "EC2KeyPair",
     Description="Name of an existing EC2 KeyPair to enable SSH "
-                "access to the instance",
+                "access to the instance (mandatory)",
     Type="String",
```
wallrj: There's a special type for key pair name:
 * `AWS::EC2::KeyPair::KeyName`: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html
 * https://blogs.aws.amazon.com/application-management/post/Tx3DV2UYG9SC38G/Using-the-New-CloudFormation-Parameter-Types

Worth trying that?

admin/installer/cloudformation.py
```diff
@@ -70,19 +70,29 @@ def _sibling_lines(filename):
 # Keys corresponding to CloudFormation user Inputs.
 access_key_id_param = template.add_parameter(Parameter(
     "AmazonAccessKeyID",
-    Description="Your Amazon AWS access key ID",
+    Description="Your Amazon AWS access key ID (mandatory)",
```
wallrj: This is better, but I can still leave the field empty and click next.
It seems unbelievable that you can't just flag a Parameter as required....but the documentation suggests it's not possible.
Perhaps we can create a "Condition" which checks that the key and secret key are not empty before continuing with the stack creation?
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html

admin/installer/cloudformation.py
myechuri: Thanks for suggesting, @wallrj . i had tried that, but there is no way to express mandatory constraint except setting MinLength to 1. So, decided to not use it.

admin/installer/cloudformation.py
myechuri: > This is better, but I can still leave the field empty and click next.

Validation is performed after your click ``Create``.

wallrj:
Thanks @myechuri 

The validation patterns don't seem to work for me.

Looking at the JSON template it looks like the regexp we supply to Troposphere is being mangled:

```
    "Parameters": {
        "AmazonAccessKeyID": {
            "AllowedPattern": "[\\w]+",
            "Description": "Your Amazon AWS access key ID (mandatory)",
            "MaxLength": "32",
            "MinLength": "16",
            "NoEcho": true,
            "Type": "String"
        },
        "AmazonSecretAccessKey": {
            "Description": "Your Amazon AWS secret access key (mandatory)",
            "MinLength": "1",
            "NoEcho": true,
            "Type": "String"
        },
        "EC2KeyPair": {
            "AllowedPattern": "[ -~]*",
            "ConstraintDescription": "can contain only ASCII characters.",
            "Description": "Name of an existing EC2 KeyPair to enable SSH access to the instance (mandatory)",
            "MaxLength": "255",
            "MinLength": "1",
            "Type": "String"
        },
        "VolumeHubToken": {
            "Default": "",
            "Description": "Your Volume Hub token (optional). You'll find the token at https://volumehub.clusterhq.com/v1/token.",
            "Type": "String"
        }
    },
```

the min and max length constraints don't seem to work either. I was able to enter a single character in each field and proceed to the stack creation.

I made one or two other suggestions above.

Let's see if we can make this better before merging.

wallrj:
Ah, the Splunk Cloudformation template work around the lack of required parameters by specifying a default for each of those fields:
 * https://github.com/splunk/splunk-aws-cloudformation/blob/master/templates/splunk_cluster.template

So perhaps we could say :+1: 
```
secret_access_key_param = template.add_parameter(Parameter(
    "AmazonSecretAccessKey",
    Description="Your Amazon AWS secret access key (mandatory)",
    Type="String",
    NoEcho=True,
    MinLength="1",

))
```

admin/installer/cloudformation.py
wallrj: Ah! Sorry. I see that now. 
Let me try this again...

admin/installer/cloudformation.py
```diff
@@ -70,19 +70,29 @@ def _sibling_lines(filename):
...
     Type="String",
+    NoEcho=True,
+    AllowedPattern="[\w]+",
```
wallrj: :question: Maybe add a ConstraintDescription here too saying "can contain only alphanumeric characters and  underscores" (at least that's how the Python RE docs describe `\w`) 
 * https://docs.python.org/2/library/re.html#regular-expression-syntax

We could also link to https://docs.aws.amazon.com/IAM/latest/APIReference/API_AccessKey.html in the constraint description.

admin/installer/cloudformation.py
wallrj: :question: Perhaps we could add an *invalid* default value? It seems like it might be worth giving it another try since the validation should quickly report whether the named key is available in this region and I think @robhaswell  and someone else already stumbled across that problem.

admin/installer/cloudformation.py
```diff
@@ -70,19 +70,29 @@ def _sibling_lines(filename):
...
+    MinLength="1",
 ))
 keyname_param = template.add_parameter(Parameter(
     "EC2KeyPair",
     Description="Name of an existing EC2 KeyPair to enable SSH "
-                "access to the instance",
+                "access to the instance (mandatory)",
     Type="String",
+    MinLength="1",
+    AllowedPattern="[\x20-\x7E]*",
```
wallrj: :question: This is rendered as `"AllowedPattern": "[ -~]*"` in the JSON file, but I'm surprised to find that that range works in Python too. I thought character ranges had to be `a-zA-Z0-9` ....but the validation does seem to work, so fine.

wallrj:
@myechuri I tried it again after logging out and back in again.

As you said, the validation takes place when you click create and the patterns do seem to catch empty and invalid fields.

So merge when you've answered or addressed my latest review comments above.

Thanks.

admin/installer/cloudformation.py
myechuri: Fixed.

admin/installer/cloudformation.py
myechuri: Thanks for trying out the template after logging out and back in, @wallrj !

admin/installer/cloudformation.py
myechuri: Output: 

```
Error
Template validation error: Parameter AmazonAccessKeyID failed to satisfy constraint: can only contain [A-Za-z0-9_]. Reference: https://docs.aws.amazon.com/IAM/latest/APIReference/API_AccessKey.html
```

myechuri:
Endtoend test with the latest commit:

```
ubuntu@ip-172-31-25-23:~/flocker/admin/installer$ KEEP_STACK=1 CLOUDFORMATION_TEMPLATE_URL=https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json KEY_PAIR=$KEY ACCESS_KEY_ID=$ACCESS_KEY_ID SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY VOLUMEHUB_TOKEN='' trial flocker.acceptance.endtoend.test_installer
  DockerComposeTests
    test_docker_compose_up_postgres ... Warning: Permanently added '54.183.182.190' (ECDSA) to the list of known hosts.
                                   [OK]

Ran 1 tests in 670.570s

PASSED (successes=1)
ubuntu@ip-172-31-25-23:~/flocker/admin/installer$
```

PR #2598, opened 2016-02-04
[FLOC-3976 ] support for batch cloud provisioning

avg-I:
`IProvisioner.create_nodes()` is a new interface that allows to request multiple nodes at once.
The method returns a list of Deferreds with one Deferred for each requested node.
The implementations request multiple nodes from a platform at once if the platform permits (AWS)
or one by one if not (Rackspace). But in either case the waiting for nodes to become running and accessible is done in parallel and asynchronously.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -25,7 +25,6 @@ from twisted.python.usage import Options, UsageError
 from twisted.python.filepath import FilePath
 from twisted.python.failure import Failure
 from twisted.internet.defer import inlineCallbacks, returnValue, succeed
-from twisted.internet.threads import deferToThreadPool
 from twisted.python.reflect import prefixedMethodNames
 
 from effect import parallel
@@ -636,12 +635,15 @@ class LibcloudRunner(object):
         :return: Deferred that fires with the created node when
             the provisioning is completed.
         """
+        print "re-creating and re-trying node", name
+
         def create_attempt():
-            d = deferToThreadPool(
-                reactor,
-                reactor.getThreadPool(),
-                self._create_node,
-                name,
+            # A list with a single Deferred must be returned.
+            [d] = self.provisioner.create_nodes(
+                reactor=reactor,
+                names=[name],
+                distribution=self.distribution,
+                metadata=self.metadata,
             )
 
             def provision(node):
@@ -671,6 +673,37 @@ class LibcloudRunner(object):
         d.addErrback(error)
         return d
 
+    def _provision_with_retry(self, reactor, name, result):
+        def provision(node):
+            print "created node", name
+            print "provisioning..."
+            self.nodes.append(node)
+            d = self._provision_node(reactor, node)
+
+            def provisioning_failed(failure):
+                print "provisioning failed", name
+                # Destroy a node if we failed to provision it.
+                self._destroy_node(node)
+                return failure
+
+            # Make sure to return the node.
+            d.addCallbacks(lambda _: node, errback=provisioning_failed)
+            return d
+
+        result.addCallback(provision)
+
+        # If the initial bulk creation failed for this node for whatever
+        # reason or if its provisioning failed, then beging the process
+        # for this node from the very start.
+        def retry_on_error(failure):
+            print "bulk creation failed", name
+            print "error:", failure
+            write_failure(failure)
+            return self._create_and_provision(reactor, name)
+
+        result.addErrback(retry_on_error)
+        return result
+
     def _create_nodes(self, reactor, names):
         """
         Create and provision a number of nodes with the given names.
@@ -681,10 +714,15 @@ class LibcloudRunner(object):
         :return: List of Deferreds each firing when the corresponding node
             is created and provisioned.
         """
-        reactor.suggestThreadPoolSize(len(names))
-        return [
-            self._create_and_provision(reactor, name) for name in names
-        ]
+        results = self.provisioner.create_nodes(
+            reactor=reactor,
+            names=names,
+            distribution=self.distribution,
+            metadata=self.metadata,
+        )
+        for name, result in zip(names, results):
+            self._provision_with_retry(reactor, name, result)
+        return results
 
     def _make_node_name(self, tag, index):
         """
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -674,6 +674,22 @@ class LibcloudRunner(object):
         return d
 
     def _provision_with_retry(self, reactor, name, result):
+        """
+        Wait for a node to get created and the provision it.
+        Re-create the node and retry the provisioning if either the node
+        creation or provisioning fails.
+
+        This method is used in conjuction with bulk creation of new nodes,
+        so that a failure for any single node can be retried individually
+        (rather than failing all the nodes and starting from scratch).
+
+        :param reactor: The reactor.
+        :param str name: The name of the node.
+        :param Deferred result: Deferred that fires with the node when it's
+            created.
+        :return: Deferred that fires with the node when it's created and
+            provisioned.
+        """
         def provision(node):
             print "created node", name
             print "provisioning..."
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -638,7 +638,8 @@ class LibcloudRunner(object):
         print "re-creating and re-trying node", name
 
         def create_attempt():
-            # A list with a single Deferred must be returned.
+            # A list with a single Deferred must be returned as we are
+            # requesting a single node.
             [d] = self.provisioner.create_nodes(
                 reactor=reactor,
                 names=[name],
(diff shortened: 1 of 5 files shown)


admin/acceptance.py
```diff
@@ -636,12 +635,15 @@ def _create_and_provision(self, reactor, name, retries=10):
...
         """
+        print "re-creating and re-trying node", name
+
         def create_attempt():
-            d = deferToThreadPool(
-                reactor,
-                reactor.getThreadPool(),
-                self._create_node,
-                name,
+            # A list with a single Deferred must be returned.
```
brendan-donegan: I'm not sure this comment is useful unless it says *why* this is the case.

avg-I: I assumed it to be sort of self-documented in the code as `create_nodes` is called with a single node name.

admin/acceptance.py
```diff
@@ -636,12 +635,15 @@ def _create_and_provision(self, reactor, name, retries=10):
         :return: Deferred that fires with the created node when
             the provisioning is completed.
         """
+        print "re-creating and re-trying node", name
```
mpgomez: We could use eliot instead of print?

admin/acceptance.py
mpgomez: Yes, it is kind of self-documented, but I would find it useful to have some extra info about why (ie, because it is called with a name list with only one name on it). I guess if you are familiar with the code it's perfectly self-documented, not that obvious if you have never used cloud provisioner

admin/acceptance.py
```diff
@@ -671,6 +673,37 @@ def error(failure):
...
+            d.addCallbacks(lambda _: node, errback=provisioning_failed)
+            return d
+
+        result.addCallback(provision)
+
+        # If the initial bulk creation failed for this node for whatever
+        # reason or if its provisioning failed, then beging the process
+        # for this node from the very start.
+        def retry_on_error(failure):
+            print "bulk creation failed", name
```
mpgomez: Same as above. Eliot is already being used for some outputs in this file, and we use it as well in benchmark code, it may be a good thing to be consistent with what we use - either print or Eliot, but using both doesn't look very consistent.

admin/acceptance.py
```diff
@@ -671,6 +673,37 @@ def error(failure):
         d.addErrback(error)
         return d
 
+    def _provision_with_retry(self, reactor, name, result):
```
mpgomez: Maybe we could have a docstring for this function?

flocker/provision/_aws.py
```diff
@@ -130,6 +132,51 @@ def _wait_until_running(instance):
         raise FailedToRun(instance.state_reason)
 
 
+def _async_wait_until_running(reactor, instance):
+    """
+    Wait until a instance is running.
+
+    :param reactor: The reactor.
+    :param boto.ec2.instance.Instance instance: The instance to wait for.
+    :return: Deferred that fires when the instance has become running
```
mpgomez: Please, put the ":" after Deferred :)

flocker/provision/_aws.py
```diff
@@ -314,6 +376,98 @@ def _set_metadata(self, instance, metadata):
             )
         return False
 
+    def create_nodes(self, reactor, names, distribution, metadata={}):
+        size = self._default_size
```
mpgomez: Maybe some docstrings could be added here?

flocker/provision/_libcloud.py
```diff
@@ -341,6 +344,270 @@ def _cleanup_node_named(self, name):
...
+            )
+
+            def failed_to_run(failure):
+                d = _retry_exception_async(reactor, node.destroy)
+                d.addCallback(lambda _: failure)
+                return d
+
+            d.addErrback(failed_to_run)
+            return d.addActionFinish()
+
```
mpgomez: It seems that there is an extra blank line

flocker/provision/_libcloud.py
```diff
@@ -371,3 +638,66 @@ def _retry_exception(f, steps=(0.1,) * 10, sleep=sleep):
...
+    with action.context():
+        d = loop_until(reactor, make_call, steps)
+        d = DeferredContext(d)
+        d.addCallbacks(
+            lambda _: saved_result[0],
+            errback=lambda _: saved_failure[0],
+        )
+        return d.addActionFinish()
+
+
```
mpgomez: And an extra blank line here as well

mpgomez:
LGTM! Thanks Andriy for making our lives so much easiier with all the setup cluster tools! 
I have added a few comments, please address any you think you need to address.

flocker/provision/_libcloud.py
```diff
@@ -341,6 +344,270 @@ def _cleanup_node_named(self, name):
...
+        d.addCallback(check_state)
+        return d
+
+    def _wait_until_running(self, reactor, node):
+        """
+        Wait until the node is running and its network interface is configured.
+
+        This method fails if the node does not reach the expected state until
+        the predefined timeout expires or if the node goes into an error state
```
james-w: I'm not sure what the predefined timeout is. This method seems to
define the steps itself.

flocker/provision/_libcloud.py
```diff
@@ -341,6 +344,270 @@ def _cleanup_node_named(self, name):
...
+        """
+        Determine whether a compute instance is in one of the given states.
+
+        :param target_node: A libcloud node object representing the compute
+            instance to examine.
+        :param set target_states: The states to compare the compute instance's
+            actual state against.
+
+        :return: ``True`` if the compute instance is currently in one of
+            ``target_states``, ``False`` otherwise.
```
james-w: It actually returns a deferred that fires with that result.

admin/acceptance.py
avg-I: That makes sense but `print`s are now pervasive in that code. I've created an issue: https://clusterhq.atlassian.net/browse/FLOC-4096

admin/acceptance.py
avg-I: Okay.

admin/acceptance.py
avg-I: As a matter of fact, eliot logging is "almost unused" in this file. There are just a few instances of `write_failure`.

admin/acceptance.py
avg-I: Done.

james-w:
Hi, Thanks for this change, I think it's a really good idea. I added a couple of
minor comments of my own.

I have a couple of major questions as well:

  1) It doesn't seem like this changes the single-creation methods of the provisioners to use these new methods. Can't we do that so we don't have two paths to create a node?

  2) There aren't any direct tests of the new methods. I think this just continues the current standard around this code though, so not a blocker.

So I would suggest:

1) small cleanups suggested on the review.
2) rewrite create_node to use create_nodes. Possibly even remove any use of create_node and delete it.

3) Consider using eliot. I think Pilar's argument for this is good, but there may be a good reason not to.

Up to you whether to request a re-review of this based on the size of your changes as part of this.

Thanks,

James

flocker/provision/_aws.py
avg-I: Not sure, why?

flocker/provision/_aws.py
avg-I: Done, copied from the interface.

flocker/provision/_libcloud.py
avg-I: I think that there should be 2 blank lines between the class definition and the subsequent unrelated function. `flake8` doesn't complain here.

flocker/provision/_libcloud.py
avg-I: Same comment :-)

flocker/provision/_libcloud.py
avg-I: Yes, I referred to the hardcoded steps as the 'predefined timeout' (in the contrast to a timeout that could have been provided as an argument).

flocker/provision/_libcloud.py
avg-I: Fixed.

avg-I:
@mpgomez and @james-w, thank you for the review!

Regarding the larger comment:

> 1) small cleanups suggested on the review.

Done.

> 2) rewrite create_node to use create_nodes. Possibly even remove any use of create_node and delete it.

The reason that I haven't done that is that I wanted to improve performance of `setup-cluster` without compromising the stability of `run-acceptance-tests`.  One significant difference between the old interface and the new one is that the old one is completely synchronous (it returns an object with `INode` interface) while the new one returns a list of `Deferred`. So, the change from the old one to the new one would not be mechanical. I've created an issue for this: https://clusterhq.atlassian.net/browse/FLOC-4097

> 3) Consider using eliot. I think Pilar's argument for this is good, but there may be a good reason not to.

As I noted earlier the code didn't use eliot before this change. I do agree that all the existing `print`s in the cluster runners should be switched to eliot, so https://clusterhq.atlassian.net/browse/FLOC-4096

PR #2599, opened 2016-02-04
[FLOC-3974] Convert JSON benchmarking files to CSV

zubron:
This changes adds a script which can be used to convert JSON results from the benchmarking tool to a set of CSV files which summarise the results. This script outputs a CSV file for each different scenario and each file contains all the results that were obtained for that scenario. The results in the file are divided by number of nodes that were used in the cluster and the number of containers started in that cluster.

This script is unfortunately not very neat or clear to read so I expect that a number of changes will need to be made :smile: The approach I took was to flatten each set of results to separate each sample that was collected. The results are then filtered and combined again before being written CSV.

### DIFF
--- a/benchmark/metrics_parser.py
+++ b/benchmark/metrics_parser.py
@@ -1,5 +1,6 @@
 # Copyright ClusterHQ Inc.  See LICENSE file for details.
 
+import argparse
 import json
 from collections import defaultdict
 
@@ -8,7 +9,6 @@ example = {"flocker_control":6,
 
 def apply_cpu_metric(result_blob):
     wall_time = float(result_blob.get(u'-- WALL --'))
-    #result_blob.remove(u'--WALL--')
     for k, v in result_blob.iteritems():
         if not k == u'-- WALL --':
             result_blob[k] = v/wall_time
@@ -27,16 +27,27 @@ def print_averages(results):
         )
         print s
 
-def main(args):
-    for json_file in args:
-        with open(json_file) as f:
-            results = json.load(f)
-            # XXX adding schema json validation
-            result_table = defaultdict(list)
-            for sample in results[u'samples']:
-                for result in sample[u'value'].itervalues():
-                    apply_cpu_metric(result)
-                    add_results_to_table(result, result_table)
 
-            print result_table
-            print_averages(result_table)
+def parse_args(args):
+    parser = argparse.ArgumentParser(
+        description="Produce CSV from benchmarking results"
+    )
+    parser.add_argument("files", nargs="*", type=argparse.FileType(),
+                        help="Input JSON files to be processed")
+
+    return parser.parse_args(args).files
+
+def main(args):
+    files = parse_args(args)
+
+    for f in files:
+        results = json.load(f)
+        # XXX adding schema json validation
+        result_table = defaultdict(list)
+        for sample in results[u'samples']:
+            for result in sample[u'value'].itervalues():
+                apply_cpu_metric(result)
+                add_results_to_table(result, result_table)
+
+        print result_table
+        print_averages(result_table)
--- a/benchmark/metrics_parser.py
+++ b/benchmark/metrics_parser.py
@@ -123,19 +123,23 @@ def handle_cputime_metric(common_props, sample):
 
     :param common_props: Common properties shared with all other samples.
     :param sample: Original sample to extract values from.
+
+    :return list: The created samples.
     """
-    cputime_sample = dict(common_props)
+    cputime_samples = []
 
     wallclock_key = u'-- WALL --'
     for data in sample['value'].itervalues():
         wall_time = data[wallclock_key]
         for process, value in data.iteritems():
             if process != wallclock_key:
+                cputime_sample = dict(common_props)
                 cputime_sample['process'] = process
                 cputime_sample['value'] = value
                 cputime_sample['wallclock'] = wall_time
+                cputime_samples.append(cputime_sample)
 
-    return cputime_sample
+    return cputime_samples
 
 
 def handle_wallclock_metric(common_props, sample):
@@ -144,10 +148,14 @@ def handle_wallclock_metric(common_props, sample):
 
     :param common_props: Common properties shared with all other samples.
     :param sample: Original sample to extract values from.
+
+    :return list: The created samples.
     """
+    wallclock_samples = []
     wallclock_sample = dict(common_props)
     wallclock_sample['value'] = sample['value']
-    return wallclock_sample
+    wallclock_samples.append(wallclock_sample)
+    return wallclock_samples
 
 
 METRIC_HANDLER = {
@@ -299,7 +307,7 @@ class BenchmarkingResults(object):
 
         for sample in results['samples']:
             if sample['success']:
-                flattened.append(
+                flattened.extend(
                     METRIC_HANDLER[metric_type](common_props, sample)
                 )
         return flattened
--- a/benchmark/metrics_parser.py
+++ b/benchmark/metrics_parser.py
@@ -21,7 +21,7 @@ def add_results_to_table(ip, result, result_table):
             result_table[ip][k] = []
             result_table[ip][k].append(v)
     else:
-        for k, v in result_table[ip].iteritems():
+        for k, v in result.iteritems():
             print "V is (2) ", v, "\n"
             result_table[ip][k].append(v)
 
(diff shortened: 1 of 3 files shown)


benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+
+import argparse
+from collections import defaultdict
+import csv
+import itertools
+import json
+
+
+WALL_CLOCK_KEY = u'-- WALL --'
+
```
brendan-donegan: I suppose this is legal, but a bit unusual to have two spaces between what are essentially variable definitions.

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+                        for process, value in data.iteritems():
+                            if process != WALL_CLOCK_KEY:
+                                doc = dict(common)
+                                doc['node_ip'] = ip
+                                doc['process'] = process
+                                doc['value'] = value
+                                doc['wallclock'] = wall_time
+                                flattened.append(doc)
+                elif metric_type == 'wallclock':
+                        doc = dict(common)
```
jongiddy: 8 space indent?

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+def wallclock_for_operation(results, operation):
+    """
+    Calculate the wallclock time for a process running in a particular
+    scenario.
+
+    By default this will return the `mean` result.
+    """
+    operation_results = itertools.ifilter(
+        lambda r: r['metric']['type'] == 'wallclock'
+        and r['operation']['type'] == operation,
```
jongiddy: Put this and other broken `and` operators at end of previous line - see https://www.python.org/dev/peps/pep-0008/#id18

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+HEADERS = [
+    u'Flocker Version',
+    u'Nodes',
+    u'Containers',
+    u'Scenario',
+    u'Control Service CPU',
+    u'Dataset Agent CPU',
+    u'Container Agent CPU',
+    u'Containers Converged Within Limit',
+    u'Scenario Requests Within Limit',
```
jongiddy: Can we set these titles to name the limit (30 or 60 seconds), so it is clear in the output when if we change them?

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+                    u'Containers': container,
+                    u'Scenario': scenario,
+                    u'Control Service CPU':
+                        cputime_for_process(results, 'flocker-control'),
+                    u'Dataset Agent CPU':
+                        cputime_for_process(results, 'flocker-dataset'),
+                    u'Container Agent CPU':
+                        cputime_for_process(results, 'flocker-contain'),
+                    u'Containers Converged Within Limit':
+                        container_convergence(results, 60),
```
jongiddy: Make this a command line option, with a default  of 60 :question:

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+                    u'Scenario Requests Within Limit':
+                        request_latency(results, 30),
```
jongiddy: Ditto command line option :question:

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        self.results.extend(self._flatten(results))
+
+    def output_csv(self, prefix):
+        """
+        Output a CSV representation of a set of results.
+        """
+        summary = self._create_summary()
+        for scenario, result in summary.iteritems():
+            write_csv(result, '{0}-{1}.csv'.format(prefix, scenario))
```
brendan-donegan: In general it's better to use names here rather than ordinals

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+        flattened = []
+        common = dict(
+            [(k, results.get(k)) for k in results.iterkeys() if k != 'samples']
+        )
+
+        metric_type = results['metric']['type']
+
+        for sample in results['samples']:
+            if sample['success']:
+                if metric_type == 'cputime':
```
jongiddy: Put the code for cputime and walltime into separate functions that return the `doc` value, and call through a dictionary:
```python
     ...
     flattened.append(METRIC_HANDLER[metric_type](some common args)

METRIC_HANDLER = {
    'cputime': handle_cputime,
    'walltime': handler_walltime,
}
def handle_walltime(some common args):
    doc = dict(common)
    doc['value'] = sample['value']
    return doc
def handle_cputime(some common args):
    ...
```

This will help both with the excessive indentation here, and when we add more metrics.

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+        return len(convergences_within_limit) / num_convergences
+
+    return None
+
+
+def request_latency(results, seconds):
+    """
+    Calculate the percentage of scenario requests have a latency under the
+    specified time limit.
+    """
```
jongiddy: Document parameters. It is not clear, for example, that `seconds` is the limit.

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+    convergence_results = [
+        r for r in results if r['metric']['type'] == 'wallclock'
+        and r['operation']['type'] == 'create-container'
+    ]
+    num_convergences = len(convergence_results)
+    if num_convergences > 0:
+        convergences_within_limit = [
+            r for r in convergence_results if r['value'] < seconds
+        ]
+        return len(convergences_within_limit) / num_convergences
```
jongiddy: This appears to divide an int by an int, which rounds down.  This would mean 19 / 20 returns 0 instead of 0.95. (Unless I'm missing something?)

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+def write_csv(results, filename):
+    with open(filename, 'w') as csvfile:
+        writer = csv.DictWriter(csvfile, fieldnames=HEADERS)
+        writer.writeheader()
+        writer.writerows(results)
+
+
+def mean(values):
+    if len(values) > 0:
+        return sum(values) / len(values)
```
jongiddy: Convert the sum of values to float to ensure that this function returns a float, and not a truncated int.

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,276 @@
...
+            write_csv(result, '{0}-{1}.csv'.format(prefix, scenario))
+
+    def _versions(self):
+        """
+        Return all unique versions present in the results.
+        """
+        versions = [
+            r['control_service']['flocker_version'] for r in self.results
+        ]
+        return list(set(versions))
```
brendan-donegan: It's okay for these to be sets as itertools just needs iterables, not lists. So you can just do:

return set(<list comprehension code>)

jongiddy:
Please add a test module to check the results of the top-level functions that process the data, to ensure that they give sensible outputs for various results (20/20, 19/20, 0/20).

If you create the METRIC_HANDLER map as above, then the individual handlers should also be testable.

benchmark/test/test_metrics_parser.py
```diff
@@ -0,0 +1,342 @@
...
+                'metric': {
+                    'type': 'cputime'
+                },
+                'process': process_name,
+                'value': 30,
+                'wallclock': 100
+            },
+        ]
+        cputime_result = cpu_usage_for_process(results, process_name)
+        self.assertEqual(cputime_result, 0.25)
```
jongiddy: :question: This seems like the wrong answer. The total CPU time is 40 seconds and the total walltime is 150 seconds. Hence the average is 26.67% CPU load.

If we observed 90% load for 1 second, but another sample showed 10% load for 60 seconds, we'd probably want the result to be closer to 10% than to 90%.

OTOH, if the above is true, why run 3 samples when we could just run 1 sample for 3 times as long? Given our sample denominators are likely to be much closer than this, it probably doesn't matter much in practice.

zubron: Yes, I see what you mean. The calculation currently works by diving the value by the wall time for each sample rather than summing the values and then dividing. Most of our benchmarking runs are currently using the `wait-100` operation which means the wall clock values for the samples will be almost equal. This approach taken will be more of an issue when we starts performing other operations where the wall time values will have more variation. I'll change it.

benchmark/metrics_parser.py
```diff
@@ -0,0 +1,353 @@
...
+    parsed_args = parser.parse_args(args)
+    return parsed_args
+
+
+def main(args):
+    parsed_args = parse_args(args)
+    results = BenchmarkingResults(
+        parsed_args.convergence_limit, parsed_args.latency_limit
+    )
+    for input_file in parsed_args.files:
```
jongiddy: If there are no files, print usage message.

jongiddy:
Looks good. Please address 2 comments, and merge when green.

PR #2600, opened 2016-02-04
[FLOC-4092] Use testtools 2.0.0, rather than our fork

jml:
testing-cabal/testtools#172 has landed & been released, so we can control our logging destiny.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,6 @@
 Generate a Flocker package that can be deployed onto cluster nodes.
 """
 
-import os
 import platform
 from setuptools import setup, find_packages
 import versioneer
@@ -99,13 +98,6 @@ setup(
     cmdclass=versioneer.get_cmdclass(),
 
     dependency_links = [
-        # Use our fork of testtools until #165, #171, and #172 are merged and
-        # released. See FLOC-3498.
-        #
-        # "git+https" weirdness is due to setuptools expecting:
-        #     vcs+proto://host/path@revision#egg=project-version
-        # See https://setuptools.readthedocs.org/en/latest/setuptools.html
-        "git+https://github.com/ClusterHQ/testtools@clusterhq-fork#egg=testtools-1.9.0chq5",  # noqa
     ],
 
     # Some "trove classifiers" which are relevant.
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -45,7 +45,7 @@ sphinx-prompt==1.0.0
 sphinx-rtd-theme==0.1.8
 sphinxcontrib-httpdomain==1.3.0
 sphinxcontrib-spelling==2.1.2
-testtools==1.9.0chq5
+testtools==2.0.0
 tox==2.1.1
 troposphere==1.3.0
 six==1.10.0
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -99,13 +98,6 @@
     cmdclass=versioneer.get_cmdclass(),
 
     dependency_links = [
```
brendan-donegan: Why not remove the whole section as opposed to just the content?

james-w:
:heavy_check_mark: Looks good. Land when green after maybe making the
change Brendan suggests

jml:
Confident the tests are not failing because of this change. Merging anyway.

wallrj:
There are still quite a few ``--dependency-links`` references lingering:

```
(master ? =)[~/.../HybridLogic/flocker]$ git grep process-dependency-links
CONTRIBUTING.rst:   pip install --process-dependency-links --editable .[dev]
CONTRIBUTING.rst:.. Need --process-dependency-links while are using a fork of testtools.
admin/release.py:        ["pip install --process-dependency-links -e .[dev]"], shell=True,
build.yaml:    # Install Flocker. Use --process-dependency-links so we can temporarily use
build.yaml:    pip install --process-dependency-links  ".[dev]" ${PIP_ADDITIONAL_OPTIONS}
docs/gettinginvolved/infrastructure/packaging.rst:   pip install --process-dependency-links .[dev]
docs/gettinginvolved/infrastructure/packaging.rst:.. Need --process-dependency-links while we're using a fork of testtools.
flocker/testtools/_flaky.py:                "testtools: pip install --upgrade --process-dependency-links "
tox.ini:    pip install --process-dependency-links .[dev]
tox.ini:    pip install --process-dependency-links .[dev]
tox.ini:    pip install --process-dependency-links .[dev]
tox.ini:    pip install --process-dependency-links .[dev]

```

jml:
Why is that a problem?

jml:
To possibly save a roundtrip…

I considered removing `--process-dependency-links`, but decided not to. I'm not fully convinced I made the right decision: I don't think there's much in it either way.

Pros:
* We don't need them now, so let's remove them, because they are clutter and because YAGNI

Cons:
* We might need them later for another fork, and they are a pain in the neck to add, because everyone has to be reminded to use them (remember all the hassle when we first started using a fork of testtools?)

In the end, I came down on the side of the "con" because I don't want to break everyone's code again when we next need a Github-based fork of something. On the other hand, Tom told me that we could upload wheels to our devpi, instead of using a fork.

I'd be happy to see them removed, and less happy but still generally OK with having to remove them myself.

wallrj:
Yeah I agree that there's no harm in leaving the `--dependency-links` argument in tox.ini.

I was really just pointing out that there a bunch of places in our documentation and code where we still talk about using a fork of testtools:
 
```
Need --process-dependency-links while are using a fork of testtools.
```

```
            # We are using a fork of testtools, which unfortunately means that
            # we need to do special things to make sure we're using the latest
            # version. Raise an error message that will help people figure out
            # what they need to do.
            raise Exception(
                "Could not reset TestCase. Maybe upgrade your version of "
                "testtools: pip install --upgrade --process-dependency-links "
                ".[dev]")

```

jml:
Mostly addressed in #2616

PR #2602, opened 2016-02-04
[FLOC-4101] Add build number to github -> jenkins links.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4101

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1,4 +1,4 @@
-# vim: ai ts=4 sts=4 et sw=4 ft=yaml fdm=indent et foldlevel=0
+# vim: ai ts=4 sts=4 et sw=4 ft=yaml et
 #
 # build.yaml
 #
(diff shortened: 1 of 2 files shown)


build.yaml
```diff
@@ -1,4 +1,4 @@
-# vim: ai ts=4 sts=4 et sw=4 ft=yaml fdm=indent et foldlevel=0
+# vim: ai ts=4 sts=4 et sw=4 ft=yaml et
```
tomprince: This is unrelated, but a pain every time I open one of these files. Personal preferences (display folding) shouldn't be hard-coded into files.

jobs.groovy
```diff
@@ -609,6 +614,11 @@ def define_job(dashProject, dashBranchName, branchName, job_type, job_name,
         wrappers build_wrappers(job_values)
         triggers build_triggers(job_type, job_values.at, branchName)
         scm build_scm(git_url, branchName, isReleaseBuild)
+        steps {
+          shell(updateGitHubStatus(
+                'pending', 'Build started',
+                branchName, dashProject, dashBranchName, job_name))
+        }
```
tomprince: This could go in `build_steps`, but we don't have all the variables we need there. I could change that, but I'm not sure the value of that.

james-w:
Nice, thanks for this. Go ahead and merge.

PR #2603, opened 2016-02-04
[FLOC-4094] Adding version checking in CommonOptions

mpgomez:
The flocker version, when passed to CommonOptions (when passed as a command like argument to the acceptance test runner or the setup cluster script, --flocker-version <flocker-version>), did not validate it to start with, meaning that, it would deploy as many nodes as requested (instances), and later on, when trying to actually install flocker, would fail, and delete the nodes.
By validating the version within the options parser, we will fail sooner, and avoid creating cloud instances that we won't use at the end.


admin/test/test_acceptance.py
```diff
@@ -205,3 +207,19 @@ def test_non_json_message(self):
             )],
             self._convert(NON_JSON_JOURNAL_EXPORT),
         )
+
+
```
mpgomez: We didn't have any tests for CommonOptions that I could find, but I decided it would be a good idea to add one anyway

flocker/common/version.py
```diff
@@ -101,7 +101,7 @@ def installable_release(self):
         return self.release
 
 
-def _parse_version(version):
+def parse_version(version):
```
mpgomez: Now I am using it outside version.py, so renamed the function to be public

admin/acceptance.py
```diff
@@ -964,6 +964,16 @@ def postOptions(self):
         provider = self['provider'].lower()
         provider_config = self['config'].get(provider, {})
 
+        if self.get('flocker-version') is not None:
+            try:
+                parse_version(self['flocker-version'])
+            except UnparseableVersion:
+                raise UsageError(
+                    "Flocker vesion {!r} is not a valid version".format(
```
wallrj: Typo "vesion"

admin/test/test_acceptance.py
```diff
@@ -205,3 +207,19 @@ def test_non_json_message(self):
...
+class CommonOptionsParser(TestCase):
```
wallrj: The convention is that TestCase subclass names end with `Tests`....so `CommonOptionsParserTests`.

admin/test/test_acceptance.py
```diff
@@ -205,3 +207,19 @@ def test_non_json_message(self):
...
+    """
+    Tests for ``CommonOptions``
+    """
+    def test_invalid_version(self):
```
wallrj: Needs a docstring.

admin/test/test_acceptance.py
```diff
@@ -205,3 +207,19 @@ def test_non_json_message(self):
...
+
+        arg_options = "--distribution ubuntu-14.04 --provider AWS " \
+            "--config-file acceptance.yml " \
+            "flocker-version invalid-version"
+        common_options = CommonOptions(arg_options)
```
wallrj: `CommonOptions.__init__` expects an argument `top_level`, not the unparsed arguments.

wallrj:
```
> /home/richard/projects/HybridLogic/flocker-2/admin/test/test_acceptance.py(222)test_invalid_version()
-> exception = self.assertRaises(
(Pdb) common_options
{'dataset-backend': 'zfs', 'number-of-nodes': 2, 'config-file': None, 'branch': None, 'provider': None, 'build-server': 'http://build.clusterhq.com/', 'distribution': None, 'variants': [], 'flocker-version': None}
(Pdb) common_options.top_level
'--distribution ubuntu-14.04 --provider AWS --config-file acceptance.yml flocker-version invalid-version'
```

I'd have thought this should have blown up, but it didn't

admin/test/test_acceptance.py
```diff
@@ -205,3 +207,19 @@ def test_non_json_message(self):
...
+        self.assertRaises(
+            UsageError,
+            common_options.parseOptions, arg_options
+        )
```
wallrj: And if you drop to pdb here you'll see that the exception is different than expected because you need to supply a `list` rather than a `str` to parseOptions.

```
(Pdb) exception
UsageError('Wrong number of arguments.',)
(Pdb) list
220  	        common_options = CommonOptions(arg_options)
221  	        exception = self.assertRaises(
222  	            UsageError,
223  	            common_options.parseOptions, arg_options
224  	        )
225  ->	        import pdb; pdb.set_trace()
```

``assertRaises`` returns the exception so that you can make further assertions about the attributes of the exception.

wallrj:
Thanks @mpgomez 

I verified that the validation works:

```
$ ./admin/run-acceptance-tests --flocker-version=x.y.z --distribution=centos-7 --provider=rackspace
run-acceptance-tests: Flocker vesion 'x.y.z' is not a valid version

```

But the tests need some more work.

Please resubmit when you've checked the tests and got green builds.

mpgomez:
Implemented review comments, and waiting for a green build.
Thanks Richard for the review!

admin/acceptance.py
```diff
@@ -964,6 +964,16 @@ def postOptions(self):
...
+        if self.get('flocker-version') is not None:
+            try:
+                parse_version(self['flocker-version'])
+            except UnparseableVersion:
+                raise UsageError(
+                    "Flocker version {!r} is not a valid version".format(
+                        self['flocker-version']
+                    )
+                )
+
```
wallrj: :question: ``twisted.python.options.Options`` supports "Type Enforcement" 
 * https://twistedmatrix.com/documents/current/core/howto/options.html#type-enforcement
Consider moving this into a separate function which might be reusable elsewhere (I'd use it in my publish-installer-images script.) and which wouldn't need to handle the `None` case since that's a default value.

admin/test/test_acceptance.py
```diff
@@ -150,7 +152,8 @@ def test_cluster_id(self):
 MESSAGE=time="2015-10-02T13:33:26.192780138Z" level=info msg="GET /v1.20/containers/json"
 """
 
-class JournaldJSONFormatter(TestCase):
+
+class JournaldJSONFormatterTests(TestCase):
```
wallrj: Oh, there are other such test names. I see. Good spot.

admin/test/test_acceptance.py
wallrj: Yep, this is a great improvement.

admin/test/test_acceptance.py
```diff
@@ -205,3 +208,27 @@ def test_non_json_message(self):
...
+    """
+    def test_invalid_version(self):
+        """
+        An ``UsageError`` is thrown if an invalid flocker version is passed
+        to the ``CommonOptions``.
+        """
+        arg_options = (
+            "--distribution ubuntu-14.04 --provider AWS "
+            "--flocker-version invalid-version"
+        )
```
wallrj: :question: Personally, I'd make this a tuple rather than splitting it later. Imaging if one of the option values contained whitespace.

wallrj: I might also make `invalid-version` a variable and then make an assertion that the invalid option_value is included in the error message...which is a nice touch by the way. There's nothing more annoying than a validation error that doesn't include the invalid value.

admin/test/test_acceptance.py
```diff
@@ -205,3 +208,27 @@ def test_non_json_message(self):
...
+        common_options = CommonOptions(self.mktemp())
+        exception = self.assertRaises(
+            UsageError,
+            common_options.parseOptions, arg_options.split()
+        )
+
+        self.assertIn(
+            "Flocker version", exception.args[0]
```
wallrj: :question: This doesn't seem specific enough. Imagine if command line errors always printed the ``--help`` and if the ``--help`` included the text ``Flocker Version``.

wallrj: :question: I might  do ``assertIn(invalid_version, unicode(exception)``

wallrj:
Thanks @mpgomez 

This looks much better. I made some suggestions. Answer or address those and then merge when green.

admin/acceptance.py
jongiddy: This a nice model. However, we don't do this anywhere else, so it causes a consistency problem.  We should probably adopt this model universally, so developers understand the idiom.

PR #2605, opened 2016-02-04
[FLOC-4105] Extract the lint output in a useful way.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4105

See http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/extract-lint-output-broken/job/run_lint/1/violations/ for a demonstration ([diff](https://github.com/ClusterHQ/flocker/compare/extract-lint-output-FLOC-4105...extract-lint-output-broken))

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -1260,6 +1260,7 @@ job_type:
                    *run_lint, *exit_with_return_code_from_test ]
           }
       timeout: 10
+      publish_lint: true
       directories_to_delete: []
 
 
--- a/build.yaml
+++ b/build.yaml
@@ -878,7 +878,9 @@ common_cli:
   run_lint: &run_lint |
     # run flake8 lint tests on Flocker source code
     flake8 --format=pylint --output flake8.lint flocker
+    JOB_EXIT_STATUS="$( updateExitStatus $? )"
     pylint flocker > pylint.lint
+    JOB_EXIT_STATUS="$( updateExitStatus $? )"
 
   lint_artifacts: &lint_artifacts
     - flake8.lint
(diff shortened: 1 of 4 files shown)


tomprince:
I haven't thought whether it would be useful to also include the raw output as artifacts (which would be fairly straightforward).

build.yaml
```diff
@@ -876,16 +876,14 @@ common_cli:
...
-    unset PIP_INDEX_URL
+    # Disable exiting on error, so we can run multiple lints
+    set +e
     # run flake8 lint tests on Flocker source code
-    tox -e lint
-
-  install_flake8: &install_flake8 |
-    # flake8 is used by lint tests on Flocker source code
-    pip install flake8 ${PIP_ADDITIONAL_OPTIONS}
+    flake8 --format=pylint --output flake8.lint flocker
```
james-w: I think tox ignores a couple of files. This should too I guess. It does reveal that the two definitions will be a pain to keep in sync. Could you perhaps rewrite the tox commands to do these things and to write the files using tee?

james-w: Oh sorry. I see you found a better way to do the ignores. I still think that one way to run these things would be better, but it's perhaps not worth it.

james-w:
Hi. I think this is fine. It would be good to avoid the duplication though.

Please merge when green.

build.yaml
tomprince: I agree that it would be better to have one way of doing things, but I don't think tox is the right way. That involves building a second virtualenv and installing everything in it as well, which takes a significant portion of the time of the build.

james-w: Then how about putting this in a script in admin so devs can run it and deleting the tox definition?

tomprince: - I think a developer probably (?) would prefer the output to the screen (but personally, I just use a editor plugin).
- `admin` is getting rather disorganized.

I think getting rid of the duplication would be good, but I think this is enough of an improvement that I don't want to block it.

PR #2606, opened 2016-02-04
[DO-NOT-MERGE] Broken lints.

tomprince:
This is just here to enable looking at the Jenkins violation reports for failing lints.


james-w:
I think this has served it's purpose so closing it to clean up. Apologies if not.

PR #2609, opened 2016-02-05
[FLOC-4110] Fix log message attribute so it gets displayed correctly.

jongiddy:
The `setup-cluster-containers` script should print a final status message before exiting. To do this it uses the `eliot_output` function which maps Eliot messages to display strings.

The final log message used `action_type` instead of `message_type`, so does not get mapped correctly.

Simple change to use correct attribute.


james-w:
:+1: looks good. Merge when you are happy.

PR #2610, opened 2016-02-05
[FLOC-4111] Add package source to get_trial_environment parameters

jongiddy:
Commit https://github.com/ClusterHQ/flocker/commit/f72dd544395a4b9317f010962f369f01ac8d3c8b added an extra parameter to the function `get_trial_environment` to pass the package versions info to subsequent stages.  The  change is very useful for the benchmarking workflow, as we are currently passing the branch name manually.

However, for benchmarking, we call the `get_trial_environment` function from `cluster_setup.py`, but it has not been changed there.

This PR fixes that omission.

### DIFF
--- a/admin/cluster_setup.py
+++ b/admin/cluster_setup.py
@@ -401,11 +401,13 @@ def main(reactor, args, base_path, top_level):
     if options['no-keep']:
         print("not keeping cluster")
     else:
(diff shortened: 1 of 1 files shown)


james-w:
:+1: Looks good. Merge when you are happy.

PR #2611, opened 2016-02-05
[FLOC-4090] Refactor eliot_output

dwgebler:
Fixes FLOC-4090. Removes the three separate definitions of `eliot_output` and refactors it in to one function inside `flocker.common.script`.

### DIFF
--- a/flocker/common/test/test_script.py
+++ b/flocker/common/test/test_script.py
@@ -5,6 +5,7 @@
 import sys
 
 from eliot.testing import validateLogging, assertHasMessage
+from eliot import Message
 
 from twisted.internet import task
 from twisted.internet.defer import succeed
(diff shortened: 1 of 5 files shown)


flocker/common/test/test_script.py
```diff
@@ -300,3 +301,34 @@ def test_error(self, logger):
...
+            self.flushed_data = self.buffered_data
+            self.buffered_data = b''
+
+    def test_destination_stdout(self):
+        """
+        Eliot messages are written to stdout.
+        """
+        fake_stdout = self.StubStdout()
+        message_formats = {'flocker:eliot:test': 'Running %(command)s\n'}
+        eliot_to_stdout(message_formats, {}, stdout=fake_stdout)
```
james-w: Does this need to be reversed in some way so that the rest of the testsuite
doesn't print eliot logs to stdout?

james-w:
:+1: Hi Dave, thanks for the refactoring. It looks good to me. I have one question about
if the testsuite needs to clean up after itself.

Merge when you are happy.

flocker/common/test/test_script.py
dwgebler: `add_destination` I think is cumulative, so anything that runs after this test will continue to log messages to the stub, as well as whatever other destinations are in place. I don't think that's problematic, given the actual use cases in the codebase for `eliot_to_stdout`, but @itamarst can probably better answer the question.

PR #2612, opened 2016-02-05
[FLOC-3758]  add --log-file option to the benchmark script

avg-I:
The option allows to specify a log file to use.
The log file is appended to.
By default the log messages still go to `stderr`.

### DIFF
--- a/benchmark/script.py
+++ b/benchmark/script.py
@@ -26,8 +26,6 @@ from benchmark.cluster import BenchmarkCluster
 from benchmark._driver import driver
 
 
-to_file(sys.stderr)
-
 # Change this number when changing the format of the output JSON
 OUTPUT_VERSION = 1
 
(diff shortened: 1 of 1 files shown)


benchmark/script.py
```diff
@@ -244,6 +241,22 @@ def main(argv, environ, react=react):
     except UsageError as e:
         usage(options, e.args[0])
 
+    if options['log-file'] is not None:
+        try:
+            log_file = open(options['log-file'], 'a')
```
brendan-donegan: I would actually specify 'w+', if the command was run twice in a row with the same log file name the user would expect the contents to be replaced, not appended to. At least that's my opinion. Change that and then merge if tests pass.

brendan-donegan:
Seems good apart from the comment about overwriting rather than appending the log file. Merge after changing that and checking the tests pass.

benchmark/script.py
avg-I: I guess that different people may have different expectations about the behavior. I do not have a strong opinion on this issue, but I followed an example set in `run-acceptance-tests` (`admin/acceptance.py`) where a log file is opened in `"a"` mode.  So, I've got used to that behavior.
I guess we could add another option which would tell whether to append to a log or to overwrite it, but that feels like an overkill.

PR #2613, opened 2016-02-06
[FLOC-3942] Allow CloudFormation user to select cluster size between 3 and 10 nodes

myechuri:
Enable CloudFormation user to create a stack with size between 3 and 10 nodes (inclusive). 

Since CloudFormation cannot insert user-input number of nodes during stack creation workflow into template, the second best alternative is to parameterize Troposphere template to take in cluster size as input parameter:

```
root@ip-172-31-25-23:/home/ubuntu/flocker/admin/installer# python cloudformation.py --help
usage: cloudformation.py [-h] [-s SIZE]

Create CloudFormation template for cluster of desired size (default: 3 nodes).

optional arguments:
  -h, --help            show this help message and exit
  -s SIZE, --size SIZE  an integer corresponding to desired number of nodes in
                        the cluster. Supported sizes: min=3 max=10
root@ip-172-31-25-23:/home/ubuntu/flocker/admin/installer# 
```

Testing
----------

* New unit tests:

```
root@ip-172-31-25-23:/home/ubuntu/flocker# trial admin.installer.test.test_cloudformation.ClusterSizeLimitsTests
admin.installer.test.test_cloudformation
  ClusterSizeLimitsTests
    test_invalid_cluster_size ...                                          [OK]
    test_valid_cluster_size ...                                            [OK]

Ran 2 tests in 4.618s

PASSED (successes=2)

```
* Manual testing with max cluster size (10 nodes):

Time taken for stack creation: 12 minutes

```
ubuntu@ip-172-31-5-125:~$ export FLOCKER_CERTS_PATH=/etc/flocker
ubuntu@ip-172-31-5-125:~$ export FLOCKER_USER=user1
ubuntu@ip-172-31-5-125:~$ export FLOCKER_CONTROL_SERVICE=52.23.161.41
ubuntu@ip-172-31-5-125:~$ flockerctl status
SERVER     ADDRESS       
1d25f65b   172.31.10.114 
76e1482d   172.31.13.25  
f9f9e474   172.31.11.28  
e09eaf6a   172.31.13.139 
c8854560   172.31.13.18  
2cf443cc   172.31.10.139 
0f3f03ea   172.31.14.248 
0b0fe734   172.31.0.15   
529cc119   172.31.6.141  

ubuntu@ip-172-31-5-125:~$ flockerctl ls
DATASET   SIZE   METADATA   STATUS   SERVER 

ubuntu@ip-172-31-5-125:~$ export DOCKER_HOST=tcp://52.23.161.41:2376
ubuntu@ip-172-31-5-125:~$ export DOCKER_TLS_VERIFY=1
ubuntu@ip-172-31-5-125:~$ docker info
Containers: 9
Images: 18
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 9

```

* Automated endtoend test with largest supported cluster size (10 nodes):

```
ubuntu@ip-172-31-25-23:~/flocker$ echo $CLOUDFORMATION_TEMPLATE_URL
https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster-10-nodes.cloudformation.json
ubuntu@ip-172-31-25-23:~/flocker$ KEEP_STACK=1 CLOUDFORMATION_TEMPLATE_URL=$CLOUDFORMATION_TEMPLATE_URL KEY_PAIR=$KEY ACCESS_KEY_ID=$ACCESS_KEY_ID SECRET_ACCESS_KEY=$SECRET_ACCESS_KEY VOLUMEHUB_TOKEN='' trial flocker.acceptance.endtoend.test_installer
could not remove FilePath('/home/ubuntu/flocker/_trial_temp'), caught OSError [Errno 13]: Permission denied
  DockerComposeTests
    test_docker_compose_up_postgres ... Warning: Permanently added '54.183.54.196' (ECDSA) to the list of known hosts.
                                   [OK]

Ran 1 tests in 684.012s

PASSED (successes=1)
```


admin/installer/test/test_cloudformation.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        Attempt to create CloudFormation template for unsupported cluster size.
+        """
+        output = None
+        try:
+            self._run_cloudformation_with_cluster_size(cluster_size)
+        except CalledProcessError as e:
+            output = e.output
+        self.assertEquals(True,
+                          'InvalidClusterSizeException' in output)
```
brendan-donegan: assertIn('InvalidClusterSizeException', output) is easier to read

admin/installer/test/test_cloudformation.py
```diff
@@ -0,0 +1,98 @@
...
+        result = run_process(run_process_args)
+        return result
+
+    @given(cluster_size=valid_cluster_size)
+    def test_valid_cluster_size(self, cluster_size):
+        """
+        Create CloudFormation template for supported cluster size.
+        """
+        result = self._run_cloudformation_with_cluster_size(cluster_size)
+        self.assertEqual(True,
```
brendan-donegan: assertIn

admin/installer/cloudformation.py
```diff
@@ -179,7 +211,8 @@ def _sibling_lines(filename):
 # Gather WaitConditions
 wait_condition_names = []
 
-for i in range(NUM_NODES):
+# for i in range(NUM_NODES):
```
james-w: Not needed any more.

admin/installer/cloudformation.py
```diff
@@ -64,6 +69,33 @@ def _sibling_lines(filename):
...
+                        type=int,
+                        help='an integer corresponding to desired '
+                             'number of nodes in the cluster. '
+                             'Supported sizes: min={0} max={1}'.format(
+                                 MIN_CLUSTER_SIZE,
+                                 MAX_CLUSTER_SIZE))
+    size = parser.parse_args().size
+    if size < MIN_CLUSTER_SIZE or size > MAX_CLUSTER_SIZE:
+        raise InvalidClusterSizeException(size)
+    return size
```
james-w: :eyes: As soon as we need another option then the signature of this function
will need to change. Consider changing it now to return the result of
parser.parse_args()

admin/installer/cloudformation.py
```diff
@@ -64,6 +69,33 @@ def _sibling_lines(filename):
...
+                        default=MIN_CLUSTER_SIZE,
+                        type=int,
+                        help='an integer corresponding to desired '
+                             'number of nodes in the cluster. '
+                             'Supported sizes: min={0} max={1}'.format(
+                                 MIN_CLUSTER_SIZE,
+                                 MAX_CLUSTER_SIZE))
+    size = parser.parse_args().size
+    if size < MIN_CLUSTER_SIZE or size > MAX_CLUSTER_SIZE:
+        raise InvalidClusterSizeException(size)
```
james-w: :eyes: You can encode this as a function that you use in place of 'int' for 'type'
in 'add_argument' if you like.

admin/installer/cloudformation.py
```diff
@@ -64,6 +69,33 @@ def _sibling_lines(filename):
...
+    return size
+
+num_nodes = _get_cluster_size()
```
james-w: :eyes: Would be nice to have all this code in a function rather than at
the top level. It avoid leaking globals in to functions and it not being
caught.

admin/installer/test/test_cloudformation.py
```diff
@@ -0,0 +1,98 @@
...
+from flocker.testtools import TestCase, run_process
+
+from .. import MIN_CLUSTER_SIZE, MAX_CLUSTER_SIZE
+
+# A Hypothesis strategy for generating supported cluster size.
+valid_cluster_size = integers(min_value=MIN_CLUSTER_SIZE,
+                              max_value=MAX_CLUSTER_SIZE)
+
+# A Hypothesis strategy for generating unsupported cluster size.
+too_small_cluster_size = integers(min_value=0,
```
james-w: I'm not sure min_value should be set here. It seems like a negative value
passed in would be accepted by argparse, then rejected by the check
on size, so should be tested.

admin/installer/test/test_cloudformation.py
```diff
@@ -0,0 +1,98 @@
...
+                                  max_value=MIN_CLUSTER_SIZE-1)
+too_big_cluster_size = integers(min_value=MAX_CLUSTER_SIZE+1,
+                                max_value=100)
```
james-w: max_value of 100 seems rather arbitrary. I think I can see a rationale for doing
this, but I think letting hypothesis do its thing and not implying there is something
special about a size of 100 would be better.

admin/installer/test/test_cloudformation.py
```diff
@@ -0,0 +1,98 @@
...
+    :returns: Fully qualified pathname of cloudformation.py
+    :rtype: twisted.python.filepath.FilePath
+    """
+    root_path = b'/'
+    cloudformation_path_suffix = b'flocker/admin/installer'
+    cloudformation_file = b'cloudformation.py'
+    for root, dirs, files in walk(root_path):
+        if root.endswith(cloudformation_path_suffix) \
+           and cloudformation_file in files:
+            return FilePath("/".join((root, cloudformation_file)))
```
james-w: This walks from the root trying to find something that looks like the right path?

This seems likely to bite people later who have two trees around (Jenkins slaves
certainly do)

Can this not be written in terms of __file__?

james-w:
Hi,

Thanks for adding this.

I'm curious why you set a max cluster size?

I left a few comments, most are minor, but I think the filesystem walking
needs to be fixed before this lands, otherwise it is likely to cause
no end of confusion.

Thanks,

James

admin/installer/test/test_cloudformation.py
myechuri: Fixed, thanks!

```
ubuntu@ip-172-31-25-23:~/flocker$ trial admin.installer.test.test_cloudformation.ClusterSizeLimitsTests
admin.installer.test.test_cloudformation
  ClusterSizeLimitsTests
    test_invalid_cluster_size ...                                          [OK]
    test_valid_cluster_size ...                                            [OK]

Ran 2 tests in 4.653s

PASSED (successes=2)
ubuntu@ip-172-31-25-23:~/flocker$ 
```

admin/installer/test/test_cloudformation.py
myechuri: Fixed.

admin/installer/cloudformation.py
myechuri: Sorry, removed!

admin/installer/cloudformation.py
myechuri: Sounds good, thanks for suggesting! Updated.

admin/installer/cloudformation.py
myechuri: Thanks for the suggestion. Reworked it a little. Please let me know if it looks ok now? Thanks!

admin/installer/cloudformation.py
myechuri: Updated. Although slightly unhappy about performance impact of calling ``_get_cluster_size`` twice at that two callsites for ``num_nodes``.

admin/installer/test/test_cloudformation.py
myechuri: Good point, thanks for catching! Updated.

admin/installer/test/test_cloudformation.py
myechuri: Updated. Thanks!

admin/installer/test/test_cloudformation.py
myechuri: Thanks for catching a valid failure case! Updated to using ``__file__``.

myechuri:
@james-w @brendan-donegan : thanks a lot for the review comments! Addressed all the suggestions.

> I'm curious why you set a max cluster size?

Because we want to advertise a cluster size that is supported + tested internally. For example, when i attempted to create a cluster size of 20 nodes, stack creation failed because i hit maximum EC2 instance limit on ``flocker-training`` account. While traiging that failure, ran into following Amazon documentation: http://aws.amazon.com/ec2/faqs/#How_many_instances_can_I_run_in_Amazon_EC2

```
Q: How many instances can I run in Amazon EC2?

You are limited to running up to 20 On-Demand Instances, purchasing 20 Reserved Instances, and requesting Spot Instances per your dynamic Spot limit per region. New AWS accounts may start with limits that are lower than the limits described here. Certain instance types are further limited per region as follows:...
```
For these reasons, i set current limit for installer to 10 to be conservative - we can always increase the upper bound if we setup scale testbeds in-house and we hear from users that they couldn't use our CloudFormation template because their cluster size requirement was too big.

Please let me know if this does not sound reasonable. Thanks!

james-w:
Hi,

Thanks for the changes. I think it all looks good, land when green.

On the globals thing, my point was less that the specific function
shouldn't be at the top-level, but that the function call shouldn't be.

Rather than performing the job of the script at the global level

```
opts = parse_args()
do_the_work(opts)
```

putting this in a main function then calling it

```
def main(args):
    opts = parse_args()
    do_the_work(opts)
    return 0

if __name__ == '__main__':
    sys.exit(main(sys.argv[1:]))
```

means that you don't leak any globals.

What you have is fine though, so don't feel like you have to change this
before landing.

Thanks,

James

myechuri:
> On the globals thing, my point was less that the specific function
shouldn't be at the top-level, but that the function call shouldn't be.

Thanks a lot, @james-w , for the detailed explanation! Yes, Richard Wall suggested refactoring this module as well - we will address it as a follow up. Thanks!

PR #2614, opened 2016-02-06
[FLOC-4118] A command line tool to cleanup leaked acceptance test volumes

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4118

This is a port of the code from https://github.com/ClusterHQ/build.clusterhq.com/pull/109.

It's blocking at the moment, but in a follow up I can also port the Instance cleanup code and run instance and volume cleanup in parallel threads.

Next I'll add this to a jenkins cron job.
 * https://clusterhq.atlassian.net/browse/FLOC-4119

### DIFF
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -114,7 +114,7 @@ def _describe_volume(volume):
     }
 
 
-@attributes(["lag"])
+@attributes(["lag", "marker"])
 class CleanVolumes(object):
     """
     Destroy volumes that leaked into the cloud from the acceptance and
@@ -174,7 +174,7 @@ class CleanVolumes(object):
         :return: ``True`` if it does, ``False`` if it does not.
         """
         try:
-            return UUID(cluster_id).node == MARKER
+            return UUID(cluster_id).node == self.marker
         except:
             err(None, "Could not parse cluster_id {!r}".format(cluster_id))
             return False
@@ -303,7 +303,11 @@ class CleanupCloudResourcesOptions(Options):
         [u"volume-lag", None, timedelta(minutes=30),
          u"The oldest in minutes a volume may be "
          u"without being considered for deletion.\n",
-         lambda option_value: timedelta(minutes=int(option_value))]
+         lambda option_value: timedelta(minutes=int(option_value))],
+        [u"marker", None, MARKER,
+         u"The marker which is used "
+         u"as the ``node`` in the acceptance test cluster UUID.",
+         lambda option_value: int(option_value, base=16)]
     ]
 
     def postOptions(self):
@@ -327,7 +331,10 @@ def cleanup_cloud_resources_main(reactor, args, base_path, top_level):
         )
         raise SystemExit(1)
 
-    cleaner = CleanVolumes(lag=options["volume-lag"])
+    cleaner = CleanVolumes(
+        lag=options["volume-lag"],
+        marker=options["marker"],
+    )
 
     actions = cleaner.start(config=options["config-file"])
     d = succeed(actions)
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -69,9 +69,12 @@ def _get_volume_creation_time(volume):
     """
     Extract the creation time from an AWS or Rackspace volume.
 
-    libcloud doesn't represent volume creation time uniformly across
+    XXX: libcloud doesn't represent volume creation time uniformly across
     drivers.  Thus this method only works on drivers specifically accounted
-    for.
+    for. Should be extended or refactored for GCE support.
+
+    :param libcloud.compute.base.StorageVolume volume: The volume to query.
+    :returns: The datetime when the ``volume`` was created.
     """
     try:
         # AWS
@@ -87,7 +90,14 @@ def _get_volume_creation_time(volume):
 
 def _get_volume_region(volume):
     """
-    Get the name of the region the volume is in.
+    Extract the region name from an AWS or Rackspace volume.
+
+    XXX: libcloud doesn't represent volume creation time uniformly across
+    drivers.  Thus this method only works on drivers specifically accounted
+    for. Should be extended or refactored for GCE support.
+
+    :param libcloud.compute.base.StorageVolume volume: The volume to query.
+    :returns: The region of the ``volume``
     """
     return (
         # Rackspace
@@ -101,6 +111,9 @@ def _describe_volume(volume):
     """
     Create a dictionary giving lots of interesting details about a cloud
     volume.
+
+    :param libcloud.compute.base.StorageVolume volume: The volume to query.
+    :returns: A JSON serializable dict of ``volume`` information.
     """
     return {
         'id': volume.id,
@@ -162,6 +175,7 @@ class CleanVolumes(object):
         """
         Extract the Flocker-specific cluster identifier from the given volume.
 
+        :param libcloud.compute.base.StorageVolume volume: The volume to query.
         :raise: ``KeyError`` if the given volume is not tagged with a Flocker
             cluster id.
         """
@@ -294,6 +308,7 @@ def _yaml_configuration_path_option(option_name, option_value):
 
 class CleanupCloudResourcesOptions(Options):
     """
+    Command line options for ``cleanup_cloud_resources``.
     """
     optFlags = [
         ["dry-run", None,
@@ -302,7 +317,7 @@ class CleanupCloudResourcesOptions(Options):
 
     optParameters = [
         [u"config-file", None, None,
-         u"The config file containing cloud credentials.\n",
+         u"An acceptance.yml file containing cloud credentials.\n",
          lambda option_value: _yaml_configuration_path_option(
              u"config-file", option_value
          )],
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -159,7 +159,7 @@ class CleanVolumes(object):
         for extra in config.get("extra-aws", []):
             extra_driver_config = base_ec2.copy()
             extra_driver_config.update(extra)
-            drivers.append(get_ec2_driver(config))
+            drivers.append(get_ec2_driver(extra_driver_config))
         return drivers
 
     def _get_cloud_volumes(self, drivers):
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -1,0 +1,29 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+"""
+Utilities for cloud resource cleanup.
+"""
+import sys
+
+from twisted.internet.defer import succeed
+from twisted.python.usage import Options, UsageError
+
+
+class CleanupCloudResourcesOptions(Options):
+    """
+
+    """
+
+
+def cleanup_cloud_resources_main(reactor, args, base_path, top_level):
+    options = CleanupCloudResourcesOptions()
+
+    try:
+        options.parseOptions(args)
+    except UsageError as e:
+        sys.stderr.write(
+            "Usage Error: %s: %s\n" % (
+                base_path.basename(), e
+            )
+        )
+        raise SystemExit(1)
+    return succeed(None)
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -120,14 +120,15 @@ class CleanVolumes(object):
     Destroy volumes that leaked into the cloud from the acceptance and
     functional test suites.
     """
-    def start(self, config):
+    def start(self, config, dry_run=False):
         """
         Clean up old volumes belonging to test-created Flocker clusters.
         """
         drivers = self._get_cloud_drivers(config)
         volumes = self._get_cloud_volumes(drivers)
         actions = self._filter_test_volumes(self.lag, volumes)
-        self._destroy_cloud_volumes(actions.destroy)
+        if not dry_run:
+            self._destroy_cloud_volumes(actions.destroy)
         return actions
 
     def _get_cloud_drivers(self, config):
@@ -294,6 +295,11 @@ def _yaml_configuration_path_option(option_name, option_value):
 class CleanupCloudResourcesOptions(Options):
     """
     """
+    optFlags = [
+        ["dry-run", None,
+         "Just print the calculated actions. Don't delete anything."]
+    ]
+
     optParameters = [
         [u"config-file", None, None,
          u"The config file containing cloud credentials.\n",
@@ -311,8 +317,7 @@ class CleanupCloudResourcesOptions(Options):
     ]
 
     def postOptions(self):
-        """
-        """
+        self["dry-run"] = bool(self["dry-run"])
         if self["config-file"] is None:
             raise UsageError("Missing --config-file option.")
 
@@ -336,7 +341,10 @@ def cleanup_cloud_resources_main(reactor, args, base_path, top_level):
         marker=options["marker"],
     )
 
-    actions = cleaner.start(config=options["config-file"])
+    actions = cleaner.start(
+        config=options["config-file"],
+        dry_run=options["dry-run"],
+    )
     d = succeed(actions)
     d.addCallback(print_result)
     return d
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -300,6 +300,10 @@ class CleanupCloudResourcesOptions(Options):
          lambda option_value: _yaml_configuration_path_option(
              u"config-file", option_value
          )],
+        [u"volume-lag", None, timedelta(minutes=30),
+         u"The oldest in minutes a volume may be "
+         u"without being considered for deletion.\n",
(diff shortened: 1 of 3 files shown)


wallrj:
I haven't attempted to fix the bugs in the existing code.
It will still fail with `in-use` errors just like the buildbot version does:
 * http://build.clusterhq.com/builders/clean-old-resources

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+         lambda option_value: _yaml_configuration_path_option(
+             u"config-file", option_value
+         )],
+        [u"volume-lag", None, timedelta(minutes=30),
+         u"The oldest in minutes a volume may be "
+         u"without being considered for deletion.\n",
+         lambda option_value: timedelta(minutes=int(option_value))],
+        [u"marker", None, MARKER,
+         u"The marker which is used "
+         u"as the ``node`` in the acceptance test cluster UUID.",
```
sarum90: :question: I'm not super familiar with the UUID version 1 spec (http://tools.ietf.org/html/rfc4122#section-4.1.6), so ``node`` looked a bit confusing here. perhaps just call it the rightmost 48 bits of the UUID? Totally optional.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+def get_ec2_driver(aws):
+    """
+    Get a libcloud EC2 driver given some credentials and other configuration.
+    """
+    ec2 = get_driver(Provider.EC2)(
+        aws['access_key'], aws['secret_access_token'],
+        region=aws['region'],
+    )
+    return ec2
+
```
sarum90: :question: Looks like http://libcloud.readthedocs.org/en/latest/compute/drivers/gce.html exists as well.  Not sure if you should have to deal with this right now or just leaving a note for later.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    else:
+        return parse_date(date_string)
+
+
+def get_rackspace_driver(rackspace):
+    """
+    Get a libcloud Rackspace driver given some credentials and other
+    configuration.
+    """
+    rackspace = get_driver(Provider.RACKSPACE)(
```
sarum90: :question: One question on the approach -- I think it's probably more important for us to have something in place rather than do everything precisely correct so more just take this into consideration rather than go off and implement it.

So we already call `get_driver(Provider.RACKSPACE)` in `flocker.provision._rackspace.rackspace_provisioner`. This suggested to me that we might actually want to extend the provisioner to have whatever functionality we need to clean up volumes rather than do a separate libcloud interaction here.

Pros:

- Makes it obvious that you have to add this functionality when you add a new cloud (For instance, I know for GCE I had to implement ``IProvisioner``, I didn't know that for GCE I had to find this script and add GCE support).

- We might need to ensure that the way we query for volumes is consistent with how we are storing the volumes. (This might even suggest that we need a new endpoint in the IBlockDeviceAPI). For instance, for GCE to get the cluster_id you have to query the description, parse it as JSON, and extract a specific key. We might want to change that API to add a specific tag for the cluster. That requires logic in the cleanup code that must be in-sync with the implementation of the IBlockDeviceAPI. Seems like it be nice if that was behind the same interface.

Cons:

- This is implemented.

- `libcloud` already gives us a common interface to multiple clouds, why are we just adding provisioner layers plumbing through to libcloud?

- Really we to be able to query for all ways that we have ever stored volume data, that sort of code might be better served in isolated cleanup code rather than confusing our `pristine` driver implementations.

---------------------

Something else occurred to me, what if we just extend the provisioner interface to have a method that gets a libcloud driver?  You could create the drivers by creating all provisioners in https://github.com/ClusterHQ/flocker/blob/master/flocker/provision/__init__.py#L16, and then we wouldn't even have to change this file to get support for other clouds.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+
+def _get_volume_creation_time(volume):
+    """
+    Extract the creation time from an AWS or Rackspace volume.
+
+    libcloud doesn't represent volume creation time uniformly across
+    drivers.  Thus this method only works on drivers specifically accounted
+    for.
+    """
+    try:
```
sarum90: This looks like something that wants to be behind a one-implementation-per-cloud-type interface.

In theory by the structure of the code we already know from which cloud the volume came from, any way we could change the structure of the code to only execute one of these code blocks?

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+        # experimentation indicates timestamps are in UTC (which of course
+        # is the only reasonable thing).
+        return parse_date(
+            volume.extra['created_at']
+        ).replace(tzinfo=tzutc())
+
+
+def _get_volume_region(volume):
+    """
+    Get the name of the region the volume is in.
```
sarum90: :question: If you choose not to put this behind an interface, could you change this comment to something more like the one above?

Specifically, note that this is "best effort" or "Only works for AWS and Rackspace" and maybe also grumble about libcloud not actually presenting a uniform interface to volume data.

Looks like this is just used for logging on `--dry_run` so best effort is probably fun, but perhaps even call that out with a `N.B.`

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    except KeyError:
+        # Rackspace.  Timestamps have no timezone indicated.  Manual
+        # experimentation indicates timestamps are in UTC (which of course
+        # is the only reasonable thing).
+        return parse_date(
+            volume.extra['created_at']
+        ).replace(tzinfo=tzutc())
+
+
+def _get_volume_region(volume):
```
sarum90: :question: just fyi, GCE calls "the thing that represents the location your resource is in / routing information for requests" a `zone` so maybe "routing_location" or some higher level abstraction would make sense for this. Although `region` is what is used elsewhere in our code, so perhaps it isn't worth the confusion to change it.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    volume.
+    """
+    return {
+        'id': volume.id,
+        'creation_time': _format_time(
+            _get_volume_creation_time(volume),
+        ),
+        'provider': volume.driver.name,
+        'region': _get_volume_region(volume),
+        # *Stuffed* with non-JSON-encodable goodies.
```
sarum90: :question: what sort of goodies? I'm just actually curious.

Also any sense in just printing the repr of the whole volume?

sarum90:
Throughout there seem to be missing parameter type documentation and return type documentation, and a few method docstrings.   I'll comment on the ones I can't figure out, but we should change our pylint rules if we really want all of them documented rather than having me comment on every one.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+        region=aws['region'],
+    )
+    return ec2
+
+
+def _get_volume_creation_time(volume):
+    """
+    Extract the creation time from an AWS or Rackspace volume.
+
+    libcloud doesn't represent volume creation time uniformly across
```
sarum90: :question: perhaps explicitly note that the type is a subclass of ``libcloud.compute.base.StorageVolume`

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+        if not dry_run:
+            self._destroy_cloud_volumes(actions.destroy)
+        return actions
+
+    def _get_cloud_drivers(self, config):
+        """
+        From the private buildbot configuration, construct a list of all of the
+        libcloud drivers where leaked volumes might be found.
+        """
+        base_ec2 = config["aws"]
```
sarum90: nit: inline below just like for rackspace. Better yet, leverage https://github.com/ClusterHQ/flocker/blob/master/flocker/provision/__init__.py#L16 somehow so you don't have to explicitly list all supported clouds here as well.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+
+        drivers = [
+            get_rackspace_driver(config["rackspace"]),
+            get_ec2_driver(base_ec2),
+        ]
+
+        for extra in config.get("extra-aws", []):
```
sarum90: Document this functionality. Looks like you have two aws config blobs in your acceptance.yaml or something? Was that a testing thing? What is extra-aws supposed to look like?

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    Format a time to ISO8601 format.  Or if there is no time, just return
+    ``None``.
+    """
+    if when is None:
+        return None
+    return datetime.isoformat(when)
+
+
+def _get_tag(volume, tag_name):
```
sarum90: I think you actually just want `_get_flocker_cluster_id(volume)`.

I don't think we want to commit to every cloud having a concept of "tags", that tags are key:value pairs, or that flocker-cluster-id will be the key we will use for this tag.

For instance, on GCE tags are more like flags (just a set of strings that match `/[a-z]([a-z0-9-]{0,62}[a-z0-9])?/` you can flag an instance with), and thus are not used for storing cluster id in the current implementation. So instead we set the free-from description field to a JSON blob that we parse to get the cluster id.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    libcloud doesn't represent tags uniformly across drivers.  Thus this method
+    only works on drivers specifically account for.
+
+    :raise: ``KeyError`` if the tag is not present.
+    """
+    return volume.extra.get("tags", volume.extra.get("metadata"))[tag_name]
+
+
+@attributes(["destroy", "keep"])
+class VolumeActions(object):
```
sarum90: :question: any reason not to use a PClass?

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+            u"Unable to parse YAML from {}. "
+            u"Error message: {}.".format(
+                option_name, yaml_path.path, unicode(e)
+            )
+        )
+    return configuration
+
+
+class CleanupCloudResourcesOptions(Options):
+    """
```
sarum90: nit: empty docstring. Is this some weird quirk about `Options` classes, or should this be filled with goodies or removed?

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+         u"without being considered for deletion.\n",
+         lambda option_value: timedelta(minutes=int(option_value))],
+        [u"marker", None, MARKER,
+         u"The marker which is used "
+         u"as the ``node`` in the acceptance test cluster UUID.",
+         lambda option_value: int(option_value, base=16)]
+    ]
+
+    def postOptions(self):
+        self["dry-run"] = bool(self["dry-run"])
```
sarum90: :question: curious, does this make --dry-run=no do something funny? I just don't know twisted options, probably fine as is.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    """
+    """
+    optFlags = [
+        ["dry-run", None,
+         "Just print the calculated actions. Don't delete anything."]
+    ]
+
+    optParameters = [
+        [u"config-file", None, None,
+         u"The config file containing cloud credentials.\n",
```
sarum90: I _think_ (or maybe hope) this is the same as `acceptance.yml`, but not immediately clear. Want to leave me a breadcrumb of something to that extent.

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+from libcloud.compute.providers import get_driver, Provider
+
+from twisted.internet.defer import succeed
+from twisted.python.filepath import FilePath
+from twisted.python.log import err
+from twisted.python.usage import Options, UsageError
+
+# Marker value defined by ``flocker.testtools.cluster_utils.MARKER``.  This
+# should never change and should always identify test-created clusters.
+from flocker.testtools.cluster_utils import MARKER
```
sarum90: I think benchmarking might also  use this marker. Perhaps they need a new one? Or we could default to more than 30 minutes. Certainly things created over a day ago should be safe to clean up.

docs/gettinginvolved/acceptance-testing.rst
```diff
@@ -321,6 +321,27 @@ And then run the acceptance tests on those nodes using the following command:
      --flocker-version='' \
      flocker.acceptance.obsolete.test_containers.ContainerAPITests.test_create_container_with_ports
 
+
+Test Cleanup
+============
+
+``admin/cleanup_cloud_resources`` creates a number of test datasets.
```
sarum90: I think this was supposed to say admin/run_acceptance_test

admin/cleanup.py
```diff
@@ -0,0 +1,379 @@
...
+    )
+    if len(actions.destroy) > 0:
+        # We fail if we destroyed any volumes because that means that
+        # something is leaking volumes.
+        raise SystemExit(1)
+
+
+def _serialize_volume_actions(actions):
+    serializable = {}
+    for kind in ('destroy', 'keep'):
```
sarum90: :question: should we print out the keep? I guess it's nice if you want to be like, "why isn't my volume being deleted", but it's kind of noisy in general. (running this in us-west-2 gets 1517 lines of output). That said I can always pipe into less or use jq, so maybe better as is.

sarum90:
Looks great! ran it locally and it seems to work (although with --dry_run enabled).

Some concerns throughout, but nothing too large. I think some of my comments drifted a little too far down a rabbit hole. For instance, I'm not sure if it's actually worth the work to make a new volume interface. Might be nicer, but it's not too much work to add GCE even to this existing script.  It's not super obvious where all you need to add code when you add a new cloud, but even after just doing this review I think I know the things that would have to be added.

admin/cleanup.py
wallrj: It is specifically used as the UUID node field here:
 * https://github.com/ClusterHQ/flocker/blob/6adff92a1e95a99afa509461941bfbf348e3638c/flocker/testtools/cluster_utils.py#L76

admin/cleanup.py
wallrj: I'm going to limit this branch to porting the existing code. I created a followup issue:
 * https://clusterhq.atlassian.net/browse/FLOC-4128

admin/cleanup.py
wallrj: I won't attempt to do this here. I created a followup issue (also mentioned above);
 * https://clusterhq.atlassian.net/browse/FLOC-4128

admin/cleanup.py
wallrj: Looks like @brendan-donegan 's working on this from a different angle:
 * https://clusterhq.atlassian.net/browse/FLOC-3981

I guess we should swap notes.

admin/cleanup.py
wallrj: I'll got with the docstring change for now. 
Can add an interface as part of the GCE support.

admin/cleanup.py
wallrj: understood.

admin/cleanup.py
wallrj: It includes "Optional provider specific attributes":
 * https://libcloud.readthedocs.org/en/latest/compute/api.html#libcloud.compute.base.StorageVolume

I guess we could / should list the `state` attribute too, but I think that's duplicated in in the extras dict too.

admin/cleanup.py
wallrj: Done.

admin/cleanup.py
wallrj: I haven't tried this option but I think it's so you can add other regions to cleanup and have those merged with the credentials of the primary region.
Again, this is just something that's present in the existing buildbot job.

admin/cleanup.py
wallrj: I'll defer this until we add GCE support.

admin/cleanup.py
wallrj: In a followup issue.

admin/cleanup.py
wallrj: An oversight. I've added a docstring now.

admin/cleanup.py
wallrj: ``t.p.options.Options`` sets the value to 0 or 1. I just wanted to get the types correct as soon as possible.

wallrj: https://twistedmatrix.com/documents/current/core/howto/options.html#boolean-options

admin/cleanup.py
wallrj: Done.

admin/cleanup.py
wallrj: And it turns out to be broken. We should be passing ``extra_config`` rather than ``config`` to ``get_ec2_driver``.

I'll fix it in a follow up.

wallrj:
There are some build failures caused by Jenkins bugs. I'm going to merge this anyway.

PR #2616, opened 2016-02-08
[FLOC-4113] Expand functional test set up docs

dwgebler:
Fixes FLOC-4113. Explains that cloud backend tests must be run from an instance of the cloud provider and how to set up an instance to run the functional tests.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -62,9 +62,7 @@ Install Flocker's development dependencies in a ``virtualenv`` by running the fo
 .. prompt:: bash $
 
    mkvirtualenv flocker
-   pip install --process-dependency-links --editable .[dev]
-
-.. Need --process-dependency-links while are using a fork of testtools.
+   pip install --editable .[dev]
 
 .. _Docker: https://www.docker.com/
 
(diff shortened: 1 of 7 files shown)


james-w:
Hi,

Thanks for fixing this, it would have helped me a lot.

My only comments are that the instructions for getting the dev
env don't quite match everywhere else:

  1. tox does pip install -r requirements.txt before .[dev]. It might well be implicit though?

  2. We have --process-dependency-links everywhere else. It's not needed today because we dropped the testtools fork, but I think we should remain consistent, and either include it here, or remove it from everywhere else.

Also, having a command of some sort that wrapped these steps would be useful, to reduce some of this duplication.

Thanks,

James

james-w:
Oh, merge when you are happy.

dwgebler:
> tox does pip install -r requirements.txt before .[dev]. It might well be implicit though?

`pip install -e .[dev]` or `pip install --editable .[dev]` is functionally equivalent, our `setup.py` reads the requirements file to build its dependency list.

> We have --process-dependency-links everywhere else. It's not needed today because we dropped the testtools fork, but I think we should remain consistent, and either include it here, or remove it from everywhere else.

I've now removed other docs references to `process-dependency-links` in this branch. It's deprecated and gives a warning and seeing as we no longer need it, seems to make more sense to take it out everywhere.

james-w:
--process-dependency-links is still used in every script that installs flocker though.

dwgebler:
> --process-dependency-links is still used in every script that installs flocker though

Also removed from scripts now.

PR #2619, opened 2016-02-08
[FLOC-4119] Configure Jenkins to run ``cleanup_cloud_resources`` daily

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4119

I'm not sure whether Jenkins will add my new build step automatically.

If it appears to work, I'll remove the ``--dry-run`` option and rebuild.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -666,7 +666,6 @@ common_cli:
     ${venv}/bin/python setup.py --version
     # Cleanup leaked or stale cloud instances and volumes.
     ${venv}/bin/python admin/cleanup_cloud_resources \
-    --dry-run \
     --config-file /tmp/acceptance.yaml \
     -- 1>cleanup_cloud_resources.stdout 2>cleanup_cloud_resources.stderr
 
--- a/build.yaml
+++ b/build.yaml
@@ -662,6 +662,14 @@ common_cli:
   cleanup_old_jenkins_osx_yosemite_boxes: &cleanup_old_jenkins_osx_yosemite_boxes |
     admin/ci-tools/cleanup-vagrant-boxes clusterhq/jenkins-osx-yosemite
 
+  cleanup_cloud_resources: &cleanup_cloud_resources |
+    ${venv}/bin/python setup.py --version
+    # Cleanup leaked or stale cloud instances and volumes.
+    ${venv}/bin/python admin/cleanup_cloud_resources \
+    --dry-run \
+    --config-file /tmp/acceptance.yaml \
+    -- 1>cleanup_cloud_resources.stdout 2>cleanup_cloud_resources.stderr
+
   new_vagrantfile_for_osx_yosemite: &new_vagrantfile_for_osx_yosemite |
     # this builds a Vagrantfile which will use the latest jenkins-osx-yosemite
     # vagrant box built overnight.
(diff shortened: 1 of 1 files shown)


wallrj:
Hmm. The build failed, but it's not clear why: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/sweep-the-cloud-from-jenkins-FLOC-4119/job/_cleanup_cloud_resources/1/console

I don't think Jenkins even attempted to run my command.

wallrj:
I figured it out. The command was being run but Jenkins doesn't show the commands it's running and the stdout and stderr had disappeared with the slave.
I'm now archiving the stdout and stderr to separate files.

You can see the list of volumes that would be destroyed by downloading the stdout file and  using the following jq line:

```
jq '.destroy[]| {id: .id, state: .extra | gsub("['"'"']"; "\"")| gsub("None"; "null") | gsub("datetime.datetime\\([^)]+\\)"; "null") | fromjson| .state, region: .region, provider: .provider, created: .creation_time}' ~/Downloads/cleanup_cloud_resources.stdout
```

For the time being, the script will fail because Rackspace has a bunch of volumes that are in an inconsistent state and which we cannot delete.
I've contacted Rackspace and asked them to fix those.
 * https://mycloud.rackspace.com/cloud/929000/tickets#160209-dfw-0000148

You can see the errors in stderr.

I was planning that I'd tackle https://clusterhq.atlassian.net/browse/FLOC-3480 next and attempt to force destroy those volumes, but I don't think the Rackspace API policy allows it.
At least I haven't been able to use the ``cinder force-delete` or `cinder reset-state` commands with the inconsistent volumes.

build.yaml
```diff
@@ -662,6 +662,13 @@ common_cli:
   cleanup_old_jenkins_osx_yosemite_boxes: &cleanup_old_jenkins_osx_yosemite_boxes |
     admin/ci-tools/cleanup-vagrant-boxes clusterhq/jenkins-osx-yosemite
 
+  cleanup_cloud_resources: &cleanup_cloud_resources |
+    ${venv}/bin/python setup.py --version
+    # Cleanup leaked or stale cloud instances and volumes.
+    ${venv}/bin/python admin/cleanup_cloud_resources \
+    --config-file /tmp/acceptance.yaml \
+    -- 1>cleanup_cloud_resources.stdout 2>cleanup_cloud_resources.stderr
```
james-w: Did you consider using tee here?

Using tee with stderr>stdout redirection would:

  * Allow us to see what the job is doing while it is running.
  * Keep interleaving of messages in the output files.

However, it would mean that didn't know whether a message went to which channel,
and perhaps lose the ability to process the output.

james-w:
Hi

Thanks for adding the job.

I think this is fine, feel free to land it when you are happy.

However, given that this is a very static job, and we don't really need a copy
we aren't going to run for every PR, you could consider adding this to
ci-platform as a "seed job". It would then be deployed only once at the
top-level, like ReconnectDisconnectedSlaves. You can grab the XML
config for the job you ran from the master, tweak it and drop it in
the directory in ci-platform (I don't think it needs to be named explicitly
in ansible, but it might.)

I think that would be a better setup if you are willing to change your
approach slightly.

Thanks,

James

wallrj:
Thanks @james-w 

> However, given that this is a very static job, and we don't really need a copy we aren't going to run for every PR, you could consider adding this to ci-platform as a "seed job".
> You can grab the XML config for the job you ran from the master, tweak it and drop it in the directory in ci-platform (I don't think it needs to be named explicitly in ansible, but it might.)

I searched for xml files containing `admin/cleanup_cloud_resources` in `/var/*/jenkins` but didn't find anything (in a reasonable time) and the ci-platform docs don't say much about this process.

```
[centos@ci-live ~]$ sudo find /tmp /opt/jenkins /var/*/jenkins/ -iname '*.xml' -not -ipath '*fingerprints*' -print0 | sudo xargs  -0 egrep 'admin/cleanup_cloud_resources'

```

And in any case, the `admin/cleanup_cloud_resources` command imports stuff from flocker so I can imagine future situations where I'd want to run this build job for a specific Flocker branch.

So I think I'll leave it in build.yml for now. 

I won't do the `tee` thing because, as you say, it's useful to be able to examine the clean JSON results and the logged errors separately.

PR #2620, opened 2016-02-09
[FLOC-4099] Disable API response validation

jongiddy:
Profiling shows that the JSON schema validation of control service API responses causes a large proportion of CPU load on the control service for clusters containing a large number of containers. See the comments for issue FLOC-4099 for some numbers.

This PR disables validation of the responses for contexts other than unit testing.

The principle being moved towards is to validate data when it arrives at the service and then to assume it is valid from there on, including on the way out.  This should be true if 1. all input data is validated, and 2. there are no bugs.

FLOC-4121 will review and, if required, update the situation for validating all input data, and validation for unit tests is kept to detect bugs.

In addition, I updated the `benchmark.yml` file defining the benchmark scenarios to reflect the actual tests we've been running.

### DIFF
--- a/flocker/restapi/test/test_infrastructure.py
+++ b/flocker/restapi/test/test_infrastructure.py
@@ -21,6 +21,7 @@ from twisted.web.http import (
     BAD_REQUEST, INTERNAL_SERVER_ERROR, PAYMENT_REQUIRED, GONE,
     NOT_ALLOWED, NOT_FOUND, OK)
 
+from .. import _infrastructure
 from .._infrastructure import (
     EndpointResponse, user_documentation, structured, UserDocumentation)
 from .._logging import REQUEST, JSON_REQUEST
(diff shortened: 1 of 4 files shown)


flocker/restapi/test/test_infrastructure.py
```diff
@@ -577,6 +581,26 @@ def test_responseValidationError(self, logger):
...
+    @validateLogging(_assertRequestLogged(b"/foo/badresponse", b"GET"))
+    def test_responseNoValidation(self, logger):
+        """
+        If _validate_responses is False, then JSON is not validated.
+        """
+        def cleanup(saved=_infrastructure._validate_responses):
+            _infrastructure._validate_responses = saved
+        self.addCleanup(cleanup)
+
+        _infrastructure._validate_responses = False
```
james-w: This can use self.patch.

james-w:
Hi,

Thanks for the performance improvement.

I don't really like the global state, but given that all of this is built with
decorators I don't think the alternatives are much better.

Merge when you are happy.

PR #2621, opened 2016-02-09
[FLOC-3862] Do not accept floats for maximum size of a dataset

jongiddy:
The `maximum_size` property of datasets is an integer:
- The API docs text for `maximum_size` state it is an (optional) integer.
- The Dataset PClass in `flocker.apiclient` constrins it to an integer.
- No fractional value of bytes is valid (in fact only multiples of 1024 are valid).

However, the server-side JSON schema allows the type of `maximum_size` to be `number` which permits floats. The Dataset PClass on the server side does not constrain the type.  So, a float type is not detected as an error until it is passed back to the client, after the control service data is made invalid.

The test that should have detected this fails since the value tested did not meet a different constraint (multiple of 1024).

This PR fixes the test and the constraint to only allow integers.


james-w:
:+1: Thanks for fixing this. Merge when happy.

PR #2622, opened 2016-02-09
[FLOC-4124] Patch cluster management scripts to use eliot_to_stdout

brendan-donegan:
eliot_to_stdout moved and it's usage in e.g. cluster_setup.py wasn't caught as there were no tests of any kind for that. This PR:

- Updates cluster_setup.py and cluster_add_nodes.py to use eliot_to_stdout (via a convenience function)
- Patches eliot_to_stdout to be less verbose when used in the cluster scripts (this was done on the advice of dwgebler)
- Added a basic test for the options parsing. Note that I chose to introduce use of mock in order to get around the fact that the option parser was embedding quite a lot of logic to connect to the cloud provider.

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -1432,7 +1432,7 @@ def main(reactor, args, base_path, top_level):
     """
     options = RunOptions(top_level=top_level)
 
-    eliot_logging_acceptance()
+    configure_eliot_logging_for_acceptance()
     try:
         options.parseOptions(args)
     except UsageError as e:
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -1254,7 +1254,11 @@ ACTION_START_FORMATS = {
 }
 
 
-def eliot_logging_acceptance():
+def configure_eliot_logging_for_acceptance():
+    """
+    Set up eliot logging for use in scripts which are used in
+    acceptance tests
+    """
     eliot_to_stdout(MESSAGE_FORMATS, ACTION_START_FORMATS)
 
 
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -1254,6 +1254,10 @@ ACTION_START_FORMATS = {
 }
 
 
+def eliot_logging_acceptance():
+    eliot_to_stdout(MESSAGE_FORMATS, ACTION_START_FORMATS)
+
+
 def capture_upstart(reactor, host, output_file):
     """
     SSH into given machine and capture relevant logs, writing them to
@@ -1424,7 +1428,7 @@ def main(reactor, args, base_path, top_level):
     """
     options = RunOptions(top_level=top_level)
 
-    eliot_to_stdout(MESSAGE_FORMATS, ACTION_START_FORMATS)
+    eliot_logging_acceptance()
     try:
         options.parseOptions(args)
     except UsageError as e:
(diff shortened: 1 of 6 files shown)


admin/acceptance.py
```diff
@@ -1254,6 +1254,10 @@ def parseArgs(self, *trial_args):
 }
 
 
+def eliot_logging_acceptance():
```
james-w: We put docstrings on pretty much every function.

Also it's not clear to me from the name what this function does. There's no verb,
so I don't know what the subject is. "configure_eliot_for_acceptance" or something
would solve that.

admin/test/test_cluster_setup.py
```diff
@@ -0,0 +1,25 @@
...
+
+from flocker.testtools import TestCase
+
+from ..acceptance import CommonOptions
+from ..cluster_setup import RunOptions
+
+
+class RunOptionsTest(TestCase):
+
+    @mock.patch.object(CommonOptions, 'postOptions')
```
james-w: Rather than mocking I think subclassing in the test module would make sense here.

```python
class RunOptionsForTest(RunOptions):


```

james-w:
Hi,

Thanks for fixing these issues.

I have suggested a couple of small changes.

Why did you choose to test the options parsing in this branch?
I guess importing the module to get the options subclass provides
a basic check that would have caught this, and it's a framework for
adding more tests?

I'm -1 on adding a dependency on mock for this, and have suggested
an alternative approach (the one proposed by @jml)

Please resubmit when you've made the changes, or let's further discuss
how to approach this.

Thanks,

James

admin/test/test_cluster_setup.py
brendan-donegan: That actually changes the nature of the test, as I'm mocking the postOptions of CommonOptions (the parent) rather than of RunOptions. This requires more code and would rapidly become ugly. I'm willing to change it here for the sake of getting this landed, but I'd question whether it's correct to avoid writing certain kinds of tests in order to avoid adding mock as a dependency.

admin/test/test_cluster_setup.py
```diff
@@ -0,0 +1,27 @@
...
+from flocker.testtools import TestCase
+
+from ..acceptance import CommonOptions
+from ..cluster_setup import RunOptions
+
+
+class RunOptionsForTest(RunOptions):
+
+    def postOptions(self):
+        pass
```
james-w: I would suggest a comment to make it clear why we are doing this.

flocker/common/script.py
```diff
@@ -164,15 +164,15 @@ def eliot_output(message):
         action_type = message.get('action_type')
         action_status = message.get('action_status')
 
-        message_format = '%s'
+        message_format = ''
```
james-w: If we aren't printing the messages do they go somewhere useful?

james-w:
:+1: Great, thanks for changing this. I left a couple of small comments.
Sorry for not asking the second on the first review. Merge when you are happy.

flocker/common/script.py
avg-I: @james-w The messages still go to other destinations if configured. E.g. `run-acceptance-tests` and `setup-cluster` add `FileDestination` so that all eliot log messages go to a log file.

PR #2624, opened 2016-02-09
[FLOC-4102] Update the publish-installer-images script to support specifying a Docker version

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4102

I've redone the way we install Docker so that I can specify the Docker version.

Also took the opportunity to support "FLOCKER_VERSION" rather than "FLOCKER_BRANCH".

Also added AMI tags for all the versions that were specified.

I've run the scripts (as documented) and you can see the resulting images by logging into AWS AMI page.


admin/installer/packer/provision_ubuntu-14.04_docker.sh
```diff
@@ -1,12 +1,24 @@
 #!/bin/bash
 set -ex
 
-# Install Docker like Vagrant does
-# https://github.com/mitchellh/vagrant/blob/master/plugins/provisioners/docker/cap/debian/docker_install.rb#L13
+# Install Docker according to their distribution specific instructions:
+# https://docs.docker.com/engine/installation/ubuntulinux/
+apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D
```
james-w: :eyes: You could consider inlining the ascii-armoured key in to this script.

Keyservers aren't very reliable, and this is the most brittle part of the docker install
process on our Jenkins slave image.

It's only at packer time, rather than cloudformation time, but the key is static
data, so avoiding the request will make this process ~1% more reliable or
something.

Getting the key isn't the easiest thing, if you have it in a keyring then
 `gpg --armor --output docker.pub --export docker@docker.com` will do it.
`sudo apt-key adv --armor --output docker.pub --export docker@docker.com`
if you have an Ubuntu machine that has installed docker.

If you don't have it in a keying then
`gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D` (which can use a
throwaway keyring if you like) will get it.

The apt-key step would then me `apt-key adv --import <keyfile>` (keyfile can be `-` to
read from stdin)

james-w:
Looks good, thanks for the further improvements.

I have a suggestion about avoiding the keyserver, as our experience
with Jenkins suggests that it's the most brittle part of the docker install.

Merge when you are happy.

PR #2625, opened 2016-02-10
[FLOC-4145] Test type of JSON schema validation errors

jongiddy:
FLOC-3862 PR #2621 was an error in validation that occurred partly because the test that was supposed to fail due to the value being the wrong type actually failed due to the value not being a multiple of 1024.  Since the failures were not distinguished, the bug slipped through testing.

This PR separates the tests by expected validation error. This ensures that we are testing the correct failure reason. Using this new code, the FLOC-3862 test was evidently in the wrong section.

In addition, we now have some tests which actually have multiple validation errors, but with the new structure, we can see that they are failing for the desired reason. https://github.com/ClusterHQ/flocker/blob/81dfb8d68964608e11f188e5d45812ae29c18a30/flocker/control/test/test_schemas.py#L90-L97 is an example. The dataset_id UUID is invalid, which would give a `pattern` error, but it is clear we are failing for the desired reason: an additional `extra` property.

During the cleanup, a few other problems were fixed:
- An environment stanza could contain an empty variable name with any type.
- The `ConfigurationContainersSchemaTests` was redefined, meaning some tests were not being run.
- The `ConfigurationDatasetsUpdateSchemaTests` used a common stanza with other dataset commands, but for update they were all testing the same error (`additionalProperties`). This PR separates the update failures into a simpler set.
- The test that `ConfigurationDatasetsCreateSchemaTests` fails if `primary` is not present, was actually failing because the `dataset_id` UUID format was invalid.

Note, this PR broadly keeps the same tests, but in a different order (to group failures by validation error), and with some fixes. It does not add or modify any tests to improve completeness of the validation.  

There was only one change to the schema (to make "" an invalid environment variable name).

### DIFF
--- a/flocker/control/schema/types.yml
+++ b/flocker/control/schema/types.yml
@@ -150,6 +150,7 @@ definitions:
     patternProperties:
       "^.+$":
         type: string
+    additionalProperties: false
 
   volume:
     title: "A volume attached to a container"
--- a/flocker/control/test/test_schemas.py
+++ b/flocker/control/test/test_schemas.py
@@ -333,7 +333,7 @@ ConfigurationContainersSchemaTests = build_schema_test(
                 'node_uuid': a_uuid,
                 'image': 'postgres',
                 'name': 'postgres',
-                'volumes': [{'dataset_id': "x" * 36}],
+                'volumes': [{'dataset_id': a_uuid}],
             },
         ],
         INVALID_WRONG_TYPE: [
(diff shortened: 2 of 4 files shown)


james-w:
Hi, Thanks for improving our tests. I think this branch is good. The only way I can
think to improve it would be using references for the validation error types rather
than strings, if possible. Merge when happy.

PR #2626, opened 2016-02-10
[FLOC 4103] Remove all references to uft

davidoliverSP2:
Fixes 4103

Several commands were pre-fixed ``uft-``. This PR removes that. Additionally, there were 2 instances of recommending use of ``flocker-deploy``, which were removed.


james-w:
:+1: looks good, merge when happy.

PR #2627, opened 2016-02-10
[FLOC 3967] Move Appendix example to Acceptance Testing doc

davidoliverSP2:
Fixes 3967

The example in the Appendix was only be used in the Acceptance Testing/Configuration File documentation. This PR moves the example to the Acceptance Testing doc, and removes the Appendix.


james-w:
:+1: great, thanks for the improvement. Merge when happy.

PR #2629, opened 2016-02-10
[FLOC 3600] Add "Communicating with the API" section to API doc

davidoliverSP2:
Fixes 3600

@itamarst identified some key information missing from our API documentation. This PR adds it, in the same format as was originally documented in the issue (I thought this was the simplest way to present this information)

When reviewing please let me know if this could be expanded to help users understand this addition.

I also fixed the following header, which did not use title case.

### DIFF
--- a/docs/reference/api.rst
+++ b/docs/reference/api.rst
@@ -2,9 +2,16 @@
 REST API Endpoints
 ==================
 
+Communicating with the API
+==========================
+
+* The API is available on port 4523 of the control service, using HTTPS.
+* The client should validate the identity of the server using standard TLS validation, using :file:`cluster.crt` as the CA certificate.
+* The client should authenticate to the server by using the :file:`api.key` and :file:`api.crt` files.
+
 .. _conditional requests:
 
-Conditional requests
+Conditional Requests
 ====================
 When using the API to create, delete or otherwise modify datasets you may wish to do so only if certain conditions apply.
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks @davidoliverSP2 

The new bullet points are really useful. 

I have some further suggestions:
 * I think all the REST API docs should be on one page. So move all the REST API information that currently exists at the top of  http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/reference-APIinfo-FLOC-3600/reference/index.html#flocker-rest-api into http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/reference-APIinfo-FLOC-3600/reference/api.html
 * I think the new "communicating-with-the-api" section should include some examples of using ``curl`` to access the API.  Personally I think all the content from http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/reference-APIinfo-FLOC-3600/kubernetes-integration/generate-api-certificates.html?highlight=curl#using-an-api-certificate-to-authenticate should be moved here.
 * I've said it before, but I'd like "REST API" to be a top level menu item.

Answer or address or create followup issues for the points above and then merge.

davidoliverSP2:
Thanks @wallrj 

I agree with you with the API docs being all one page, and a top level menu item. I've created https://clusterhq.atlassian.net/browse/FLOC-4179

PR #2630, opened 2016-02-10
[FLOC-4167] build all AWS flocker jobs on medium instances 

mdevilliers:
Change all builds to use the same sized instance (medium) on AWS.

The belief is a homogeneous environment should give us
- greater reuse of provisioned slaves
- faster builds 

https://clusterhq.atlassian.net/browse/FLOC-4167

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -978,7 +978,7 @@ job_type:
   run_trial:
     # http://build.clusterhq.com/builders/flocker-centos-7
     run_trial_on_AWS_CentOS_7:
-      on_nodes_with_labels: 'aws-centos-7-SELinux-T2Small'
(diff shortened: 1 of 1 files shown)


sarum90:
Awesome.

Especially with how long the current time to first job run is on the small instances, I'm hopeful this will help a lot.  I'm a little bit curious about the larger world this lives in. Presumably the Mediums have more executors and that's taken into account when spinning up instances? I'm only asking because ideally we would only require N/X medium instances to do the work of N small instances in parallel. If we don't have the executors configured correctly, we might not see that improvement.

Either way, this change is obviously correct, and I believe it will make the world better by decreasing time to first test, Thanks!

mdevilliers:
looks like our jenkins instance has become sentinent....

Azulinho:
Haha lastpass being naughty again

PR #2632, opened 2016-02-10
[FLOC-4155] Cleanup leaked cloud instances from jenkins

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4155

I've ported the code from build.clusterhq.com with a few modifications:
 * I got rid of the ``deferToThread`` -- the operations are no longer asynchronous. @sarum90 and I discussed it and decided that it would be more straight forward and easier to maintain if the delete operations we're performed in sequence. In the future we could speed things up by running parallel cleanup operations for each cloud provider and each region. It's important that the nodes are destroyed before the volumes though.
 * I was tempted to rewrite it as Effects and performers but decided that that can be done in a followup issue.
 * I was also tempted to add Eliot logging to each of the actions, but  in the interests of keeping the branch small, I can do that in a follow up.
 * There are some Rackspace volumes which simply can't be deleted. I've reported those to Rackspace support and am waiting for them to fix their MySQL database for us (or whatever is involved).

I've temporarily configured Jenkins to use ``run-acceptance-tests --keep`` and manually forced a build on AWS and Rackspace.
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/cleanup_leaked_cloud_instances-FLOC-4155/job/_run_acceptance_on_AWS_CentOS_7_with_EBS/1/
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/cleanup_leaked_cloud_instances-FLOC-4155/job/_run_acceptance_on_Rackspace_Ubuntu_Trusty_with_Cinder/1/

The idea is that 2h later I can then force the cleanup_cloud_resources and see it clean up the nodes that were left behind...

...~2h later, I ran the script and it deleted the two left over AWS servers in the screenshot:
screenshot from 2016-02-10 17-51-02

The log of the cleanup is here:
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/cleanup_leaked_cloud_instances-FLOC-4155/job/_cleanup_cloud_resources/1/

### DIFF
--- a/admin/cleanup.py
+++ b/admin/cleanup.py
@@ -14,9 +14,9 @@ from characteristic import attributes
 from dateutil.parser import parse as parse_date
 from dateutil.tz import tzutc
 
+from libcloud.compute.base import NodeState, StorageVolume
 from libcloud.compute.providers import get_driver, Provider
 
-from twisted.internet.defer import succeed
 from twisted.python.filepath import FilePath
 from twisted.python.log import err
 from twisted.python.usage import Options, UsageError
@@ -146,8 +146,6 @@ class CleanVolumes(object):
         drivers = self._get_cloud_drivers(config)
         volumes = self._get_cloud_volumes(drivers)
         actions = self._filter_test_volumes(self.lag, volumes)
-        if not dry_run:
-            self._destroy_cloud_volumes(actions.destroy)
         return actions
 
     def _get_cloud_drivers(self, config):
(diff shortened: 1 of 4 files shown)


build.yaml
```diff
@@ -467,7 +467,11 @@ common_cli:
     # which are made available through a webserver running on port 80.
     # We pass the URL of our Jenkins Slave to the acceptance test nodes
     # through the --build-server parameter below.
+    #
+    # XXX Temporarily disable node cleanup so that there's something for the
```
sarum90: Now that you've proved that this works I think you mean to remove this line?

admin/cleanup.py
```diff
@@ -26,20 +26,12 @@
...
-    date_string = node.extra.get("created", node.extra.get("launch_time"))
-    if date_string is None:
-        return None
-    else:
-        return parse_date(date_string)
+# By default only nodes with names beginning with these prefixes will be
+# considered for cleanup.
+DEFAULT_NODE_NAME_PREFIXES = (
+    'acceptance-test-',
+    'client-test-'
```
sarum90: I guess I should probably rename "acceptance-test-DO-NOT-DELETE-MARCUS-PROD-SERVER"...

admin/cleanup.py
```diff
@@ -65,6 +57,22 @@ def get_ec2_driver(aws):
     return ec2
 
 
+def _get_node_creation_time(node):
+    """
+    Get the creation time of a libcloud node.
+
+    Rackspace and EC2 store the information in different metadeta.
```
sarum90: :question: Perhaps comment on which one is EC2 and which is Rackspace ("launch_time" vs "created")

admin/cleanup.py
```diff
@@ -230,22 +251,107 @@ def _filter_test_volumes(self, maximum_age, volumes):
...
-            try:
-                volume.destroy()
-            except:
-                err(None, "Destroying volume.")
+
+@attributes(['lag', 'prefixes'])
+class CleanAcceptanceNodes(object):
+    """
+    :ivar timedelta lag: The age of nodes to destroy.
+    :param prefixes: List of prefixes of nodes to destroy.
```
sarum90: nit: Prefer not to mix param/ivar.

admin/cleanup.py
```diff
@@ -230,22 +251,107 @@ def _filter_test_volumes(self, maximum_age, volumes):
...
+    def start(self, config):
+        # Get the libcloud drivers corresponding to the acceptance tests.
+        rackspace = get_rackspace_driver(config["rackspace"])
+        ec2 = get_ec2_driver(config["aws"])
+        drivers = [rackspace, ec2]
+
+        # Get the prefixes of the node names, appending the creator from
+        # the config.
+        creator = config['metadata']['creator']
+        prefixes = tuple(map(lambda prefix: prefix + creator, self.prefixes))
```
sarum90: :question: It'd be cool if this could somehow re-use the code from https://github.com/ClusterHQ/flocker/blob/master/admin/acceptance.py#L767. In general, I just have some concern about the fact that if I change that code, I wouldn't necessarily think to change this code.

That said, that's probably some sort of ideal world that we don't need to live in.

admin/cleanup.py
```diff
@@ -230,22 +251,107 @@ def _filter_test_volumes(self, maximum_age, volumes):
...
-        Unconditionally and irrevocably destroy all of the given cloud volumes.
-        """
-        for volume in volumes:
-            try:
-                volume.destroy()
-            except:
-                err(None, "Destroying volume.")
+
+@attributes(['lag', 'prefixes'])
+class CleanAcceptanceNodes(object):
```
sarum90: :question: Any reason not to use PClass?

admin/cleanup.py
```diff
@@ -230,22 +251,107 @@ def _filter_test_volumes(self, maximum_age, volumes):
...
+            #
+            # > The complete fix for this issue is expected in the next
+            # > Openstack iteration (mid August).  Until then what can be done
+            # > is just to issue another delete against the same node.  The
+            # > servers are only billed when they are in Active (green) status,
+            # > so the deleted nodes are not billed.
+            #
+            # So consider any nodes in that state as potential destruction
+            # targets.
+            and node.state in (NodeState.RUNNING, NodeState.TERMINATED)
```
sarum90: :question: Still a bit of a plug for abstracting this behind cloud-specific integrations. That way you could hide the ugly comment above in some specific rackspace implementation, rather than also doing it for EC2 (I think you are also doing it for EC2 here, which I guess isn't ideal?).

sarum90:
This looks great.

Any concern that it might delete "acceptance nodes" from benchmarking, or is that taken care of?

build.yaml
wallrj: Done.

admin/cleanup.py
wallrj: Probably, but the script will currently only cleanup AWS instances in the region where Jenkins runs acceptance tests.

admin/cleanup.py
wallrj: Done.

admin/cleanup.py
wallrj: Done.

admin/cleanup.py
wallrj: I added an XXX comment.

admin/cleanup.py
wallrj: ``PClass`` would be better, but I'm just copying code from Buildbot in this branch, not trying to make improvements.

admin/cleanup.py
wallrj: I added a big comment at the top of the file.

wallrj:
> Any concern that it might delete "acceptance nodes" from benchmarking, or is that taken care of?

I was concerned about that, but I heard today that the benchmark clusters are running in a separate AWS account. So I think we're safe.

Rackspace assure me that they've now deleted the broken Cinder volumes, so I'll see if the script now runs cleanly.

wallrj:
There is only one remaining error which is caused by a volume attached to one of @itamarst 's acceptance test nodes. The cleanup tool shouldn't attempt to delete such volumes. See: 
 * https://clusterhq.atlassian.net/browse/FLOC-4188

I'll merge this when the other tests have passed.

sarum90:
Nice, Thanks for entertaining my comments.

PR #2633, opened 2016-02-11
[FLOC-3922] Lint benchmark directory

jongiddy:
This PR adds the `benchmark` directory to the lint tests.

Code in the directory was changed to pass the lint tests.  Most were stylistic issues, but there was one bug where the variable `t0` was used in a closure within a loop, meaning that some elapsed time measurements may have been shorter than the actual time taken, probably by a fraction of a second.

### DIFF
--- a/benchmark/metrics/__init__.py
+++ b/benchmark/metrics/__init__.py
@@ -1,2 +1,7 @@
 from .cputime import CPUTime
 from .wallclock import WallClock
+
+__all__ = [
+    'CPUTime',
+    'WallClock',
+]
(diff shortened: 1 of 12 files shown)


james-w:
This should also change tox.ini, or delete the lint environment from there
(and probably add a script that developers can use to run lint checks
locally.)

james-w:
The changes in the current PR are fine though.

jongiddy:
Tests in Jenkins have passed, but GitHub is not showing the correct status. Merging based on direct view of Jenkins.

PR #2635, opened 2016-02-11
[FLOC-4177] Fix cputime error under load

jongiddy:
The CPU time command used for benchmarking requires that at least one of the named processes exist, in order to distinguish an error result from an empty result.

The code uses `ps` for this extra process name, due to the command actually calling `ps`.  However, it has been observed that under heavy load (which is often seen in the benchmarks), the short-lived `ps` process doesn't see itself.  This causes a `KeyError` when we try to remove the unwanted `ps` data.

This PR uses the name of the init process (process 1) on each node instead, which should be more robust.

### DIFF
--- a/benchmark/metrics/cputime.py
+++ b/benchmark/metrics/cputime.py
@@ -3,6 +3,8 @@
 CPU time metric for the control service benchmarks.
 """
 
+import subprocess
+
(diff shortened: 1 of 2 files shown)


james-w:
:+1: thanks for the fix.

PR #2638, opened 2016-02-12
[FLOC-4086] Warn users about needing access to S3 buckets during CloudFormation

myechuri:
If user's AWS account does not have a Policy that allows him/her to access S3 buckets, CloudFormation stack creation fails. Instead of wasting user's time and EC2 resources, it would be better to have user check a box during CloudFormation workflow confirming S3 access.

This PR also includes an unrelated change where user is allowed to choose from existing KeyPairs instead of typing in the name of the KeyPair.


wallrj:
Thanks @myechuri 

The new S3 prompt will hopefully make the user think.
It's a shame there isn't a way to test the user's access rights automatically.
I wonder if we could eventually distribute the certificates without an S3 bucket.

The new SSH Key selector is much nicer.

Please merge and update the live template.

myechuri:
Thanks, @wallrj ! Merged, updated and tested master template.

admin/installer/cloudformation.py
```diff
@@ -150,25 +150,32 @@ def _validate_cluster_size(size):
...
 ))
 keyname_param = template.add_parameter(Parameter(
     "EC2KeyPair",
-    Description="Name of an existing EC2 KeyPair to enable SSH "
-                "access to the instance (mandatory)",
+    Description="Required: Name of an existing EC2 KeyPair to enable SSH "
+                "access to the instance",
+    Type="AWS::EC2::KeyPair::KeyName",
+))
+secret_access_key_param = template.add_parameter(Parameter(
```
wallrj: Sorry, I missed this.
We're overwriting the `secret_access_key_param` variable.
Which means that the node cloud-init scripts get `secret_access_key="Yes"`.

We need to fix this and redeploy the cloudformation template.

wallrj:
@myechuri 

Sorry I missed it, but this should be reverted so that we can fix the problem described above.

I'll do that.

PR #2641, opened 2016-02-13
[FLOC-4086] Warn users about needing access to S3 buckets during CloudFormation

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4086
Based on #2638 with a bug fix.

If user's AWS account does not have a Policy that allows him/her to access S3 buckets, CloudFormation stack creation fails. Instead of wasting user's time and EC2 resources, it would be better to have user check a box during CloudFormation workflow confirming S3 access.

This PR also includes an unrelated change where user is allowed to choose from existing KeyPairs instead of typing in the name of the KeyPair.

### DIFF
--- a/admin/installer/cloudformation.py
+++ b/admin/installer/cloudformation.py
@@ -141,7 +141,7 @@ template = Template()
 # Keys corresponding to CloudFormation user Inputs.
 access_key_id_param = template.add_parameter(Parameter(
     "AmazonAccessKeyID",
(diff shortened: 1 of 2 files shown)


wallrj:
I uploaded template built from this branch:
 * https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.cloudformation-s3-access-FLOC-4086-2.json

PR #2643, opened 2016-02-15
[FLOC-4140] Handle Jenkins skipping 'builds' in the job dict.

james-w:
Also add some tests for the merge_pr module. They are far from
complete.

### DIFF
--- a/admin/merge_pr.py
+++ b/admin/merge_pr.py
@@ -164,9 +164,9 @@ def format_status(status):
     :return unicode: a string formatting of the pair
          for display to the user.
     """
-    GitHub, jenkins = status
-    return "{}: {} ({})".format(
-        GitHub['context'], jenkins.value, GitHub['target_url'])
+    github, jenkins = status
+    return u"{}: {} ({})".format(
+        github['context'], jenkins.value, github['target_url'])
 
 
 def infinite_sleeps(sleep_for):
@@ -236,6 +236,12 @@ class JenkinsResults(Values):
     PASSED = ValueConstant('success')
     FAILED = ValueConstant('failed')
     NOTRUN = ValueConstant('notrun')
+    # Sometimes Jenkins replies with not much info,
+    # e.g. no 'builds' in the result just after a
+    # build finished. We use this status if that
+    # happens so we can ignore the build for that
+    # cycle
+    UNKNOWN = ValueConstant('unknown')
 
 
 def jenkins_result_from_api(result):
@@ -279,6 +285,26 @@ def properties_to_params(props):
     return None
 
 
+def jenkins_info_from_response(project):
+    """
+    Get the Jenkins job information from the Jenkins API response.
+
+    :param dict project: the API response as a dict.
+    :return (JenkinsResults, dict): The first element will
+        be a JenkinsResults value corresponding to the status
+        of the job. The second element will be
+        the parameters to start a new Jenkins build if needed.
+    """
+    if project.get('inQueue', False):
+        return JenkinsResults.RUNNING, {}
+    if 'builds' not in project:
+        return JenkinsResults.UNKNOWN, {}
+    if len(project['builds']) < 1:
+        return JenkinsResults.NOTRUN, {}
+    result = jenkins_result_from_api(project['builds'][0]['result'])
+    return result, properties_to_params(project['property'])
+
+
 def get_jenkins_info(jenkins_session, status):
     """
     Get the Jenkins job info for a GitHub status, if any.
@@ -289,18 +315,13 @@ def get_jenkins_info(jenkins_session, status):
     :return (JenkinsResults, dict): The first element will
         be a JenkinsResults value if the information can
         be found, None if not. The second element will be
-        the paramters to start a new Jenkins build if needed.
+        the parameters to start a new Jenkins build if needed.
     """
     if status['context'].startswith('jenkins-'):
         jenkins_url = status['target_url']
         project = jenkins_session.get(
             jenkins_url + JENKINS_BUILD_INFO_PATH).json()
-        if project.get('inQueue', False):
-            return JenkinsResults.RUNNING, {}
-        if len(project['builds']) < 1:
-            return JenkinsResults.NOTRUN, {}
-        result = jenkins_result_from_api(project['builds'][0]['result'])
-        return result, properties_to_params(project['property'])
+        return jenkins_info_from_response(project)
     return None, None
 
 
--- a/admin/test/test_merge_pr.py
+++ b/admin/test/test_merge_pr.py
@@ -7,7 +7,7 @@ from datetime import datetime
 import os
 import subprocess
 
-from hypothesis import assume, given
+from hypothesis import given
 from hypothesis.strategies import (
     booleans,
     dictionaries,
@@ -19,7 +19,7 @@ from hypothesis.strategies import (
     sampled_from,
     text,
     )
-from pyrsistent import pmap
+from pyrsistent import pmap, plist
 
 from flocker.testtools import TestCase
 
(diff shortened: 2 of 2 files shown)


admin/test/test_merge_pr.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual({}, params)
+
+    @given(jenkins_build_results)
+    def test_no_builds(self, info):
+        assume(not info.get('inQueue', False))
+        info = dict(info)
+        info['builds'] = []
+        result, params = merge_pr.jenkins_info_from_response(info)
+        self.assertEqual(merge_pr.JenkinsResults.NOTRUN, result)
+        self.assertEqual({}, params)
```
james-w: I'm not particularly happy with these tests, but I think it's a reasonable approach
to get some better coverage than hardcoding a single example, and the intent of
the assume/mutation is fairy understandable.

admin/merge_pr.py
```diff
@@ -279,6 +285,26 @@ def transform(x):
     return None
 
 
+def jenkins_info_from_response(project):
+    """
+    Get the Jenkins job information from the Jenkins API response.
+
+    :param dict project: the API response as a dict.
+    :return (JenkinsResults, dict): The first element will
```
sarum90: :question: I'm not crazy about having tuples as return types (reading `maybe_retry_jobs`, specifically `lambda x: x[1][0] in [JenkinsResults.FAILED, JenkinsResults.NOTRUN]` was sort of tricky to figure out). But it seems that this fits the existing style and API, so probably not something to be fixed in this PR.

admin/test/test_merge_pr.py
```diff
@@ -0,0 +1,212 @@
...
+
+class StatusesTests(TestCase):
+    """
+    Tests for interpretation of commit statuses.
+
+    https://developer.github.com/v3/repos/statuses/
+    """
+
+    @given(commit_statuses)
+    def test_final_status_one(self, status):
```
sarum90: :question: nit: docstrings. These tests are pretty simple and straightforward, but flocker codebase is pretty uniform in having docstrings on all test methods:

```
 ♦ ~/dev/flocker (master)$ git grep -A 1 'def test_' | grep '"""' | wc
    2049    4308  103346
 ♦ ~/dev/flocker (master)$ git grep -A 1 'def test_' | grep 'def test_' | wc
    2085    6549  166892
```

So I guess only ~36/2085 tests are missing docstrings. It's not technically in our handbook, so ultimately your call if it's worth adding them.

admin/test/test_merge_pr.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual(target, merge_pr.final_status([status2, status1]))
+
+    @given(commit_statuses)
+    def test_not_success(self, status):
+        assume(status['state'] != u"success")
+        self.assertEqual(True, merge_pr.not_success(status))
+
+    @given(commit_statuses)
+    def test_not_success_success(self, status):
+        status['state'] = u"success"
```
sarum90: :question: I think it's slightly better to use a different strategy:
```
fixed_dictionaries(
     {'updated_at': datetimes,
      'state': just('success'),
      'context': text(),
      'target_url': text(),
      })
```
Not required for merge, but note that as the code is currently, this test might be run many times where hypothesis only varies the state key, which will be sort of lost testing.

Might consider turning commit_statuses into a function that lets you override the strategies:
```
def commit_statuses(**kwargs):
  return fixed_dictionaries(dict({
     'updated_at': datetimes,
      'state': text(),
      'context': text(),
      'target_url': text(),
      }, **kwargs))
```

admin/test/test_merge_pr.py
```diff
@@ -0,0 +1,212 @@
...
+        self.assertEqual(merge_pr.JenkinsResults.RUNNING, result)
+        self.assertEqual({}, params)
+
+    @given(jenkins_build_results)
+    def test_builds_not_present(self, info):
+        assume(not info.get('inQueue', False))
+        info = dict(info)
+        if 'builds' in info:
+            del info['builds']
+        result, params = merge_pr.jenkins_info_from_response(info)
```
sarum90: :question: This test looks really weird, The assume will through out ~half of the examples which isn't great, and then after mutations I think you are left with something that looks like:
```
one_of(
  just(pmap()),
  just(pmap(inQueue=False)),
  fixed_dictionaries(dict(inQueue=just(False), property=dictionaries(text(), text())))
```

Hypothesis will still try to come up with a bunch of different `builds=lists(jenkins_builds)`.

I guess this is starting to get borderline with me as far as if this really gets that much benefit from using hypothesis, especially when combined with the assume's and the mutations causing hypothesis to generate a bunch of examples that won't even expand test coverage.

admin/test/test_merge_pr.py
sarum90: Hrm... I actually struggle with this one as far as deciding if hypothesis adds test coverage or just complexity over single example tests. It seems after `assume`s and mutations, hypothesis is usually left to fiddle with keys on the dictionaries that obviously have not bearing on the output.

It's tempting to have just a single test that asserts different things depending on the example hypothesis generates, but then you are just re-implementing your production code in your test... And you would still sort of want the single example style test to verify the logic in your test code.

So there are 2 concerns with using hypothesis in my mind, 1 is code complexity and the other is test runtime. Obviously these have to be balanced against the enhanced code coverage. You have some nice clean strategies here, so I think the complexity is well within our budget, but I'm a little concerned about test time (especially with use of assume and the mutations).

I noticed running this class of test suites takes significantly longer than the rest of the file. 4 seconds isn't great for testing pure-functional code. Perhaps consider trying to get that down? Simplifying the strategy for the property dict I think has the best bang for your buck, something like `property=dictionaries(text(max_size=2), text(max_size=2), max_size=2)` runs a lot faster, and I don't think it cheapens your test that much if at all.

```
flocker-review ♦ ~/dev/flocker-review (merge-pr-no-builds-FLOC-4140)$ trial admin.test.test_merge_pr.JenkinsResultsTests                                                                                    
admin.test.test_merge_pr
  JenkinsResultsTests
    test_builds_not_present ...                                            [OK]
    test_in_queue ...                                                      [OK]
    test_no_builds ...                                                     [OK]
    test_result_types ...                                                  [OK]

Ran 4 tests in 3.838s

PASSED (successes=4)
flocker-review ♦ ~/dev/flocker-review (merge-pr-no-builds-FLOC-4140)$ trial admin.test.test_merge_pr.{SmokeTests,StatusesTests,URLTests}
admin.test.test_merge_pr
  SmokeTests
    test_help ...                                                          [OK]
  StatusesTests
    test_final_status_many ...                                             [OK]
    test_final_status_one ...                                              [OK]
    test_format_status ...                                                 [OK]
    test_not_success ...                                                   [OK]
    test_not_success_success ...                                           [OK]
  URLTests
    test_pr_api_url ...                                                    [OK]
    test_url_path ...                                                      [OK]
    test_url_path_no_hostname ...                                          [OK]
    test_url_path_parts ...                                                [OK]
    test_url_path_parts_no_hostname ...                                    [OK]

Ran 11 tests in 0.610s

PASSED (successes=11)
```

sarum90:
Looks great!

A couple recommendations so that hypothesis doesn't slow your tests down like crazy. Nice to see this script getting hardened over time.

admin/merge_pr.py
james-w: Yeah, it's not that nice, I agree.

The reason it's like this is for that filter() that you mention, so that we can get the
information, then filter it on parts of that info, but use other parts.

Better schemes might be to build a dict of all the info, then filter the keys. The
functions that use the results can then just do a dict lookup for what they
need.

Alternatively I could create some object that would hold all the info with
names rather than positions, perhaps a namedtuple or a class.

james-w:
Hi Marcus, thanks for the review, good points all around. Please take a look at
the updated version and see what you think. Functions that take
arguments to override the strategies are a nice solution I think, though
the implementation of the one for the Jenkins results is a bit hairy.

admin/merge_pr.py
sarum90: Yea, since working here I've started just writing 2 attribute PClasses for things like this, but our requirements for docstring sometimes make that have more overhead than it probably should.

sarum90:
Tests look great!

I'm much more confident these are not going to take a long time, and much happier with strategy injection over mutation. With more complicated hypothesis tests I think you'll also get more bang for your buck if you can express strategies that let hypothesis concentrate fuzzing on the important parts of your inputs, so I think it'a  good policy to have in general.

:+1:

PR #2644, opened 2016-02-16
Benchmark back off when not converged floc 4059

mpgomez:
When waiting for convergence, we were looping really fast (every 0.1 seconds) checking the state.
It was a problem for bulk creation and deletion of datasets, as we were overloading AWS.

I have added an exponential backoff. I may make sense to retry fast the first one or two times, but if it is saturated, it is better to loop slower.

It improves the issues for bulk creation and deletion (tried with 28 node cluster with 20 containers each, only 3 failures; we were having ~150 failures)


flocker/node/test/test_loop.py
```diff
@@ -58,6 +58,7 @@
 NO_OP = NoOp(sleep=timedelta(seconds=300))
 
 
+
```
zubron: This extra blank line is causing a lint failure.

james-w:
Hi,

I think the code is good. It could perhaps benefit from a couple
of direct tests of the new Class.

I'm not sure if this is the best approach to this problem (which we
should definitely tackle, I noticed this problem myself a few days
ago, the rate at which we poll AWS will surely lead to RequestLimitExceeded)

This approach will slow down the whole convergence loop if one
volume takes a while to appear. That might not be desired behaviour.
A couple of alternatives:

  * Have the convergence loop run at the current speed, but don't query every iteration. Some sort of stateful layer that caches results and only updates at a slower frequency would work for this, but with some new problems.

  * Itamar mentioned per-dataset convergence loops. That would allow for basically this code to work exactly how we want, but requires converting to that scheme, which would probably be a significant chunk of work. It also makes batch requests to the backend harder. With the current scheme making 1 describe volumes call for all oustanding volumes would be straightforward. Per-dataset loops would make it much harder.

Overall I think this is a problem we should fix, as it will be a major problem for scalability. This seems like a perfectly reasonable way to do that for little cost, without ruling out other solutions in the future. However I may be overlooking a major downside. I propose that we merge this code, but I would like other opinions.

I'm happy to make any changes needed for this branch if Pilar would like.

Thanks,

James

flocker/node/test/test_loop.py
```diff
@@ -538,8 +539,8 @@ def test_convergence_sent_state_fail_resends_alternating(self):
             client=client, configuration=configuration, state=state))
 
         # Wait for all three iterations to occur.
-        reactor.advance(_UNCONVERGED_DELAY.delay_seconds)
-        reactor.advance(_UNCONVERGED_DELAY.delay_seconds)
+        reactor.advance(_UNCONVERGED_DELAY)
+        reactor.advance(_UNCONVERGED_DELAY*2)
```
zubron: If I'm reading this test correctly, the reactor should now advance by the value provided by `UnconvergedDelay.sleep()` which in this case would be `_UNCONVERGED_DELAY * 4`. It would probably be better to have the factor that the delay increases by as a constant.

sarum90: +1

flocker/node/_loop.py
```diff
@@ -501,15 +523,23 @@ def got_local_state(local_state):
...
             )
             if isinstance(action, NoOp):
+                # If we have converged, we need to reset the sleep delay
+                # in case there were any incremental back offs while
+                # waiting to converge.
+                self._unconverged_sleep.reset_delay()
                 # We've converged, we can sleep for NoOp's sleep duration.
                 # We add some jitter so not all agents wake up at exactly
                 # the same time, to reduce load on system:
+                self._unconverged_sleep.reset_delay()
```
sarum90: nit: I don't think there is benefit in calling this twice, I suspsect some sort of merge error.

flocker/node/_loop.py
```diff
@@ -532,8 +562,8 @@ def got_local_state(local_state):
         # converging again; hopefully next time we'll have more success.
         def error(failure):
             writeFailure(failure, self.fsm.logger)
-            # We should retry quickly to redo the failed work:
-            return _UNCONVERGED_DELAY
+            # We should retry to redo the failed work:
+            return self._unconverged_sleep.sleep()
```
sarum90: +1 to this. This part of the PR absolutely should be merged.

flocker/node/_loop.py
```diff
@@ -501,15 +523,23 @@ def got_local_state(local_state):
...
                     action.sleep.total_seconds())
             else:
                 # We're going to do some work, we should do another
-                # iteration quickly in case there's followup work:
-                sleep_duration = _UNCONVERGED_DELAY
+                # iteration, but chances are that if, for any reason,
+                # the backend is saturated, by looping too fast, we
+                # will only make things worse, so there is an incremental
+                # back off in the sleep interval.
+                sleep_duration = self._unconverged_sleep.sleep()
```
sarum90: :grey_question: I'm 50/50 on this. There is definitely something smarter we could do here.

The bad: It's only 6 hits for this to max out at the 400 seconds. I guess I don't have a good feel for how often this will or won't occur, It seems like it'd be kind of bad if there is a 400 second wait before any volume could be attached. Could we add logging and then run the benchmarking cluster with this to gather more data? A new blockdevice goes from non-existing->CREATE->non-manifest->ATTACH->attached_no_fs->CREATE_FS->attached->MOUNT->mounted. I think that means we're already at waiting 25 seconds if another configuration change comes sometime during that process.

The good: This does put a limit at how quickly we will call discover_state (and thus backend list volume calls). It's a shame for a single blockdevice between CREATE_FS and MOUNT we can't do a cheaper discover_state that just dos lsblkid, and not querying the cloud.

Hrm... I think my main concerns could probably be addressed by seeing the results of some benchmarking runs on this code.

sarum90:
Just hopped in here since I noticed this has been stuck in needs review for awhile.

The exponential back-off on the error conditions I absolutely agree with, but the exponential backoff when we are doing real work I have a bit more trouble with... We won't schedule the next loop of convergence until the previous round is done anyway (and that involves waiting for AWS to create volumes, or creating filesysyems, etc.) and it is sort of concerning to me that in the happy case we might have to keep waiting longer and longer.

For instance I did some log grepping on some of the acceptance tests run from this PR and confirmed that we waiting 1.6 seconds from create_fs to mount. That's not terrible, but it would be pretty bad if we added two volumes in series and that got up to ~20 seconds, or 3 in series and got that up to a couple hundred seconds.

I think it's possible that this should still go in with a few caveats:

- I think the maximum wait should be less for the happy case (line 542 should probably cap out at O(10 seconds) in my opinion) (alternatively, if we already have benchmarking results for time-to-converge, and can show this does not effect that negatively even when we get to the 400 second wait situation might be fine without that). Although... if that doesn't fix the original problem, that'd be a shame.
- I think we should probably add a few tests for the new class
- I think we should improve logging so that it's easier to tell from the logs when we are in an exponential back-off situation.

@james-w @mpgomez @zubron Does it seem reasonable to do those three things before merge? For now I'll put back in code-me as I don't think it's clear who will be taking the cleanup coding task.

james-w:
I made some adjustments to this branch, as well as some general cleanups/tests/docstrings.

The big change is that the sleep maxes out at 10 seconds now. This isn't as good for
the error case, particularly backoff when getting RequestLimitExceded. However,
I believe that reducing the polling of AWS while waiting for volumes will
mean we see that much less often. Obviously some data to back this up
would be preferable, but I'm not really sure how to reproduce what the
benchmark team were seeing with this new code.

james-w:
It sounds like Pilar was testing this branch by trying to start a large cluster using the benchmarking
tools. Doing that would probably be a good test of the new numbers in this branch, and perhaps
suggest whether different maximums for error/success cases would be useful.

sarum90:
Thanks for the commits James. I'll take reviewing, and I'll try to figure out how to actually run the benchmarking cluster to verify this still fixes the issue they were seeing.

brendan-donegan:
You can ask me or Bridget if you need help with that

On Wed, 24 Feb 2016, 00:38 Marcus Ewert <notifications@github.com> wrote:

> Thanks for the commits James. I'll take reviewing, and I'll try to figure
> out how to actually run the benchmarking cluster to verify this still fixes
> the issue they were seeing.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ClusterHQ/flocker/pull/2644#issuecomment-187988973>.
>
-- 

Brendan Donegan - 07771376827
Lead QA Engineer @ ClusterHQ

sarum90:
1) Got myself some shiny new credentials to spin up a cluster in the benchmarking account.
2) Made an acceptance.yaml for the new credentials.
3) Kicked off http://build.clusterhq.com/builders/flocker-omnibus-ubuntu-14.04/builds/8173 to rebuild the packages so I can actually build a cluster using the code in this branch.

sarum90:
Thanks @brendan-donegan! So far so good. I did enough code reviews of that code that I can kinda fake my way through, with help from http://doc-dev.clusterhq.com/gettinginvolved/benchmarking.html#benchmarking-configuration-file where I need it.

sarum90:
Hrm.. a fake run of run-acceptance-tests resulted in `exceptions.OSError: [Errno 24] Too many open files` on my Mac... so... we must be using something else to spin up 28 nodes....

sarum90:
Well, that failed quicker, but I still hit `exceptions.OSError: [Errno 24] Too many open files`. So I guess I'll bump up my file limits. Although that might wait until tomorrow as I think I'm done for the day.

brendan-donegan:
Yes, there is setup-cluster, this is what we were normally using

On Wed, 24 Feb 2016, 01:45 Marcus Ewert <notifications@github.com> wrote:

> Well, that failed quicker, but I still hit exceptions.OSError: [Errno 24]
> Too many open files. So I guess I'll bump up my file limits. Although
> that might wait until tomorrow as I think I'm done for the day.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ClusterHQ/flocker/pull/2644#issuecomment-188004130>.
>
-- 

Brendan Donegan - 07771376827
Lead QA Engineer @ ClusterHQ

zubron:
Ah, you got in there just before me @brendan-donegan! @sarum90, I will try to run the benchmarking setup scripts today to test this change and we can chat later about it. Thanks!

zubron:
After upping my file limits to avoid the issue that @sarum90 hit, I successfully deployed a cluster of 28 nodes, each with 20 stateful containers, with no errors!
`Started 560 containers with 0 failures.`
Looking good! :smile:

sarum90:
Code looks great, I think I'm happy with the test coverage too. It sort of would be nice to add logging that we are waiting, but not as important now that we cap out at 10 seconds.

Because this seems like a pretty significant change, I just clicked the "run acceptance tests on AWS" buttons in Jenkins (http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/benchmark-back-off-when-not-converged-FLOC-4059/). I'll keep an eye on them today, if they are green I'll merge.

sarum90:
Also thanks @zubron for running the test!

sarum90:
Looks like everything is green, including acceptance tests, so I'll go ahead and merge.

PR #2645, opened 2016-02-17
[FLOC-4190] Fix a bug in the initialize-release tool

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4190

### DIFF
--- a/admin/functional/test_release.py
+++ b/admin/functional/test_release.py
@@ -4,6 +4,7 @@
 Tests for ``admin.release``.
 """
 
+from git import Repo
 from testtools.matchers import Contains, Equals
 from twisted.python.filepath import FilePath
 from flocker.testtools import TestCase, run_process
@@ -65,12 +66,25 @@ class InitializeReleaseScriptTests(TestCase):
 
     def test_run(self):
         """
-        The script can be run.
+        The script creates a new release version branch in a new working
+        directory and prints some shell commands to ``stdout``.
         """
+        expected_version = "9.9.9"
         script_path = self.make_temporary_directory()
+        repo_path = script_path.sibling(
+            'flocker-release-{}'.format(expected_version)
+        )
         result = run_process(
-            [INITIALIZE_RELEASE_PATH.path, '--flocker-version=0.0.0'],
+            [INITIALIZE_RELEASE_PATH.path,
+             '--flocker-version={}'.format(expected_version)],
             cwd=script_path.path
         )
         self.expectThat(result.status, Equals(0))
-        self.expectThat(script_path.children(), Equals(['foo']))
+        self.expectThat(
+            result.output,
+            Contains('export VERSION={}'.format(expected_version))
+        )
+        self.expectThat(
+            Repo(path=repo_path.path).active_branch.name,
+            Equals('release/flocker-{}'.format(expected_version))
+        )
--- a/admin/functional/test_release.py
+++ b/admin/functional/test_release.py
@@ -1,0 +1,76 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Tests for ``admin.release``.
+"""
+
+from testtools.matchers import Contains, Equals
+from twisted.python.filepath import FilePath
+from flocker.testtools import TestCase, run_process
+
+
+def repository_path():
+    """
+    :return: The nearest ``FilePath`` in parents of this module that have a
+        ``flocker/__init__.py`` file that matches what will be imported as
+        ``flocker``.
+    """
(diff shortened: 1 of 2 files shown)


admin/release.py
```diff
@@ -1,4 +1,4 @@
-# -*- test-case-name: admin.test.test_release -*-
+# -*- test-case-name: admin.test.test_release,admin.functional.test_release -*-  # noqa
```
sarum90: :question: I don't have emacs or emacs plugins that parse this, so I'll have to trust your that this is the right form. (Just fyi, I'm sure it's right, but I can't verify)

admin/release.py
```diff
@@ -879,7 +875,7 @@ def initialize_release(version, path, top_level):
         os.environ["LDFLAGS"] = '-L{}/lib" CFLAGS="-I{}/include'.format(
             brew_openssl, brew_openssl)
     check_call(
-        ["pip install -e .[dev]"], shell=True,
+        ["pip", "install", "-e" ".[dev]"],
```
sarum90: missing comma after -e I think? Might still work cause single letter args sometimes parse all characters after the first as the arg parameter. If that's the case and it is intentional, prefer to delete the space between the strings.

admin/functional/test_release.py
```diff
@@ -0,0 +1,90 @@
...
+        self.expectThat(result.status, Equals(0))
+        self.expectThat(result.output, Contains('Usage: initialize-release'))
+
+    def test_run(self):
+        """
+        The script creates a new release version branch in a new working
+        directory and prints some shell commands to ``stdout``.
+        """
+        expected_version = "9.9.9"
+        script_path = self.make_temporary_directory()
```
sarum90: :question: Just FYI I don't think this cleans up the directory, just puts it in the temp folder. probably okay, but thought you should know.

admin/functional/test_release.py
```diff
@@ -0,0 +1,90 @@
...
+        repo_path = script_path.sibling(
+            'flocker-release-{}'.format(expected_version)
+        )
+        result = run_process(
+            [INITIALIZE_RELEASE_PATH.path,
```
sarum90: :question: I don't remember off hand and should be getting ready to go here pretty soon. Does this also make the virtualenv for you? If so we might want to add cleanup to this test. If not, fine as is.

sarum90:
Looks good, thanks for maintaining these scripts.

One real comment to address before merging.

tomprince:
I'd strongly suggest, instead of fixing the script, deleting it, and replacing the instructions to use it with prose instructions to create a new work-copy and virtualenv.

tomprince:
To expand on that:
- less code to maintain
- when releases get automated, the jenkins abstraction around getting a clean environment will be used, rather than this.

admin/release.py
wallrj: Fair point. I tested it in emacs, but this is also how `trial  --testmodule=admin/release.py` works. We use that feature for running our flocker.node.agent functional tests.

admin/release.py
wallrj: Fixed.

admin/functional/test_release.py
wallrj: That's what I want. So I can see what the script created if the test fails.

admin/functional/test_release.py
wallrj: It does create a virtualenv, but inside the newly cloned working copy.
```
ls -la _trial_temp/admin.functional.test_release/InitializeReleaseScriptTests/test_run/flocker-release-9.9.9/flocker-9.9.9/bin/activate
```

So no need to clean up.

wallrj:
> I'd strongly suggest, instead of fixing the script, deleting it, and replacing the instructions to use it with prose instructions to create a new work-copy and virtualenv.

I kind of agree. But I notice that the command also handles some stuff to do with installing OpenSSL on the OSX platform. I don't want to have to document that right now.
This command obviously worked for @zubron  during the last few releases, so I'll leave it for now and it can be deleted if / when we do fully automate the release process.

zubron:
@wallrj: I ended up only doing one release which used this script. I fixed one bug in it during that release [FLOC-3842](https://clusterhq.atlassian.net/browse/FLOC-3842) and then I encountered another bug that didn't affect me until later in the process. See [FLOC-3918](https://clusterhq.atlassian.net/browse/FLOC-3918).

wallrj:
@zubron Oh, right :frowning: so this script hasn't got a great track record then. Perhaps next release I'll take another look and see if it can be deleted and documented instead. (as @tomprince suggested)

PR #2648, opened 2016-02-19
[FLOC-4194] Add GCE functional test that verifies the attached to other node exception.

sarum90:
Refactors the provisioner so that it can be used by code that does not need to log into the instance. Adds a new test to the GCE functional tests that actually spins up a second instance and attaches a volume to that instance just to verify the exception raised by the BlockDeviceAPI.

There are other tests that will also use this infrastructure forthcoming.

### DIFF
--- a/flocker/node/agents/gce.py
+++ b/flocker/node/agents/gce.py
@@ -200,6 +200,31 @@ def _get_metadata_path(path):
     return r.text
 
 
+def get_machine_zone():
+    """
+    Returns the zone that the GCE instance running this code is running within.
+
+    It does this by querying the metadata server.
+
+    :returns unicode: The zone where the curring instance is running.
+    """
+    # Transform "projects/<project-number>/zones/us-central1-f" to
+    # "us-central1-f".
+    return _get_metadata_path('instance/zone').split('/')[-1]
+
+
+def get_machine_project():
+    """
+    Returns the project that the GCE instance running this code is running
+    within.
+
+    It does this by querying the metadata server.
+
+    :returns unicode: The project under which the curring instance is running.
+    """
+    return _get_metadata_path('project/project-id')
+
+
 # The prefix added to dataset_ids to turn them into blockdevice_ids.
 _PREFIX = 'flocker-v1-'
 
--- a/flocker/node/agents/gce.py
+++ b/flocker/node/agents/gce.py
@@ -206,7 +206,7 @@ def get_machine_zone():
 
     It does this by querying the metadata server.
 
-    :returns unicode: The zone where the curring instance is running.
+    :returns unicode: The zone where the current instance is running.
     """
     # Transform "projects/<project-number>/zones/us-central1-f" to
     # "us-central1-f".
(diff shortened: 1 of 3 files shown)


flocker/node/agents/gce.py
```diff
@@ -200,6 +200,31 @@ def _get_metadata_path(path):
     return r.text
 
 
+def get_machine_zone():
+    """
+    Returns the zone that the GCE instance running this code is running within.
+
+    It does this by querying the metadata server.
+
+    :returns unicode: The zone where the curring instance is running.
```
zubron: `current` rather than `curring`?

flocker/node/agents/gce.py
```diff
@@ -200,6 +200,31 @@ def _get_metadata_path(path):
...
+
+
+def get_machine_project():
+    """
+    Returns the project that the GCE instance running this code is running
+    within.
+
+    It does this by querying the metadata server.
+
+    :returns unicode: The project under which the curring instance is running.
```
zubron: Same as above.

flocker/provision/_gce.py
```diff
@@ -272,32 +297,71 @@ def _create_gce_instance_config(instance_name, project, zone, machine_type,
...
+        with start_action(
+            action_type=u"flocker:provision:gce:destroy",
+            instance_id=self.name,
+        ):
+            operation = self.compute.instances().delete(
+                project=self.project,
+                zone=self.zone,
+                instance=self.name
+            ).execute()
+            wait_for_operation(self.compute, operation, timeout_steps=[5]*60)
```
zubron: :question: Any particular reason for changing the timeout values?

flocker/provision/_gce.py
```diff
@@ -364,25 +419,86 @@ def reboot(self):
...
 
+    def create_instance(self, instance_name, image=None, **kwargs):
+        """
+        Create a GCE instance.
+
+        :param unicode instance: The name of the instance to create.
+        :param unicode image: The link to the GCE image to use for the base
+            disk. Defaults to the latest CentOS 7 image if unspecified.
+        :param **kwargs: Other keyword arguments for the creation of the
+            instance see documentation of ``_create_gce_instance_config``
```
zubron: Use two separate sentences here. `...instance. See documentation...`

zubron:
Thanks @sarum90. This looks good. Please address the comments and merge when the tests are green.

flocker/node/agents/gce.py
sarum90: Thanks, fixed.

flocker/node/agents/gce.py
sarum90: Thanks. I even had spell check on... but alas: http://dictionary.reference.com/browse/curring

flocker/provision/_gce.py
sarum90: Nice catch, you're keeping me honest :).

Yea I did see it take somewhere between 3 and 4 minutes once when I ran the test.

As this is test infra code, the right answer seemed to be just wait up to 5 minutes. I also figured if it was going to take that long we probably don't need to poll every second.

Added a comment so people don't think about making it shorter.

flocker/provision/_gce.py
sarum90: Done.

sarum90:
Thanks for the review @zubron! I realized I should run the GCE acceptance tests as well to verify I didn't break them, so I'm going to do that before I merge. Otherwise the changes are all comment changes, so I might just wait for run_lint.

PR #2649, opened 2016-02-19
[FLOC-4197] Remote test-redirects and update documentation redirect documentation

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4197

Since #2503 we no longer add `latest` and ` devel` 302 redirects. 

Instead we now copythe released documentation to a version key and to the `latest` or `devel` key, replacing the content that was there from the last release.

This means there's very little point in the `test-redirects` script which only tested the `latest` and `devel` redirects.

I considered updating the tool to check the redirects that are generated from `docs/redirects.yaml` but that seems like a different piece of work which can be done in another branch if we think it's important.

I'll follow this with https://clusterhq.atlassian.net/browse/FLOC-4199 which will add some version checking to `admin/publish-docs`` .


admin/testbrew.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+/#!/usr/bin/env python
```
zubron: I'm guessing this was an accidental commit but please remove the `/`.

zubron:
Thanks @wallrj. Please address the one comment above and merge when the lint test is green.

PR #2651, opened 2016-02-20
[FLOC-4200] Retry scp operations in test_installer

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4200

* Added a new ``upload`` function based on the existing ``flocker.common.runner.download`` 
* Refactored that function to share a common scp function.
* Added a new error case for parsing SCPs `lost connection` error message into an Exception that we can catch from the retry helpers.

### DIFF
--- a/flocker/common/functional/test_runner.py
+++ b/flocker/common/functional/test_runner.py
@@ -15,7 +15,13 @@ from flocker.testtools import AsyncTestCase, random_name
 
 
 class UploadTests(AsyncTestCase):
+    """
+    Tests for ``flocker.common.runner.upload``.
+    """
     def test_upload_file(self):
+        """
+        A single file can be uploaded.
+        """
         expected_content = random_name(self)
         local_path = self.make_temporary_file()
         local_path.setContent(expected_content)
@@ -25,6 +31,9 @@ class UploadTests(AsyncTestCase):
         )
 
     def test_upload_directory(self):
+        """
+        A directory can be uploaded recursively.
+        """
         local_path = self.make_temporary_directory()
         local_path.child('child1').open('w')
         return self.assert_upload(
@@ -33,6 +42,15 @@ class UploadTests(AsyncTestCase):
         )
 
     def assert_upload(self, local_path, matcher):
+        """
+        Assert that the ``local_path`` can be uploaded to and then downloaded
+        from a remote SSH server and that the contents are preserved.
+
+        :param FilePath local_path: The local file or directory to upload.
+        :param matcher: A ``testtools`` matcher which will be compared to the
+            downloaded ``FilePath.path``.
+        :returns: A ``Deferred`` that fires when the assertion is complete.
+        """
         self.ssh_server = create_ssh_server(
             base_path=self.make_temporary_directory()
         )
--- a/flocker/common/functional/test_runner.py
+++ b/flocker/common/functional/test_runner.py
@@ -65,12 +65,11 @@ class UploadTests(AsyncTestCase):
                 local_path=download_path,
                 port=self.ssh_server.port,
                 identity_file=self.ssh_server.key_path,
-                recursive=True,
             )
         d.addCallback(download)
 
         def check(ignored):
             self.assertThat(download_path.path, matcher)
-        d.addBoth(check)
+        d.addCallback(check)
 
         return d
--- a/flocker/common/functional/test_runner.py
+++ b/flocker/common/functional/test_runner.py
@@ -10,7 +10,7 @@ from twisted.internet import reactor
 from flocker.testtools.ssh import (
     create_ssh_server
 )
-from flocker.common.runner import download_file, upload
+from flocker.common.runner import download, upload
 from flocker.testtools import AsyncTestCase, random_name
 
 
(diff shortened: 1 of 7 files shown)


wallrj:
Hmm. I seem to have made things worse. The scp process now  fails every time.

http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/less-flaky-test-installer-FLOC-4200/job/_run_admin_test_test_installer/3/console

```
20:26:06 
20:26:06 c7a1ceed-857c-4a89-a5f8-fa06497e77b1 -> /2/81
20:26:06 2016-02-20 20:19:19.035518Z
20:26:06   message_type: 'flocker.common.runner:scp:stderr'
20:26:06   line: 'debug2: we did not send a packet, disable method\r'
20:26:06 
20:26:06 c7a1ceed-857c-4a89-a5f8-fa06497e77b1 -> /2/82
20:26:06 2016-02-20 20:19:19.035942Z
20:26:06   message_type: 'flocker.common.runner:scp:stderr'
20:26:06   line: 'debug1: No more authentication methods to try.\r'
20:26:06 
20:26:06 c7a1ceed-857c-4a89-a5f8-fa06497e77b1 -> /2/83
20:26:06 2016-02-20 20:19:19.036345Z
20:26:06   message_type: 'flocker.common.runner:scp:stderr'
20:26:06   line: 'Permission denied (publickey).\r'
20:26:06 
20:26:06 c7a1ceed-857c-4a89-a5f8-fa06497e77b1 -> /2/84
20:26:06 2016-02-20 20:19:19.036842Z
20:26:06   message_type: 'flocker.common.runner:scp:stderr'
20:26:06   line: 'lost connection'
20:26:06 
20:26:06 c7a1ceed-857c-4a89-a5f8-fa06497e77b1 -> /2/2/2
20:26:06 2016-02-20 20:19:19.040130Z
20:26:06   action_type: 'flocker.common.runner:run'
20:26:06   action_status: 'failed'
20:26:06   exception: 'twisted.internet.error.ProcessTerminated'
20:26:06   reason: 'A process has ended with a probable error condition: process ended with exit code 1.'
```

I wonder if an ssh-agent is running and whether the vagrant-buildbot key is loaded.

flocker/common/runner.py
```diff
@@ -256,7 +340,59 @@ def scp_failed(reason):
...
+        return context.addActionFinish()
+
+
+def download(reactor, username, host, remote_path, local_path,
+             port=22, identity_file=None):
+    """
+    Run the local ``scp`` command to download a file or directory from a remote
+    host and kill it if the reactor stops.
+
+    See ``scp`` for parameter and return type documentation.
```
sarum90: :question: Could you have used kwargs here? seems like it might be less maintenance burden.

flocker/common/runner.py
```diff
@@ -209,45 +250,88 @@ def run_ssh(reactor, username, host, command, **kwargs):
...
+        port=22, identity_file=None, direction=DOWNLOAD):
+    """
+    :param reactor: A ``twisted.internet.reactor``.
+    :param bytes username: The SSH username.
+    :param bytes host: The SSH host.
+    :param FilePath remote_path: The path to the remote file.
+    :param FilePath local_path: The path to the local file.
+    :param int port: The SSH TCP port.
+    :param FilePath identity_file: The path to an SSH private key.
+    :param direction: One of ``DOWNLOAD`` or ``UPLOAD``.
```
sarum90: :question: prefer no default direction. Defaulting to either and having the user expect the wrong one could result in them overwriting a file they wanted to copy elsewhere. This seems precisely like the sort of file you don't want overwritten.

flocker/common/functional/test_runner.py
```diff
@@ -0,0 +1,93 @@
...
+from twisted.internet import reactor
+
+from flocker.testtools.ssh import (
+    create_ssh_server
+)
+from flocker.common.runner import download, upload
+from flocker.testtools import AsyncTestCase, random_name
+
+
+class UploadTests(AsyncTestCase):
```
sarum90: :question: One could imagine that we should also have download tests, and tests for logging... but... as this is currently just a test util, that is probably overkill.

sarum90:
Looks great. Feel free to merge so long as you have a passing test.

Perhaps remove the default on direction, but my other comments are more passing observations than anything else.

flocker/common/runner.py
wallrj: Yeah, ``kwargs`` is better. And it forces the caller to supply key word arguments for clarity.
I was originally trying to keep the old ``download_file`` function whose argument names didn't match up.

Done.

flocker/common/runner.py
wallrj: Good point. Done.

flocker/common/functional/test_runner.py
wallrj: Well the test case does exercise ``download`` too. But yeah, ideally there'd be some more tests here. I'll leave it for now though.

wallrj:
Just to be clear, the real problem here was that I hadn't included the `setup_authentication` step in `build.yaml`.

The tests happened to pass as long as the acceptance tests had recently been run on the slave node, leaving behind the correct private SSH key loaded in a running `ssh-agent` process.

The tests started failing after the private SSH key got changed on the slave. 
It's not clear why it changed, but `/home/ubuntu/.ssh/id_rsa` was originally `vagrant-buildbot` and got replaced with the `jenkins-slave` key, breaking the cloudformation test.

https://clusterhq.atlassian.net/browse/FLOC-4214

PR #2656, opened 2016-02-24
[FLOC-4184] add initial segment support

msh100:
Closes FLOC-4184


wallrj:
Thanks @msh100 

@ferrantim explained that we want to add this to the current docs for 1.10.2 in which case we'll need to do a docs release.
And to do a docs release, this change / PR first needs to be merged into `release/flocker-1.10.2` branch.

For example, @tomprince just did this for some API docs changes:
 * #2660 was first merged into the release branch 
 * Then in #2666 `1.10.2.post1` was released.

So I'll create a new branch and close this PR.

PR #2658, opened 2016-02-25
[FLOC-3829] Have the GCE driver implement ICloudAPI

sarum90:
Implement the ICloudAPI in the GCE IBlockDeviceAPI.

Also, actually test all of the endpoints at the functional level.

The GCE functional tests now are considerably slower to run since they actually spin up 2nd instances for some tests.  Since they are only run nightly on Jenkins this seemed reasonable.

### DIFF
--- a/flocker/node/agents/functional/test_gce.py
+++ b/flocker/node/agents/functional/test_gce.py
@@ -92,7 +92,7 @@ class GCEComputeTestObjects(Fixture):
                 else:
                     raise e
 
-        self.addCleanup(destroy_best_effort(instance))
+        self.addCleanup(lambda: destroy_best_effort(instance))
         return instance
 
 
--- a/flocker/node/agents/functional/test_gce.py
+++ b/flocker/node/agents/functional/test_gce.py
@@ -41,6 +41,8 @@ from ..test.blockdevicefactory import (
     get_minimum_allocatable_size, get_device_allocation_unit
 )
 
+from testtools.matchers import MatchesAll, Contains, Not
+
 from ....testtools import TestCase
 
 
@@ -143,6 +145,64 @@ class GCEBlockDeviceAPITests(TestCase):
         self.assertEqual([cluster_2_dataset_id],
                          list(x.dataset_id
                               for x in gce_block_device_api_2.list_volumes()))
+
+    def test_list_live_nodes_pagination_and_removal(self):
+        """
+        list_live_nodes should be able to walk pages to get all live nodes and
+        should not have nodes after they are destroyed.
+        """
+        api = gceblockdeviceapi_for_test(self)
+
+        # Set page size to 1 to force pagination after we spin up a second
+        # node.
+        api._page_size = 1
+
+        gce_fixture = self.useFixture(GCEComputeTestObjects(
+            compute=api._compute,
+            project=get_machine_project(),
+            zone=get_machine_zone()
+        ))
+
+        other_instance_name = u"functional-test-" + unicode(uuid4())
+        other_instance = gce_fixture.create_instance(instance_name)
+
+        self.assertThat(
+            api.list_live_nodes(),
+            MatchesAll(
+                Contains(other_instance_name),
+                Contains(api.compute_instance_id())
+            )
+        )
+
+        api.stop_node(other_instance_name)
+
+        self.assertThat(
+            api.list_live_nodes(),
+            MatchesAll(
+                Not(Contains(other_instance_name)),
+                Contains(api.compute_instance_id())
+            )
+        )
+
+        api.start_node(other_instance_name)
+
+        self.assertThat(
+            api.list_live_nodes(),
+            MatchesAll(
+                Contains(other_instance_name),
+                Contains(api.compute_instance_id())
+            )
+        )
+
+        other_instance.destroy()
+
+        self.assertThat(
+            api.list_live_nodes(),
+            MatchesAll(
+                Not(Contains(other_instance_name)),
+                Contains(api.compute_instance_id())
+            )
+        )
 
     def test_attach_elsewhere_attached_volume(self):
         """
--- a/flocker/node/agents/functional/test_gce.py
+++ b/flocker/node/agents/functional/test_gce.py
@@ -26,6 +26,7 @@ to all Google Cloud services in the same project.``
 from uuid import uuid4
 from fixtures import Fixture
 from characteristic import attributes
+from googleapiclient.errors import HttpError
 
 from ..blockdevice import AlreadyAttachedVolume
 
@@ -33,7 +34,7 @@ from ..gce import get_machine_zone, get_machine_project
 from ....provision._gce import GCEInstanceBuilder
 
 from ..test.test_blockdevice import (
-    make_iblockdeviceapi_tests
+    make_iblockdeviceapi_tests, make_icloudapi_tests
 )
 
 from ..test.blockdevicefactory import (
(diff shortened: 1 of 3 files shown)


flocker/node/agents/gce.py
```diff
@@ -350,7 +353,7 @@ def _disk_resource_description(self):
         """
         return u"flocker-v1-cluster-id: " + unicode(self._cluster_id)
 
-    def _do_blocking_operation(self, function, **kwargs):
+    def _do_blocking_operation(self, function, max_wait=60, **kwargs):
```
justnoise: :grey_question: Nitpick: timeout is a more typical way of expressing max_wait.

Yes, I had to find _something_ in here :)

justnoise:
I can't find much to fault in here.  Implementation seems pretty close to what's done for EBS and Cinder with the addition of walking GCE response pages.

flocker/node/agents/gce.py
sarum90: Lol, Thanks-fixed. Actually that is definitely something I would have complained about later. I sincerely appreciate you catching that before I put that into master.

flocker/node/agents/functional/test_gce.py
```diff
@@ -78,7 +81,18 @@ def create_instance(self, instance_name):
...
+
+        def destroy_best_effort(inst):
+            try:
+                inst.destroy()
+            except HttpError as e:
+                if e.resp.status == 404:
+                    # The test must have already destroyed the instance.
+                    pass
+                else:
+                    raise e
```
wallrj: You probably know this already, but you lose the stack trace if you `raise e`. Instead just `raise`.
 * http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html

And maybe do 

```
if e.resp.status != 404:
    raise
```

...while you're here.

flocker/node/agents/functional/test_gce.py
```diff
@@ -78,7 +81,18 @@ def create_instance(self, instance_name):
...
+
+        self.addCleanup(lambda: destroy_best_effort(instance))
```
wallrj: :question: nit this could be `self.addCleanup(destroy_best_effort, instance)` I think

flocker/node/agents/gce.py
```diff
@@ -544,3 +549,42 @@ def destroy_volume(self, blockdevice_id):
...
+                project=self._project,
+                zone=self._zone,
+                maxResults=self._page_size,
+                pageToken=page_token
+            ).execute()
+            page_token = result.get('nextPageToken')
+            nodes.extend(result.get('items', []))
+            done = not page_token
+        return set(node.get("name") for node in nodes
+                   if node.get("status") == "RUNNING")
```
wallrj: :question: Might be better to use `node["name"]` and `node["status"]` so we get an clear error if the keys of the node dictionary ever change.

flocker/node/agents/gce.py
```diff
@@ -544,3 +549,42 @@ def destroy_volume(self, blockdevice_id):
...
+
+        Note this is only used in the functional tests of start_node.
+
+        :param unicode node_id: The compute_instance_id of the node to stop.
+        """
+        self._do_blocking_operation(
+            self._compute.instances().stop,
+            timeout_sec=5*60,
+            instance=node_id
+        )
```
wallrj: :question: Sounds like a useful method that should eventually be part of the ICloudAPI if we want to share the test_list_live_nodes test across other drivers. 
For now, I wonder if it should be private so that it's clearly not part of ICloudAPI.

flocker/node/agents/functional/test_gce.py
```diff
@@ -144,6 +169,64 @@ def test_multiple_cluster(self):
                          list(x.dataset_id
                               for x in gce_block_device_api_2.list_volumes()))
 
+    def test_list_live_nodes_pagination_and_removal(self):
+        """
+        list_live_nodes should be able to walk pages to get all live nodes and
+        should not have nodes after they are destroyed.
+        """
```
wallrj: :question: Why bother testing pagination here? I'd remove that part.
Seems like this'd be a useful `ICloudAPI` test in the future.
Other drivers probably don't / can't support the `stop_node` API so perhaps it should just rely on creating and terminating test nodes.

wallrj:
Thanks @sarum90 and @justnoise 

I left a few comments, some optional, some questions.

Please answer or address those and merge when green.

I also left a comment about the `already_attached` tests situation https://clusterhq.atlassian.net/browse/FLOC-1839

wallrj:
Oh, and I forgot to say, do you think it'd be worth making the GCE volume description a JSON blob from the get go to allow extra information to be added in future? 
Here:
```
    def _disk_resource_description(self):
        """
        Returns the value to be used in the description field of the disk
        resources for this cluster.

        :returns unicode: The value for the description.
        """
        return u"flocker-v1-cluster-id: " + unicode(self._cluster_id)

```

flocker/node/agents/functional/test_gce.py
sarum90: Thanks you for the reminder! When I wrote that I felt like I was doing it wrong.

flocker/node/agents/gce.py
sarum90: Done (above as well)

flocker/node/agents/gce.py
sarum90: +1 I also winder if this should actually be part of the provisioner instead of part of this interface.

For now though, just doing a GCE-specific implementation.

Changing to private though, as it is only used by this and the immediate functional tests of this.

flocker/node/agents/functional/test_gce.py
sarum90: So spinning up a second node is expensive, and I need at least 2 live nodes in order to verify pagination is working... sort of bad, but I went for 1 large large test that tests everything that requires having a second node spun up rather than two proper unit tests.

I will ~~make a sacrifice to the testing gods in order to ask for forgiveness for this uncleanliness~~ add a comment reflecting this.

flocker/node/agents/gce.py
sarum90: Also start_node and what is now _stop_node are only ever used in tests.  Really `ICloudAPI` is known to be an ugly interface at this point.  Somewhere that is cited for the reason that start_node isn't properly tested in the make_icloudapi_tests test suite :/

sarum90:
Thanks for the review! I addressed comments, and plan on re-kicking the test on jenkins, waiting for green and merging.

Re the GCE volume description:

GCE filters on those fields are really basic, essentially only allowing for exact string match. Thus, I'd like the content of description to be a pure function of the cluster-id in case we need to filter by cluster id. The driver does not currently use GCE filters, but for now I did not want to put us in a position where we would be unable to use filter on cluster id in the future (This would make getting a list of all volumes in a given cluster faster). I have a feeling if it were a JSON blob it would probably not be a pure function of cluster id.

So from my current understanding of the GCE API:

+1 to JSON being the answer if we want to add metadata in the future, but -1 to JSON being the answer if we need to do fancier filtering in the future, so I'd rather leave it as is for now and let future feature decisions decide which way to go.

Absolutely if we need more metadata in the future we should do that though. Shame on the GCE API about not having better ways to filter.

flocker/node/agents/functional/test_gce.py
sarum90: +1, But somehow I missed this comment before I kicked off my most recent test branch.

Is there a technical reason to prefer one over the other?

sarum90:
Got all greens (Including the GCE functional tests) So I'll be merging now.

PR #2659, opened 2016-02-25
[FLOC-4327] Use `PSet` for `required_config` in `BackendDescription`

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4237

This also makes it require unicode. I don't think anybody has started using this.

### DIFF
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -438,7 +438,7 @@ class BackendDescription(PClass):
     # out.
     needs_cluster_id = field(type=bool, mandatory=True)
     # Config "dataset" keys required to initialize this backend.
-    required_config = pset_field(str)
+    required_config = pset_field(unicode)
     api_factory = field(mandatory=True)
     deployer_type = field(
         mandatory=True,
@@ -469,15 +469,15 @@ _DEFAULT_BACKENDS = [
         name=u"openstack", needs_reactor=False, needs_cluster_id=True,
         api_factory=cinder_from_configuration,
         deployer_type=DeployerType.block,
-        required_config=set(["region", ]),
+        required_config={u"region"},
     ),
     BackendDescription(
         name=u"aws", needs_reactor=False, needs_cluster_id=True,
         api_factory=aws_from_configuration,
         deployer_type=DeployerType.block,
-        required_config=set(
-            ["region", "zone", "access_key_id", "secret_access_key", ]
-        ),
+        required_config={
+            u"region", u"zone", u"access_key_id", u"secret_access_key",
+        },
     ),
 ]
 
--- a/flocker/node/script.py
+++ b/flocker/node/script.py
@@ -426,8 +426,9 @@ class BackendDescription(PClass):
     :ivar api_factory: An object which can be called with some simple
         configuration data and which returns the API object implementing this
         storage backend.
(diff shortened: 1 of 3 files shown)


sarum90:
Looks good to me :)

Any reason not to use frozenset? I guess we're already off in pyrsistent land here.

flocker/node/script.py
```diff
@@ -438,7 +439,7 @@ class BackendDescription(PClass):
     # out.
     needs_cluster_id = field(type=bool, mandatory=True)
     # Config "dataset" keys required to initialize this backend.
-    required_config = field(type=set, mandatory=True, initial=set())
+    required_config = pset_field(unicode)
```
sarum90: Sorry, thinking about this again... it'd really be annoying if someone develops a driver with flocker 1.10 and forgets to put the "u" in front of the strings, and then later upgrades and their plugin breaks.

Could we add a factory to this field that will turn "str" values into "unicode" or instead just not specify a type of the set?

sarum90: More importantly, I think we need an initial value, otherwise we'll break for old drivers.

sarum90:
Even though I doubt people are using this today, on the off chance that someone in the future develops a driver on flocker 1.10, and decides to use this feature it'd be really crappy if it immediately broke when they updated. I think we should maybe set the type to (unicode, str) rather than just unicode (even though it makes me sad to do that).

Otherwise looks great.

sarum90:
Is the fear python3?
```
flocker-review ♦ ~/dev/flocker (enable-gce-driver-and-acceptance-tests-FLOC-3826)$ python3
Python 3.5.0 (default, Sep 23 2015, 04:42:00) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> b"cat" in set([u"cat"])
False
>>> u"cat" in set([b"cat"])
False
>>>
```

I still think we should accept either, and then translate it into a set of unicodes if we really need a set of unicodes later.

Cause python2 seems more lenient on this point:
```
flocker-review ♦ ~/dev/flocker (enable-gce-driver-and-acceptance-tests-FLOC-3826)$ python
Python 2.7.10 (default, Sep 23 2015, 04:34:14) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.72)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> b"cat" in set([u"cat"])
True
>>> u"cat" in set([b"cat"])                                                                                                                                                                                                                                                                                                                                               
True
>>> quit()
```

tomprince:
This is definitely something that matters mostly for python3 (although, the python3 behavior is better).

While in general, I agree the idea that this is an incompatible change, nobody is using `required_config` yet (see [this search](https://github.com/search?q=FLOCKER_BACKEND+required_config&ref=reposearch&type=Code&utf8=✓)). Given that we can update the docs for 1.10.2 to have the new constraint, and there are no users. I'm not convinced that it is worth supporting `bytes` as well as `unicode` here.

tomprince:
Regarding `pset` vs. `frozenset`, it is mainly a consistency thing. Nested `pyrsistent` objects behave nicely together (`.transform` for example).

PR #2662, opened 2016-02-26
[FLOC-4199] Verify that the version.html value matches the version being published

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4199

Sorry this branch grew so big.
I added the check for version and was faced with updating dozens of tests with fixed dictionaries of fake S3 keys.
So instead I modified FakeAWS to use an immutable FakeAWSState and added some canned states to test_release.
The tests can then transform the canned state, adding modifying keys, route and redirection rules.
Trouble is that that then broke a bunch of other test modules that also use FakeAWS so I had to go and update those too.

I also deleted a bunch of ``publish_docs`` which are concerned with publishing to a staging bucket. That is no longer part of the release process. We only ever use ``admin/publish_docs --production``.

### DIFF
--- a/admin/aws.py
+++ b/admin/aws.py
@@ -548,13 +548,18 @@ class FakeAWS(object):
         """
         See :class:`UploadToS3`.
         """
-        bucket = self.s3_buckets[intent.target_bucket]
+
         with intent.file.open() as source_file:
             content = source_file.read()
+        # XXX: Need to think about this.
+        # The fake currently only allows unicode content.
         content_type = intent.content_type
         if content_type is not None:
             content = ContentTypeUnicode(content, content_type)
-        bucket[intent.target_key] = content
+        self.state = self.state.transform(
+            ['s3_buckets', intent.target_bucket, intent.target_key],
+            content
+        )
 
     def get_dispatcher(self):
         """
--- a/admin/aws.py
+++ b/admin/aws.py
@@ -10,6 +10,7 @@ from effect import Effect, sync_performer, TypeDispatcher
 from effect.do import do, do_return
 
 import boto
+from boto.s3.website import RoutingRules
 from pyrsistent import PClass, pmap_field, PMap, field, discard, freeze
 from twisted.python.filepath import FilePath
 
@@ -438,7 +439,7 @@ class FakeAWSState(PClass):
     """
     routing_rules = pmap_field(
         key_type=unicode,
-        value_type=boto.s3.website.RoutingRules
+        value_type=RoutingRules
     )
     s3_buckets = pmap_field(
         key_type=unicode,
--- a/admin/aws.py
+++ b/admin/aws.py
@@ -7,9 +7,11 @@ Effectful interface to boto.
 import os
 from characteristic import attributes, Attribute
 from effect import Effect, sync_performer, TypeDispatcher
-from effect.do import do
+from effect.do import do, do_return
 
 import boto
+
+from twisted.python.filepath import FilePath
 
 
 @attributes([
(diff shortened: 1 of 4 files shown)


admin/aws.py
```diff
@@ -380,11 +425,31 @@ def __new__(cls, value, content_type):
...
+    """
+    The immutable state of ``FakeAWS``
+
+    :ivar routing_rules: Dictionary of routing rules for S3 buckets. They are
+        represented as dictonaries mapping key prefixes to replacements. Other
+        types of rules and attributes are supported or represented.
+    :ivar s3_buckets: Dictionary of fake S3 buckets. Each bucket is represented
+        as a dictonary mapping keys to contents. Other attributes are ignored.
+    :ivar cloudfront_invalidations: List of
+        :class:`CreateCloudFrontInvalidation` that have been requested.
```
wallrj: Document ``error_key``

admin/aws.py
```diff
@@ -499,3 +584,11 @@ def get_dispatcher(self):
...
                 self._perform_create_cloudfront_invalidation,
         })
+
+
+def fake_aws(routing_rules, s3_buckets):
+    initial_state = FakeAWSState(
+        routing_rules=routing_rules,
+        s3_buckets=freeze(s3_buckets),
+    )
+    return FakeAWS(state=initial_state)
```
wallrj: Remove this.

admin/release.py
```diff
@@ -287,6 +319,7 @@ def publish_docs(flocker_version, doc_version, environment, routing_config):
             UpdateS3ErrorPage(bucket=configuration.documentation_bucket,
                               target_prefix=version_prefix))
 
+    # XXX: We also need to calculate and invalidate the changed "latest" keys.
```
wallrj: This is a bug in the invalidation rules.

tomprince:
As indicated on the issue, we shouldn't drop support for staging, as that allows for testing the script.

admin/aws.py
```diff
@@ -281,11 +284,52 @@ def perform_download_s3_key(dispatcher, intent):
...
+@do
+def perform_read_s3_key(dispatcher, intent):
+    """
+    See :class:`ReadS3Key`.
+    """
+    target_file = FilePath(
+        u'/tmp/{}.perform_read_s3_key'.format(
+            __file__.replace(u"/", "!"),
+        )
+    ).temporarySibling()
```
tomprince: Looking at `DownloadS3Key`, it would be less code and simpler to just download to a string (which `boto` can do), than do some gyrations with a temporary file, just to reuse a bit of code.

admin/aws.py
```diff
@@ -399,14 +464,17 @@ class FakeAWS(object):
         :class:`CreateCloudFrontInvalidation` that have been requested.
     """
     def __init__(self):
-        self.cloudfront_invalidations = []
+        self.initial_state = self.state
```
tomprince: This attribute seems unused?

admin/aws.py
```diff
@@ -380,11 +425,31 @@ def __new__(cls, value, content_type):
...
+    routing_rules = pmap_field(
+        key_type=unicode,
+        value_type=RoutingRules
+    )
+    s3_buckets = pmap_field(
+        key_type=unicode,
+        value_type=PMap
+    )
+    error_key = pmap_field(key_type=unicode, value_type=unicode)
+    cloudfront_invalidations = field(initial=freeze([]))
```
tomprince: :exclamation: If you used more specific types for all these, you woudn't need to litter the test code with tons of `freeze` calls, which are just noise.

admin/release.py
tomprince: Is there a ticket for this?

tomprince:
- Doesn't verify that code does the right thing for subdirectories
- It doesn't appear that `test_overwrites_existing_documentation` actually tests that anything actually overrides any content (although it is hard to be sure, since once can't tell what the test is actually testing without having several different parts of the file open at once).
- There are no running tests for deleting keys. I realize that the invalidation is broken for them, but we should still test that the part that works actually does.

As a general comment, I find the new tests harder to reason about, since I need to keep in my head what the different state constants mean as a I look at the tests. So, I don't think the less repetitive nature of the new tests is actually a good thing. Thinking about it a bit more, I wonder if having:
- a way to merge parts of buckets
- a way to generate a document tree rooted at a particular bucket prefix
would allow the best of both worlds? That is, all the information needed to reason about a given tests in that test and not having dozens of repetitive lines in each test?

However, I think that test change is big enough to warrant separating it out from the actual functional change in this issue.

PR #2665, opened 2016-02-27
[FLOC-4248] Cloudformation stub script

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4248

* Moved the body of the cloudformation.py script into a main function
* Moved and simplified the argparse code so that I can easily add new options.
* Removed the existing functional tests. We really only need to run unit tests on the argparse options but even those tests seem unnecessary at this stage.
* Updated docs


admin/installer/cloudformation.py
```diff
@@ -72,332 +136,275 @@ def _sibling_lines(filename):
         return f.readlines()
 
 
-def _get_cluster_size():
+def print_flocker_docker_template(options):
```
tomprince: I don't think this should take an argparse option object. Instead, it should take just the options it needs, and the caller can do the processing needed to get the argparse options out.

tomprince: Also, I'd consider moving the `print` out of this function.

tomprince:
This looks reasonable. I have a couple of minor suggestions about the interface to generating the template. Please address and merge.

PR #2667, opened 2016-02-29
[FLOC-4254] Add initial segment support for 1.10.2

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4254

Release maintenance branch based on #2656

### DIFF
--- a/docs/_themes/clusterhq/layout_docs.html
+++ b/docs/_themes/clusterhq/layout_docs.html
@@ -284,5 +284,17 @@
     </script>
     <!-- Javascript include for form feedback submit -->
     <script type="text/javascript" src="{{ pathto('_static/js/feedback.js', 1) }}"></script>
+
+
+    <!-- Snippet from Segment.com for segment integration -->
+    <script type="text/javascript">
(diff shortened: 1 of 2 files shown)


tomprince:
Looking at https://segment.com/docs/libraries/analytics.js/quickstart/ it suggests turning off other integrations. It looks like we have at least
- optimizely
- heatmap.me
- google analytics
already installed. Should we consider using the segment integration instead? (cc: @msh100)

tomprince:
I also see some intercom code.

tomprince:
Also, I don't know any javascript, but it looks like there might be a race between loading the analyitics code, and trying to run it. It isn't clear to me why some of the code is inline and some in an external script.

tomprince:
I've made some comments, about trying to understand the reasoning behind how this change is implemented. it might just be ignorance on my part, feel free to ignore them and merge.

wallrj:
Thanks @tomprince 

I'll wait for @msh100 to respond before proceeding.

Meanwhile I've forced a "release" style build on the branch. 

 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/release-maintenance-flocker-1.10.2-segment-docs-FLOC-4254-2/

msh100:
@tomprince I was aware of a potential race condition which I informed Michael about, however I need to test it built somewhere that isn't local to confirm that. 

Where does Jenkins push to?

tomprince:
@msh100 You can find a link in the console output of a sphinx build, near the end.

msh100:
@tomprince Regarding this code co-existing with other integrations, the only integration enabled from within segment itself is Mixpanel which is not in anyway implemented in the documentation at the moment. Absolutely those integrations should be removed if we plan to move them over to Segment at some point, it's my understanding this is purely being used for testing at the moment.

Regarding the race condition: All code regarding Segment is in the document ready callback, this function waits for the page to become ready before execution (this includes loading of assets). I don't believe there is a race condition in this instance.

If you feel any specific part of the JS needs explaining more, let me know and I will add comments.

One issue I did notice which does appear to be a problem is a race condition between sending the information about the page to Segment and the changes which were implemented to improve SEO, JS injection of the canonical link path. Segment appears to show pages at a different path than they were actually accessed, I think we need @ferrantim to confirm if this is okay or not.

ferrantim:
I spoke to marcus and it looks like the one race condition won't be a problem when this is live.  So I am ok with this.

"
sometimes segment sees it as “/en/version/path"

[4:02] 
other times it is the real path

[4:03] 
when it is live in production, I dont think this will be an issue, because it will be the right path I guess"

wallrj:
Superseded by #2933

PR #2673, opened 2016-03-02
[FLOC-4257] A tool to print the latest official regional AMI IDs for Ubuntu

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4257

This is part of our effort to update CloudFormation templates during Flocker releases.
You can read our draft docs in #2672 outlining a rather manual process which calls this tool as a first step. Later it can be automated.

This PR is one way to get the AMI IDs.
It seemed like it might be simpler and more correct to directly to Ubuntu.com as the source of truth, but it wasn't as easy as i thought since there are no APIs just text files and something called `simplestreams` which turned out to be unusably slow.
 * http://discourse.ubuntu.com/t/which-ami-in-aws-does-correspond-to-ubuntu-14-04-2/2195/2

The alternative is to query the AWS API for images belonging to owner `099720109477` and sort by date:
 * http://ubuntu-smoser.blogspot.com/2011/07/how-to-find-right-ubuntu-ami-on-ec2.html
 * https://askubuntu.com/questions/53582/how-do-i-know-what-ubuntu-ami-to-launch-on-ec2

That might be a better and more flexible AMI search tool but I'd need a way to query all regions quickly and to be sure that images from all regions are the same (but perhaps that's not so important).

I suppose I could instantiate a separate BOTO3 driver for all available regions and query them in parallel with `deferToThread`.

This is good enough for now and I might investigate the alternative in a followup issue.

### DIFF
--- a/admin/ubuntu.py
+++ b/admin/ubuntu.py
@@ -1,0 +1,186 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Helpers for Ubuntu cloud images.
+"""
+
+import json
+import sys
+from urllib2 import urlopen
+
+from pyrsistent import PClass, field, pmap, freeze
+from twisted.python.usage import Options, UsageError
+
+URL_BASE = u"https://cloud-images.ubuntu.com/query"
+URL_LATEST_TEMPLATE = URL_BASE + u"/{release_cycle}.latest.txt"
+URL_DETAILS_TEMPLATE = (
+    URL_BASE + u"/{ubuntu_name}/{ubuntu_variant}/{release_cycle}.current.txt"
+)
+
+
+def url_mangle(template, template_variables):
+    template_variables = template_variables.copy()
+    release_cycle = template_variables["release_cycle"]
+    if release_cycle == u"release":
+        release_cycle = u"released"
+    template_variables["release_cycle"] = release_cycle
+    return template.format(**template_variables)
+
+
+def new_record_type(name, headings):
+    return type(
+        name,
+        (PClass,),
+        {heading: field(type=unicode, mandatory=True) for heading in headings}
+    )
+
+
+HEADINGS_BASE = (
+    "ubuntu_name",
+    "ubuntu_variant",
+    "release_cycle",
+    "release_date",
+)
+
+HEADINGS_DETAIL = HEADINGS_BASE + (
+    "ec2_image_type",
+    "architecture",
+    "region",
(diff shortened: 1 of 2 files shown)


admin/ubuntu.py
```diff
@@ -0,0 +1,186 @@
...
+    print json.dumps(
+        {
+            r.region: r.ami_id
+            for r in latest(
+                release_cycle=options["release-cycle"],
+                ubuntu_name=options["ubuntu-name"],
+                ubuntu_variant=u"server",
+                ec2_image_type=u'ebs',
+                architecture=u'amd64',
+                hypervisor=u'hvm',
```
myechuri: Any particular reason for ``architecture`` and ``hypervisor`` choices? Thanks.

admin/ubuntu.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+    )
+
+
+class AMISearchUbuntuOptions(Options):
+    """
+    Options.
+    """
+    optParameters = [
+        ['release-cycle', None, u"daily",
```
myechuri: Should the default release cycle be ``release`` instead of ``daily``? Thanks.

myechuri:
Thanks for the change, @wallrj ! The changes look good. Can you please clarify on the default release cycle choice, then proceed with the merge? Thanks!

admin/ubuntu.py
wallrj: For `hypervisor` it's a choice between `hvm` or `paravirtualization`  and we're using the former in our installer images because it's now the recommended image type. More here:
 * https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/virtualization_types.html

And it's `amd64` vs `i386`. Centos use `x86_64` . Since we only ever install Flocker on 64bit machines it didn't seem like something we need to make configurable.

admin/ubuntu.py
wallrj: I chose daily as that's what I've always used when looking up the AMI IDs manually.

I'd been assuming that `released` meant that you get an Ubuntu image based on the packages at the time of first release, but that's not it:
 * https://cloud-images.ubuntu.com/query/trusty/server/released.current.txt

Were updated on 20160222. I'll stick with daily for now, but might change if I find a good explanation of the released list.

wallrj:
Thanks @myechuri 

I'll merge this as-is and make further improvements as I add it to the release process.

PR #2675, opened 2016-03-03
[FLOC-4265] Refactor `create_node` vs. `create_nodes`.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4265

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -816,7 +816,7 @@ class LibcloudRunner(object):
         print "Assigning random tag:", self.random_tag
         names = [self._make_node_name(self.random_tag, index)
                  for index in range(self.num_nodes)]
-        self.nodes = yield gather_deferreds(
+        yield gather_deferreds(
             self._create_and_provision_nodes(reactor, names)
         )
 
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -23,7 +23,6 @@ from bitmath import GiB
 from twisted.internet.error import ProcessTerminated
 from twisted.python.usage import Options, UsageError
 from twisted.python.filepath import FilePath
-from twisted.python.failure import Failure
 from twisted.internet.defer import inlineCallbacks, returnValue, succeed
 from twisted.python.reflect import prefixedMethodNames
 
@@ -603,29 +602,6 @@ class LibcloudRunner(object):
         # operation fails in a way such that it isn't clear if the instance has
         # been created or not.
         self.random_tag = b32encode(os.urandom(8)).lower().strip("\n=")
-
-    def _create_node(self, name):
-        """
-        Create a cloud node with the given name.
-
-        Additional properties of the node are defined in the instance.
-        :param str name: The name.
-        :return: Node instance.
-        """
-        try:
-            print "Creating node {}".format(name)
-            node = self.provisioner.create_node(
-                name=name,
-                distribution=self.distribution,
-                metadata=self.metadata,
-            )
-            self.nodes.append(node)
-            return node
-        except BaseException:
-            print "Error creating node %s" % (name,)
-            print "It may have leaked into the cloud."
-            write_failure(Failure())
-            raise
 
     def _provision_node(self, reactor, node):
         """
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -814,37 +814,9 @@ class LibcloudRunner(object):
         :return Cluster: The cluster to connect to for acceptance tests.
         """
         print "Assigning random tag:", self.random_tag
-        for index in range(self.num_nodes):
-            name = self._make_node_name(self.random_tag, index)
-            try:
-                print "Creating node %d: %s" % (index, name)
-                node = yield self.provisioner.create_node(
-                    name=name,
-                    distribution=self.distribution,
-                    metadata=self.metadata,
-                )
-            except:
-                print "Error creating node %d: %s" % (index, name)
-                print "It may have leaked into the cloud."
-                raise
-
-            yield remove_known_host(reactor, node.address)
-            self.nodes.append(node)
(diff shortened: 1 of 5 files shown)


admin/acceptance.py
```diff
@@ -604,29 +603,6 @@ def __init__(self):
         # been created or not.
         self.random_tag = b32encode(os.urandom(8)).lower().strip("\n=")
 
-    def _create_node(self, name):
```
tomprince: This is just dead code.

flocker/provision/_libcloud.py
```diff
@@ -197,153 +195,6 @@ def get_ssh_key(self):
             # https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_KeyPairInfo.html
             return None
 
-    def create_node(self, name, distribution, metadata={}):
```
tomprince: I *think* the logic here (and for AWS) is mostly the same for both the single node and multiple node code paths, but it is a bit hard to follow the code to be sure. The one difference that I am aware of is that the single-node libcloud case retries on failure, but this is handled at higher level in the multi-node case (in `admin.acceptance`).

tomprince:
It would be good to get a second pair of eyes to verify that the new code-paths are equivalent to the existing ones.

flocker/provision/_aws.py
```diff
@@ -473,7 +413,7 @@ def make_node(ignored, name, instance):
             # so we can leave action_completion alone now.
             return results
 
-    def _async_get_node(self, reactor, instance, metadata):
+    def _get_node(self, reactor, instance, metadata):
         """
         Configure the given AWS instance, wait until it's running
         and create an ``AWSNode`` object for it.
```
zubron: This isn't an issue with your change but I find this docstring a little confusing since an `AWSNode` object isn't created as part of this function. That appears to happen within `make_node` in `create_nodes`.

zubron:
Thanks @tomprince. I think this is a great change to make. This is my first time properly looking through this code though so I will probably put it back up for a second review.

I think that it looks good however I do find this code quite difficult to follow, particularly due to the inconsistencies in naming between each of the provisioners. For example, GCE has `_create_node` however AWS has `make_node`, `_get_node` and `_run_nodes`, and it's not entirely clear what the responsibility of each of those is.  I don't know if trying to add some consistency is something that should be done as part of this change, but I thought I'd mention it here anyway.

flocker/provision/_aws.py
```diff
@@ -297,66 +297,6 @@ def get_ssh_key(self):
...
-            poll_until(lambda: self._set_metadata(instance, metadata),
-                       repeat(1, INSTANCE_TIMEOUT))
-            try:
-                _wait_until_running(instance)
-                return instance
-            except FailedToRun:
-                instance.terminate()
-                return None     # the instance is in the wrong state
-
     def _run_nodes(self, count, image_id, size, diskmap):
```
zubron: An example of the naming issues I mentioned in another comment. This function is creating instances and the name doesn't quite reflect that.

tomprince:
>  I do find this code quite difficult to follow.

Me too. But I think that is beyond the scope of this ticket.

tomprince:
To expand on that, this PR is mostly deleting code in `flocker.provision`, and changing the caller in `admin.acceptance`.

admin/acceptance.py
sarum90: Confirmed in review.

admin/acceptance.py
```diff
@@ -838,37 +814,9 @@ def start_cluster(self, reactor):
...
-            zfs_commands = parallel([
-                configure_zfs(node, variants=self.variants)
-                for node in self.nodes
-            ])
-            commands = commands.on(success=lambda _: zfs_commands)
-
-        yield perform(make_dispatcher(reactor), commands)
+        names = [self._make_node_name(self.random_tag, index)
+                 for index in range(self.num_nodes)]
+        self.nodes = yield gather_deferreds(self._create_nodes(reactor, names))
```
sarum90: :question: nit: if you wouldn't mind renaming _create_nodes to  _create_and_provision_nodes it would be more obvious that this is correct without having to look at the implementations of everything.

flocker/provision/_libcloud.py
sarum90: Going through line by line. Will comment on all differences.

flocker/provision/_libcloud.py
```diff
@@ -197,153 +195,6 @@ def get_ssh_key(self):
...
-        :return libcloud.compute.base.Node: The created node.
-        """
-        size = self._default_size
-
-        image_name = self._image_names[distribution]
-
-        create_node_arguments = self._create_node_arguments()
-
-        node, addresses = self._create_with_retry(
-            name=name,
```
sarum90: in create_nodes this just logs the instance_count rather than the all of the names. Probably okay.

flocker/provision/_libcloud.py
```diff
@@ -197,153 +195,6 @@ def get_ssh_key(self):
...
-            image=get_image(self._driver, image_name),
-            size=get_size(self._driver, size),
-            ex_keyname=self._keyname,
-            ex_metadata=metadata,
-            **create_node_arguments
-        )
-
-        public_address = addresses[0]
```
sarum90: Public address is gotten differently, but that shouldn't matter.

sarum90:
I also think this refactor is safe and should be approximately a No-op.

So long as you get passing acceptance tests on cinder and aws I think that should be enough confidence to merge.

Worst case scenario our retry logic is subtly different than it used to be.

tomprince:
Turns out, this retries provisioning forever, which means that any errors in provisioning are obscured.

PR #2678, opened 2016-03-04
[FLOC-4271] Make the P2P deployer not require the container agent running.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4271

This is a follow up to #2272 for the obsolete ZFS backend.

I'm doing this so I can test the code-paths for provisioning hooks with in-tree code.


sarum90:
Thanks!

I'm not going to dive too deeply into the details of this one.

Looks like you are correctly getting rid of some code that is no longer needed (+ a test) now that we have leases, and modifying another test to use a constant that is probably used elsewhere and kept up to day.

Merge once tests age green.

PR #2680, opened 2016-03-04
[FLOC-3510] Cinder v2 API paging support

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-3510

These changes allow the OpenStack backend to use either Cinder V1 or Cinder V2
Configurable via a new optional key in the dataset configuration file.
The default remains v1 until we drop support for Rackspace.

@wallrj has a devstack environment running if the reviewer wants to try out the branch there.

Just ask and I'll add your keys.


flocker/node/agents/cinder.py
```diff
@@ -541,7 +547,7 @@ def attach_volume(self, blockdevice_id, attach_to):
     def detach_volume(self, blockdevice_id):
         try:
             cinder_volume = self.cinder_volume_manager.get(blockdevice_id)
-        except CinderNotFound:
+        except CinderClientNotFound:
```
wallrj: Latest `python-cinderclient` raises a cinder rather than a keystone derived exception.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -113,29 +113,23 @@ def cinder_client(self):
...
         :return: A ``cinderclient.Cinder`` instance.
         """
-        try:
-            config = get_blockdevice_config(ProviderType.openstack)
-        except InvalidConfig as e:
-            self.skipTest(str(e))
-        session = get_keystone_session(**config)
-        region = get_openstack_region_for_test()
-        return get_cinder_v1_client(session, region)
+        return self.api.cinder_volume_manager
```
wallrj: The tests below need direct access to a `python-cinderclient` `VolumeManager`.
We used to create it from scratch, but now we need to be sure that we're using the same v1 or v2 that is being used by the Flocker IBlockDeviceAPI driver.
So I return a reference to Flocker driver's `python-cinderclient` VolumeDriver instead.
Unfortunately that causes a fair amount of churn below, because the Flocker driver only has a reference to the VolumeManager, whereas we use to return a CinderClient (which *has* a VolumeManager (volumes) attribute).

flocker/node/agents/functional/test_cinder.py
```diff
@@ -168,10 +162,16 @@ def test_name(self):
...
+        volume = cinder_client.get(
+            flocker_volume.blockdevice_id
+        )
         self.assertEqual(
-            cinder_client.volumes.get(
-                flocker_volume.blockdevice_id).display_name,
-            u"flocker-{}".format(dataset_id))
+            # Why? Because v1 calls it display_name and v2 calls it
+            # name.
+            getattr(volume, volume.NAME_ATTR),
```
wallrj: i.e. `cinderclient.v1.client.Volume.name` vs `cinderclient.v2.client.Volume.display_name`

flocker/node/agents/functional/test_cinder_behaviour.py
```diff
@@ -123,3 +110,104 @@ def test_updated_metadata_is_listed(self):
...
+    deleted_volume_ids = set(v.id for v in volumes)
+
+    def all_gone():
+        found_volume_ids = set(
+            v.id for v in cinder_volume_manager.list()
+        )
+        return len(deleted_volume_ids.intersection(found_volume_ids)) == 0
+    return poll_until(all_gone, repeat(1, 60))
+
+
```
wallrj: I'll disable these tests before merging.
They only make sense when run in an OpenStack environment where `osapi_max_limit = 5`.

requirements.txt
```diff
@@ -47,21 +51,21 @@ pycparser==2.14
 pycrypto==2.6.1
 pyOpenSSL==0.15.1
 pyrsistent==0.11.9
-python-cinderclient==1.1.1
+python-cinderclient==1.6.0
```
wallrj: This is the important update. 
It turns out that  `python-cinderclient==1.1.1` didn't make any use of the `next` paging links returned in Cinder V2 response body. 
Upgrading `python-cinderclient` required an update to `python-novaclient` and a bunch of transient dependencies too.
This seems to be the minimum required to make pip / setuptools happy.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -224,7 +224,9 @@ def test_verify_false(self):
         config['peer_verify'] = False
         session = get_keystone_session(**config)
         region = get_openstack_region_for_test()
-        cinder_client = get_cinder_v1_client(session, region)
+        cinder_client = get_cinder_client(
+            session, region, version=config["cinder_api_version"]
```
wallrj: This is broken. 
Should re-use the same driver loading function as the IBlockDeviceAPI tests.

flocker/node/agents/cinder.py
```diff
@@ -852,19 +858,19 @@ def get_keystone_session(**config):
         )
 
 
-def get_cinder_v1_client(session, region):
+def get_cinder_client(session, region, version):
```
moshez: Do we expect that get_cinder_v1_client is internal-only? The path to it is fully public, so anyone using flocker-as-a-library can get burned by that. Otherwise we should support a get_cinder_v1 -> get_cinder shim?

moshez:
Summary of face-to-face discussion with @wallrj :
It seems like either making it a separate backend or making the cinder backend dynamically detect which API versions are supported would be better, so he is going to massively update this PR after deciding which direction to take.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -198,57 +197,47 @@ class CinderHttpsTests(TestCase):
...
+        config = self.config.copy()
+        path = self.make_temporary_directory()
         RootCredential.initialize(path, b"mycluster")
-        try:
-            config = get_blockdevice_config(ProviderType.openstack)
-        except InvalidConfig as e:
-            self.skipTest(str(e))
         config['backend'] = 'openstack'
         config['auth_plugin'] = 'password'
-        config['password'] = 'password'
```
wallrj: This test was broken. 
The ``UnauthorizedFailure`` was caused by the incorrect password rather than incorrect CA certificate. 
I fixed it and simplified the test too.
No need to get a cinder client to verify authentication, all you need is the Keystone session.

wallrj:
@moshez I switched to the automatic detection of highest supported Cinder API version.

See what you think.

flocker/node/agents/cinder.py
```diff
@@ -852,17 +854,53 @@ def get_keystone_session(**config):
...
 
 
-def get_cinder_v1_client(session, region):
+SUPPORTED_VERSIONS = (
+    (u"2", u"volumev2"),
+    (u"1", u"volume"),
+)
+
+
+class CinderAPIVersionDetectionFailure(Exception):
```
moshez: Probably should derive from EnvironmentError.

https://docs.python.org/2/library/exceptions.html#exceptions.EnvironmentError

flocker/node/agents/cinder.py
```diff
@@ -852,17 +854,53 @@ def get_keystone_session(**config):
...
     """
     Create a Cinder (volume) client from a Keystone session.
 
+    Try all ``SUPPORTED_VERSIONS`` in order and return the first client to
+    successfully complete a ``list`` API call.
+
     :param keystoneclient.Session session: Authenticated Keystone session.
     :param str region: Openstack region.
-    :return: A cinderclient.Client
+    :return: A cinderclient.Client (v2 else v1)
```
moshez: I'd move the "v2 else v1" up into the prose description and make it more verbose, and keep this line the same.

flocker/node/agents/cinder.py
```diff
@@ -852,17 +854,53 @@ def get_keystone_session(**config):
...
     """
-    return CinderClient(
-        session=session, region_name=region, version=1
-    )
+    endpoint_errors = []
+    for version, service_type in SUPPORTED_VERSIONS:
```
moshez: where is service_type used?

If it's not used, let's name the variable 'dummy' or '_' to make it clear it's not used. I'm not completely sure if it's used at all?

flocker/node/agents/cinder.py
```diff
@@ -902,8 +946,30 @@ def cinder_from_configuration(region, cluster_id, **config):
...
+        v1 uses display_name rather than name.
+        """
+        return self._client_v2.create(
+            size=size,
+            metadata=metadata,
+            name=display_name
+        )
+
+
+CINDER_V1_ADAPTERS = {
```
moshez: I'd prefer unifying this data structure with the supported versions data structure, as they should never be out of sync.

flocker/node/agents/cinder.py
```diff
@@ -902,8 +946,30 @@ def cinder_from_configuration(region, cluster_id, **config):
...
+        cinder_volume_manager=wrapped_cinder_volume_manager,
         nova_volume_manager=logging_nova_volume_manager,
         nova_server_manager=logging_nova_server_manager,
         cluster_id=cluster_id,
     )
+
+
+class Cinder1to2Adapter(proxyForInterface(ICinderVolumeManager, "_client_v2")):
+    """
+    Deal with annoying differences in the method signatures between
```
moshez: Shouldn't this be singular, not plural?

flocker/node/agents/functional/test_cinder.py
```diff
@@ -198,57 +197,47 @@ class CinderHttpsTests(TestCase):
...
         try:
-            cinder_client.authenticate()
-            return True
-        except Unauthorized:
-            return False
+            self.config = get_blockdevice_config(ProviderType.openstack)
+        except InvalidConfig as e:
+            self.skipTest(str(e))
+        auth_url = self.config['auth_url']
+        if not auth_url.startswith(u"https://"):
```
moshez: Mild preference for using urlparse and .schema rather than this :)

flocker/node/agents/functional/test_cinder.py
```diff
@@ -198,57 +197,47 @@ class CinderHttpsTests(TestCase):
...
     def test_verify_false(self):
         """
         With the peer_verify field set to False, connection to the
         OpenStack servers always succeeds.
         """
-        try:
-            config = get_blockdevice_config(ProviderType.openstack)
-        except InvalidConfig as e:
-            self.skipTest(str(e))
+        config = self.config.copy()
```
moshez: Why do we need this .copy? every setUp will recreate the config object, no?
[If not, I'd push the .copy into setUp]

flocker/node/agents/functional/test_cinder.py
```diff
@@ -198,57 +197,47 @@ class CinderHttpsTests(TestCase):
...
         config['peer_verify'] = False
-        session = get_keystone_session(**config)
-        region = get_openstack_region_for_test()
-        cinder_client = get_cinder_v1_client(session, region)
-        self.assertTrue(self._authenticates_ok(cinder_client))
+        session = get_keystone_session(**self.config)
```
moshez: If this is using self.config, who uses the "peer_verify = false" thing?

requirements.txt
moshez: How did you verify that this is the minimum?

stmcginnis: 1.6.1 is the official client as of the Mitaka release and the recommended one to use. I forget where we fixed the paging, but at least for other reasons I would recommend going to 1.6.1.

wallrj: Thanks @stmcginnis and @moshez 

I should have been clearer. 

I meant these are the minimum set of package upgrades in order for `pip` to satisfy all  dependencies after I'd updated to the latest ``python-cinderclient==1.6.1``.

I wasn't particularly scientific. I just upgraded, to latest, every package that was causing a conflict and repeated until I could run ``pip install --editable .[dev]`` cleanly.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: I missed that. It was left over from a previous implementation where I attempted to lookup the available Cinder API services in the OpenStack Keystone service catalog. 
Unfortunately, Rackspace deny access to the service_catalog endpoint.

I've removed the ``service_type``.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/cinder.py
wallrj: Done.

flocker/node/agents/functional/test_cinder.py

flocker/node/agents/functional/test_cinder.py
wallrj: You're right. No need to copy. Removed.

flocker/node/agents/functional/test_cinder.py
wallrj: Damn. I fixed one  bug and introduced another. Thanks for spotting. Fixed.

flocker/node/test/test_script.py
```diff
@@ -433,15 +480,18 @@ def test_default_openstack(self):
         """
         An OpenStack backend is available by default.
         """
+        mimic_port = mimic_for_test(test_case=self)
```
wallrj: The problem here is that the Cinder backend driver factory now calls ``VolumeManager.list`` in order to detect whether the v1 or v2 API is available. 
These are unit tests though, so they really shouldn't depend on having responsive Cinder endpoints.

However, I found that mimic now has a Cinder v2 simple list endpoint which is just enough to get this test passing. 
It seems OTT for such a basic test....but it works.

Alternatives:
 * Remove the test.
 * Monkey patch ``cinderclient.v2.volumes.VolumeManager.list`` for this test.
 * Implement a lazy loading ``ICinderClient`` and alter ``cinder_from_configuration`` to use it. The Cinder API version will be automatically detected on first use.

flocker/node/test/test_script.py
```diff
@@ -303,6 +307,49 @@ def test_logging(self, log_name):
...
+    test_case.addCleanup(cleanup)
+
+    poll_until(
+        predicate=log.exists,
+        steps=repeat(1, 5)
+    )
+
+    def port_from_log():
+        for line in log.open():
+            match = re.search(r"Site starting on (\d+)$", line)
```
exarkun: Might be an irrelevant tangent, but you can probably tell the child to listen on an inherited fd using, eg, the systemd endpoint (writing that, now seems silly to only have support for this via "systemd" instead of a generic inherited-fd endpoint, oh well).  It's probably a little fiddling around to set things up as systemd would have (and thus make the systemd endpoint actually work) but it means not having to parse the logs to discover the port.

wallrj:
I ran the functional test on a devstack server and they passed.

```
FAILED (skips=90, errors=35, successes=16)
(3510)[centos@richardw-cinder-machine-5 flocker]$ sudo /usr/bin/env FLOCKER_FUNCTIONAL_TEST=TRUE FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE=$PWD/acceptance.yml FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_SECTION=devstack-openstack FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER=devstack-openstack $(type -p trial) flocker.node.agents.functional
  BlockDeviceAPIDestroyTests
    test_destroy_timesout ...                                              [OK]
  CinderAttachmentTests
    test_get_device_path_no_attached_disks ...                             [OK]
  CinderBlockDeviceAPIInterfaceTests
    test_attach_attached_volume ...                                        [OK]
    test_attach_elsewhere_attached_volume ...                              [OK]
    test_attach_unattached_volume ...                                      [OK]
    test_attached_volume_listed ...                                        [OK]
    test_compute_instance_id_nonempty ...                                  [OK]
    test_compute_instance_id_unicode ...                                   [OK]
    test_detach_volume ...                                                 [OK]
    test_device_size ...                                                   [OK]
    test_foreign_cluster_volume ...                                        [OK]
    test_foreign_volume ...                                                [OK]
    test_get_device_path_device ...                                        [OK]
    test_get_device_path_device_repeatable_results ...                     [OK]
    test_list_attached_and_unattached ...                                  [OK]
    test_multiple_volumes_attached_to_host ...                             [OK]
    test_name ...                                                          [OK]
    test_reattach_detached_volume ...                                      [OK]
  CinderCloudAPIInterfaceTests
    test_current_machine_is_live ...                                       [OK]
    test_list_live_nodes ...                                               [OK]
  CinderHttpsTests
    test_verify_ca_path_no_match_fails ...                            [SKIPPED]
    test_verify_false ...                                             [SKIPPED]
  VirtIOCinderAttachmentTests
    test_disk_attachment_fails_with_conflicting_disk ...              [SKIPPED]
    test_get_device_path_correct_with_attached_disk ...               [SKIPPED]
    test_get_device_path_virtio_blk_error_without_udev ...            [SKIPPED]
    test_get_device_path_virtio_blk_symlink ...                       [SKIPPED]
...
```

So I'll merge this now and do some acceptance testing tomorrow.

PR #2683, opened 2016-03-05
[FLOC-4276] Add hooks to support provisioning external backends.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4276

I have a couple of questions:
- should the new configuration be at the top-level, or should there be a separate description of the acceptance testing configuration. I suspect there will need to be more configuration related to tests in the future ([this](https://github.com/ClusterHQ/flocker/blob/61a38a5124e28834e172ac12912424b152660b04/flocker/acceptance/testtools.py#L230) for example will need to be parameterized for ceph).
- Should these variables be actual functions, or fully-qualified names? Both here and in the [ceph driver](https://github.com/ClusterHQ/ceph-flocker-driver/compare/ceph-provision#diff-ed1a3ddb9abeaaaae06d9a573081f693R37) to avoid having to import test-code at runtime (either to avoid potential for import loops or bringing in additional run-time dependencies).

### DIFF
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -27,7 +27,7 @@ from twisted.python.failure import Failure
 from twisted.internet.defer import inlineCallbacks, returnValue, succeed
 from twisted.python.reflect import prefixedMethodNames
 
-from effect import parallel
+from effect import parallel, Func, Effect
 from txeffect import perform
 
 from uuid import UUID
@@ -41,7 +41,7 @@ from flocker.common import (
 )
 from flocker.common.script import eliot_to_stdout
 
-from flocker.provision import PackageSource, Variants, CLOUD_PROVIDERS
+from flocker.provision import PackageSource, Variants, CLOUD_PROVIDERS, Cluster
 from flocker.provision._ssh import (
     run_remotely,
     ensure_agent_has_ssh_key,
@@ -56,7 +56,6 @@ from flocker.provision._install import (
 )
 from flocker.provision._ca import Certificates
 from flocker.provision._ssh._conch import make_dispatcher
-from flocker.provision._common import Cluster
 from flocker.testtools.cluster_utils import (
     make_cluster_id, Providers, TestTypes
 )
@@ -553,11 +552,15 @@ def configured_cluster_for_nodes(
         dataset_backend_config_file=dataset_backend_config_file
     )
 
-    configuring = perform(
-        make_dispatcher(reactor),
-        configure_cluster(
+    commands = Effect(Func(
+        lambda: dataset_backend.acceptance_configure_cluster(cluster)
+    )).on(
+        success=lambda _: configure_cluster(
             cluster, dataset_backend_configuration, provider, logging_config
-        )
+        ),
+    )
+    configuring = perform(
+        make_dispatcher(reactor), commands,
     )
     configuring.addCallback(lambda ignored: cluster)
     return configuring
--- a/admin/acceptance.py
+++ b/admin/acceptance.py
@@ -53,7 +53,6 @@ from flocker.provision._install import (
     install_flocker,
     configure_cluster,
     configure_node,
-    configure_zfs,
 )
 from flocker.provision._ca import Certificates
 from flocker.provision._ssh._conch import make_dispatcher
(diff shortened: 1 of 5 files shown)


admin/acceptance.py
```diff
@@ -554,11 +552,15 @@ def configured_cluster_for_nodes(
         dataset_backend_config_file=dataset_backend_config_file
     )
 
-    configuring = perform(
-        make_dispatcher(reactor),
-        configure_cluster(
+    commands = Effect(Func(
```
sarum90: :question: This is different from how we use Effect elsewhere in the provisioning code. Trying to think through if that will cause us any troubles.

- Do we still generate our docs using Effect? Will this break that?
- Are there any meaningful unit tests that we write or want to write against our provisioning code that this will slow down?

The answer to the first bullet should be answered by Jenkins, and I think the answer to the second is a tentative no, so should be fine just to do inline like this. If you think the answer to the second is actually yes, please refactor into a new proper intent rather than just using Func.

flocker/node/backends.py
```diff
@@ -90,6 +90,23 @@ class BackendDescription(PClass):
     :ivar deployer_type: A constant from ``DeployerType`` indicating which kind
         of ``IDeployer`` the API object returned by ``api_factory`` is usable
         with.
+
+    The following fields are experimental, and subject to change
```
sarum90: :question: Maybe put these fields behind an interface. `IBackendProvisioner` or something. Just so there are more places you could put the warning.

admin/acceptance.py
```diff
@@ -641,9 +643,10 @@ def _provision_node(self, reactor, node):
         """
         commands = node.provision(package_source=self.package_source,
                                   variants=self.variants)
-        if self.dataset_backend == backends.ZFS:
-            zfs_commands = configure_zfs(node, variants=self.variants)
-            commands = commands.on(success=lambda _: zfs_commands)
+        commands = commands.on(
+            success=lambda _:
```
sarum90: below you added the node=node scoping trick, but here you did not. Is there a reason why it's needed below and not here?

sarum90:
Regarding your questions... I think it might be cleaner if you let backends optionally define an interface, and you call some method on that interface for all backend-specific decisions.

Then if Ceph needs something parametrized, that can be part of it's interface implementation.

sarum90:
Code looks great.

A couple things here and there to consider, but nothing blocking (given that the tests appear to have passed).

:+1:

admin/acceptance.py
tomprince: It is the usual thing about capturing loop variables.

admin/acceptance.py
tomprince: > This is different from how we use Effect elsewhere in the provisioning code.

Effect performers can return effects, which then get run (much like deferreds). I'm presuming at some point, these hooks will be used by people that don't want to learn about effect, so I want to make sure that they get called at the appropriate time (hence the `Func`). We don't need this for our own code, since the functions we call just return an `Effect` instead of having side-effects.

> Do we still generate our docs using Effect? Will this break that?

Yes and no. The docs hook into lower level tasks. Even if they hit this, as you point out, there can be a no-op implementation.

> Are there any meaningful unit tests [...]

I'm not sure what a meaningful unit test would look like. In any case, I don't think it would be too hard to introduce an intent later.

flocker/node/backends.py
```diff
@@ -90,6 +90,23 @@ class BackendDescription(PClass):
...
+    without notice.
+
+    :ivar acceptance_provision_node: Function that will provision a
```
jml: Should this really be "acceptance"? Rather than naming it after how we're using it now, could we find a name that's more intrinsic to what it is?

jml:
Hey Tom,

You asked if I had any input. You don't need to follow up on what I say here, it's just food for thought.

I had thought that @wallrj was going to update our acceptance tests to use the new installer that he & Madhuri worked on. I personally find it difficult to reason about a change to what looks like a provisioning API (although is it?) when I don't really understand which of our several provisioning mechanisms it is for.

I'm probably barking up the wrong tree. Nevertheless, I hope this helps.

jmrl

PR #2684, opened 2016-03-05
[FLOC-4277] Update CentOS base image.

tomprince:
https://clusterhq.atlassian.net/browse/FLOC-4277

### DIFF
--- a/docs/installation/setup-aws.rst
+++ b/docs/installation/setup-aws.rst
@@ -22,17 +22,18 @@ You can also refer to `the full documentation for interacting with EC2 from Amaz
 .. note:: 
    If you want to choose a CentOS 7 AMI, you might not find it listed in the provided link. 
    
-   Use the links below to launch instances using CentOS 7 AMIs (specifically ``CentOS 7 x86_64 (2014_09_29) EBS HVM``):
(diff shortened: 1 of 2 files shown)


wallrj:
Thanks @tomprince 

I wondered if the documentation could also / instead link to:
 * https://cloud-images.ubuntu.com/locator/ec2/ and 
 * https://wiki.centos.org/Cloud/AWS#head-224024c7b3b083bd574bec6861bcdfd3487a5418 or
 * https://aws.amazon.com/marketplace/pp/B00O7WM7QW

I added a tool recently to get the latest Ubuntu AMIs. `./admin/ami-search-ubuntu`
I should extend that to get Centos-7 AMIs too. And maybe for other clouds.

Anyway, I checked all the links and they seem to work.
And the loopback acceptance tests all passed which demonstrates that the `flocker/provision/_aws.py
` AMIs also work.

Please merge.

wallrj:
And we should probably update @lukemarsden 's terraform template too:
 * https://github.com/ClusterHQ/unofficial-flocker-tools/blob/master/unofficial_flocker_tools/terraform_templates/variables.tf#L17

PR #2685, opened 2016-03-07
avoid recreating already deleted dataset

zhenghc:
when use p2p type storage backend. we notice that on primary node of deleted dataset. dataset will be created and then deleted constantly.

this exclude already deleted dataset in local_desired_datasets

### DIFF
--- a/flocker/node/_p2p.py
+++ b/flocker/node/_p2p.py
@@ -354,7 +354,8 @@ def find_dataset_changes(uuid, current_state, desired_state):
                             # datasets; this is wrong. See FLOC-2060.
                             in (node.manifestations or {}).values())
                         for node in current_state.nodes}
-    local_desired_datasets = desired_datasets.get(uuid, set())
+    local_desired_datasets = set(dataset for dataset in desired_datasets.get(uuid, set())
+                                 if dataset.deleted is False)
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks for the patch @zhenghc 

Unfortunately we no longer test or support this ZFS backend so it's going to be difficult to merge this change.
Can you use one of the block device backends instead?
 * https://docs.clusterhq.com/en/latest/flocker-features/storage-backends.html

And if not, we'd be interested to learn more about your ZFS use case and about how you plan to use Flocker.

wallrj:
Hey @zhenghc 

We discussed this internally and decided to do some manual testing of your patch.

I recreated the problem today and verified that your patch prevents recreation of deleted datasets.

Later this week I'll see if I can add a short unit test for the code. 
It should be simple to write a test for the calculate_changes logic without us having to install ZFS build slaves.
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/node/test/test_p2p.py#L355

Or if you have time to write test that would be greatly appreciated.

Thanks.

wallrj:
@zhenghc I've copied your contribution to another branch (https://github.com/ClusterHQ/flocker/pull/2689) so that I can run it through our build system.

Take a look, test the branch if you can and comment on that PR if you  have any suggestions.

PR #2687, opened 2016-03-08
[FLOC-4281] publish-installer-images now prints image ID map

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4281

Publishing to S3 is unnecessary complication.
I've removed all that code and replaced with printing the AMI map to stdout.

We're aiming for a process that looks something like this: #2672

Here's a sample output

```
(4281)(print-image-ids-FLOC-4281 ? =)[~/.../HybridLogic/flocker]$ DOCKER_VERSION=1.10.0       SWARM_VERSION=1.1.3       ./admin/publish-installer-images         --template=docker 2>stderr.log | tee stdout.log
{"us-west-1": "ami-916b18f1"}
(4281)(print-image-ids-FLOC-4281 ? =)[~/.../HybridLogic/flocker]$ jq . stdout.log 
{
  "us-west-1": "ami-916b18f1"
}

```

### DIFF
--- a/admin/installer/_images.py
+++ b/admin/installer/_images.py
@@ -355,7 +355,7 @@ def publish_installer_images_effects(options):
     )
     yield Effect(
         intent=StandardOut(
-            content=json.dumps(thaw(ami_map), encoding="utf-8")
+            content=json.dumps(thaw(ami_map), encoding="utf-8") + b"\n"
         )
     )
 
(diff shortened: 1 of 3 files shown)


myechuri:
Looks good. Thanks, @wallrj !

PR #2692, opened 2016-03-14
Split title on CloudFormation document [FLOC 4286]

davidoliverSP2:
Fixes 4286

Following feedback from the designers, this splits the long title on the CloudFormation doc into a main header and sub header.

### DIFF
--- a/docs/docker-integration/cloudformation.rst
+++ b/docs/docker-integration/cloudformation.rst
@@ -6,9 +6,12 @@
         .toctree-wrapper { display:none; }
     </style>
 
-================================================================
-Installing Flocker with Docker Swarm on AWS using CloudFormation
-================================================================
+====================================
+Installing Flocker with Docker Swarm 
+====================================
+
+on AWS using CloudFormation
+===========================
 
 In this guide you will learn how to quickly deploy a cluster of servers running Flocker and Docker Swarm.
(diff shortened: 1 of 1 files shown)


docs/docker-integration/cloudformation.rst
```diff
@@ -6,9 +6,12 @@
...
     </style>
 
-================================================================
-Installing Flocker with Docker Swarm on AWS using CloudFormation
-================================================================
+====================================
+Installing Flocker with Docker Swarm 
+====================================
+
+on AWS using CloudFormation
```
sarum90: Just noting that this is not the same split as on the image on the issue.

That split was:
```
Installing Flocker
with Docker Swarm on AWS using CloudFormation
```
Is there intentionally a difference in where you are splitting the line?

Also maybe use re-structured-tesxt's subtitle feature? http://docutils.sourceforge.net/docs/user/rst/quickstart.html#document-title-subtitle

Seems more representative of the intent of this text.

sarum90:
Looks good, Thanks!

Consider comment and merge whenever.

Not sure how to resolve the inconsistencies in the ticket description and the demo screenshot.

davidoliverSP2:
thanks @sarum90 - good spot on the subtitle syntax!

I've reverted the title to the original screenshot, as it looked a bit cleaner.

PR #2695, opened 2016-03-14
[FLOC-4300] Centos gce container agent fix floc 4300

sarum90:
We do not fully support the container agent anymore (and haven't for awhile),
so it is unlikely that we would even fix a bug identified by the container
agent's logs. Also, now that we start and stop the container agent a lot, there
appears to be some race conditions where the logs from the container agent are
not collected by flocker-diagnostics, ~~potentially because the container agent~~
~~has not logged it's first line of output~~ **because it is not yet reported by `initctl`**
by the time we run flocker-diagnostics.

Rather than diving in the deep end and figuring out the root cause of these
flakes, just allow the test to pass even if the container agent logs are
missing, since their presence adds little value.

### DIFF
--- a/flocker/acceptance/endtoend/test_diagnostics.py
+++ b/flocker/acceptance/endtoend/test_diagnostics.py
@@ -94,11 +94,11 @@ class DiagnosticsTests(AsyncTestCase):
             ])
             self.expectThat(
                 actual_basenames,
-                MatchesAny([
+                MatchesAny(
                     Equals(expected_basenames),
                     Equals(expected_basenames.union(
                         container_agent_basenames)),
-                ])
+                )
             )
 
         verifying = downloading.addCallback(verify_archive)
--- a/flocker/acceptance/endtoend/test_diagnostics.py
+++ b/flocker/acceptance/endtoend/test_diagnostics.py
@@ -99,7 +99,6 @@ class DiagnosticsTests(AsyncTestCase):
                     Equals(expected_basenames + container_agent_basenames),
                 ])
             )
-            self.assertEqual(expected_basenames, actual_basenames)
 
         verifying = downloading.addCallback(verify_archive)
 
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks @sarum90 

I reminded myself of the ``flocker-diagnostics`` code and it looks like only exports the logs of Flocker services that are enabled. 
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/node/diagnostics.py#L222

It determines available services on Ubuntu using ``initctl``.

The acceptance tests enable services using this script (run on the node)
 * https://github.com/ClusterHQ/flocker/blob/master/flocker/acceptance/node_scripts/enable_service.py#L15

So perhaps that isn't doing quite enough to ensure that ``flocker-container-agent`` is once again listed by ``initctl``.

On the other hand, the change you made to the test makes sense and is the expedient solution, given that we're removing the container agent...soon.

Please merge when green unless you have second thoughts about diving in.

wallrj:
Also strange that this should only happen on GCE nodes. Perhaps GCE Ubuntu images have some variation on standard Ubuntu service configuration.

sarum90:
Agreed about concerns about this only happening on GCE, and thanks for looking into how the diagnostics tool works.

I'm still inclined to just do the expedient solution in the interest of unblocking me so I can eventually get around to removing the container agent.

If GCE has some oddity with it's service configuration I would posit that it's a difference we don't really care about unless some of the other acceptance tests fail. I'll probably just put this in the list of things to check first when the tests fail rather than proactively investigate this oddity. I think there are other oddities in how flocker works that are higher on my list of things to investigate.

sarum90:
Also thanks for the review @wallrj! Nice to have someone else at least aware of the corner I'm cutting here.

sarum90:
Updated the above PR description with your investigation of flocker-diagnostics.

Running `admin/merge-pr` presently.

PR #2696, opened 2016-03-14
[FLOC-4255] Add documentation for Flocker deploy on RHEL 7.2

myechuri:
Add documentation for installing flocker node and flocker cli on RHEL 7.2. Relevant sections start at http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/add-rhel-documentation-FLOC-4255/docker-integration/manual-install.html

Setting up RHEL CI environment will follow this PR.

### DIFF
--- a/docs/docker-integration/install-node.rst
+++ b/docs/docker-integration/install-node.rst
@@ -21,6 +21,12 @@ Installing the Flocker Node Services
 * :ref:`aws-install-docker`
 * :ref:`rackspace-install-docker`
 
+.. _rhel-7-install-docker:
+
+.. include:: ../installation/install-node.rst
+   :start-after: .. begin-body-installing-node-rhel
+   :end-before: .. end-body-installing-node-rhel
+
 .. _centos-7-install-docker:
 
 .. include:: ../installation/install-node.rst
--- a/docs/installation/enabling-control-service.rst
+++ b/docs/installation/enabling-control-service.rst
@@ -32,10 +32,10 @@ On AWS, an external firewall is used instead, which will need to be configured s
 RHEL 7.2
 ========
 
-   .. prompt:: bash root@rhel:~$
+.. prompt:: bash root@rhel:~$
       
-      systemctl enable flocker-control
-      systemctl start flocker-control
+   systemctl enable flocker-control
+   systemctl start flocker-control
 
 The control service needs to be accessible remotely.
 You will need to configure FirewallD to allow access to the control service HTTP API and for agent connections.
--- a/docs/installation/enabling-control-service.rst
+++ b/docs/installation/enabling-control-service.rst
@@ -32,7 +32,7 @@ On AWS, an external firewall is used instead, which will need to be configured s
 RHEL 7.2
 ========
 
-.. prompt:: bash root@rhel:~$
+.. prompt:: bash [root@rhel]#
       
    systemctl enable flocker-control
    systemctl start flocker-control
--- a/docs/installation/enabling-control-service.rst
+++ b/docs/installation/enabling-control-service.rst
@@ -29,6 +29,27 @@ For more details on configuring the firewall, see the `FirewallD documentation`_
 
 On AWS, an external firewall is used instead, which will need to be configured similarly.
 
+RHEL 7.2
+========
+
+   .. prompt:: bash root@rhel:~$
+      
+      systemctl enable flocker-control
+      systemctl start flocker-control
+
+The control service needs to be accessible remotely.
+You will need to configure FirewallD to allow access to the control service HTTP API and for agent connections.
+Note that on some environments, in particular AWS, the ``firewalld`` package is not installed and the ``firewall-cmd`` program will not be found.
+If that is the case then just skip these commands.
+Otherwise run:
+
+.. task:: open_control_firewall centos-7
+   :prompt: [root@control-node]#
+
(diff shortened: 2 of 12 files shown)


docs/installation/install-node.rst
```diff
@@ -44,6 +44,54 @@ If you do not have any nodes, the following guides will help you set some up, wi
...
+      ssh root@<your-first-node>
+
+#. **Install the** ``clusterhq-flocker-node`` **package:**
+
+   To install ``clusterhq-flocker-node`` on RHEL 7 you must install the RPM package provided by the ClusterHQ repository.
+   The commands below will install the two repositories and the ``clusterhq-flocker-node`` package.
+   
+   Run the following commands as root on the target node:
+
+   .. prompt:: bash root@rhel:~$
```
davidoliverSP2: If this is run as root, should the prompt directive be `.. prompt:: bash [root@rhel]#`?

docs/installation/install-node.rst
```diff
@@ -44,6 +44,54 @@ If you do not have any nodes, the following guides will help you set some up, wi
...
+#. **Install the** ``clusterhq-flocker-docker-plugin`` **package:**
+
+   At this point you can choose to install the Flocker plugin for Docker.
+   Run the following command as root on the target node:
+
+   .. prompt:: bash [root@rhel]#
+   
+      yum install -y clusterhq-flocker-docker-plugin
+
+   .. XXX FLOC-3454 to create a task directive for installing the plugin
```
davidoliverSP2: I added indents here (and below where the same comment was used), as it was causing the numbering to get misaligned.

docs/installation/enabling-control-service.rst
```diff
@@ -29,6 +29,27 @@ For more details on configuring the firewall, see the `FirewallD documentation`_
 
 On AWS, an external firewall is used instead, which will need to be configured similarly.
 
+RHEL 7.2
+========
+
+.. prompt:: bash root@rhel:~$
```
davidoliverSP2: The other instructions in this document use prompt directives using root - should this be `.. prompt:: bash [root@rhel]#?`

davidoliverSP2:
Thanks @myechuri 

I added a few commits that cleaned up some indenting - I added an indent to a comment (which I don't think you added in this PR), and removed some indents from code blocks.

docs/installation/install-node.rst
myechuri: Thanks for pointing out! Fixed.

docs/installation/install-node.rst
myechuri: Thanks!

docs/installation/enabling-control-service.rst
myechuri: Fixed.

myechuri:
Thanks for the quick review, @davidoliverSP2 ! Addressed your review comments.

docs/docker-integration/install-node.rst
```diff
@@ -21,6 +21,12 @@ Installing the Flocker Node Services
 * :ref:`aws-install-docker`
 * :ref:`rackspace-install-docker`
 
+.. _rhel-7-install-docker:
```
davidoliverSP2: Does this need to be replicated in the Kubernetes and Standalone instructions? Currently the RHEL instructions for Installing the Node services are only appearing in the Docker integration instructions.

docs/installation/install-client.rst
```diff
@@ -57,6 +57,19 @@ On Ubuntu 14.04 (64-bit), the Flocker CLI can be installed from the ClusterHQ re
 
 .. begin-body-installing-client-linux
 
+Installing on RHEL 7.2
```
davidoliverSP2: These instructions have been embedded into the `installing-client-linux` section - I'm not sure if this was intentional - it probably makes sense unless there is a scenario where we would want to publish instructions for "other linux distributions" but not RHEL. 

As it stands, every version of our install instructions (Docker/Kubernetes/Standalone) all have RHEL in the installing the client, but not all of them have RHEL in the node instructions (see my previous comment)

davidoliverSP2:
Thanks @myechuri 

I've raised a few comments relating to whether the RHEL instructions need to appear for both installing the client and node services - currently they appear in all of the client instructions, but only the Docker integration for the node services.

I would suggest best practice is to label up the RHEL instructions in the `installation/install-client.rst` file as you have in the `installation/install-node.rst` document, then in each of the integration folders add an include directive. 

If you'd rather pass this over to me, I'm happy to pick it up - just move the Jira ticket to me!

docs/docker-integration/install-node.rst
myechuri: Thanks for pointing out, @davidoliverSP2 . Added.

myechuri:
> I would suggest best practice is to label up the RHEL instructions in the installation/install-client.rst file as you have in the installation/install-node.rst document, then in each of the integration folders add an include directive.

Thanks for suggesting, @davidoliverSP2 . Refactored client install instructions accordingly - definitely makes it cleaner.

myechuri:
@davidoliverSP2 : addressed your comments and verified that the staging results comply: http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/add-rhel-documentation-FLOC-4255

Please let me know if you have any further comments. Thanks!

PR #2701, opened 2016-03-15
[FLOC-4301] Add GCE node setup instructions, and reference GCE throughout the docs.

sarum90:
Went through the docs and added GCE to every location Rackspace was mentioned.

Along the way I realized we have a layout file that is no longer referenced (I think), so I just deleted it.

@davidoliverSP2 Would you mind confirming my docs changes, especially that it is safe to delete layout_homepage?

@justnoise, would you mind reviewing the GCE directions? Feel free to suggest rewording, I typed this up pretty quickly.


docs/_themes/clusterhq/layout.html
```diff
@@ -1,6 +1,4 @@
-{%- if meta is defined and meta.get('layout') == 'homepage' -%}
-{%- extends 'layout_homepage.html' -%}
```
wallrj: :+1: We discussed this and `layout_homepage.html` seems to be a legacy of previous documentation where the first documentation page was set up to look like part of the main website.

sarum90:
http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/add-gce-where-aws-is-FLOC-4301/ to review the docs.

docs/gettinginvolved/example-acceptance.yml
```diff
@@ -45,3 +50,7 @@ storage-drivers:
     username: "your.rackspace.username"
     api_key: "33333333333333333333333333333333"
     auth_url: "https://identity.api.rackspacecloud.com/v2.0"
+
+  gce:
+    project: "your-project-name"
+    zone: "us-central1-f"
```
wallrj: :question: I thought the zone was automatically detected in the GCE driver?

docs/gettinginvolved/plugins/building-driver.rst
```diff
@@ -12,6 +12,7 @@ These drivers include:
 
 * `OpenStack Cinder <https://github.com/ClusterHQ/flocker/blob/master/flocker/node/agents/cinder.py>`_
 * `Amazon EBS <https://github.com/ClusterHQ/flocker/blob/master/flocker/node/agents/ebs.py>`_
+* `Google Compute Engine PD <https://github.com/ClusterHQ/flocker/blob/master/flocker/node/agents/gce.py>`_
```
wallrj: :question: `PD`? I think I asked that before. Maybe say "Persistent disk".

docs/installation/setup-gce.rst
```diff
@@ -0,0 +1,31 @@
...
+You can get a Flocker cluster running using Google Compute Engine.
+You'll need to setup at least two nodes.
+
+#. Create a new cloud server:
+
+   * Visit the `Google Cloud Console <https://console.cloud.google.com/>`_.
+   * Navigate to the project you want to use.
+   * Use the menu in the upper left to navigate to ``Compute Engine``.
+   * Click ``Create Instance``.
+   * Choose a name and a zone for your instance. Zone must be the same for all instances.
```
wallrj: :question: If Adam was reviewing this he'd probably say "semantic line feeds."

docs/installation/setup-gce.rst
```diff
@@ -0,0 +1,31 @@
...
+   * Choose a machine type.
+   * For the Boot Disk, choose a supported Linux distribution (either CentOS 7 or Ubuntu 14.04) as your image.
+   * Check ``Allow API access to all Google Cloud services in the same project``.
```
justnoise: I _think_ they changed the look and feel of the instance creation page.  There appears to be an "Identity and API" box where the old trusty checkbox used to be.

wallrj:
Thanks @sarum90 

Looks good to me.
I didn't realise that include  the cloud specific instructions in each stack section.
If we're going to do that, it seems like we may as well just include the instructions directly rather than a links to an include page.
Not something to worry about here though.

Please merge.

wallrj:
Also a bit ugly that the search results end up with four copies of the GCE instructions:
 * http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/add-gce-where-aws-is-FLOC-4301/search.html?q=Setting+Up+Nodes+Using+Google+Compute+Engine&check_keywords=yes&area=default

justnoise:
On this page [http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/add-gce-where-aws-is-FLOC-4301/supported/index.html] we probably need to update the button/link for PD.  Under the "IaaS Block Storage" section.

docs/installation/setup-gce.rst
sarum90: Ah referencing http://rhodesmill.org/brandon/2012/one-sentence-per-line/

Interesting stuff. Although not the way I would want to solve that problem :)

I'd prefer having semantic `diff` programs. If I win the lottery I'll make one for C++ to show you what I'm talking about. Essentially, in my mind the problem is his diff program is unaware that newlines are not significant in rst files. Instead, he should use a diff program that knows that newlines are not significant. Similarly for something like C++, for some changes I would much rather review the AST diff than the actual code diff.

docs/gettinginvolved/example-acceptance.yml
sarum90: It is in the prod case (where this is only consumed by dataset agents), but this is the `acceptance.yml`, so it is also consumed the trial process on your laptop which needs project/zone because it also attempts to interact with GCE block devices.

docs/gettinginvolved/plugins/building-driver.rst
sarum90: I always try to mimic Amazon EBS but no one seems to buy it.

Updated to see Persistent Disk.

sarum90:
+1 about the multiple search results being confusing, but I think that's true for all of our node setup instructions these days.

PR #2703, opened 2016-03-16
[FLOC-3404] Update cffi and cryptography dependencies

sarum90:
This fixes installing on OSX which is currently broken on El Capitan.

Verified fix manually in a VirtualBox VM on my machine. (Built an sdist, uploaded it to S3, made a new tap ruby file, uploaded that to S3, installed using brew from that file).

I do not know why the automated OSX builder did not catch this. If I was crossing all of my t's and dotting all of my i's I would look into it, but I think that's probably lower priority than getting the fix in (and some of the other work I'm working on presently).

### DIFF
--- a/docs/releasenotes/index.rst
+++ b/docs/releasenotes/index.rst
@@ -16,6 +16,7 @@ Next Release
   This could lead to the appearance of data loss, as different volumes get used.
   Now, even if multiple volumes are created, only a single volume will be used.
   This was particularly likely to occur on AWS.
+* Fixed brew tap for flocker client tools on OSX, which had regressed on Yosemite.
 
 This Release
 ============
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,9 +6,9 @@ bitmath==1.2.3-4
 boto==2.38.0
 boto3==1.2.2
 botocore==1.3.17
-cffi==1.1.2
+cffi==1.5.2
 characteristic==14.3.0
-cryptography==1.1.2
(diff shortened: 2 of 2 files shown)


sarum90:
Just realized our CI is named in such a way that suggests that it is running Yosemite. So... that might be why it broke on El Capitan without us noticing.

zubron:
Thanks @sarum90! Looks good. I just installed flocker using the homebrew recipe you referenced in the issue and it installed successfully.

Given that this won't update the existing recipe, do we need to do another release to fix OS X installation or can we just fix it manually?

sarum90:
Oh I hadn't thought about fixing it manually, I was just going to wait for the next release.

Which reminds me to add this to release notes...

sarum90:
@wallrj and I just discussed. We think we'll wait for the next release.

Seems sort of evil to either A) change our docs to reference the tap recipe I made. or B) manually change the sdist of Flocker 1.10.2 so that it is different than the linux packages.

zubron:
Yeah, that's probably for the best! I do think that we should probably release 1.10.3 fairly soon to fix this though.

sarum90:
Yea I think next Tuesday will be a proper release.

sarum90:
Considering the release notes will be reviewed as part of the release process, I'm going to merge this even though I haven't technically gotten a review of the release notes yet.

zubron:
Sounds good! Thanks again! :)

PR #2706, opened 2016-03-17
[FLOC-4215] Separate GCE operations (VolumeManager) from the GCE driver.

sarum90:
Also test that when operations are duplicated errors are thrown. This caught some errors that existed in some detachVolume code I wrote in a different branch, because that code is always executed in this test.

This is being done in preparation of making this driver threadsafe. The the locking will only have to be done inside the VolumeManager, which is nice.


flocker/node/agents/functional/test_gce.py
```diff
@@ -126,6 +141,56 @@ def test_attach_elsewhere_attached_volume(self):
         pass
 
 
+class _RepeatProxy(object):
+    """
+    Implementation of a proxy for an interface that calls each method of the
+    underlying interface twice in a row, and returns the value from the first
```
justnoise: The code seems to be returning the value from the second call.

sarum90: Nice catch, will fix.

flocker/node/agents/gce.py
```diff
@@ -822,6 +724,249 @@ def _stop_node(self, node_id):
 
         :param unicode node_id: The compute_instance_id of the node to stop.
         """
+        self._volume_manager.stop_node(node_id)
+
+
+class IGCEVolumeManager(Interface):
```
justnoise: :question: We call this a GCEVolumeManager but it has operations in the interface that operate on nodes.  Is this named correctly?

flocker/node/agents/functional/test_gce.py
```diff
@@ -322,13 +388,91 @@ def test_attach_elsewhere_attached_volume(self):
             attach_to=api.compute_instance_id(),
         )
 
+    def test_duplicated_calls(self):
+        """
+        Verify that if every call to the :class:`GCEAtomicOperation` s is
```
justnoise: Hey there Lone Wolf 'S'! What are you doing in this docstring?

flocker/node/agents/functional/test_gce.py
```diff
@@ -322,13 +388,91 @@ def test_attach_elsewhere_attached_volume(self):
...
+        )
+
     def test_list_volumes_walks_pages(self):
         """
         Ensure that we can walk multiple pages returned from listing GCE
         volumes.
         """
         api = gceblockdeviceapi_for_test(self)
-        self.patch(api, '_page_size', 1)
+        api = api.set('_page_size', 1)
```
justnoise: Just curious: set == patch?  Couldn't find this by grepping testtools or flocker.

flocker/node/agents/functional/test_gce.py
```diff
@@ -322,13 +388,91 @@ def test_attach_elsewhere_attached_volume(self):
...
+        )
+
+        self.assertThat(
+            api.get_device_path(volume.blockdevice_id).path,
+            Contains('/dev/sd')
+        )
+
+        api.detach_volume(
+            blockdevice_id=volume.blockdevice_id,
+        )
```
justnoise: :grey_question: possibly add a comment that detach_volume isn't in the interface.  Not sure what our commenting guidelines are but I had to double check to make sure this wasn't calling it 4 times.

flocker/node/agents/gce.py
```diff
@@ -794,12 +704,8 @@ def list_live_nodes(self):
...
         nodes = []
         while not done:
-            result = self._compute.instances().list(
-                project=self._project,
-                zone=self._zone,
-                maxResults=self._page_size,
-                pageToken=page_token
-            ).execute()
+            result = self._volume_manager.list_nodes(page_token,
+                                                        self._page_size)
```
justnoise: Indentation is funky

flocker/node/agents/functional/test_gce.py
sarum90: I have a feeling your reviews will be the most fun.

flocker/node/agents/functional/test_gce.py
sarum90: so since the API is now a PClass you can't change it, but you can create a new instance that is identical _except_ that it has the `_page_size` attribute set to 1 instead of whatever it previously was set to. `.set()` is pyrsistent's syntactic sugar on that operation. You don't have to worry about setting it back to something else at the end of the test because, well, you have a new instance of the object that only you have a reference to. The old one is still available for whomever else was using it (although in this case the old one has no more references after line 475 (I think pyrsistent has some magic that makes the operation efficient in this case)).

flocker/node/agents/gce.py
```diff
@@ -822,6 +724,249 @@ def _stop_node(self, node_id):
...
+    def create_disk(self, name, size, description, gce_disk_type):
+        sizeGiB = int(size.to_GiB())
+        config = dict(
+            name=name,
+            sizeGb=sizeGiB,
+            description=description,
+            type="projects/{project}/zones/{zone}/diskTypes/{type}".format(
+                project=self._project, zone=self._zone, type=gce_disk_type)
+        )
+        # TODO(mewert): Verify timeout and error conditions.
```
justnoise: Do we need to do more verification of timeouts?

flocker/node/agents/functional/test_gce.py
sarum90: Oh actually this was my bad from testing. This should be removed... and I probably have something else to fix now...

Thanks for pointing this out.

flocker/node/agents/gce.py
sarum90: Heh, originally that was named AtomicOperations, and then I changed to VolumeManager on request from Richard because that maps to what we call a similar object in the Cinder driver. But your right, this has more than just volume operations. Renaming to just plain old GCEOperations.

flocker/node/agents/gce.py
sarum90: merging is hard.

admin/merge_pr.py
```diff
@@ -166,7 +166,7 @@ def format_status(status):
     """
     github, jenkins = status
     return u"{}: {} ({})".format(
-        github['context'], jenkins.value, github['target_url'])
```
sarum90: This change shouldn't be here....

justnoise:
:+1: Better late than never

sarum90:
1) Only the cinder test is pending anymore.
2) This code does not touch the cinder driver.
3) I have another PR I want to send out that depends on this one.

Conclusion. I'm merging this as is. Thanks for the review @justnoise!

PR #2708, opened 2016-03-18
[FLOC-4309] Gce blockdevice threadsafe.

sarum90:
In reviewing logs, @justnoise and I noticed that we had some concurrent GCE requests that appeared to come back mangled (or hang abnormally long).

The GCE python bindings are based on httplib2, which is not threadsafe, so we either need to create 1 backend per thread or add some other serialization method.  To me it seemed most simple to just add a mutex around all calls to the GCE API.

I'm still debugging some failures that I'm seeing in the acceptance tests, but it seems as if this change does not make thing worse, so I consider this ready for review.

### DIFF
--- a/flocker/acceptance/testtools.py
+++ b/flocker/acceptance/testtools.py
@@ -911,6 +911,13 @@ class Cluster(PClass):
                 d.addBoth(
                     lambda _: async_api.destroy_volume(volume.blockdevice_id)
                 )
+                # Consume failures and write them out. Failures might just
+                # indicate that the cluster beat us to deleting the volume
+                # which should not cause the test to fail, we only want the
+                # test to fail if list_volumes still returns a non-empty list.
+                # The construction of api_clean_state should prevent this from
+                # masking significant failures to clean up volumes.
+                d.addErrback(write_failure)
                 return d
 
             cleaning_volumes = api_clean_state(
--- a/flocker/node/agents/gce.py
+++ b/flocker/node/agents/gce.py
@@ -656,8 +656,13 @@ class GCEBlockDeviceAPI(PClass):
 
                 if potentially_detaching_error is not None:
                     try:
+                        # We want to poll until _get_attached_to no longer
+                        # returns a compute_instance_id, and instead raises an
+                        # UnattachedVolume exception.
                         poll_until(
-                            lambda: self._get_attached_to(blockdevice_id),
+                            lambda: not bool(
+                                self._get_attached_to(blockdevice_id)
+                            ),
                             [1] * VOLUME_DETATCH_TIMEOUT
                         )
                         raise GCEVolumeException(
--- a/flocker/node/agents/gce.py
+++ b/flocker/node/agents/gce.py
@@ -899,7 +899,7 @@ class GCEOperations(PClass):
 
         # A custom sleep function that drops the lock while the actual sleeping
         # is going on.
-        def custom_sleep(*args, **kwargs):
+        def lock_dropped_sleep(*args, **kwargs):
             self._lock.release()
             try:
                 return sleep(*args, **kwargs)
@@ -912,7 +912,7 @@ class GCEOperations(PClass):
         try:
             operation = function(**args).execute()
             return wait_for_operation(
-                self._compute, operation, [1]*timeout_sec, sleep)
+                self._compute, operation, [1]*timeout_sec, lock_dropped_sleep)
         finally:
             self._lock.release()
 
--- a/flocker/node/agents/gce.py
+++ b/flocker/node/agents/gce.py
@@ -11,7 +11,9 @@ driver:
 - Python API: https://google-api-client-libraries.appspot.com/documentation/compute/v1/python/latest/ # noqa
 - Python Oauth: https://developers.google.com/identity/protocols/OAuth2ServiceAccount#authorizingrequests # noqa
 """
+import time
 from uuid import UUID
+from threading import Lock
 
 import requests
 from bitmath import GiB, Byte
@@ -853,6 +855,7 @@ class GCEOperations(PClass):
     _compute = field(mandatory=True)
     _project = field(type=unicode, mandatory=True)
     _zone = field(type=unicode, mandatory=True)
+    _lock = field(mandatory=True, initial=Lock())
 
     def _do_blocking_operation(self,
                                function,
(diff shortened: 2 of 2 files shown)


flocker/node/agents/gce.py
```diff
@@ -891,11 +894,27 @@ def _do_blocking_operation(self,
...
         args = dict(project=self._project, zone=self._zone)
         args.update(kwargs)
-        operation = function(**args).execute()
-        return wait_for_operation(
-            self._compute, operation, [1]*timeout_sec, sleep)
+        self._lock.acquire()
+        try:
+            operation = function(**args).execute()
+            return wait_for_operation(
+                self._compute, operation, [1]*timeout_sec, sleep)
```
sarum90: Crap.. I should probably use that fancy custom sleep function I wrote...

sarum90:
Apologies for just tagging on some acceptance test fix-ups along the way. Iterating pretty hard on acceptance tests today. Send me a ping when you start reviewing so I know to stop pushing little fixes.

flocker/node/agents/gce.py
```diff
@@ -891,11 +899,27 @@ def _do_blocking_operation(self,
...
-        operation = function(**args).execute()
-        return wait_for_operation(
-            self._compute, operation, [1]*timeout_sec, sleep)
+        self._lock.acquire()
+        try:
+            operation = function(**args).execute()
+            return wait_for_operation(
+                self._compute, operation, [1]*timeout_sec, lock_dropped_sleep)
+        finally:
+            self._lock.release()
```
wallrj: :grey_question: The code would be a littler shorter (clearer?) if you used a `with` statement instead of `try...finally...`
 * https://docs.python.org/2/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement

flocker/node/agents/gce.py
```diff
@@ -891,11 +899,27 @@ def _do_blocking_operation(self,
         :returns dict: A dict representing the concluded GCE operation
             resource.
         """
+        if sleep is None:
+            sleep = time.sleep
+
+        # A custom sleep function that drops the lock while the actual sleeping
+        # is going on.
```
wallrj: :grey_question: Maybe make this comment a docstring of the nested function.

flocker/acceptance/testtools.py
```diff
@@ -911,6 +911,13 @@ def detach_and_destroy_volume(volume):
                 d.addBoth(
                     lambda _: async_api.destroy_volume(volume.blockdevice_id)
                 )
+                # Consume failures and write them out. Failures might just
+                # indicate that the cluster beat us to deleting the volume
+                # which should not cause the test to fail, we only want the
+                # test to fail if list_volumes still returns a non-empty list.
+                # The construction of api_clean_state should prevent this from
+                # masking significant failures to clean up volumes.
+                d.addErrback(write_failure)
```
wallrj: :grey_question: Sounds correct. But does this also apply to non GCE tests? I remember you saying something about instantiating another copy of the GCE driver in the TestCase in order to do cleanup, but I don't think we do that for other drivers.

flocker/node/agents/gce.py
```diff
@@ -654,8 +656,13 @@ def detach_volume(self, blockdevice_id):
 
                 if potentially_detaching_error is not None:
                     try:
+                        # We want to poll until _get_attached_to no longer
+                        # returns a compute_instance_id, and instead raises an
+                        # UnattachedVolume exception.
```
wallrj: :+1: Great. Nice fix.

wallrj:
Thanks @sarum90 

It looks great. 

 * All the GCE API calls in `_operations` are now performed with a lock.
 * I wondered if the calls to the metadata service needed some locks too, but they all seem to be performed during driver initialization.
 * I wondered if the lock.acquire should be done with a timeout, to escape from possible deadlock situations, but we discussed and couldn't think of any.
 * The non-thread safety related bug fixes all make sense.

wallrj:
One parting thought that we discussed yesterday.
The Cinder driver also has an Interface wrapping the Cloud API operations so perhaps we can implement a general wrapper that  locks the operations for all drivers.
I don't think the EBS driver has an interface for the boto operations, yet, but could easily.

flocker/acceptance/testtools.py
sarum90: Ah, yea this is relatively new (I think it was added in the past month, perhaps as part of the acceptance test for the AWS duplicate dataset bug?). I think this might be occurring in GCE more than in other drivers because GCE sometimes has some very slow operations, but I think we now do this for all backends.

It _might_ have something to do with the dataset flickering over the flocker API though....
Sounds like people have seen that on non-GCE though.

I believe:

- This cleanup code is new, and not yet hardened (although it runs on all backends).
- It may or may not have similar failure modes on EBS and Cinder, but it depends on a lot on the details of how the endpoint API behaves while being queried from multiple nodes about the same dataset, and I'm not sure if I see too much value in looking at those details in the short term.
- Catching errors here is still correct, as we should never get out of this cleanup method until list_volumes on the API returns an empty list.
- This change hardens the cleanup code in all cases.

flocker/node/agents/gce.py
sarum90: Thanks yea, sorry about the original bug.
I guess this might be an argument for having a mock version of GCEOperations (so I could unit test this logic)... although the failure modes here don't affect production flocker in negative ways, just the cleanup code in our acceptance tests.

flocker/node/agents/gce.py
sarum90: Changed.

flocker/node/agents/gce.py
sarum90: Nice catch! yea much better.

I still have to do `try... finally...` for the lock_dropped_sleep, but all others changed to use with.

sarum90:
Thanks for the review Richard!

+1 to using an interface wrapper. Added a comment to https://clusterhq.atlassian.net/browse/FLOC-4305

Might be complicated by the _do_blocking_operation in GCE. Perhaps that should be factored back up into the higher level somehow...

For now I think I'll keep with the manually locking here that only surrounds calls to execute() (+ the slightly more complicated logic around sleeping).

Going to run merge-pr now. Thanks!

PR #2709, opened 2016-03-19
[FLOC-4310] Increase all acceptance timeouts.

sarum90:
GCE unfortunately can be very slow at doing volume operations (specifically attach and detach) (O(90 seconds)). Most of our tests have to do 2 operations in series (First they have to delete old volumes, then the have to create a volume for their own test). Thus, it sort of makes sense to me to have a much larger default timeout on our acceptance tests (I had been going through 1 by 1 and identified tests that needed a longer timeout experimentally, but after seeing 1 test that had an 80 second detach in cleanup followed by a 70 second attach in the actual test itself, it seems like this will be a quicker solution.)

Decided to make the change in three places:

1) test_dataset
2) test_container
3) test_dockerplugin

removed old timeout extensions that were less than 5 minutes.


sarum90:
Against the wrong branch

PR #2710, opened 2016-03-19
[FLOC-4310] Increase all acceptance test timeouts.

sarum90:
GCE unfortunately can be very slow at doing volume operations (specifically attach and detach) (O(90 seconds)). Most of our tests have to do 2 operations in series (First they have to delete old volumes, then the have to create a volume for their own test). Thus, it sort of makes sense to me to have a much larger default timeout on our acceptance tests (I had been going through 1 by 1 and identified tests that needed a longer timeout experimentally, but after seeing 1 test that had an 80 second detach in cleanup followed by a 70 second attach in the actual test itself, it seems like this will be a quicker solution.)

Decided to make the change in three places:

1) test_dataset
2) test_container
3) test_dockerplugin

removed old timeout extensions that were less than 5 minutes.

### DIFF
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -27,7 +27,7 @@ from ...node import backends
 from ..testtools import (
     require_cluster, require_moving_backend, create_dataset,
     skip_backend, get_backend_api, verify_socket,
-    get_default_volume_size,
+    get_default_volume_size, ACCEPTANCE_TEST_TIMEOUT
 )
 
 
@@ -35,6 +35,8 @@ class DatasetAPITests(AsyncTestCase):
     """
     Tests for the dataset API.
     """
+
+    run_test_with = async_runner(timeout=ACCEPTANCE_TEST_TIMEOUT)
 
     @flaky(u'FLOC-3207')
     @require_cluster(1)
@@ -210,8 +212,6 @@ class DatasetAPITests(AsyncTestCase):
 
     @flaky(u'FLOC-3341')
     @require_moving_backend
-    # GCE sometimes takes 1 full minute to detach a disk.
-    @run_test_with(async_runner(timeout=timedelta(minutes=4)))
     @require_cluster(2)
     def test_dataset_move(self, cluster):
         """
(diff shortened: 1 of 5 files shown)


sarum90:
Soooo much confidence in this change.

I made 3 branches with the changes from this branch so that I could run the GCE acceptance tests in parallel:

http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/increase-all-acceptance-FLOC-4310/
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/gce-blockdevice-threadsafe-FLOC-4309-2/
http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/gce-blockdevice-threadsafe-FLOC-4309-3/

You can look through the history of the GCE acceptance tests on all of those branches and you should find ~10 passes in each of the branches, and a few failures.

The remaining failures are:

- Sometimes on CentOS, we get an error about needed a tty to run Sudo. We try to workaround this by using a startup script, but evidently that does not always work. This does *not* indicate and production code issue, and might be fixed by @justnoise in the near future (as part of some other work he is working on), so I just hit retry and don't worry about it.
- Sometimes Jenkins kills slaves mid test run. Again obviously *not* a production code issue, so I hit retry and move on with my life. I anticipate this will go away as we make CI improvements sometime.

justnoise:
This looks good to me.  One question: Since increasing timeouts can hide other bugs (but is necessary for GCE), do you think it would make engineering sense to increase this timeout ONLY for GCE?  This could be done by either using a fancier decorator or through some other means.

sarum90:
Agreed that we might want to do this only for GCE in order to not mask other bugs. However, I don't think the added complexity (there are already so many ways to set the timeout) is worth the risk of hiding other bugs.

Also, it might not be the case that this is only needed for GCE. There are quite a few existing flocker bugs around timeouts, some of which might be fixed by this change (Thanks to @wallrj for compiling this list):

Some other acceptance test timeout issues:
https://clusterhq.atlassian.net/browse/FLOC-3196
https://clusterhq.atlassian.net/browse/FLOC-3207
https://clusterhq.atlassian.net/browse/FLOC-3341
https://clusterhq.atlassian.net/browse/FLOC-1522

For now I think I'll proceed on the risky ground of increasing the timeout across the board. If we find a bug that this is masking later I'll invest the time in to have different timeouts for the different providers.

justnoise:
:+1:

PR #2714, opened 2016-03-22
Acceptance Test Provisioner: SSH Client spawns a tty

justnoise:
This is the first in a set of two PRs that'll allow the acceptance tests to run as the default user.  Running acceptance tests without a tty cause flakes on rackspace.  Running the tests as the root user requires us to modify the root account and sshd.  This has slowed down development on GCE.

In the future these changes will be used to allow the provisioning code to be reused to build a better flocker installer.

### DIFF
--- a/flocker/provision/_ssh/_conch.py
+++ b/flocker/provision/_ssh/_conch.py
@@ -26,7 +26,7 @@ from twisted.conch.endpoints import (
 from twisted.conch.ssh.common import NS
 import twisted.conch.ssh.session as session
 from twisted.conch.client.knownhosts import KnownHostsFile
-from twisted.internet.defer import Deferred, inlineCallbacks
+from twisted.internet.defer import Deferred, inlineCallbacks, CancelledError
 from twisted.internet.endpoints import UNIXClientEndpoint, connectProtocol
 from twisted.internet.error import ConnectionDone
 from twisted.protocols.basic import LineOnlyReceiver
@@ -100,7 +100,7 @@ class CommandChannelWithTTY(_CommandChannel):
         Create a pty by sending a pty-req to the server
         """
         term = 'xterm'
-        winSize = (25,80,0,0)
+        winSize = (25, 80, 0, 0)
         ptyReqData = session.packRequest_pty_req(term, winSize, '')
         self.conn.sendRequest(self, 'pty-req', ptyReqData)
         command = self.conn.sendRequest(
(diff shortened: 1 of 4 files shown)


flocker/provision/_ssh/_conch.py
```diff
@@ -20,10 +20,13 @@
     _NewConnectionHelper,
     # https://twistedmatrix.com/trac/ticket/7862
     _ReadFile, ConsoleUI,
```
wallrj: :question: I was wondering if there's a public API that we could use to request a PTY.

I read:
 *  http://as.ynchrono.us/2011/08/twisted-conch-in-60-seconds-protocols.html


which talks about overriding `request_pty_req` on the session. Is that an option here?

flocker/provision/_ssh/_conch.py
```diff
@@ -87,6 +90,59 @@ def lineReceived(self, line):
...
+    Subclass that spawns a TTY when connecting.
+    """
+    def _executeCommand(self, connection, protocolFactory):
+        """
+        Given a secured SSH connection, try to execute a command in a new
+        channel created on it and associate the result with a protocol from the
+        given factory.
+
+        @param connection: See L{SSHCommandClientEndpoint.existingConnection}'s
+            C{connection} parameter.
```
wallrj: :question: Why use Twisted style docstrings? So that we can contribute this upstream? In which case maybe add a Twisted ticket reference.

flocker/provision/_ssh/_conch.py
```diff
@@ -87,6 +90,59 @@ def lineReceived(self, line):
...
+
+        @param protocolFactory: See L{SSHCommandClientEndpoint.connect}'s
+            C{protocolFactory} parameter.
+
+        @return: See L{SSHCommandClientEndpoint.connect}'s return value.
+        """
+
+        def disconnectOnFailure(passthrough):
+            # Close the connection immediately in case of cancellation, since
+            # that implies user wants it gone immediately (e.g. a timeout):
```
wallrj: :question: Comment could just as well be a docstring on the nested function.

wallrj:
Thanks @justnoise 

* I tested the branch with the standalone script that you supplied and saw that the sudo command worked. 
* Also noted that the script fails when tested against master.

* I attempted to run a sudo command using my openssh client and saw `sudo: sorry, you must have a tty to run sudo`

* Then tried again with `ssh -t` (which seems to be equivalent to the feature in this branch) and the sudo command worked.

I left a few comments above. Feel free to address those in followup branches though, since you've got a rare green build on this PR at the moment.

Merge when you're happy with everything.

flocker/provision/_ssh/_conch.py
justnoise: This is actually copy-pasted directly from SSHCommandClientEndpoint's implementation.  I could delete the docstring.  Do you think that would be better?

flocker/provision/_ssh/_conch.py
justnoise: Once again, just copy and pasted from twisted's implementation.

PR #2720, opened 2016-03-25
[FLOC-4322] Benchmarking speedups end of March

sarum90:
This branch includes all of the benchmarking speedups that have been done so far this week.

This includes:

1) Removing logging (It was expensive and it was removed. We hypothesize debugging might be harder now)
2) Using a forked branch of Pyrsistent (Thanks @zubron and @moshez for setting that up!)
3) Some twiddles around not calling pyrsistent.set() if we don't have to (actually acknowledging that it is an expensive operation)
4) Decreasing the poll time of the container agent.
5) Modifying flocker to be compatible with the latest version of pyrsistent (which our custom fork is based off of). (This mostly includes working around https://github.com/tobgu/pyrsistent/issues/77 putting all callable initial values inside other callables, although PClass.evolver().data was also changed to PClass._to_dict() as the former API was removed).
6) Adding the ability to profile the container and the dataset agents.
7) Adding some caches here and there (A couple of them are only of size 1, because they are merely to address responding quickly to requests we've already had recently).
8) Coalescing "PClass.set(x=1).set(y=2) into PClass.set(x=1, y=2)".
9) Turning on pyrsistent's C extension.
10) Fixing a bug were we would poll the dataset agent every 3 seconds if there was a deleted dataset in the cluster. It appears as if deleted datasets are wrongly observed by flocker to be "REGISTERED", which is supposed to indicate "Some node has created a dataset, and communicated that to me using the control service, but the cloud API does not report it to me, I assume this means I'm on AWS, and AWS is lying to me due to eventual consistency." Prior to this PR, for datasets that are not related to the current node, we would decide to wake up in 3 seconds to see if AWS was telling the truth yet. In practicality, I don't think there is any reason to wake up in 3 seconds to respond to datasets that are unrelated to us (desired state of non-manifest), so changed an entry in a table that makes that go back to the 60 second sleep.
11) Turned a list into a set turning a quadratic algorithm into a linear one.

Shout outs to @zubron, @moshez, and @wallrj  for helping with all of these fix-ups.


sarum90:
Well, looks like I've got a bunch of change detector tests to fix up tomorrow.

flocker/node/script.py
```diff
@@ -66,6 +68,40 @@ def flocker_dataset_agent_main():
     agent_script = AgentScript(service_factory=service_factory.get_service)
     options = DatasetAgentOptions()
 
+    # Use CPU time instead of wallclock time.
+    # The control service does a lot of waiting and we do not
```
zubron: This was copied (somewhat in a rush) from `flocker/control/script.py` and was going to be included as part of [FLOC-3142](https://clusterhq.atlassian.net/browse/FLOC-3142). Do you think we should just tidy this up here or include it as part of a different PR?

sarum90: I'm fine with either here or in a different PR. +1 to cleaning up. Will address before merging.

Please add all sorts of comments like this to this PR; most of my making of this PR was:

* I want to have tests running on all these changes.
* I want to have a single location for people to add TODOs before we merge this (as github comments).

build.yaml
```diff
@@ -318,6 +318,7 @@ common_cli:
 
     # using the caching-layer, install all the dependencies
     pip install --upgrade pip
+    pip install -r requirements.txt ${PIP_ADDITIONAL_OPTIONS}
```
wallrj: Add an explanatory comment here. Why is it not sufficient to run the next command which parses `setup.py` which includes all of `requirements.txt`?

build.yaml
```diff
@@ -318,6 +318,7 @@ common_cli:
...
     pip install . ${PIP_ADDITIONAL_OPTIONS}
     pip install ".[dev]" ${PIP_ADDITIONAL_OPTIONS}
```
wallrj: :question: Unrelated to this branch but why do we install dev requirements in separate step? 

And I was going to add "And in my experience this command doesn't work anyway. It has to be `pip install Flocker[dev]`."

But it seem newer `pip`s handle it.

flocker/__init__.py
```diff
@@ -11,27 +11,6 @@
...
-    performance. Once we have benchmarking framework we can assess
-    risk/benefit ratio better.
-
-    The mechanism for disabling extensions is documented at
-    https://github.com/tobgu/pyrsistent/blob/master/CHANGES.txt#L17-L19.
-    """
-    import os
-    os.environ[b"PYRSISTENT_NO_C_EXTENSION"] = b"1"
-_disable_pyrsistent_c_extensions()
-del _disable_pyrsistent_c_extensions
```
wallrj: :question: Do we still think this gave a significant speedup? ISTR it did, but not as much as other Python fixes. Perhaps we should disable it again to avoid the risk of segfaults in the field.

@itamarst or @exarkun the Pyrsistent changelog says segfaults were fixed in 0.11.7. Did we consciously keep the c extension disabled after we upgraded to 0.11.9?

wallrj: https://pyrsistent.readthedocs.org/en/latest/changes.html

flocker/control/_model.py
```diff
@@ -1128,11 +1128,14 @@ def update_node(self, node_state):
         [original_node] = nodes
         updated_node = original_node.evolver()
         for key, value in node_state.items():
-            if value is not None:
```
wallrj: :question: We should add a comment explaining this optimization. So that future maintainers don't revert it.
We should also create a Pyrsistent issue suggesting that `PClass.set` does this check by default and returns `self` if not modified....although I suppose you might be supplying a value which is equal but which is not a reference to the same existing value....in which case the new value should replace the old.

flocker/control/_model.py
```diff
@@ -1128,11 +1128,14 @@ def update_node(self, node_state):
...
+            if value is not None and value != updated_node[key]:
                 updated_node = updated_node.set(key, value)
-        return self.set(
-            "nodes", self.nodes.discard(original_node).add(
-                updated_node.persistent()))
+        updated_node = updated_node.persistent()
```
wallrj: :question: Add an explanatory comment here too.

flocker/control/_persistence.py
```diff
@@ -173,7 +173,7 @@ def default(self, obj):
             result[_CLASS_MARKER] = obj.__class__.__name__
             return result
         elif isinstance(obj, PClass):
-            result = obj.evolver().data
```
wallrj: :question: Create an upstream issue about the usefulness of this being a public method and link to it here.

flocker/control/_persistence.py
```diff
@@ -316,10 +316,14 @@ def update_leases(transform, persistence_service):
         persistence service has saved.
     """
     config = persistence_service.get()
-    new_config = config.set("leases", transform(config.leases))
-    d = persistence_service.save(new_config)
-    d.addCallback(lambda _: new_config.leases)
-    return d
```
wallrj: :question: Comment explaining the optimization.

flocker/control/httpapi.py
```diff
@@ -134,6 +136,29 @@ def render_if_matches(self, request, **route_arguments):
     return render_if_matches
```
wallrj: I suggest we move the optimizations in this module to a separate PR and add some explanatory comments justifying  our use of caching.

flocker/node/_container.py
```diff
@@ -570,17 +570,13 @@ def _restart_for_application_change(
...
         # Check volumes separately.
-        comparable_state = state.set(volume=None)
-        comparable_configuration = configuration.set(volume=None)
-
         # Restart policies don't implement comparison usefully.  See FLOC-2500.
-        restart_state = comparable_state.restart_policy
-        comparable_state = comparable_state.set(restart_policy=RestartNever())
-        comparable_configuration = comparable_configuration.set(
-            restart_policy=RestartNever()
-        )
```
wallrj: :question: We could include this part in a datastructure-optimizations branch along with the similar changes above and below.

flocker/node/_container.py
```diff
@@ -714,4 +710,4 @@ def calculate_changes(self, desired_configuration, current_cluster_state,
             phases.append(in_parallel(changes=start_restart))
 
         return sequentially(changes=phases,
-                            sleep_when_empty=timedelta(seconds=1))
```
wallrj: :question: And this can be in a separate PR. I think this was one of the more significant optimizations. So add an explanatory comment too.

flocker/node/_loop.py
```diff
@@ -379,21 +379,12 @@ class ConvergenceLoopOutputs(Names):
     [_FIELD_CONNECTION, _FIELD_LOCAL_CHANGES], [],
     u"Send the local state to the control service.")
```
wallrj: :question: Shall we add logging changes to a separate PR and create followup issues for adding the logging back, but with limits on number of datasets and containers logged. I think we still get a log of the updated configuration in the flocker-control logs, right? And when DeploymentState changes. So it shouldn't hinder our debugging too badly while we figure out the right thing to do.

flocker/node/agents/blockdevice.py
```diff
@@ -1577,11 +1577,9 @@ def from_state_and_config(discovered_dataset, desired_dataset):
         Discovered.NON_EXISTENT: CreateBlockDeviceDataset,
         # Other node will detach:
         Discovered.ATTACHED_ELSEWHERE: DoNothing,
-        # Some backends (AWS in particular) are only eventually consistent, so
-        # list_volumes might not report the volume even after it has already
-        # been created. We poll waiting for the backend to start reporting the
-        # created volume.
-        Discovered.REGISTERED: Poll,
```
wallrj: :question: Let's add this to a separate PR and issue.

flocker/node/agents/blockdevice.py
```diff
@@ -1961,7 +1959,7 @@ def _calculate_desired_state(
         # `ATTACHED_ELSEWHERE` need the same behavior as `NON_MANIFEST`, so we
         # don't check them either.
 
-        not_in_use_datasets = not_in_use(local_datasets.values())
```
wallrj: :question: This one needs a comment and could be bundled into the datastructure-optimizations branch (mentioned above.)

flocker/node/script.py
wallrj: I agree that agent profiling support should be in a separate PR.

flocker/node/script.py
```diff
@@ -294,7 +365,7 @@ class AgentServiceFactory(PClass):
     # This should have an explicit interface:
     # https://clusterhq.atlassian.net/browse/FLOC-1929
     deployer_factory = field(mandatory=True)
-    get_external_ip = field(initial=_get_external_ip, mandatory=True)
```
wallrj: :question: These all belong in #2725 where @moshez switches us to a vendored Pyrsistent, I think. Or alternatively we can create an "upgrade-pyristent" branch and *then* switch to using our vendored repo once we've merged all the optimizations there.

flocker/restapi/_infrastructure.py
```diff
@@ -267,11 +267,9 @@ def loadAndDispatch(self, request, **routeArguments):
 
                 def got_result(result):
                     code = OK
-                    json = result
                     if isinstance(result, EndpointResponse):
                         code = result.code
-                        json = result.result
-                    eliot_action.add_success_fields(code=code, json=json)
```
wallrj: :question: This can be bundled with the other logging reduction changes, but I don't think this one needs a followup. Logging the JSON responses seems excessive and I don't remember it being useful while debugging acceptance tests.

requirements.txt
```diff
@@ -46,7 +46,7 @@ pyasn1-modules==0.0.8
 pycparser==2.14
 pycrypto==2.6.1
 pyOpenSSL==0.15.1
-pyrsistent==0.11.9
```
wallrj: :question: This is already moved to #2725

flocker/node/script.py
moshez: I'm not moving to a vendored pyrsistent, just a forked one.
Vendoring is a longer-term thing.

wallrj:
I made some suggestions about how we should split this into smaller PRs and add some explanatory comments here and there. We can close this once all the individual bits have been merged.

flocker/__init__.py
sarum90: I've had flocker running on 58 nodes over the weekend with the C extension enabled and the cluster is still functioning. Just presenting data.

It definitely gave a significant speedup in some situations, and I actually think that unless the segfault is extremely frequent flocker is designed to be impervious to segfaults (assuming systemd would just respawn the daemon). It's only scary if the segfaults are indicative of worse bugs. IMHO, not as bad as the multithreading issues in the drivers.

sarum90:
Thanks @wallrj for adding the comments. I agree we should do them, and it's nice to have a list of things we can iterate on so we know when we are done.

flocker/node/script.py
```diff
@@ -294,7 +365,7 @@ class AgentServiceFactory(PClass):
...
+    get_external_ip = field(initial=(lambda: _get_external_ip), mandatory=True)
```
wallrj: @moshez: Ah, I was misusing the term "vendored". But the point I was really trying to make is that the fixes here and below are all required for our tests to pass with the latest version (or our fork) of Pyrsistent.

flocker/node/_container.py
wallrj: @itamarst Suggested this in https://clusterhq.atlassian.net/browse/FLOC-3374 along with some logging reduction.

wallrj: I created https://clusterhq.atlassian.net/browse/FLOC-4325 and #2727

flocker/node/_loop.py
wallrj: Created https://clusterhq.atlassian.net/browse/FLOC-4326

flocker/restapi/_infrastructure.py
wallrj: Instead I created https://clusterhq.atlassian.net/browse/FLOC-4327 and #2729

flocker/node/agents/blockdevice.py
wallrj: Created https://clusterhq.atlassian.net/browse/FLOC-4328 and

flocker/__init__.py
wallrj: Created https://clusterhq.atlassian.net/browse/FLOC-4330 and

flocker/control/_model.py
wallrj: Created https://clusterhq.atlassian.net/browse/FLOC-4332 and

wallrj:
We merged all these changes as separate PRs.

PR #2723, opened 2016-03-25
[FLOC-4324] Fast serialization of configuration.

sarum90:
This depends pretty heavily on the other benchmarking speedups that we have already put forward, so I'm opening this against that branch.

Basic idea is to cache the serialization of pyrsistent objects for their entire lifetime. Since pyrsistent objects are usually composed of different pyrsistent object that we have probably serialized, the anticipated benefit is very large.


flocker/control/_persistence.py
```diff
@@ -162,37 +164,106 @@ def upgrade_from_v3(cls, config):
...
-                    "Datetime without a timezone: {}".format(obj))
-            return {_CLASS_MARKER: u"datetime",
-                    "seconds": timegm(obj.utctimetuple())}
-        return JSONEncoder.default(self, obj)
+    # Ensure this is a quick function for basic types:
+    if input_object is None:
+        return None
+    t = type(input_object)
+    is_pyrsistent = False
+    if (
```
moshez: why not:

     if isinstance(input_object, (types.NoneType, str, unicode, int, long, float, bool)):
         return input_object

Instead of lines 230-242

and

    if isinstance(input_object, (dict, list, tuple)):
        obj = input_object

instead of lines 243-246

It seems clearer. If there is a performance benefit in the less clear version, I think we should put a specific comment to that.

flocker/control/_persistence.py
```diff
@@ -202,7 +273,7 @@ def wire_encode(obj):
     :param obj: An object from the configuration model, e.g. ``Deployment``.
     :return bytes: Encoded object.
     """
-    return dumps(obj, cls=_ConfigurationEncoder)
+    return dumps(_cached_dfs_serialize(_cached_dfs_serialize(obj)))
```
moshez: Why the double serialize here?

Also, would prefer (regardless of whether double-serialize is needed or not) to make this into vertical rather than horizontal:

    serialized = _cache_dfs_serialize(obj)
    really_serialized = _cache_dfs_serialize(obj) # needed because ...
    return dumps(really_serialized)

flocker/control/_persistence.py
```diff
@@ -239,8 +310,8 @@ def to_unserialized_json(obj):
     :param obj: An object that can be passed to ``wire_encode``.
     :return: Python object that can be JSON serialized.
     """
-    # Worst implementation everrrr:
-    return loads(wire_encode(obj))
+    # Best implementation everrrr:
+    return _cached_dfs_serialize(obj)
```
moshez: clearly, the best implementation ever is:

    to_serialized_json = _cached_dfs_serialize

:)

flocker/control/_persistence.py
sarum90: 2 reasons:

1) Performance (as you mentioned) - I'll add a comment (https://gist.github.com/sarum90/00d837d89f9a99c5ac7a9e09a713eb1b)

2) Semantics are slightly different:
```
>>> class A(dict):
...   def __init__(self, a):
...     self.bbb = a
... 
>>> 
>>> d = A(23)
>>> d.bbb
23
>>> isinstance(d, dict)
True
>>> type(d)
<class '__main__.A'>
>>>
```
Possibly not worth guarding against the later one, but I did anyway.

flocker/control/_persistence.py
sarum90: Wow yea that's not needed. Something something typo something something.

Thanks for catching that.

flocker/control/_persistence.py
sarum90: :) right?

Leaving as is for the docstring, and cause I'm always unsure how to put docstrings on assignments.

sarum90:
Closing in favor of #2736, which is opened against the right branch, and rebased off of master rather than off of the big franken branch

PR #2725, opened 2016-03-28
Use forked version of Pyrsistent

moshez:
This allows us to use a forked version
of Pyrsistent. Currently, it points
to a CHQ fork that has nothing in it
but a "forking stamp". This should be
updated to a hash that has the speed-ups
we need.

### DIFF
--- a/admin/packaging.py
+++ b/admin/packaging.py
@@ -857,6 +857,10 @@ def omnibus_package_builder(
         steps=(
             InstallVirtualEnv(virtualenv=virtualenv),
             InstallApplication(virtualenv=virtualenv,
+                               package_uri='pip==8.1.1'),
+            InstallApplication(virtualenv=virtualenv,
+                               package_uri='-r/flocker/requirements.txt'),
+            InstallApplication(virtualenv=virtualenv,
                                package_uri=package_uri),
             # get_package_version_step must be run before steps that reference
             # rpm_version
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -173,7 +173,7 @@ class _ConfigurationEncoder(JSONEncoder):
             result[_CLASS_MARKER] = obj.__class__.__name__
             return result
         elif isinstance(obj, PClass):
-            result = obj.evolver().data
+            result = obj._to_dict()
             result[_CLASS_MARKER] = obj.__class__.__name__
             return result
         elif isinstance(obj, PMap):
--- a/admin/test/test_packaging.py
+++ b/admin/test/test_packaging.py
@@ -893,11 +893,15 @@ class OmnibusPackageBuilderTests(TestCase):
 
         package_files = FilePath('/package-files')
 
+        virtualenv = VirtualEnv(root=expected_virtualenv_path)
         expected = BuildSequence(
             steps=(
                 # clusterhq-python-flocker steps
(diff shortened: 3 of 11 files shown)


sarum90:
We have some change detector tests that detected your change!

```
23:36:44 * expected: <InstallApplication(virtualenv=<VirtualEnv(root=FilePath('/opt/flocker'))>, package_uri='https://www.example.com/foo/Bar-1.2.3.whl')> !=
23:36:44   actual:   <InstallApplication(virtualenv=<VirtualEnv(root=FilePath('/opt/flocker'))>, package_uri='pip==8.1.1')>
23:36:44 * expected: <GetPackageVersion(virtualenv=<VirtualEnv(root=FilePath('/opt/flocker'))>, package_name='flocker')> !=
23:36:44   actual:   <InstallApplication(virtualenv=<VirtualEnv(root=FilePath('/opt/flocker'))>, package_uri='-r/flocker/requirements.txt')>
```

flocker/control/_persistence.py
```diff
@@ -173,7 +173,7 @@ def default(self, obj):
             result[_CLASS_MARKER] = obj.__class__.__name__
             return result
         elif isinstance(obj, PClass):
-            result = obj.evolver().data
+            result = obj._to_dict()
```
sarum90: Added.

sarum90:
Just occurred to me, we probably have `pip install .` in our documentation a bunch. http://doc-dev.clusterhq.com/gettinginvolved/infrastructure/packaging.html?highlight=pip for instance.

We should probably add --process-dependency-links to those invocations. in the docs.

admin/packaging.py
```diff
@@ -857,6 +857,10 @@ def omnibus_package_builder(
         steps=(
             InstallVirtualEnv(virtualenv=virtualenv),
             InstallApplication(virtualenv=virtualenv,
+                               package_uri='pip==8.1.1'),
```
sarum90: Sorry side thought here.

Any chance that if we just update pip or virtualenv in the docker image, we wouldn't have to do any of this?

sarum90:
Updated the docs / tox.ini / and some I think unused stuff in build.yaml to have `--process-dependency-links`. So that will re-trigger the tests.

I'm going to build an RPM on this branch using buildbot, then spin up a little cluster, install flocker from the branch, and verify that the forked version of pyrsistent is being used.

sarum90:
Actually... I'm just going to try to catch the acceptance tests while they are running.

admin/packaging.py
moshez: I'd like the logic to be self-contained here.

sarum90:
SSH'd to the acceptance test while it was running.

did:
```
/opt/flocker/bin/python
>>> import pyrsistent
>>> help(pyrsistent)     
```
And the help window indicated that it was the ClusterHQ fork.

That was the main test I wanted to run. Given that passing test and the fact that I find the above code to be maintainable enough, I'm fine with this being merged once the tests have completed, so long as they are green.

sarum90:
I'm going to run the merge-pr script on this so that it will be automatically merged once the tests are green.

sarum90:
This has merge conflicts now... that's probably my fault, sorry.

sarum90:
I'll do the manual merge.

PR #2726, opened 2016-03-28
Acceptance tests no root floc 4316

justnoise:
Well this got out of hand quickly...  There was a smaller PR but our docs are generated from our provisioning code so it's necessary to update those concurrently.  Since this is a large change, I'm very happy to walk through the changes with the reviewer.

This PR exists to support a new tool that allows our users to use our provisioning code to provision a cluster easily.  To ensure a user doesn't need to allow root to ssh into their machine, the provisioning code has been changed to run everything through sudo.  (yes, this assumes that the user of the provisioning tool will have passwordless sudo enabled).

A few new Effect classes have been added for the sudo commands, the nastiest of which is sudo_script.  This packages multiple bash commands into a shell quoted string and runs them using `bash -c 'commands'`.


wallrj:
@justnoise there are some conflicts here so I'll wait for you to merge the branch forward.

admin/client.py
```diff
@@ -279,6 +296,8 @@ def get_steps_pkg(distribution, package_source=PackageSource()):
     package_manager = DOCKER_IMAGES[distribution].package_manager
     steps = [
         ensure_minimal_setup(package_manager),
+        allow_notty_sudo(distribution),
```
justnoise: Sadly we still need a tty for running a script on docker (but not for ssh)

admin/client.py
```diff
@@ -279,6 +296,8 @@ def get_steps_pkg(distribution, package_source=PackageSource()):
...
+        install_distro_package("curl", distribution),
```
justnoise: It's possible this should go in the ensure_minimal_setup but I'm unsure if we should pollute that function with installing curl.

flocker/provision/_install.py
```diff
@@ -743,7 +754,12 @@ def _remove_private_key(content):
     if trim_end <= trim_start:
         # strangely short key, keep all content
         return content
-    return content[:trim_start] + '...REMOVED...' + content[trim_end:]
+    end_of_key = end+len(suffix)
+    trimmed_key = (content[:trim_start] +
+                   '...REMOVED...' +
+                   content[trim_end:end_of_key])
+    # keep looking for more keys
+    return trimmed_key + _remove_private_keys(content[end_of_key:])
```
justnoise: Probably uncecessary but updated this to remove multiple keys from the output (used to stop after a single key)

justnoise:
OK this is ready for review... hopefully for good this time.

flocker/provision/_install.py
```diff
@@ -140,6 +140,15 @@ def apt_get_install(args, sudo=False):
     )
 
 
+def install_distro_package(package_name, distribution, sudo=False):
+    if type(package_name) != list:
+        package_name = [package_name]
+    if is_centos(distribution):
```
myechuri: Address RHEL.

flocker/provision/_install.py
```diff
@@ -915,21 +940,27 @@ def task_enable_flocker_control(distribution, action="start"):
...
                 content=dedent('''\
                     start on runlevel [2345]
                     stop on runlevel [016]
                     '''),
             ),
-            run("echo 'flocker-control-api\t4523/tcp\t\t\t# Flocker Control API port' >> /etc/services"),  # noqa
-            run("echo 'flocker-control-agent\t4524/tcp\t\t\t# Flocker Control Agent port' >> /etc/services"),  # noqa
-            run_from_args(['service', 'flocker-control', action.lower()]),
+            sudo_script(dedent("""\
+            if grep -q flocker-control-api /etc/services; then
```
myechuri: Bug: need to make sure it's NOT in /etc/services

flocker/provision/_install.py
```diff
@@ -915,21 +940,27 @@ def task_enable_flocker_control(distribution, action="start"):
...
+                echo 'flocker-control-api\t4523/tcp\t\t\t# Flocker Control API port' >> /etc/services
+            fi""")),  # noqa
+            sudo_script(dedent("""\
+            if grep -q flocker-control-agent /etc/services; then
```
myechuri: Same Bug.

wallrj:
Branch has merge conflicts and the certificate distribution problem is going to go away in the next version of Flocker.

PR #2728, opened 2016-03-29
[FLOC-4326] Do not log configuration and deployment state during CONVERGE

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4326

Copied from #2720 with some additional test fixes.

### DIFF
--- a/flocker/node/_loop.py
+++ b/flocker/node/_loop.py
@@ -379,21 +379,13 @@ LOG_SEND_TO_CONTROL_SERVICE = ActionType(
     [_FIELD_CONNECTION, _FIELD_LOCAL_CHANGES], [],
     u"Send the local state to the control service.")
 
-_FIELD_CLUSTERSTATE = Field(
-    u"cluster_state", to_unserialized_json,
-    u"The state of the cluster, according to control service.")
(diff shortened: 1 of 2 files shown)


flocker/node/_loop.py
```diff
@@ -524,8 +516,7 @@ def _maybe_send_state_to_control_service(self, state_changes):
             return succeed(None)
 
     def output_CONVERGE(self, context):
-        with LOG_CONVERGE(self.fsm.logger, cluster_state=self.cluster_state,
-                          desired_configuration=self.configuration).context():
+        with LOG_CONVERGE(self.fsm.logger).context():
```
sarum90: Add a comment with a follow-up issue to re-enable some subset of the logging.

sarum90:
Thanks for the test fixes!

I think we should probably add a comment with a FLOC-XXXX issue opened to follow this up. I see this as a fairly significant tech debt / risk as it will make acceptance tests much harder to debug.

flocker/node/_loop.py
wallrj: Done.

PR #2729, opened 2016-03-29
[FLOC-4327] Stop logging the response body of REST API requests

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4327

Copied from #2720 with some test alterations.

### DIFF
--- a/flocker/restapi/_logging.py
+++ b/flocker/restapi/_logging.py
@@ -37,4 +37,4 @@ JSON_REQUEST = ActionType(
     LOG_SYSTEM + u":json_request",
     [JSON],
     [RESPONSE_CODE],
-    u"A request containing JSON request and response bodies.")
+    u"A request containing JSON request body and HTTP response code.")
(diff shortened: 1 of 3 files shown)


flocker/restapi/_logging.py
```diff
@@ -36,5 +36,5 @@
 JSON_REQUEST = ActionType(
     LOG_SYSTEM + u":json_request",
     [JSON],
-    [RESPONSE_CODE, JSON],
+    [RESPONSE_CODE],
     u"A request containing JSON request and response bodies.")
```
zubron: ❓ Is this description still correct, should we be referring to the response body here?

zubron:
Looks good, thanks @wallrj! Just one point about maybe updating an eliot action description. Please address if you think you need to and then merge.

wallrj:
Thanks @zubron I fixed the description and added an explanatory comment. I'll merge when the builds pass.

wallrj:
I'm ignoring the two cinder failures above. We're seeing them on master and in other branches. Probably due to Rackspace problems.

PR #2731, opened 2016-03-29
[FLOC-4328] Don't wake up for datasets that have been deleted on other nodes

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4328

Copied from #2720 with a test.

It took me ages to figure out that the problem was only occurring when the deleted dataset had been attached to some other volume.
For now I've modified one of the existing tests. Perhaps this case deserves a separate test though.

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -1864,7 +1864,7 @@ class CalculateDesiredStateTests(TestCase):
 def assert_calculated_changes(
         case, node_state, node_config, nonmanifest_datasets, expected_changes,
         additional_node_states=frozenset(), leases=Leases(),
-        discovered_datasets=None, persistent_state=None,
+        discovered_datasets=None
 ):
     """
     Assert that ``BlockDeviceDeployer`` calculates certain changes in a certain
@@ -1907,7 +1907,6 @@ def assert_calculated_changes(
         case, deployer, node_state, node_config,
         nonmanifest_datasets, additional_node_states, set(),
         expected_changes, local_state, leases=leases,
-        persistent_state=persistent_state,
     )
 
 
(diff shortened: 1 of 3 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2143,11 +2144,29 @@ def test_deleted_ignored(self):
...
+        )
 
         assert_calculated_changes(
             self, local_state, local_config,
             nonmanifest_datasets={},
             expected_changes=in_parallel(changes=[]),
+            discovered_datasets=[
+                foreign_registered_dataset
+            ],
+            persistent_state=PersistentState().transform(
```
sarum90: :question: +1 on getting this in a unit test. This file is full with tests that are sort of... tricky to maintain. There is a very big overhead to re-architecting this part of flocker. I don't think this makes the problem any worse, but it's something I'm always on the lookout for.

sarum90:
@wallrj Care to review again?

I changed the tests a little bit after our chat in the office.

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2144,6 +2143,18 @@ def test_deleted_ignored(self):
                 metadata={u"foo": u"bar"},
             )
         )
+
+        # Upon deletion, datasets to not get unregistered, so they will still
```
wallrj: typo "to not" > "do not"

flocker/node/agents/test/test_blockdevice.py
```diff
@@ -2144,10 +2144,34 @@ def test_deleted_ignored(self):
             )
         )
 
+        # Upon deletion, datasets to not get unregistered, so they will still
```
wallrj: typo "to not"

wallrj:
Thanks @sarum90 

Your changes look good to me. But I don't see anything here that calls `discover_state`.  but you've pointed out that there is already a test for the case where it returns DiscoveredDataset.REGISTERED.

Let's merge this.

PR #2736, opened 2016-03-30
[FLOC-4324] Caching json serialization

sarum90:
This modifies our JSON serialization to not use the JSONEncoder scheme, and instead turns the entire Pyrsistent object into something that is serializable outside of the realm of the JSONEncoder.

This enabled us to be able to add a cache that stores the values of the encoding at the subtree level, which should lead to great speedups when serializing the configuration or the state inside flocker.

### DIFF
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -207,7 +207,7 @@ def _to_serializables(obj):
 
 def _is_pyrsistent(obj):
     """
-    Boolean check if an object is a pyrsistent object.
+    Boolean check if an object is an instance of a pyrsistent object.
     """
     return isinstance(obj, (PRecord, PClass, PMap, PSet, PVector))
 
@@ -229,8 +229,10 @@ def _cached_dfs_serialize(input_object):
     # Ensure this is a quick function for basic types:
     if input_object is None:
         return None
+
+    # Note that ``type(x) == str or type(x) == int`` is faster than
+    # ``isinstance(x, (str, int))``.
     t = type(input_object)
-    is_pyrsistent = False
     if (
             t is str or
             t is unicode or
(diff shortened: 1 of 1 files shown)


flocker/control/_persistence.py
```diff
@@ -162,37 +164,110 @@ def upgrade_from_v3(cls, config):
...
-            return {_CLASS_MARKER: u"datetime",
-                    "seconds": timegm(obj.utctimetuple())}
-        return JSONEncoder.default(self, obj)
+    # Ensure this is a quick function for basic types:
+    if input_object is None:
+        return None
+
+    # Note that ``type(x) == str or type(x) == int`` is faster than
+    # ``isinstance(x, (str, int))``.
+    t = type(input_object)
```
moshez: Nitpick: not a huge fan of one-letter variables, how about input_type?

flocker/control/_persistence.py
```diff
@@ -162,37 +164,110 @@ def upgrade_from_v3(cls, config):
...
+    if (
```
moshez: Nitpick:

What about

    ## Somewhere outside this function
    _ATOMIC_TYPES = set([str, unicode, int, long, float, bool])

    ## Here
    if t in _ATOMIC_TYPES:
        return input_object

sarum90: Oh cool, sets are optimized for small numbers of objects: https://gist.github.com/sarum90/00d837d89f9a99c5ac7a9e09a713eb1b

Just wanted to double check -- I'll definitely switch to that.

sarum90:
I'm going to wait for another full run of the tests (This code is pretty essential to the way flocker works).

Thanks for the reviews! Much nicer on the eyes and actually faster now :)

PR #2738, opened 2016-03-30
[FLOC-4334] Cache flocker-control responses.

sarum90:
In order to speed up response times from the control agent, this PR adds a simple cache to be used in the case where the deployment state or configuration has not changed since the last time it was requested over the API.

Note: This assumes that the callers of the caching functions are well-behaved and don't mutate the values they receive. If they mutate the values they receive they will be mutating the value in the cache. I believe this is true, and that it is fairly easy to reason about it being true in the future.

### DIFF
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -143,7 +143,7 @@ def _extract_containers_state(deployment_state):
     container in the deployment state.
 
     N.B.: Callers of this function should not mutate the returned value as this
-    function uses an lru_cache.
+    function uses an lru_cache (FLOC-4345 to clean up).
 
     :param DeploymentState deployment_state: The deployment state to extract
         the containers from.
@@ -152,9 +152,7 @@ def _extract_containers_state(deployment_state):
     """
     result = []
     for node in deployment_state.nodes:
-        if node.applications is None:
-            continue
-        for application in node.applications:
+        for application in (node.applications or ()):
             container = container_configuration_response(
                 application, node.uuid)
             container[u"running"] = application.running
@@ -1257,7 +1255,7 @@ def _containers_from_deployment(deployment):
     Extract the containers from the supplied deployment instance.
 
     N.B.: Callers of this function should not mutate the returned value as this
-    function uses an lru_cache.
+    function uses an lru_cache (FLOC-4345 to clean up).
 
     :param Deployment deployment: A ``Deployment`` describing the state
         of the cluster.
--- a/flocker/control/httpapi.py
+++ b/flocker/control/httpapi.py
@@ -142,6 +142,9 @@ def _extract_containers_state(deployment_state):
     Turn the deployment state into serializable simple types for every
     container in the deployment state.
 
+    N.B.: Callers of this function should not mutate the returned value as this
+    function uses an lru_cache.
+
     :param DeploymentState deployment_state: The deployment state to extract
         the containers from.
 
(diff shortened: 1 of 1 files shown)


flocker/control/httpapi.py
```diff
@@ -134,6 +136,32 @@ def render_if_matches(self, request, **route_arguments):
     return render_if_matches
 
 
+@lru_cache(1)
+def _extract_containers_state(deployment_state):
+    """
+    Turn the deployment state into serializable simple types for every
+    container in the deployment state.
+
+    N.B.: Callers of this function should not mutate the returned value as this
```
moshez: Can we return a pvector instead of Asking Nicely(TM)? This scares me. A lot.

flocker/control/httpapi.py
```diff
@@ -134,6 +136,32 @@ def render_if_matches(self, request, **route_arguments):
...
+    :param DeploymentState deployment_state: The deployment state to extract
+        the containers from.
+
+    :returns: The states of the containers in simple types.
+    """
+    result = []
+    for node in deployment_state.nodes:
+        if node.applications is None:
+            continue
+        for application in node.applications:
```
moshez: You can harmonize with

    for application in (node.applications or ()):

flocker/control/httpapi.py
```diff
@@ -523,7 +551,8 @@ def get_containers_configuration(self):
         :return: A ``list`` of ``dict`` representing each of the containers
             that are configured to exist anywhere on the cluster.
         """
-        return list(containers_from_deployment(self.persistence_service.get()))
+        return list(
+            _containers_from_deployment(self.persistence_service.get()))
```
moshez: Did you just change the formatting here?

flocker/control/httpapi.py
```diff
@@ -1231,18 +1251,26 @@ def datasets_from_deployment(deployment):
...
 
 
-def containers_from_deployment(deployment):
+@lru_cache(1)
+def _containers_from_deployment(deployment):
     """
     Extract the containers from the supplied deployment instance.
 
+    N.B.: Callers of this function should not mutate the returned value as this
+    function uses an lru_cache.
```
moshez: See above

flocker/control/httpapi.py
sarum90: Yea +1000... let me do some profiling.

The thing is... we need to return something _deeply_ pyrsistent, and then call serialize() on it externally.

This sounds expensive in the general case.

flocker/control/httpapi.py
sarum90: I changed `containers_from_deployment` to be a private function to this file, made the width of this line 80 characters :(.

Made private since it's API is so terrible. (can't mutate result)

flocker/control/httpapi.py
sarum90: Opened up FLOC-4345 to make this better, preliminary attempts to make the issue go away proved to be not successful.

flocker/control/httpapi.py
sarum90: Some people don't like that formation. I happen to like it though.

Done.

flocker/control/httpapi.py
sarum90: Added pointer to the FLOC-4345 issue here as well.

sarum90:
Thanks for the review!

I was hopeful to be able to resolve the nastiness immediately, sadly it doesn't seem like that was possible.

That said this should be pretty close to ready to merge, will merge once tests are green.

sarum90:
I'm going to merge this even though the acceptance tests haven't passed on this exact hash.

They passed on the previous commit, so I'm not too worried.

PR #2739, opened 2016-03-30
[FLOC-4336] Update pyrsistent to a version with __eq__ and __hash__ optimizations

sarum90:
Now that we have a mechanism to use our own fork of pyrsistent, this is the commit to use the latest hash of master.

By policy this hash matches exactly a valid hash from ClusterHQ/pyrsistent/master, specifically:

https://github.com/ClusterHQ/pyrsistent/commit/5b722c3ad75ed2da414a40e6d27af373a8610018


requirements.txt
```diff
@@ -1,3 +1,4 @@
+git+https://github.com/ClusterHQ/pyrsistent@5b722c3ad75ed2da414a40e6d27af373a8610018#egg=pyrsistent-0.11.13chq
```
sarum90: This PR is built on top of https://github.com/ClusterHQ/flocker/pull/2725

Once that is merged the only change in this PR should be the hash in this line.

PR #2742, opened 2016-04-04
[FLOC-4357] Removed one of the points of logging

sarum90:
Flocker had a lot of nested logging on API requests. This drives CPU usage higher than it should be for high numbers of requests per second.

Removing one of these layers in an attempt to reduce CPU usage.

After spinning up 2 15 node clusters, one with this change and one without this change, I observed the following when I ran the 10 QPS read benchmark:

Without this PR:

flocker-control CPU: 25%

With this PR:

flocker-control CPU: 23%

This was on ubuntu rather than CentOS, but now that we neutered the JSON logs, it seems to me that these are adding no value, and getting the extra 2 percent back is sort of a big deal.

### DIFF
--- a/flocker/apiclient/test/test_client.py
+++ b/flocker/apiclient/test/test_client.py
@@ -47,7 +47,7 @@ from ...control import (
     NodeState, NonManifestDatasets, Dataset as ModelDataset, ChangeSource,
     DockerImage, UpdateNodeStateEra,
 )
-from ...restapi._logging import JSON_REQUEST
+from ...restapi._logging import REQUEST
 from ...restapi import _infrastructure as rest_api
 from ... import __version__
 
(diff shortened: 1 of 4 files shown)


wallrj:
I looked back to see why we added the logging in the first place and found #1225 

We *reluctantly" removed logging of the JSON response body, for performance reasons, but at the time we talked about restoring that logging at some point when we'd figured out how to do it efficiently.

This change takes us further away from that goal.

If I remember correctly, the expensive part in these logs, is looking up the parent task from the stack.

If Eliot could delay that operation until the observer consumes (serializes) the messages, perhaps we could write a log observer with a ring buffer and which only logs to Journalctl when an error occurs.

This seems like a great speed up for the time being though.

Please merge.

sarum90:
Merging as requested. Acknowledged about logging.

I think we might need to do something smarter in general than logging on every request though. We'll get it right somehow.

PR #2743, opened 2016-04-04
[FLOC-4355] Install Pyrsistent from a tag

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4355

This works with https://github.com/ClusterHQ/pyrsistent/pull/7

Allows us to install Flocker as follows:

```
pip install --process-dependency-links --editable .[dev]
...

$ pip freeze | grep pyrsistent
pyrsistent==0.11.12+chq1
...

$ flocker --version
1.10.3.dev2+374.g36a6122

```

It's no longer necessary to do a separate ``pip install -r requirements.txt` first, but I haven't modified the docs yet. I can if we agree this is a better / more reliable way to install our fork of pyrsistent.

This is how we used to install our fork of testtools:
 * https://github.com/ClusterHQ/flocker/pull/2324


moshez:
How did you test that this installs the right Pyrsistent?

[The difference is that because eliot depend on pyrsistent, sometimes pip gets confused and installs the wrong one. Note that testing is easy because I put a docstring in our fork's pyrsistent.]

moshez:
OK tested it myself.

    [root@cb735b8f51f5 flocker.local]# python -c 'import pyrsistent;help(pyrsistent)' | grep flocker
        distributions of flocker, in case there are

wallrj:
I made this change as part of https://github.com/ClusterHQ/flocker/pull/2792 and https://github.com/ClusterHQ/pyrsistent/pull/11

PR #2746, opened 2016-04-05
[FLOC 4361] Update the label on the Supported Platforms/RHEL box 

davidoliverSP2:
Fixes 4361

This moves RHEL before the experimental CentOS box, and removes the coming soon label.

It was noted during this work that the CTA on the supported platform links (to Learn More) directs the user to a rather unhelpful homepage destination. This has been raised as a bug (FLOC 4362)

### DIFF
--- a/docs/supported/index.rst
+++ b/docs/supported/index.rst
@@ -188,16 +188,15 @@ Supported Operating Systems
 	        <a href="../index.html" class="button">Learn more</a>
 	    </div>
 	    <div class="pod-boxout pod-boxout--4up pod-boxout--orchestration">
+			<img src="../_static/images/rhel72x.png" alt="RHEL 7 logo"/>
+			<span>RHEL 7</span>
+	        <a href="../index.html" class="button">Learn more</a>
+	    </div>
+	    <div class="pod-boxout pod-boxout--4up pod-boxout--orchestration">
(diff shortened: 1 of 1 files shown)


docs/supported/index.rst
```diff
@@ -188,17 +188,16 @@ Supported Operating Systems
 	        <a href="../index.html" class="button">Learn more</a>
 	    </div>
 	    <div class="pod-boxout pod-boxout--4up pod-boxout--orchestration">
+			<img src="../_static/images/rhel72x.png" alt="RHEL 7 logo"/>
+			<span>RHEL 7</span>
+	        <a href="../index.html" class="button">Learn more</a>
```
zubron: ❓ Not specific to this change, but all the supported configurations (except CoreOS) link to back to the beginning of the docs - is this intended?

zubron:
Thanks for another quick change @davidoliverSP2! Just one question on where the buttons link to, but since that hasn't changed with this branch, I'm happy for this to be merged.

docs/supported/index.rst
wallrj: I think it's because the CoreOS installation instructions are off on our blog somewhere while the other "traditional" OS installations all require the user to choose their orchestration framework first.

wallrj:
I'll merge this.

PR #2749, opened 2016-04-05
[FLOC-4363] Increase other acceptance test timeouts

sarum90:
Now that we restart the nodes every time we toggle running the container agent or not, our tests take a lot longer to run.

Thus, we should increase all acceptance test timeouts rather than just the ones that GCE was slow on.


wallrj:
LGTM. Please merge when the acceptance tests pass.

sarum90:
All tests but the rackspace ones have passed. I'd rather get this in and re-run the tests on master since master has the other fix as well.

PR #2752, opened 2016-04-07
[FLOC-4369] Update all the openstack dependencies and their transisent dependencies

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4369

PR #2753, opened 2016-04-07
[FLOC-4368] Simpler Cinder HTTPs tests

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4368

Stacked on #2752

These tests were using region dependent, but needn't be.
Instead they now only create a keystone session, which is not tied to a region.

The tests were also broken, in so far as the "rackspace" keystone auth driver doesn't support the peer_verify or peer_ca_path options. 

I now skip the tests unless the driver supports those options.

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -197,28 +197,58 @@ class CinderHttpsTests(TestCase):
     """
     Test connections to HTTPS-enabled OpenStack.
     """
-    def setUp(self):
-        super(CinderHttpsTests, self).setUp()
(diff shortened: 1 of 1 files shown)


sarum90:
Looks like good test cleanup.

A little sad to have tests that we don't have any location we can run though :/ Maybe add a comment that you basically can't run these unless you find a way to run them on swisscom or hopefully in our home grown openstack?

Definitely an improvement either way.

wallrj:
The `jenkins-run_client_installation_on_Ubuntu_Wily` failure is unrelated. Merging.

PR #2755, opened 2016-04-07
[FLOC-4370] Removed most references to homebrew throughout code and docs.

sarum90:
As far as I can tell, on a clean installation of El Capitan, system python /
sytem OpenSSL is capable of installing the flocker command line utilities from
pip, and they are functional. Similarly most other maintenance tasks that used
to require specifically setting flags to use different versions of libSSL / or
specifically installing packages using brew before pip installing flocker are
no longer needed.

Combined with a desire to reduce our release/maintenance burden of creating a
new brew recipe, I've just removed most references to brew/homebrew throughout
our codebase. The one exception is the release process, which still uses
Homebrew to install utilities to create packages.


docs/gettinginvolved/infrastructure/release-process.rst
```diff
@@ -47,7 +46,7 @@ Software
 .. prompt:: bash $
 
    brew tap stepanstipl/noop
-   brew install createrepo dpkg libffi openssl
+   brew install createrepo dpkg
```
wallrj: :question: I'm inclined to remove these instructions too and say that the release process must be performed from a standard Centos7 machine....and I think we'd both like to see the Vagrant dev environment re-instated. I'll file a followup issue if you agree.

docs/installation/install-client.rst
```diff
@@ -109,7 +109,7 @@ Whenever you need to run Flocker CLI commands, ensure you are in the virtualenv:
 .. version-prompt:: bash alice@mercury:~$ auto
 
    alice@mercury:~$ source flocker-client/bin/activate
-   alice@mercury:~$ flocker-deploy --version
```
wallrj: :question: That's been annoying me too. We should remove ``flocker-deploy`` to avoid any confusion. https://clusterhq.atlassian.net/browse/FLOC-4371

docs/installation/install-client.rst
```diff
@@ -119,38 +119,29 @@ Whenever you need to run Flocker CLI commands, ensure you are in the virtualenv:
...
+Install ``pip`` and ``virtualenv`` on your machine at the system level:
 
 .. prompt:: bash alice@mercury:~$
 
-   brew uninstall flocker-<old version>
+   sudo python -m ensurepip
+   sudo pip install virtualenv
 
-Add the ``ClusterHQ/tap`` tap to Homebrew and install ``flocker``:
+To install ``flocker-cli`` in a Python virtualenv, run:
```
wallrj: :question: I think it's confusing to highlight ``flocker-cli`` this way, since there's no such tool or package, but I see that's how it's formatted elsewhere....so fine. Might be better if we talked about installing ``flocker-ca`` specifically since that's the only client command of interest these days....until we have a proper ``flocker-volumes`` tool....and even then perhaps we should make it a ``flocker volumes`` and a ``flocker ca`` style subcommand.

wallrj:
Thanks @sarum90 

Looks good. I haven't tested the new client installation instructions, but I trust you have. A few points below.

* Some lint errors
  http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/remove-homebrew-recipe-FLOC-4370/job/run_lint/1/

* build.yml has an ``is_osx`` function which no longer seems to be used.

* Some remaining Mac references

```
$ git grep -i 'mac\W'
.gitignore:# Mac stuff
docs/gettinginvolved/client-testing.rst:ClusterHQ has a Mac with the ability to start an OS X Virtual Machine.
...
```

http://clusterhq-staging-docs.s3-website-us-east-1.amazonaws.com/remove-homebrew-recipe-FLOC-4370/gettinginvolved/client-testing.html#os-x


Please merge when you're happy with things.

docs/installation/install-client.rst
sarum90: Yea, just following existing practice for now. Acknowledged about maybe we should change this... although flocker-cli is future resistant :), in that it will always be accurate even if we change what is in that package. Mostly just not changing because not changing is easier than changing.

docs/gettinginvolved/infrastructure/release-process.rst
sarum90: Yea 👍 no need to like "officially support" it. But it is useful for what it is.

docs/installation/install-client.rst
sarum90: I'll leave that for addressing in the follow up.

sarum90:
Fixed lint, removed a few more references. Kicking tests and starting the merge-pr script momentarily.

sarum90:
The remaining failing test is actually failing on master as well (http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/view/master/job/master/job/run_client_installation_on_Ubuntu_Wily/425/console)

So I'm going to merge this as is.

PR #2760, opened 2016-04-14
[FLOC-2090] Use the AgentService to load functional testing configuration

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2090


build.yaml
```diff
@@ -327,26 +327,37 @@ common_cli:
...
+    export FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_SECTION=storage-drivers.aws
+    AWS_ZONE=$(\
+        wget -q -O - \
+        http://169.254.169.254/latest/meta-data/placement/availability-zone\
+    )
+    echo "AWS_ZONE=${AWS_ZONE}"
+    export FLOCKER_FUNCTIONAL_TEST_CONFIG_OVERRIDE_zone="${AWS_ZONE}"
+    # There isn't a separate meta-data endpoint to get the instance region.
+    # So assume that region is always ZONE without the trailing character.
+    export FLOCKER_FUNCTIONAL_TEST_CONFIG_OVERRIDE_region="${AWS_ZONE:0:-1}"
```
sarum90: Nice! yea, this is how we should always be specifying this.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -869,3 +881,237 @@ def setUp(self):
...
+    backend, api_args = backend_and_api_args_from_configuration(config)
+    api = get_api(
+        backend=backend,
+        api_args=api_args,
+        reactor=reactor,
+        cluster_id=make_cluster_id(
+            # XXX I assume that this function will only be used from functional
+            # tests.  Acceptance tests will load the API via the AgentService.
+            test_type=TestTypes.FUNCTIONAL,
+            # XXX This should match the backend, but Providers are
```
sarum90: Any chance we could just remove this parameter then?

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -869,3 +881,237 @@ def setUp(self):
...
+            "Section: %s, "
+            "Configuration File: %s" % (config_section, config_file_path)
+        )
+
+    # XXX A hack to work around the fact that the sub-sections of
+    # storage-drivers in acceptance.yml do not all have a ``backend`` key.
+    if "backend" not in config:
+        config["backend"] = section
+
+    # XXX Allow certain configuration to be overridden from environment
```
sarum90: Why the XXX?

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -869,3 +881,237 @@ def setUp(self):
...
+    # variables.
+    # This allows us to dynamically use the zone where the functional test
```
sarum90: ❓ Cool mechanism. I'm somewhat concerned about the discoverability though. Not so concerned that you should necessarily change it, but I frequently grep for environment variables when I want to figure out what they do in code, and I would have to figure out that I should have only grepped for the prefix.

flocker/node/backends.py
```diff
@@ -159,3 +159,13 @@ class BackendDescription(PClass):
 LOOPBACK = backend_loader.get('loopback')
 ZFS = backend_loader.get('zfs')
 GCE = backend_loader.get('gce')
+
+
+def backend_and_api_args_from_configuration(configuration):
+    """
+    """
```
sarum90: Docstring requested.

flocker/node/script.py
```diff
@@ -428,16 +431,14 @@ class AgentService(PClass):
     :ivar node_credential: Credentials with which to configure this agent.
     :ivar ca_certificate: The root certificate to use to validate the control
         service certificate.
-    :ivar backend_name: The name of the storage driver to instantiate.  This
-        must name one of the items in ``backends``.
+    :ivar backend_description: XXX
```
sarum90: request dockstring

sarum90:
This is much easier to follow (from what I can tell). Couple of documentation fix-ups requested, and a few questions. In general though this looks great. Merge after you revise.

flocker/node/agents/testtools/_blockdevice.py
wallrj: I removed it and  completely removed the Providers constants and some unused helper functions. I *think* that's the right thing to do. At least I couldn't think of a good reason to encode "OPENSTACK", "GCE" etc into the cluster_id....that should be obvious from the cloud where we find the lingering cluster, right? (7d2ea42efa9942378f340ce76c7f7b3ff8de85b6)

flocker/node/agents/testtools/_blockdevice.py
wallrj: Removed.

flocker/node/agents/testtools/_blockdevice.py
wallrj: Yeah, that's a good point. Hopefully the fact that the last part is lower case will make that somewhat obvious. eg `FLOCKER_FUNCTIONAL_TEST_CONFIG_OVERRIDE_region` but perhaps I should make the value a JSON blob...

wallrj: I tried  parsing the environment variable value as JSON or YAML but it seems ....error prone. Especially when I'm trying to insert multiple values and values expanded from bash variables.
I'll leave the current mechanism for now (undocumented) and hopefully we can remove it as soon as we update all the drivers to auto-detect their zone.

flocker/node/backends.py
wallrj: I hate this function, but I'll leave it and documented it for now. 
I think we can probably ditch it if / when we make our default backends into plugins. eg
``backend: flocker.node.agents.cinder``

flocker/node/script.py
wallrj: done.

PR #2758, opened 2016-04-13
[FLOC-4374] Allow get_blockdeviceapi_with_cleanup to be used for 3rd party backend drivers

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4374

PR #2762, opened 2016-04-14
Review Branch (DO NOT MERGE)

sarum90:
This reviews https://github.com/ClusterHQ/flocker/pull/2758  (against it's logical head)


build.yaml
```diff
@@ -347,6 +349,7 @@ common_cli:
     # bootstrapping. These are copied from the Jenkins Master /etc/slave_config
     # directory.
     export FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE=/tmp/acceptance.yaml
+    export FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_SECTION=storage-drivers.gce
     export FLOCKER_FUNCTIONAL_TEST_CLOUD_PROVIDER=gce
```
sarum90: Is this necessary still? Prefer duplicate data over the wrong abstraction, but this looks like it might actually just be the wrong abstraction.

If it's not easy to remove for other structural reasons, feel free to ignore.

flocker/node/agents/testtools/_cinder.py
```diff
@@ -3,14 +3,19 @@
...
 from ..cinder import (
     ICinderVolumeManager,
     INovaVolumeManager,
 )
+from ._blockdevice import require_backend
+
+
+# Raise SkipTest unless the functional test configuration has an openstack
+# backend.
+require_backend_openstack = require_backend('openstack')
```
sarum90: Any particular reason to make this different than the other require backends? (i.e. just use require_backend('openstack') everywhere?)

Not worth changing / both are obviously equally expressive. Just curious if there is an underlying reason.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -1085,33 +1053,22 @@ def _gce(cluster_id, config):
...
+    up at the end of the current test run.
 
     :param TestCase test_case: The running test.
-    :param provider: A provider type the ``IBlockDeviceAPI`` is to be
-        compatible with.  A value from ``ProviderType``.
-
     :raises: ``SkipTest`` if either:
         1) A ``FLOCKER_FUNCTIONAL_TEST_CLOUD_CONFIG_FILE``
         was not set and the default config file could not be read, or,
         2) ``FLOCKER_FUNCTIONAL_TEST`` environment variable was unset.
```
sarum90: I think this is no longer true.

build.yaml
sarum90: Never-mind, I see this is used to get the minimum blockdevice size. I suppose we can also use this to create a thing that spins up a node and runs the tests at some point in the future.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -891,117 +891,81 @@ class InvalidConfig(Exception):
...
-class ProviderType(Names):
-    """
-    Kinds of compute/storage cloud providers for which this module is able to
-    build ``IBlockDeviceAPI`` providers.
-    """
-    openstack = NamedConstant()
-    aws = NamedConstant()
-    gce = NamedConstant()
-    rackspace = NamedConstant()
+# XXX Remember to comment on or close FLOC-2584.
```
sarum90: Remember to comment on or close FLOC-2584.

Honestly... I'd just close it. This is the sort of issue I don't fully understand why we create. It's not like there's ever a time where we just choose to take a random issue off the backlog, and what needs to be done to clean it up is pretty obvious to someone looking through the code. I'm not sure that the issue provides enough relevant content above what the code was to outweigh the fact that it could become stale while the code changes.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -891,117 +891,81 @@ class InvalidConfig(Exception):
...
     Validate and load cloud provider's yml config file.
     Default to ``~/acceptance.yml`` in the current user home directory, since
     that's where buildbot puts its acceptance test credentials file.
     """
-    config = get_blockdevice_config(provider_type)
-    provider = _provider_for_provider_type(provider_type)
+    config = get_blockdevice_config()
+    backend_name = config.pop('backend')
+    provider = Providers.lookupByName(backend_name.upper())
     factory = _BLOCKDEVICE_TYPES[provider]
```
sarum90: This still won't work for external drivers, correct?

Just a passing observation, obviously not worse than what was here.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -891,117 +891,81 @@ class InvalidConfig(Exception):
...
+            "Configuration File: %s" % (config_section, config_file_path)
         )
 
-    # XXX - If provider type is Rackspace, we add the auth_plugin config
-    # here.  The CI system should be configured to provide this
-    # parameter as part of the config, to match the documented way of
-    # configuring Rackspace.
-    if provider_environment == ProviderType.rackspace:
-        section['auth_plugin'] = 'rackspace'
+    # XXX A hack to work around the fact that the sub-sections of
```
sarum90: Hrm... This hack... half of me wants you to remove it, and instead re-deploy Jenkins... but that might be busy work. Let's chat about that.

flocker/node/agents/testtools/_blockdevice.py
```diff
@@ -891,117 +891,81 @@ class InvalidConfig(Exception):
...
-            "Unsupported provider. "
-            "Supplied provider: %s, "
-            "Available providers: %s" % (
-                provider_name,
-                ', '.join(p.name for p in ProviderType.iterconstants())
-            )
-        )
+    section = None
+    for section in config_section.split('.'):
+        config = config.get(section)
```
sarum90: nit, perhaps use `(config or {}).get(section)` or otherwise guard against missing sections before the last one.

specifically the following case:

```
config_section = 'a.b.c.d'
config = {a: {b: {}}}
```
will error out with "NoneType has no attribute get" when section='d' and config equals `None` from the previous attempt at `getting` `c` from `b`.

build.yaml
wallrj: Yeah, I was tempted to get rid of it and instead introduce  `FUNCTIONAL_TEST_MINIMUM_ALLOCATABLE_SIZE` and `FUNCTIONAL_TEST_ALLOCATION_UNIT` which again, would help 3rd party driver developers to use our factory functions.

flocker/node/agents/testtools/_cinder.py
wallrj: I did this to avoid writing `require_backend('openstack')` in lots of places across two test modules...but on reflection it's probably not worth the indirection. I'll remove it.

flocker/node/agents/testtools/_blockdevice.py
wallrj: Hopefully the next branch will address this.

flocker/node/agents/testtools/_blockdevice.py
wallrj: @sarum90 I'll leave it in this branch and we can discuss removing it as soon as I land the segredos fix.

flocker/node/agents/testtools/_blockdevice.py
wallrj: Done.

wallrj:
Thanks @sarum90 I addressed all your comments and merged the original branch.

PR #2765, opened 2016-04-15
[FLOC-4331] Restore some logging

sarum90:
We turned a lot of logging off in order to meet Swisscom benchmarking requirements.

This turns back on only the node configuration log in the agents during convergence.

The bad news is that this does log all datasets that this node has deleted (since the beginning of time). Although from preliminary benchmarking the additional CPU usage of adding this logging back is negligible (spun up 280 containers on 20 nodes with and without this PR, and couldn't see any real difference in CPU usage with the write benchmark). Although, it is probably worth mentioning that from master I was able to delete all of them all at once, whereas with this  PR I started to get timeout errors.

Nevertheless this PR will make our acceptance tests diagnosable once more.

### DIFF
--- a/flocker/common/logging.py
+++ b/flocker/common/logging.py
@@ -1,0 +1,75 @@
+# -*- test-case-name: flocker.common.test.test_logging -*-
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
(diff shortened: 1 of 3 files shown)


moshez:
LGTM

PR #2768, opened 2016-04-22
[FLOC-4376] Fix client on Ubuntu Wily tests

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4376

The Ubuntu 15.10 `pip` works (with the Ubuntu packaged python-requests) until you use it to install a newer version of requests, which no longer has the required `from requests.compat import IncompleteRead` API.

See:
 * https://bugs.launchpad.net/ubuntu/+source/python-pip/+bug/1306991/comments/47

My solution is to install the latest version of Pip in the package build images and never install the Ubuntu or Centos packaged `pip`.

We should probably do the same in the Jenkins slave images...except there, we'd need to install the latest version of virtualenv in order to get the latest version of pip for the slaves to install Flocker during tests.

Note that since building and pushing `clusterhqci/fpm-ubuntu-wily` here: 
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/fix-client-wily-tests-FLOC-4376/view/cron/job/_run_docker_build_ubuntu_wily_fpm/

The Ubuntu wily client installation tests now pass on master here: 
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/view/master/job/master/view/client/job/run_client_installation_on_Ubuntu_Wily/lastSuccessfulBuild/

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -587,8 +587,10 @@ common_cli:
           --assumeyes epel-release" >> Dockerfile
     echo "RUN env URLGRABBER_DEBUG=1 yum install --assumeyes \
           git ruby-devel python-devel libffi-devel openssl-devel \
-          python-pip rpmlint" >> Dockerfile
+          rpmlint" >> Dockerfile
     echo "RUN env URLGRABBER_DEBUG=1 yum update --assumeyes" >> Dockerfile
+    echo "ADD https://bootstrap.pypa.io/get-pip.py /tmp/" >> Dockerfile
+    echo 'RUN ["/usr/bin/env", "python2.7", "/tmp/get-pip.py"]' >> Dockerfile
     echo "COPY requirements.txt /tmp/" >> Dockerfile
     # Is this unused? --process-depency-links added but sort of untested.
     echo "RUN pip install --process-dependency-links -r /tmp/requirements.txt" >> Dockerfile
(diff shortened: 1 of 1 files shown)


sarum90:
Looks good to me. Sort of sad about reproducibility (https://bootstrap.pypa.io/get-pip.py is expected to point at newer and newer versions of pip over time, right?) but I'd much rather have green tests over red tests (and I think that's true of `apt-get install python-pip` as well, so this certainly is not a regression).

sarum90:
Thanks for cleaning this up!

wallrj:
> Sort of sad about reproducibility (https://bootstrap.pypa.io/get-pip.py is expected to point at newer and newer versions of pip over time, right?)

Right, but it turns out I can pin the version that it installs.


```
(get_pip) [/tmp]$ python get-pip.py pip==7.0.0
Collecting pip==7.0.0
  Downloading pip-7.0.0-py2.py3-none-any.whl (1.1MB)
    100% |████████████████████████████████| 1.1MB 649kB/s 
Collecting setuptools
  Using cached setuptools-20.10.1-py2.py3-none-any.whl
Collecting wheel
  Downloading wheel-0.29.0-py2.py3-none-any.whl (66kB)
    100% |████████████████████████████████| 71kB 4.9MB/s 
Installing collected packages: pip, setuptools, wheel
Successfully installed pip-7.0.0 setuptools-20.10.1 wheel-0.29.0
You are using pip version 7.0.0, however version 8.1.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

```

I think I'd rather have the latest version of `pip` and stop worrying about it....but we just talked about reproducibility, etc...

PR #2769, opened 2016-04-26
[FLOC-4381] Work around the broken lshw encoder in flocker-benchmark hardware-report

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4381

The problem is described here:
 * https://bugs.launchpad.net/ubuntu/+source/lshw/+bug/1405873

The output is broken in various ways so doesn't seem worth attempting to fix it up.
 * http://www.ezix.org/project/query?status=assigned&status=new&status=reopened&component=lshw&description=~json&order=priority

I'm also not convinced that we need to restrict the information output by the tool. 
It's probably easier to extract the information we need before it gets recorded in the benchmark results database....if we ever get around to deploying that.

I guess it may already be deployed somewhere but I don't know how to access it:
 * https://github.com/ClusterHQ/benchmark-server

Any ideas @avg-I ?

### DIFF
--- a/flocker/node/benchmark.py
+++ b/flocker/node/benchmark.py
@@ -50,9 +50,7 @@ def hardware_report(options):
     """
     Print a hardware report to stdout.
     """
-    sys.stdout.write(
-        list_hardware(['processor', 'memory', 'network', 'disk', 'volume'])
-    )
+    sys.stdout.write(list_hardware())
     return succeed(None)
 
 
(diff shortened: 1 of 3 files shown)


avg-I:
I think that Bridgett worked on deploying a "production" instance of the benchmark server. Maybe Marcus knows about it as well.

sarum90:
I don't know of any deployed 'production' benchmarking server. Seems reasonable to simplify the interface to this function if we have no consumers of it with the arguments.  I'm pretty sure today the results of benchmark.py are manually consumed when it runs, so we can filter then, and reduce the complexity of the code here.

Feel free to merge whenever.

Thanks!

Issue #120, opened 2014-06-23
Create a clusterhq/test-http-server docker container for use in functional tests


wallrj:
**Replaced by** https://clusterhq.atlassian.net/browse/FLOC-120

We're currently using `openshift/busybox-http-app` in our `geard` functional tests.
 * https://github.com/hybridlogic/flocker/blob/master/flocker/node/functional/test_gear.py#L63

But it's an externally controlled container and could change (or go away) at any time, breaking our functional tests.

The test http server also seems to return non-compliant headers which makes it difficult for us to parse its response using standards compliant tools like twisted.web.

```
(flocker-91)[vagrant@localhost vagrant]$ trial flocker.node.functional.test_gear.GearClientTests.test_add_with_port
flocker.node.functional.test_gear
  GearClientTests
    test_add_with_port ...                                               [FAIL]

===============================================================================
[FAIL]
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/flocker-91/lib/python2.7/site-packages/twisted/trial/_synctest.py", line 356, in assertEqual
    % (msg, pformat(first), pformat(second)))
twisted.trial.unittest.FailTest: not equal:
a = 'HTTP/1.1 200 OK\nContent-Type: text/plain;charset=utf-8\nContent-Length: 12\n\nHello world!\n'
b = 'Hello world'


flocker.node.functional.test_gear.GearClientTests.test_add_with_port
-------------------------------------------------------------------------------
Ran 1 tests in 22.254s
```

adamtheturtle:
We are moving our development planning to JIRA. This issue is now being tracked at https://clusterhq.atlassian.net/browse/FLOC-120. You are welcome to file additional issues in GitHub if that's easier for you.

PR #2771, opened 2016-04-27
[FLOC-4382] Docker plugin tests now wait for test volumes to be mounted

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4382

The `docker volume create` API returns immediately and the `docker create --volume-plugin=flocker ...`  && `docker start` APIs don't wait for the *existing* flocker volume to be mounted before starting the container.

The test has been failing since 2016-04-12....around the time of the Docker 1.11.0 release so perhaps that's a change in Docker behaviour.

 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/view/master/job/master/view/cron/job/_run_acceptance_on_AWS_CentOS_7_with_EBS/lastCompletedBuild/testReport/flocker.acceptance.endtoend.test_dockerplugin/DockerPluginTests/test_create_sized_volume_with_v2_plugin_api/

```
Failing for the past 12 builds (Since Failed#120 )
Took 8 ms.
```
And build #120 was for revision 312403043142d0119ee43fb1e110325241bda47b on April 12.

 * https://github.com/docker/docker/blob/master/CHANGELOG.md#1110-2016-04-13

So I've added a poll_until after the Docker volume create step which waits until ``docker volume ls`` reports a mountpoint for the volume.

Also added some better error reporting to the `lsblk` command since that was failing with stderr message "/dev/xvdf is not a block device" but that was  missing from the test data.

### DIFF
--- a/flocker/acceptance/endtoend/test_dockerplugin.py
+++ b/flocker/acceptance/endtoend/test_dockerplugin.py
@@ -15,7 +15,7 @@ from bitmath import GiB
 
 from eliot import Message
 
-from ...common import loop_until
+from ...common import loop_until, poll_until
 from ...common.runner import run_ssh
 
 from ...dockerplugin.test.test_api import volume_expression
@@ -145,6 +145,17 @@ class DockerPluginTests(AsyncTestCase):
         """
         result = client.create_volume(name, u'flocker', driver_opts)
         self.addCleanup(client.remove_volume, name)
+
+        # The docker volume API doesn't block.
+        # Wait until the volume has been mounted.
+        poll_until(
+            predicate=lambda: list(
+                v for v in client.volumes()["Volumes"]
+                if v["Name"] == name and v["Mountpoint"]
+            ),
+            steps=[1]*60
+        )
+
         return result
 
     def _test_sized_vol_container(self, cluster, node):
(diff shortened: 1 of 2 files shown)


wallrj:
The GCE acceptance test failure is fixed in #2769: 
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/lsblk-FLOC-4382/job/_run_acceptance_on_GCE_CentOS_7_with_GCE/2/

flocker/acceptance/endtoend/test_dockerplugin.py
```diff
@@ -145,6 +145,17 @@ def _create_volume(self, client, name, driver_opts):
         """
         result = client.create_volume(name, u'flocker', driver_opts)
         self.addCleanup(client.remove_volume, name)
+
+        # The docker volume API doesn't block.
+        # Wait until the volume has been mounted.
```
sarum90: I think we should consider renaming this function with this addition. It used to just be a light wrapper on the underlying create_volume API, but now that you added the wait until, perhaps rename this mount_volume or at least change the docstring?

sarum90:
Looks good, thanks!

PR #2775, opened 2016-05-02
[FLOC-4312] Port acceptance tests to RHEL 7.2

myechuri:
Not ready for review.

Port acceptance tests to run on RHEL 7.2. Tested standalone on RHEL buildslave reporting to a staged buildmaster.

### DIFF
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -301,10 +301,7 @@ class AWSProvisioner(PClass):
 
     def create_node(self, name, distribution, metadata={}):
         size = self._default_size
-        disk_size = 8
-        if distribution == 'rhel-7.2':
-            # RHEL 7.2 official image snapshot is 10GB
-            disk_size = 10
+        disk_size = 10
 
         with start_action(
             action_type=u"flocker:provision:aws:create_node",
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -36,6 +36,7 @@ _usernames = {
     'centos-7': 'centos',
     'ubuntu-14.04': 'ubuntu',
     'ubuntu-15.10': 'ubuntu',
+    'rhel-7.2': 'ec2-user',
 }
 
 
--- a/flocker/provision/_aws.py
+++ b/flocker/provision/_aws.py
@@ -48,6 +48,8 @@ IMAGE_NAMES = {
     # https://cloud-images.ubuntu.com/locator/ec2/
     'ubuntu-14.04': 'ubuntu/images/hvm-ssd/ubuntu-trusty-14.04-amd64-server-20160222',  # noqa
     'ubuntu-15.10': 'ubuntu/images/hvm-ssd/ubuntu-wily-15.10-amd64-server-20160226',  # noqa
+    # RHEL 7.2 HVM GA image
+    'rhel-7.2': 'RHEL-7.2_HVM_GA-20151112-x86_64-1-Hourly2-GP2',  # noqa
 }
 
 BOTO_INSTANCE_NOT_FOUND = u'InvalidInstanceID.NotFound'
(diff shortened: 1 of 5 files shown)


flocker/provision/_install.py
```diff
@@ -519,14 +545,18 @@ def task_package_install(package_name, distribution,
...
+            package_distribution = 'centos-7'
         result_path = posixpath.join(
-            '/results/omnibus/', package_source.branch, distribution)
+            '/results/omnibus/', package_source.branch, package_distribution)
         base_url = urljoin(package_source.build_server, result_path)
     else:
         base_url = None
 
-    if is_centos(distribution):
+    if is_centos(distribution) or is_rhel(distribution):
```
moshez: We should probably have a function "is_redhatish"?

flocker/provision/_install.py
```diff
@@ -1250,6 +1280,7 @@ def maybe_disable(unit):
 _flocker_uninstallers = {
     "ubuntu-14.04": _uninstall_flocker_ubuntu1404,
     "centos-7": _uninstall_flocker_centos7,
+    "rhel-7.2": _uninstall_flocker_centos7,
```
moshez: Why is this not failing on ubuntu 15.10?

flocker/provision/_install.py
```diff
@@ -519,14 +545,18 @@ def task_package_install(package_name, distribution,
     :return: a sequence of commands to run on the distribution
     """
     if package_source.branch:
-        # A development branch has been selected - add its Buildbot repo
```
moshez: This function mixes in distro-agnostic code and distro-specific code. Would prefer if it was split out.

flocker/provision/_aws.py
```diff
@@ -36,6 +36,7 @@
     'centos-7': 'centos',
     'ubuntu-14.04': 'ubuntu',
     'ubuntu-15.10': 'ubuntu',
+    'rhel-7.2': 'ec2-user',
```
moshez: There are a lot of places which we need to change to support a new distribution. We should probably localize them into the [Strategy Pattern](https://en.wikipedia.org/wiki/Strategy_pattern)

flocker/provision/_install.py
myechuri: Fixed.

flocker/provision/_install.py
myechuri: 15.10 support is still partial (https://docs.clusterhq.com/en/latest/supported/index.html#supported-operating-systems does not list 15.10, and we do not have tests running on 15.10 except for benchmark work). Thanks for pointing out. @wallrj is planning to work on 16.10 support - we will update the thread with 15.10 plan soon.

myechuri: Confirming that 15.10 support is no longer required, based on in-person discussions. We only have client support on 15.10 for benchmarking work.

flocker/provision/_install.py
myechuri: Thanks for suggesting! Fixed.

flocker/provision/_aws.py
myechuri: Deferring to [FLOC-4386](https://clusterhq.atlassian.net/browse/FLOC-4386) .

flocker/provision/_install.py
```diff
@@ -95,16 +95,29 @@ def tag_as_test_install(flocker_version, distribution, package_name):
...
+def is_centos_or_rhel(distribution):
+    """
+    Determine whether the named distribution is a version of CentOS or RHEL.
+
+    :param bytes distribution: The name of the distribution to inspect.
+
+    :return: ``True`` if the distribution named is a version of CentOS or RHEL,
+        ``False`` otherwise.
+    """
+    return (distribution.startswith("centos-") or
```
moshez: Please use

    distribution.startswith(('censtos-', 'rhel-'))

myechuri: Thanks! Updated.

PR #2778, opened 2016-05-06
[FLOC-4391] Update to docker-compose 1.7.1 use the latest compose file format

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4391

The problem:
 * Our cloudformation template installs the `:latest` version of docker-swarm.
 * A new version of docker-swarm was released recently  `1.2.1 (2016-05-03)`.
 * Our cloudformation template installs an old version of `docker-compose` 
 * Our tutorial and tests both use `docker-compose` v1 configuration file format rather than v2.
 * The old docker-compose and the  latest docker-swarm no longer create volumes (via volume plugin) on-demand.

Updating to latest docker-compose and using the v2 compose file format allows us to specify the volumes and the plugin options in a separate section of the config file.
And with this, `docker-compose` will now create the listed volumes before starting the services.

I also encountered https://clusterhq.atlassian.net/browse/FLOC-4392 (flocker-docker-plugin leaks file descriptors) while working on this.

Here are the results of the  updated installer test:
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/compose-file-v2-FLOC-4391/job/_run_admin_test_test_installer/1/


admin/installer/setup_client.sh
```diff
@@ -13,7 +13,7 @@ apt-get update
 sudo apt-get install -y postgresql-client
 
 # Get docker-compose.
-curl -L https://github.com/docker/compose/releases/download/1.5.2/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
+curl -L https://github.com/docker/compose/releases/download/1.7.1/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
```
wallrj: We should probably pre-install this in the image to reduce the risk of failures caused by flaky github services or network connections.

admin/installer/setup_swarm_manager.sh
```diff
@@ -11,10 +11,10 @@ DOCKER_CERT_HOME='/root/.docker'
 # Create Swarm cluster, and publish Swarm cluster ID to S3 bucket.
 rm -rf /tmp/swarm-config
 mkdir -p /tmp/swarm-config
-docker pull swarm:1.0.1
-docker run --rm -v ${DOCKER_CERT_HOME}:${DOCKER_CERT_HOME} swarm create > /tmp/swarm-config/swarm_cluster_id
+docker pull swarm:1.2.1
```
wallrj: This step isn't strictly necessary since we already `docker pull swarm 1.2.1` in the image building scripts.

admin/installer/setup_swarm_node.sh
```diff
@@ -7,8 +7,8 @@ set -ex
 DOCKER_CERT_HOME='/root/.docker'
 
 # Gather Swarm cluster id from S3 bucket.
-swarm_cluster_id=$(s3cmd_wrapper get --config=/root/.s3cfg s3://${s3_bucket}/swarm-config/swarm_cluster_id -) 
+swarm_cluster_id=$(s3cmd_wrapper get --config=/root/.s3cfg s3://${s3_bucket}/swarm-config/swarm_cluster_id -)
 
 # Start the Swarm node.
-docker pull swarm:1.0.1
-docker run -d -v ${DOCKER_CERT_HOME}:${DOCKER_CERT_HOME} swarm join --addr=$(/usr/bin/ec2metadata --public-ipv4):2375 token://$swarm_cluster_id
+docker pull swarm:1.2.1
```
wallrj: Same here.

admin/test/installer/postgres/docker-compose-node1.yml
```diff
@@ -1,12 +1,22 @@
...
-  volumes:
-     - 'postgres:/var/lib/postgresql/data'
+version: "2"
+
+volumes:
+  postgres:
+    driver: "flocker"
+    driver_opts:
+      size: "10GiB"
+      profile: "silver"
```
wallrj: I included driver options into the test, to ensure that they continue to work with compose and to reduce the size and cost of the volumes created by this test.

wallrj: Probably could switch this to bronze.

admin/test/installer/postgres/docker-compose-node1.yml
```diff
@@ -1,12 +1,22 @@
...
+  postgres:
+    image: "postgres:9.5"
+    ports:
+       -  "5432:5432"
+    environment:
+       - "constraint:flocker-node==1"
+       - "POSTGRES_USER=flocker"
+       - "POSTGRES_PASSWORD=flocker"
+       - "PGDATA=/var/lib/postgresql/data"
+    network_mode: "bridge"
```
wallrj: The containers wouldn't start unless I specified a network_mode.

admin/test/installer/postgres/docker-compose-node1.yml
```diff
@@ -1,12 +1,22 @@
...
+volumes:
+  postgres:
+    driver: "flocker"
+    driver_opts:
+      size: "10GiB"
+      profile: "silver"
+
+services:
+  postgres:
+    image: "postgres:9.5"
```
wallrj: I pinned the version of Postgres too.

admin/test/test_installer.py
```diff
@@ -35,7 +35,7 @@
 POSTGRESQL_USERNAME = 'flocker'
 POSTGRESQL_PASSWORD = 'flocker'
 
-CLOUDFORMATION_TEMPLATE_URL = "https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.json"  # noqa
+CLOUDFORMATION_TEMPLATE_URL = "https://s3.amazonaws.com/installer.downloads.clusterhq.com/flocker-cluster.cloudformation.FLOC-4391.json"  # noqa
```
wallrj: This is temporary. In order to test the new cloudformation script without overriding the existing one.

admin/test/test_installer.py
```diff
@@ -387,49 +386,6 @@ def _cleanup_flocker(self):
...
-                self.docker_host,
-                self.compose_node2.path,
-                'rm', '-f'
-            ).addErrback(
-                # This sometimes fails with exit code 255
-                # and a message ValueError: No JSON object could be decoded
-                lambda failure: failure.trap(ProcessTerminated)
-            )
-        )
-        return gather_deferreds([d_node1_compose, d_node2_compose])
```
wallrj: `docker-compose down` stops and removes the containers related to the compose file so these commands are no longer needed.

admin/test/test_installer.py
```diff
@@ -502,13 +458,23 @@ def upload_docker_compose_files():
...
 
-        # This isn't in the tutorial, but docker-compose doesn't retry failed
-        # pulls and pulls fail all the time.
+        def cleanup_container(ignored):
+            self.addCleanup(
+                remote_docker_compose,
+                self.client_node_ip,
+                self.docker_host,
+                self.compose_node1.path,
+                'down'
```
wallrj: This is a new command which both stops and removes the containers....which simplifies things here and in the docs (later).

admin/test/test_installer.py
```diff
@@ -502,13 +458,23 @@ def upload_docker_compose_files():
...
+
+        # docker-compose doesn't retry failed pulls and pulls fail all the
+        # time.
         def pull_postgres():
-            return remote_docker(
+            return remote_docker_compose(
                 self.client_node_ip,
                 self.docker_host,
-                'pull', 'postgres:latest'
+                self.compose_node1.path, 'pull'
```
wallrj: Using `docker-compose pull` ensures that we're pulling the same version of the image that's used in the config file and pulls to all the docker swarm hosts.

coreos/fleet/fleet-flocker-control.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
```
wallrj: Oops. This shouldn't be here. I'll delete it.

docs/docker-integration/tutorial-swarm-compose.rst
```diff
@@ -92,22 +92,6 @@ The Docker Compose files both have the same layout, as illustrated below, except
 
   * Flocker will automatically provision this volume on-demand if it doesn't already exist.
 
-Step 2.1: Optionally specify size and profile
----------------------------------------------
```
wallrj: I removed this whole section in favour of adding an example of the plugin options into the sample configuration file. 
I think the options are self-explanatory.
These old commands no-longer work because `docker-compose` adds a prefix to the names of the volumes it expects (or creates).
These old commands would only work if we specified `external` volumes in the compose file.
 * https://docs.docker.com/compose/compose-file/#volume-configuration-reference

admin/test/installer/postgres/docker-compose-node1.yml
myechuri: @wallrj : switching to bronze might be slightly better since bronze is mentioned in the tutorial?

myechuri:
Looks good, @wallrj . Please go ahead after pending CI checks complete and changing CLOUDFORMATION_TEMPLATE_URL. Thanks!

PR #2780, opened 2016-05-09
[FLOC-4395] Update Flocker to use pyrsistent with PClass.set optimizations.

sarum90:
I just added an optimization to pyrsistent that increases the chance of getting the exact same object for equivalent PClasses.  Preliminary tests in flocker indicate that this greatly helps our comparisons when we are looking up PClasses in our lru caches.


sarum90:
https://github.com/ClusterHQ/pyrsistent/commit/f881840273db5292c64289d9bccab162ad41dfec is the commit we'll be pinned to after this.

wallrj:
LGTM. 

It'd be nice if we also incremented a version number, but I remember now that that ball's in my court  (#7) :frowning:

sarum90:
I think that might actually be referring to https://github.com/ClusterHQ/pyrsistent/pull/7 but #7 seemed topical as well.

sarum90:
Thanks, merging.

PR #2781, opened 2016-05-09
[FLOC-4396] Only send updates on the second to nodes (for broadcast)

sarum90:
First pass implementation of batched updates.

I'll probably revisit this to reduce calls further. This greatly reduces the number of broadcasts, but a decent number of requests, especially when we are busy, are done via the non-broadcast "send update once last update is acknowledged" codepath. It would be much faster (just for amortized logging reasons) to figure out how to collapse all of those into a single call as well. I'll be looking to amortize all of those as well.

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -519,6 +519,10 @@ class ControlAMPService(Service):
     :ivar dict _current_command: A dictionary containing information about
         connections to which state updates are currently in progress.  The keys
         are protocol instances.  The values are ``_UpdateState`` instances.
+    :ivar IReactorTime _reactor: An ``IReactorTime`` provider to be used to
+        schedule delays in sending updates.
+    :ivar bool _broadcast_scheduled: Whether there is a currently pending
+        broadcast of state to all connections.
     """
     logger = Logger()
 
@@ -534,6 +538,8 @@ class ControlAMPService(Service):
         :param context_factory: TLS context factory.
         """
         self.connections = set()
+        self._reactor = reactor
+        self._broadcast_scheduled = False
         self._current_command = {}
         self.cluster_state = cluster_state
         self.configuration_service = configuration_service
@@ -546,8 +552,7 @@ class ControlAMPService(Service):
             )
         )
         # When configuration changes, notify all connected clients:
-        self.configuration_service.register(
-            lambda: self._send_state_to_connections(self.connections))
+        self.configuration_service.register(self._schedule_broadcast_update)
 
     def startService(self):
         self.endpoint_service.startService()
@@ -709,6 +714,28 @@ class ControlAMPService(Service):
         """
         self.connections.remove(connection)
 
+    def _execute_broadcast_update(self):
+        """
+        Actually executes a broadcast update to all current connections.
+        """
+        self._broadcast_scheduled = False
+        self._send_state_to_connections(self.connections)
+
+    def _schedule_broadcast_update(self):
+        """
+        Ensure that there is a pending broadcast update call.
+
+        This is called when the state or configuration is updated, to trigger
+        a broadcast of the current state and configuration to all nodes.
+
+        In general, it only schedules an update to be broadcast 1 second later
+        so that if we recieve multiple updates within that second they are
+        coalesced down to a single update.
+        """
+        if not self._broadcast_scheduled:
+            self._broadcast_scheduled = True
+            self._reactor.callLater(1.0, self._execute_broadcast_update)
+
     def node_changed(self, source, state_changes):
         """
         We've received a node state update from a connected client.
@@ -719,7 +746,7 @@ class ControlAMPService(Service):
             providers representing the state change which has taken place.
         """
         self.cluster_state.apply_changes_from_source(source, state_changes)
-        self._send_state_to_connections(self.connections)
+        self._schedule_broadcast_update()
 
 
 class IConvergenceAgent(Interface):
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -510,6 +510,13 @@ class _UpdateState(PClass):
     next_scheduled = field()
 
 
+# The control service waits this long before sending any update to an agent.
+# This allows for a batch of updates to build up, and effectively puts a cap on
+# the maximum number of updates the control node will have to send over any
+# fixed period of time.
+CONTROL_SERVICE_BATCHING_DELAY = 1.0
+
+
 class ControlAMPService(Service):
     """
     Control Service AMP server.
@@ -523,6 +530,9 @@ class ControlAMPService(Service):
         schedule delays in sending updates.
     :ivar bool _broadcast_scheduled: Whether there is a currently pending
         broadcast of state to all connections.
+    :ivar IDelayedCall _current_pending_broadcast_delayed_call: The
+        ``IDelayedCall`` provider for the currently pending call to broadcast a
+        state/configuration update.
     """
     logger = Logger()
 
@@ -540,6 +550,7 @@ class ControlAMPService(Service):
         self.connections = set()
         self._reactor = reactor
         self._broadcast_scheduled = False
+        self._current_pending_broadcast_delayed_call = None
         self._current_command = {}
         self.cluster_state = cluster_state
         self.configuration_service = configuration_service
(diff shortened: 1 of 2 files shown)


flocker/control/_protocol.py
```diff
@@ -709,6 +714,28 @@ def disconnected(self, connection):
...
+
+    def _schedule_broadcast_update(self):
+        """
+        Ensure that there is a pending broadcast update call.
+
+        This is called when the state or configuration is updated, to trigger
+        a broadcast of the current state and configuration to all nodes.
+
+        In general, it only schedules an update to be broadcast 1 second later
+        so that if we recieve multiple updates within that second they are
```
wallrj: typo "receive"

flocker/control/_protocol.py
```diff
@@ -709,6 +714,28 @@ def disconnected(self, connection):
...
+        coalesced down to a single update.
+        """
+        if not self._broadcast_scheduled:
+            self._broadcast_scheduled = True
+            self._reactor.callLater(1.0, self._execute_broadcast_update)
```
wallrj: :question: Might be nice to make it `callLater(BROADCAST_DELAY)`

flocker/control/test/test_protocol.py
```diff
@@ -563,6 +563,7 @@ def test_nodestate_notifies_all_connected(self):
             self.client.callRemote(NodeStateCommand,
                                    state_changes=(NODE_STATE,),
                                    eliot_context=TEST_ACTION))
+        self.reactor.advance(10.0)
```
wallrj: :question: Then this could be `advance(BROADCAST_DELAY + 1)` which might be clearer.

flocker/control/test/test_protocol.py
```diff
@@ -574,6 +575,55 @@ def test_nodestate_notifies_all_connected(self):
...
+
+        agents = [FakeAgent(), FakeAgent()]
+        clients = list(AgentAMP(Clock(), agent) for agent in agents)
+        servers = list(LoopbackAMPClient(client.locator) for client in clients)
+
+        for server in servers:
+            self.control_amp_service.connected(server)
+
+        initial_update_counts = list(
+            agent.cluster_updated_count for agent in agents)
```
wallrj: :question: Is this `1` because of the cluster state unicast that happens when the agent first connects to the control service?

wallrj: Maybe add a comment to that effect.

flocker/control/_protocol.py
moshez: We should probably save the result of the callLater (for example, to properly cancel if we're shutting down).

flocker/control/test/test_protocol.py
```diff
@@ -574,6 +575,55 @@ def test_nodestate_notifies_all_connected(self):
...
+
+        for i in xrange(10):
+            new_state = NODE_STATE.set(
+                'applications',
+                NODE_STATE.applications.add(
+                    Application(name=('app-%d' % i),
+                                image=DockerImage.from_string('image-%d' % i))
+                )
+            )
```
wallrj: :question: So this is like a single client sending 10 state updates instantly. 
 * Is that realistic? 
 * Does it matter if it's not realistic?
 * Wouldn't this be coalesced by the only-send-update-if-agent-acknowledged-the-last-update optimization, rather than the delay-update-by-1s optimization?

wallrj: I'm talking about the `_delayed_update_connection` thing in case that's not clear.

flocker/control/test/test_protocol.py
```diff
@@ -679,26 +729,60 @@ def test_configuration_change(self):
...
         client = AgentAMP(Clock(), agent)
-        service = build_control_amp_service(self)
+        service_clock = Clock()
+        service = build_control_amp_service(self, service_clock)
         service.startService()
         server = LoopbackAMPClient(client.locator)
         service.connected(server)
 
         service.configuration_service.save(TEST_DEPLOYMENT)
+        service_clock.advance(10.0)
```
wallrj: :question: maybe use `BROADCAST_DELAY + 1` here too and elsewhere in the tests....or add an explanatory comment.

flocker/control/test/test_protocol.py
```diff
@@ -679,26 +729,60 @@ def test_configuration_change(self):
...
+        service.connected(server)
+
+        initial_updates_count = agent.cluster_updated_count
+        for i in xrange(10):
+            service.configuration_service.save(
+                TEST_DEPLOYMENT.transform(
+                    ['nodes'],
+                    lambda nodes: nodes.union(frozenset([
+                        Node(hostname='%d.example.com' % i)
+                    ]))
```
wallrj: :question: Pylint calls this out, and now I look at it, I'm not sure I understand it. Why not `nodes.add(Node(hostname='%d.example.com' % i)` ?

flocker/control/test/test_protocol.py
sarum90: * Probably not realistic. (state updates should actually come from different nodes)
* Close enough approximation (If I take out the code I wrote, this test fails, yay TDD)
* I actually had to remove the DelayedAmpClient code from this test so that this was not the case. Without the delayed update code the clients respond immediately, so cluster_updated_count increments 10 times in this loop, just as desired.

wallrj:
Thanks @sarum90 

This looks great. A few further points:
 * I'm can't quite picture how this optimization interacts with the older optimization of only sending updates to clients that have acknowledged the last update. That was done to prevent unacknowledged updates stacking up at the agent and using tons of buffer memory while the updates are processed.
 * Do we still need that mechanism? I guess so.
 * In the PR you describe this as a NAGLE, but it's not really, right? NAGLE is to do with unacknowledged updates and maybe a moving timer that only fires after 1s of inactivity and unsent updates....or after 2s if updates are still coming in. In the past you've described this as batching, which seems more accurate.
 * There's a lint error...a weird one...
 * Perhaps we could measure the effectiveness of this change by counting the number of logged state updates in acceptance test remote_logs, before and after.

Please merge when you're happy with the branch.

wallrj:
> I'm can't quite picture how this optimization interacts with the older optimization of only sending updates to clients that have acknowledged the last update. 

Sorry @sarum90 I see that you've commented on this in the PR description. I missed that.

sarum90:
> Perhaps we could measure the effectiveness of this change by counting the number of logged state updates in acceptance test remote_logs, before and after.

With: 356 send_cluster_state instances: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/FLOC-4396-only-send-updates-on-the-second/job/run_acceptance_loopback_on_AWS_Ubuntu_Trusty_for_flocker.acceptance/1/artifact/remote_logs.log

Without: 482 send_cluster_state instances: http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/master/job/run_acceptance_loopback_on_AWS_Ubuntu_Trusty_for_flocker.acceptance/390/artifact/remote_logs.log

flocker/control/_protocol.py
sarum90: Done

flocker/control/_protocol.py
sarum90: Used a slightly different name for the constant. Added storing the callLater, and a test to verify that we cancel all callLaters once stopService has been called.

flocker/control/test/test_protocol.py
sarum90: Did *2 rather than +1.... no real reason.

flocker/control/test/test_protocol.py
sarum90: I actually think this might be filled with a bunch of zeros, or maybe [1, 2, 3, 4, 5...] from the initial state updates. The real reason for taking these values is just because I only want to measure the number of times things are sent between the node state updates.

flocker/control/test/test_protocol.py
sarum90: Done throughout.

flocker/control/test/test_protocol.py
sarum90: That's much better. But I also just stumbled upon `arbitrart_tranformation` which conveys what I want.

PR #2782, opened 2016-05-10
[FLOC-4397] Amortize updates some more

sarum90:
This PR makes it so that the control node will only ever send updates to nodes once every second even when it is recovering from a delayed update for a node that took awhile to acknowledge the original update.

### DIFF
--- a/flocker/control/test/test_protocol.py
+++ b/flocker/control/test/test_protocol.py
@@ -829,8 +829,8 @@ class ControlAMPServiceTests(ControlTestCase):
         service_clock.advance(CONTROL_SERVICE_BATCHING_DELAY*2)
 
         # Before any of the nodes respond, update configuration again
-        service.configuration_service.save(
-            arbitrary_transformation(configuration))
+        final_configuration = arbitrary_transformation(configuration)
+        service.configuration_service.save(final_configuration)
         service_clock.advance(CONTROL_SERVICE_BATCHING_DELAY*2)
 
         initial_update_counts = list(
@@ -842,8 +842,8 @@ class ControlAMPServiceTests(ControlTestCase):
             # Let some negligible amount of time pass
             service_clock.advance(0.001)
 
-        # Each agent should have the original delayed response, but they should
-        # not get the second response until a second has passed.
+        # Agents should not get any update until CONTROL_SERVICE_BATCHING_DELAY
+        # has passed, even for these delayed updates.
         self.assertEqual(
             [0] * len(agents),
             list(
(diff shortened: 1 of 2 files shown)


flocker/control/test/test_protocol.py
```diff
@@ -798,6 +800,66 @@ def test_coalesced_sends_within_time(self):
...
+            agent.cluster_updated_count for agent in agents
+        )
+
+        for server in delayed_servers:
+            server.respond()
+            # Let some negligible amount of time pass
+            service_clock.advance(0.001)
+
+        # Each agent should have the original delayed response, but they should
+        # not get the second response until a second has passed.
```
wallrj: :question: The word "original" confuses me a bit here. 
The one an only state / configuration  update will contain the *second* configuration change , right?

flocker/control/test/test_protocol.py
```diff
@@ -798,6 +800,66 @@ def test_coalesced_sends_within_time(self):
...
+        self.assertEqual(
+            [0] * len(agents),
+            list(
+                agent.cluster_updated_count - c
+                for agent, c in zip(agents, initial_update_counts)
+            )
+        )
```
wallrj: :question: Did you consider storing the actual state objects received instead of / as well as the state update count?

I think the tests might be a little clearer if you could assert which state had been sent and received by the agents.

wallrj:
Thanks @sarum90 

This looks great. Some thoughts:

 * Did you get any sense of the CPU reduction from this change?
 * This change makes it more likely that all agents will  receive updates at the same time even if some of them are slower to acknowledge the updates. So the entire cluster may experience regular CPU spikes. Should we introduce some jitter?
 * Is "amortize" the right word here? I had to look it up, but amortize seems to imply that the agents receive every status update, but at a fixed rate. Where as here it seems like were writing off some of the state updates that should have been delivered to the agents. Am I being pedantic? :-)

Good night. Please merge when your happy with the shape of the branch.

sarum90:
- Unfortunately I don't know the CPU savings from this change (and shame on me I haven't actually tried it on a full cluster yet). What I can say off the record is that after I did the broadcast optimization it reduced CPU nicely. Once that was done, I found that more CPU was spent on these delayed sending than in the broadcasts, and that a lot of the costs from the delayed updates are eliminated if you group all the sendings together (for instance you only have to serialize the config and the state once (or look up the serialization in a cache which is still a bit expensive), and send it to all of the connections, rather than serializing it once per connection). So I'm hopeful to see a similar CPU improvement.  The other thing is I think this will allow me to not actually take out the logging of the state/config that we are sending to all the nodes. this PR guarantees that we only log it at most once per second, which seems much more agreeable than logging it as fast as we can.  I'll be testing this at scale tomorrow, and promise to roll back if anything terrible happens.

- So I think this is only sending updates to the nodes in `_connections_pending_update` so while all of the delayed nodes will get CPU spikes at the same time, the whole cluster still gets spikes on the previous schedule. We've always done broadcasts in the past, and cross cluster CPU spike hasn't been too much of a concern so far, so I think I'd wait to try to add jitter till we have evidence we need it. Certainly don't want to add jitter that undoes the work of batching all the updates together.

- So this might send state to the nodes a bit less for the batching argument, but I think a lot of the savings are just because we will now only be logging once for all of the delayed updates that happen within a second, rather than logging for each delayed update that happens within a second. That sort of optimization seems very much along the lines of: https://en.wikipedia.org/wiki/Amortized_analysis. Either way it's just in the PR description/branch name, so only of concern for historical audits of the code. I don't think I've dropped that term in the code anywhere.

Now on to code comments...

flocker/control/test/test_protocol.py
sarum90: Yup that was an old confused comment. Fixed.

flocker/control/test/test_protocol.py
sarum90: Added check at the end. The agents aren't sending any state, just receiving, so I just verified at the end that they did get the correct final configuration.

PR #2784, opened 2016-05-10
[FLOC-4398] use murmur hash instead of sha

sarum90:
In profiling the control agent I got to a point where it was clear that using a secure hashing function to attempt to uniquely identify a configuration was consuming a significant amount of our time. This specifically happens in workloads with a lot of configuration changes (like deleting all datasets at the end of a benchmarking run). These might not be particularly relevant workloads in the real world, but one could make an argument that since the flocker control node requires making individual requests for creating or deleting datasets there are situations where this should be faster; if anyone wants to delete N disks, they have to make N requests in quick succession.

Given that our specific use case does not require that the hash be cryptographically secure, we can get a lot of CPU back by simply using a faster hashing algorithm.

### DIFF
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -4,11 +4,12 @@
 Persistence of cluster configuration.
 """
 
-from json import dumps, loads
-from uuid import UUID
+from base64 import b16encode
 from calendar import timegm
 from datetime import datetime
-from hashlib import sha256
+from json import dumps, loads
+from mmh3 import hash_bytes
+from uuid import UUID
 
 from eliot import Logger, write_traceback, MessageType, Field, ActionType
 
@@ -515,7 +516,7 @@ class ConfigurationPersistenceService(MultiService):
         """
         config = Configuration(version=_CONFIG_VERSION, deployment=deployment)
         data = wire_encode(config)
-        self._hash = sha256(data).hexdigest()
+        self._hash = b16encode(hash_bytes(data)).lower()
         self._config_path.setContent(data)
 
     def save(self, deployment):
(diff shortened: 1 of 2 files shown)


flocker/control/_persistence.py
```diff
@@ -4,11 +4,12 @@
...
 """
 
-from json import dumps, loads
-from uuid import UUID
+from base64 import b16encode
 from calendar import timegm
 from datetime import datetime
-from hashlib import sha256
+from json import dumps, loads
+from mmh3 import hash_bytes
```
moshez: Can you do

    from mmh3 import hash_bytes as mmh3_hash_bytes

So below it's clearer what hashing you're using.

myechuri:
@sarum90 : are there benchmark numbers for time spent on hashing before and after the change? Thanks!

sarum90:
Unfortunately there are only "take my word for it" numbers from when I had a large cluster spun up and was doing profiling. Not the best for documenting my work and justifying the changes. If you want me to improve the process let me know. Currently optimizing for velocity to get to Q3 requirements, and trying not to invest too heavily in improving the process when we are thinking about deprecating some of this code in the medium-term.

On the cluster I was playing with though, for a workload of "deleting all of the datasets":

Before: >5% of all CPU usage spent in hashing. (more expensive than writing the config to disk!)
After: <0.1% of all CPU usage spent in hashing. Or maybe even less, it was basically negligible.

flocker/control/_persistence.py
sarum90: yea that's definitely better, 👍 

Also apologies about sorting imports. Old habit.

sarum90:
Even though I got the green checkmark from the previous run... I think I'll wait for another green checkmark before merging. Thanks for the reviews!

myechuri:
> Before: >5% of all CPU usage spent in hashing. (more expensive than writing the config to disk!)
> After: <0.1% of all CPU usage spent in hashing. Or maybe even less, it was basically negligible.

Thanks for sharing!

PR #2785, opened 2016-05-10
[FLOC-4399] Only lookup the value once in the cache.

sarum90:
`bool(key in dict)` is one lookup in the dictionary, and `dict[key]` is another lookup.  We have already determined that looking up a large PClass in a dict is a slow operation, so we should avoid that. Using `dict.get` and then checking for None (or using `try...except IndexError`) should eliminate one of the checks.

### DIFF
--- a/flocker/control/_persistence.py
+++ b/flocker/control/_persistence.py
@@ -248,8 +248,9 @@ def _cached_dfs_serialize(input_object):
     else:
         if _is_pyrsistent(input_object):
             is_pyrsistent = True
-            if input_object in _cached_dfs_serialize_cache:
-                return _cached_dfs_serialize_cache[input_object]
+            cached_value = _cached_dfs_serialize_cache.get(input_object)
+            if cached_value is not None:
+                return cached_value
         obj = _to_serializables(input_object)
 
     result = obj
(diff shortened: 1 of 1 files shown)


moshez:
I agree it's a good idea, but this *is* a trade-off in code adherence to best practices in return for performance. Therefore, we should document that in a comment, and put numbers there.

(try/except is slower because exceptions are pretty slow in Python, but using None as a sentinel is never a great idea -- but it is the best option here).

sarum90:
I'll add some sort of comment...

I can make a new sentinel, that's no issue.

Actually on second though ... I'm not convinced performance is critical enough here to avoid the exception path (we don't want to navigate recursively down a PClass, but unwinding a few stack frames is probably okay). The main reason I didn't do that originally is because the exception isn't really exceptional (it's guaranteed to happen the first call for instance.)

I'll just switch to an exception style. It'll be more pythonic anyway. This is not a tight loop so the exception will be fine.

sarum90:
Eh actually you're right we don't want to incur the exception drop on the first invocation for every pyrsistent object, we should use a sentinel. I'll just add a comment.

PR #2786, opened 2016-05-10
[FLOC-4400] Change the default machine size in GCE tests.

sarum90:
This is a change that I've done locally for all of the benchmarking clusters that I've run. Decided it was probably safe to push to master and it puts GCE at par with the AWS instances that we spin up for tests.

On AWS we have the default set to m3.large, which is 2 CPU and 7.5 GiB of RAM. The n1-standard-2 on GCE is the equivalent machine. The n1-standard-1 was chosen previously because I incorrectly thought that flocker should never use more than 1 CPU. Turns out the control node on large clusters easily consumes an entire CPU and then some, and we need the spare for the agents and the journalctl daemon.


wallrj:
lgtm

PR #2787, opened 2016-05-10
[FLOC-4401] enable persistent connections in flocker client code

sarum90:
Two things going on here:

1) There `persistent` keyword only exists on `treq.request` and friends, not `treq.HTTPClient.request` and friends. Unfortunately treq.HTTPClient.request takes `kwargs` so it never tells you if you are passing it an argument it's never heard of, instead it just silently does nothing with the argument.
2) `treq.HTTPClient.request` by-default does not use persistent connections.

These two facts combine to show us that we had a lot of `persistent=False` lines in our code that could just be deleted.

It does not seem that `treq` today offers a convenient interface to use persistent connections (despite the fact that it appears to). Thus, we just remove all references to persistent connections, and hope the default stays such that persistent connections are not used.

~~After that bit of cleanup, I went about modifying our code that creates treq HTTPClients so that it could create ones that had connection pools with persistence, and the appropriate plumbing to (A) test that the object created was at least trying to be persistent and (B) actually use persistent connections from our benchmarking scripts.~~

~~This is mostly done so that we can spin up large clusters and launch many many containers without spending lots of time generating SSH keys and negotiating new connections.  If SwissCom is having troubles with the CPU usage of the control node, we can help them change their flocker client code to also use persistent connections.~~


flocker/apiclient/_client.py
```diff
@@ -578,7 +578,8 @@ class FlockerClient(object):
     A client for the Flocker V1 REST API.
     """
     def __init__(self, reactor, host, port,
-                 ca_cluster_path, cert_path, key_path):
+                 ca_cluster_path, cert_path, key_path,
+                 persistent_connections=False):
```
wallrj: Document this new argument.

Vagrantfile
```diff
@@ -0,0 +1,35 @@
...
+  # Use the git configuration from the host in the VM, if it is in the expected
+  # location
+  if File.exists?(File.join(Dir.home, ".gitconfig"))
+    config.vm.provision "file", source: File.join(Dir.home, ".gitconfig"), destination: ".gitconfig"
+  end
+
+  if Vagrant.has_plugin?("vagrant-cachier")
+    config.cache.scope = :box
+  end
+end
```
wallrj: :question: Nice! Are we sneaking this in the back door or is this a mistake?

sarum90: Oops, total mistake.... but I kinda don't want to undo it... maybe I should.  Anyway I'll wait until you are done with your review before I decide.

flocker/apiclient/test/test_client.py
```diff
@@ -736,8 +756,8 @@ def create_client(self):
         self.addCleanup(self.cluster_state_service.stopService)
         self.addCleanup(self.persistence_service.stopService)
         credential_set, _ = get_credential_sets()
-        credentials_path = FilePath(self.mktemp())
-        credentials_path.makedirs()
+        self.credentials_path = FilePath(self.mktemp())
+        self.credentials_path.makedirs()
```
wallrj: :question: Maybe switch to one of the new helper methods while you're here.

flocker/apiclient/test/test_client.py
```diff
@@ -751,11 +771,8 @@ def create_client(self):
         api_service.startService()
         self.addCleanup(api_service.stopService)
 
-        credential_set.copy_to(credentials_path, user=True)
-        return FlockerClient(reactor, b"127.0.0.1", self.port,
-                             credentials_path.child(b"cluster.crt"),
-                             credentials_path.child(b"user.crt"),
-                             credentials_path.child(b"user.key"))
+        credential_set.copy_to(self.credentials_path, user=True)
+        return self._create_flocker_client(False)
```
wallrj: :question: It'd be nice to use keyword argument here so that the reader doesn't have to go looking at the `create_flocker_client` signature.

flocker/apiclient/test/test_client.py
```diff
@@ -798,6 +815,21 @@ def test_logging(self, logger):
         return d
 
     @capture_logging(None)
+    def test_persistent_connections(self, logger):
+        """
+        ``FlockerClient`` can use a pool with persistent connections when
+        specified.
```
wallrj: :question: This is the bit that I'm slightly confused about. What does a pool of non-persistent connections even mean? 

And in my tests this morning it appeared that with `persistent_connections=True` my client established a new connection for each request and yet held it open....at least as far as I could see by running `netstat -antp` on client and server.
I may have been mistaken though.

wallrj: Here's someone else saying something similar:
 * http://blog.mailgun.com/stress-testing-http-with-twisted-python-and-treq/#comment-1087501883

Also note that `maxPersistentPerHost = 2 ` by default.

flocker/ca/_validation.py
```diff
@@ -130,11 +132,16 @@ def treq_with_authentication(reactor, ca_path, user_cert_path, user_key_path):
...
     :return: ``treq`` compatible object.
     """
     ca = Certificate.loadPEM(ca_path.getContent())
     user_credential = UserCredential.from_files(user_cert_path, user_key_path)
     policy = ControlServicePolicy(
         ca_certificate=ca, client_credential=user_credential.credential)
-    return HTTPClient(Agent(reactor, contextFactory=policy))
+    pool = None
+    if persistent_connections:
+        pool = HTTPConnectionPool(reactor, persistent_connections)
```
wallrj: :question: Perhaps we should always supply a connection pool, persistent or not.

wallrj:
Thanks @sarum90 

I've commented (above) that I didn't understand the behaviour and the number of connections I was seeing when I tested the FlockerClient with persistent_connections=True.

Having read the code, it seems that `HTTPConnectionPool` can container a mixture of non-persistent connections and up to `maxPersistentConnections` persistent connections.

There seem to be a bunch of bugs and confusion around the HTTPConnectionPool and persistent connections:
 * https://twistedmatrix.com/trac/search?q=persistent&noquickjump=1&ticket=on

I also tried to see, by profiling on the client side, whether this does indeed reduce the time spent in the SSL handshake.

As far as I could tell it doesn't. Perhaps because the HTTPConnectionPool doesn't actually limit the number of connections. Instead what seems to happen is that you create as many connections as there are outstanding HTTP requests.

I also think that I may have got confused and mislead you about persistent connection being the answer to the SSL handshake CPU problems.

I think I was mixing up persistent connections with HTTP pipelining, which is where you can send multiple HTTP requests down the same connection without waiting for responses to previous requests. That feature seems not to be implemented in Twisted web so far and may never be. Instead I think people are moving to HTTP2 multiplexing.

Anyway, you should double check whether this does have the desired effect and merge if your convinced or test it first on a benchmark cluster to verify the CPU savings.

sarum90:
I'll probably.... just remove all of the fake persistent args and get rid of the rest of this PR?

no sense doing it if it didn't work. (I assumed it did... kind of sad at treq for not actually making this work). I'll send this out for re-review once I've reduced this PR.

flocker/apiclient/test/test_client.py
sarum90: I _thought_ that a non-persistent pool was just like a pool with standing size of 0 that boosts to more on-demand ... effectively just a source of connections to a given server. I thought it was just misnamed pool because the implementation for persistent connections would obviously have to be a pool.

Obviously I was wrong, thanks for digging in! Removing all this nonsense as treq doesn't appear to offer a good way to do what we want to do.

wallrj:
Aha, I think this has the answer:
 * https://github.com/twisted/treq/issues/71#issuecomment-53835158

So the connection pool doesn't limit the total connections. 

It'd be interesting to see whether the combination of persistent=True and increasing the pool size (and a larger number of idle connections) would lower the number of new connections.

Also possibly limit the parallelism in the client library.

But pipelining (if it were supported) would be even more effective I think and allow the benchmarks to run with even SSL handshakes.

I've merged forward to force some new builds...after upgrading treq and twisted recently.

wallrj:
Looks good @sarum90 

So the upshot of this will be:
 * Places where we use treq directly will use persistent connections, by way of the global treq persistent connection pool,
 * ...except where we use `treq.get/post(...., persistent=False)`, as in the testtools, which start and then query a test HTTP server.
 * And the Flocker API client bypasses  the `treq.api`, instead instantiating a `treq.client.HTTPClient` and supplying it with a custom `Agent` *without* specifying an ``HTTPConnectionPool`` ...which will result in a new non-persistent connection pool for each request.

I guess I'm a bit confused why your original code (reverted in 64beddaf3d0514ff081387bb3de3c5b1d9bc13cd) didn't seem to work. 
Having dug into the t.w.Agent code, it looks correct.
I wonder if I was using it correctly?

Anyway. The cleanups here make sense, but we should probably update the PR title before merging and this PR shouldn't close FLOC-4401.
Please merge

wallrj:
Oh, it looks like you updated the issue title. So we *can* close it.

sarum90:
Thanks for the review. Merging.

PR #2788, opened 2016-05-10
[FLOC-4402] Update the benchmark.yml

sarum90:
Missing stanzas from benchmark.yml that are needed to run the 10 QPS benchmark scenarios.


wallrj:
LGTM.

PR #2790, opened 2016-05-11
[DO NOT MERGE] Add flocker volumes to cluster

sarum90:
Totally hacky implementation  of a command line utility that lets you list all volumes your backend can see (not just the flocker ones), and add some to a flocker cluster of your choice.

This definitely needs some work, but is useful to play around with the workflows.


wallnerryan:
thanks @sarum90 trying this out now. Has possible uses for certain customers as well. Will let you know how it goes.

sarum90:
cleaning up old PRs.

PR #2791, opened 2016-05-16
[FLOC-4405] Stop using  devpi.clusterhq.com

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4405

I decided to stop using devpi.clusterhq.com for the following reasons:
 * It seems to be incompatible with pip 8.1.2 and manylinux wheel files
 * It's a maintenance headache
 * It is not configured securely.

With manylinux wheel files now on PyPI, we may even speed up the builds by going direct to PyPI and hopefully the cloud providers will provide their own transparent Python package caches.

Let's see.

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -254,34 +254,9 @@ common_cli:
     # Report the version of Python we're using, to aid debugging.
     ${venv}/bin/python --version
 
-  setup_pip_cache: &setup_pip_cache |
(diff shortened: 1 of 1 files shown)


sarum90:
Thanks for fixing! 

Looks good to me. It'd be sad if this makes all of our builds slower, but not as sad as failing our GCE tests randomly.

PR #2793, opened 2016-05-19
[FLOC-4412] Upgrade to Hypothesis 3 

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4412

Broken out from https://github.com/ClusterHQ/flocker/pull/2792

### DIFF
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -149,7 +149,7 @@ DISCOVERED_DATASET_STRATEGY = tagged_union_strategy(
     DiscoveredDataset,
     {
         'dataset_id': uuids(),
-        'maximum_size': integers(min_value=1),
+        'maximum_size': integers(min_value=512),
         'mount_point': builds(FilePath, sampled_from([
             '/flocker/abc', '/flocker/xyz',
         ])),
@@ -1404,6 +1404,9 @@ class BlockDeviceCalculatorTests(TestCase):
         """
         Cleanup after running a hypothesis example.
         """
+        deployer = getattr(self, 'deployer', None)
+        if deployer is None:
+            return
         umount_all(self.deployer.mountroot)
         detach_destroy_volumes(self.deployer.block_device_api)
 
--- a/flocker/node/agents/test/test_blockdevice.py
+++ b/flocker/node/agents/test/test_blockdevice.py
@@ -1517,7 +1517,7 @@ class BlockDeviceCalculatorTests(TestCase):
                       e.iteration_count)
 
 
-class TranistionTests(TestCase):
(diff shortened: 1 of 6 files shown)


flocker/node/agents/test/test_blockdevice.py
```diff
@@ -169,12 +169,10 @@
 
 DESIRED_DATASET_ATTRIBUTE_STRATEGIES = {
     'dataset_id': uuids(),
-    'maximum_size': integers(min_value=0).map(
-        lambda n: (
-            LOOPBACK_MINIMUM_ALLOCATABLE_SIZE +
-            n * LOOPBACK_ALLOCATION_UNIT
-        )
-    ),
+    'maximum_size': integers(
```
sarum90: ❓ I think the other formation is (a  lot) better for hypothesis:

```
integers(min_value=0, max_value = 10).map(lambda n: LOOPBACK_MINIMUM_ALLOCATABLE_SIZE+n*LOOPBACK_ALLOCATION_UNIT)
```

Otherwise I think hypothesis is like "try 512... okay now try 576... okay now try 513... okay now try 514" which is a whole bunch of wasted examples (as they would all be run as 512) (and real test time) that cheapens hypothesis's ability to fuzz the relevant part of the test input search space.

Apologies I don't have time to hack together a minimal example to show you right now... but I'm happy to make one if you need assistance.

sarum90:
Good stuff, Thanks!

Potentially fix the one strategy... but it's not something I care too deeply about.

I might make a gist later to show why it's better to do the other way.

flocker/node/agents/test/test_blockdevice.py
wallrj: Yeah, that does make sense.
I'm also inclined to make this a fixed value `just(LOOPBACK_MINIMUM_ALLOCATABLE_SIZE)` 
....otherwise it seems like this is an in appropriate place to test for the ability of the loopback driver to handle a range of device sizes.
It might make more sense if we supported resizing or if the tests weren't actually calling the backend.

sarum90: I agree.

Our general philosophy with hypothesis in CHQ seems to be "assume you are forgetting about some edge case, be very liberal with your strategies, and let hypothesis show you".

I think I agree with that strategy _if_ the tests you are running are actually unit tests (can run in under 1ms). In this particular case, loop devices take a lot longer to interact with than 1ms, so I think we should trim the search space for hypothesis. Thus, I think you are correct, we should use a `just(LOOPBACK_MINIMUM_ALLOCATABLE_SIZE)` for this. Especially considering loop devices are only intended for use in test code. exercising weird corner cases doesn't get us much.

PR #2794, opened 2016-05-19
[FLOC-4413] Fix golden files

sarum90:
#I've done 2 things in this PR.

1) I've added a regression test to verify that HEAD can always generate the latest golden file. This should fix the problem I encountered where I could no longer generate the golden file from master, and was confused as to why.

2) I've added additional infrastructure to enable us to generate multiple golden files per config version. This is important because version 4 added some complexity to the config file that should be tested. However, this cannot be added to the version 4 golden file because the version 4 golden file must be exactly the upgraded version 3 golden file. This complexity does not exist in versions 1 through 3, so it only needs to be tested for versions 4+. Thus when we add version 5 we need to test that it can be upgraded to with 2 different versions:

- One that is valid to upgrade all the way from version 1 to version 5 (and 2 to 5 and 3 to 5)
- One that tests the new complexity added in version 4, and verifies that we can upgrade from that configuration to version 5.

The proposed solution is to enable additional golden files for all configurations 4+, where a new golden file can be added every time there is additional complexity in the configuration/state.


flocker/control/test/test_persistence.py
```diff
@@ -761,11 +768,51 @@ def test_upgrade_configuration_versions(self, versions):
...
+            path = configs_dir.child(
+                b"configuration_%d_v%d.json" % (i, _CONFIG_VERSION)
+            )
+            self.assertEqual(
+                encoding, path.getContent(),
+                "Golden test file %s can not be generated from HEAD. Please "
+                "review the python files in that directory to re-generate "
+                "that file if you have intentionally changed the backing test "
+                "data. You might need to rev the model if you are "
+                "intentionally changing the model."
```
sarum90: incomplete format string.

flocker/control/test/test_persistence.py
```diff
@@ -38,17 +38,17 @@
     Port, Link, Leases, Lease, BlockDeviceOwnership, PersistentState,
     )
 
-# The UUID values for the Dataset and Node in TEST_DEPLOYMENT match
-# those in the versioned JSON configuration files used by tests in this
-# module. If these values are changed, you will also need to regenerate
-# the test JSON files using the scripts provided in the
-# flocker/control/test/configurations/ directory, using the correct
-# commit checkout to generate JSON appropriate to each config version.
+# The UUID values for the Dataset and Node in the follow TEST_DEPLOYMENTs match
```
wallrj: typo "follow"

sarum90: fixed.

flocker/control/test/configurations/configuration_v4.py
```diff
@@ -1,11 +1,22 @@
...
 
 from flocker.control._model import Configuration
 from flocker.control._persistence import wire_encode
-from flocker.control.test.test_persistence import TEST_DEPLOYMENT
+from flocker.control.test.test_persistence import TEST_DEPLOYMENTS
+
+_VERSION = 4
 
 if __name__ == "__main__":
-    print wire_encode(Configuration(version=4, deployment=TEST_DEPLOYMENT))
```
wallrj: :question: So the script now writes the configuration files directly. I wondered if this process might be described in our docs and need updating, but only found the instructions in the test_persistence docstrings. That's fine.

wallrj: And I suppose we should never need to re-run the v1, v2 and v3 scripts again....at least not from master...only from the revisions that they were written for.

flocker/control/test/configurations/configuration_v4.py
```diff
@@ -1,11 +1,22 @@
...
+    myfile = FilePath(__file__)
+    for i, deployment in enumerate(TEST_DEPLOYMENTS, start=1):
```
wallrj: :question: `enumerate(start=...` Cool. Didn't know about that.

flocker/control/test/test_persistence.py
```diff
@@ -38,17 +38,17 @@
...
-# module. If these values are changed, you will also need to regenerate
-# the test JSON files using the scripts provided in the
-# flocker/control/test/configurations/ directory, using the correct
-# commit checkout to generate JSON appropriate to each config version.
+# The UUID values for the Dataset and Node in the following TEST_DEPLOYMENTs
+# match those in the versioned JSON configuration files used by tests in this
+# module.  If these values are changed, you will also need to regenerate the
+# test JSON files using the scripts provided in the
+# flocker/control/test/configurations/ directory, using the correct commit
+# checkout to generate JSON appropriate to each config version.
```
wallrj: :question: Yeah, so the docs say nothing about whether the script prints or writes the output to a file.

flocker/control/test/test_persistence.py
```diff
@@ -38,17 +38,17 @@
...
 DATASET = Dataset(dataset_id=u'4e7e3241-0ec3-4df6-9e7c-3f7e75e08855',
                   metadata={u"name": u"myapp"})
 NODE_UUID = UUID(u'ab294ce4-a6c3-40cb-a0a2-484a1f09521c')
 MANIFESTATION = Manifestation(dataset=DATASET, primary=True)
-TEST_DEPLOYMENT = Deployment(
```
wallrj: :question: The v1, v2, and v4 scripts won't care about this renaming because they're only ever run from old revisions.

flocker/control/test/test_persistence.py
```diff
@@ -761,11 +768,51 @@ def test_upgrade_configuration_versions(self, versions):
...
+                b"configuration_%d_v%d.json" % (i, _CONFIG_VERSION)
+            )
+            self.assertEqual(
+                encoding, path.getContent(),
+                "Golden test file %s can not be generated from HEAD. Please "
+                "review the python files in that directory to re-generate "
+                "that file if you have intentionally changed the backing test "
+                "data. You might need to rev the model if you are "
+                "intentionally changing the model." % path.path
+            )
```
wallrj: Looks good. I tried running this test against master, but it failed for not finding the new second configuration file.

```
(4413-master) (master ✓? =)[~/.../HybridLogic/flocker]$ trial flocker.control.test.test_persistence.LatestGoldenFilesValid
flocker.control.test
  test_persistence
    LatestGoldenFilesValid ...                                          [ERROR]

[ERROR]
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/trial/runner.py", line 605, in loadByNames
    things.append(self.findByName(name))
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/trial/runner.py", line 409, in findByName
    return reflect.namedAny(name)
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/python/reflect.py", line 303, in namedAny
    topLevelPackage = _importAndCheckStack(trialname)
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/python/reflect.py", line 242, in _importAndCheckStack
    return __import__(importName)
  File "/home/richard/projects/HybridLogic/flocker/flocker/control/test/test_persistence.py", line 79, in <module>
    'configurations').child(b"configuration_1_v1.json").getContent()
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/python/filepath.py", line 294, in getContent
    fp = self.open()
  File "/home/richard/.virtualenvs/4413-master/lib/python2.7/site-packages/twisted/python/filepath.py", line 1010, in open
    return open(self.path, mode + 'b')
exceptions.IOError: [Errno 2] No such file or directory: '/home/richard/projects/HybridLogic/flocker/flocker/control/test/configurations/configuration_1_v1.json'

flocker.control.test.test_persistence.LatestGoldenFilesValid
Ran 1 tests in 0.001s

FAILED (errors=1)

```

flocker/control/test/test_persistence.py
```diff
@@ -761,11 +768,51 @@ def test_upgrade_configuration_versions(self, versions):
...
+            path = configs_dir.child(
+                b"configuration_%d_v%d.json" % (i, _CONFIG_VERSION)
+            )
+            self.assertEqual(
+                encoding, path.getContent(),
+                "Golden test file %s can not be generated from HEAD. Please "
+                "review the python files in that directory to re-generate "
+                "that file if you have intentionally changed the backing test "
+                "data. You might need to rev the model if you are "
+                "intentionally changing the model." % path.path
```
wallrj: typo "rev the model" ??

wallrj: We normally write `% (path.path,)` to avoid any confusion  while unpacking.

wallrj:
Thanks @sarum90 

I must say I've struggled to understand this code and therefore the changes to it.

Here's my understanding:

v4 test data is different than we have in the sample configuration file.
It includes a new  `blockdevice_ownership` section.

```
{
  "deployment": {
    "persistent_state": {
      "blockdevice_ownership": {
        "values": {
          "0": [
            [
              {
                "hex": "b229d949-0856-4011-96e5-3dd0a5586180",
                "$__class__$": "UUID"
              },
              "block-device-id"
            ]
          ],
          "_t": "a"
        }
      }
    }
  }
}
```
https://benjamine.github.io/jsondiffpatch/demo/

* Note: It'd be nice if the configurations were stored in pretty printed form to make it easier to review the changes.     

* The existing tests demonstrate that the sample configuration can be upgraded between all configuration revisions.

* But the latest sample  doesn't contain an example of the `blockdevice_ownership` attribute and therefore there's nothing to test that the upgrade code does the correct thing in the situation where the blockdevice_ownership should be populated during upgrade...is that it?

Thanks for writing a comprehensive PR description, but perhaps you could quickly explain the failure circumstances to me before you merge this.

But yes, I think this is good to merge.

flocker/control/test/test_persistence.py
sarum90: Yup. Tried to express intent with the new names.

flocker/control/test/test_persistence.py
sarum90: Oh yea I probably should have made an intermediate version that had this test but didn't expect the new names of the files. Apologies for that.

I verified this by changing the new files (Just add a random character anywhere, but maintain valid JSON) and seeing it fail.

flocker/control/test/test_persistence.py
sarum90: yea I mean "rev" as in "increase the version and do whatever special version bump dance you have to do".

In this case, increase the version and write an upgrade script. Reworded to explicitly say that rather than use the word `rev`.

sarum90:
Updated the PR description. Hopefully everything is more clear now.

PR #2799, opened 2016-05-24
[FLOC-4417] Applications should be a pmap in the data model.

sarum90:
Note that this is merging against a staging branch. I have further changes that I want to make to version 5 of the model before merging it into master, so I will be building them up in the FLOC-4416 branch and then submitting that separately. That way "Version 5 in master" will never mean 2 different things.


flocker/control/test/configurations/configuration_v5.py
```diff
@@ -0,0 +1,22 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+# Generate a v5 configuration.
+# Hash to recreate: XXX: Put the hash of the commit here.
```
sarum90: Going to wait until the PR to master to fill this in.

flocker/control/_config.py
```diff
@@ -1296,7 +1296,7 @@ def deployment_from_configuration(deployment_state, deployment_configuration,
             raise ConfigurationError("No known node with address {}.".format(
                 hostname))
         node = Node(uuid=node_states[hostname].uuid,
-                    applications=frozenset(node_applications),
+                    applications={app.name: app for app in node_applications},
```
wallrj: I'd forgotten about all this code...for parsing YAML configuration files supplied to the _compose endpoint by `flocker-deploy`
We should remove all that code, right?
Not in this branch, but it would be good housekeeping.

flocker/control/_model.py
```diff
@@ -110,23 +111,33 @@ def pmap_field(
     :param initial: An initial value for the field.  This will first be coerced
         using the field's factory.  If not given, the initial value is an empty
         map.
+    :param factory: A factory used to convert input arguments to the stored
+        value whenever it is set. Note that this will be composed with the
+        constructor for the ``CheckedPMap`` class constructed for this field.
```
wallrj: Should this work?:

```
class Foo(PClass):
    bar = pmap_field(key_type=bytes, value_type=int, factory=int)


In [58]: Foo().bar.set('moo', .1)
CheckedValueTypeError                     Traceback (most recent call last)
<ipython-input-58-5a65a2a9ff99> in <module>()
----> 1 Foo().bar.set('moo', .1)

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_pmap.pyc in set(self, key, val)
    157         pmap({'a': 1, 'c': 4, 'b': 2})
    158         """
--> 159         return self.evolver().set(key, val).persistent()
    160 
    161     def remove(self, key):

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_checked_types.pyc in set(self, key, value)
    458         def set(self, key, value):
    459             _check_types([key], self._destination_class._checked_key_types, self._destination_class, CheckedKeyTypeError)
--> 460             _check_types([value], self._destination_class._checked_value_types, self._destination_class)
    461             self._invariant_errors.extend(data for valid, data in (invariant(key, value)
    462                                                                    for invariant in self._destination_class._checked_invariants)

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_checked_types.pyc in _check_types(it, expected_types, source_class, exception_type)
    171                     expected_types=tuple(get_type(et).__name__ for et in expected_types),
    172                     actual_type=actual_type.__name__)
--> 173                 raise exception_type(source_class, expected_types, actual_type, e, msg)
    174 
    175 

CheckedValueTypeError: Type StrIntPMap can only be used with ('int',), not float

In [59]: Foo().bar.set('moo', 1)
Out[59]: StrIntPMap({'moo': 1})

```

I expected the `int` factory to coerce the float to int before it gets validated and assigned.

wallrj: Ah, no. I get it. The factory is used when assigning a new value for the pmap_field attribute, not an element in the pmap.

flocker/control/_model.py
```diff
@@ -414,6 +425,26 @@ def key_match_invariant(pmap):
...
+        resulting Mapping.
+    :param obj: Either something that is a Mapping (which is immediately
+        returned), or an iterable to be converted into a Mapping.
+
+    :returns: A Mapping of the items in ``obj`` from
+        ``getattr(o, attribute)`` to the items in obj (``o`` in ``obj``).
+    """
+    if isinstance(obj, Mapping):
+        return obj
+    return {getattr(a, attribute): a for a in obj}
```
wallrj: I guess it doesn't matter too much, but this gives a non-obvious error if you accidentally assign e.g. a list or set


```
In [81]: Foo().transform(['bar'], {"abc", 123})
AttributeError                            Traceback (most recent call last)
<ipython-input-81-90b6a18eed00> in <module>()
----> 1 Foo().transform(['bar'], {"abc", 123})

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_pclass.pyc in transform(self, *transformations)
    133         for other persistent types work as expected.
    134         """
--> 135         return transform(self, transformations)
    136 
    137     def __eq__(self, other):

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_transformations.pyc in transform(structure, transformations)
     41     r = structure
     42     for path, command in _chunks(transformations, 2):
---> 43         r = _do_to_path(r, path, command)
     44     return r
     45 

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_transformations.pyc in _do_to_path(structure, path, command)
     50 
     51     kvs = _get_keys_and_values(structure, path[0])
---> 52     return _update_structure(structure, kvs, path[1:], command)
     53 
     54 

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_transformations.pyc in _update_structure(structure, kvs, path, command)
     89             if result is not v:
     90                 e[k] = result
---> 91     return e.persistent()

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_pclass.pyc in persistent(self)
    229     def persistent(self):
    230         if self._pclass_evolver_data_is_dirty:
--> 231             return self._pclass_evolver_original.__class__(**self._pclass_evolver_data)
    232 
    233         return self._pclass_evolver_original

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_pclass.pyc in __new__(cls, **kwargs)
     50         for name, field in cls._pclass_fields.items():
     51             if name in kwargs:
---> 52                 value = field.factory(kwargs[name])
     53                 _check_and_set_attr(cls, field, name, value, result, invariant_errors)
     54                 del kwargs[name]

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.py in <lambda>(x)
    136 
    137     if input_factory:
--> 138         factory = lambda x: mapping_factory(input_factory(x))
    139     else:
    140         factory = mapping_factory

<ipython-input-69-59b958d20f7e> in <lambda>(x)
      1 class Foo(PClass):
----> 2     bar = pmap_field(key_type=bytes, value_type=int, factory=lambda x: _turn_iterable_to_mapping_from_attribute('name', x))
      3 

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.py in _turn_iterable_to_mapping_from_attribute(attribute, obj)
    443     if isinstance(obj, Mapping):
    444         return obj
--> 445     return {getattr(a, attribute): a for a in obj}
    446 
    447 

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.py in <dictcomp>((a,))
    443     if isinstance(obj, Mapping):
    444         return obj
--> 445     return {getattr(a, attribute): a for a in obj}
    446 
    447 

AttributeError: 'int' object has no attribute 'name'

```


Maybe the function could  do a little extra validation of the supplied objects and raise a meaningful error message.

flocker/control/_model.py
```diff
@@ -451,7 +482,10 @@ def __new__(cls, hostname=None, **kwargs):
         return PClass.__new__(cls, **kwargs)
 
     uuid = field(type=UUID, mandatory=True)
-    applications = pset_field(Application)
+    applications = pmap_field(
+        unicode, Application, invariant=_keys_match("name"),
+        factory=lambda x: _turn_iterable_to_mapping_from_attribute('name', x)
```
wallrj: Just remind me of the motivation for this. 
Are just trying to avoid changing existing code that supplies Application sets rather than mappings.
It seems confusing to be able to supply a list of applications here:

```
In [112]: Node(uuid=uuid4(), applications=[Application(name=u"database1", image=DockerImage(repository="clusterhq/foodb", tag="latest"))])
Out[112]: Node(applications=UnicodeApplicationPMap({u'database1': Application(memory_limit=None, name=u'database1', links=LinkPSet([]), environment=pmap({}), command_line=None, image=DockerImage(tag='latest', repository='clusterhq/foodb'), restart_policy=RestartNever(), volume=None, running=True, ports=PortPSet([]), cpu_shares=None)}), manifestations=UnicodeManifestationPMap({}), uuid=UUID('4701d23b-de5e-4e85-9e57-431345cecd29'))

In [113]: Node(uuid=uuid4(), applications=[Application(name="database1", image=DockerImage(repository="clusterhq/foodb", tag="latest"))])
CheckedKeyTypeError                       Traceback (most recent call last)
<ipython-input-113-1f36d68ca295> in <module>()
----> 1 Node(uuid=uuid4(), applications=[Application(name="database1", image=DockerImage(repository="clusterhq/foodb", tag="latest"))])

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.py in __new__(cls, hostname, **kwargs)
    480                  "a bug.", DeprecationWarning, stacklevel=2)
    481             kwargs["uuid"] = ip_to_uuid(hostname)
--> 482         return PClass.__new__(cls, **kwargs)
    483 
    484     uuid = field(type=UUID, mandatory=True)

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_pclass.pyc in __new__(cls, **kwargs)
     50         for name, field in cls._pclass_fields.items():
     51             if name in kwargs:
---> 52                 value = field.factory(kwargs[name])
     53                 _check_and_set_attr(cls, field, name, value, result, invariant_errors)
     54                 del kwargs[name]

/home/richard/projects/HybridLogic/flocker/flocker/control/_model.py in <lambda>(x)
    136 
    137     if input_factory:
--> 138         factory = lambda x: mapping_factory(input_factory(x))
    139     else:
    140         factory = mapping_factory

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_checked_types.pyc in __new__(cls, initial, size)
    409         evolver = CheckedPMap.Evolver(cls, pmap())
    410         for k, v in initial.items():
--> 411             evolver.set(k, v)
    412 
    413         return evolver.persistent()

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_checked_types.pyc in set(self, key, value)
    457 
    458         def set(self, key, value):
--> 459             _check_types([key], self._destination_class._checked_key_types, self._destination_class, CheckedKeyTypeError)
    460             _check_types([value], self._destination_class._checked_value_types, self._destination_class)
    461             self._invariant_errors.extend(data for valid, data in (invariant(key, value)

/home/richard/.virtualenvs/4417/lib/python2.7/site-packages/pyrsistent/_checked_types.pyc in _check_types(it, expected_types, source_class, exception_type)
    171                     expected_types=tuple(get_type(et).__name__ for et in expected_types),
    172                     actual_type=actual_type.__name__)
--> 173                 raise exception_type(source_class, expected_types, actual_type, e, msg)
    174 
    175 

CheckedKeyTypeError: Type UnicodeApplicationPMap can only be used with ('unicode',), not str


```

Also note that Application seems to allow a name to be str where the pmap_field only accepts unicode.

flocker/control/_model.py
```diff
@@ -723,35 +757,35 @@ def move_application(self, application, target_node):
         """
         deployment = self
         for node in deployment.nodes:
-            for container in node.applications:
```
wallrj: This is the (or part of the) crux of this branch, right?

We no longer have to loop through the entire set. Just lookup the application by name.

Cool.

flocker/control/_model.py
```diff
@@ -723,35 +757,35 @@ def move_application(self, application, target_node):
...
-                            dataset_id = application.volume.dataset.dataset_id
-                            node = node.transform(
-                                ("manifestations", dataset_id), discard
-                            )
-                        # Before updating the deployment instance.
-                        deployment = deployment.update_node(node)
-                        deployment = deployment.update_node(target_node)
+                            ("manifestations", dataset_id), discard
+                        )
+                    # Finally we can now remove the manifestation from the
```
wallrj: This comment is incomplete.

flocker/control/_model.py
```diff
@@ -983,7 +1017,16 @@ def __new__(cls, **kwargs):
...
     uuid = field(type=UUID, mandatory=True)
     hostname = field(type=unicode, factory=unicode, mandatory=True)
-    applications = pset_field(Application, optional=True, initial=None)
+    applications = pmap_field(
+        unicode, Application, optional=True, initial=None,
+        invariant=_keys_match("name"),
+        factory=(
+            lambda x: (
+                x if x is None
+                else _turn_iterable_to_mapping_from_attribute('name', x)
```
wallrj: Can't the check for `None` be moved into `turn_iterable_to_mapping_from_attribute`?

flocker/control/_persistence.py
```diff
@@ -164,6 +164,32 @@ def upgrade_from_v3(cls, config):
...
+            nodes = decoded_config[u"deployment"][u"nodes"]
+        except KeyError:
+            pass
+        else:
+            new_nodes = []
+            for n in nodes:
+                new_node = n
+                new_node[u"applications"] = {
+                    u"values": [(a[u"name"], a) for a in n[u"applications"]],
+                    _CLASS_MARKER: "PMap"
```
wallrj: Why isn't `values` a dictionary? `dict((a.name, a) for a in n[u"applications"])`

flocker/control/httpapi.py
```diff
@@ -152,7 +152,10 @@ def _extract_containers_state(deployment_state):
     """
     result = []
     for node in deployment_state.nodes:
-        for application in (node.applications or ()):
+        applications = ()
+        if node.applications is not None:
+            applications = node.applications.values()
```
wallrj: Much clearer.

flocker/control/httpapi.py
```diff
@@ -698,9 +701,8 @@ def create_container_configuration(
         # Check if container by this name already exists, if it does
         # return error.
         for node in deployment.nodes:
-            for application in node.applications:
-                if application.name == name:
-                    raise CONTAINER_NAME_COLLISION
+            if name in node.applications:
+                raise CONTAINER_NAME_COLLISION
```
wallrj: Much simpler.

flocker/control/test/configurations/configuration_1_v5.json
```diff
@@ -0,0 +1 @@
...ataset_id": "4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}, "$__class__$": "AttachedVolume"}, "memory_limit": null, "name": "myapp", "links": [], "$__class__$": "Application", "restart_policy": {"$__class__$": "RestartNever"}, "image": {"tag": "7.6", "repository": "postgresql", "$__class__$": "DockerImage"}, "command_line": null, "environment": {"values": [], "$__class__$": "PMap"}, "running": true, "ports": [], "cpu_shares": null}]], "$__class__$": "PMap"}, "manifestations": {"values": [["4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", {"dataset": {"deleted": false, "dataset_id": "4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}]], "$__class__$": "PMap"}, "uuid": {"hex": "ab294ce4-a6c3-40cb-a0a2-484a1f09521c", "$__class__$": "UUID"}, "$__class__$": "Node"}], "leases": {"values": [], "$__class__$": "PMap"}, "$__class__$": "Deployment"}, "version": 5, "$__class__$": "Configuration"}
```
wallrj: missing newline.

flocker/control/test/configurations/configuration_2_v5.json
```diff
@@ -0,0 +1 @@
...ataset_id": "4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}, "$__class__$": "AttachedVolume"}, "memory_limit": null, "name": "myapp", "links": [], "$__class__$": "Application", "restart_policy": {"$__class__$": "RestartNever"}, "image": {"tag": "7.6", "repository": "postgresql", "$__class__$": "DockerImage"}, "command_line": null, "environment": {"values": [], "$__class__$": "PMap"}, "running": true, "ports": [], "cpu_shares": null}]], "$__class__$": "PMap"}, "manifestations": {"values": [["4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", {"dataset": {"deleted": false, "dataset_id": "4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}]], "$__class__$": "PMap"}, "uuid": {"hex": "ab294ce4-a6c3-40cb-a0a2-484a1f09521c", "$__class__$": "UUID"}, "$__class__$": "Node"}], "leases": {"values": [], "$__class__$": "PMap"}, "$__class__$": "Deployment"}, "version": 5, "$__class__$": "Configuration"}
```
wallrj: and here

flocker/control/test/configurations/configuration_v5.py
wallrj: Ok.

wallrj:
Thanks @sarum90 

Looks good and I  think I get it.

Sorry it took so long to review.

Please answer or address the comments then merge.

flocker/control/_model.py
sarum90: Yea +1, just like factory for the normal `field`. Maybe not the best... but fit the existing style.

flocker/control/_config.py
sarum90: Yea I think we should probably remove it. I think it just gets obviously nuked in a move to something like Flocker-Lite.

flocker/control/_model.py
sarum90: Hrm... This factory intentionally enables you to pass an iterable of things and turns it into a mapping from an attribute of the objects to the actual objects.

In the use case here, that enables us to hand an iterable of applications, and this function turns it into a mapping from application.name to applications.

If you hand it in an iterable of object that don't have a 'name' attribute... then you should get an attribute error describing that the name attribute is missing... So the AttributeError is definitely the thing that should be raised, and I don't have a mechanism to add context to the exception (... in an iterable handed to the factory of a PClass field).

Expanded the docstring to leave a breadcrumb for people who might encounter this Exception.

flocker/control/_model.py
sarum90: Motivation of turning into a pmap rather than pset: Enforces the `unique primary key`-ness of the field in the data model. Also conceptually I want to be able to do things like "add a dataset to the node with UUID foo" rather than "remove the node with UUID foo from this set, and insert a new object into the set with an additional dataset".  Encoding it this way makes it _much_ easier to write a general-purpose differ.

Motivation for the factory: Now we have duplicate data. You need to specify the name of the application as application.name _and_ as the key to the object in the mapping. Rather, when you are initially constructing a bunch of applications, just specify application.name, and this factory will turn it into a mapping. Conveniently this also saved me some trouble migrating all of the tests to the new format.

Added a constraint to Application that the type of name should be a unicode. Really it was totally unconstrained in the past.

flocker/control/_model.py
sarum90: The crux is more about being able to write a general-purpose differ and have it do  the "right thing" with our deeply pyrsistent configuration and state.

However, as a side effect there are lots of places where it was clear that what we wanted was a pmap for O(1) lookup, but instead did O(n) lookup because we had a PSet. These are just bonus points :).

flocker/control/_model.py
sarum90: De-spliced it, thanks!

flocker/control/_model.py
sarum90: Sure, this is needed because here the field is optional, whereas in the other location it is not optional.  But... that is more clearly communicated as a parameter to the factory than a `x if x is None` here.

flocker/control/_persistence.py
sarum90: Agreed that it should be, unfortunately that's not the way our wire_encoding works :(

https://github.com/ClusterHQ/flocker/blob/d0f65ca210a23cc3152c5ca68c1c35ae1c1d8659/flocker/control/_persistence.py#L191

(Also see other upgraders in this file)

Bonus points: Because this is in the mechanism that we use to load configs generated from previous versions it would be quite a bit of work to change this now :(.

flocker/control/httpapi.py
sarum90: Yea, turning this into: `((node.applications or {}).values())` just seemed too mean.

flocker/control/test/configurations/configuration_1_v5.json
sarum90: Acknowledged.

This is generated and read by a script though:

`flocker/control/test/configurations/configuration_v5.py`

So I guess not tempted to add it manually right now? Not trying to be difficult, just trying to both:

A) have the code actually and truly generation these files
B) Not have random code that just adds a newline added to the generation script (And I think it would also have to be added to the verification test that I added earlier).

flocker/control/test/configurations/configuration_2_v5.json
sarum90: ditto above.

sarum90:
Even though just merging into a staging branch... I'm going to let the tests run out, because the additional type constraint on applications might yield some interesting failures.

sarum90:
Errors appear... unrelated to the application stuff.

Good enough for me to merge into the staging branch. I'll re-run the full gamut of tests on that branch before merging to master.

PR #2800, opened 2016-05-24
[FLOC-4418] Nodes should be a pmap.

sarum90:
Built on FLOC-4417, this PR changes the data model from a set of Nodes to a mapping from node UUID to nodes.


sarum90:
If you review this before #2799 is merged, please instead review on the following PR: https://github.com/ClusterHQ/flocker/pull/2801

It should reflect only the differences that this is adding on top of #2799

wallrj:
Thanks @sarum90 

I reviewed the changes in https://github.com/ClusterHQ/flocker/pull/2801

Everything looks fine, but I'm starting to have some doubts about switching to node / application maps vs sets. 
Is it as blindingly obvious as we first thought?

 * Sets ensure that an object appears only once in the collection while with maps (right now), it's up to the caller to ensure that the object matches the key in the map....
   * ah, now I go read `_keys_match` invariant function so forget that.

 * And perhaps the map is even better, because it prevents accidentally adding e.g. Nodes with duplicate UUIDs but differing in other attributes..which would have been possible with sets.

 * But I can't help feeling that a `set` of Nodes is more natural and that we should (in theory) be able to add a `KeyedPSet` type which automatically maps the inserted objects by one or more attributes behind the scenes. 

Just a thought.

Please merge and proceed unless you agree strongly with my comments.

sarum90:
With the `_keys_match` invariant and the factory method... I think this now is effectively a `KeyedPSet`?

I'll be updating this branch to use the new optional kwarg

PR #2801, opened 2016-05-24
DO NOT MERGE (For review of https://github.com/ClusterHQ/flocker/pull/2800 only)

sarum90:
This is the logical review of https://github.com/ClusterHQ/flocker/pull/2800, which is built on https://github.com/ClusterHQ/flocker/pull/2799


sarum90:
Wasn't used in review.

PR #2804, opened 2016-05-31
[FLOC-4416] Use maps for maps

sarum90:
This combines FLOC-4417 and FLOC-4418 into a single PR to atomically rev the version of the configuration.

As all of this code has already been reviewed in those two branches, I intend to run this through all of the current tests, and then if green, just merge this to master as-is.

The only modification since the last review was filling in a hash to flocker/control/test/configurations/configuration_v5.py.


sarum90:
The previous test run should be valid still. So I'm going to merge as is.

PR #2805, opened 2016-06-01
[FLOC-4425] Upgrade virtualenv package requirement

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4425

Upgraded to virtualenv-15.0.2
 * https://virtualenv.pypa.io/en/latest/changes/#id1

Which has the latest version of pkg_resources (vendored into setuptools) which is able to parse the so called environment markers in package version strings e.g.

``nomenclature >= "0.1.0"; sys_platform == "linux2"``

### DIFF
--- a/flocker/restapi/docs/publicapi.py
+++ b/flocker/restapi/docs/publicapi.py
@@ -12,7 +12,7 @@ from yaml import safe_load
 from docutils import nodes
 
 from sphinxcontrib import httpdomain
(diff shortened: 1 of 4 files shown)


sarum90:
Nice catch! Looks like the fix worked.

Bonus points to make  jenkins-_run_sphinx_link_check  pass as well... but that looks like an unrelated regression in the links in our docs.

Thanks for digging into this!

sarum90:
Please merge whenever.

PR #2806, opened 2016-06-02
[FLOC-4421] Diffing of objects

sarum90:
In order to be able to just send Configuration and State diffs to all of the nodes, Flocker needs to have a mechanism to compute the diff between two different Configurations or two different States.

This Adds a generic "Diffs between pyrsistent objects" module, and adds a bunch of hypothesis strategies to generate deployments in order to assist in testing the diffing algorithm.


flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+# -*- test-case-name: flocker.control.test.test_diffing -*-
+
```
wallrj: Add a module docstring.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+class _DiffChange(Interface):
+    """
+    Interface for a diff change.
+
+    This is simply something that can be applied to an object to create a new
+    object.
+
+    This interface is created more to simply documentation than for any of the
+    actual zope.interface mechanisms.
+    """
```
wallrj: Typo "more to simply documentation"
And maybe drop the word "simply" from the line above while you're here.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Apply this diff change to the passed in object and return a new object
+        that is obj with the ``self`` diff applied.
+
+        :param object obj: The object to apply the diff to.
+
+        :returns: A new object that is the passed in object with the diff
+            applied.
+        """
+        pass
```
wallrj: No need for `pass` if there's a docstring.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+    ``PMap`` inside a nested object tree.
+
+    :ivar path: The path in the nested object tree of the object to be removed
+        from the import set.
+    """
+
+    path = field(
+        type=PVector,
+        factory=freeze
+    )
```
wallrj: Would https://pyrsistent.readthedocs.io/en/latest/api.html#pyrsistent.pvector_field work here?

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+
+    changes = field(
+        type=PVector,
+        factory=freeze
+    )
+
+    def apply(self, obj):
+        for c in self.changes:
+            obj = c.apply(obj)
+        return obj
```
wallrj: :question: This looks like a situation where it'd be faster to operate on ``obj.evolver()`` rather than the immutable object.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+
+def _create_diffs_for_sets(current_path, set_a, set_b):
+    """
+    Computes a series of ``_DiffChange`` s to turn ``set_a`` into ``set_b``
+    assuming that these sets are at ``current_path`` inside a nested pyrsistent
+    object.
+
+    :param current_path: And iterable of pyrsistent object describing the path
+        inside the root pyrsistent object where the other arguments are
+        located.  See ``PMap.transform`` for the format of this sort of path.
```
wallrj: typo "And iterable of pyrsistent object"

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+    PSet,
+    field,
+    freeze,
+    pvector,
+)
+
+from zope.interface import Interface, implementer
+
+
+class _DiffChange(Interface):
```
wallrj: The naming convention for interfaces is `IDiffChange`

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+    :param set_b: The desired output set.
+
+    :returns: An iterable of ``_DiffChange`` s that will turn ``set_a`` into
+        ``set_b``.
+    """
+    resulting_diffs = pvector([]).evolver()
+    for item in set_a.difference(set_b):
+        resulting_diffs.append(
+            _Remove(path=current_path.append(item))
+        )
```
wallrj: :question: Might be neat if `_Remove` and `_Add` had the same signature. 
`_Remove(path=current_path, item=item)` should work I think.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+
+def _create_diffs_for_mappings(current_path, mapping_a, mapping_b):
+    """
+    Computes a series of ``_DiffChange`` s to turn ``mapping_a`` into
+    ``mapping_b`` assuming that these mappings are at ``current_path`` inside a
+    nested pyrsistent object.
+
+    :param current_path: And iterable of pyrsistent object describing the path
+        inside the root pyrsistent object where the other arguments are
+        located.  See ``PMap.transform`` for the format of this sort of path.
```
wallrj: same typo here.

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+    :param mapping_a: The desired input mapping.
+
+    :param mapping_b: The desired output mapping.
+
+    :returns: An iterable of ``_DiffChange`` s that will turn ``mapping_a``
+        into ``mapping_b``.
+    """
+    resulting_diffs = pvector([]).evolver()
+    a_keys = frozenset(x for x in mapping_a.iterkeys())
+    b_keys = frozenset(x for x in mapping_b.iterkeys())
```
wallrj: :question: Why not `frozenset(mapping_a.keys())` ?

flocker/control/_diffing.py
```diff
@@ -0,0 +1,247 @@
...
+        a_dict = subobj_a._to_dict()
+        b_dict = subobj_b._to_dict()
+        return _create_diffs_for_mappings(current_path, a_dict, b_dict)
+    elif isinstance(subobj_a, PMap) and isinstance(subobj_b, PMap):
+        return _create_diffs_for_mappings(
+            current_path, subobj_a, subobj_b)
+    elif isinstance(subobj_a, PSet) and isinstance(subobj_b, PSet):
+        return _create_diffs_for_sets(
+            current_path, subobj_a, subobj_b)
+    return pvector([_Set(path=current_path, val=subobj_b)])
```
wallrj: :question: I'm having to think about this default. Maybe add an explanatory comment.

flocker/control/test/test_diffing.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+        self.assertThat(
+            wire_decode(encoded_removal_diff).apply(d),
+            Equals(deployment)
+        )
+
+    def test_set_diffing_smart(self):
+        """
+        Small modifications to a sets have diffs that are small. Their reverse
+        is also small.
```
wallrj: typo "to a sets"

flocker/control/_diffing.py
wallrj: And `coverage` shows some missing coverage in this function.

```
coverage run --source flocker.control._diffing --branch $(type -p trial) flocker.control.test.test_diffing
```

```
(4421) (FLOC-4421-flocker-diffing ? =)[~/.../HybridLogic/flocker]$ coverage report
Name                          Stmts   Miss Branch BrPart  Cover
---------------------------------------------------------------
flocker/control/_diffing.py      68      4     40      3    94%

```


```
>     if subobj_a == subobj_b:
!         return pvector([])
>     elif type(subobj_a) != type(subobj_b):
!         return pvector([_Set(path=current_path, val=subobj_b)])
>     elif isinstance(subobj_a, PClass) and isinstance(subobj_b, PClass):
>         a_dict = subobj_a._to_dict()
>         b_dict = subobj_b._to_dict()
>         return _create_diffs_for_mappings(current_path, a_dict, b_dict)
>     elif isinstance(subobj_a, PMap) and isinstance(subobj_b, PMap):
>         return _create_diffs_for_mappings(
>             current_path, subobj_a, subobj_b)
>     elif isinstance(subobj_a, PSet) and isinstance(subobj_b, PSet):
>         return _create_diffs_for_sets(
>             current_path, subobj_a, subobj_b)
!     return pvector([_Set(path=current_path, val=subobj_b)])

```

wallrj:
Thanks @sarum90 

This looks great.

I see now why the Node and Application pmaps make this so much easier.

I noted some missing test coverage which we should probably address before merging.

flocker/control/_diffing.py
sarum90: Oh! Thanks for that idea. I was trying to get away with just using hypothesis... but I should absolutely add smaller unit tests, and I'll use coverage to verify line coverage with those tests.

flocker/control/_diffing.py
sarum90: done

flocker/control/_diffing.py
sarum90: Chose `_IDiffChange` as this is a private interface.

flocker/control/_diffing.py
sarum90: done

flocker/control/_diffing.py
sarum90: A little funky because I don't care what the type of the members of the field are... but that might express what this is better.

flocker/control/_diffing.py
sarum90: Noted.

Turns out Evolvers don't actually have a `.transform` method, so this change is a bit non-trivial.

Something similar would be to aggregate all of the transforms, and then apply them all in a single call to obj.transform.

But... That seems like the sort of change that can be done in a subsequent review, and can be done on-demand if required by performance analysis.

flocker/control/_diffing.py
sarum90: Fixed.

flocker/control/_diffing.py
sarum90: Yea not sure what I was doing here. Clearly nothing worth trying to fix.

flocker/control/_diffing.py
sarum90: Yea that's much cleaner. Nice catch! Not sure what I was thinking.

flocker/control/_diffing.py
sarum90: Fixed.

flocker/control/_diffing.py
sarum90: Added some unit tests:
```
Name                          Stmts   Miss Branch BrPart  Cover
---------------------------------------------------------------
flocker/control/_diffing.py      67      0     36      0   100%
```

sarum90: And a comment

flocker/control/test/test_diffing.py
sarum90: Fixed.

flocker/control/_diffing.py
wallrj: @sarum90 's comment here is related to the issue described  in https://github.com/tobgu/pyrsistent/issues/89 and offers a possible alternative implementation than https://github.com/ClusterHQ/flocker/pull/2839

PR #2807, opened 2016-06-02
[FLOC-4420] Generation Hashes

sarum90:
This PR adds code to compute consistent 128-bit hashes. This means I can compute the hash of the configuration on the control node, send it to an agent, the agent can compute the hash of its configuration, and if they are equal it is safe to assume that they actually are the same object.

Note that we are using mmh3 for hashing rather than a security grade hash. I believe this to be okay we are sending the hash over the same protocol that we are sending the full object over, so any malicious MITM could just replace the whole content, there's no benefit from being able to contrive a version of the config that matches a given hash.

Also see https://en.wikipedia.org/wiki/Birthday_problem#Probability_table for why I believe there is an _extremely_ low probability of hash collisions (You would need 2.6×10^18 deployments before you'd have a 1% chance of having a hash collision (128-bits)).


flocker/control/_persistence.py
```diff
@@ -306,6 +310,77 @@ def _cached_dfs_serialize(input_object):
...
+    """
+    # Ensure this is a quick function for basic types:
+    # Note that ``type(x) in frozenset([str, int])`` is faster than
+    # ``isinstance(x, (str, int))``.
+    input_type = type(input_object)
+    if (
+            input_object is None or
+            input_type in _BASIC_JSON_TYPES
+    ):
+        if input_type == unicode or input_type == bytes:
```
wallrj: :question: Maybe `input_type in (unicode, bytes)` or `input_type in abc.types.StringTypes`

flocker/control/_persistence.py
```diff
@@ -306,6 +310,77 @@ def _cached_dfs_serialize(input_object):
...
+            # Add a token to identify this as a string.
+            object_to_process = b''.join([_STR_TOKEN, bytes(input_object)])
```
wallrj: :question: I don't know what `bytes(u"some unicode string")` does. Which encoding does it use? It might be clearer to say `input_object.encode('utf8')`

wallrj: It's not immediately obvious why we need to prepend the `_STR_TOKEN`. 
Perhaps expand the comment....I'll read on.

flocker/control/_persistence.py
```diff
@@ -306,6 +310,77 @@ def _cached_dfs_serialize(input_object):
...
+        # different hashes.
+        object_to_process = frozenset(object_to_process.iteritems()).union(
+            [_MAPPING_TOKEN]
+        )
+
+    if isinstance(object_to_process, Set):
+        def xor_bytes(a, b):
+            for i in xrange(len(a)):
+                a[i] ^= ord(b[i])
+            return a
```
wallrj: :question: Why nested? Make this a module function if possible.

flocker/control/_persistence.py
```diff
@@ -306,6 +310,77 @@ def _cached_dfs_serialize(input_object):
...
+
+        sub_hashes = (generation_hash(x) for x in object_to_process)
+        result = bytes(
+            reduce(xor_bytes, sub_hashes, bytearray(_NULLSET_TOKEN))
+        )
```
wallrj: Huh, that's neat!

wallrj: I wondered if there might be something in the `cryptography` library that does this even faster.

flocker/control/test/test_persistence.py
```diff
@@ -830,3 +834,125 @@ def test_can_create_latest_golden(self):
...
+
+class GenerationHashTests(TestCase):
+    """
+    Tests for generation_hash.
+    """
+
+    @given(st.data())
+    def test_no_hash_collisions(self, data):
+        """
+        Hashes of different deployments do not have hash collisions.
```
wallrj: I struggle to understand this test. Advanced hypothesis by the looks of it! Please expand the docstring and add some comments.

wallrj:
Looks great @sarum90 

You're obviously having fun! :smile: 

Please force a build and answer / address my inline comments before merging.

Maybe check code coverage too (as in the previous PR)

sarum90:
I actually did the code coverage this time, This PR has full coverage of all new lines of code. Thanks for reminding me how to do so in the previous one. I'll be sure to resolve all of your other concerns before merging.

flocker/control/_persistence.py
sarum90: It's just so that the hash of `unicode(5)` is different than the hash of `int(5)`. I'll add a comment to that effect.

flocker/control/_persistence.py
sarum90: Split out due to unicode handling.

flocker/control/_persistence.py
sarum90: Added comment, fixed encoding.

flocker/control/_persistence.py
sarum90: A long time ago I was advised to use nested methods for scoping reasons. But I think there are better reasons to make it a module function (don't accidentally inherit surrounding variables). If you are in the module function camp I'll gladly move it to a module level function.

sarum90: moved to module level, added docstring, and made the argument names more descriptive.

flocker/control/_persistence.py
sarum90: Yea glanced around and didn't see anything, but there definitely are faster implementations in C, so please let me know if you find something faster.

I think the bytearray is already much faster than the first version of this that I wrote.

flocker/control/test/test_persistence.py
sarum90: Done

PR #2808, opened 2016-06-02
[FLOC-4371] Remove flocker-deploy

myechuri:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4371

``flocker-deploy`` is no longer relevant. Having it in our codebase and packing it as part of ``flocker-cli`` enables partners/customers to (mis)use it. This PR removes traces of ``flocker-deploy`` in our repo so that it is not longer available when flocker packages are installed.


wallrj:
Thanks @myechuri 

Looks good. A few points to address before merging:
 * Add a line to the release notes -- explaining that flocker-deploy has been removed.
   There are quite a few references to it in CHQ and external blog posts and I guess people might still be reading those: https://www.google.co.uk/search?q=%22flocker-deploy%22 so an explanation in the release notes should help avoid confusion.

 * Some of the cli installation tests rely on `flocker-deploy --version`  ...they could be changed to call `flocker-ca --version`

```
$ ./admin/run-client-tests --pip  --distribution=ubuntu-14.04 --branch=remove-flocker-deploy-FLOC-4371
...
++ flocker-deploy --version
+ test 1.12.0 = 1.12.0

```
Those tests don't seem to be run by Jenkins and I find it hard to run the tests against this branch because they rely on there being a Python wheel `.whl` file to install from S3. But those tests will start to fail when we do the next release so we should fix them here.

* The entire `flocker.control._config` module is now unused. Either remove it and any affected tests here, or create a followup issue and add a link to the top of that module.

* The `flocker.cli.configue_ssh` remains, but there's no command line tool to invoke it. It'd be nice if we could still perform the SSH key distribution for testing / demoing the zfs backend. Create a followup issue and link to it from the module explaining why it's still there.

* The `/configuration/_compose` REST endpoint remains and should be removed at the same time we remove the `flocker.control._config` module.

Address or answer those points and then merge.

PR #2809, opened 2016-06-03
[FLOC-4423] A utility to assist in generation tracking.

sarum90:
This PR creates a utility that will be used in the next PR to track the current generation of the configuration and state, and be able to upgrade from older versions of the config and state to the latest version.


flocker/control/_generations.py
```diff
@@ -0,0 +1,112 @@
...
+        ``_GenerationRecord`` to the queue (and drop one if the queue is full).
+        """
+        if latest == self._latest_object:
+            return 
+        latest_hash = make_generation_hash(latest)
+
+        if (self._latest_object is not None and
+                latest_hash != self._latest_hash):
+            new_diff = create_diff(self._latest_object, latest)
+            if self._latest_hash is not None:
```
wallrj: :question: I can't think why this check is necessary? `make_generation_hash` always returns a `GenerationHash`.

wallrj: Coverage report agrees.

flocker/control/_model.py
```diff
@@ -1232,10 +1232,50 @@ def get_information_wipe(self):
...
+
+
+class GenerationHash(PClass):
+    """
+    Generation hash uniquely defines a configuration or state. A value of None
+    indicates that there is no configuration or state known.
+
+    Note that this is a tuple of ints rather than ``bytes`` because the flocker
+    ``wire_encode`` function assumes that ``bytes`` can be encoded as JSON
+    strings, which is not true of generation hashes.
```
wallrj: :question: Did you consider `base64` encoding instead?

flocker/control/test/test_diffing.py
```diff
@@ -54,6 +56,24 @@ def test_deployment_diffing(self, deployment_a, deployment_b):
             Equals(deployment_b)
         )
 
+    @given(
+        st.lists(deployment_strategy(), min_size=3, max_size=10)
+    )
+    def test_deployment_diffing_composable(self, deployments):
+        """
+        Diffs should compose to create an aggregate diff.
```
wallrj: :question: Haha, It took me a while to get my head around this test. Add a little comment outlining the strategy for my benefit.

flocker/control/test/test_generations.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def test_basic_use_works(self):
+        """
+        After inserting a bunch of deployments into a ``GenerationTracker`` in
+        sequence, the ``Diff`` returned from ``get_diff_from_hash_to_latest``
+        can be applied to convert each of the deployments to the latest
+        deployment.
+        """
+        deployments = list(deployment_strategy().example() for _ in xrange(5))
+        deployments[3] = deployments[1]
```
wallrj: :question: It's not obvious to me why ^^ ?. Add a comment.

wallrj: :question: Yeah, you'll have to explain your strategy  / intent in this test too. I can't work it out :confused:

flocker/control/testtools.py
```diff
@@ -312,3 +353,22 @@ def deployment_strategy(draw, min_number_of_nodes=1):
...
+
+    :param int number_of_deployments: The number of deployments to create.
+
+    :returns: A strategy to create ``number_of_deployments`` ``Deployment`` s.
+    """
+    node_uuid_pool = draw(node_uuid_pool_strategy())
+    return tuple(
+        draw(deployment_strategy(node_uuid_pool=node_uuid_pool))
+        for _ in xrange(number_of_deployments)
+    )
```
wallrj: Cool.

flocker/control/test/test_generations.py
```diff
@@ -0,0 +1,82 @@
...
+class GenerationTrackerTests(TestCase):
+
+    def test_basic_use_works(self):
+        """
+        After inserting a bunch of deployments into a ``GenerationTracker`` in
+        sequence, the ``Diff`` returned from ``get_diff_from_hash_to_latest``
+        can be applied to convert each of the deployments to the latest
+        deployment.
+        """
+        deployments = list(deployment_strategy().example() for _ in xrange(5))
```
wallrj: :question: Can you use `related_deployments_strategy` here?

wallrj:
Looks great @sarum90 

It's taken me a while to get my head around the code in this branch, especially some of the tests.
Consider adding some further explanatory docstrings and comments and then merge.

flocker/control/_generations.py
sarum90: Removed, Thanks!

That was left over from an earlier version of this object that was harder to reason about.

flocker/control/_model.py
sarum90: Yea, that'd be equivalent, and would serialize to something smaller as well, I guess then in make_generation_hash I could compute the base64 encoding and put it in this object. I don't think there is a strong argument to go for one over the other.

On 1 hand base64 serializes smaller, on the other hand python doesn't have a good type to verify that a value is a base64 string... (I guess I could write a pyrsistent invariant or something...)

flocker/control/test/test_diffing.py
sarum90: Added.

flocker/control/test/test_generations.py
sarum90: Lol, yea I wrote that strategy specifically to use it in locations like this, and then just forgot about it :p. Using that instead.

flocker/control/test/test_generations.py
sarum90: Yea... it was kind of convoluted. Added comments.

The code above is a bit more interesting if the queue has repeated deployments (the line labelled `if record.generation_hash == generation_hash:` gets hit multiple times).

Most of the other things are just trying to double check that inserting the same object as the latest a bunch is a no-op, and that you always get the same diff out of the tracker.

flocker/control/testtools.py
sarum90: Thanks.

Yea I like this idea in general... if I had more time and thought this code was long-term important I would have done more strategies like this... generate a pool of node uuids, generate a pool of dataset ids, generate a pool of applications, create a bunch of deployments from that, etc.

sarum90:
Added comments, waiting for tests.

I _think_ this might all get a bit clearer with my next PR.... which I'll throw together now :).

sarum90:
The only most recent failure was because Jenkins killed the slave. I've seen all of the other tests pass with the code that is running, so I'm going to merge.

Thanks for the review Richard!

PR #2812, opened 2016-06-03
[FLOC-4428] Guard against a pyrsistent.tranform call.

sarum90:
In profiling I discovered that there is a specific pyrsistent.transform call that takes quite a bit of time.

I have reason to believe that a simple guard might assist in speeding up some of the performance.

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -380,16 +380,21 @@ class ControlServiceLocator(CommandLocator):
     @SetBlockDeviceIdForDatasetId.responder
     def set_blockdevice_id(self, dataset_id, blockdevice_id):
         deployment = self.control_amp_service.configuration_service.get()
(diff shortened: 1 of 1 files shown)


wallrj:
Thanks @sarum90 

Good catch.
This seems to be called as part of @tomprince 's dataset to blockdevice_id registration process.
It makes me wonder why the dataset registration is happening multiple times for an individual dataset.
Possible reasons:
 * All agents are discovering the new block device and attempting to register it. I glanced through the code and it looks like this is possible. `f.n.a.blockdevice.discover_state` looks at *all* volumes when marking those that need to be REGISTERed. 
   * Perhaps it should only attempt to register datasets that the configuration says are primary on *this*  host?

* Perhaps agents are attempting to register the datasets over and over. 
  * I can't think of a reason for that happening.

Anyway, this seems like a sensible change and the tests all pass, so I'll merge.

PR #2818, opened 2016-06-08
[FLOC-4307] Allow tuning of container swappiness

myechuri:
[Not ready for review.]


flocker/control/_model.py
```diff
@@ -314,6 +314,10 @@ class Application(PClass):
     :ivar command_line: Custom command to run using the image, a ``PVector``
         of ``unicode``. ``None`` means use default.
 
+    :ivar swappiness: Tunable swappiness of the container (default is None,
+        which disables swapping). Valid value is in between (including)
+        [0, 100].
+
```
wallrj: :question: Maybe check the default behaviour. Docker docs say:
> By default, if you are not using --memory-swappiness, memory swappiness value will be inherited from the parent.

* https://docs.docker.com/engine/reference/run/#/swappiness-constraint

flocker/control/_model.py
```diff
@@ -328,6 +332,7 @@ class Application(PClass):
                         type=PMap)
     running = field(mandatory=True, initial=True, type=bool)
     command_line = pvector_field(unicode, optional=True, initial=None)
+    swappiness = field(mandatory=False, initial=None)
```
wallrj: :question: `type=int` (probably?)

flocker/control/_persistence.py
```diff
@@ -193,6 +193,36 @@ def upgrade_from_v4(cls, config):
...
+                    app[1].update({u'swappiness': None})
+                new_node = node[1]
+                new_node[u"applications"][u"values"] = applications
+                new_node_values.append((uuid, new_node))
+            decoded_config[u"deployment"][u"nodes"] = {
+                u"values": new_node_values,
+                _CLASS_MARKER: "PMap"
+            }
+        return dumps(decoded_config)
+
```
wallrj: :question: You probably don't need this upgrade script, since `swappiness` is an optional attribute in the model (`mandatory=False`)

flocker/control/httpapi.py
```diff
@@ -691,6 +692,9 @@ def create_container_configuration(
         :param command_line: If not ``None``, the command line to use when
             running the Docker image's entry point.
 
+        :param swappiness: Tune container's memory swappiness
+            (default of None is no swap).
```
wallrj: :question: Again, check the meaning of the default.

flocker/control/schema/types.yml
```diff
@@ -197,6 +197,14 @@ definitions:
       type: string
       maxLength: 4096
 
+  swappiness:
+    title: "Swappiness"
+    description: |
+      Tune a container's memory swappiness behavior (default of 0 is no swap).
```
wallrj: :question: Is the default  `0` or `None` (i.e. omitted)

flocker/control/test/configurations/configuration_1_v5.json
```diff
@@ -1 +1 @@
...
...3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}, "$__class__$": "AttachedVolume"}, "memory_limit": null, "name": "myapp", "links": [], "$__class__$": "Application", "restart_policy": {"$__class__$": "RestartNever"}, "image": {"tag": "7.6", "repository": "postgresql", "$__class__$": "DockerImage"}, "command_line": null, "environment": {"values": [], "$__class__$": "PMap"}, "running": true, "ports": [], "cpu_shares": null}]], "$__class__$": "PMap"}, "manifestations": {"values": [["4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", {"dataset": {"deleted": false, "dataset_id": "4e7e3241-0ec3-4df6-9e7c-3f7e75e08855", "metadata": {"values": [["name", "myapp"]], "$__class__$": "PMap"}, "maximum_size": null, "$__class__$": "Dataset"}, "primary": true, "$__class__$": "Manifestation"}]], "$__class__$": "PMap"}, "uuid": {"hex": "ab294ce4-a6c3-40cb-a0a2-484a1f09521c", "$__class__$": "UUID"}, "$__class__$": "Node"}]], "$__class__$": "PMap"}, "leases": {"values": [], "$__class__$": "PMap"}, "$__class__$": "Deployment"}, "version": 5, "$__class__$": "Configuration"}
```
wallrj: :question: Might not be necessary. (see above)

flocker/control/test/persisted_model.json
```diff
@@ -34,6 +34,7 @@
                     "running": [
                         "__builtin__.bool"
                     ],
+                    "swappiness": [],
```
wallrj: :question: Oh, I'd forgotten about this model change test. I'll have to remind myself how this works....

wallrj:
Thanks @madhuri


Looks good.

* Ignore my comments about "not requiring an upgrade script" . I'm wrong about about that and you're doing the right thing. I was looking for equivalent upgrade scripts for things like container environment variables and cpu shares. There aren't eny, but that's the point. We forgot to add the upgrade script for those things and that's why @itamarst added the "missing upgrade test." (AFAIK)

 * The other test failure in `test_consistent_hash` is a side effect of a hard coded mm3 hash in the test. See @sarum90 's comment the test docstring. The hash changed because of the new swappiness attribute so you just need to update that hash.

* I'm not clear what the default "swappiness" value will be. Some places say that it "disables" swap. Other places say it defaults to `0` and other places say it defaults to `None` which from the docs appears to mean that the swappiness is inherited from the parent namespace.

I'd probably also recommend merging the branch forward to pick up various fixes that have been merged into master since this branch started.

flocker/control/_model.py
myechuri: Fixed.

myechuri:
Thanks for the review, @wallrj !

PR #2830, opened 2016-06-15
[FLOC-4445] Redirect obsolete documentation to latest

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4445

I set up Google webmaster tools and that confirms that people are searching for "install flocker" and getting a load of ancient pages;

screenshot from 2016-06-15 17-32-40

I considered adding a robots.txt to "Dissallow"  crawlers from finding all those  old docs, but am nervous whether that might spoil our search ranking.
 * https://en.wikipedia.org/wiki/Robots_exclusion_standard
 * https://support.google.com/webmasters/topic/4598466

We could just delete some of those *really* old docs. Google would then remove them from it's index.
That might have the same negative effect on our search ranking.

And finally I investigated "sitemap" files.....but that doesn't really seem to buy us much...assuming that Google can follow all our links.
 * https://support.google.com/webmasters/answer/156184?hl=en

So I've updated the redirect rules and done a test publication of the docs here:
You should see a 301 redirect for each of the versions in the PR:

```
[richard@richardw-pet-instance-1 flocker]$ curl --insecure -v https://docs.staging.clusterhq.com/en/1.11.0/
* About to connect() to docs.staging.clusterhq.com port 443 (#0)
*   Trying 52.85.178.188...
* Connected to docs.staging.clusterhq.com (52.85.178.188) port 443 (#0)
* Initializing NSS with certpath: sql:/etc/pki/nssdb
* skipping SSL peer certificate verification
* SSL connection using TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256
* Server certificate:
*       subject: CN=docs.staging.clusterhq.com,OU=Gandi Standard SSL,OU=Domain Control Validated
*       start date: Jan 22 00:00:00 2015 GMT
*       expire date: Jan 22 23:59:59 2016 GMT
*       common name: docs.staging.clusterhq.com
*       issuer: CN=Gandi Standard SSL CA 2,O=Gandi,L=Paris,ST=Paris,C=FR
> GET /en/1.11.0/ HTTP/1.1
> User-Agent: curl/7.29.0
> Host: docs.staging.clusterhq.com
> Accept: */*
> 
< HTTP/1.1 301 Moved Permanently
< Content-Length: 0
< Connection: keep-alive
< Date: Wed, 15 Jun 2016 16:37:48 GMT
< Location: https://docs.staging.clusterhq.com/en/latest/
< Server: AmazonS3
< X-Cache: Miss from cloudfront
[...]

PR #2834, opened 2016-06-16
[FLOC-4448] Do root redirects instead of prefix based redirects for versions of docs prior to the latest restructuring.

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4448

@ferrantim noticed that the redirects we introduced in #2830 broke lots of the links from our blog posts.

Instead we've decided to redirect versions prior to 1.11.0 to the root of the docs site.
1.11.0 and 1.12.0 docs will redirect to the corresponding page in `en/latest`.

@ferrantim Needed this urgently so I already installed these redirects on the live 1.13.0 documentation using the following command.

```
admin/publish-docs --flocker-version=1.13.0 --doc-version=1.13.0 --production
```

Which re-publishes the existing 1.13.0 docs and re-installs the S3 redirects from the current branch.

You can test the effects with curl commands like:

```
# Should redirect to /en/latest/
curl -L -v https://docs.clusterhq.com/en/1.0.0/foo/bar > /dev/null

# Should redirect to /en/latest/foo/bar ...which will then be a 404 because that page doesn't exist.
curl -L -v https://docs.clusterhq.com/en/1.12.0/foo/bar > /dev/null
```


wallrj:
Approved by @ferrantim  in #flocker-team on Slack.
Merging.

PR #2833, opened 2016-06-16
[FLOC-4447] Add google-compute-engine as a requirement

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4447

This is caused by:
 * https://github.com/GoogleCloudPlatform/compute-image-packages/issues/262
 * https://github.com/GoogleCloudPlatform/compute-image-packages/pull/263

I've added google-compute-engine package here but in re-running the `admin/update-requirements` tool I've pulled in a few other library upgrades and a new version of setuptools. 
All good stuff I'm sure and all the tests still pass.

I had some trouble with the GCE functional tests:
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/google-compute-engine-FLOC-4447/job/_run_trial_for_gce_storage_driver_on_CentOS_7_flocker.node.agents.functional/
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/google-compute-engine-FLOC-4447/job/_run_trial_for_gce_storage_driver_on_Ubuntu_trusty_flocker.node.agents.functional/

But that turned out be the fault of  the old version of virtualenv installed on our static GCE slaves.

I manually upgraded `pip` and `virtualenv` as the jenkins user on  those slaves.

We should really update the VM images too, but this works for now.
(as discussed with @sarum90 a week or two ago)

PR #2837, opened 2016-06-17
[FLOC-4451] Bundle eliot-tree in the Flocker package

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4451
Stacked on #2833 (review that one first)

We always used to bundle it but I missed  it out when I landed #2792

Unfortunately in re-running the `update-requirments` tool I've pulled in a just released version of python-novaclient.
Their release notes are non-existent but the changes I can see, look minor;
 * https://github.com/openstack/python-novaclient

I'll run all the acceptance tests to ensure that it hasn't broken anything.
If I have, I'll pin to the previous version.

### DIFF
--- a/admin/build_targets/centos-7/Dockerfile
+++ b/admin/build_targets/centos-7/Dockerfile
@@ -6,7 +6,7 @@
 FROM clusterhqci/fpm-centos-7
 MAINTAINER ClusterHQ <contact@clusterhq.com>
 COPY requirements/*.txt /requirements/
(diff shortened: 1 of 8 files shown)


exarkun:
I verified that installing the packages from this branch on the Docker images centos:7, ubuntu:14.04, and ubuntu:15.14 provides an `eliot-tree` executable.  The changes themselves seem mostly sensible (though I won't claim to completely understand all aspects of Flocker packaging).  Based on the test results, I think this is good to merge.  Thanks.

PR #2845, opened 2016-06-28
[FLOC-4339] Run all tests on Ubuntu 16.04

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4339

PR #2851, opened 2016-07-01
[FLOC-4463] Run unit tests on Ubuntu 16.04

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4463

This fixes a couple of `admin.tests` and updates build.yaml to run all our standard unit tests on the xenial build slave.

The lintian error is that the package description must have more than one word:
 * https://lintian.debian.org/tags/description-too-short.html

Once this is merged I'll merge forward https://github.com/ClusterHQ/flocker/pull/2845 (the remaining tests) and put it up for review

### DIFF
--- a/admin/test/test_packaging.py
+++ b/admin/test/test_packaging.py
@@ -766,7 +766,7 @@ class LintPackageTests(TestCase):
             vendor="Acme Corporation",
             maintainer='Someone <noreply@example.com>',
             architecture="all",
-            description="A Test Package.\n\n" + ("Lorum ipsum. " * 100),
+            description="A Test Package\n\nThe description.",
             category="none",
             dependencies=[]
         ).run()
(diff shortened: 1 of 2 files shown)


admin/test/test_packaging.py
```diff
@@ -199,7 +199,7 @@ def assert_deb_content(test_case, expected_paths, package_path):
             continue
         actual_paths.add(FilePath('/').descendant(f.segmentsFrom(output_dir)))
 
-    test_case.assertEqual(expected_paths, actual_paths)
+    test_case.assertEqual(set(), expected_paths.difference(actual_paths))
```
myechuri: Can you please explain why this is better than original? Thanks!

wallrj: Yeah, it's to check that the package *contains* the expected_paths  rather than checking that the list of packaged files exactly matches `expected_paths`.

And if any of `expected_paths` are not included in the package they will be included in the assertion failure message making it easier to debug the failure.

PR #2839, opened 2016-06-21
[FLOC-4452] Delay invariant checks until all diff operations have been applied

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4452

PR #2858, opened 2016-07-07
[DO NOT REVIEW OR MERGE] Move pyrsistent serialization and hashing helpers to a new module.

wallrj:
Here's an idea for re-organizing the serialization stuff so that I can e.g. use wire_encode from _diffing in #2839.

Unfortunately it's not quite enough. 

Perhaps instead, `wire_decode` should accept an `acceptable_classes` argument rather than having the global `SERIALIZABLE_CLASSES` variable.
After all we'd never expect to find _diffing classes in the control service configuration which also uses the wire_encode format.

Needs more thought.

### DIFF
--- a/flocker/control/_diffing.py
+++ b/flocker/control/_diffing.py
@@ -257,9 +257,3 @@ def compose_diffs(iterable_of_diffs):
             pvector().evolver()
         ).persistent()
     )
-
-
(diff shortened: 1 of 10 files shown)


Azulinho:
I think you can do a lot better than this, seriously

PR #2867, opened 2016-07-20
[FLOC-4474] Ubuntu 16.04 installation documentation

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4474
Stacked on: https://github.com/ClusterHQ/flocker/pull/2853

I know there's lot's of duplication here but I haven't got time to fix that now :frowning: 

I've performed a test release (1.13.1.dev2) to check that the documentation looks ok when published and to allow us to test the installation instructions
 * https://docs.clusterhq.com/en/devel/flocker-standalone/manual-install.html

PR #2853, opened 2016-07-02
[FLOC-4341] Ubuntu 16.04 acceptance tests

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4341

PR #2872, opened 2016-07-25
[FLOC-4349] Use Ubuntu 16.04 in cloudformation template

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4349
Stacked on: #2845, #2853, #2867 

I built new images and a new cloudformation template and ran the tests manually:

```
(4475) richard@richardw-pet-ubuntu1604:~/flocker$ KEEP_STACK=TRUE CLOUDFORMATION_TEMPLATE_URL=https://s3.amazonaws.com/installer.downloads.clusterhq.c
om/flocker-cluster.cloudformation.1.13.1.dev2.json ACCEPTANCE_YAML=/home/richard/acceptance.yml trial admin.test.test_installer                       
admin.test.test_installer
  DockerComposeTests
    test_docker_compose_up_postgres ...                                    [OK]

Ran 1 tests in 748.989s

PASSED (successes=1)
```


admin/acceptance.py
```diff
@@ -1466,7 +1466,7 @@ def cluster_cleanup():
     try:
         yield runner.ensure_keys(reactor)
         cluster = yield runner.start_cluster(reactor)
-        if options['distribution'] in ('centos-7', 'rhel-7.2'):
+        if options['distribution'] in ('centos-7', 'rhel-7.2', 'ubuntu-16.04'):
```
moshez: Why does this mismatch the DISTRIBUTIONS?

admin/installer/_images.py
```diff
@@ -53,7 +52,7 @@ class AWS_REGIONS(Values):
 
 DEFAULT_IMAGE_BUCKET = u'clusterhq-installer-images'
 DEFAULT_BUILD_REGION = AWS_REGIONS.US_WEST_1
-DEFAULT_DISTRIBUTION = u"ubuntu-14.04"
+DEFAULT_DISTRIBUTION = u"ubuntu-16.04"
```
moshez: Can we have the default just be the first element of DISTRIBUTIONS?
This way there's only one place to change.

admin/installer/_images.py
```diff
@@ -144,7 +143,7 @@ class PackerConfigure(PClass):
     :ivar template: The prototype configuration to use as a base. One of
         `docker` or `flocker`.
     :ivar distribution: The operating system distribution to install.
-        ubuntu-14.04 is the only one implemented so far.
+        ubuntu-16.04 is the only one implemented so far.
```
moshez: This is really weird. It seems like we just change the hard coded default but keep it as a hard coded default. Can we get rid of this?

admin/installer/flocker-configuration-getter.sh
```diff
@@ -23,15 +23,25 @@ chmod --recursive a-rwx,u+rwX "${TMP_DIR}"
 mv "${TMP_DIR}" "${FLOCKER_CONFIG_DIRECTORY}"
 
 if test "${node_number}" -eq "0"; then
-    service flocker-control restart
+    # Enable flocker-control on node0 and disable the flocker agent services.
```
moshez: There's a lot of copy-pasta here and weird abstractions.
It might be time to consider rewriting in Python, and having proper functions like "systemd_setup_service(name)" and "systemd_unsetup_service(name)" and then doing set-based service setup/unsetup so that the input can be more declarative.

admin/acceptance.py
wallrj: Because this branch is for distros where we use `journalctl` to capture the remote logs.
And I think RHEL7 isn't in the `DISTRIBUTIONS` because the RHEL7 support isn't completely finished :frowning:

admin/installer/_images.py
wallrj: There was a plan to have cloudformation templates for other distributions, but we never got round to it....and probably never will.
I'll remove this.

wallrj:
I removed the unused distribution option and manually tested.

```
(4349) richard@richardw-pet-ubuntu1604:~/flocker$ ./admin/publish-installer-images --source-ami-map "$(<ami_map_ubuntu.json)" --template docker
....
1469527760,,ui,say,--> amazon-ebs: AMIs were created:\n\nus-west-1: ami-62f3b202
{"us-west-1": "ami-62f3b202"}
```

Merging.

Issue #2854, opened 2016-07-04
docker plugin not working with docker 1.12.0-rc3


Nomon:
There is a new field passed to the docker plugin called 'ID':
https://github.com/docker/docker/pull/21015

The plugin (flocker 1.13.0) fails to mount the volume with the following error:
```
{"exception": "exceptions.TypeError", 
     "reason": "volumedriver_mount() got an unexpected keyword argument 'ID'",
     "timestamp": 1467628063.957701, 
     "traceback": "
    Traceback: <type 'exceptions.TypeError'>: volumedriver_mount() got an unexpected keyword argument 'ID'
      /opt/flocker/local/lib/python2.7/site-packages/twisted/internet/defer.py:150:maybeDeferred
      /opt/flocker/local/lib/python2.7/site-packages/flocker/restapi/_infrastructure.py:264:loadAndDispatch 
      /opt/flocker/local/lib/python2.7/site-packages/twisted/internet/defer.py:150:maybeDeferred 
      /opt/flocker/local/lib/python2.7/site-packages/flocker/dockerplugin/_api.py:118:wrapped
     --- <exception caught here> ---  
     /opt/flocker/local/lib/python2.7/site-packages/twisted/internet/defer.py:150:maybeDeferred",
  "message_type": "eliot:traceback",
  "task_uuid": "6d68e5a3-1ef3-44e1-a477-e1cac6d76da2",
  "task_level": [3
}
```

wallnerryan:
@Nomon thanks for reporting this, I was away on some time off but will be digging into 1.12 and making sure things work correctly and making needed changes, will update this issue as we go along

wallrj:
Looks like 1.12 was released last week and all our Flocker Docker plugin tests are now broken :-(
 * http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/view/master/job/master/view/cron/job/_run_acceptance_on_GCE_Ubuntu_Xenial_with_GCE/7/#showFailuresLink

@wallnerryan Thanks for creating the Jira issue. 
Let me know if you want me to review / help with anything
 * https://clusterhq.atlassian.net/browse/FLOC-4480

wallnerryan:
should be fixed now that https://github.com/ClusterHQ/flocker/pull/2880 is merged. Next release will/should include this.

PR #2876, opened 2016-08-01
add ID to Mount / Unmount for docker 1.12

wallnerryan:
Because docker added IDs to Mount() / Unmount()

Right now we choose to do nothing with the ID but add it to our plugin layer so that calls will continue to work with docker 1.12

https://clusterhq.atlassian.net/browse/FLOC-4480
https://github.com/ClusterHQ/flocker/issues/2854

### DIFF
--- a/flocker/dockerplugin/_api.py
+++ b/flocker/dockerplugin/_api.py
@@ -197,11 +197,12 @@ class VolumePlugin(object):
 
     @app.route("/VolumeDriver.Unmount", methods=["POST"])
     @_endpoint(u"Unmount")
-    def volumedriver_unmount(self, Name):
+    def volumedriver_unmount(self, Name, ID):
         """
         The Docker container is no longer using the given volume.
 
         :param unicode Name: The name of the volume.
+        :param string ID: A unique ID for caller that requested the mount
 
         For now this does nothing. In FLOC-2755 this will release the
         lease acquired for the dataset by the ``VolumeDriver.Mount``
@@ -313,7 +314,7 @@ class VolumePlugin(object):
 
     @app.route("/VolumeDriver.Mount", methods=["POST"])
     @_endpoint(u"Mount")
-    def volumedriver_mount(self, Name):
+    def volumedriver_mount(self, Name, ID):
         """
         Move a volume with the given name to the current node and mount it.
 
@@ -321,6 +322,11 @@ class VolumePlugin(object):
         dataset is mounted locally.
 
         :param unicode Name: The name of the volume.
+        :param string ID: A unique ID for caller that requested the mount
+
+        Right now we do nothing with ID, but it can be used to track
+        the mount calls when multiple containers are trying to mount
+        the same volume.
 
         :return: Result that includes the mountpoint.
         """
--- a/flocker/dockerplugin/_api.py
+++ b/flocker/dockerplugin/_api.py
@@ -197,7 +197,7 @@ class VolumePlugin(object):
 
     @app.route("/VolumeDriver.Unmount", methods=["POST"])
     @_endpoint(u"Unmount")
-    def volumedriver_unmount(self, Name, ID):
+    def volumedriver_unmount(self, Name, ID=None):
         """
         The Docker container is no longer using the given volume.
 
@@ -314,7 +314,7 @@ class VolumePlugin(object):
 
     @app.route("/VolumeDriver.Mount", methods=["POST"])
     @_endpoint(u"Mount")
-    def volumedriver_mount(self, Name, ID):
+    def volumedriver_mount(self, Name, ID=None):
         """
         Move a volume with the given name to the current node and mount it.
 
--- a/flocker/dockerplugin/test/test_api.py
+++ b/flocker/dockerplugin/test/test_api.py
@@ -119,7 +119,8 @@ class APITestsMixin(APIAssertionsMixin):
         unmount_id = ''.join(random.choice(
             '0123456789abcdef') for n in xrange(64))
         return self.assertResult(b"POST", b"/VolumeDriver.Unmount",
-                                 {u"Name": u"vol", unicode(unmount_id): u""},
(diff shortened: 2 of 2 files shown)


flocker/dockerplugin/_api.py
```diff
@@ -197,11 +197,12 @@ def volumedriver_remove(self, Name):
 
     @app.route("/VolumeDriver.Unmount", methods=["POST"])
     @_endpoint(u"Unmount")
-    def volumedriver_unmount(self, Name):
+    def volumedriver_unmount(self, Name, ID):
```
wallrj: I've got a feeling that this is going to break with `Docker < 1.12.0` which won't supply an `ID`.
If I'm right, then we should give it a default value and add a test that the API can can be called with and without the ID.

flocker/dockerplugin/_api.py
```diff
@@ -313,14 +314,19 @@ def got_state(datasets):
 
     @app.route("/VolumeDriver.Mount", methods=["POST"])
     @_endpoint(u"Mount")
-    def volumedriver_mount(self, Name):
+    def volumedriver_mount(self, Name, ID):
```
wallrj: Here too.

flocker/dockerplugin/_api.py
```diff
@@ -313,14 +314,19 @@ def got_state(datasets):
...
         dataset is mounted locally.
 
         :param unicode Name: The name of the volume.
+        :param string ID: A unique ID for caller that requested the mount
+
+        Right now we do nothing with ID, but it can be used to track
+        the mount calls when multiple containers are trying to mount
+        the same volume.
 
         :return: Result that includes the mountpoint.
```
wallrj: :question: Maybe double check whether the return format has changed. Do we have to include the supplied ID in the json that we return from the API. 
If so, then we'll have to also update the JSON schema files.

wallnerryan: The return schema is the same, ID is only passed in the request and not the response. See here https://docs.docker.com/engine/extend/plugins_volume/

flocker/dockerplugin/_api.py
wallnerryan: good point. (Another reason our tests should include other docker versions)

wallnerryan: will verify it breaks with < 1.12, then make changes.

flocker/dockerplugin/test/test_api.py
```diff
@@ -115,8 +116,12 @@ def test_unmount(self):
         """
         ``/VolumeDriver.Unmount`` returns a successful result.
         """
+        unmount_id = ''.join(random.choice(
+            '0123456789abcdef') for n in xrange(64))
         return self.assertResult(b"POST", b"/VolumeDriver.Unmount",
```
wallrj: I looked at https://github.com/docker/docker/pull/21015 and ID comes from https://github.com/docker/docker/blob/master/pkg/stringid/stringid.go ....and which seems to be a / the container ID.

You could do something like  `os.urandom(32).encode('hex')` (we use that elsewhere in Flocker) which might be make the intended ID format a little clearer....although when I tested it seems to hang when there's insufficient entropy. And I guess that urandom may not be hooked up to trial's random seed option....so forget this comment :-)

wallrj:
Thanks @wallnerryan 

Please address or answer the comments above and merge when all the tests have passed.

Probably worth forcing the acceptance tests too....the  non-loopback ones listed under the "cron"

flocker/dockerplugin/_api.py
wallnerryan: looks like docker uses `1234` to test, not an actual ID. https://github.com/cpuguy83/docker/blob/2b6bc294fc7f9e08a9091833b021b7d2a01ad2a6/volume/local/local_test.go#L184

wallnerryan: confirmed this does break.

```
WARN[1065] 716e7f439070882877e3f401d06edc76b6c2d59cd48fb819b8fa3d80394afa02 cleanup: Failed to umount volumes: TypeError: volumedriver_unmount() takes exactly 3 arguments (2 given)
ERRO[1065] Handler for POST /v1.22/containers/716e7f439070882877e3f401d06edc76b6c2d59cd48fb819b8fa3d80394afa02/start returned error: TypeError: volumedriver_mount() takes exactly 3 arguments (2 given)
docker: Error response from daemon: TypeError: volumedriver_mount() takes exactly 3 arguments (2 given).
root@mha-aws-demo0:/home/ubuntu# docker version
Client:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 19:55:25 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.10.0
 API version:  1.22
 Go version:   go1.5.3
 Git commit:   590d5108
 Built:        Thu Feb  4 19:55:25 2016
 OS/Arch:      linux/amd64
```

flocker/dockerplugin/test/test_api.py
wallnerryan: Yeah, saw this too, tried to duplicate the stringid non cryptic id in python.

Im fine either way, the question is do we want to test multiple IDs with hypothesis or even test for malformed IDs ?

wallnerryan:
@wallrj how come the acceptance tests under cron dont kick off automatically ?

wallrj:
> how come the acceptance tests under cron dont kick off automatically ?

Because they're so slow. 

We made a decision that it was sufficient to see the loopback acceptance tests pass .
The loopback accceptance tests use the loopback backend and therefore don't  include any of the multi-node, dataset moving tests.
The full acceptance tests are run nightly on master. 
And the idea is that we check the build results on master and notice any new errors.

wallnerryan:
part of combined branch https://github.com/ClusterHQ/flocker/pull/2880

PR #2879, opened 2016-08-03
fix acceptance tests as of docker 1.12 release

wallnerryan:
fixes https://clusterhq.atlassian.net/browse/FLOC-4483

*NOTE*- this only fixes acceptance testing for Docker >= 1.12. This will likely break acceptance tests for older versions of docker < 1.12. If we plan on fixing that we should create a seperate issue that allows us to flip flop these type of configurations based on the docker version we want to test.


wallnerryan:
part of combines branch https://github.com/ClusterHQ/flocker/pull/2880

PR #2880, opened 2016-08-03
[FLOC-4480, FLOC-4483] Fix docker systemd dropin configuration files and add docker plugin ID for Docker 1.12

wallnerryan:
In order for one of these to pass, its needs the other. Creating one branch for both.

See:
 - https://clusterhq.atlassian.net/browse/FLOC-4480
 - https://clusterhq.atlassian.net/browse/FLOC-4483

NOTE- this only fixes acceptance testing for Docker >= 1.12. This will likely break acceptance tests for older versions of docker < 1.12. If we plan on fixing that we should create a seperate issue that allows us to flip flop these type of configurations based on the docker version we want to test.

### DIFF
--- a/flocker/provision/_install.py
+++ b/flocker/provision/_install.py
@@ -834,7 +834,7 @@ def task_enable_docker(distribution):
     # Use the Flocker node TLS certificate, since it's readily
     # available.
     docker_tls_options = (
-        '--tls=true --tlscacert=/etc/flocker/cluster.crt'
+        '--tlsverify --tlscacert=/etc/flocker/cluster.crt'
         ' --tlscert=/etc/flocker/node.crt --tlskey=/etc/flocker/node.key'
         ' -H=0.0.0.0:2376')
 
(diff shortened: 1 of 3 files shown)


flocker/provision/_install.py
```diff
@@ -834,10 +834,13 @@ def task_enable_docker(distribution):
     # Use the Flocker node TLS certificate, since it's readily
     # available.
     docker_tls_options = (
-        '--tlsverify --tlscacert=/etc/flocker/cluster.crt'
+        '--tls=true --tlscacert=/etc/flocker/cluster.crt'
```
wallrj: I think we should keep `'--tlsverify`. It ensures that docker daemon only accepts TLS connections from clients with a certificates signed by the Docker CA.

```
  --tls                                    Use TLS; implied by --tlsverify
  --tlscacert=~/.docker/ca.pem             Trust certs signed only by this CA
  --tlscert=~/.docker/cert.pem             Path to TLS certificate file
  --tlskey=~/.docker/key.pem               Path to TLS key file
  --tlsverify                              Use TLS and verify the remote
```

wallrj:
Thanks @wallnerryan 

Looks good.

* A couple of the Rackspace acceptance test nodes got killed before the tests finished....I've restarted those.
* I'd like to put back the ``--tlsverify`` flag for security.

I'll do that and kick off the tests again and then merge.

PR #2893, opened 2016-08-31
[FLOC-4490] Run Rackspace acceptance tests from Ubuntu Xenial

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4490

I *think* this might fix the problem by virtue of Ubuntu 16.04 having a more modern OpenSSL.

If this doesn't work, I'll try https://libcloud.readthedocs.io/en/latest/other/ssl-certificate-validation.html#changing-used-ssl-tls-version

I've also added  few more GCE Ubuntu 16.04 slaves in:
 * https://github.com/ClusterHQ/segredos/pull/33

Also helps avoid "slave went offline" errors during acceptance tests for rackspace.
If it proves successful, we can move the other acceptance tests here.


build.yaml
```diff
@@ -495,7 +496,7 @@ common_cli:
     # we provide one by starting a webserver and pointing the tests
     # to look over there
     REPO_PATH=/results/omnibus/${TRIGGERED_BRANCH}/${DISTRIBUTION_NAME}
-    DOC_ROOT=/usr/share/nginx/html
+    DOC_ROOT=/var/www/html
```
wallnerryan: naive question as im less familiar with editing this file. Will this break any backward compatibility? or will all cases no use `/var/www/html`

build.yaml
```diff
@@ -774,7 +775,6 @@ run_trial_cli: &run_trial_cli [
   *setup_venv,
   *setup_flocker_modules,
   *setup_coverage,
-  *setup_aws_env_vars,
```
wallnerryan: what env vars were these? Do we not need them in gce?

wallnerryan:
Few comments to be addressed. Otherwise seeing a failed test.

```
}}}

  File "/tmp/32302/local/lib/python2.7/site-packages/testtools/twistedsupport/_runtest.py", line 386, in _log_user_exception
    raise e
testtools.twistedsupport._spinner.TimeoutError: <flocker.acceptance.endtoend.test_diagnostics.DiagnosticsTests.test_export id=0x7f4afd0abb90> took longer than 300.0 seconds

]
test: flocker.acceptance.endtoend.test_dockerplugin.DockerPluginTests.test_create_container_with_v2_plugin_api
Build timed out (after 90 minutes). Marking the build as failed.
Build was aborted
Archiving artifacts
Archiving artifacts
ERROR: Failed to archive artifacts: run-acceptance-tests.log
java.io.IOException: java.io.IOException: Failed to extract /tmp/workspace/ClusterHQ-flocker/rackspace-libcloud-tls-FLOC-4490/_run_acceptance_on_Rackspace_Ubuntu_Trusty_with_Cinder/transfer of 1 files
```

not sure if this had to do with jenkins master disk space issues or if its related, but make sure its unrelated.

Otherwise, after the comments and the above test, + slack conversation LGTM

build.yaml
wallrj: This seems to be the standard directory for Nginx on Ubuntu 16.04 and all tests that require a local package repository are now run from Ubuntu 16.04.

build.yaml
wallrj: They're the variables required to run the `IBlockDeviceAPI` functional tests on AWS nodes.
I'm pretty sure they have been copy pasted all over the place without them actually being required.

wallrj:
Thanks @wallnerryan 

The Rackspace + Ubuntu 14.04 + cinder acceptance tests have been failing on master for a month.

* http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/master/view/on_Rackspace/job/_run_acceptance_on_Rackspace_Ubuntu_Trusty_with_Cinder/245/

Mostly due to "slave went offline" errors and at other times due to similar test timeouts.

I'm hoping that if we can eliminate the "slave went offline" errors by launching these tests from GCE, we'll be able to see clearly why the tests are timing out.

I'll merge this, see how it goes and we can revert back if it causes problems.

PR #2895, opened 2016-09-02
[FLOC-4491] Close the Cinder client persistent connection instead of flaky monkey patching of the server side.

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4491

The original test monkey patches twisted.web.http, aiming to force the server to make all HTTP connections non-persistent.
The reason for that was to avoid lingering persistent  HTTPChannels  in the reactor after the test....leading to Unclean reactor errors in the test.
But that doesn't work because the client has negotiated a persistent connection and thinks it can re-use it.
The client interprets the empty response which results from server TCP RESET as a ``BadStatusLine`` and aborts the connection...but only if the RST is received slowly...after the client has sent a subsequent HTTP request.
I *think* that's why this test was especially flaky on my devstack environment where the host machine is over loaded and the network packets are delayed.

```
  File "/home/ubuntu/.venvs/4489/local/lib/python2.7/site-packages/keystoneclient/session.py", line 449, in _send_request
    raise exceptions.ConnectionRefused(msg)
keystoneauth1.exceptions.connection.ConnectFailure: Unable to establish connection to http://127.0.0.1:53813/mimicking/CinderApi-0ea185/ORD/v2/5062690
87629312/volumes/detail?limit=2. (ConnectionError(ProtocolError('Connection aborted.', BadStatusLine("''",)),))


flocker.node.agents.functional.test_cinder.CinderFromConfigurationTests.test_no_retry_authentication
Ran 1 tests in 0.042s
```

The test now passes reliably (>1000) times on my devstack system

```
PASSED (successes=1)
Test Pass 1004
  CinderFromConfigurationTests
    test_no_retry_authentication ... Version 1 is deprecated, use alternative version 2 instead.
WARNING:cinderclient.api_versions:Version 1 is deprecated, use alternative version 2 instead.
Version 1 is deprecated, use alternative version 2 instead.
WARNING:cinderclient.api_versions:Version 1 is deprecated, use alternative version 2 instead.
                                      [OK]

Ran 1 tests in 0.046s

PASSED (successes=1)
Test Pass 1005

```

Previously the test failed on the second attempt.

### DIFF
--- a/flocker/node/agents/functional/test_cinder.py
+++ b/flocker/node/agents/functional/test_cinder.py
@@ -802,17 +802,17 @@ class CinderFromConfigurationTests(AsyncTestCase):
             cluster_id=make_cluster_id(TestTypes.FUNCTIONAL),
         )
 
-    def test_no_retry_authentication(self):
+    def _build_and_test_api(self, listening_port):
         """
-        The API object returned by ``cinder_from_configuration`` will retry
-        authentication even when initial authentication attempts fail.
+        Build the CinderBlockDeviceAPI configured to connect to the Mimic
+        server at ``listening_port``.
+        Patch twisted.web to force the mimic server to drop incoming
+        connections.
+        And attempt to interact with the disabled API server first and then
+        after re-enabling it to show that the API will re-authenticate even
+        after an initial failure.
         """
         import twisted.web.http
-        self.patch(
-            twisted.web.http.HTTPChannel,
-            'checkPersistence',
-            lambda self, request, version: False
-        )
         patch = MonkeyPatcher()
         patch.addPatch(
             twisted.web.http.HTTPChannel,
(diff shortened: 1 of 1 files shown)


flocker/node/agents/functional/test_cinder.py
```diff
@@ -802,79 +802,93 @@ def test_no_immediate_authentication(self):
...
+            backend=backend,
+            api_args=api_args,
+            reactor=object(),
+            cluster_id=make_cluster_id(TestTypes.FUNCTIONAL),
+        )
+        # List volumes with API patched to close incoming connections.
+        try:
+            result = api.list_volumes()
+        except ConnectFailure:
+            # Can't use self.asserRaises here because that would call the
```
wallnerryan: nit, `assertRaises` typo.

flocker/node/agents/functional/test_cinder.py
```diff
@@ -802,79 +802,93 @@ def test_no_immediate_authentication(self):
...
-        finishing = listing_volumes1.addCallback(check_successful_connection)
+        # List volumes with API re-enabled
+        result = api.list_volumes()
+        self.assertEqual([], result)
+
+        # Close the connection from the client side so that the mimic server
+        # can close down without leaving behind lingering persistent HTTP
+        # channels which cause dirty reactor errors.
+        # XXX: This is gross. Perhaps we need ``IBlockDeviceAPI.close``
+        (api
```
wallnerryan: this gets the client session and closes, yes agreed this is ugly, but what benefits would IBlockDeviceAPI.close() have other than code clarity? (assessing whether is worth adding instead of having this here)

flocker/node/agents/functional/test_cinder.py
```diff
@@ -802,79 +802,93 @@ def test_no_immediate_authentication(self):
...
+         ._cinder_volumes
+         .api
+         .client
+         .session
+         .session.close())
 
-        return finishing
+    def test_no_retry_authentication(self):
+        """
+        The API object returned by ``cinder_from_configuration`` will retry
```
wallnerryan: slightly confused by naming, `test_no_retry..` seems to indicate no retries, yet the description says it will?

wallnerryan:
@wallrj few nits, and one question about `IBlockDeviceAPI.close()` otherwise LGTM. Address my questions/comments then feel free to merge when happy and tests pass.

flocker/node/agents/functional/test_cinder.py

flocker/node/agents/functional/test_cinder.py
wallrj: Only that. The reason I have to reach for such a deeply nested attribute is because we've got a bunch of API wrappers that:
 * Add logging
 * Adapt for differences between the Cinder v1 client code and Cinder v2 client.
 * Perform lazy Keystone authentication and session initialization.

I'll leave it like this for now.

flocker/node/agents/functional/test_cinder.py
wallrj: Yep, it looks like a typo or a test that I intended to implement and then abandoned.
I've renamed it.

PR #2836, opened 2016-06-17
[FLOC-4450] Cancel ping timeout action when AMP connection is lost

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4450

Stacked on #2833 (review that one first)

I refactored the tests a little. It seemed cleaner to have all the connection ping tests together.

I'll trigger the acceptance tests and hopefully we won't see any...

```
AttributeError: 'NoneType' object has no attribute 'abortConnection' after agent disconnects from control service
```

....messages in the remote_logs.log

PR #2898, opened 2016-09-09
[FLOC-4494] Catch a circular call to cancel the AMP protocol inactivity timeout

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4494

What seems to happen is that if a blocking agent operation takes too long, the AMP connection pings are not responded to this means that the unresponsive peer timeout is reached and the connection is aborted.
And in aborting the connection we attempt to cancel the timer....which has already fired.

Putting this up for review now, to see if it fixes the ubuntu 14.04 errors and meanwhile I'll see if I can figure out a unit test for this.

This may have been caused by #2836

### DIFF
--- a/flocker/control/_protocol.py
+++ b/flocker/control/_protocol.py
@@ -57,6 +57,7 @@ from twisted.protocols.amp import (
     Argument, Command, Integer, CommandLocator, AMP, Unicode,
     MAX_VALUE_LENGTH,
 )
+from twisted.internet.error import AlreadyCalled
 from twisted.internet.task import LoopingCall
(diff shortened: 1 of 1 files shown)


wallrj:
My mistake. This isn't related to the Rackspace + Ubuntu 14.04 failures.
Those are caused by https://github.com/ClusterHQ/flocker/pull/2479

I'll create a separate issue for this.

wallrj:
Wrong issue link. See https://github.com/ClusterHQ/flocker/pull/2899

PR #2901, opened 2016-09-13
[FLOC-2966] Use OpenStack metadata service or configdrive to get instance ID

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-2966

Builds upon and replaces: https://github.com/ClusterHQ/flocker/pull/2479

I've made it so that the Cinder driver will now attempt to lookup the instance ID from:
1. configdrive (if there is one and it can be mounted)
2. metadata service (if it is accessible)
3. IP address matching (intersection rather than subset)

The configdrive mechanism works on Rackspace.
Configdrive also works on devstack mitaka (although it's an optional parameter of the `nova boot` command, so isn't always going to be available)
Metadataservice works on devstack mitaka but not on Rackspace (who seem to block  / drop packets to the service IP)
The IP address matching mechanism may not be necessary and we should discuss whether to drop it.


flocker/node/agents/cinder.py
```diff
@@ -27,6 +30,8 @@
 from novaclient.client import Client as NovaClient
 from novaclient.exceptions import NotFound as NovaNotFound
 from novaclient.exceptions import ClientException as NovaClientException
+import requests
+from requests.exceptions import ConnectionError, Timeout
```
wallrj: Ignore this. It's a Python3 built-in exception.

wallrj: Ignore this. It's a Python3 built-in exception.

flocker/node/agents/cinder.py
```diff
@@ -457,37 +487,140 @@ def allocation_unit(self):
...
+
+    def _compute_instance_id_from_metadata_service(self):
+        """
+        Attempt to access instance ID from the metadata service.
+
+        http://docs.openstack.org/admin-guide/compute-networking-nova.html#metadata-service
+        """
+        try:
+            response = requests.get(
+                'http://169.254.169.254/openstack/latest/meta_data.json',
```
wallnerryan: i dont know much about the metadata service but is `latest` the version of openstack or version of metadata?

flocker/node/agents/cinder.py
```diff
@@ -457,37 +487,140 @@ def allocation_unit(self):
...
+        """
+        methods = [
+            self._compute_instance_id_from_configdrive,
+            self._compute_instance_id_from_metadata_service,
+            self._compute_instance_id_from_ipaddresses,
+        ]
+        for method in methods:
+            result = method()
+            if result is not None:
+                return result
```
wallnerryan: if configdrive or metadata service aren't available, then still theres a concern of overlapping IPs with intersection when using overlays correct? Looking at the code again it seems its goes through server by server and therefore may choose the wrong server if it is the first one with the intersection.

Maybe worth a comment or a seperate issue but I dont think its worth blocking this PR over. Recommendation would be for metadata or configdrive etc as you mention in the PR comments.

wallnerryan:
Overall looks good. I didnt see any tests for configdrive or metadata service. Is this something we could add tests for to create a fake configdrive or mock the metadata url response etc?

wallnerryan:
Possible to create a fake configdrive or mock the metadata url response to tests?

wallnerryan:
Overall looks good, please answer the above and make sure tests pass and are unrelated to the PR if they fail. I recommend adding tests for configdrive and metadata service but I will leave this up to  you given the conversation we had.

wallrj:
Replaced by #2908 and #2909

PR #2906, opened 2016-09-22
[FLOC-4497] Move run_process to a public module

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4497

I need this for running the mount and blkid commands in  #2908

We can also use this in block_device_manager which has its own, less useful sub process runner.

### DIFF
--- a/flocker/common/process.py
+++ b/flocker/common/process.py
@@ -1,0 +1,63 @@
+# Copyright ClusterHQ Inc.  See LICENSE file for details.
+
+"""
+Subprocess utilities.
+"""
+from subprocess import PIPE, STDOUT, CalledProcessError, Popen
+
+from eliot import Message, start_action
+from pyrsistent import PClass, field
+
+
+class _CalledProcessError(CalledProcessError):
+    """
+    Just like ``CalledProcessError`` except output is included in the string
+    representation.
+    """
+    def __str__(self):
+        base = super(_CalledProcessError, self).__str__()
+        lines = "\n".join("    |" + line for line in self.output.splitlines())
+        return base + " and output:\n" + lines
+
+
+class _ProcessResult(PClass):
+    """
+    The return type for ``run_process`` representing the outcome of the process
+    that was run.
+    """
+    command = field(type=list, mandatory=True)
+    output = field(type=bytes, mandatory=True)
+    status = field(type=int, mandatory=True)
+
+
+def run_process(command, *args, **kwargs):
+    """
+    Run a child process, capturing its stdout and stderr.
+
+    :param list command: An argument list to use to launch the child process.
+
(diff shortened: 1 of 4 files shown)


flocker/common/test/test_process.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(
+            _ProcessResult(
+                command=command.commandline(),
+                status=command.returncode,
+                output=command.stdout + command.stderr,
+            ),
+            result
+        )
+
+    def check_run_process_error(self, expected_returncode):
```
wallnerryan: nit: add comment under definition as to what this does.

flocker/common/test/test_process.py
```diff
@@ -0,0 +1,94 @@
...
+            "--stdout", self.stdout,
+            "--stderr", self.stderr,
+        ]
+
+
+class RunProcessTests(TestCase):
+    """
+    Tests for ``run_process``.
+    """
+    def command_for_test(self, returncode):
```
wallnerryan: nit: add comment as to what this does

flocker/common/test/sample_script.py
```diff
@@ -0,0 +1,41 @@
...
+    options = ScriptOptions()
+    options.parseOptions(sys.argv[1:])
+    sys.stdout.write(options["stdout"])
+    sys.stdout.flush()
+    sys.stderr.write(options["stderr"])
+    sys.stderr.flush()
+    returncode = options["returncode"]
+    if returncode < 0:
+        os.kill(os.getpid(), abs(returncode))
+        time.sleep(10)
```
wallnerryan: is there a reason this sleeps for ten seconds, seems long.

wallrj: I just wanted to sleep some time in order to receive the signal.
TBH I'm not sure whether os.kill blocks until the delivery of the signal or how signals are scheduled by the kernel. 
The sleep may not be necessary but I'll leave it there, reduce the time and add a comment.

flocker/common/test/test_process.py
wallrj: Done.

flocker/common/process.py
```diff
@@ -0,0 +1,63 @@
...
+
+    :param list command: An argument list to use to launch the child process.
+
+    :raise CalledProcessError: If the child process has a non-zero exit status.
+
+    :return: A ``_ProcessResult`` instance describing the result of the child
+         process.
+    """
+    kwargs["stdout"] = PIPE
+    kwargs["stderr"] = STDOUT
```
wallrj: It's an assignment to the `kwargs` dict so it shouldn't cause `KeyError`.
It'll override the `stdout` and `stderr` kwargs supplied by the caller, but that's an important behaviour of this function.

PR #2908, opened 2016-09-26
[FLOC-4498] Get compute_instance_id from metadata file on configdrive 

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4498
Stacked on: #2910
This replaces #2901

This is a better tested version of the config drive code in #2901.
This will also work on OpenStack if nodes are booted with the configdrive option.

I've run the Rackspace acceptance tests for this branch.
If my calculations are correct, we should be  able to see flocker-dataset-agent successfully mounting the ConfigDrive (config-2) and getting the instance ID from there.

This is followed by #2909 a branch that adds a mechanism to query the metadata service if the configdrive is not available.


flocker/common/_filepath.py
```diff
@@ -38,3 +42,55 @@ def make_directory(path):
...
+    An ``IFilePath`` which when used as a context manager will remove itself.
+    """
+    def __init__(self, path):
+        self._path = path
+
+    @property
+    def path(self):
+        return self._path.path
+
+    def __eq__(self, other):
```
wallrj: I don't think so....but could / should probably add a few more tests for this.
I'm delegating to the wrapped ``FilePath.__eq__`` method which will compare its `.path` byte string  (b"/foo/bar/baz") with `other.path`.

vchernoy: I am just learning the code, most of my comments will be false-positive.

flocker/node/agents/cinder.py
```diff
@@ -484,9 +544,14 @@ def allocation_unit(self):
 
     def compute_instance_id(self):
         """
-        Find the ``ACTIVE`` Nova API server with a subset of the IPv4 and IPv6
-        addresses on this node.
+        Attempt to retrieve node UUID from the metadata in a config drive.
+        Fall back to finging the ``ACTIVE`` Nova API server with an
```
vchernoy: finging => finding

flocker/common/_filepath.py
```diff
@@ -38,3 +42,55 @@ def make_directory(path):
     if not path.isdir():
         path.makedirs()
     return path
+
+
+class IFilePathExtended(IFilePath):
+    """
+    ``IFilePath`` plus some other ``FilePath`` methods that we need in Flocker.
+    """
+    def remove():
```
wallrj: It's a `zope.interface` thing. 
TBH I never really understood why until now.
The docs have this to say:
> Interfaces document how an object is used. When calling instance methods, you don’t pass a self argument, so a self argument isn’t included in the interface signature. The self argument in instance methods is really an implementation detail of Python instances. Other objects, besides instances can provide interfaces and their methods might not be instance methods. For example, modules can provide interfaces and their methods are usually just functions. Even instances can have methods that are not instance methods.

https://docs.zope.org/zope.interface/README.html

wallrj:
Thanks for the review @vchernoy 

Your comments made me realise that I'd  overcomplicated things.

I've reorganised so that ``mount`` and ``temporary_mount`` can be used for mounting filesystems by label or by device path.
Also removed the unnecessary ``config_drive`` context manager and added more reliable cleanup of loop devices and mounts when tests fail.

I'll merge this if If all the tests pass.

vchernoy:
All this Python and Twisted look to me overcomplicated (over abstracted). Too many (abstract) classes, interfaces and hierarchies. I feel like it is Java style but harder to understand because it has everything-dynamic.

IMHO, having an interface and abstract class for FilePath is over kill, unless I miss something, which is very likely.

Will try to make next review quicker.

flocker/node/agents/cinder.py
```diff
@@ -60,6 +64,51 @@
...
+            content = metadata_file.getContent()
+        except IOError as e:
+            Message.new(
+                message_type=(
+                    u"flocker:node:agents:blockdevice:openstack:"
+                    u"compute_instance_id:metadata_file_not_found"),
+                error_message=unicode(e),
+            ).write()
+            return
+        try:
```
vchernoy: probably you can merge the two try-statements.

flocker/node/agents/cinder.py
```diff
@@ -60,6 +64,51 @@
...
+            return json.loads(content)
+        except ValueError as e:
+            Message.new(
+                message_type=(
+                    u"flocker:node:agents:blockdevice:openstack:"
+                    u"compute_instance_id:metadata_file_not_json"),
+                error_message=unicode(e),
+            ).write()
+            return
```
vchernoy: I guess return is for extra-safety.

PR #2910, opened 2016-09-27
[FLOC-4501] Skip upgrade tests on Openstack backends

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4501

``test_upgrade`` downgrades to a version of Flocker where `compute_instance_id` fails because it can't deal with floating IP addresses on  Rackspace + Ubuntu 14.04.

The test times out and Flocker is never upgraded again and all remaining tests time out.

Once we've released 1.16.0 we can re-enable this test and also remove the skip for Ubuntu 16.04

### DIFF
--- a/flocker/acceptance/endtoend/test_dataset.py
+++ b/flocker/acceptance/endtoend/test_dataset.py
@@ -102,9 +102,14 @@ class DatasetAPITests(AsyncTestCase):
         ),
     )
     @skip_backend(
-        unsupported={backends.LOOPBACK},
-        reason="Does not maintain compute_instance_id across restarting "
-               "flocker (and didn't as of most recent release).")
+        unsupported={backends.LOOPBACK, backends.OPENSTACK},
+        reason=(
+            "Loopback backend does not maintain compute_instance_id across "
+            "restarting flocker (and didn't as of most recent release). "
+            "OpenStack backend (as of 1.15.0) cannot reliably identify "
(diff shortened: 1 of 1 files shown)


flocker/acceptance/endtoend/test_dataset.py
```diff
@@ -102,9 +102,14 @@ def _get_package_source(self, default_version=None):
...
     @skip_backend(
-        unsupported={backends.LOOPBACK},
-        reason="Does not maintain compute_instance_id across restarting "
-               "flocker (and didn't as of most recent release).")
+        unsupported={backends.LOOPBACK, backends.OPENSTACK},
+        reason=(
+            "Loopback backend does not maintain compute_instance_id across "
+            "restarting flocker (and didn't as of most recent release). "
+            "OpenStack backend (as of 1.15.0) cannot reliably identify "
+            "the compute node if it has floating IP addresses."
```
wallnerryan: nit: maybe worth putting a link the the text to where we track the bug? Or to a place where there is more detail to someone who comes long after. up to you :)

wallrj: I've created an issue to re-enable the tests after 1.16.0 which is blocked by the 1.16 release issue: 
 * https://clusterhq.atlassian.net/browse/FLOC-2966

I'll add a link in a followup branch.

PR #2912, opened 2016-09-29
[FLOC-4504] Update to latest versions of Python dependencies and transient  dependencies

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4504

Also fixed the `update-requirments` tool to be based off of alpine linux which is both a smaller image and which has a modern openssl library against which we can install latest version of cryptography.

There's an ongoing effort to create cryptography wheel files, here:
 * https://github.com/reaperhulk/cryptography-manylinux1


.pylintrc
```diff
@@ -74,9 +74,6 @@ enable=import-error,
        global-at-module-level,
        bad-open-mode,
        redundant-unittest-assert,
-       # Has common issues with our style due to
-       # https://github.com/PyCQA/pylint/issues/210
-       unused-variable,
```
wallrj: The new version of pylint seems to be better at finding unused variables and I didn't want to fix them all in this branch.

wallnerryan: :)

admin/requirements.Dockerfile
```diff
@@ -6,16 +6,17 @@
 # temporary directory containing a copy of Flocker/requirements and the
 # entrypoint script.
 
-FROM quay.io/pypa/manylinux1_x86_64:latest
+FROM alpine:latest
```
wallnerryan: cool, def in favor of this.

admin/requirements.py
```diff
@@ -14,7 +14,7 @@
 from twisted.python.usage import Options, UsageError
 from twisted.python.filepath import FilePath
 
-MANYLINUX_IMAGE = "quay.io/pypa/manylinux1_x86_64:latest"
+ALPINE_IMAGE = "alpine:latest"
```
wallnerryan: i am always skeptical about using `:latest` for the reason that src code will randomly change to use a new version of a container if it comes out on quay/dockerhub. `latest` oftem means the `most recently pushed`. So as long as we know this shouldn't introduce any issues im OK with it, but figured i'd put the point here.

requirements/admin.txt
```diff
@@ -1,67 +1,68 @@
...
-awscli==1.10.44
-bitmath==1.3.0.2
-boto==2.41.0
+awscli==1.10.67
+bitmath==1.3.1.2
+boto==2.42.0
 characteristic==14.3.0
 deb-pkg-tools==1.36
-docker-py==1.8.1
+docker-py==1.10.3
```
wallnerryan: this probably brings some major changes. E.g. supporting `swarm-mode` which may be useful in testing things in the future with docker api instead of container api.

https://github.com/docker/docker-py/blob/master/docs/change_log.md#1103

admin/requirements.py
wallrj: :+1: I agree but I'll leave it for now....to save having to re-run all the tests :frowning:

requirements/admin.txt
wallrj: Yeah. I'm quite surprised that all our tests still pass.

PR #2915, opened 2016-10-03
[FLOC-4507] Disallow installation of pycparser wheel file. 

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4507
Workaround for: https://github.com/eliben/pycparser/issues/148

This is a quick fix.

Long term we should:
 * Include hashes in our requirements files - so that pip wouldn't even attempt to install the broken file.
 * Monitor: https://github.com/pypa/pip/issues/3996 which would make pip not attempt to install the wheel file if the hash was for an sdist.
 * Have our own wheelhouse of verified dependencies and only install from there.


wallrj:
This has been fixed upstream by removing the offending Wheel file.

PR #2919, opened 2016-10-04
[FLOC-4411] Use the latest version of oauth2client

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4411

Our update-requirements tool dragged in a few other dependencies that have been updated since last week  when we last ran the tool.

PR #2928, opened 2016-10-11
[FLOC-4512] Fix the tox -e lint command and use it in Jenkins

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4512

This goes beyond the Jira issue as follows:
- updated various dependencies
- tidied up the tox config file
- Speeds up tox setup by not creating an sdist
- Fixes a bug where tox couldn't be run multiple times because its call to python setup.py develop conflicted with the `pip install --editable .` in our dev-requiremnts.txt file.


wallrj:
I want to get this merged so that we don't fall too far behind on dependencies and so that I can update #2919.
So I did a self review of this branch :slightly_frowning_face:  
LGTM :slightly_smiling_face:

PR #2936, opened 2016-11-03
[FLOC-4523] Add FlockerHub menu links

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4523

I couldn't easily  get the new drop down menus working so opted to stick with the current menu style.


ferrantim:
In the interest of expediency, let's just use 
FlockerHub
Fli
Flocker 
Blog

as the menu options for now.

We can figure out the drop down later.  Mainly want to get rid of VolumeHub and dvol for now.  Thanks!

PR #2944, opened 2016-11-08
Remove zendesk

ferrantim:


wallnerryan:
@wallrj this is a blocker for having intercom on fli-docs and flockerhub-docs which already have this change, do you think we can expedite this?

wallrj:
Thanks @ferrantim. Closing in favour of #2945 which merges into the release branch.

PR #2946, opened 2016-11-09
[FLOC-4530] Pass the swappiness configuration value to the Docker API

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4530

In https://github.com/ClusterHQ/flocker/pull/2818 we added swappiness parameter to the container API but the implementation wasn't complete.
The swappiness was not passed through to the Docker API.
And the swappiness was not inspected by the ApplicationDeployer.

This branch adds that missing piece and a functional test that demonstrates that swappiness is set and discovered.

### DIFF
--- a/flocker/node/_docker.py
+++ b/flocker/node/_docker.py
@@ -495,7 +495,7 @@ class DockerClient(object):
             long_timeout=600):
         self.namespace = namespace
         self._client = dockerpy_client(
-            version="1.15", base_url=base_url,
+            version="1.20", base_url=base_url,
             long_timeout=timedelta(seconds=long_timeout),
         )
         self._image_cache = LRUCache(100)
@@ -717,6 +717,7 @@ class DockerClient(object):
                 binds=binds,
                 port_bindings=port_bindings,
                 restart_policy=restart_policy_dict,
+                mem_swappiness=swappiness,
             )
             # We're likely to get e.g. pvector, so make sure we're passing
             # in something JSON serializable:
@@ -971,9 +972,9 @@ class DockerClient(object):
                 # mem_limit in containers without specified limits, however
                 # Docker returns the values in these cases as zero, so we
                 # manually convert.
-                cpu_shares = data[u"Config"][u"CpuShares"]
+                cpu_shares = data[u"HostConfig"][u"CpuShares"]
                 cpu_shares = None if cpu_shares == 0 else cpu_shares
-                mem_limit = data[u"Config"][u"Memory"]
+                mem_limit = data[u"HostConfig"][u"Memory"]
                 mem_limit = None if mem_limit == 0 else mem_limit
                 restart_policy = self._parse_restart_policy(
                     data[U"HostConfig"][u"RestartPolicy"])
@@ -988,7 +989,8 @@ class DockerClient(object):
                     mem_limit=mem_limit,
                     cpu_shares=cpu_shares,
                     restart_policy=restart_policy,
-                    command_line=command)
+                    command_line=command,
+                    swappiness=data[u"HostConfig"][u"MemorySwappiness"])
                 )
             return result
         return deferToThread(_list)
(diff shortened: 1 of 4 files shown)


flocker/node/_docker.py
```diff
@@ -976,9 +972,9 @@ def _list():
                 # mem_limit in containers without specified limits, however
                 # Docker returns the values in these cases as zero, so we
                 # manually convert.
-                cpu_shares = data[u"Config"][u"CpuShares"]
+                cpu_shares = data[u"HostConfig"][u"CpuShares"]
                 cpu_shares = None if cpu_shares == 0 else cpu_shares
-                mem_limit = data[u"Config"][u"Memory"]
+                mem_limit = data[u"HostConfig"][u"Memory"]
                 mem_limit = None if mem_limit == 0 else mem_limit
```
wallnerryan: Do we need to do a similar thing for swappiness here?

flocker/node/test/test_container.py
```diff
@@ -460,6 +460,34 @@ def test_command_line(self):
...
+        ``DockerClient.add`` which is used when creating a Unit.
+        """
+        expected_swappiness = 99
+        fake_docker = FakeDockerClient()
+        deployer = ApplicationNodeDeployer(u'example.com', fake_docker)
+
+        application_name = u'site-example.com'
+        application = Application(
+            name=application_name,
+            image=DockerImage(repository=u'clusterhq/postgresql',
```
wallnerryan: Why dont we use a smaller image here? Any reason we are using postgresql just to test the swappiness? Could mean longer pulls for new test nodes.

wallrj:
In https://github.com/ClusterHQ/flocker/pull/2946#discussion_r87477767 @wallnerryan wrote:
> Do we need to do a similar thing for swappiness here?

I've been thinking about it.
* The intention (I think) was that Flocker would impose a default swappiness of 0 (do not swap)....that's the default in the container API if a swappiness is not supplied.
* And there's a configuration upgrade script which will set the swappiness of all existing containers to 0.
* With this change, `DockerClient` will return  `Unit(..., swappiness = -1,...)` (the value returned by `docker inspect` when the swappiness has not been defined.
* That should be ok, since `Unit` defines swappiness as an integer type....it doesn't enforce postive integers.
* The deployer will turn that into an `Application(..., swappiness = -1,...)`... which is also allowed
* That means that all discovered containers will be found to have a different swappiness value than configured...
* Which means that *all containers will be restarted* when the upgraded flocker-container-agent is started!

Which might very well be a problem; ie unscheduled restart and all containers restarting at once may very well overload the server.

I guess I should go back to the drawing board and make the default swappiness None :cry:

wallnerryan:
@wallrj 

The intention was for default to be `0` but im not sure that is the best way, that makes swisscom happy but i guess container agent will never be used by anyone but them in the short term.

My vote would not to over complicate it and just set the default to 0, but make it configurable if they want/need it. Swappiness should not be set to -1, but rather 0 be default and placed into the HostConfig for the DockerClient. That way, in thoery, the restart issue should only happen if they containers swappiness is changed out of band, which im not sure if possible, but rather any changes would need to happen in the deployment yaml.

I _think_ strongly forcing `0` by default should work, and since this this is really only impactful in the short term for swisscom, I think this may be ok. Thoughts?

wallrj:
I've suggested (as a work around) changing the memory.swappiness value in the parent dockerd namespace.
I'll see if that is sufficient to address the problem and come back to this if not.

```
# View the swappiness setting of the docker process namespace
# cat  /sys/fs/cgroup/memory/docker/memory.swappiness
60

# Containers inherit the swappiness from docker namespace
# docker run --rm busybox cat /sys/fs/cgroup/memory/memory.swappiness
60

# Change the docker daemon swappiness
# echo 0 >  /sys/fs/cgroup/memory/docker/memory.swappiness

# Containers now have the new swappiness value by default.
# docker run --rm busybox cat /sys/fs/cgroup/memory/memory.swappiness
0
```

PR #2957, opened 2016-11-22
[FLOC-4537] build docker image for flocker control

vchernoy:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4537

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -561,6 +561,22 @@ common_cli:
         dockerfiles/dataset
     JOB_EXIT_STATUS="$( updateExitStatus $? )"
 
+
+  build_docker_image_flocker_control: &build_docker_image_flocker_control |
+    # Build Flocker Docker image
+    BUILD_SERVER_IP="$(wget -qO- http://instance-data/latest/meta-data/public-ipv4)"
+    REPOSITORY_PATH="results/omnibus/${TRIGGERED_BRANCH}/${DISTRIBUTION_NAME}"
+    docker build \
+        --rm \
+        --tag "${DOCKER_IMAGE}" \
+        --label "FLOCKER_VERSION=${FLOCKER_VERSION}" \
+        --label "GIT_COMMIT=${GIT_COMMIT}" \
+        --label "TRIGGERED_BRANCH=${TRIGGERED_BRANCH}" \
+        --label "BUILD_SERVER_IP=${BUILD_SERVER_IP}" \
+        --build-arg "FLOCKER_REPOSITORY=http://${BUILD_SERVER_IP}/${REPOSITORY_PATH}" \
+        dockerfiles/control
+    JOB_EXIT_STATUS="$( updateExitStatus $? )"
+
   # These are the docker images we will be using during our tests.
   # We build them every 24 hours, making sure we have the latest OS updates
   # installed on those images.
@@ -1175,6 +1191,9 @@ job_type:
                    *build_repo_metadata,
                    'export DOCKER_IMAGE=clusterhqci/flocker-dataset-agent:${GIT_COMMIT}',
                    *build_docker_image_flocker,
+                   *push_image_to_dockerhub,
+                   'export DOCKER_IMAGE=clusterhqci/flocker-control:${GIT_COMMIT}',
+                   *build_docker_image_flocker_control,
                    *push_image_to_dockerhub,
                    *clean_packages,
                    *exit_with_return_code_from_test ]
--- a/dockerfiles/control/Dockerfile
+++ b/dockerfiles/control/Dockerfile
@@ -18,7 +18,10 @@ RUN \
         clusterhq-python-flocker=${FLOCKER_VERSION} \
         clusterhq-flocker-node=${FLOCKER_VERSION}
 
+VOLUME /etc/flocker
+VOLUME /var/lib/flocker
+
 EXPOSE 4523
 EXPOSE 4524
 
-ENTRYPOINT ["/usr/sbin/flocker-control", "--port=tcp:4523", "--agent-port=tcp:4524", "--journald", "--certificates-directory=/etc/flocker", "--data-path=/var/lib/flocker"] 
+ENTRYPOINT ["/usr/sbin/flocker-control", "--port=tcp:4523", "--agent-port=tcp:4524", "--certificates-directory=/etc/flocker", "--data-path=/var/lib/flocker"] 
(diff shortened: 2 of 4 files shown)


dockerfiles/control/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+ARG FLOCKER_REPOSITORY="https://clusterhq-archive.s3.amazonaws.com/ubuntu/16.04/amd64"
+RUN echo "deb ${FLOCKER_REPOSITORY} /" > /etc/apt/sources.list.d/clusterhq.list
+RUN \
+    apt-get --yes update \
+    && apt-get --yes install --no-install-recommends --allow-unauthenticated \
+        clusterhq-python-flocker=${FLOCKER_VERSION} \
+        clusterhq-flocker-node=${FLOCKER_VERSION}
+
+EXPOSE 4523
+EXPOSE 4524
```
wallrj: :question: I wonder if we should also declare the `VOLUME`s that we expect to be mounted? I've never really understood the point of it, since you can override any directory inside the container with `docker run --volume ...` . Might be worth looking into: https://docs.docker.com/engine/reference/builder/#/volume

docs/gettinginvolved/infrastructure/packaging.rst
```diff
@@ -146,3 +149,35 @@ To check the image, run the container with the argument ```--version```:
...
+        --privileged \
+        --volume /flocker:/flocker:shared \
+        --volume /etc/flocker:/etc/flocker \
+        --volume /dev:/dev \
+        --detach \
+        clusterhqci/flocker-dataset-agent:master
+
+
+flocker-control
+---------------
```
wallrj: Add a note about how the `flocker-control` Docker image is built using the same `docker build ...` command line as for `flocker-dataset` but substituting the `control/Dockerfile`.

wallrj:
I think you mentioned before that the Dockerfile `MAINTAINER` directive has been deprecated. I can't find anything about that, but I did find https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/label which has guidlines about what labels to include. And I notice that the Kubernetes Docker images have flags like this.
Perhaps we should update our labels....not here....in a followup branch.

wallrj:
Also noticed this warning in the build output:
```
09:35:40 Flag --email has been deprecated, will be removed in 1.13.
09:35:44 Login Succeeded
09:35:44 The push refers to a repository [docker.io/clusterhqci/flocker-dataset-agent]
```
-- http://ci-live.clusterhq.com:8080/job/ClusterHQ-flocker/job/docker-image--flocker-control-FLOC-4537/job/run_docker_build_on_Ubuntu_Xenial/1/console

So perhaps  here (or in a followup branch) we should update the `docker login ...` command in `build.yaml`.
 * https://github.com/ClusterHQ/flocker/blob/master/build.yaml#L539

And as a followup perhaps we can start looking into how we can perform the Docker image build from Travis-CI...:
 * https://docs.travis-ci.com/user/docker/#Pushing-a-Docker-Image-to-a-Registry

Or maybe use Dockerhub automated builds:
 * https://docs.docker.com/docker-hub/builds/

Or Quay.io
 * https://docs.quay.io/guides/building.html

We should chat to @moshez about the ideal set of build artifacts. I'm thinking it could be as simple as e.g.
1. Docker image
2. RPMs and Debs with systemd and firewalld configuration for running Flocker Docker images....or maybe via Rocket / Systemd-nspawn for people who want to install and run Flocker themselves
3.  Kubernetes Helm chart for people who want Flocker to be managed by Kubernetes.
...and make a plan for a build pipeline. 
Something like:
 1. `python setup.py sdist`
 2. `python setup.py wheel`
 3. `pex build --requirement flocker/requirements/... 
 4.  `docker build ...`
 5. `rpmbuild ...` (includes systemd and firewalld service files that start and manage the Docker image on Centos-7)
 6. `Kubernetes Helm charts`
 7. etc

build.yaml
```diff
@@ -1176,6 +1192,9 @@ job_type:
                    'export DOCKER_IMAGE=clusterhqci/flocker-dataset-agent:${GIT_COMMIT}',
                    *build_docker_image_flocker,
                    *push_image_to_dockerhub,
+                   'export DOCKER_IMAGE=clusterhqci/flocker-control:${GIT_COMMIT}',
+                   *build_docker_image_flocker_control,
+                   *push_image_to_dockerhub,
```
vchernoy: right, they both contain the same data and share the same layers.

dockerfiles/control/Dockerfile
vchernoy: a... right, mapping and exposing ports and mounting volumes is done in CLI. The reason we can add this in the Dockerfile is for declaration purpose. So we probably want to do that for both images.

dockerfiles/control/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+
+ENTRYPOINT ["/usr/sbin/flocker-control", "--port=tcp:4523", "--agent-port=tcp:4524", "--journald", "--certificates-directory=/etc/flocker", "--data-path=/var/lib/flocker"]
```
vchernoy: What I tried is: 
1. I took dataset docker image and run a new container overriding the entry point and providing port mapping and mount points.
2. Then checked that it works by doing round-trip tests (moving volumes back and forth).
3. And finally I derived the Dockerfile from the docker-run parameters.

So I planed to check the Dockerfile when Jenkins will build the image, but as you pointed, it does not do it automatically, which explains why I could not find (docker-pull) it.

vchernoy: I checked the image, it seems it works but I cannot explain yet why we do not see logs, will look into it.

wallrj: Just remove the ``--journald`` argument and it will log to stdout. See `flocker-control --help`

vchernoy:
see [deprecated maintainer](https://github.com/docker/docker/blob/master/docs/deprecated.md#maintainer-in-dockerfile)

docs/gettinginvolved/infrastructure/packaging.rst
vchernoy: addded

dockerfiles/control/Dockerfile
vchernoy: removed

PR #2964, opened 2016-11-29
[FLOC-4543] Break apart the docs checks and run them from travis-ci

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4543

I've made the following changes:
 * builds are now performed by a script in the repository -- my intention is that we'll be able to extend this to perform more complex builds for things which we don't want to configure in tox.
 * I've added a complimentary `after_script` helper script which uploads the generated HTML docs to s3://clusterhq-staging-docs
 * I've copied the S3 credentials from https://github.com/ClusterHQ/segredos/blob/master/ci-platform/all/all.yaml#L141 into the Travis-CI UI as encrypted environment variables. These should only be made available to builds triggered from ClusterHQ/flocker branches and PRs (not for PRs for Flocker forks) See https://docs.travis-ci.com/user/environment-variables/#Defining-Variables-in-Repository-Settings
 * I've deleted all the documentation related build steps from the Jenkins build.yml file.

I've done this to get ideas about how we can perform more complicated build steps from Travis.

Next I'd like to try launching `run-acceptance-tests` and the cloudformation tests from Travis since that's the build step that is so unreliable.


vchernoy:
Great step!

PR #2965, opened 2016-11-29
[FLOC-4534] Install Kubernetes Flocker plugin

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4534
Stacked on: #2932

This is a first draft rather than the finished product.


admin/acceptance.py
```diff
@@ -1455,6 +1463,7 @@ def main(reactor, args, base_path, top_level):
     reached_finally = False
 
     def cluster_cleanup():
+        return
```
vchernoy: dead code?

PR #2971, opened 2016-12-07
[FLOC-4549] Update Python requirements

wallrj:
Fixes: https://clusterhq.atlassian.net/browse/FLOC-4549

### DIFF
--- a/admin/packaging.py
+++ b/admin/packaging.py
@@ -606,6 +606,11 @@ IGNORED_WARNINGS = {
         # E.g.
         # /opt/flocker/lib/python2.7/site-packages/sphinx/locale/bn/LC_MESSAGES/sphinx.mo
         'file-not-in-%lang',
+
+        # Twisted 16.6 includes an executable C source file.
+        # https://twistedmatrix.com/trac/ticket/8921
+        'spurious-executable-perm /opt/flocker/lib/python2.7/site-packages/twisted/internet/iocpreactor/iocpsupport/iocpsupport.c', # noqa
+
     ),
 # See https://www.debian.org/doc/manuals/developers-reference/tools.html#lintian  # noqa
     PackageTypes.DEB: (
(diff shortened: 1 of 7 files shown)


admin/packaging.py
```diff
@@ -606,6 +606,11 @@ def __str__(self):
         # E.g.
         # /opt/flocker/lib/python2.7/site-packages/sphinx/locale/bn/LC_MESSAGES/sphinx.mo
         'file-not-in-%lang',
+
+        # Twisted 16.6 includes an executable C source file.
```
wallnerryan: What do we use this for? Im not familiar with iocsupport.c so that would help in the comments for other readers.

Assuming we build this before and changing this to use a binary?

setup.cfg
```diff
@@ -12,6 +12,6 @@ parentdir_prefix = Flocker-
 
 [flake8]
 exclude=flocker/_version.py,flocker/restapi/docs/hidden_code_block.py
-# Ignore defaults + E731
+# Ignore defaults + E731, E305
 # http://pep8.readthedocs.io/en/latest/intro.html#error-codes
-ignore=E123,E226,E241,E731,W503
+ignore=E123,E226,E241,E731,E305,W503
```
wallnerryan: 305 is "expected 2 blank lines after end of function or class"

Is there a reason we dont want this anymore? I'm indifferent, just curious.

admin/packaging.py
wallrj: It's part of Twisted's MS Windows event loop.
Nothing to do with Flocker, but it gets installed when we install the Twisted package.
And unfortunately triggers our rpmlint packaging checks.

wallrj:
Thanks for the review @wallnerryan 

> What's the driving factors that made us do an overall update of deps?

1. To keep Flocker up-to-date and secure
2. The requirements files are written by a tool I wrote, `admin/update-requirements` which doesn't allow you to update a single requirment.
3. I wanted to add a new dependency (Alchimia) in #2969 but didn't want to run `admin/update-requirements` there and add a load of unrelated package version changes in that branch.

Once this is merged, I can merge the stacked branches forward so that they have sane changesets.

PR #2973, opened 2016-12-07
Harden flocker-cluster-id EBS parsing

sgnn7:
If there is a volume in EBS with a tag key of `flocker-cluster-id` but
the value is not a valid UUID, any Flocker agent connected to that
account will throw an exception during iteration and will be unusable.

The change here catches bad values and makes them process as just
another volume that's not part of the asking cluster rather than
throwing the mentioned exceptions.


wallrj:
Thanks @sgnn7 

I'll write a test for this to ensure that it stays fixed and merge this into a ClusterHQ/flocker branch so that we can run the full set of tests.

wallrj:
Moved to https://github.com/ClusterHQ/flocker/pull/2974

sgnn7:
:+1:

